PR #2, opened 2023-03-16
ImmutAfterInit: misuse runtime detection

00xc:
* Use `SyncUnsafeCell` instead of `UnsafeCell` + `unsafe impl Sync`.
* Pass Copy types by value in `ImmutAfterInitCell` methods.
* Remove a double initialization before implementing runtime detection of this condition.
* Back `ImmutAfterInitCell` by an `Option` in order to be able to detect uninitialized cells.
* Return `Result` from public API calls to handle double initialization and uninitialized access.

### DIFF
--- a/src/utils/immut_after_init.rs
+++ b/src/utils/immut_after_init.rs
@@ -12,6 +12,13 @@ use core::ops::Deref;
 #[cfg(debug_assertions)]
 use core::sync::atomic::{AtomicBool, Ordering};
 
+#[cfg(not(debug_assertions))]
+pub type ImmutAfterInitResult<T> = Result<T, core::convert::Infallible>;
+
+#[cfg(debug_assertions)]
+pub type ImmutAfterInitResult<T> = Result<T, ImmutAfterInitError>;
+
+#[cfg(debug_assertions)]
 #[derive(Clone, Copy, Debug)]
 pub enum ImmutAfterInitError {
     AlreadyInit,
@@ -85,7 +92,7 @@ impl<T: Copy> ImmutAfterInitCell<T> {
         self.init.store(true, Ordering::Relaxed);
     }
 
-    fn check_init(&self) -> Result<(), ImmutAfterInitError> {
+    fn check_init(&self) -> ImmutAfterInitResult<()> {
         #[cfg(debug_assertions)]
         if !self.init.load(Ordering::Relaxed) {
             return Err(ImmutAfterInitError::Uninitialized);
@@ -93,7 +100,7 @@ impl<T: Copy> ImmutAfterInitCell<T> {
         Ok(())
     }
 
-    fn check_uninit(&self) -> Result<(), ImmutAfterInitError> {
+    fn check_uninit(&self) -> ImmutAfterInitResult<()> {
         #[cfg(debug_assertions)]
         if self.init.load(Ordering::Relaxed) {
             return Err(ImmutAfterInitError::AlreadyInit);
@@ -114,7 +121,7 @@ impl<T: Copy> ImmutAfterInitCell<T> {
         (*self.data.get()).assume_init_ref()
     }
 
-    fn try_get_inner(&self) -> Result<&T, ImmutAfterInitError> {
+    fn try_get_inner(&self) -> ImmutAfterInitResult<&T> {
         self.check_init()?;
         unsafe { Ok(self.get_inner()) }
     }
@@ -124,7 +131,7 @@ impl<T: Copy> ImmutAfterInitCell<T> {
     /// Must **not** get called on an already initialized instance!
     ///
     /// * `v` - Initialization value.
-    pub fn init(&self, v: &T) -> Result<(), ImmutAfterInitError> {
+    pub fn init(&self, v: &T) -> ImmutAfterInitResult<()> {
         self.check_uninit()?;
         unsafe { self.set_inner(v) };
         Ok(())
@@ -261,7 +268,7 @@ impl<'a, T> ImmutAfterInitRef<'a, T> {
     /// * `r` - Reference to the value to make the `ImmutAfterInitRef` to refer
     ///         to. By convention, the referenced value must have been
     ///         initialized already.
-    pub fn init_from_ref<'b>(&self, r: &'b T) -> Result<(), ImmutAfterInitError>
+    pub fn init_from_ref<'b>(&self, r: &'b T) -> ImmutAfterInitResult<()>
     where
         'b: 'a,
     {
@@ -298,10 +305,7 @@ impl<'a, T: Copy> ImmutAfterInitRef<'a, T> {
     /// * `cell` - The value to make the `ImmutAfterInitRef` to refer to. By
     ///            convention, the referenced value must have been initialized
     ///            already.
-    pub fn init_from_cell<'b>(
-        &self,
-        cell: &'b ImmutAfterInitCell<T>,
-    ) -> Result<(), ImmutAfterInitError>
+    pub fn init_from_cell<'b>(&self, cell: &'b ImmutAfterInitCell<T>) -> ImmutAfterInitResult<()>
     where
         'b: 'a,
     {
--- a/src/utils/immut_after_init.rs
+++ b/src/utils/immut_after_init.rs
@@ -67,7 +67,7 @@ pub enum ImmutAfterInitError {
 /// }
 /// ```
 ///
-pub struct ImmutAfterInitCell<T: Copy> {
+pub struct ImmutAfterInitCell<T> {
     #[doc(hidden)]
     data: UnsafeCell<MaybeUninit<T>>,
     // Used to keep track of the initialization state. Even though this
@@ -76,7 +76,7 @@ pub struct ImmutAfterInitCell<T: Copy> {
     init: AtomicBool,
 }
 
-impl<T: Copy> ImmutAfterInitCell<T> {
+impl<T> ImmutAfterInitCell<T> {
     /// Create an unitialized `ImmutAfterInitCell` instance. The value must get
     /// initialized by means of [`Self::init()`] before first usage.
     pub const fn uninit() -> Self {
@@ -159,7 +159,7 @@ impl<T: Copy> ImmutAfterInitCell<T> {
     }
 }
 
-impl<T: Copy> Deref for ImmutAfterInitCell<T> {
+impl<T> Deref for ImmutAfterInitCell<T> {
     type Target = T;
 
     /// Dereference the wrapped value. Must **only ever** get called on an
@@ -169,8 +169,8 @@ impl<T: Copy> Deref for ImmutAfterInitCell<T> {
     }
 }
 
-unsafe impl<T: Copy> Send for ImmutAfterInitCell<T> {}
-unsafe impl<T: Copy> Sync for ImmutAfterInitCell<T> {}
+unsafe impl<T> Send for ImmutAfterInitCell<T> {}
+unsafe impl<T> Sync for ImmutAfterInitCell<T> {}
 
 /// A reference to a memory location which is effectively immutable after
 /// initalization code has run.
@@ -296,7 +296,7 @@ impl<'a, T> ImmutAfterInitRef<'a, T> {
     }
 }
 
-impl<'a, T: Copy> ImmutAfterInitRef<'a, T> {
+impl<'a, T> ImmutAfterInitRef<'a, T> {
     /// Initialize an uninitialized `ImmutAfterInitRef` instance to point to
     /// value wrapped in a [`ImmutAfterInitCell`].
     ///
--- a/src/utils/immut_after_init.rs
+++ b/src/utils/immut_after_init.rs
@@ -9,6 +9,14 @@ use core::marker::Copy;
 use core::marker::PhantomData;
 use core::mem::MaybeUninit;
 use core::ops::Deref;
+#[cfg(debug_assertions)]
+use core::sync::atomic::{AtomicBool, Ordering};
+
+#[derive(Clone, Copy, Debug)]
+pub enum ImmutAfterInitError {
+    AlreadyInit,
+    Uninitialized,
+}
 
 /// A memory location which is effectively immutable after initalization code
 /// has run.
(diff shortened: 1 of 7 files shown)


src/utils/immut_after_init.rs
```diff
@@ -102,7 +118,7 @@ impl<T: Copy> Deref for ImmutAfterInitCell<T> {
     /// Dereference the wrapped value. Must **only ever** get called on an
     /// initialized instance!
     fn deref(&self) -> &T {
-        unsafe { (&*self.data.get()).assume_init_ref() }
+        unsafe { self.as_ref() }.unwrap()
```
nicstange, line 128: Would it perhaps make sense to restrict this misuse tracking to debug builds only?
Because there are some values wrapped in `ImmutAfterInitCell<>` whose accesses might be performance-sensitive, like `ENCRYPT_MASK`.

00xc: > Would it perhaps make sense to restrict this misuse tracking to debug builds only? Because there are some values wrapped in `ImmutAfterInitCell<>` whose accesses might be performance-sensitive, like `ENCRYPT_MASK`.

We could use [`Option::unwrap_unchecked()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_unchecked), and use debug assertions (as explained [here](https://stackoverflow.com/a/39205417)) to conditionally compile one or another.

00xc:
> FWIW, I have a very similar, conflicting commit as your [a7da78a](https://github.com/coconut-svsm/svsm/pull/2/commits/a7da78ad1792066d6de2ae26af3989b2161ab614) in my pending ELF PR: [e2acf29](https://github.com/coconut-svsm/svsm/commit/e2acf29c58ab39fb516e046ee76ac5e13ed44e08)
> 
> The reasoning for having the initializer functions to take references rather than values had been to not rely on compiler optimizations too much when initializing large objects like the SnpCpuidTable blob. That is, my concern had been that temporaries on the stack could perhaps get created.

I see. I think this is a common pattern for Rust and should not be an issue, especially on release builds.

00xc:
TODO for me:
* Rebase based on nicstange's changes, when/if they are merged (or at least based on [e2acf29](https://github.com/coconut-svsm/svsm/commit/e2acf29c58ab39fb516e046ee76ac5e13ed44e08))
* Refactor to use an `initialized` flag instead of an `Option`. The main issue is that there are no guarantees on the layout of the `Option`, so, in debug builds, we cannot copy into it from a pointer without first copying to the stack, and then from the stack into the Option (via something like `Option::replace()`), which will be in static storage for regular users of the API.

src/utils/immut_after_init.rs
Freax13: Removing the check is unsound in Rust as it would allow creating references to uninitialized memory.

Freax13: Yes and no. Without the patches from this pr, the use of `assume_init_ref` unsound. With the patches from this pr, `assume_init_ref` is no longer used and the call to `unwrap()` ensures that the value is actually initialized.

00xc: I understand, but there's the performance point Nicolai raised above.

Freax13: The problem is that when you allow code like this you're essentially at the same security level as if you were writing C (as in you just have to be careful) and all of Rust's guarantees fly out the window. Now, there's inherently nothing wrong with writing C, but the whole point of Rust is usually to avoid UB foot-guns like that. 

Regarding performance, have you considered making frequently accessed variables atomic e.g. using `AtomicU64`? On x86, reading from an atomic in Rust compiles into the same `mov` instruction as a regular read, so from a performance perspective there's no difference. The difference would be that this would be easy to implement in 100% safe Rust.

00xc: Yes, using a separate flag is what we will end up doing, see my [comment](https://github.com/coconut-svsm/svsm/pull/2#issuecomment-1474186036) from a couple of weeks ago. There are two issues with this though:

1. **It does not solve the performance concern**. An `Option<T>` (for a `T` that does not allow for the [null pointer optimization](https://doc.rust-lang.org/core/option/index.html#representation)) already gets compiled down to a `T` plus a tag to represent `Some` or `None`, (so `1` or `0` essentially). IOW, using a separate flag is the same as using an `Option` in terms of performance (`unwrap()` is just a check with a panic in one of the branches). On the upside, we can directly copy to the data buffer via raw pointers. An `Option` does not allow this due to the lack of guarantees regarding layout.
2. **Using atomics concedes that `ImmutAfterInit*` types will be thread safe**. This would imply adding extra logic for synchronized initialization, which would add needless overhead, since we only initialize these structures from single-CPU contexts. This is why I'm inclined to use a regular `bool` instead of an `AtomicBool`.

nicstange: Without commenting on this particular change, just as a generic remark: the main intent for introducing the `ImmutAfterInit` was to encapsulate a couple of plain `static mut` variables with their common pattern of being initialized once at bootup and effectively becoming RO thereafter, so that their respective accesses didn't require `unsafe{}` sprinkled all over the place, but are well confined within the ImmutAfterInitCell implementation. Without any additional runtime checks enabled in production, the pattern is inherently unsafe, I think, but wrapping all occurrences thereof in an `ImmutAfterInitCell`, which can be easily grepped for, at least makes it reviewable. That being said, I personally think that the RO-after-init pattern is so straight-forward -- and the initialization order at boot so deterministic, that the overhead of checking all RO accesses on production builds is not really justified. A runtime detection feature enabled only for debug builds OTOH would certainly help identifying bugs within the initialization sequence at boot.

Ultimately, only real world benchmarks would show whether or not my performance concerns are void. For the sake of keeping changes incremental, I'd propose to gate any runtime RO-access checks behind a rust feature in a first step, which could potentially get enabled by default (or removed alltogether) even for release builds later on.

Freax13: > 1. **It does not solve the performance concern**.

Is there actually a performance problem here? Has this been benchmarked? IMHO this seems a bit like premature optimization.

>   2. **Using atomics concedes that `ImmutAfterInit*` types will be thread safe**. This would imply adding extra logic for synchronized initialization, which would add needless overhead, since we only initialize these structures from single-CPU contexts. 

Note that `ImmutAfterInit*` is currently already (incorrectly) marked as thread safe:
https://github.com/coconut-svsm/svsm/blob/907a177ba56d96c3b7d48605c6fed12f9d109e1f/src/utils/immut_after_init.rs#L109-L110

> This is why I'm inclined to use a regular `bool` instead of an `AtomicBool`.

In case you want the interior mutability of an atomic, but want to restrict it to be non thread safe, you could also use `Cell<bool>`.

---

Other than that, are you aware of [`OnceCell`](https://doc.rust-lang.org/core/cell/struct.OnceCell.html) in the standard library? This type seems very similar to `ImmutAfterInit*`.

00xc: > > 1. **It does not solve the performance concern**.
> 
> Is there actually a performance problem here? Has this been benchmarked? IMHO this seems a bit like premature optimization.

Not that I'm aware of.

> > 2. **Using atomics concedes that `ImmutAfterInit*` types will be thread safe**. This would imply adding extra logic for synchronized initialization, which would add needless overhead, since we only initialize these structures from single-CPU contexts.
> 
> Note that `ImmutAfterInit*` is currently already (incorrectly) marked as thread safe:
> 
> https://github.com/coconut-svsm/svsm/blob/907a177ba56d96c3b7d48605c6fed12f9d109e1f/src/utils/immut_after_init.rs#L109-L110

This is only done so that the compiler allows global access. It's the same concept as `SyncUnsafeCell`, which I use in my PR.

> > This is why I'm inclined to use a regular `bool` instead of an `AtomicBool`.
> 
> In case you want the interior mutability of an atomic, but want to restrict it to be non thread safe, you could also use `Cell<bool>`.

Yes, that was the idea I had in mind :)

> Other than that, are you aware of [`OnceCell`](https://doc.rust-lang.org/core/cell/struct.OnceCell.html) in the standard library? This type seems very similar to `ImmutAfterInit*`.

The main problem this has is it does not allow for reinitialization AFAICT, which we do use in some places (see `ImmutAfterInitCell::reinit()`), although I have not evaluated if we could refactor the code in a way that it would remove this need.

Freax13: > > Other than that, are you aware of [`OnceCell`](https://doc.rust-lang.org/core/cell/struct.OnceCell.html) in the standard library? This type seems very similar to `ImmutAfterInit*`.
> 
> The main problem this has is it does not allow for reinitialization AFAICT, which we do use in some places (see `ImmutAfterInitCell::reinit()`), although I have not evaluated if we could refactor the code in a way that it would remove this need.

`OnceCell` does allow reinitialization, if you can get a mutable reference to it. For `ImmutAfterInitCell::reinit` the caller is already required to uphold the same guarantees as are required for a mutable reference, so this would be a large change in practice. Perhaps what is needed here is a wrapper type that allows read access from safe code and write access from unsafe code. This would allow most code to read from the `OnceCell` and restrict unsafe to the few places where the value is changed.

Freax13: > > > 2. **Using atomics concedes that `ImmutAfterInit*` types will be thread safe**. This would imply adding extra logic for synchronized initialization, which would add needless overhead, since we only initialize these structures from single-CPU contexts.
> > 
> > 
> > Note that `ImmutAfterInit*` is currently already (incorrectly) marked as thread safe:
> > https://github.com/coconut-svsm/svsm/blob/907a177ba56d96c3b7d48605c6fed12f9d109e1f/src/utils/immut_after_init.rs#L109-L110
> 
> This is only done so that the compiler allows global access. It's the same concept as `SyncUnsafeCell`, which I use in my PR.

Still, as is, `ImmutAfterInitCell` is marked as thread-safe, which is technically wrong if `T` is not `Sync`. `SyncUnsafeCell` can this because it doesn't allow safe code any access to the contained value from a shared reference and because it requires `T` to be `Sync`.

00xc: So I've looked into using an `OnceCell`, but AFAICT there is no way to initialize it from a pointer (`OnceCell::set()` takes `T` by value), which runs into the same issue as using an `Option`: in debug builds the value to be inserted is copied first to the stack and then to the backing memory location.

Freax13: Doesn't `ImmutCell` initialize by value as well?
https://github.com/coconut-svsm/svsm/blob/d462fd3fc0eb791a9050abefe4a72cb7ef11b754/src/utils/immut_after_init.rs#L86-L91

> which runs into the same issue as using an `Option`: in debug builds the value to be inserted is copied first to the stack and then to the backing memory location.

Why is this a problem? Performance?

00xc: Yes, but more importantly it is about stack usage. Nicolai raised this internally when I opened the PR; there might be issues when copying big blobs of data (e.g. `SnpCpuidTable`) since they might overflow the reduced stack space available during the early stages.

00xc:
I've simplified the PR:
* No longer uses `SyncUnsafeCell` (it introduces a dependency on a nightly feature).
* No longer takes advantage of the `Copy` trait (it is no longer used in the main branch, plus it proved to have some problems). Initialization is done by copying from a pointer as before.
* No longer uses an `Option` to track state, but an `AtomicBool`, since it implements `Sync` and has a simple API.

nicstange (commented):
I like this PR, but have three suggestions though. I don't want to delay the merge any further just because of nitpicking, so it might make sense to merge it as-is and do a follow-up patch -- up to Joerg to decide.

1.) You removed the `#[repr(transparent)]` from `ImmutAfterInitCell`. Have you tried whether `#[cfg_attr(not(debug_assertions), repr(transparent))]` would work? This way the `ImmutAfterInitCell` would remain to be guaranteed a zero-cost wrapper for non-debug builds.

2.) For `not(debug_assertions)`, I would turn the new `ImmutAfterInitError` into a  [`core::convert::Infallible`](https://doc.rust-lang.org/core/convert/enum.Infallible.html). That would potentially enable the compiler to optimize the error checks at call sites away for non-debug builds. Not sure compilers are clever enough today already, but when `Infallible` is made the never-type (`!`) one day, they probably would.

3.) In my personal opinion, removing the `unsafe` marker from the `fn init()` can be debated about: from my POV it serves as an indicator to users that they need to think carefully when invoking it, i.e. that there is no possible code path where an `ImmutAfterInitCell` can get dereferenced while still in uninitialized state. Your patch set here certainly helps a lot in establishing confidence into that,  but it still doesn't provide a hard guarantee, which can be obtained only by reviewing the code. Having the `unsafe{}` markers at the potentially problematic call sites, would at least help to draw reviewers' attention at the right spots. In either case, the move to drop `unsafe` from `fn init()` might perhaps have deserved a separate patch on its own. Just saying ;)

00xc:
> I like this PR, but have three suggestions though. I don't want to delay the merge any further just because of nitpicking, so it might make sense to merge it as-is and do a follow-up patch -- up to Joerg to decide.
> 
> 1.) You removed the `#[repr(transparent)]` from `ImmutAfterInitCell`. Have you tried whether `#[cfg_attr(not(debug_assertions), repr(transparent))]` would work? This way the `ImmutAfterInitCell` would remain to be guaranteed a zero-cost wrapper for non-debug builds.

I thought about this, but I did not like the idea of having different layout guarantees for debug and release builds. Anyhow, for release builds the struct only has one field, I do not see how it can be a non-zero-cost wrapper.

> 2.) For `not(debug_assertions)`, I would turn the new `ImmutAfterInitError` into a [`core::convert::Infallible`](https://doc.rust-lang.org/core/convert/enum.Infallible.html). That would potentially enable the compiler to optimize the error checks at call sites away for non-debug builds. Not sure compilers are clever enough today already, but when `Infallible` is made the never-type (`!`) one day, they probably would.

Done in the updated PR.

> 3.) In my personal opinion, removing the `unsafe` marker from the `fn init()` can be debated about: from my POV it serves as an indicator to users that they need to think carefully when invoking it, i.e. that there is no possible code path where an `ImmutAfterInitCell` can get dereferenced while still in uninitialized state. Your patch set here certainly helps a lot in establishing confidence into that, but it still doesn't provide a hard guarantee, which can be obtained only by reviewing the code. Having the `unsafe{}` markers at the potentially problematic call sites, would at least help to draw reviewers' attention at the right spots.

I mean, technically the unsafe part is dereferencing the cell, not so much initializing it. I think having it being unsafe might raise awareness as you mention, but it also clutters the code in places where UB is really not possible. If we want to be extra safe we could add a dedicated method to access the cell contents and have that be unsafe.

00xc:
@nicstange do you think we could also remove the `T: Copy` constraint? I don't think it is needed at all.

00xc:
With [10b7610](https://github.com/coconut-svsm/svsm/pull/2/commits/10b761095b957d3465846881ea4d8483a20d64c0) we could perhaps have a slightly better initialization of the filesystem. Instead of:

https://github.com/coconut-svsm/svsm/blob/631ae3b8db8f22915449bbb4326c8233f1171729/src/fs/fs.rs#L129-L136

We could remove `SvsmFs::initialize()` and have something like:

```rust
static FS_ROOT: ImmutAfterInitCell<SvsmFs> = ImmutAfterInitCell::uninit();

pub fn initialize_fs() { 
    let root_dir = Arc::new(RamDirectory::new()); 
    FS_ROOT.init(&root_dir);
}
```

PR #3, opened 2023-03-16
Stage2: Extract C-bit position from CPUID page

Freax13:
Previously the C-bit position was requested from the hypervisor by executing a SEV Information request via the GHCB MSR protocol. This allows a malicous host to feed us an incorrect value. Instead the C-bit position should be extracted from the CPUID page.

### DIFF
--- a/src/boot_stage2.rs
+++ b/src/boot_stage2.rs
@@ -121,33 +121,17 @@ global_asm!(
 
     get_pte_c_bit:
         /*
-         * Determine the PTE C-bit position. The user could mistakenly attempt
-         * to launch the SVSM in a non-SEV-SNP environment and we should handle
-         * this gracefully here, so that a meaningful error can be reported at a
-         * later stage when the console is functional. Ultimately, the C-bit
-         * is found from CPUID 0x8000001f[%ebx]. The cpuid insn cannot be used
-         * under SEV-ES or SEV-SNP, because the HV would fail to emulate it at
-         * this point. Under SEV-SNP, there is the CPUID page, but that's not
-         * available under either SEV or SEV-ES, where the C-bit would also
-         * strictly be needed to proceed. For SEV-ES, the GHCB MSR protocol can
-         * be used to retrieve the information, but not on plain SEV -- on the
-         * latter, we're left only with the cpuid insn. For code simplicity, this
-         * approach will also be used on SEV-SNP, the more secure CPUID page will
-         * be examined at a later stage. First read from the SEV_STATUS MSR
-         * to figure out whether any and which of SEV/SEV-ES/SEV-SNP is enabled.
-         * The MSR might not exist if SEV is not supported at all, but this
-         * will be handled gracefully by __rdmsr_safe.
+         * Check that the SNP_Active bit in the SEV_STATUS MSR is set.
          */
         movl $0xc0010131, %ecx
         call __rdmsr_safe
         testl %ecx, %ecx
-        js .Lno_sev
+        js .Lno_sev_snp
 
-        testl $0x01, %eax
-        jz .Lno_sev
+        testl $0x04, %eax
+        jz .Lno_sev_snp
 
-        testl $0x06, %eax
-        jz .Lsev_no_es
+        /* Determine the PTE C-bit position from the CPUID page. */
 
         /* Read the number of entries. */
         mov CPUID_PAGE, %eax
@@ -157,7 +141,7 @@ global_asm!(
     .Lcheck_entry:
         /* Check that there is another entry. */
         test %eax, %eax
-        je .Lno_sev
+        je .Lno_sev_snp
 
         /* Check the input parameters of the current entry. */
         cmpl $0x8000001f, (%ecx) /* EAX_IN */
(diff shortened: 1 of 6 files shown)


nicstange:
Hi,

many thanks for your PR!

However, at that early stage the SVSM code is supposed to be fairly tolerant and to come up by all means to the point where a console is available and meaningful error information can be reported. That is, if a user accidentally launches the SVSM in e.g. a SEV-ES environment rather than SNP, the SVSM should report that and abort. For making this to work, the initial C-bit must get retrieved via the GHCB MSR protocol, because that's available under SEV-ES as well while the CPUID page is not.

Once the SVSM has confirmed that the environment looks like it's a SNP one by all means, it will verify that the C-bit position as announced in the CPUID page matches what has previously been retrieved from the GHCB MSR protocol.

For reference, see Stage2's `setup_env()`, the `paging_init()` called therefrom does the aforementioned C-bit verification.

Do you agree that the issue your patch addresses is already handled by this or am I missing something?

Thanks!

Nicolai

Freax13:
> However, at that early stage the SVSM code is supposed to be fairly tolerant and to come up by all means to the point where a console is available and meaningful error information can be reported. That is, if a user accidentally launches the SVSM in e.g. a SEV-ES environment rather than SNP, the SVSM should report that and abort. For making this to work, the initial C-bit must get retrieved via the GHCB MSR protocol, because that's available under SEV-ES as well while the CPUID page is not.

I was under the impression that the code I changed was already only used for SEV-SNP. Turns out you're correct and it's also used for SEV-ES. I'll fix that. Perhaps we should read the `SEV_STATUS` MSR to check if the `SNP_Active` bit is set, use the CPUID page for SEV-SNP and fall back to the previous code for SEV-ES. We already do something similar for SEV.

> Once the SVSM has confirmed that the environment looks like it's a SNP one by all means, it will verify that the C-bit position as announced in the CPUID page matches what has previously been retrieved from the GHCB MSR protocol.
> 
> For reference, see Stage2's `setup_env()`, the `paging_init()` called therefrom does the aforementioned C-bit verification.

Is that safe? My main concern here is that if an attacker were to provide a wrong C-bit location, the SVSM's memory would be shared and could be modified by the attacker. The attacker could then corrupt the stack and leak the Secrets page. By the time `paging_init` verifies the C-bit location, the attack would have already been executed.

nicstange:
> > However, at that early stage the SVSM code is supposed to be fairly tolerant and to come up by all means to the point where a console is available and meaningful error information can be reported. That is, if a user accidentally launches the SVSM in e.g. a SEV-ES environment rather than SNP, the SVSM should report that and abort. For making this to work, the initial C-bit must get retrieved via the GHCB MSR protocol, because that's available under SEV-ES as well while the CPUID page is not.
> 
> I was under the impression that the code I changed was already only used for SEV-SNP. Turns out you're correct and it's also used for SEV-ES. I'll fix that. Perhaps we should read the `SEV_STATUS` MSR to check if the `SNP_Active` bit is set, use the CPUID page for SEV-SNP and fall back to the previous code for SEV-ES. We already do something similar for SEV.

We talked about it internally and came to the conclusion it'd be best to keep things simple there and remove all that error reporting functionality for non-SNP environments. That is, verify whether the SEV_STATUS MSR reports that SNP is active and go into a hlt loop if not. If active, try to get the C-bit from the CPUID page. If the corresponding CPUID entry isn't found -> hlt loop again.

For extra safety, it might be worthwhile to confirm the C-bit encryption control works as intended by using an approach similar to what the Linux kernel does in `arch/x86/kernel/sev_verify_cbit.S`:  push a `rdrand` value on the stack before paging is enabled and read it back and compare it to the expected value (held in some register all the time) afterwards.

In either case, removing support for that error reporting under non-SNP will enable some cleanups to the Stage2's  setup_env() phase as well.

Now the question: would you like to continue working on this and implement the above based on your existing CPUID page examination code from this patch here or shall I give it a shot?

> > Once the SVSM has confirmed that the environment looks like it's a SNP one by all means, it will verify that the C-bit position as announced in the CPUID page matches what has previously been retrieved from the GHCB MSR protocol.
> > For reference, see Stage2's `setup_env()`, the `paging_init()` called therefrom does the aforementioned C-bit verification.
> 
> Is that safe? My main concern here is that if an attacker were to provide a wrong C-bit location, the SVSM's memory would be shared and could be modified by the attacker. The attacker could then corrupt the stack and leak the Secrets page. By the time `paging_init` verifies the C-bit location, the attack would have already been executed.

You're right, I think.

Freax13:
> > > However, at that early stage the SVSM code is supposed to be fairly tolerant and to come up by all means to the point where a console is available and meaningful error information can be reported. That is, if a user accidentally launches the SVSM in e.g. a SEV-ES environment rather than SNP, the SVSM should report that and abort. For making this to work, the initial C-bit must get retrieved via the GHCB MSR protocol, because that's available under SEV-ES as well while the CPUID page is not.
> > 
> > 
> > I was under the impression that the code I changed was already only used for SEV-SNP. Turns out you're correct and it's also used for SEV-ES. I'll fix that. Perhaps we should read the `SEV_STATUS` MSR to check if the `SNP_Active` bit is set, use the CPUID page for SEV-SNP and fall back to the previous code for SEV-ES. We already do something similar for SEV.
> 
> We talked about it internally and came to the conclusion it'd be best to keep things simple there and remove all that error reporting functionality for non-SNP environments. That is, verify whether the SEV_STATUS MSR reports that SNP is active and go into a hlt loop if not. If active, try to get the C-bit from the CPUID page. If the corresponding CPUID entry isn't found -> hlt loop again.
> 
> For extra safety, it might be worthwhile to confirm the C-bit encryption control works as intended by using an approach similar to what the Linux kernel does in `arch/x86/kernel/sev_verify_cbit.S`: push a `rdrand` value on the stack before paging is enabled and read it back and compare it to the expected value (held in some register all the time) afterwards.
> 
> In either case, removing support for that error reporting under non-SNP will enable some cleanups to the Stage2's setup_env() phase as well.
> 
> Now the question: would you like to continue working on this and implement the above based on your existing CPUID page examination code from this patch here or shall I give it a shot?

Sure, I'd love to. I'll start working on it over the weekend.

Freax13:
> > > However, at that early stage the SVSM code is supposed to be fairly tolerant and to come up by all means to the point where a console is available and meaningful error information can be reported. That is, if a user accidentally launches the SVSM in e.g. a SEV-ES environment rather than SNP, the SVSM should report that and abort. For making this to work, the initial C-bit must get retrieved via the GHCB MSR protocol, because that's available under SEV-ES as well while the CPUID page is not.
> > 
> > 
> > I was under the impression that the code I changed was already only used for SEV-SNP. Turns out you're correct and it's also used for SEV-ES. I'll fix that. Perhaps we should read the `SEV_STATUS` MSR to check if the `SNP_Active` bit is set, use the CPUID page for SEV-SNP and fall back to the previous code for SEV-ES. We already do something similar for SEV.
> 
> We talked about it internally and came to the conclusion it'd be best to keep things simple there and remove all that error reporting functionality for non-SNP environments. That is, verify whether the SEV_STATUS MSR reports that SNP is active and go into a hlt loop if not. If active, try to get the C-bit from the CPUID page. If the corresponding CPUID entry isn't found -> hlt loop again.

I have removed all code that was used for error reporting on non-SEV envs from the raw assembly part of stage2 as well as `setup_env`. I also replaced all code that extracts the C-bit location from the SEV information GHCB response in favor of code that extracts it from the CPUID page. 

> For extra safety, it might be worthwhile to confirm the C-bit encryption control works as intended by using an approach similar to what the Linux kernel does in `arch/x86/kernel/sev_verify_cbit.S`: push a `rdrand` value on the stack before paging is enabled and read it back and compare it to the expected value (held in some register all the time) afterwards.

I haven't implemented this yet. This could probably be done in a separate pr though.

joergroedel:
Looks good to me, thanks for tackling this!

PR #4, opened 2023-03-17
Load the SVSM kernel from an ELF file

nicstange:
- Implement a minimal ELF file reader
- Make the Stage2 to read the SVSM kernel from an ELF (instead of raw binary) and load it according to the layout described therein.
- Let the SVSM kernel determine its own memory layout to transfer over to the runtime page tables from the ELF metadata as well.
- Implement address space randomization in Stage2.

Future plans include
- to load the symbol table from the ELF as well and to e.g. pretty-print stacktraces and
- to use the unwinding records from the ELF for unwinding, alleviating the need for the current framepointer based approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,11 +30,11 @@ stage1/stage2.bin:
 	cargo build ${CARGO_ARGS} --bin stage2
 	objcopy -O binary ${STAGE2_ELF} $@
 
-stage1/kernel.bin:
+stage1/kernel.elf:
 	cargo build ${CARGO_ARGS} --bin svsm
-	objcopy -O binary ${KERNEL_ELF} $@
+	objcopy -O elf64-x86-64 --strip-unneeded ${KERNEL_ELF} $@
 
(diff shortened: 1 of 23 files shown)


joergroedel:
I really like these changes, but want to go this way in somewhat smaller steps.

The reason is that the ASLR changes drop the assumption that the the virtual base is aligned to a 2M boundary, which will hurt us down the road. So here are the steps I propose:

1. Submit only the ELF loader changes, they make the bulk of this PR anyway
2. Then we separate the kernel text+data+bss mapping from the SVSM memory mapping, best into a separate top-level page-table entry. Since the kernel image is still in the SVSM memory mapping, make sure it is mapped RO or even unmapped there.
3. Implement ASLR based on this change, which allows to have more bits of randomness. Even the top-level page-table index can be random within certain bounds.

nicstange:
I'm closing this PR, a new one with the first ELF loader part only will follow shortly.

PR #7, opened 2023-03-22
Add error handling for GHCB operations

00xc:
Add error types for GHCB and MSR protocol operations.

These errors are mostly only to be handled from two sites, namely the firmware setup in `svsm_main()` and the `PerCpu` handlers. At the first one, we just panic as before, printing a more verbose message. At the second group, we ignore the error types for now to simplify the patches, but handling will be added in a separate PR.


joergroedel:
Thanks, this is a nice improvement!

PR #8, opened 2023-03-24
Improve error typing

00xc:
Add a new generic error type, `SvsmError`. This type can be expanded over time with additional carried information or new variants to represent different error conditions.

In general, internally each subsystem may have its own internal error type (e.g. `GhcbError`), but must return in its public interface(s) an `SvsmError`, which may contain this internal (but not private) error type with additional details (e.g. `SvsmError::Ghcb`).

For guest requests, we define a way to translate these general errors to error codes (`SvsmError` -> `SvsmReqError`) via the From trait.

Unfortunately, due to the pervasiveness of error handling and strong typing, this requires a lot of small cascading changes, so making commits self-contained was not an easy task :)

### DIFF
--- a/src/mm/alloc.rs
+++ b/src/mm/alloc.rs
@@ -4,6 +4,7 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
+use crate::error::SvsmError;
 use crate::locking::SpinLock;
 use crate::types::{PhysAddr, VirtAddr, PAGE_SHIFT, PAGE_SIZE};
 use crate::utils::{align_up, zero_mem_region};
@@ -283,9 +284,9 @@ impl MemoryRegion {
         Page::from_mem(info)
     }
 
-    pub fn get_page_info(&self, vaddr: VirtAddr) -> Result<Page, ()> {
+    pub fn get_page_info(&self, vaddr: VirtAddr) -> Result<Page, SvsmError> {
         if vaddr == 0 || !self.check_virt_addr(vaddr) {
-            return Err(());
+            return Err(SvsmError::Mem);
         }
 
         let pfn = (vaddr - self.start_virt) / PAGE_SIZE;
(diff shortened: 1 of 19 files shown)


00xc:
> Hmm, this fails to boot for me early in stage2 when compiled as a release build.

Looking into it

00xc:
The fact that this only fails in release mode is quite suspicious.

It's breaking in:
```
stage2_main()
-> init_valid_bitmap_alloc()
    -> vb_ref.clear_all()
        -> ptr::write_bytes(self.bitmap, 0, index);
```

`self.bitmap` is coming from `allocate_pages()`, whose error type I changed in [48b31b03a473](https://github.com/coconut-svsm/svsm/pull/8/commits/48b31b03a4735852cbc51fc7c5dbf32be0fc02aa#diff-4466e802ff4c9e56138b26df9909800209d3f332fe2ef08078fc99cc484617d5R372) (`svsm/error: add error typing in the memory management subsystem`). There is no crash before this commit.

`self.bitmap` and `index` seem to be the same before and after the commit, so something is getting optimized out along the way I'd say. I'll keep looking at it next week.

src/mm/alloc.rs
```diff
@@ -368,41 +369,37 @@ impl MemoryRegion {
...
+        let vaddr = self.allocate_page()?;
 
         unsafe {
             asm!("rep stosq",
-                in("rdi") vaddr.unwrap(),
+                in("rdi") vaddr,
                 in("rax") 0,
                 in("rcx") PAGE_SIZE / 8,
                 options(att_syntax));
         }
```
00xc, line 397: Ok so this syntax cleanup unlocked an optimization that exposed a bug we had here. `rdi` should be set as inout, so that the compiler knows that the contents it points to are changed. This is similar to the bug we fixed in 35b8a31aa2b9.

00xc: Fixed in #9

PR #10, opened 2023-03-27
Fix trivial clippy warnings

00xc:
A few fixes for some clippy warnings which should not introduce any functional changes.


joergroedel:
Looks good, thanks!

PR #18, opened 2023-03-28
Allow guest VMPL compile-time configuration

### LINKED ISSUE #11: Make guest VMPL compile-time configurable
Currently the guest VMPL is hard-coded to VMPL-1 throughout the code.

Introduce a single global constant which defines the guest VMPL level and use it everywhere in the code. Try to run a guest at VMPL-2 and VMPL-3.

Note that the VMPLCK keys in the secret page need to be cleared for all lower-level VMPLs before the page is passed to the guest.

00xc:
Remove all hardcoded instances of VMPL=1 for the guest, and replace them with a new constant, `GUEST_VMPL`. Introduce as well a new value in RMPFlags so that users of the structure can easily refer to the guest VMPL.

The best home for the new constant seems to be `src/types.rs`, but this could be changed.

I have verified that a guest can be booted in VMPL 2 by simply changing the value of the constant.

Fixes #11.

### DIFF
--- a/src/cpu/percpu.rs
+++ b/src/cpu/percpu.rs
@@ -338,7 +338,7 @@ impl PerCpu {
             return Err(SvsmError::Mem);
         }
 
-        let vaddr = allocate_new_vmsa(RMPFlags::VMPL1)?;
+        let vaddr = allocate_new_vmsa(RMPFlags::GUEST_VMPL)?;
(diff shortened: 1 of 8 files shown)


joergroedel:
Works like a charm, booted a guest at VMP-3 with a debug and release build.

PR #24, opened 2023-04-20
Replace slot-based virtual memory allocation with bitmap allocator

roy-hopkins:
The current method of mapping a virtual memory address to a physical address involves mapping individual pages using slots, with a guard page between each virtual address. It is not possible to create a virtual memory mapping that spans multiple pages.

This new virtual address allocator allows dynamic allocation of virtual addresses that can span multiple 4K or 2M pages using a bitmap to keep track of virtual pages that are in use.

This simplifies the development of future capabilities and in particular simplifies parsing of variable size structures such as those used in the files of a firmware file system.

### DIFF
--- a/src/mm/virtualrange.rs
+++ b/src/mm/virtualrange.rs
@@ -1,0 +1,111 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Roy Hopkins <rhopkins@suse.de>
+
+use crate::types::{VirtAddr, PAGE_SHIFT, PAGE_SIZE, PAGE_SIZE_2M, PAGE_SHIFT_2M};
+use crate::locking::SpinLock;
+use crate::error::SvsmError;
+use crate::utils::bitmap_allocator::{BitmapAllocator1024, BitmapAllocator};
+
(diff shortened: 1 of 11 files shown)


src/mm/virtualrange.rs
```diff
@@ -0,0 +1,111 @@
...
+        panic!("Attempted to allocate page map with more than 4K pages");
+    }
+    pm4k.start_virt = SVSM_PERCPU_TEMP_BASE_4K;
+    pm4k.page_count = page_count;
+    pm4k.bits.set(0, page_count, false);
+
+    let mut pm2m = VIRTUAL_MAP_2M.lock();
+    let page_count = (SVSM_PERCPU_TEMP_END_2M - SVSM_PERCPU_TEMP_BASE_2M) / PAGE_SIZE_2M;
+    if page_count > BitmapAllocator1024::CAPACITY {
+        panic!("Attempted to allocate page map with more than 4K pages");
```
joergroedel, line 67: This panic message is the same as above. Can you please add the page-size to each message to make it clear for which size the allocation failed?

src/utils/bitmap_allocator.rs
```diff
@@ -0,0 +1,493 @@
+
```
joergroedel, line 1: Please add a copyright header to each new file.

joergroedel (commented):
Really cool, thanks a lot! I left a few minor comments, otherwise this looks great to me.

joergroedel:
Thanks a lot, this is a great addition and simplification of internal interfaces.

PR #25, opened 2023-04-26
Add GDB stub to debug builds.

roy-hopkins:
This PR adds a GDB stub capability to coconut-svsm which activates on panics and forced breakpoints within debug builds. The stub provides a debug capability from the main SVSM entry point through to the launch of OVMF. It remains active at this point and allows debugging of the kernel request loop. 

The ability to debug the request loop will be useful for anyone developing future SVSM modules, such as a TPM module.

The stub is implemented behind a cargo 'feature' meaning that it is completely excluded from release builds.


src/debug/gdbstub.rs
```diff
@@ -0,0 +1,488 @@
...
+                "lea try_done(%rip), %rax",
+                out("rax") td,
+                options(att_syntax)
+            );
+            PF_RETURN_ADDR = td;
+        }
+    }
+
+    #[allow(named_asm_labels)]
+    pub fn safe_access_memory(addr: VirtAddr, write: bool, value: u8) -> Result<u8, ()> {
```
joergroedel, line 180: This can probably be implemented with the exception fixup code that is already available in SVSM.

src/mm/pagetable.rs
```diff
@@ -203,6 +203,14 @@ impl PageTable {
         })
     }
 
+    // When GDB is enabled allow the executable memory to be writable to
```
joergroedel, line 206: I think it is better to use temporary mapping for pages that need to be written to. That way code pages can stay read-only.

joergroedel (commented):
Thanks a lot, this is a great addition and very helpful for debugging!

Two more things:

1. Can the gdbstub (optionally) make use of debug registers (DR0-3, DR6-7)? We can check SEV features if debug register swap is enabled and use them in this case.
2. Are there implications when ASLR for the kernel is implemented? Will that break gdbstub? In that case we should make this feature mutually exclusive with ASLR once we get it.

src/console.rs
```diff
@@ -5,21 +5,17 @@
...
 use core::fmt;
 use log;
 
-pub trait ConsoleWriter {
-    fn put_byte(&self, _ch: u8) {}
-}
-
 pub struct Console {
-    writer: *mut dyn ConsoleWriter,
+    writer: *mut dyn Terminal,
```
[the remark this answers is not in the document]
roy-hopkins: Done

src/mm/pagetable.rs
roy-hopkins: Done.

joergroedel:
This is merged now, with a commit on-top that excludes GDB stub from debug builds. The makefile now takes the FEATURES=enable-gdb variable to enable GDB stub for debug or release builds.

PR #26, opened 2023-04-26
Introduce VirtAddr and PhysAddr types.

00xc:
At the moment, `VirtAddr` and `PhysAddr` are simply typedefs, meaning that one might use either one interchangeably without the compiler giving any warnings. Introduce two distinct new types, plus an `Address` trait implemented by both for common operations.

The new types provide several facilities to convert easily to and from other types such as `usize`, `u64` and raw pointers via the `From` trait.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -1,3 +1,9 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Carlos López <carlos.lopez@suse.com>
+
 use crate::types::PAGE_SIZE;
 use core::fmt;
 use core::ops;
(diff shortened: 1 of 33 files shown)


joergroedel (commented):
Looks great, thanks for doing this! One thing I was wondering is whether we can also have define From traits to convert between VirtAddr and PhysAddr which does virt_to_phys/phys_to_virt() implicitly?

00xc:
> Looks great, thanks for doing this! One thing I was wondering is whether we can also have define From traits to convert between VirtAddr and PhysAddr which does virt_to_phys/phys_to_virt() implicitly?

This would be possible, although introducing a dependency on a global mutable (`KERNEL_MAPPING`) seemed a bit strange to me in a `From` conversion, especially since it can panic. We could add it as [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) perhaps? This would increase the amount of error handling though.

joergroedel:
> This would be possible, although introducing a dependency on a global mutable (`KERNEL_MAPPING`) seemed a bit strange to me in a `From` conversion, especially since it can panic. We could add it as [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) perhaps? This would increase the amount of error handling though.

Yeah, you are right, keeping the Virt<->Phys address conversion explicit is better and makes the code easier to read. I will merge this as-is.

PR #27, opened 2023-04-28
Initial implementation for a RAM filesystem

joergroedel:
Support for a RAM filesystem and introduction of a general filesystem interface into the SVSM code-base. The filesystem is populated from a packaged archive file at boot time. The archive file is optional for now and requires a separate tool to create.

### DIFF
--- a/src/boot_stage2.rs
+++ b/src/boot_stage2.rs
@@ -16,6 +16,8 @@ global_asm!(
         .globl startup_32
         startup_32:
 
+        /* Save pointer to startup structure in ESI */
+        movl %esp, %esi
         /*
          * Load a GDT. Despite the naming, it contains valid
          * entries for both, "legacy" 32bit and long mode each.
@@ -187,8 +189,6 @@ global_asm!(
         movw %ax, %gs
         movw %ax, %ss
 
-        pushq   %rdi
-
         /* Clear out .bss and transfer control to the main stage2 code. */
         xorq %rax, %rax
         leaq _bss(%rip), %rdi
@@ -197,7 +197,7 @@ global_asm!(
         shrq $3, %rcx
         rep stosq
 
-        popq    %rdi
+        movq %rsi, %rdi
         jmp stage2_main
 
         .data
--- a/src/fs/api.rs
+++ b/src/fs/api.rs
@@ -46,8 +46,8 @@ pub trait File {
 pub trait Directory {
     fn list(&self) -> Vec<FileName>;
     fn lookup_entry(&self, name: FileName) -> Result<DirEntry, SvsmError>;
-    fn create_file(&mut self, name: FileName) -> Result<Arc<dyn File>, SvsmError>;
-    fn create_directory(&mut self, name: FileName) -> Result<Arc<dyn Directory>, SvsmError>;
+    fn create_file(&self, name: FileName) -> Result<Arc<dyn File>, SvsmError>;
+    fn create_directory(&self, name: FileName) -> Result<Arc<dyn Directory>, SvsmError>;
 }
 
 pub enum DirEntry {
--- a/src/fs/api.rs
+++ b/src/fs/api.rs
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2023 SUSE LLC
+//
+// Author: Joerg Roedel <jroedel@suse.de>
+
+extern crate alloc;
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+
+use crate::error::SvsmError;
+use crate::string::FixedString;
+
+/// Maximum supported length for a single filename
+const MAX_FILENAME_LENGTH: usize = 64;
+pub type FileName = FixedString<MAX_FILENAME_LENGTH>;
+
+#[derive(Copy, Clone, Debug)]
+pub enum FsError {
+    Inval,
+    FileExists,
+    FileNotFound,
+}
+
+macro_rules! impl_fs_err {
+    ($name:ident, $v:ident) => {
+        pub fn $name() -> Self {
+            Self::$v
+        }
+    };
+}
+
+impl FsError {
+    impl_fs_err!(inval, Inval);
+    impl_fs_err!(file_exists, FileExists);
+    impl_fs_err!(file_not_found, FileNotFound);
+}
+
+pub trait File {
+    fn read(&self, buf: &mut [u8], offset: usize) -> Result<usize, SvsmError>;
+    fn write(&self, buf: &[u8], offset: usize) -> Result<usize, SvsmError>;
+    fn truncate(&self, size: usize) -> Result<usize, SvsmError>;
+    fn size(&self) -> usize;
+}
+
+pub trait Directory {
+    fn list(&self) -> Vec<FileName>;
+    fn lookup_entry(&self, name: FileName) -> Result<DirEntry, SvsmError>;
+    fn create_file(&mut self, name: FileName) -> Result<Arc<dyn File>, SvsmError>;
+    fn create_directory(&mut self, name: FileName) -> Result<Arc<dyn Directory>, SvsmError>;
+}
+
+pub enum DirEntry {
+    File(Arc<dyn File>),
+    Directory(Arc<dyn Directory>),
+}
+
+impl DirEntry {
+    pub fn is_file(&self) -> bool {
+        matches!(self, Self::File(_))
+    }
+
+    pub fn is_dir(&self) -> bool {
+        matches!(self, Self::Directory(_))
+    }
+}
+
+impl Clone for DirEntry {
+    fn clone(&self) -> Self {
+        match self {
+            DirEntry::File(f) => DirEntry::File(f.clone()),
+            DirEntry::Directory(d) => DirEntry::Directory(d.clone()),
+        }
+    }
+}
+
+pub struct DirectoryEntry {
+    pub name: FileName,
+    pub entry: DirEntry,
+}
+
+impl DirectoryEntry {
+    pub fn new(name: FileName, entry: DirEntry) -> Self {
+        DirectoryEntry { name, entry }
+    }
+}
--- a/src/fs/api.rs
+++ b/src/fs/api.rs
@@ -48,6 +48,7 @@ pub trait Directory {
     fn lookup_entry(&self, name: FileName) -> Result<DirEntry, SvsmError>;
     fn create_file(&self, name: FileName) -> Result<Arc<dyn File>, SvsmError>;
     fn create_directory(&self, name: FileName) -> Result<Arc<dyn Directory>, SvsmError>;
+    fn unlink(&self, name: FileName) -> Result<(), SvsmError>;
 }
 
 pub enum DirEntry {
--- a/src/fs/fs.rs
+++ b/src/fs/fs.rs
@@ -171,6 +171,25 @@ fn walk_path(path_items: &Vec<&str>) -> Result<Arc<dyn Directory>, SvsmError> {
     Ok(current_dir)
 }
 
+fn walk_path_create(path_items: &Vec<&str>) -> Result<Arc<dyn Directory>, SvsmError> {
+    let mut current_dir = unsafe { FS_ROOT.root_dir() };
+
+    for item in path_items.iter() {
+        let dir_name = FileName::from(*item);
+        let lookup = current_dir.lookup_entry(dir_name);
+        let dir_entry = match lookup {
+            Ok(entry) => entry,
+            Err(_) => DirEntry::Directory(current_dir.create_directory(dir_name)?),
+        };
+        current_dir = match dir_entry {
+            DirEntry::File(_) => return Err(SvsmError::FileSystem(FsError::file_not_found())),
+            DirEntry::Directory(dir) => dir,
+        };
+    }
+
+    Ok(current_dir)
+}
+
 pub fn open(path: &str) -> Result<FileHandle, SvsmError> {
     let mut path_items = split_path(path)?;
     let file_name = FileName::from(path_items.pop().unwrap());
@@ -188,6 +207,21 @@ pub fn create(path: &str) -> Result<FileHandle, SvsmError> {
     let mut path_items = split_path(path)?;
     let file_name = FileName::from(path_items.pop().unwrap());
     let current_dir = walk_path(&path_items)?;
+    let file = current_dir.create_file(file_name)?;
+
+    Ok(FileHandle::new(&file))
+}
+
+/// Creates a file with all sub-directories
+pub fn create_all(path: &str) -> Result<FileHandle, SvsmError> {
+    let mut path_items = split_path(path)?;
+    let file_name = FileName::from(path_items.pop().unwrap());
+    let current_dir = walk_path_create(&path_items)?;
+
+    if file_name.length() == 0 {
+        return Err(SvsmError::FileSystem(FsError::inval()));
+    }
+
     let file = current_dir.create_file(file_name)?;
 
     Ok(FileHandle::new(&file))
--- a/src/fs/fs.rs
+++ b/src/fs/fs.rs
@@ -1,0 +1,460 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2023 SUSE LLC
+//
+// Author: Joerg Roedel <jroedel@suse.de>
+
+use super::ramfs::RamDirectory;
+use super::*;
+
+use crate::error::SvsmError;
+use crate::locking::SpinLock;
+
+use core::cmp::min;
+
+extern crate alloc;
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+
+struct RawFileHandle {
+    file: Arc<dyn File>,
+    current: usize,
+}
+
+impl RawFileHandle {
+    pub fn new(file: &Arc<dyn File>) -> Self {
+        RawFileHandle {
+            file: file.clone(),
+            current: 0,
+        }
+    }
+
+    pub fn read(&mut self, buf: &mut [u8]) -> Result<usize, SvsmError> {
+        let result = self.file.read(buf, self.current);
+        if let Ok(v) = result {
+            self.current += v;
+        }
+        result
+    }
+
+    pub fn write(&mut self, buf: &[u8]) -> Result<usize, SvsmError> {
+        let result = self.file.write(buf, self.current);
+        if result.is_ok() {
+            self.current += result.unwrap();
+        }
+        result
+    }
+
+    pub fn truncate(&mut self, offset: usize) -> Result<usize, SvsmError> {
+        self.file.truncate(offset)
+    }
+
+    pub fn seek(&mut self, pos: usize) {
+        self.current = min(pos, self.file.size());
+    }
+
+    pub fn size(&self) -> usize {
+        self.file.size()
+    }
+}
+
+pub struct FileHandle {
+    // Use a SpinLock here because the read operation also needs to be mutable
+    // (changes file pointer). Parallel reads are still possible with multiple
+    // file handles
+    handle: SpinLock<RawFileHandle>,
+}
+
+unsafe impl Sync for FileHandle {}
+
+impl FileHandle {
+    pub fn new(file: &Arc<dyn File>) -> Self {
+        FileHandle {
+            handle: SpinLock::new(RawFileHandle::new(file)),
+        }
+    }
+
+    pub fn read(&self, buf: &mut [u8]) -> Result<usize, SvsmError> {
+        self.handle.lock().read(buf)
+    }
+
+    pub fn write(&self, buf: &[u8]) -> Result<usize, SvsmError> {
+        self.handle.lock().write(buf)
+    }
+
+    pub fn truncate(&self, offset: usize) -> Result<usize, SvsmError> {
+        self.handle.lock().truncate(offset)
+    }
+
+    pub fn seek(&self, pos: usize) {
+        self.handle.lock().seek(pos);
+    }
+
+    pub fn size(&self) -> usize {
+        self.handle.lock().size()
+    }
+}
+
+struct SvsmFs {
+    root: Option<Arc<RamDirectory>>,
+}
+
+unsafe impl Sync for SvsmFs {}
+
+impl SvsmFs {
+    pub const fn new() -> Self {
+        SvsmFs { root: None }
+    }
+
+    pub fn initialize(&mut self, root: &Arc<RamDirectory>) {
+        assert!(!self.initialized());
+        self.root = Some(root.clone());
+    }
+
+    #[cfg(test)]
+    pub fn uninitialize(&mut self) {
+        self.root = None;
+    }
+
+    pub fn initialized(&self) -> bool {
+        self.root.is_some()
+    }
+
+    pub fn root_dir(&self) -> Arc<dyn Directory> {
+        assert!(self.initialized());
+        self.root.as_ref().unwrap().clone()
+    }
+}
+
+static mut FS_ROOT: SvsmFs = SvsmFs::new();
+
+pub fn initialize_fs() {
+    let root_dir = Arc::new(RamDirectory::new());
+    unsafe {
+        FS_ROOT.initialize(&root_dir);
+    }
+}
+
+#[cfg(test)]
+fn uninitialize_fs() {
+    unsafe {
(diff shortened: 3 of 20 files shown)


src/string.rs
```diff
@@ -75,6 +76,22 @@ impl<const N: usize> PartialEq<&str> for FixedString<N> {
...
 }
 
+impl<const N: usize> PartialEq<FixedString<N>> for FixedString<N> {
+    fn eq(&self, other: &FixedString<N>) -> bool {
+        if self.len != other.len {
+            return false;
+        }
+
+        for i in 0..self.len {
+            if self.data[i] != other.data[i] {
```
00xc, line 90: An iterator would be more conventional here. Something like:
```rust
for (a, b) in self.data.iter().zip(&other.data).take(self.len) {
    ...
}
```

Or even:
```rust
self.data
	.iter()
	.zip(&other.data)
	.take(self.len)
	.all(|(a, b)| *a == *b)
```

src/fs/fs.rs
```diff
@@ -0,0 +1,411 @@
...
+            file: file.clone(),
+            current: 0,
+        }
+    }
+
+    pub fn read(&mut self, buf: &mut [u8]) -> Result<usize, SvsmError> {
+        let result = self.file.read(buf, self.current);
+        if result.is_ok() {
+            self.current += result.unwrap();
+        }
```
00xc, line 36: Instead of checking via `is_ok()` and then unwrapping one can do:

```rust 
if let Ok(result) = self.file.read(buf, self.current) {
    self.current += result;
}
```

src/fs/fs.rs
```diff
@@ -0,0 +1,411 @@
...
+#[cfg(test)]
+fn uninitialize_fs() {
+    unsafe {
+        FS_ROOT.uninitialize();
+    }
+}
+
+fn split_path_allow_empty(path: &str) -> Vec<&str> {
+    let mut path_items: Vec<&str> = path.split('/').collect();
+    path_items.retain(|x| x.len() > 0);
```
00xc, line 136: Filtering before collecting could reduce the initial allocation size. Something like:

```rust
fn split_path_allow_empty(path: &str) -> Vec<&str> {
    path.split('/')
        .filter(|x| x.len() > 0)
        .collect()
}
```

src/fs/ramfs.rs
```diff
@@ -219,6 +219,19 @@ impl Directory for RamDirectory {
...
         Ok(new_dir)
     }
+
+    fn unlink(&self, name: FileName) -> Result<(), SvsmError> {
+        let mut vec = self.entries.lock();
+        let pos = vec.iter().position(|e| e.name == name);
+
+        match pos {
+            Some(idx) => {
+                vec.remove(idx);
```
00xc, line 231: If we do not care about the order of entries the entry could be removed with `vec.swap_remove(idx)`, which is O(1).

src/fs/ramfs.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        while cap > self.capacity {
+            self.increase_capacity()?;
+        }
+
+        Ok(())
+    }
+
+    fn read_from_page(&mut self, buf: &mut [u8], offset: usize) {
+        let page_index = offset % PAGE_SIZE;
```
00xc, line 54: Maybe I'm missing something but this does not need to be `&mut self`, right?

src/fs/ramfs.rs
```diff
@@ -0,0 +1,223 @@
...
+impl File for RawRamFile {
+    fn read(&mut self, buf: &mut [u8], offset: usize) -> Result<usize, SvsmError> {
+        let mut current = offset;
+        let mut len = buf.len();
+        let mut bytes: usize = 0;
+        let mut buf_offset = 0;
+
+        while len > 0 {
+            let page_end = min(page_align_up(current + 1), self.size);
+            let page_len = min(page_end - current, len);
```
00xc, line 85: If `current` is greater than `self.size` then `page_end - current` will overflow, which, as far as I know, causes a panic in debug builds. We should probably either check this and return an error, or use something like [`wrapping_sub()`](https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_sub).

00xc: Also, even in non-debug builds, since the substraction would result in a large number, using `min()` will make `page_len = len`, and `len` is `buf.len()` on the first iteration. This will trigger the following assertion in `read_from_page()` if the destination buffer is larger than a page:

```rust
assert!(page_end <= PAGE_SIZE);
```

So `wrapping_sub()` should be out of the question.

src/string.rs
joergroedel: Thanks, changed it to the later version.

src/fs/fs.rs
joergroedel: Right, but in this case it leads to uglier code because I need to re-assemble the return type if I do the 'if let' unwrapping.

joergroedel: Right, but in this case it leads to uglier code because I need to re-assemble the return type if I do the 'if let' unwrapping.

src/fs/fs.rs
joergroedel: That is better, fixed it. Thanks.

src/fs/fs.rs
00xc: Then it can still be done without unwrap, I just collapsed more than was needed:

```rust
let result = self.file.read(buf, self.current);
if let Ok(v) = result {
    self.current += v;
}
result
```

src/fs/fs.rs
00xc: I just noticed `x.len() > 0` perhaps should be replaced by `!x.is_empty()`, clippy will probably complain otherwise.

src/fs/ramfs.rs
joergroedel: Good question, usually we do not care about the order, but the unit-tests do. This can be fixed, though.

roy-hopkins (commented):
Looks good. I just had one thought:

If the pre-built filesystem image is going to contain a significant volume of files then unpacking the entire filesystem into RAM on every boot may not be efficient. Would it make sense to implement a read-only filesystem using the same interface over the top of something like the firmware file system that OVMF uses? RamFS could then be reserved just for dynamic or modified files.

src/fs/ramfs.rs
joergroedel: Right, fixed that. Also removed the unnecessary mut from the write function signature.

src/fs/ramfs.rs
joergroedel: I think it can be resolved by initializing current by:
`let mut current = min(offset, self.size);`
This way if offset is > self.size the read call will just return 0 bytes read.

src/fs/fs.rs
joergroedel: Okay, changed that, thanks.

src/fs/ramfs.rs
joergroedel: Okay, no current tests rely on this, so I changed it. Future tests need to take this into account when comparing directory listings with expected output.

src/fs/api.rs
```diff
@@ -0,0 +1,89 @@
...
+    FileNotFound,
+}
+
+impl FsError {
+    pub fn file_exists() -> Self {
+        Self::FileExists
+    }
+
+    pub fn file_not_found() -> Self {
+        Self::FileNotFound
```
00xc, line 35: Any particular reason we have these constructors? If not, just as a suggestion, we could probably get away with simply constructing the particular variant inline wherever needed.

src/fs/fs.rs
```diff
@@ -0,0 +1,411 @@
...
+pub fn open(path: &str) -> Result<FileHandle, SvsmError> {
+    let mut path_items = split_path(path)?;
+    let file_name = FileName::from(path_items.pop().unwrap());
+    let current_dir = walk_path(&path_items)?;
+
+    let dir_entry = current_dir.lookup_entry(file_name)?;
+
+    match dir_entry {
+        DirEntry::Directory(_) => Err(SvsmError::FileSystem(FsError::file_not_found())),
+        DirEntry::File(f) => Ok(FileHandle::new(&f)),
```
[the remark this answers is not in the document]
joergroedel: Multiple handles for the same file are allowed and should work. The file and directory operations itself are behind spin-locks, so data structure updates are serialized (should maybe converted to RWLock). The file and directory structs are also kept in reference-counted pointers, so that the data does not go away as long as there are still references to it.

One thing that popped up in my mind while thinking about this is whether the file handling is safe against files getting smaller. Because a file-handles current field could then point beyond the end-of-file. I need to check the code and possibly write a unit-test for that case (and add truncate support to actually shrink files).

src/fs/api.rs
joergroedel: Yes, I tried to construct them inline and that leads to long and hard-to-read lines. Maybe we can construct these functions via macros as done elsewhere.

joergroedel:
> Looks good. I just had one thought:
> 
> If the pre-built filesystem image is going to contain a significant volume of files then unpacking the entire filesystem into RAM on every boot may not be efficient. Would it make sense to implement a read-only filesystem using the same interface over the top of something like the firmware file system that OVMF uses? RamFS could then be reserved just for dynamic or modified files.

So the reason for the RamFS and unpacking is that the file-system contains the binaries we want to execute. Unpacking into a RamFS allows to execute them in-place, without copying the binary contents to a different memory location.

But the underlying filesystem API allows multiple backend implementations. A DirEntry can point to any struct implementing the Directory trait, so that we can implement "mounting" of other file-systems, like a read-only FS that gives access to the firmware file system.

The plan is to use the filesystem interface also for the persistency layer once that gets implemented.

src/fs/api.rs
joergroedel: Converted the code to macros.

src/fs/api.rs
```diff
@@ -0,0 +1,95 @@
...
+    Directory(Arc<dyn Directory>),
+}
+
+impl DirEntry {
+    pub fn is_file(&self) -> bool {
+        match self {
+            DirEntry::File(_) => true,
+            _ => false,
+        }
+    }
```
00xc, line 65: This could be shortened to:

```rust
pub fn is_file(&self) -> bool {
    matches!(self, Self::File(_))
}
```

Same for `is_dir()` below.

src/fs/api.rs
```diff
@@ -0,0 +1,95 @@
...
+pub struct DirectoryEntry {
+    pub name: FileName,
+    pub entry: DirEntry,
+}
+
+impl DirectoryEntry {
+    pub fn new(name: FileName, entry: DirEntry) -> Self {
+        DirectoryEntry {
+            name: name,
+            entry: entry,
```
00xc, line 93: Field names can be omitted here (clippy will complain)

src/fs/api.rs
joergroedel: Fixed, thanks.

src/fs/api.rs
joergroedel: Fixed, thanks.

src/fs/fs.rs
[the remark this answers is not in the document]
joergroedel: I converted the RamFile and RamDirectory structures to RWLock, FileHandle still uses SpinLock as the read() method needs to be mutable there as well. Parallel reads are still possible via multiple file handles.
I also added a unit-test for reading/writing/truncating a file via multiple file-handles.

src/fs/ramfs.rs
```diff
@@ -0,0 +1,258 @@
...
+        Ok(bytes)
+    }
+
+    fn write(&mut self, buf: &[u8], offset: usize) -> Result<usize, SvsmError> {
+        let mut current = offset;
+        let mut bytes: usize = 0;
+        let mut len = buf.len();
+        let mut buf_offset: usize = 0;
+
+        self.set_capacity(offset + len)?;
```
00xc, line 109: This addition could overflow if the caller supplies a big enough offset. Perhaps we could use `checked_add` to harden it against these edge cases:

```rust
let capacity = offset.checked_add(len)
    .ok_or(SvsmError::FileSystem(FsError::inval()))?;
self.set_capacity(capacity)?;
```

src/fs/ramfs.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        while len > 0 {
+            let page_len = min(PAGE_SIZE - page_offset(current), len);
+            let buf_end = buf_offset + page_len;
+
+            self.write_to_page(&buf[buf_offset..buf_end], current);
+            self.size = max(self.size, current + page_len);
```
00xc, line 116: Same here, `current + page_len` could overflow with a big enough offset.

00xc: Nevermind, `set_capacity` should fail before that happens.

joergroedel:
Population of RamFS is now implemented too and code is rebased to latest main branch. Moving this out of draft state now.

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+
+struct PackItHeader {
+    /// Header Magic (PKIT)
+    magic: [u8; 4],
+    /// Header Size
+    header_size: u32,
+}
+
+impl PackItHeader {
+    pub const fn new() -> Self {
```
00xc, line 28: Since this structure is not pub, its methods don't need to be pub either. pub only affects visibility at the module level.

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut hdr = PackItHeader::new();
+        hdr.magic.copy_from_slice(&buf[0..4]);
+        hdr.header_size = u32::from_le_bytes(buf[4..8].try_into().unwrap());
+
+        hdr
+    }
+
+    pub fn check_magic(&self) -> bool {
+        self.magic == PACKIT_MAGIC
+    }
```
00xc, line 49: We should probably replace the `try_into()` for a returned SvsmError. On top of that, we could also check the magic bytes at the end of `load()`, returning an error if they do not match.

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+            hdr_type,
+            name_len,
+            file_size,
+            name: result.unwrap(),
+        }
+    }
+
+    pub fn valid(&self) -> bool {
+        self.hdr_type == 1 && self.name_len > 0
+    }
```
00xc, line 102: Same here. Since this function can fail in multiple places, perhaps we could return a `Result<Self, SvsmError>`. As before, we could move the check in `valid()` to this function.

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+        if result.is_err() {
+            return FileHeader::new();
+        }
+
+        FileHeader {
+            hdr_type,
+            name_len,
+            file_size,
+            name: result.unwrap(),
+        }
```
00xc, line 97: We can remove the unwrap with `let-else` syntax:

```rust
let Ok(name) = String::from_utf8(Vec::from(&buf[12..header_len])) else {
    return FileHeader::new();
};

FileHeader { hdr_type, name_len, file_size, name }
```

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    pub fn valid(&self) -> bool {
+        self.hdr_type == 1 && self.name_len > 0
+    }
+
+    pub fn file_name(&self) -> &String {
```
00xc, line 104: Returning `&str` here is a bit more conventional.

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+    let size = pend - pstart;
+
+    if size == 0 {
+        return Ok(());
+    }
+
+    log::info!("Unpacking FS archive...");
+
+    let guard = PerCPUPageMappingGuard::create(pstart.page_align(), pend.page_align_up(), 0)
+        .expect("Failed to map FS archive");
```
00xc, line 137: This is fine, but we could also use the question mark operator.

src/fs/init.rs
00xc: (Meant to select line 104 only)

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+    let mut current = hdr.len();
+    while current < size {
+        let fh = FileHeader::load(&data[current..]);
+        if !fh.valid() {
+            log::error!("Invalid data in FS archive");
+            return Err(SvsmError::FileSystem(FsError::inval()));
+        }
+
+        let start = current + fh.header_size();
+        let end = start + fh.file_size();
```
00xc, line 157: These look dangerous to me, especially the second one, for two reasons: they could go out of bounds in `data`, and they could overflow when adding them.

`current` comes from `PackItHeader::len()`, which returns `PackItHeader::header_size`, which is never validated (although it is bounded by `size` in this loop). `FileHeader::file_size()` returns `FileHeader::file_size + 12`, which is also not validated.

We should use a [safe slicing method](https://doc.rust-lang.org/std/primitive.slice.html#method.get) when using these values below, e.g.:

```rust
let file_data = data.get(start..end)
    .ok_or(SvsmError::FileSystem(FsError::inval()))?;
let written = file.write(file_data)?;
```

This would at least ensure that we do not go out of bounds, avoiding a panic. The values could still accidentally overflow I suspect, panicking in debug builds. If we want to be extra careful, we could use `checked_add`, e.g.:

```rust
let start = current.checked_add(fh.header_size())
    .ok_or(SvsmError::FileSystem(FsError::inval()))?;
let end = start.checked_add(fh.file_size())
    .ok_or(SvsmError::FileSystem(FsError::inval()))?;
```

Take these with a grain of salt, I have not tested whether they can actually overflow.

src/fs/init.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    pub fn load(buf: &[u8]) -> Self {
+        if buf.len() < 12 {
+            return FileHeader::new();
+        }
+        let hdr_type = u16::from_le_bytes(buf[0..2].try_into().unwrap());
+        let name_len = u16::from_le_bytes(buf[2..4].try_into().unwrap());
+        let file_size = u64::from_le_bytes(buf[4..12].try_into().unwrap());
+
+        let header_len: usize = (12 + name_len).into();
```
00xc, line 81: This can also overflow, since the addition is done before the conversion to `usize`.

00xc: So this should work fine:

```rust
let header_len = name_len as usize + 12;
```

src/fs/fs.rs
```diff
@@ -0,0 +1,465 @@
...
+    }
+}
+
+pub fn create(path: &str) -> Result<FileHandle, SvsmError> {
+    let mut path_items = split_path(path)?;
+    let file_name = FileName::from(path_items.pop().unwrap());
+    let current_dir = walk_path(&path_items)?;
+
+    if file_name.length() == 0 {
```
osteffenrh, line 211: Why this check here?

This should not happen, because `split_paths_allow_empty` filters these.

Somehow I would have expected this check in the lower level create function(s).
But if it is desired/needed here, then I would also expect it for `mkdir()`?

src/fs/init.rs
joergroedel: Okay, fixed that, thanks.

src/fs/init.rs
joergroedel: Changed the function to return Result<Self, SvsmError> and added the magic check at the end.

src/fs/init.rs
joergroedel: Done.

src/fs/init.rs
joergroedel: Looks much better, thanks.

src/fs/init.rs
joergroedel: Yes, makes sense, done.

src/fs/init.rs
joergroedel: Both checks make sense, I added them. Thanks.

src/fs/init.rs
joergroedel: Fixed, thanks.

src/fs/ramfs.rs
```diff
@@ -0,0 +1,263 @@
...
+            base_pages + 1
+        } else {
+            base_pages
+        };
+
+        while self.pages.len() > new_pages {
+            self.pages.pop();
+        }
+        self.capacity = new_pages * PAGE_SIZE;
+        self.size = size;
```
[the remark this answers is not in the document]
joergroedel: Right. I also added code to clear the pages that are removed from the file. Thanks.

src/boot_stage2.rs
```diff
@@ -197,7 +197,7 @@ global_asm!(
         shrq $3, %rcx
         rep stosq
 
-        popq    %rdi
+        movq %rsi, %rdi
```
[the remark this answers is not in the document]
joergroedel: Yes, the 32-bit registers are zero-extended, so this is guaranteed.

src/fs/ramfs.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        // Write second buffer at offset 4096 - 256 - cross-page write
+        let mut buf2 = [0xaau8; 512];
+        file.write(&mut buf2, PAGE_SIZE - 256)
+            .expect("Failed to write file cross-page");
+        assert!(file.size() == PAGE_SIZE + 256);
+
+        // Read back and check first buffer
+        let size = file
+            .read(&mut buf1, 0)
```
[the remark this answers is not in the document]
joergroedel: Fixed, thanks.

src/fs/ramfs.rs
```diff
@@ -0,0 +1,263 @@
...
+            .expect("Failed to read from offset 0");
+        assert!(size == 512);
+
+        for byte in buf1.iter() {
+            assert!(*byte == 0xff);
+        }
+
+        // Read back and check second buffer
+        let size = file
+            .read(&mut buf2, PAGE_SIZE - 256)
```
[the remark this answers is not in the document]
joergroedel: Fixed, thanks.

src/fs/fs.rs
joergroedel: Right, thanks, that check is redundant. I removed it.

joergroedel:
Updated the PR, all review comments received so far are now addressed.

src/fs/init.rs
```diff
@@ -0,0 +1,163 @@
...
+impl PackItHeader {
+    const fn new() -> Self {
+        PackItHeader {
+            magic: [0; 4],
+            header_size: 8,
+        }
+    }
+
+    fn load(buf: &[u8]) -> Result<Self, SvsmError> {
+        assert!(buf.len() >= 8);
```
00xc, line 36: I'd vote to return an error here instead to keep everything consistent.

src/fs/init.rs
```diff
@@ -0,0 +1,163 @@
...
+
+struct FileHeader {
+    name_len: u16,
+    file_size: u64,
+    name: String,
+}
+
+impl FileHeader {
+    fn load(buf: &[u8]) -> Result<Self, SvsmError> {
+        assert!(buf.len() >= 12);
```
00xc, line 65: Same here, returning an error seems more natural.

src/fs/init.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        if hdr_type != 1 || name_len == 0 {
+            log::error!("Invalid file header in archive");
+            return Err(SvsmError::FileSystem(FsError::inval()));
+        }
+
+        Ok(FileHeader {
+            name_len,
+            file_size,
+            name: name,
```
00xc, line 92: Field could be omitted here, clippy will complain.

00xc: Also we could use `Self`. Just a nitpick 😄

src/fs/init.rs
joergroedel: I changed it to an assert because it checks for a violation in the function input contract, which is a programmer error if it happens. The assert gives us the exact place in the code which makes such issues faster to track down.

src/fs/init.rs
joergroedel: Same here.

src/fs/init.rs
00xc: But the size of the slice comes directly from the size of the region passed down to `populate_ram_fs()`, so I'd say this does not come from a programmer error directly, no?

joergroedel: Yeah, okay, it can happen with a wrongly packed archive file. I changed to to log an error in this case and return Err. Same for the other case below.

00xc (approved):

PR #28, opened 2023-05-04
Make virtualrange allocators per-cpu

joergroedel:
The virtualrange allocators are global, but the address-space areas it covers are per-cpu. Change it to make the allocators per-cpu too.

### DIFF
--- a/src/cpu/percpu.rs
+++ b/src/cpu/percpu.rs
@@ -16,14 +16,16 @@ use crate::locking::{LockGuard, RWLock, SpinLock};
 use crate::mm::alloc::{allocate_page, allocate_zeroed_page};
 use crate::mm::pagetable::{get_init_pgtable_locked, PageTable, PageTableRef};
 use crate::mm::stack::{allocate_stack_addr, stack_base_pointer};
+use crate::mm::virtualrange::VirtualRange;
 use crate::mm::{
-    virt_to_phys, SVSM_PERCPU_BASE, SVSM_PERCPU_CAA_BASE, SVSM_PERCPU_VMSA_BASE,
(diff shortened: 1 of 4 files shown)


src/cpu/percpu.rs
```diff
@@ -467,6 +479,20 @@ impl PerCpu {
...
+        let page_count = (SVSM_PERCPU_TEMP_END_4K - SVSM_PERCPU_TEMP_BASE_4K) / PAGE_SIZE;
+        assert!(page_count <= VirtualRange::CAPACITY);
+        self.vrange_4k
+            .init(VirtAddr::from(SVSM_PERCPU_TEMP_BASE_4K), page_count);
+
+        // Initialize 2M range
+        let page_count = (SVSM_PERCPU_TEMP_END_2M - SVSM_PERCPU_TEMP_BASE_2M) / PAGE_SIZE_2M;
+        assert!(page_count <= VirtualRange::CAPACITY);
+        self.vrange_2m
+            .init(VirtAddr::from(SVSM_PERCPU_TEMP_BASE_2M), page_count);
```
roy-hopkins, line 494: I've just noticed a bug (my bug) in the VirtualRange initialisation. `VirtualRange::alloc()` returns a contiguous virtual address range based on the allocated page count therefore it needs to know the page size. At the moment it incorrectly assumes it's 4k. Therefore the `init()` function will need to take the page size, or at least the page shift as a parameter.

I can fix this in a separate PR if you would prefer?

roy-hopkins (commented):
Looks great.

src/cpu/percpu.rs
joergroedel: Yes, please send a separate PR for this fix, ideally based on this PR. I will merge this later today.

PR #29, opened 2023-05-04
Add page shift size to virtual allocator to fix issue with multi-page 2M allocations.

roy-hopkins:
The virtual address bitfield allocator allocates a range based on the number of pages and the page size. The page size was always assumed to be 4K, so when the allocator was used with 2M pages an incorrect range was returned. This PR fixes that issue.


joergroedel (approved):
Looks good now and all tests passed. Thanks

PR #36, opened 2023-05-11
Add Github actions

00xc:
Add automated actions to be executed on every push to the repository and every pull request. At the moment, every step of the build process is done except the final linking, because the runner image (ubuntu-latest) is on Ubuntu 22.04, which does not have a new enough binutils version. The actions run all the tests via make test, as well as cargo fmt to keep the code format consistent.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,43 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Install latest nightly
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: nightly
+            profile: minimal
+            override: true
+            components: rustfmt, rust-src
+
+      # ubuntu-latest does not have binutils 2.39, which we need for
+      # ld to work, so build all the objects without performing the
+      # final linking step.
+      - name: Build
+        run: make stage1/stage1.o stage1/reset.o
+
+      - name: Run tests
+        run: make test
+
+      - name: Format
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check


00xc:
> About the binutils problem:
> 
> One could [run the jobs in a container](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container), with a suitable image (whatever that could be).
> 
> Building a custom image from a dockerfile and maintaining that on ghcr.io requires some effort.

Yes, I also very briefly looked into [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners), but they also seem overkill.

> Ubuntu 23.04 for example has binutils 2.40, but there I get some linker error related to PIE

Hm, interesting, it works fine on my machine with binutils 2.39 (openSUSE Leap 15.4).

osteffenrh:
Quoting Carlos López (2023-05-11 14:22:51)
[...]
>     Ubuntu 23.04 for example has binutils 2.40, but there I get some linker
>     error related to PIE
>
> Hm, interesting, it works fine on my machine with binutils 2.39 (openSUSE Leap
> 15.4).
>

It works on Fedora 38 (binutils-2.39-9), but not on Fedora 37 (binutils
too old).

Here is it complains about on Ubuntu 23.04 when I run `make`:
```
[...]
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
objcopy -O elf64-x86-64 --strip-unneeded
"target/svsm-target/"debug"/svsm" stage1/kernel.elf
cc    -c -o stage1/stage1.o stage1/stage1.S
cc -o stage1/stage1 stage1/stage1.o stage1/reset.o -nostdlib
-Wl,--build-id=none -Wl,-Tstage1/stage1.lds
/usr/bin/ld: stage1/stage1.o: relocation R_X86_64_32 against
`.startup.text' can not be used when making a PIE object; recompile
with -fPIE
/usr/bin/ld: failed to set dynamic section sizes: bad value
collect2: error: ld returned 1 exit status
make: *** [Makefile:41: stage1/stage1] Error 1
```

joergroedel:
Nice work, thanks. As a next step, can Github actions also do per-commit checks? I'd like to check commits for author Signed-off-by's an other style issues.

00xc:
> Nice work, thanks. As a next step, can Github actions also do per-commit checks? I'd like to check commits for author Signed-off-by's an other style issues.

Yes, we can add arbitrary scripts to do this kind of thing. I'll look into it.

osteffenrh:
There is the GitHub DCO App that can do that: https://probot.github.io/apps/dco/
It can be added to a repo by just a few clicks.

Example PRs:
 - Not signed off: https://github.com/osteffenrh/coconut-svsm/pull/1
 - Signed off: https://github.com/osteffenrh/coconut-svsm/pull/2

There is also a python script ([dco-check](https://github.com/christophebedard/dco-check)) as an alternative.

PR #50, opened 2023-06-19
Tasks: Add support for task context switching

roy-hopkins:
Draft PR for review and comments.

Implements a mechanism to switch the current CPU between different task contexts that include per-task pagetables and will in the future support per-task virtual memory mapping.

The current version creates a global task list and launches an initial task on each virtual CPU. Additional tasks can be created and added to the task list which will then be activated by a subsequent call to schedule(). This has been tested locally but is not currently called within this PR.

### DIFF
--- a/src/cpu/percpu.rs
+++ b/src/cpu/percpu.rs
@@ -25,6 +25,7 @@ use crate::mm::{
 use crate::sev::ghcb::GHCB;
 use crate::sev::utils::RMPFlags;
 use crate::sev::vmsa::{allocate_new_vmsa, VMSASegment, VMSA};
+use crate::task::TaskPointer;
 use crate::types::{PAGE_SHIFT, PAGE_SHIFT_2M, PAGE_SIZE, PAGE_SIZE_2M, SVSM_TR_FLAGS, SVSM_TSS};
 use alloc::vec::Vec;
 use core::cell::UnsafeCell;
@@ -182,6 +183,8 @@ pub struct PerCpu {
     pub vrange_4k: VirtualRange,
     /// Address allocator for per-cpu 2m temporary mappings
     pub vrange_2m: VirtualRange,
+
+    pub current_task: Option<TaskPointer>,
 }
 
 impl PerCpu {
@@ -199,6 +202,7 @@ impl PerCpu {
             reset_ip: 0xffff_fff0u64,
             vrange_4k: VirtualRange::new(),
             vrange_2m: VirtualRange::new(),
+            current_task: None,
         }
     }
 
--- a/src/cpu/registers.rs
+++ b/src/cpu/registers.rs
@@ -5,32 +5,7 @@
 // Author: Roy Hopkins <rhopkins@suse.de>
 
 #[repr(C, packed)]
-pub struct X86Regs {
-    pub r15: usize,
-    pub r14: usize,
-    pub r13: usize,
-    pub r12: usize,
-    pub r11: usize,
-    pub r10: usize,
-    pub r9: usize,
-    pub r8: usize,
-    pub rbp: usize,
-    pub rdi: usize,
-    pub rsi: usize,
-    pub rdx: usize,
-    pub rcx: usize,
-    pub rbx: usize,
-    pub rax: usize,
-    pub vector: usize,
-    pub error_code: usize,
-    pub rip: usize,
-    pub cs: usize,
-    pub flags: usize,
-    pub rsp: usize,
-    pub ss: usize,
-}
-
-#[repr(C, packed)]
+#[derive(Default)]
 pub struct X86GeneralRegs {
     pub r15: usize,
     pub r14: usize,
@@ -50,6 +25,7 @@ pub struct X86GeneralRegs {
 }
 
 #[repr(C, packed)]
+#[derive(Default)]
 pub struct X86SegmentRegs {
     pub cs: usize,
     pub ds: usize,
--- a/src/cpu/registers.rs
+++ b/src/cpu/registers.rs
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Roy Hopkins <rhopkins@suse.de>
+
+#[repr(C, packed)]
+pub struct X86Regs {
+    pub r15: usize,
+    pub r14: usize,
+    pub r13: usize,
+    pub r12: usize,
+    pub r11: usize,
+    pub r10: usize,
+    pub r9: usize,
+    pub r8: usize,
+    pub rbp: usize,
+    pub rdi: usize,
+    pub rsi: usize,
+    pub rdx: usize,
+    pub rcx: usize,
+    pub rbx: usize,
+    pub rax: usize,
+    pub vector: usize,
+    pub error_code: usize,
+    pub rip: usize,
+    pub cs: usize,
+    pub flags: usize,
+    pub rsp: usize,
+    pub ss: usize,
+}
+
+#[repr(C, packed)]
+pub struct X86GeneralRegs {
+    pub r15: usize,
+    pub r14: usize,
+    pub r13: usize,
+    pub r12: usize,
+    pub r11: usize,
+    pub r10: usize,
+    pub r9: usize,
+    pub r8: usize,
+    pub rbp: usize,
+    pub rdi: usize,
+    pub rsi: usize,
+    pub rdx: usize,
+    pub rcx: usize,
+    pub rbx: usize,
+    pub rax: usize,
+}
+
+#[repr(C, packed)]
+pub struct X86SegmentRegs {
+    pub cs: usize,
+    pub ds: usize,
+    pub es: usize,
+    pub fs: usize,
+    pub gs: usize,
+    pub ss: usize,
+}
+
+#[repr(C, packed)]
+pub struct X86InterruptFrame {
+    pub rip: usize,
+    pub cs: usize,
+    pub flags: usize,
+    pub rsp: usize,
+    pub ss: usize,
+}
--- a/src/cpu/tsc.rs
+++ b/src/cpu/tsc.rs
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Roy Hopkins <rhopkins@suse.de>
+
+use core::arch::asm;
+
+pub fn rdtsc() -> u64 {
+    let eax: u32;
+    let edx: u32;
+
+    unsafe {
+        asm!("rdtsc",
+             out("eax") eax,
+             out("edx") edx,
+             options(att_syntax));
+    }
+    (eax as u64) | (edx as u64) << 32
+}
--- a/src/task/schedule.rs
+++ b/src/task/schedule.rs
@@ -1,0 +1,210 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Roy Hopkins <rhopkins@suse.de>
+
+extern crate alloc;
+
+use core::cell::RefCell;
+
+use super::Task;
+use super::{tasks::TaskRuntime, TaskState, INITIAL_TASK_ID};
+use crate::cpu::percpu::{this_cpu, this_cpu_mut};
+use crate::error::SvsmError;
+use crate::locking::SpinLock;
+use alloc::boxed::Box;
+use alloc::rc::Rc;
+use intrusive_collections::{intrusive_adapter, Bound, KeyAdapter, RBTree, RBTreeLink};
+
+pub type TaskPointer = Rc<TaskNode>;
+
+pub struct TaskNode {
+    link: RBTreeLink,
+    pub task: RefCell<Box<Task>>,
+}
+
+intrusive_adapter!(pub TaskNodeAdapter = Rc<TaskNode>: TaskNode { link: RBTreeLink });
+impl<'a> KeyAdapter<'a> for TaskNodeAdapter {
+    type Key = u64;
+    fn get_key(&self, node: &'a TaskNode) -> u64 {
+        node.task.borrow().runtime.value()
+    }
+}
+
+pub struct TaskRBTree {
+    tree: Option<RBTree<TaskNodeAdapter>>,
+}
+
+impl TaskRBTree {
+    pub fn tree(&mut self) -> &mut RBTree<TaskNodeAdapter> {
+        self.tree
+            .get_or_insert_with(|| RBTree::<TaskNodeAdapter>::new(TaskNodeAdapter::new()))
+    }
+
+    pub fn get_task(&self, id: u32) -> Option<Rc<TaskNode>> {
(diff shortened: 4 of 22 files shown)


src/task/tasklist.rs
```diff
@@ -0,0 +1,186 @@
...
+
+use super::INITIAL_TASK_ID;
+use super::{task::TaskState, Task};
+
+pub static TASKLIST: SpinLock<TaskList> = SpinLock::new(TaskList::new());
+
+#[derive(Copy, Clone)]
+pub struct TaskList {
+    pub root: Option<*mut Task>,
+}
```
00xc, line 19: I have not tested this at all, but could we use the [LinkedList](https://doc.rust-lang.org/alloc/collections/linked_list/struct.LinkedList.html) type in the alloc module? It would save us a lot of unsafe code.

src/task/tasklist.rs
```diff
@@ -0,0 +1,186 @@
...
+        }
+        if unsafe { (*task).state } != TaskState::RUNNING {
+            panic!("Attempt to launch a non-running initial task");
+        }
+        unsafe { (*task).state = TaskState::SCHEDULED };
+        this_cpu_mut().current_task = Some(task);
+        unsafe { (*task).set_current(None) };
+    }
+
+    pub fn schedule() {
```
00xc, line 86: This should probably be a standalone function, no? I find it quite confusing that this is an associated function on `TaskList`, but takes a lock on `TASKLIST`, which itself is a lock around a `TaskList` instance.

src/cpu/percpu.rs
```diff
@@ -182,6 +183,8 @@ pub struct PerCpu {
     pub vrange_4k: VirtualRange,
     /// Address allocator for per-cpu 2m temporary mappings
     pub vrange_2m: VirtualRange,
+
+    pub current_task: Option<*mut Task>,
```
00xc, line 187: If we are guarding against NULL with an `Option` we could use a [`ptr::NonNull`](https://doc.rust-lang.org/core/ptr/struct.NonNull.html) to improve struct layout.

src/task/tasklist.rs
roy-hopkins: Good point. I looked into using a LinkedList or something similar but there seemed to be some challenges making it work in Rust so I (temporarily) opted for a simple but messy solution. However, I'll go back and see if I can find something safer and more suitable.

src/task/tasklist.rs
roy-hopkins: Yes you're right. But also a couple of the other functions should also be standalone. I'll fix this.

src/cpu/percpu.rs
00xc: Can't we use None to detect that initial state?

roy-hopkins:
I've just pushed V2 of this draft PR which has some significant changes to the previous version. In particular I wanted to address the use of a self-implemented linked list, as commented on above and also add the ability to support more advanced scheduling from the outset. So I've made the following big changes:

* Replaced the linked list in task.rs and tasklist.rs with an RB tree implemented using an imported crate in a new file: schedule.rs
* Used Box to allocate memory and manage the lifetime of the Task structures.
* Let the task tree (in TASKS) own the tasks themselves but use reference counting for extending the lifetime where necessary. Note here though that if an additional reference to a task is held then it is not possible to mutate the task during scheduling and a runtime panic will be generated.
* Use the sorting of the RB tree on CPU usage to determine the next task to schedule
* Implement two different ways of measuring CPU usage: switch count and rdtsc.

The memory management/borrowing/reference counting in this implementation was particularly tricky to get right. I'd appreciate it if reviewers could pay particular attention to this area of task ownership.

src/task/task.rs
```diff
@@ -0,0 +1,404 @@
...
+        let new_task_addr = (self as *mut Task) as u64;
+
+        // Keep track of when the new task was scheduled in
+        self.runtime.schedule_in();
+
+        // Switch to the new task
+        unsafe {
+            asm!(
+                r#"
+                    int     $0x20
```
joergroedel, line 255: Okay, as I understand this, the code switches a task via a software interrupt to simplify switching the contents of the GPRs. The GPRs are then stored/restored in/from the struct task.

Can we instead write some assembly which stores the GPRs directly on the task stack, switches the stack pointer, and restores the GPRs? This would avoid the overhead of having a separate context store and the software IRQ.

joergroedel (commented):
In general this looks good to me. The task switching code is a good start and the beginnings of per-task kernel memory are also there. As one of the next steps we probably need an address allocator to manage the kernel task dynamic memory and the user-space address layout as well. I will look into this going forward.

src/cpu/registers.rs
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Roy Hopkins <rhopkins@suse.de>
+
+#[repr(C, packed)]
+pub struct X86Regs {
```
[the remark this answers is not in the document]
roy-hopkins: Yes. I'll remove it.

src/cpu/tsc.rs
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Roy Hopkins <rhopkins@suse.de>
+
+use core::arch::asm;
+
+pub fn rdtsc() -> u64 {
```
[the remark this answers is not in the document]
roy-hopkins: I was in two minds about this - renaming the file msr.rs/new file/just putting it in there anyway. I'll move it to msr.rs.

src/task/task.rs
roy-hopkins: It wasn't so much to simplify the switching of the contents of the GPRs. It was more to prove that switching via an ISR was possible for the case if/we want preemptive switching via a timer interrupt. It's fairly trivial to switch it away from using a software interrupt so I'll make the change.

The GPRs are stored on the stack in this implementation. For convenience (for the gdbstub) a pointer to the context on the stack is maintained to find it easy to inspect the registers.

roy-hopkins:
> As one of the next steps we probably need an address allocator to manage the kernel task dynamic memory and the user-space address layout as well. I will look into this going forward.

I haven't started thinking about user-space address layout at the moment but I have used the current bitmap allocator to implement a per task kernel address allocator in my WIP PR for module loading in my own fork. https://github.com/roy-hopkins/svsm/pull/1.

src/task/schedule.rs
```diff
@@ -0,0 +1,230 @@
...
+pub struct TaskRBTree {
+    tree: Option<RBTree<TaskNodeAdapter>>,
+}
+
+impl TaskRBTree {
+    pub fn tree(&mut self) -> &mut RBTree<TaskNodeAdapter> {
+        if self.tree.is_none() {
+            self.tree = Some(RBTree::<TaskNodeAdapter>::new(TaskNodeAdapter::new()));
+        }
+        self.tree.as_mut().unwrap()
```
00xc, line 56: This saves us an unwrap:

```rust
self.tree.get_or_insert_with(|| RBTree::new(TaskNodeAdapter::new()))
```

src/task/schedule.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    pub fn get_task(&mut self, id: u32) -> Option<Rc<TaskNode>> {
+        let mut cursor = self.tree().front();
+        while cursor.get().is_some() {
+            if cursor.get().unwrap().task().id == id {
+                return Some(cursor.clone_pointer().unwrap());
+            }
+            cursor.move_next();
+        }
+        None
```
00xc, line 67: This saves us an unwrap:

```rust
let mut cursor = self.tree().front();
while let Some(node) = cursor.get() {
    if node.task().id == id {
        return Some(cursor.clone_pointer().unwrap())
    }
    cursor.move_next();
}
None
```

src/task/task.rs
```diff
@@ -0,0 +1,404 @@
...
+            state: TaskState::RUNNING,
+            affinity: None,
+            id: TASK_ID_ALLOCATOR.next_id(),
+            runtime: TaskRuntimeImpl::new(),
+            next: core::ptr::null_mut(),
+        });
+        Ok(task)
+    }
+
+    pub fn set_current(&mut self, previous_task: Option<*mut Task>) {
```
00xc, line 237: As before, `Option<*mut Task>` could perhaps be switched to `Option<NonNull<Task>>`, since we use `None` to signify null.

00xc: Also, does it need to be `*mut`?

roy-hopkins: Yes. In fact, I've just working on this and have changed this to use a raw pointer instead as the null pointer is handled in the assembly code in `set_current()`.

src/task/schedule.rs
```diff
@@ -0,0 +1,230 @@
...
+    let task_ptr = {
+        let mut task = task_node.task_mut();
+        if task.state != TaskState::RUNNING {
+            panic!("Attempt to launch a non-running initial task");
+        }
+        this_cpu_mut().current_task = Some(task_node.clone());
+        task.state = TaskState::SCHEDULED;
+        task.as_mut() as *mut Task
+    };
+    unsafe { (*task_ptr).set_current(None) };
```
00xc, line 92: Not a huge fan of this unsafe, I wonder if we could remove it. If `Task::set_current()` took `&self` instead of `&mut self` we could avoid it I think. There are two things that prevent this:

1. This line in `Task::set_current()`:
```rust
let new_task_addr = (self as *mut Task) as u64;
```
But perhaps this could be `*const`?

2. The fact that `TaskRuntime::schedule_in()` takes `&mut self` (and therefore `CountRuntime::schedule_in()`. If we could make the trait signature take `&self` that would be great for this. For `CountRuntime` this is trivial, we simply need to use an `AtomicU64` instead of an `u64`:
```rust
pub struct CountRuntime {
    count: AtomicU64,
}

impl TaskRuntime for CountRuntime {
    fn schedule_in(&mut self) {
        self.count.fetch_add(1, Ordering::SeqCst);
    }
    ...
}
```
I'm unsure if we could use a more relaxed ordering though. For `TscRuntime` we could do the same via `AtomicU64::store()`

src/task/task.rs
```diff
@@ -0,0 +1,404 @@
...
+    pub fn new() -> Self {
+        Self { runtime: 0 }
+    }
+}
+
+/// Tracks task runtime based on the number of times the task has been
+/// scheduled
+pub struct CountRuntime {
+    count: u64,
+}
```
00xc, line 142: Perhaps this should be `#[repr(transparent)]` so it's guaranteed that it doesn't affect the layout of `Task`, which is `#[repr(C)]`.

src/task/task.rs
```diff
@@ -0,0 +1,404 @@
...
+    pub affinity: Option<u32>,
+
+    /// ID of the task
+    pub id: u32,
+
+    /// Amount of CPU resource the task has consumed
+    pub runtime: TaskRuntimeImpl,
+
+    /// Next task in the task list
+    pub next: *mut Task,
```
00xc, line 209: Unused, right?

src/task/task.rs
roy-hopkins: The C layout for the task structure is to allow the assembly code to access only the first couple of members. The remaining layout of the task structure is unimportant. This isn't actually very obvious or clear in the code though so I agree it makes sense to add `#[repr(transparent)]`.

src/task/schedule.rs
roy-hopkins: This is a tricky one. Certainly 2. is easy - in fact I've just moved `TaskRuntime::schedule_in()` from this function in the the scheduler itself, so that solves that.

1. Is a bit harder as `set_current()` does indeed mutate the task in that the contents will certainly have changed by the time the function returns, even if it happens in the context of another task.

The bigger problem though is that the task structure itself uses runtime borrow checking. Therefore to call it without a pointer means having an active borrow which remains in scope until the task context is resumed. This prevents the scheduler (and the gdbstub) from being able to access the task.

src/cpu/percpu.rs
roy-hopkins: The `current_task` now is an `Option<Rc<TaskNode>>` instead of a pointer so the comment no longer applies.

src/task/schedule.rs
```diff
@@ -0,0 +1,214 @@
...
+    Ok(node_ret)
+}
+
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub fn close_task(task: TaskPointer) -> Result<(), SvsmError> {
+    // All tasks are protected via the task tree lock
+    let mut tl = TASKS.lock();
+    let mut cursor = unsafe {
+        tl.tree()
+            .cursor_mut_from_ptr(task.as_ref() as *const TaskNode)
```
00xc, line 109: Just a nitpick, but the reference will automatically be coerced to a pointer, this would do:

```
tl.tree()
    .cursor_mut_from_ptr(task.as_ref())
```

Or if you want to be more explicit:

```
tl.tree()
    .cursor_mut_from_ptr(Rc::as_ptr(&task))
```

src/task/schedule.rs
```diff
@@ -0,0 +1,214 @@
...
+        None => id == INITIAL_TASK_ID,
+    }
+}
+
+fn update_current_task(tree: &mut RBTree<TaskNodeAdapter>) -> TaskPointer {
+    // This function leaves the CPU in an invalid state as it takes the current
+    // task, replacing it with None. The caller must assign a new task or reassign
+    // the current task to the CPU before resuming.
+    let task_ptr = this_cpu_mut().current_task.take().unwrap();
+    let mut task_cursor = unsafe { tree.cursor_mut_from_ptr(task_ptr.as_ref() as *const TaskNode) };
```
00xc, line 133: Same thing here about the reference to pointer coercion.

src/task/schedule.rs
```diff
@@ -0,0 +1,214 @@
...
+    assert!(
+        !task_cursor.is_null(),
+        "The current task is not in the task tree"
+    );
+    let task_node = task_cursor.remove().unwrap();
```
00xc, line 138: This assert can be removed since the unwrap() already would cause a panic. You can use expect to add a message to the panic.

src/task/schedule.rs
```diff
@@ -0,0 +1,214 @@
...
+        // Find the task with the lowest runtime that is eligible to run
+        // on this CPU
+        let mut cursor = tree.lower_bound(Bound::Included(&0));
+        while cursor.get().is_some() {
+            let candidate_task = cursor.get().unwrap().task.borrow();
+            // If the runtime of the candidate task is greater than the task that was
+            // current running on the CPU then don't switch
+            if candidate_task.state == TaskState::RUNNING
+                && (candidate_task.affinity.is_none()
+                    || (candidate_task.affinity.unwrap() == this_cpu().get_apic_id()))
```
00xc, line 179: We can use `Option::filter()` here as well to avoid the unwrap. Or as an alternative, we can use `map()` + `unwrap_or()`:

```rust
if candidate_task.state == TaskState::RUNNING
    && candidate_task
        .affinity
        .map(|af| af == this_cpu().get_apic_id())
        .unwrap_or(true)
{
    break;
}
```

roy-hopkins:
Thanks to the people that looked at and reviewed this. I'll close this draft now and raise new PRs based on this work.

PR #52, opened 2023-06-19
Implement a log buffer to store log messages.

### LINKED ISSUE #12: Implement a log buffer to keep log in memory
The SVSM prints all log messages directly to the serial console. In a production environment this could be a security problem as it reveals information about the guest.

Implement a log buffer which stores all log messages in memory. On boot the log messages are stored on a single page and later moved to a `File` structure. The `File` can optionally be made visible through the COCONUT file system.

The log buffer must use line buffering by collecting output per `Task` until a full line can be flushed to the common `File` buffer. Such a line buffer already exists for user-mode printouts, this can be extended to have a separate line buffer for the kernel prints as well.

For debug builds additionally print the messages to the console. For release builds do not print anything to the console by default. Add a flag to set in the code to re-enable console printing also for release builds to make debugging issues easier.

Stage2 can be ignored for this work, having an implementation in the kernel is sufficient for now.

vsntk18:
Implemented a  global buffer to store log messages and passed this buffer from stage2 to svsm kernel. Replaced console logging with buffer logging.

Percpu line buffer will probably be necessary, and  is not yet implemented.   As a rust newbee, I appreciate your feedback on if I am using the right data structures in the right way.  Thanks!

Fixes: https://github.com/coconut-svsm/svsm/issues/12


00xc:
Probably a good idea to rebase from the main branch to fix the cargo fmt checks. Also, make sure your author email matches your Signed-off-by email!

vsntk18:
> Probably a good idea to rebase from the main branch to fix the cargo fmt checks. Also, make sure your author email matches your Signed-off-by email!
> 
> EDIT: nevermind, no need for a rebase, but the fmt checks should be fixed.

sorry, will fix the fmt checks. I was under the impression that commit hooks report the deviations.

src/log_buffer.rs
```diff
@@ -0,0 +1,292 @@
...
+
+#[inline]
+fn encode_state(full: usize, rc: usize, wc: usize, head: usize, tail: usize) -> usize {
+    let t = tail & TAIL_MASK;
+    let h = (head << 16) & HEAD_MASK;
+    let w = (wc << 32) & WRITER_MASK;
+    let r = (rc << 48) & READER_MASK;
+    let f = (full << 63) & FULL_MASK;
+    t | h | w | r | f
+}
```
00xc, line 50: Can we get this into a struct? Something like:

```rust
#[derive(Clone, Copy, Debug)]
struct LogBufferState {
    full: usize,
    rc: usize,
    ...
}

impl LogBufferState {
    fn encode(self) -> usize { ... }
    fn decode(enc: usize) -> Self { ... }
}
```

Then we can do:

```rust
let pos = self.state.load(Ordering::Acquire);
let state = State::decode(pos);
...
let new_pos = state.encode();
if self
    .state
    .compare_exchange(pos, new_pos, Ordering::Acquire, Ordering::Relaxed)
    .is_ok()
{
    break;
}
```

It would be even more conventional to implement encoding and decoding via the From trait:

```rust
impl From<usize> for LogBufferState { ... }
impl From<LogBufferState> for usize { ... }
```

But that's not strictly necessary.

src/log_buffer.rs
```diff
@@ -0,0 +1,292 @@
...
+
+    pub fn write_log(&mut self, s: &str) {
+        let mut pos;
+        let mut new_pos;
+        let mut head;
+        let mut new_head;
+        let mut tail;
+        let mut reader_count;
+        let mut writer_count;
+        let mut is_full;
```
00xc, line 77: This is C-style coding, it's very uncharacteristic and best avoided in my opinion.

00xc:
> sorry, will fix the fmt checks. I was under the impression that commit hook reports the deviations.

The git hook should warn you, but won't fix the issues automatically.

src/log_buffer.rs
00xc: Unless the value of a variable needs to be kept for the next loop iteration, I'd say just shadow them by using `let`.

src/svsm.rs
```diff
@@ -310,9 +312,10 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
 }
 
 #[no_mangle]
-pub extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: VirtAddr) {
+pub extern "C" fn svsm_start(li: &KernelLaunchInfo,  mi: &MigrateInfo) {
     let launch_info: KernelLaunchInfo = *li;
-    let vb_ptr = vb_addr.as_mut_ptr::<u64>();
+    let migrate_info: &MigrateInfo = mi;
```
00xc, line 317: This is not necessary, right? We can simply use `mi`.

src/log_buffer.rs
vsntk18: Yes, using a struct makes much more sense.  Also implemented encoding and decoding via From train.

src/log_buffer.rs
vsntk18: Yes, it's really ugly.  Got rid of it.

src/svsm.rs
vsntk18: Removed this redundant variable.

vsntk18:
Reworked the findings so far.
Added percpu line buffer implementation.

vsntk18:
Status so far on this PR:
1. Implemented a global log buffer
2.  Added initialization and migration code for the log buffer.
3.  Implemented a percpu line buffer
4. Added a  feature called enable-console-log to print log messages to console.

src/line_buffer.rs
```diff
@@ -0,0 +1,128 @@
+#[cfg(feature = "enable-console-log")]
```
joergroedel, line 1: For new files, please add our default file header as present in other source files.

src/line_buffer.rs
```diff
@@ -0,0 +1,128 @@
...
+use core::fmt;
+use core::fmt::Write;
+
+const LINE_BUFFER_SIZE: usize = 256;
+
+#[derive(Debug)]
+pub struct LineBuffer {
+    buf: [u8; LINE_BUFFER_SIZE],
+    head: usize,
+    tail: usize,
```
joergroedel, line 16: A line buffer does not need head and tail pointers, a len is enough. Maybe you can use FixedString for it? The idea is to copy the contents of the line buffer to the global log once a newline character is encountered or the line buffer is full.

src/log_buffer.rs
```diff
@@ -0,0 +1,220 @@
...
+            tail: state & TAIL_MASK,
+            head: (state & HEAD_MASK) >> 16,
+            wc: (state & WRITER_MASK) >> 32,
+            rc: (state & READER_MASK) >> 48,
+            full: (state & FULL_MASK) >> 63,
+        }
+    }
+}
+
+impl From<LogBufferState> for usize {
```
joergroedel, line 74: When we use locking there is no need to keep all state in a usize, you can just track separate usize members for head and tail. The wc and rc members are probably for locking, please use the RWLock or SpinLock implementations already present in the code-base.

joergroedel (changes requested):
Provided some comments in the reviews. Please also move the logbuffer specific code into a separate sub-directory/module.

src/log_buffer.rs
vsntk18: Wouldn't RWLock be too restrictive here?   It allows either multiple readers or a single writer in the critical section.

But keeping all state in a usize, we could allow either multiple writers or multiple readers (ok, multiple readers are probably rare), if I understand correctly.

src/line_buffer.rs
vsntk18: I had thought of using FixedString.  Since it is implemented as a char array than a u8 array, the buffer can store only 1/4 of the size with u8 array.   Is there a better reason to use FixedString here?

src/line_buffer.rs
```diff
@@ -0,0 +1,128 @@
...
+            if b == 0xa {
+                // write to global log buffer when '\n' character is encountered
+                if self.tail <= self.head {
+                    let st = core::str::from_utf8(&self.buf[self.tail..self.head]).unwrap();
+                    unsafe { LB.write_log(st) };
+                } else {
+                    let st1 = core::str::from_utf8(&self.buf[self.tail..]).unwrap();
+                    let st2 = core::str::from_utf8(&self.buf[..self.head]).unwrap();
+                    let st = &[st1, st2].concat();
+                    unsafe { LB.write_log(st) };
```
00xc, line 41: Since the line buffer and the global buffer are backed by u8s perhaps could skip all this intermediate utf8 checking via `core::str::from_utf8`

src/line_buffer.rs
joergroedel: The char type has an encode_utf8() method which gives you an u8 slice to store in the log buffer. This way we can buffer per-cpu (or per-task once we have tasks) as chars and write it as u8 to the global log buffer.

src/log_buffer.rs
joergroedel: Lets keep things simple for now and switch to a lock implementation that we already have. The log-buffer is not performance critical, and once the system is running writes and reads are rare.

src/line_buffer.rs
vsntk18: ok, I will adapt the pre-commit hook to check for this.

vsntk18:
> Provided some comments in the reviews. Please also move the logbuffer specific code into a separate sub-directory/module.

vsntk18:
In this PR:
1. Moved log buffer implementation to separate directory.
2. Moved line buffer implementation to src/cpu directory.
3.  Reworked line buffer to use len member instead of head and tail.  However kept the implementation as u8 array not as a FixedString,  since it seemed simple.  May be this is still arguable.
4. skipped all the intermediate utf8 checking as suggested by Carlos.

src/cpu/line_buffer.rs
```diff
@@ -0,0 +1,119 @@
...
+            buf: FixedString::new(),
+        }
+    }
+
+    pub fn write_buffer(&mut self, s: &str) {
+        for c in s.chars() {
+            self.buf.push(c);
+            if c == '\n' || self.buf.length() == LINE_BUFFER_SIZE {
+                // when buffer is full or '\n' character is encountered
+                let mut buf_str: [u8; LINE_BUFFER_SIZE * 4] = [0; LINE_BUFFER_SIZE * 4];
```
00xc, line 35: This could simply be:

```rust
let mut buf_str = [0u8; LINE_BUFFER_SIZE * 4];
```

src/string.rs
```diff
@@ -36,12 +38,18 @@ impl<const T: usize> FixedString<T> {
     pub fn length(&self) -> usize {
         self.len
     }
+
+    pub fn reset(&mut self) {
+        self.len = 0;
+        self.current = 0;
+    }
```
00xc, line 45: This should probably be named `clear()`, just like the `String` method:
https://doc.rust-lang.org/std/string/struct.String.html#method.clear

src/string.rs
```diff
@@ -104,6 +112,20 @@ impl<const T: usize> fmt::Display for FixedString<T> {
...
+    fn next(&mut self) -> Option<Self::Item> {
+        let ret = if self.current == self.len {
+            None
+        } else {
+            Some(self.data[self.current])
+        };
+        self.current += 1;
+        ret
+    }
+}
```
00xc, line 127: This is overkill in my opinion. We can remove the `current` field and the `reset()` method by simply doing:

```rust
impl<const T: usize> FixedString<T> {
    ...
    pub fn iter(&self) -> impl Iterator<Item = char> + '_ {
        self.data.iter().take(self.len).copied()
    }
}
```

This would require doing `for ch in self.buf.iter()` instead in `LineBuffer::write_buffer()`.

00xc: Or rather than removing `reset()` I would replace it with `clear()p  and have it set `self.len` to 0, since I think that's necessary for the `LineBuffer` logic.

Cargo.toml
```diff
@@ -27,6 +27,8 @@ packit = { git = "https://github.com/coconut-svsm/packit", version = "0.1.0" }
 [build-dependencies]
 
 [features]
-default = ["enable-stacktrace"]
+default-release = ["enable-stacktrace"]
+default-debug = ["enable-stacktrace", "enable-console-log"]
```
00xc, line 31: The `default` feature has a special meaning, because it is always autoselected by cargo unless `--no-default-features` is passed. This is not super relevant because we build through a Makefile and pass features explicitly., but I'd rather keep it. Maybe we should have `default` and `default-debug`.

src/cpu/line_buffer.rs
vsntk18: Done!

Cargo.toml
vsntk18: ok,  now there are two features `default` and `default-debug`

src/log_buffer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+        if place_left <= len {
+            self.full = true;
+        }
+
+        /* update the tail offset if this write results
+         *  in a full buffer.
+         */
+        if self.full {
+            self.tail = self.head;
+        }
```
joergroedel, line 46: This empties the buffer when full, right? Please only advance tail so that the last line gets overwritten.

src/log_buffer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+            ret = self.buf[st.tail..st.head].to_vec();
+        } else {
+            ret = self.buf[st.tail..].to_vec();
+            ret.extend_from_slice(&self.buf[..st.head]);
+        }
+
+        self.state.update_for_read();
+
+        ret
+    }
```
joergroedel, line 112: A Vec<u8> is not a good target for reading. Reading the log will most likely write the log data into a guest-provided buffer. We will decide later how to type that, so this function can be removed for now.

joergroedel (changes requested):
Can you please also add unit-tests to test the basic functionality and the edge scenarios like buffer overflow and wraparound of the buffer?

src/log_buffer/mod.rs
00xc: Maybe taking a `&mut Vec` or a `&mut [u8]` would be the cleanest way to do it, but having these types in guest memory breaks Rust's ownership model (since the guest may write to that memory). Perhaps taking a pointer + length and doing a volatile write would be safer. Anyhow, I agree that removing this function for now is better.

vsntk18: Yes, of course. I was just using this function just to check if log buffer was getting written correctly.

src/log_buffer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub fn write_log(&mut self, s: &FixedString<LINE_BUFFER_SIZE>) {
+        let len = s.length();
+        let mut head = self.state.update_for_write(len);
+
+        let mut tmp: [u8; 4] = [0; 4];
+        for ch in s.iter() {
+            ch.encode_utf8(&mut tmp);
+            self.buf[head] = tmp[0];
+            head = (head + 1) % BUF_SIZE;
+            for item in tmp.iter().skip(1) {
```
roy-hopkins, line 85: If a multibyte character is followed by a single byte character then `tmp[1..3]` may contain non-zero values when processing the single byte character. You can clear `tmp` between each character, or it might make sense to use the return value from `encode_utf8` instead:

```
        let mut tmp: [u8; 4] = [0; 4];
        for ch in s.iter() {
            let utf = ch.encode_utf8(&mut tmp);
            for item in utf.as_bytes().iter() {
                self.buf[head] = *item;
                head = (head + 1) % BUF_SIZE;
            }
        }
```

vsntk18: Nice catch!  I thought the encode_utf8() would fill in the zeroes for the remaining bytes. But it doesn't.  Thanks for pointing out.

src/log_buffer/mod.rs
vsntk18: This actually just updates the tail pointer when the log_write() makes the buffer full.

vsntk18:
> Can you please also add unit-tests to test the basic functionality and the edge scenarios like buffer overflow and wraparound of the buffer?

Added some unit tests.

vsntk18:
Needs rebase due to percpu related changes in 7e5a9a72cac97361fe0890084b16777e3ce685cb

kernel/src/log_buffer/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+            self.buf = (*lb).lock().buf;
+        }
+    }
+
+    pub fn write_log(&mut self, s: &FixedString<LINE_BUFFER_SIZE>) {
+        self.buf.write(&s.to_string());
+    }
+
+    // A method used for testing
+    pub fn read_log(&mut self) -> Vec<u8> {
```
00xc, line 52: If it's only used for testing gate it behind `#[cfg(test)]` to avoid unused warnings.

kernel/src/log_buffer/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    /// # Safety
+    /// This fuction is called during the initiliazation phase
+    /// by a single thread on boot core. Hence dereferencing
+    /// the raw pointer passed as an argument is here is safe.
+    pub unsafe fn migrate(&mut self, lb: *const SpinLock<LogBuffer>) {
+        unsafe {
+            self.buf = (*lb).lock().buf;
+        }
+    }
```
00xc, line 45: Why not pass a regular reference to the spinlock?

kernel/src/log_buffer/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+/// This fuction is called during the initiliazation phase
+/// by a single thread on boot core. Hence dereferencing
+/// the raw pointer passed as an argument is here is safe.
+pub unsafe fn migrate_log_buffer(log_buf: *const SpinLock<LogBuffer>) {
+    unsafe {
+        LB.lock().migrate(log_buf);
+    }
+}
+
+static mut LB: SpinLock<LogBuffer> = SpinLock::new(LogBuffer::new());
```
00xc, line 71: Why does it need to be `mut`?

kernel/src/log_buffer/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+#[cfg(test)]
+const BUF_SIZE: usize = 64;
+
+#[cfg(test)]
+use crate::types::LINE_BUFFER_SIZE;
+
+#[test]
+fn test_read_write_normal() {
```
00xc, line 93: All of these should be in a test module, e.g.

```rust
#[cfg(test)]
mod test {
    use crate::types::LINE_BUFFER_SIZE;

    const BUF_SIZE: usize = 64;

    #[test]
    fn test_read_write_normal() {
        // ...
    }
}
```

kernel/src/utils/string_ring_buffer.rs
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2024 SUSE LLC
+//
+// Author: Roy Hopkins <roy.hopkins@suse.com>
```
00xc, line 5: Probably forgot to update author name :)

kernel/src/utils/string_ring_buffer.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl<const T: usize> Default for StringRingBuffer<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[test]
+fn test_ring_one_string() {
```
00xc, line 84: This test and the ones below should also be in a test module.

kernel/src/cpu/line_buffer.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+impl log::Log for BufferLogger {
+    fn enabled(&self, _metadata: &log::Metadata<'_>) -> bool {
+        true
+    }
+
+    fn log(&self, record: &log::Record<'_>) {
+        let comp: &'static str = self.component;
+        let line_buf: &mut LineBuffer = &mut this_cpu().get_line_buffer();
```
00xc, line 72: There's no need to add the type here, `RefMut` implements `Deref` and `DerefMut`, so access to the `LineBuffer` methods is transparent.

kernel/src/cpu/line_buffer.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    fn flush(&self) {}
+}
+
+static BUFFER_LOGGER: ImmutAfterInitCell<BufferLogger> = ImmutAfterInitCell::uninit();
+
+pub fn install_buffer_logger(component: &'static str) {
+    BUFFER_LOGGER
+        .init(&BufferLogger::new(component))
+        .expect("already initialized the logger");
```
00xc, line 118: I'd say return the error here and panic at the caller, just like we had before.

kernel/src/log_buffer/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// by a single thread on boot core. Hence dereferencing
+    /// the raw pointer passed as an argument is here is safe.
+    pub unsafe fn migrate(&mut self, lb: *const SpinLock<LogBuffer>) {
+        unsafe {
+            self.buf = (*lb).lock().buf;
+        }
+    }
+
+    pub fn write_log(&mut self, s: &FixedString<LINE_BUFFER_SIZE>) {
+        self.buf.write(&s.to_string());
```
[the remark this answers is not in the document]
vsntk18: Changed write() interface to take char iterator as a parameter.

kernel/src/log_buffer/mod.rs
vsntk18: done!

kernel/src/log_buffer/mod.rs
vsntk18: Yes, some compilation problems made me use that.  But I have changed to use a regular reference now.

kernel/src/log_buffer/mod.rs
vsntk18: Not necessary.  Removed it.  Thanks!

kernel/src/log_buffer/mod.rs
vsntk18: done!

kernel/src/utils/string_ring_buffer.rs
vsntk18: No, actually Roy created this file in one of the branches. I cherry picked it to my branch. So I will let it stay. :)

kernel/src/utils/string_ring_buffer.rs
vsntk18: done!

kernel/src/cpu/line_buffer.rs
vsntk18: Removed!

kernel/src/cpu/line_buffer.rs
vsntk18: done!

osteffenrh:
What is the plan for this?

vsntk18:
> What is the plan for this?

This was stalled due to lesser priority.  The changes on this PR is out-of-date, and a rebase would be necessary for the next round of review.  If you need this feature soon, I could spend some time on it next week or so.

vsntk18:
Reworked log buffer implementation.

1. There is a per-task Line buffer
2. A Global log buffer backed by a RamFile
3. Added a couple of basic test-in-svsm tests for the kernel tasks

kernel/src/fs/api.rs
```diff
@@ -302,3 +304,92 @@ impl DirectoryEntry {
...
+        }
+
+        let index = self.fill;
+        self.buffer[index] = b;
+        self.fill += 1;
+        if b == newline {
+            self.flush(is_log)?;
+        }
+
+        Ok(0)
```
joergroedel: There are a bunch of methods which only return `OK(0)` on success, which is not very useful. Either make them return the number of bytes added to the log or just `Ok(())`.

kernel/src/fs/log_buffer.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    fn truncate(&self, _size: usize) -> Result<usize, SvsmError> {
+        Err(SvsmError::FileSystem(FsError::not_supported()))
+    }
+
+    fn size(&self) -> usize {
+        0
+    }
+}
```
joergroedel: Why does `LogBuffer` need a `File` implementation? The backing memory is the `Log/logfile` file in the RamFS, no?

kernel/src/fs/log_buffer.rs
```diff
@@ -0,0 +1,131 @@
...
+#[cfg(all(test, test_in_svsm))]
+pub fn log_reset() -> Result<usize, SvsmError> {
+    let handle = open_write("Log/logfile").map_err(|_| SvsmError::LogError)?;
+    handle.truncate(0).map_err(|_| SvsmError::LogError)
+}
+
+#[derive(Debug)]
+struct LogBuffer {
+    lb: SpinLock<StdoutBuffer>,
+}
```
joergroedel: I think this struct should be a `RWLock` around a `RamFile` instance.

joergroedel (changes requested):
Hi Vasant, thanks for moving this forward. The changes go into the right direction but I have some comments and questions on the design.

kernel/src/fs/log_buffer.rs
vsntk18: This is the per-task Line Buffer.  RamFS backed file is the global log buffer. 

I am mirroring the ConsoleFile implementation. This LogBuffer is just a dummy interface to actual RamFS backed logfile. But it prohibits certain File operations like read, truncate etc which are not necessary for the Log file. At the same time, those operations are still needed and possible for the actual RamFS backed file.

vsntk18: Should I change the name to LineBuffer to avoid confusion?  I used the name LogBuffer for the sake of uniformity.

kernel/src/fs/log_buffer.rs
vsntk18: hmm, Access to RamFile is already synchronized, since it uses a RWLock for underlying RawRamFile, right?  Or, am I misunderstanding this?

kernel/src/fs/api.rs
vsntk18: will do that.

vsntk18:
Addressed some suggestions made by Joerg (few might still be open for discussion) and rebased.

kernel/src/fs/console.rs
```diff
@@ -6,62 +6,22 @@
...
-use crate::syscall::Obj;
 use alloc::string::String;
-use alloc::sync::Arc;
-
-// With the value of 224 the ConsoleBuffer struct will be exactly 256 bytes
-// large, avoiding memory waste due to internal fragmentation.
-const CONSOLE_LINE_BUFFER_SIZE: usize = 224;
 
 #[derive(Debug)]
 struct ConsoleBuffer {
```
n-ramacciotti: Could this be removed and `StdoutBuffer` used directly in `ConsoleFile`?
```
pub struct ConsoleFile {
    buffer: SpinLock<StdoutBuffer>,
}
```

kernel/src/fs/log_buffer.rs
```diff
@@ -0,0 +1,87 @@
...
+    handle.write(buf).map_err(|_| SvsmError::LogError)
+}
+
+pub fn log_read(buf: &mut [u8]) -> Result<usize, SvsmError> {
+    let handle = open_read("Log/logfile").map_err(|_| SvsmError::LogError)?;
+    handle.read(buf).map_err(|_| SvsmError::LogError)
+}
+
+#[cfg(all(test, test_in_svsm))]
+pub fn log_reset() -> Result<usize, SvsmError> {
```
n-ramacciotti: Since this is used for test in svsm, would it make sense to introduce it in the commit with the new tests ([bcef6bb](https://github.com/coconut-svsm/svsm/pull/52/changes/bcef6bb674a85185fb04d7c03a8d823b9a63b3d3)) and inside the `tests` module directly?

kernel/src/fs/log_buffer.rs
```diff
@@ -85,3 +85,47 @@ pub fn stdout_open(taskname: String) -> (Arc<dyn Obj>, Arc<dyn Obj>) {
...
+            "[SVSM test task] hello world\n"
+        );
+    }
+
+    #[test]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_log_buffer_multiple_tasks() {
+        let _ = log_reset();
+        log::info!("in test task");
+        start_kernel_task(KernelThreadStartInfo::new(task1, 1), String::from("task1"))
```
n-ramacciotti: Would it make sense to call `wait_for_termination` before checking the expected buffer?

kernel/src/fs/console.rs
vsntk18: Good idea. Let me remove this unnecessary wrapper :)

kernel/src/fs/log_buffer.rs
vsntk18: Yeah, I thought there might be legitimate use for resetting the log.  But at the moment only in the test. So I will move it inside tests module. Thanks for the suggestion!

kernel/src/fs/log_buffer.rs
vsntk18: Since there are only 3 tasks and no expectation of interrupts within the test, I didn't use wait_for_termination. I don't have a strong opinion on this. If you think there are good reasons for it, I can add it.

vsntk18:
Addressed Nicola's review comments.

kernel/src/fs/log_buffer.rs
n-ramacciotti: I'm not sure that, when `start_kernel_task` returns, `task1` has already finished. It probably is, in this case `wait_for_termination` would return immediately. Let's hear from others!

kernel/src/console.rs
n-ramacciotti: I don't know if it's necessary to keep them in the codebase, but these can be removed without consequence, as they are no longer used:
- `CONSOLE_LOGGER`
- `install_console_logger`
- `ConsoleLogger`

vsntk18: When I resumed with this PR, stage2 was still there and using console logger.  But now it's gone I think.

n-ramacciotti: > When I resumed with this PR, stage2 was still there and using console logger. But now it's gone I think.

yes it's been removed. (#1053)

syscall/Cargo.toml
```diff
@@ -11,3 +11,6 @@ zerocopy = { workspace = true, features = ["derive"] }
 
 [lints]
 workspace = true
+
+[features]
+enable-console-log = []
```
n-ramacciotti: IIUC, this feature should be also present in `userlib` and all the other userpace modules (`userinit` for the moment). And should be enabled in chain (`usermodule/enable` -> `userlib/enable` -> `syscall/enable`).

However, when enabled (even on its own by default), every call to `write_log` would cause a double write, for example:
```
[init] COCONUT-SVSM init process starting
[init] COCONUT-SVSM init process starting
```

This is due to the write to the `LOG_HANDLE`, which also prints to the console (with `kernel/enable-console-log`) and to the write to `CONSOLE_HANDLE`

That said, I'm not sure this feature should be available in `userspace`. It should probably be enough to use the kernel side feature.

kernel/src/fs/log_buffer.rs
vsntk18: With the current scheduling mechanism it should. But probably it's good practice to wait_for_termination() in case something changes.  It doesn't cause any wait at the moment anyway.  so added it.

kernel/src/console.rs
vsntk18: Removed these unused symbols.

syscall/Cargo.toml
vsntk18: Yeah, I haven't put much thought into userspace yet.  There aren't even tests yet, which I plan to add after https://github.com/coconut-svsm/svsm/pull/926 is merged. Let's revisit it as userspace evolves?

vsntk18:
Addressed comments from Nicola

syscall/Cargo.toml
n-ramacciotti: > which I plan to add after https://github.com/coconut-svsm/svsm/pull/926 is merged. Let's revisit it as userspace evolves?

I'm not sure when #926 will be merged, but I agree. For the moment I think it is sufficient to drop the feature from `syscall` and `CONSOLE_HANDLE` from `syscall/log_buffer`, and address it later (maybe with a separate PR)

vsntk18: yeah, dropped it for now.

kernel/src/fs/api.rs
```diff
@@ -302,3 +304,92 @@ impl DirectoryEntry {
         DirectoryEntry { name, entry }
     }
 }
+
+// With the value of 224 the Buffer struct will be exactly 256 bytes
+// large, avoiding memory waste due to internal fragmentation.
+const LINE_BUFFER_SIZE: usize = 224;
+
+#[derive(Debug)]
+pub struct StdoutBuffer {
```
n-ramacciotti: What about changing this to `LineBuffer`? and moving it out of `api.rs` to `kernel/src/fs/buffer.rs`?

kernel/src/fs/log_buffer.rs
```diff
@@ -0,0 +1,132 @@
...
+    handle.write(buf).map_err(|_| SvsmError::LogError)
+}
+
+pub fn log_read(buf: &mut [u8]) -> Result<usize, SvsmError> {
+    let handle = open_read("Log/logfile").map_err(|_| SvsmError::LogError)?;
+    handle.read(buf).map_err(|_| SvsmError::LogError)
+}
+
+#[derive(Debug)]
+struct LineBuffer {
```
n-ramacciotti: What about changing this to `LogFile` (or something similar), mirroring `ConsoleFile`?

kernel/src/log_buffer.rs
```diff
@@ -0,0 +1,119 @@
...
+            .as_fs()
+            .ok_or(SvsmError::LogError)
+            .map_err(print_warning)
+            .unwrap();
+        let _ = fs.write_buffer(&buf).map_err(print_warning).unwrap();
+        Ok(())
+    }
+}
+
+static LOGGER: ImmutAfterInitCell<SpinLock<LogBuffer>> = ImmutAfterInitCell::uninit();
```
n-ramacciotti, line 44: What about changing this to `BUFFER`?

kernel/src/fs/api.rs
vsntk18: I hadn't paid attention to Buffer trait. I made LineBuffer implement Buffer trait. I consolidated Console file and Log file into a single struct that way. kernel/src/fs/console.rs can be removed. How do you like this idea?

kernel/src/fs/log_buffer.rs
vsntk18: This won't be necessary after the changes I indicated in the above comment

kernel/src/log_buffer.rs
vsntk18: BUFFER would be too generic a name in my opinion

n-ramacciotti: Yeah right

kernel/src/fs/log_buffer.rs
n-ramacciotti: But you did change `LineBuffer` to `LogFile` IIUC

kernel/src/fs/api.rs
n-ramacciotti: It makes much sense!

n-ramacciotti: I'm losing track of things in the commit history. What do you think about squashing some fixes? @vsntk18

vsntk18: @n-ramacciotti  I could do that. But wouldn't it easy for you to just review the delta from the last state? For example:  commit fd444abab42d4 is the implementation of the suggestions in your last round of review.

kernel/src/fs/log_buffer.rs
vsntk18: Yeah, LineBuffer is shared by both console and logbuffer now.

kernel/src/fs/api.rs
stefano-garzarella: +1 on squashing, please. I usually review patch by patch and having things fixed later will increase a lot the time to review.

n-ramacciotti: I like the new changes in https://github.com/coconut-svsm/svsm/commit/fd444abab42d4a78ef4aaa5777e723df165893c5.

> I usually review patch by patch and having things fixed later will increase a lot the time to review.

I also prefer to review one commit at a time. Without squashing, the next round of reviews would involve a lot of back-and-forth between different commits.

kernel/src/fs/filesystem.rs
```diff
@@ -966,7 +966,7 @@ mod tests {
 
         // Check if it is removed from the listing
         let root_list = list_dir("").unwrap();
-        assert_eq!(root_list, [FileName::from("test2")]);
+        assert!(root_list.contains(&FileName::from("test2")));
```
n-ramacciotti: These tests have been changed as the root list now contains the log file. I had a similar issue in `test-in-svsm userspace PR` (where I have userspace binaries in the fs, see commit https://github.com/coconut-svsm/svsm/pull/926/changes/eef70b35df596cab1396eb50d065321f3c606560).

It makes sense to add here also:
```
assert!(!root_list.contains(&FileName::from("test1")));
```

Nit: What about moving these assert changes to a separate commit? It's just a personal preference, not a strong opinion.

vsntk18: Addressed in c62bda8f7d212821e2983ee03ebb0a4f91692806.

kernel/src/fs/console.rs
```diff
@@ -126,9 +126,12 @@ impl File for ConsoleFile {
     }
 }
 
-pub fn stdout_open() -> Arc<dyn Obj> {
+pub fn stdout_open(_name: String) -> (Arc<dyn Obj>, Arc<dyn Obj>) {
```
n-ramacciotti: What about moving these changes in a subsequent commit after `LogFile` introduction? I don't see an advantage of having two `ConsoleFile` here, considering that it also  will be dropped after

Edit: this comment is marked as outdated probably because `console.rs` has been removed, but I think it is still relevant.

vsntk18: Yeah, I suppose I can reorder those comments.

vsntk18:
Reordered few commits to address Nicola's comments. No functional changes.

n-ramacciotti:
```
log_buffer: implement Buffer trait for line buffer
This way we can consolidate console buffer and log buffer into a
single line buffer.
```
Couldn't this also be done without implementing the `Buffer` trait?

vsntk18:
> ```
> log_buffer: implement Buffer trait for line buffer
> This way we can consolidate console buffer and log buffer into a
> single line buffer.
> ```
> 
> Couldn't this also be done without implementing the `Buffer` trait?

Of course, it could be done that way. Semantically LineBuffer is a 'Buffer', so I just moved it to buffer.rs.  And, there is no real advantage. But I don't see any disadvantages either

n-ramacciotti:
> Of course, it could be done that way. Semantically LineBuffer is a 'Buffer', so I just moved it to buffer.rs. And, there is no real advantage. But I don't see any disadvantages either

Yeah, I don't see any disadvantages neither. 

My thoughts were on the commit message: from the title I understand that it's  thanks to `Buffer` trait that it is possible to obtain a single struct for both. When in reality it's just a plus

vsntk18:
> > Of course, it could be done that way. Semantically LineBuffer is a 'Buffer', so I just moved it to buffer.rs. And, there is no real advantage. But I don't see any disadvantages either
> 
> Yeah, I don't see any disadvantages neither.
> 
> My thoughts were on the commit message: from the title I understand that it's thanks to `Buffer` trait that it is possible to obtain a single struct for both. When in reality it's just a plus

ok. yeah, if you have a suggestion for a more clear commit message, please let me know.

n-ramacciotti:
> ok. yeah, if you have a suggestion for a more clear commit message, please let me know.

What about something like:
```
kernel/log_buffer: Introduce line buffer

Console and File handling require both line buffering and only differ on the destination.

Modify `ConsoleBuffer` to accept an additional parameter (`is_console`) to differentiate
and redirect output to the correct handling.

Rename `ConsoleBuffer` to `LineBuffer` to better explain its behaviour.

Additionally, given that it is a buffer, implement `Buffer` trait on top of it.
```

Note that I didn't mention `StdoutBuffer` as it is not used in the final changes and can be dropped from the history.

what do you think?

vsntk18:
> > ok. yeah, if you have a suggestion for a more clear commit message, please let me know.
> 
> What about something like:
> 
> ```
> kernel/log_buffer: Introduce line buffer
> 
> Console and File handling require both line buffering and only differ on the destination.
> 
> Modify `ConsoleBuffer` to accept an additional parameter (`is_console`) to differentiate
> and redirect output to the correct handling.
> 
> Rename `ConsoleBuffer` to `LineBuffer` to better explain its behaviour.
> 
> Additionally, given that it is a buffer, implement `Buffer` trait on top of it.
> ```
> 
> Note that I didn't mention `StdoutBuffer` as it is not used in the final changes and can be dropped from the history.
> 
> what do you think?

Yeah, that looks good to me. Thanks for the suggestion!

n-ramacciotti:
My idea was to drop entirely `StdoutBuffer` from the history. Like removing it from https://github.com/coconut-svsm/svsm/pull/52/changes/f4591554cfac61b7a86c5a5b4ff826ecfa55bd3e and directly introducing there `LineBuffer`.

vsntk18:
> My idea was to drop entirely `StdoutBuffer` from the history. Like removing it from [f459155](https://github.com/coconut-svsm/svsm/commit/f4591554cfac61b7a86c5a5b4ff826ecfa55bd3e) and directly introducing there `LineBuffer`.

yes, of course. If the changes are functionally ok, I could try to reduce the number of commits.

kernel/src/task/schedule.rs
```diff
@@ -222,7 +222,9 @@ impl RunQueue {
     ///
     /// Panics if there is no current task.
     pub fn current_task(&self) -> TaskPointer {
-        self.current_task.as_ref().unwrap().clone()
+        self.current_task
+            .as_ref()
+            .map_or(self.idle_task.as_ref().unwrap().clone(), |t| t.clone())
```
n-ramacciotti, line 240: Without this change, it doesn't print anything. It seems important, could you please mention why this is needed in the commit message?

kernel/src/task/tasks.rs
```diff
@@ -489,7 +488,12 @@ impl Task {
             runlist_link: LinkedListAtomicLink::default(),
             objs: objtree,
             wait_queue: SpinLockIrqSafe::new(WaitQueue::new()),
-        }))
+        });
+
+        if args.thread_of.is_none() {
```
n-ramacciotti, line 569: It is also worth mentioning in the commit that this is now done for both kernel and user tasks, while before it was only done for user tasks

kernel/src/svsm.rs
```diff
@@ -410,6 +412,10 @@ unsafe fn svsm_start(
         .env_setup_late(debug_serial_port)
         .expect("Late environment setup failed");
 
+    // it's important to make sure that idle task
+    // has been already created and console is
+    // initialized in env_setup_late() before this
+    install_buffer_logger();
```
n-ramacciotti, line 418: I would mention it and also explain why in the commit description

n-ramacciotti:
@vsntk18 I left some minor comments. The rest LGTM. Thanks!

kernel/src/task/schedule.rs
vsntk18: yeah, will add

kernel/src/svsm.rs
vsntk18: hmm, I am not sure about this tbh. I think it's an extra detail and would be a distraction in the commit message. But I left it as a comment so that if someone wants to move the code around, they keep this invariant in mind.

kernel/src/task/tasks.rs
vsntk18: I mentioned in the commit message now that it's done for both user and kernel tasks.

vsntk18:
dropped StdoutBuffer from history, and adjusted some commit messages.

kernel/src/svsm.rs
n-ramacciotti: Yeah, right. It is easier to spot in the code. All the new fixes and the PR in general LGTM!

vsntk18: Thanks for the review!

n-ramacciotti:
> Implemented a global buffer to store log messages and passed this buffer from stage2 to svsm kernel. Replaced console logging with buffer logging.
> 
> Percpu line buffer will probably be necessary, and is not yet implemented. As a rust newbee, I appreciate your feedback on if I am using the right data structures in the right way. Thanks!

nit: since the code has changed significantly from this version. Could this description be updated? Using also the "Fixes: issue number" tag to highlight better the relationship with the issue section (in this way it will be closed automatically once this is merged)

vsntk18:
> > Implemented a global buffer to store log messages and passed this buffer from stage2 to svsm kernel. Replaced console logging with buffer logging.
> > Percpu line buffer will probably be necessary, and is not yet implemented. As a rust newbee, I appreciate your feedback on if I am using the right data structures in the right way. Thanks!
> 
> nit: since the code has changed significantly from this version. Could this description be updated? Using also the "Fixes: issue number" tag to highlight better the relationship with the issue section (in this way it will be closed automatically once this is merged)

That description suited the first version of the PR was created. Since then there were many changes in the design etc. The most recent description is: https://github.com/coconut-svsm/svsm/pull/52#issuecomment-4260684856

Linking the Fixes: #issue is a good idea. Thanks for the suggestion.  Let me do that :)

n-ramacciotti:
> Fixes: https://github.com/coconut-svsm/svsm/issues/12
>
>Signed-off-by: Vasant Karasulli <vasant.karasulli@suse.com>

There shouldn't be a space here to be considered a trailer (by git) and (maybe, but not tested) recognized by github.

In this PR, I see:

```
Successfully merging this pull request may close these issues.

None yet
```

IMHO: commit https://github.com/coconut-svsm/svsm/pull/52/changes/bce8181214279ee66650dda350ac3b168c180140 is more suited to have this trailer

Probably it's needed in the PR description.

vsntk18:
> > Fixes: #12
> > Signed-off-by: Vasant Karasulli [vasant.karasulli@suse.com](mailto:vasant.karasulli@suse.com)
> 
> There shouldn't be a space here to be considered a trailer (by git) and (maybe, but not tested) recognized by github.
> 
> In this PR, I see:
> 
> ```
> Successfully merging this pull request may close these issues.
> 
> None yet
> ```

Are you sure?  I see Fixes tag with a space like that in many other commits, 92c082eb7cc7e3b4f68805b20e78cd2f0ac9aae1 for example
> 
> IMHO: commit [bce8181](https://github.com/coconut-svsm/svsm/commit/bce8181214279ee66650dda350ac3b168c180140) is more suited to have this trailer
> 
> Probably it's needed in the PR description.

Well, that's a toss up between 2 commits, whether the implementation of the buffer or redirecting to the buffer. But I can change it to your suggestion :)

n-ramacciotti:
> Are you sure? I see Fixes tag with a space like that in many other commits, [92c082e](https://github.com/coconut-svsm/svsm/commit/92c082eb7cc7e3b4f68805b20e78cd2f0ac9aae1) for example

I did a test and it  seems that it doesn't work neither even without the space. I'm pretty sure that git trailers should be a single block. (see [1]). I don't know how GitHub works though. 

> > IMHO: commit [bce8181](https://github.com/coconut-svsm/svsm/commit/bce8181214279ee66650dda350ac3b168c180140) is more suited to have this trailer
> > Probably it's needed in the PR description.
> 
> Well, that's a toss up between 2 commits, whether the implementation of the buffer or redirecting to the buffer. But I can change it to your suggestion :)

It's good also in the current commit. I don't have a strong preference.

[1] https://lore.kernel.org/git/xmqqr1k3w1g7.fsf@gitster.g/

vsntk18:
Addressed Jörg's review comments:
          Enabled console output by default for both DEBUG and RELEASE builds. Console output can still be disabled by defining environment variable DISABLE_CONSOLE_LOG.

vsntk18:
Rebased the PR.

Enabled console output by default for both DEBUG and RELEASE builds. Console output can still be disabled by defining environment variable DISABLE_CONSOLE_LOG.

kernel/src/fs/log_buffer.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    fn size(&self) -> usize {
+        0
+    }
+}
+
+pub fn stdout_open(taskname: String) -> (Arc<dyn Obj>, Arc<dyn Obj>) {
+    let console_file: Arc<dyn File> = Arc::new(LogFile::new(taskname.clone(), true));
+    let log_file: Arc<dyn File> = Arc::new(LogFile::new(taskname.clone(), false));
```
n-ramacciotti, line 74: Do we really need to have both `console_file` and `log_file`? IIUC, kernel tasks during logging uses only the `log_file` (obj handle 1), which will also print to the console if enabled. If a kernel task want to print to the console it can directly use `console_write`. User tasks are therefore the only user of obj handle 0 and I don't think that a user task should be able to print to the console if the kernel disabled console log.

I believe this can also be addressed in a separate and subsequent PR if we want to merge this PR soon.

vsntk18: > I don't think that a user task should be able to print to the console if the kernel disabled console log.

I am not sure about that.  Kernel might disable console log and user task still can call console_print() function to print messages to console, which in turn fetches the task specific console file.

n-ramacciotti: > Kernel might disable console log and user task still can call console_print() 

yes, exactly. I'm not sure this is a good behavior. My idea was to leave only one handler (on of `CONSOLE_HANDLE` or `LOG_HANDLE`), removing the ability for userspace to print directly to the console, but only through the console file handler, which leave the final decision to kernel (`enable_console_log`). Essentially, IMO, userspace should not be able to print directly neither it should know where the message is logged (console or log file).

vsntk18: It makes sense, yes. But I don't think user space consideration is complete with this PR, as we discussed before. I still need to add tests. I will revisit this.

n-ramacciotti: Yes, this can be addressed also later!

joergroedel (commented):
Hi Vasant,

Thanks for driving this forward. I found why I do not see anything when I build with this PR. Please also add the `enable-console-log` feature in the build recipes under `config/`.

Also there seem to be non-trivial conflicts with the user-mode testing PR I just merged, can you please have a look? I wasn't able to quickly resolve it and get `make test-in-svsm` working again.

Thanks!

vsntk18:
> Hi Vasant,
> 
> Thanks for driving this forward. I found why I do not see anything when I build with this PR. Please also add the `enable-console-log` feature in the build recipes under `config/`.
> 
> Also there seem to be non-trivial conflicts with the user-mode testing PR I just merged, can you please have a look? I wasn't able to quickly resolve it and get `make test-in-svsm` working again.
> 
> Thanks!

Ok, will take a look. Thanks for the review.

vsntk18:
Rebased the PR.

Addressed Joerg's comment:
added the enable-console-log feature in the build recipes under config/

user/lib/src/testing.rs
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 //
 
-use crate::println;
+use crate::{log_msg, print, println};
```
n-ramacciotti, line 4: This can be avoided by using "$crate" when introducing the new println and print. See https://github.com/coconut-svsm/svsm/commit/80c99e9a8e3b14992abe73b355a067ef792ba1fc

joergroedel: Let's change that as a follow-on, thanks.

joergroedel (approved):

user/lib/src/testing.rs
vsntk18: Yeah, I will send a separate small fix.

Issue #14, opened 2023-03-28
Implement VC handling in stage2 and COCONUT kernel


joergroedel:
Move away from directly making GHCB calls and implement a proper #VC exception handler to handle NAE events.

This is required for CPL-3 support, as some instructions that cause NAEs are allowed in unprivileged code (e.g. CPUID). But the console printing support also benefits from this, as its driver infrastructure can be removed so that it uses IN/OUT instructions directly.

Further, by handling #VC exceptions in stage2 already we can get console support very early in the COCONUT kernel, which makes debugging of the early boot stages much easier.

There is no reason yet to handle MMIO related NAE events, but IOIO is needed, best including INS/OUTS instructions (but these are not needed in the first step).

00xc:
Can we close this, since #55 is merged?

p4zuu:
> Can we close this, since #55 is merged?

I guess so. There are still improvement to do, like implementing all NAE events handlers, and making #VC handler IST, on which I plan to work soon, but I let @joergroedel decide on the closing

stefano-garzarella:
@joergroedel I'm closing this, feel free to re-open it if you think more work is needed.

PR #55, opened 2023-06-20
Add #VC handler

p4zuu:
This is a draft for the implementation of #14, which is implementing a #VC handler to handle NAE events.

NAE events have to be handled early, so we also set and load the IDT in stage 2.

At this point, interrupt handling in stage 2 doesn't work. I mostly moved and debug code to try to call the VC handler from a stage 2 `cpuid`, which fails. The IDT setup in this PR still handles interrupts in the SVSM kernel, but not in stage 2. My guess goes that IDT entries should be different because stage2 is a different binary (with different GDT entries?), but I'm still debugging. That's where my lack of system programming background hits.

### DIFF
--- a/src/cpu/idt/common.rs
+++ b/src/cpu/idt/common.rs
@@ -5,7 +5,6 @@
 // Author: Joerg Roedel <jroedel@suse.de>
 
 use crate::address::{Address, VirtAddr};
-use crate::cpu::insn::Instruction;
 use crate::cpu::registers::{X86GeneralRegs, X86InterruptFrame};
 use crate::types::SVSM_CS;
 use core::arch::{asm, global_asm};
@@ -44,7 +43,6 @@ pub struct X86ExceptionContext {
     pub vector: usize,
     pub error_code: usize,
     pub frame: X86InterruptFrame,
-    pub insn: Instruction,
 }
 
 #[derive(Copy, Clone, Default, Debug)]
--- a/src/cpu/idt/common.rs
+++ b/src/cpu/idt/common.rs
@@ -5,6 +5,7 @@
 // Author: Joerg Roedel <jroedel@suse.de>
 
 use crate::address::{Address, VirtAddr};
+use crate::cpu::insn::Instruction;
 use crate::cpu::registers::{X86GeneralRegs, X86InterruptFrame};
 use crate::types::SVSM_CS;
 use core::arch::{asm, global_asm};
@@ -43,6 +44,7 @@ pub struct X86ExceptionContext {
     pub vector: usize,
     pub error_code: usize,
     pub frame: X86InterruptFrame,
+    pub insn: Instruction,
 }
 
 #[derive(Copy, Clone, Default, Debug)]
--- a/src/cpu/idt/common.rs
+++ b/src/cpu/idt/common.rs
@@ -4,14 +4,8 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
-use super::control_regs::read_cr2;
-use super::percpu::this_cpu;
-use super::tss::IST_DF;
-use super::vc::handle_vc_exception;
-use super::{X86GeneralRegs, X86InterruptFrame};
 use crate::address::{Address, VirtAddr};
-use crate::cpu::extable::handle_exception_table;
-use crate::debug::gdbstub::svsm_gdbstub::handle_debug_exception;
+use crate::cpu::registers::{X86GeneralRegs, X86InterruptFrame};
 use crate::types::SVSM_CS;
 use core::arch::{asm, global_asm};
 use core::mem;
@@ -53,7 +47,7 @@ pub struct X86ExceptionContext {
 
 #[derive(Copy, Clone, Default, Debug)]
 #[repr(C, packed)]
-struct IdtEntry {
+pub struct IdtEntry {
     low: u64,
     high: u64,
 }
@@ -110,13 +104,9 @@ struct IdtDesc {
     address: VirtAddr,
 }
 
-extern "C" {
-    static idt_handler_array: u8;
-}
+pub type Idt = [IdtEntry; IDT_ENTRIES];
 
-type Idt = [IdtEntry; IDT_ENTRIES];
-
-static mut GLOBAL_IDT: Idt = [IdtEntry::no_handler(); IDT_ENTRIES];
+pub static mut GLOBAL_IDT: Idt = [IdtEntry::no_handler(); IDT_ENTRIES];
 
 pub fn idt_base_limit() -> (u64, u32) {
     unsafe {
@@ -126,20 +116,7 @@ pub fn idt_base_limit() -> (u64, u32) {
     }
 }
 
-fn init_idt(idt: &mut Idt) {
-    // Set IDT handlers
-    let handlers = unsafe { VirtAddr::from(&idt_handler_array as *const u8) };
-    for (i, entry) in idt.iter_mut().enumerate() {
-        *entry = IdtEntry::entry(handlers + (32 * i));
-    }
-}
-
-unsafe fn init_ist_vectors(idt: &mut Idt) {
-    let handler = VirtAddr::from(&idt_handler_array as *const u8) + (32 * DF_VECTOR);
-    idt[DF_VECTOR] = IdtEntry::ist_entry(handler, IST_DF.try_into().unwrap());
-}
-
-fn load_idt(idt: &Idt) {
+pub fn load_idt(idt: &Idt) {
     let desc: IdtDesc = IdtDesc {
         size: (IDT_ENTRIES * 16) as u16,
         address: VirtAddr::from(idt.as_ptr()),
@@ -162,74 +139,6 @@ pub fn triple_fault() {
     }
 }
 
-pub fn early_idt_init() {
-    unsafe {
-        init_idt(&mut GLOBAL_IDT);
-        load_idt(&GLOBAL_IDT);
-    }
-}
-
-pub fn idt_init() {
-    // Set IST vectors
-    unsafe {
-        init_ist_vectors(&mut GLOBAL_IDT);
-    }
-}
-
-#[no_mangle]
-fn generic_idt_handler(ctx: &mut X86ExceptionContext) {
-    if ctx.vector == DF_VECTOR {
-        let cr2 = read_cr2();
-        let rip = ctx.frame.rip;
-        let rsp = ctx.frame.rsp;
-        panic!(
-            "Double-Fault at RIP {:#018x} RSP: {:#018x} CR2: {:#018x}",
-            rip, rsp, cr2
-        );
-    } else if ctx.vector == GP_VECTOR {
-        let rip = ctx.frame.rip;
-        let err = ctx.error_code;
-
-        if !handle_exception_table(ctx) {
-            panic!(
-                "Unhandled General-Protection-Fault at RIP {:#018x} error code: {:#018x}",
-                rip, err
-            );
-        }
-    } else if ctx.vector == PF_VECTOR {
-        let cr2 = read_cr2();
-        let rip = ctx.frame.rip;
-        let err = ctx.error_code;
-
-        if this_cpu()
-            .handle_pf(VirtAddr::from(cr2), (err & PF_ERROR_WRITE) != 0)
-            .is_err()
-            && !handle_exception_table(ctx)
-        {
-            handle_debug_exception(ctx, ctx.vector);
-            panic!(
-                "Unhandled Page-Fault at RIP {:#018x} CR2: {:#018x} error code: {:#018x}",
-                rip, cr2, err
-            );
-        }
-    } else if ctx.vector == VC_VECTOR {
-        handle_vc_exception(ctx);
-    } else if ctx.vector == BP_VECTOR {
-        handle_debug_exception(ctx, ctx.vector);
-    } else {
-        let err = ctx.error_code;
-        let vec = ctx.vector;
-        let rip = ctx.frame.rip;
-
-        if !handle_exception_table(ctx) {
-            panic!(
-                "Unhandled exception {} RIP {:#018x} error code: {:#018x}",
-                vec, rip, err
-            );
-        }
-    }
-}
-
 #[cfg(feature = "enable-stacktrace")]
 extern "C" {
     static generic_idt_handler_return: u8;
@@ -241,30 +150,8 @@ pub fn is_exception_handler_return_site(rip: VirtAddr) -> bool {
     addr == rip
 }
 
-// Entry Code
 global_asm!(
     r#"
-        .text
-    push_regs:
-        pushq   %rax
-        pushq   %rbx
-        pushq   %rcx
-        pushq   %rdx
-        pushq   %rsi
-        pushq   %rdi
-        pushq   %rbp
-        pushq   %r8
-        pushq   %r9
-        pushq   %r10
-        pushq   %r11
-        pushq   %r12
-        pushq   %r13
-        pushq   %r14
-        pushq   %r15
-
-        movq    %rsp, %rdi
-        call    generic_idt_handler
-
         /* Needed by the stack unwinder to recognize exception frames. */
         .globl generic_idt_handler_return
     generic_idt_handler_return:
@@ -288,20 +175,6 @@ global_asm!(
         addq    $16, %rsp /* Skip vector and error code */
 
         iretq
-
-        .align 32
-        .globl idt_handler_array
-    idt_handler_array:
-        i = 0
-        .rept 32
-        .align 32
-        .if ((0x20027d00 >> i) & 1) == 0
-        pushq   $0
-        .endif
-        pushq   $i  /* Vector Number */
-        jmp push_regs
-        i = i + 1
-        .endr
         "#,
     options(att_syntax)
 );
--- a/src/cpu/idt/svsm.rs
+++ b/src/cpu/idt/svsm.rs
@@ -8,7 +8,7 @@ use super::super::control_regs::read_cr2;
 use super::super::extable::handle_exception_table;
 use super::super::percpu::this_cpu;
 use super::super::tss::IST_DF;
-use super::super::vc::handle_vc_exception;
+use super::super::vc::stage2_handle_vc_exception;
 use super::common::PF_ERROR_WRITE;
 use super::common::{
     load_idt, Idt, IdtEntry, BP_VECTOR, DF_VECTOR, GLOBAL_IDT, GP_VECTOR, PF_VECTOR, VC_VECTOR,
@@ -82,7 +82,7 @@ pub fn generic_idt_handler(ctx: &mut X86ExceptionContext) {
             );
         }
     } else if ctx.vector == VC_VECTOR {
-        handle_vc_exception(ctx);
+        stage2_handle_vc_exception(ctx);
     } else if ctx.vector == BP_VECTOR {
         handle_debug_exception(ctx, ctx.vector);
     } else {
--- a/src/cpu/idt/svsm.rs
+++ b/src/cpu/idt/svsm.rs
@@ -8,7 +8,7 @@ use super::super::control_regs::read_cr2;
 use super::super::extable::handle_exception_table;
 use super::super::percpu::this_cpu;
 use super::super::tss::IST_DF;
-use super::super::vc::stage2_handle_vc_exception;
+use super::super::vc::handle_vc_exception;
 use super::common::PF_ERROR_WRITE;
 use super::common::{
     load_idt, Idt, IdtEntry, BP_VECTOR, DF_VECTOR, GLOBAL_IDT, GP_VECTOR, PF_VECTOR, VC_VECTOR,
@@ -46,7 +46,7 @@ pub fn idt_init() {
 }
 
 #[no_mangle]
-pub fn generic_idt_handler(ctx: &mut X86ExceptionContext) {
+pub extern "C" fn generic_idt_handler(ctx: &mut X86ExceptionContext) {
     if ctx.vector == DF_VECTOR {
         let cr2 = read_cr2();
         let rip = ctx.frame.rip;
@@ -82,7 +82,7 @@ pub fn generic_idt_handler(ctx: &mut X86ExceptionContext) {
             );
         }
     } else if ctx.vector == VC_VECTOR {
-        stage2_handle_vc_exception(ctx);
+        handle_vc_exception(ctx);
     } else if ctx.vector == BP_VECTOR {
         handle_debug_exception(ctx, ctx.vector);
     } else {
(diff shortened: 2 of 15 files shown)


00xc:
> My guess goes that IDT entries should be different because stage2 is a different binary (with different GDT entries?)

I'd say you *at least* need to set up a GDT with a CPL0 code segment.

p4zuu:
> I'd say you _at least_ need to set up a GDT with a CPL0 code segment.

GDT is set and loaded in [src/boot_stage2.rs](https://github.com/coconut-svsm/svsm/blob/main/src/boot_stage2.rs#L26)
```asm
         * Load a GDT. Despite the naming, it contains valid
         * entries for both, "legacy" 32bit and long mode each.
         */
        movl $gdt64_desc, %eax
        lgdt (%eax)
```

So I guess no need to set and load it again, right? Just ensure that the IDT fields match the correct GDT entries

00xc:
> So I guess no need to set and load it again, right? Just ensure that the IDT fields match the correct GDT entries

The only GDT entry that is referenced from the IDT is the TSS, which contains the IST stacks for #DF handling, as far as I can tell. This is not set up until `bsp_percpu.load()` is called (which ends up calling `load_tss()`), followed by `idt_init()` (which calls `init_ist_vectors()`). So, by the time `early_idt_init()` is called, I don't think any entries should be different in the GDT. Have you tried calling `load_gdt()` in stage2?

p4zuu:
> The only GDT entry that is referenced from the IDT is the TSS, which contains the IST stacks for #DF handling, as far as I can tell. This is not set up until `bsp_percpu.load()` is called (which ends up calling `load_tss()`), followed by `idt_init()` (which calls `init_ist_vectors()`). So, by the time `early_idt_init()` is called, I don't think any entries should be different in the GDT.

That was my conclusion too, that's why I was struggling finding what could possibly go wrong.

>Have you tried calling `load_gdt()` in stage2?

I didn't, but I just tried and it worked :) VC handler gets called by `cpuid` in stage 2 and in the SVSM kernel as well (depending where you put the instruction of course). Thanks Carlos! I still have to clearly figure out happens

src/cpu/idt_generic_handler.rs
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Thomas Leroy <tleroy@suse.de>
```
joergroedel, line 5: When you move code to a new file, please also add the original authors to the file header.

joergroedel:
In general this looks good to me. The question is how we organize the code sharing of the IDT handlers between stage2 and SVSM kernel. The SVSM needs more complicated handlers for some things (e.g. #PF) than stage2, and we want separate handlers for both.

We can divide that by introducing features, or just include different sub-modules for SVSM and stage2. Other suggestions are also welcome.

joergroedel:
> I didn't, but I just tried and it worked :) VC handler gets called by `cpuid` in stage 2 and in the SVSM kernel as well (depending where you put the instruction of course). Thanks Carlos! I still have to clearly figure out happens

Exception handling needs a GDT to be set up, because raising the exception will reference the GDT to load the handlers code segment, and the IRET also references the GDT to jump back to the original code segment.

00xc:
> In general this looks good to me. The question is how we organize the code sharing of the IDT handlers between stage2 and SVSM kernel. The SVSM needs more complicated handlers for some things (e.g. #PF) than stage2, and we want separate handlers for both.
>
> We can divide that by introducing features, or just include different sub-modules for SVSM and stage2. Other suggestions are also welcome.

I think the feeature approach will only work if we separate things into different crates. As far as I know, there is no way to do conditional compilation depending on the binary being built (e.g. stage2 or SVSM kernel).

I'd say the best way is to separate both into different sub-modules and include them based on what's needed, as you suggest. Maybe something like `src/cpu/idt/stage2.rs` and `src/cpu/idt/svsm.rs`? And perhaps a third file for any common code and structures needed.

p4zuu:
> I'd say the best way is to separate both into different sub-modules and include them based on what's needed, as you suggest. Maybe something like `src/cpu/idt/stage2.rs` and `src/cpu/idt/svsm.rs`? And perhaps a third file for any common code and structures needed.

This looks good to me too

p4zuu:
Does it make sense to have two different IDT handler arrays, one for stage 2 handlers and one for SVSM? Or would it be better to just refill `idt_handler_array` after stage 2 with the SVSM handlers?

p4zuu:
PR is almost ready to be reviewed. I left some TODOs were I asked opened questions to me and reviewers, but here is a new one: do we need to already check CPL before the #VC handling (especially for IOIO) or should we wait for CPL-3 support to be working? Maybe @joergroedel

joergroedel:
> PR is almost ready to be reviewed. I left some TODOs were I asked opened questions to me and reviewers, but here is a new one: do we need to already check CPL before the #VC handling (especially for IOIO) or should we wait for CPL-3 support to be working? Maybe @joergroedel

I think there is no need to check CPL right now. For CPUID it doesn't matter and CPL-3 will not be able to do IOIO directly. For that an IO permission bitmap is needed which we are not planning to implement.

AdamCDunlap (commented):
I noticed that the vc handler tests I was working on were failing in weird ways and I tracked it down to the vc handler modifying the stack frame of the caller. I fixed it in https://github.com/AdamCDunlap/svsm/commit/4798647e126c5a7343f27437c91b3f1cde5a69b0 which has some more details, and some of the tests started passing (the non-passing ones are mostly because of unimplemented things)

src/cpu/idt/common.rs
```diff
@@ -39,18 +35,22 @@ pub const _HV_VECTOR: usize = 28;
...
+pub const X86_TRAP_DB: usize = 0x01;
+
 #[repr(C, packed)]
 #[derive(Default, Debug)]
 pub struct X86ExceptionContext {
     pub regs: X86GeneralRegs,
     pub vector: usize,
     pub error_code: usize,
     pub frame: X86InterruptFrame,
+    pub insn: Instruction,
```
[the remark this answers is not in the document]
p4zuu: Indeed, I noticed that too and fixed it. Thanks!

src/cpu/idt/svsm.rs
```diff
@@ -0,0 +1,136 @@
...
+
+pub fn idt_init() {
+    // Set IST vectors
+    unsafe {
+        init_ist_vectors(&mut GLOBAL_IDT);
+    }
+}
+
+#[no_mangle]
+pub fn generic_idt_handler(ctx: &mut X86ExceptionContext) {
```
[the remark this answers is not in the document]
p4zuu: Indeed, it's a residue from the initial interrupt handling code. Thanks for pointing it.

AdamCDunlap:
> +pub fn cpuid() {
> +    let mut _rax: u64 = 0;
> +    let mut _rbx: u64 = 0;
> +    unsafe {
> +        asm!(
> +        "
> +            mov $2, %rcx
> +            mov $0xD, %rax
> +            cpuid
> +            mov %rax, {rax}
> +            mov %rbx, {rbx}
> +        ", rax = out(reg) _rax, rbx = out(reg) _rbx, options(att_syntax)
> +        );
> +    }
> +}

Just from inspection, I'm worried that the compiler doesn't know that this assembly is modifying registers, so the compiler will assume they are unchanged and emit incorrect code. Additionally, llvm doesn't let you clobber %rbx even if you try to tell it, so that needs special care. Checking the emitted assembly should be able to confirm or deny this theory.

There's a core function for cpuid which should resolve both those issues: https://doc.rust-lang.org/stable/core/arch/x86_64/fn.__cpuid_count.html

I will try out your new version when you push it. I broke my own setup, though, so I'm trying to resolve that 😅

AdamCDunlap:
I ran tests with your new version and it passes the tests for cpuid, 8-bit port IO and 16-bit port IO.

p4zuu:
> Just from inspection, I'm worried that the compiler doesn't know that this assembly is modifying registers, so the compiler will assume they are unchanged and emit incorrect code. Additionally, llvm doesn't let you clobber %rbx even if you try to tell it, so that needs special care. Checking the emitted assembly should be able to confirm or deny this theory.
> 
> There's a core function for cpuid which should resolve both those issues: https://doc.rust-lang.org/stable/core/arch/x86_64/fn.__cpuid_count.html
> Just from inspection, I'm worried that the compiler doesn't know that this assembly is modifying registers, so the compiler will assume they are unchanged and emit incorrect code. Additionally, llvm doesn't let you clobber %rbx even if you try to tell it, so that needs special care. Checking the emitted assembly should be able to confirm or deny this theory.

Good catch. I checked the disassembly and I couldn't find any obvious issue. The function deals with registers without saving anything on the stack. That maybe leaves registers in an inconsistent state, but nothing stroke me about incorrect stack management. Anyway, I tried with `__cpuid_count()` and I everything runs fine. Thanks!

I pushed a new version with IOIO handling in stage 2. If your tests still pass, I will remove the Draft tag :)

00xc:
> Good catch. I checked the disassembly and I couldn't find any obvious issue. The function deals with registers without saving anything on the stack. That maybe leaves registers in an inconsistent state, but nothing stroke me about incorrect stack management. Anyway, I tried with `__cpuid_count()` and I everything runs fine. Thanks!

The generated assembly might not make it obvious. Very likely what was happening was that CPUID was modifying registers that the compiler assumed unchanged because they were not marked `out()` or `inout()`.

AdamCDunlap:
> I pushed a new version with IOIO handling in stage 2. If your tests still pass, I will remove the Draft tag :)

I tried the new version and the previously-passing tests still pass (also IOIO for 32-bit operands, which was failing before due a bug in the test). I don't have them running in stage 2 yet, though.

There's also some IO tests that are still failing when the port is encoded in the instruction, e.g:

```
fn outl_to_testdev_echo(value: u32) {
    unsafe { asm!("outl %eax, $0xe0", in("eax") value, options(att_syntax)) }
}
```

it looks like this happens when the opcode is 0xe4, 0xe5, 0xe6, or 0xe7. I'm not sure if you care about supporting this case, but I thought I'd mention it.

p4zuu:
> > I pushed a new version with IOIO handling in stage 2. If your tests still pass, I will remove the Draft tag :)
> 
> I tried the new version and the previously-passing tests still pass (also IOIO for 32-bit operands, which was failing before due a bug in the test). I don't have them running in stage 2 yet, though.

As we said, I don't expect different behavior in stage 2 and SVSM kernel, but if it's possible for you to test in stage 2, that would be safer. Let me know if there's anything I can do :)

> 
> There's also some IO tests that are still failing when the port is encoded in the instruction, e.g:
> 
> ```
> fn outl_to_testdev_echo(value: u32) {
>     unsafe { asm!("outl %eax, $0xe0", in("eax") value, options(att_syntax)) }
> }
> ```
> 
> it looks like this happens when the opcode is 0xe4, 0xe5, 0xe6, or 0xe7. I'm not sure if you care about supporting this case, but I thought I'd mention it.

Indeed, port specified in an imm8 is currently not supported. We'll add support later if needed.

AdamCDunlap:
I added testing support to stage2 and the cpuid tests pass there as well.

p4zuu:
> I added testing support to stage2 and the cpuid tests pass there as well.

Great. I'll remove the Draft tag of my PR then, and make it open for review. Your testing code looks good to me :)

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
+        SVM_EXIT_CPUID => handle_cpuid(ctx),
+        SVM_EXIT_IOIO => handle_ioio(ctx, ghcb, &insn),
+        _ => Err(SvsmError::Vc(VcError::Unsupported)),
     }
+    .unwrap_or_else(|error| {
+        panic!(
+            "Unhandled #VC exception RIP {:#018x} error code: {:#018x}: error: {:?}",
+            rip, err, error
+        )
+    });
```
00xc, line 121: Same here, the error handling should be inside the match statement.

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
+    snp_cpuid(regs)
+}
+
+fn snp_cpuid(regs: &mut X86GeneralRegs) -> Result<(), SvsmError> {
+    let mut leaf = CpuidLeaf::new(regs.rax as u32, regs.rcx as u32);
+
+    let ret = match cpuid_table_raw(leaf.cpuid_fn, leaf.cpuid_subfn, 0, 0) {
+        None => Err(SvsmError::Vc(VcError::UnknownCpuidLeaf)),
+        Some(v) => Ok(v),
+    }?;
```
00xc, line 189:
```suggestion
    let Some(ret) = cpuid_table_raw(leaf.cpuid_fn, leaf.cpuid_subfn, 0, 0) else {
        return Err(SvsmError::Vc(VcError::UnknownCpuidLeaf));
    };
```

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
+            Ok(())
+        }
+        0xED => {
+            let mut size: GHCBIOSize = GHCBIOSize::Size32;
+            let mut mask = 0xffffffff;
+            if insn.prefixes.nb_bytes > 0 {
+                // inw instruction has a 0x66 operand-size prefix for word-sized operands.
+                size = GHCBIOSize::Size16;
+                mask = 0xffff;
+            }
```
00xc, line 230:
```suggestion
            let (size, mask) = if insn.prefixes.nb_bytes > 0 {
                // inw instruction has a 0x66 operand-size prefix for word-sized operands
                (GHCBIOSize::Size16, u16::MAX as u64)
            } else {
                (GHCBIOSize::Size32, u32::MAX as u64)
            };
```

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
+            ctx.regs.rax = (ret & mask) as usize;
+            Ok(())
+        }
+        0xEE => ghcb.ioio_out(port, GHCBIOSize::Size8, out_value),
+        0xEF => {
+            let mut size: GHCBIOSize = GHCBIOSize::Size32;
+            if insn.prefixes.nb_bytes > 0 {
+                // outw instruction has a 0x66 operand-size prefix for word-sized operands.
+                size = GHCBIOSize::Size16;
+            }
```
00xc, line 242:
```suggestion
            let size = if insn.prefixes.nb_bytes > 0 {
                // outw instruction has a 0x66 operand-size prefix for word-sized operands.
                GHCBIOSize::Size16
            } else {
                GHCBIOSize::Size32
            };
```

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
+
+fn vc_decode_insn(ctx: &mut X86ExceptionContext) -> Result<Instruction, SvsmError> {
+    if !vc_decoding_needed(ctx.error_code) {
+        return Ok(Instruction::default());
+    }
+
+    /*
+     * TODO: the instruction fetch will likely to be handled differently when
+     * #VC exception will be raised from CPL > 0.
+     */
```
00xc, line 258:
```suggestion
     // TODO: the instruction fetch will likely to be handled differently when
     // #VC exception will be raised from CPL > 0.
```

src/cpu/insn.rs
```diff
@@ -0,0 +1,122 @@
...
+                Err(SvsmError::Vc(VcError::DecodeFailed))
+            }
+            _ => Err(SvsmError::Vc(VcError::DecodeFailed)),
+        }
+    }
+}
+/// # Safety
+///
+///  - rip: caller should validate that rip is not set to an address that is not supposed to
+///         be accessible.
```
00xc, line 113: Let's remove the double negative

```suggestion
///  The caller should validate that `rip` is set to a valid address.
```

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
+    let insn_raw = unsafe { insn_fetch(ctx.frame.rip as *const u8) };
```
00xc, line 259: Please add a `// SAFETY:` comment here. In fact I'm not sure this is safe, as RIP could be set close to a boundary to an unmaped page. `insn_fetch` will read `MAX_INSN_SIZE` bytes, which may or may not be in bounds.

src/cpu/insn.rs
```diff
@@ -0,0 +1,122 @@
...
+pub unsafe fn insn_fetch(rip: *const u8) -> [u8; MAX_INSN_SIZE] {
+    let mut insn_raw = [0; MAX_INSN_SIZE];
+
+    unsafe {
+        ptr::copy_nonoverlapping(rip, insn_raw.as_mut_ptr(), MAX_INSN_SIZE);
+    }
+
+    insn_raw
+}
```
00xc, line 122:
```suggestion
pub unsafe fn insn_fetch(rip: *const u8) -> [u8; MAX_INSN_SIZE] {
    rip.cast::<[u8; MAX_INSN_SIZE]>.read()
}
```

00xc (commented):
Added some comments below. Overall error handling could be cleaned up a bit. There are also are some safety concerns I have that I noted below.

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
 pub const X86_TRAP_DB: usize = 0x01;
+pub const X86_TRAP: usize = SVM_EXIT_EXCP_BASE + X86_TRAP_DB;
 
-pub fn handle_vc_exception(ctx: &mut X86ExceptionContext) {
+#[derive(Clone, Copy, Debug)]
+pub enum VcError {
+    Unsupported,
+    DecodeFailed,
+    UnknownCpuidLeaf,
+}
```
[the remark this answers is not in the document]
p4zuu: I think even all enum values should have the same fields. Would doing something like the following would look better to you?

```rust
#[derive(Clone, Copy, Debug)]
pub struct VcError {
    rip: usize,
    code: usize,
    error_type: VcErrorType,
}

#[derive(Clone, Copy, Debug)]
pub enum VcErrorType {
    Unsupported,
    DecodeFailed,
    UnknownCpuidLeaf,
}

impl From<VcError> for SvsmError {
    fn from(e: VcError) -> Self {
        Self::Vc(e)
    }
}

impl fmt::Display for VcError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "Unhandled #VC exception ")?;
        match self.error_type {
            VcErrorType::Unsupported => {
                write!(f, "unsupported #VC exception")?;
            }
            VcErrorType::DecodeFailed => {
                write!(f, "invalid instruction")?;
            }
            VcErrorType::UnknownCpuidLeaf => {
                write!(f, "unknown CPUID leaf")?;
            }
        }
        write!(
            f,
            " RIP: {:#018x}: error code: {:#018x}",
            self.rip, self.code
        )
    }
}
```

I will also change the errors returned by some functions in insn.rs`. `decode()` method currently returns a casted VcError, and I don't really like that. I think it should return something like an `InsnError::DecodeFailed`, that would be casted to a SvsmError in `vc.rs`. Would you agree?

src/cpu/insn.rs
joergroedel: The way to handle invalid RIPs here is to add an exception fixup around the memory reads and return an error from the function when a fault was detected. This can be done on-top of this PR

00xc: Could we reuse `GuestPtr`?

joergroedel:
Okay, so there have been some suggestions included by me on-top. Can you nuke this commit and rebase the base-PR to latest main branch, please? Please also address Carlos' comments when updating.

src/cpu/vc.rs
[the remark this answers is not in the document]
00xc: Yeah that looks better.

Having more specific errors is fine, my only requirement is that we have `SvsmError` in the public interfaces. I would suggest to keep it simple, but if you *really* want to have something more specific I would do:

```rust
pub enum VcError {
    DecodeFailed(InsnError),
    ...
}

impl From<InsnError> for VcError {
    fn from(err: InsnError) -> Self {
        Self::DecodeFailed(err)
    }
}
```

src/cpu/vc.rs
```diff
@@ -4,28 +4,266 @@
...
+
+        SVM_EXIT_CPUID => handle_cpuid(ctx),
+        _ => Err(SvsmError::Vc(VcError::Unsupported)),
     }
+    .unwrap_or_else(|error| {
+        panic!(
+            "Unhandled #VC exception RIP {:#018x} error code: {:#018x}: error: {:?}",
+            rip, err, error
+        )
+    });
```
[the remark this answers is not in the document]
00xc: Then I'd say (at least for now):

```rust
match err {
    X86_TRAP_DB => handle_db_exception(),
    SVM_EXIT_CPUID => {
        if let Err(e) = handle_cpuid(ctx) {
            // handle it
        }
    }
    _ => panic!("Unhandled #VC exception ..."),
}
```

src/cpu/vc.rs
p4zuu: I added a safety comment and a TODO to check that rip+MAX_INSN doesn't go on an umapped page that I will address as soon as this PR is merged

p4zuu:
Rebased on main, and added Carlos' proposals

src/cpu/vc.rs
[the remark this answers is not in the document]
p4zuu: I addressed that with the following:

```rust
    match err {
        // If the debugger is enabled then handle the DB exception
        // by directly invoking the exception handler
        X86_TRAP_DB => handle_db_exception(ctx),
        SVM_EXIT_CPUID => handle_cpuid(ctx).expect("Could not handle CPUID #VC exception"),
        _ => {
            panic!(
                "Unsupported #VC exception RIP {:#018x} error code: {:#018x}",
                rip, err,
            )
        }
    
```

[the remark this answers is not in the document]
p4zuu: At the moment error handling only results in panic, but it might change in the future

00xc:
By the way your commit messages seem to be strangely line wrapped. I suggest wrapping lines at 70 or 72 columns.

00xc (commented):
A few more extra comments

src/cpu/insn.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub length: usize,
+    pub opcode: InstructionField,
+    pub opnd_bytes: usize,
+}
+
+#[derive(Default, Debug, Copy, Clone)]
+pub struct InstructionField {
+    pub bytes: [u8; MAX_INSN_FIELD_SIZE],
+    pub nb_bytes: usize,
+}
```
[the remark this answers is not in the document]
p4zuu: Good point, that's code inspired from C without stepping back enough.
I'll start working on a refactoring but that indeed could be in another PR.
The current `InstructionField.nb_bytes` is not correlated to `InstructionField.bytes.len()` and is used differently, so I'll keep the `nb_bytes` field name or something similar and add a comment why this field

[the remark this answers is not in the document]
joergroedel: Yes, please do that in an additional PR. It makes sense to properly abstract the instruction buffer, but it can be done on-top of this one.

joergroedel:
@p4zuu Can you please work in the suggestions from @00xc and rebase to latest main branch? This should fix the CI failure too.

PR #59, opened 2023-07-04
Improve build and update dependencies

00xc:
Take advantage of conditional compilation, remove unused dependencies and update the rest of them.


00xc:
The fix for the cargo fmt failure is in #57, so I'll rebase once that is merged.

PR #61, opened 2023-07-10
Add support for defining tasks and switching between task contexts

roy-hopkins:
This PR represents a step in implementing support for CPL3 and isolating tasks that run within the context of the coconut SVSM kernel. It adds a new structure: `Task` which allows a task context to be initialised and maintained that can be swapped in and out of execution on a vCPU.

This PR does not introduce any form of scheduling of the tasks. In fact, the task code is not currently used at all within the SVSM kernel. However, the code as it stands can be used to launch an entry point that maintains its own page table (with shared access to kernel memory) in CPL0 using a per-task stack.


src/cpu/msr.rs
```diff
@@ -37,3 +37,30 @@ pub fn write_msr(msr: u32, val: u64) {
...
+
+pub fn rdtsc() -> u64 {
+    let eax: u32;
+    let edx: u32;
+
+    unsafe {
+        asm!("rdtsc",
+             out("eax") eax,
+             out("edx") edx,
+             options(att_syntax));
```
00xc, line 49: We can add `nomem, nostack` to `options()` to help the compiler a bit.

roy-hopkins:
Here's an example of how the code in this PR can be used to launch and switch between tasks.

```
diff --git a/src/svsm.rs b/src/svsm.rs
index 73a25e5..c52265e 100644
--- a/src/svsm.rs
+++ b/src/svsm.rs
@@ -9,10 +9,12 @@
 
 extern crate alloc;
 
+use alloc::boxed::Box;
 use alloc::vec::Vec;
 use svsm::fw_meta::{parse_fw_meta_data, print_fw_meta, validate_fw_memory, SevFWMetaData};
+use svsm::task::Task;
 
-use core::arch::{asm, global_asm};
+use core::arch::global_asm;
 use core::panic::PanicInfo;
 use core::slice;
 use svsm::acpi::tables::load_acpi_cpu_info;
@@ -308,6 +310,8 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
     );
 }
 
+static mut MAIN_TASK: Option<Box<Task>> = None;
+
 #[no_mangle]
 pub extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: VirtAddr) {
@@ -396,15 +400,21 @@ pub extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: VirtAddr) {
 
     log::info!("BSP Runtime stack starts @ {:#018x}", bp);
 
-    // Enable runtime stack and jump to main function
     unsafe {
-        asm!("movq  %rax, %rsp
-              jmp   svsm_main",
-              in("rax") bp.bits(),
-              options(att_syntax));
+        MAIN_TASK = Some(Task::create(svsm_main, 0).expect("Failed to create svsm_main task"));
+        MAIN_TASK
+            .as_mut()
+            .unwrap()
+            .set_current(core::ptr::null_mut());
     }
 }
 
+pub extern "C" fn test_task() {
+    log::info!("Hello from a task");
+    let main_task = unsafe { MAIN_TASK.as_mut().expect("Invalid main task") };
+    main_task.set_current(core::ptr::null_mut());
+}
+
 #[no_mangle]
 pub extern "C" fn svsm_main() {
     // The GDB stub can be started earlier, just after the console is initialised
@@ -419,6 +429,10 @@ pub extern "C" fn svsm_main() {
     // a remote GDB connection
     //debug_break();
 
+    let previous_task = unsafe { MAIN_TASK.as_mut().expect("Invalid main task") };
+    let mut task = Task::create(test_task, 0).expect("Failed to create task");
+    task.set_current(previous_task.as_mut());
+
     invalidate_stage2().expect("Failed to invalidate Stage2 memory");
 
     let fw_cfg = FwCfg::new(&CONSOLE_IO);
```

src/task/tasks.rs
```diff
@@ -0,0 +1,404 @@
...
+        movq    %gs, %rax
+        pushq   %rax
+        movq    %fs, %rax
+        pushq   %rax
+        movq    %es, %rax
+        pushq   %rax
+        movq    %ds, %rax
+        pushq   %rax
+        movq    %cs, %rax
+        pushq   %rax
```
joergroedel, line 365: No need to save and restore the segment selectors, as we are switching between SVSM kernel tasks here and they will all run with the same segments. User-space segment switches will happen in the entry/exit code.

joergroedel (commented):
I think this looks good and is a step in the right direction. Added one minor comment to the task switching assembly.

src/task/tasks.rs
roy-hopkins: The segment selectors are pushed here but they are not restored below (with a comment saying this). It's quite useful to save them because the values can then be inspected via the gdbstub.

PR #62, opened 2023-07-10
ACPI: reduce raw pointer and unsafe usage

00xc:
Reduce raw pointer and unsafe usage wherever possible in the ACPI table parsing code. This includes some potential memory safety bugs that could happen with malformed firmware data.

Most of this is done by using vectors instead of raw allocations and slicing said vectors where references are needed. The main advantage is that both vectors and slices track their sizes, so bounds checking is much simpler, while allowing to cast back to pointer types when needed.


p4zuu:
Good to see that `load_acpi_cpu_info()` gets hardened. `offset += l as usize;` in a huge unsafe block was indeed risky I think

00xc:
> Good to see that `load_acpi_cpu_info()` gets hardened. `offset += l as usize;` in a huge unsafe block was indeed risky I think

Yeah, so the loop was guarded by `while offset < len`, which only checks that the beginning of the entry is within bounds, but not the end. In other words, if my assessment is correct, these could access the content out of bounds:

https://github.com/coconut-svsm/svsm/blob/9a389c99acab96f2cf1948992165bd1969272886/src/acpi/tables.rs#L322-L324

https://github.com/coconut-svsm/svsm/blob/9a389c99acab96f2cf1948992165bd1969272886/src/acpi/tables.rs#L327-L329

https://github.com/coconut-svsm/svsm/blob/9a389c99acab96f2cf1948992165bd1969272886/src/acpi/tables.rs#L335-L337

PR #64, opened 2023-07-13
task/schedule: Implement scheduler and per-cpu runqueues

roy-hopkins:
This PR builds on the merged task switching PR https://github.com/coconut-svsm/svsm/pull/61 and also the draft scheduler PR https://github.com/coconut-svsm/svsm/pull/50 to add scheduling support to coconut-SVSM.

Two different task lists have been implemented in task/schedule:

* Firstly all newly created tasks are added to a simple doubly-linked list where they remain until the task is terminated and subsequently closed. An instance of the tasklist is provided globally and can be used to determine the status of every task.

* Secondly, an RB tree based list is implemented that sorts tasks based on their calculated execution time. The shortest running task and hence the highest priority task for scheduling is leftmost on the tree. Each CPU creates an instance of the RB tree named `runqueue` which maintains all tasks in RUNNING state that should share time from the allocated CPU.

Currently, there is no load balancing for migrating tasks between each CPU, instead each time a task is created the newly created task is allocated to the currently running CPU.

The entry point for every vCPU in the VM has been converted to launch using tasks.

### DIFF
--- a/src/cpu/percpu.rs
+++ b/src/cpu/percpu.rs
@@ -193,17 +193,11 @@ pub struct PerCpu {
     pub vrange_2m: VirtualRange,
 }
 
-impl Default for PerCpu {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl PerCpu {
-    pub fn new() -> Self {
+    fn new(apic_id: u32) -> Self {
         PerCpu {
             online: AtomicBool::new(false),
-            apic_id: 0,
+            apic_id,
             pgtbl: SpinLock::<PageTableRef>::new(PageTableRef::unset()),
             ghcb: ptr::null_mut(),
             init_stack: None,
@@ -222,8 +216,7 @@ impl PerCpu {
         let vaddr = allocate_zeroed_page()?;
         unsafe {
             let percpu = vaddr.as_mut_ptr::<PerCpu>();
-            (*percpu) = PerCpu::new();
-            (*percpu).apic_id = apic_id;
+            (*percpu) = PerCpu::new(apic_id);
             PERCPU_AREAS.push(PerCpuInfo::new(apic_id, vaddr));
             Ok(percpu)
         }
--- a/src/cpu/percpu.rs
+++ b/src/cpu/percpu.rs
@@ -25,6 +25,7 @@ use crate::mm::{
 use crate::sev::ghcb::GHCB;
 use crate::sev::utils::RMPFlags;
 use crate::sev::vmsa::{allocate_new_vmsa, VMSASegment, VMSA};
+use crate::task::RunQueue;
 use crate::types::{PAGE_SHIFT, PAGE_SHIFT_2M, PAGE_SIZE, PAGE_SIZE_2M, SVSM_TR_FLAGS, SVSM_TSS};
 use alloc::sync::Arc;
 use alloc::vec::Vec;
@@ -191,6 +192,9 @@ pub struct PerCpu {
     pub vrange_4k: VirtualRange,
     /// Address allocator for per-cpu 2m temporary mappings
     pub vrange_2m: VirtualRange,
+
+    /// Task list that has been assigned for scheduling on this CPU
+    runqueue: RWLock<RunQueue>,
 }
 
 impl PerCpu {
@@ -209,6 +213,7 @@ impl PerCpu {
             vm_range: VMR::new(SVSM_PERCPU_BASE, SVSM_PERCPU_END, PTEntryFlags::GLOBAL),
             vrange_4k: VirtualRange::new(),
             vrange_2m: VirtualRange::new(),
+            runqueue: RWLock::new(RunQueue::new(apic_id)),
         }
     }
 
@@ -554,6 +559,17 @@ impl PerCpu {
     /// * `pt` - The page table to populate the the PerCpu range into
     pub fn populate_page_table(&self, pt: &mut PageTableRef) {
         self.vm_range.populate(pt);
+    }
+
+    /// Allocate any candidate unallocated tasks from the global task list to our
+    /// CPU runqueue.
+    pub fn allocate_tasks(&mut self) {
+        self.runqueue.lock_write().allocate();
+    }
+
+    /// Access the PerCpu runqueue protected with a lock
+    pub fn runqueue(&self) -> &RWLock<RunQueue> {
+        &self.runqueue
     }
 }
 
--- a/src/locking/rwlock.rs
+++ b/src/locking/rwlock.rs
@@ -156,4 +156,32 @@ impl<T: Debug> RWLock<T> {
             data: unsafe { &mut *self.data.get() },
         }
     }
+
+    /// Waits then locks the RWLock, returning a mutable pointer to the
+    /// protected item. The lock must be released with a call to
+    /// [`Self::unlock_write_direct()`] when access to the protected resource is
+    /// no longer exclusively required.
+    pub fn lock_write_direct(&self) -> *mut T {
+        let guard = self.lock_write();
+        core::mem::forget(guard);
+        self.data.get()
+    }
+
+    /// Unlocks the RWLock, relinquishing access to the raw pointer
+    /// that was gained by a previous call to [`Self::lock_write_direct()`].
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the raw pointer returned by a
+    /// previous call to [`Self::lock_write_direct()`] is not used after
+    /// calling this function. Although the pointer may still point
+    /// to a valid object there is no guarantee of this and use of
+    /// the pointer is undefined behaviour.
+    ///
+    /// In order to gain mutable or immutable access to the object
+    /// the caller must again restablish the RWLock.
+    pub unsafe fn unlock_write_direct(&self) {
+        // There are no readers - safe to just set lock to 0
+        self.rwlock.store(0, Ordering::Release);
+    }
 }
--- a/src/task/schedule.rs
+++ b/src/task/schedule.rs
@@ -57,6 +57,7 @@ struct TaskSwitch {
 pub struct RunQueue {
     tree: Option<RBTree<TaskTreeAdapter>>,
     current_task: Option<TaskPointer>,
+    terminated_task: Option<TaskPointer>,
     id: u32,
     task_switch: TaskSwitch,
 }
@@ -69,6 +70,7 @@ impl RunQueue {
         Self {
             tree: None,
             current_task: None,
+            terminated_task: None,
             id,
             task_switch: TaskSwitch {
                 previous_task: None,
@@ -159,13 +161,30 @@ impl RunQueue {
 
     fn update_current_task(&mut self) -> Option<TaskPointer> {
         let task_node = self.current_task.take()?;
-        task_node.task.lock_write().runtime.schedule_out();
+        let task_state = {
+            let mut task = task_node.task.lock_write();
+            task.runtime.schedule_out();
+            task.state
+        };
 
-        // Reinsert the node into the tree so the position is updated with the new runtime
-        let mut task_cursor = unsafe { self.tree().cursor_mut_from_ptr(task_node.as_ref()) };
-        task_cursor.remove();
-        self.tree().insert(task_node.clone());
-        Some(task_node)
+        if task_state == TaskState::TERMINATED {
+            // The current task has terminated. Make sure it doesn't get added back
+            // into the runtime tree, but also we need to make sure we keep a
+            // reference to the task because the current stack is owned by it.
+            // Put it in a holding location which will be cleared by the next
+            // active task.
+            unsafe {
+                self.deallocate(task_node.clone());
+            }
+            self.terminated_task = Some(task_node);
+            None
+        } else {
+            // Reinsert the node into the tree so the position is updated with the new runtime
+            let mut task_cursor = unsafe { self.tree().cursor_mut_from_ptr(task_node.as_ref()) };
+            task_cursor.remove();
+            self.tree().insert(task_node.clone());
+            Some(task_node)
+        }
     }
 
     /// Helper function that determines if a task is a candidate for allocating
@@ -232,6 +251,21 @@ impl RunQueue {
             self.task_switch.next_task = None;
         }
     }
+
+    /// Deallocate a task from a per CPU runqueue but leave it in the global task list
+    /// where it can be reallocated if still in the RUNNING state.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the function is passed a valid task_node as
+    /// this function dereferences the pointer contained within the task_node. A
+    /// [`TaskPointer`] uses an [`Arc`] to manage the lifetime of the contained pointer
+    /// making it difficult to pass an invalid pointer to this function.
+    unsafe fn deallocate(&mut self, task_node: TaskPointer) {
+        let mut cursor = self.tree().cursor_mut_from_ptr(task_node.as_ref());
+        cursor.remove();
+        task_node.task.lock_write().allocation = None;
+    }
 }
 
 /// Global task list
@@ -261,6 +295,16 @@ impl TaskList {
             cursor.move_next();
         }
         None
+    }
+
+    fn terminate(&mut self, task_node: TaskPointer) {
+        // Set the task state as terminated. If the task being terminated is the
+        // current task then the task context will still need to be in scope until
+        // the next schedule() has completed. Schedule will keep a reference to this
+        // task until some time after the context switch.
+        task_node.task.lock_write().state = TaskState::TERMINATED;
+        let mut cursor = unsafe { self.list().cursor_mut_from_ptr(task_node.as_ref()) };
+        cursor.remove();
     }
 }
 
@@ -309,6 +353,15 @@ pub fn is_current_task(id: u32) -> bool {
     }
 }
 
+pub unsafe fn current_task_terminated() {
+    let mut rq = this_cpu().runqueue().lock_write();
+    let task_node = rq
+        .current_task
+        .as_mut()
+        .expect("Task termination handler called when there is no current task");
+    TASKLIST.lock().terminate(task_node.clone());
+}
+
 pub fn schedule() {
     this_cpu_mut().allocate_tasks();
 
@@ -318,4 +371,12 @@ pub fn schedule() {
             (*next_task).set_current(prev_task);
         }
     }
+
+    // We're now in the context of the new task. If the previous task had terminated
+    // then we can release it's reference here.
+    let _ = this_cpu_mut()
+        .runqueue()
+        .lock_write()
+        .terminated_task
+        .take();
 }
--- a/src/task/schedule.rs
+++ b/src/task/schedule.rs
@@ -6,17 +6,24 @@
 
 extern crate alloc;
 
+use core::ptr::null_mut;
+
 use super::Task;
+use super::{tasks::TaskRuntime, TaskState, INITIAL_TASK_ID};
 use crate::error::SvsmError;
 use crate::locking::{RWLock, SpinLock};
 use alloc::boxed::Box;
 use alloc::sync::Arc;
-use intrusive_collections::{intrusive_adapter, LinkedList, LinkedListAtomicLink};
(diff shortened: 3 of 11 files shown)


src/task/schedule.rs
```diff
@@ -0,0 +1,94 @@
...
+    let node = Rc::new(TaskNode {
+        list_link: Link::default(),
+        task: RefCell::new(task),
+    });
+    let node_ret = node.clone();
+    {
+        let mut tl = TASKLIST.lock();
+        tl.list().push_front(node);
+    }
+    Ok(node_ret)
```
00xc: I don't think we need to scope this, no? The lock will be dropped at the end of the function:

```rust
    ...
    let node = ...;
    TASKLIST.lock().push_front(node.clone());
    Ok(node)
}
```

src/task/schedule.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub fn close_task(task: TaskPointer) -> Result<(), TaskError> {
```
00xc, line 257: I would add a comment here as requested by clippy. The caller must ensure that the provided pointer comes from the global TASKLIST structure.

src/task/schedule.rs
```diff
@@ -8,25 +8,156 @@ extern crate alloc;
...
+            current_task: None,
+        }
+    }
+
+    pub fn tree(&mut self) -> &mut RBTree<TaskTreeAdapter> {
+        self.tree
+            .get_or_insert_with(|| RBTree::<TaskTreeAdapter>::new(TaskTreeAdapter::new()))
+    }
+
+    pub fn get_task(&self, id: u32) -> Option<Rc<TaskNode>> {
```
00xc, line 62: Returning an `Option<TaskPointer>` is probably more clear since we introduced that typedef.

src/task/schedule.rs
```diff
@@ -8,25 +8,156 @@ extern crate alloc;
...
+    pub fn schedule(&mut self) -> (Option<*mut Box<Task>>, *mut Task) {
+        // Update the state of the current task. This will change the runtime value which
+        // is used as a key in the RB tree therefore we need to remove and reinsert the
+        // task.
+        let current_task_node = self.update_current_task();
+
+        // Find the task with the lowest runtime. The tree only contains running tasks that
+        // are to be scheduled on this vCPU.
+        let tree = &mut self.tree();
+        let cursor = tree.lower_bound(Bound::Included(&0));
```
00xc, line 96: This does not need to be `&mut`, as `tree()` already returns a mutable reference. These two lines can be:

```rust
let cursor = self.tree().lower_bound(Bound::Included(&0));
```

src/task/schedule.rs
00xc: In my example the lock is dropped even before, since we hold no reference to the inner tasklist, so it is effectively equivalent to your code.

src/task/schedule.rs
```diff
@@ -8,25 +8,156 @@ extern crate alloc;
...
+    fn update_current_task(&mut self) -> Option<TaskPointer> {
+        // This function leaves the tree in an invalid state as it takes the current
+        // task, replacing it with None. The caller must assign a new task or reassign
+        // the current task to the tree before resuming.
+        let task_ptr = self.current_task.take();
+        if task_ptr.is_some() {
+            let mut task_cursor = unsafe {
+                self.tree()
+                    .cursor_mut_from_ptr(task_ptr.as_ref().unwrap().as_ref())
+            };
```
00xc, line 137: We can avoid the unwrap here, plus a level of indentation:

```rust
fn update_current_task(&mut self) -> Option<TaskPointer>{
    ...
    let task_ptr = self.current_task.take()?;
    let mut task_cursor = unsafe {
        self.tree()
            .cursor_mut_from_ptr(task_ptr.as_ref())
    };
    ...
    Some(task_ptr)
}
```

src/task/schedule.rs
```diff
@@ -8,25 +8,156 @@ extern crate alloc;
...
+            (None, core::ptr::null_mut())
+        } else {
+            (Some(next_task_ptr), current_task_ptr)
+        }
+    }
+
+    fn update_current_task(&mut self) -> Option<TaskPointer> {
+        // This function leaves the tree in an invalid state as it takes the current
+        // task, replacing it with None. The caller must assign a new task or reassign
+        // the current task to the tree before resuming.
```
00xc, line 131: I think this is mostly fine since it is a private function, but perhaps we could annotate it with unsafe.

src/task/schedule.rs
```diff
@@ -277,6 +277,33 @@ pub fn is_current_task(id: u32) -> bool {
...
+        current_task.runtime.terminated();
+
+        // Remove it from the CPU scheduler
+        unsafe {
+            let mut cursor = this_cpu_mut()
+                .runqueue
+                .tree()
+                .cursor_mut_from_ptr(this_task.as_ref());
+            cursor.remove();
+        }
```
00xc, line 302: So I'm not sure I fully understand, but isn't this what `TaskScheduler::deallocate()` is already doing? Any reason we're not using that function?

Also, I don't see `close_task()` called anywhere, it seems to me that functionality should be done from `TaskScheduler::deallocate()`, which should take `&mut self` (at the moment it is an associated method, which does not make much sense). Otherwise the task will stay forever in the tasklist, right?

00xc: So feel free to correct me, but I'd say we should do something like the following (have not tested it):

```rust
TASKLIST.lock().deallocate(this_task.clone(), &mut this_cpu_mut().runqueue);
```

And `deallocate()` should take care of removing the task from the runqueue and the global tasklist.

src/task/schedule.rs
roy-hopkins: The restricted scope is not required in this case - you're right. The reason it is there is because it is required in a later commit where this function automatically switches to the new task. I'll fix the scope in this commit and introduce it where it is required.

See https://github.com/roy-hopkins/svsm/commit/d3ec90c79337a328ff8ac18f1a02e74f40022415#diff-9fb0b887ee481220d8f2b10fda49c9aa708ec4835bb3e7747b3775a187ae0d88

src/task/schedule.rs
roy-hopkins: There are a couple of other places where `TaskPointer` can replace `Rc<TaskNode>`. I'll update them all.

src/task/schedule.rs
roy-hopkins: The ideas is that there is a distinction between deallocating and closing a task. 

Deallocation happens when a task is removed from the runqueue of a CPU, whether it has been terminated or not. This can happen, for example, when a task is migrated between CPUs. 

A task is closed at some point after it has terminated and deallocated and in the current implementation, removes the task from the global tasklist. If this is the last reference to the task then the task memory is freed at this point, including the task stack.

The freeing of the stack is a critical point here as when the task is flagged as terminated it is still the currently running task. By keeping the task valid in the global list until it is closed after termination means we do not have a problem with the task being freed whilst it is still active. 

However, having thought about this more, it would be much neater to remove `close_task()` and use reference counting to actually close and free terminated tasks. I've been working on this and will push the changes here once I have it working nicely.

src/task/schedule.rs
roy-hopkins: This function has now been changed and although it still removes the current task it is not actually an invalid state so I've removed the comment rather than marking the function as unsafe.

src/task/schedule.rs
roy-hopkins: I've removed the `allow` as the function now takes a safe parameter.

src/task/schedule.rs
roy-hopkins: The latest version includes a bit more of a rework than I originally planned. This is mostly due to 2 things:

1. Removing `close_task()` and replacing it purely with reference counting meant that a reference to the current task needed to be maintained until after the termination handler was called.
2. Using the gdbstub to debug changes for the above uncovered a few issues when working with multiple threads running simultaneously on multiple CPUs.

roy-hopkins:
The latest version includes a bit more of a rework than I originally planned. This is mostly due to 2 things:

1. Removing `close_task()` and replacing it purely with reference counting meant that a reference to the current task needed to be maintained until after the termination handler was called.
2. Using the gdbstub to debug changes for the above uncovered a few issues when working with multiple threads running simultaneously on multiple CPUs.

src/task/schedule.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl TaskList {
+    pub const fn new() -> Self {
+        Self { list: None }
+    }
+
+    pub fn list(&mut self) -> &mut LinkedList<TaskListAdapter> {
+        self.list
+            .get_or_insert_with(|| LinkedList::<TaskListAdapter>::new(TaskListAdapter::new()))
```
00xc, line 223: Just a nitpick, but rustc can already figure out this generic, we can simply do:

```rust
.get_or_insert_with(|| RBTree::new(TaskTreeAdapter::new()))
```

src/task/schedule.rs
```diff
@@ -9,22 +9,170 @@ extern crate alloc;
...
+    pub const fn new() -> Self {
+        Self {
+            tree: None,
+            current_task: None,
+        }
+    }
+
+    pub fn tree(&mut self) -> &mut RBTree<TaskTreeAdapter> {
+        self.tree
+            .get_or_insert_with(|| RBTree::<TaskTreeAdapter>::new(TaskTreeAdapter::new()))
```
00xc, line 61: Same here, the generic in RBTree can be omitted.

src/task/schedule.rs
```diff
@@ -9,22 +9,170 @@ extern crate alloc;
...
+                {
+                    // Restrict the scope of the mutable borrow because the insertion
+                    // into the tree also tries to borrow the task.
+                    let mut t = cursor.get().unwrap().task.borrow_mut();
+                    t.allocation = Some(id);
+                    t.runtime.set(lowest_runtime);
+                }
+                self.tree()
+                    .insert(cursor.as_cursor().clone_pointer().unwrap());
+            }
```
00xc, line 192: Unless I'm missing something about the borrowing here, we can get rid of an unwrap by using `Option::filter()`:

```rust
if let Some(task_node) = cursor.get().filter(|task_node| Self::is_cpu_candidate(...)) {
    {
        let mut t = task_node.task.borrow_mut();
        ...
    }
    ...
}
```

src/task/schedule.rs
```diff
@@ -220,5 +221,25 @@ pub fn create_task(
         task: RefCell::new(task),
     });
     TASKLIST.lock().list().push_front(node.clone());
+    // Allocate any unallocated tasks (including the newly created one)
+    // to the current CPU
+    this_cpu_mut().runqueue.allocate(this_cpu().get_apic_id());
```
00xc: It's unfortunate that we grab the `TASKLIST` lock twice in a row (here and in `RunQueue::allocate()`. I don't think it will be too important for performance so perhaps this is just a nitpick, but I wonder if we could rework this to avoid it, as it seems a bit awkward. Maybe having `allocate()` take a mutable reference to the tasklist? I think we already had something like that in a previous version of the PR. It's not pretty, but it would make the code more explicit.

Also, we can avoid getting a reference to the PerCpu struct twice:

```rust
let this_cpu = this_cpu_mut();
let mut tl = TASKLIST.lock();
tl.push_front(node.clone());
this_cpu.runqueue.allocate(this_cpu.get_apic_id(), &mut tl);
```

src/task/schedule.rs
```diff
@@ -0,0 +1,296 @@
...
+        .as_mut()
+        .expect("Task termination handler called when there is no current task");
+
+    // Set the task state as terminated. The task context is still current until
+    // schedule() is called which will then keep a reference until some time
+    // after the task has been switched.
+    task_node.task.borrow_mut().state = TaskState::TERMINATED;
+    let mut tl = TASKLIST.lock();
+    let mut cursor = unsafe { tl.list().cursor_mut_from_ptr(task_node.as_ref()) };
+    cursor.remove();
```
00xc, line 284: Just for symmetry I would have this as a method on `TaskList`, just like we have `RunQueue::deallocate()`, so something like:

```rust
TASKLIST.lock().terminate(task_node);
```

src/task/schedule.rs
roy-hopkins: The problem with grabbing the tasklist into a variable though is that we need to ensure the variable goes out of scope before `schedule()` is called. I've made the suggested change in the latest version and have introduced an explicit scope.

00xc (approved):
LGTM

roy-hopkins:
Changes since last version:

* Rebase on current main branch (to remove clippy warning)
* 7dc8633: Remove GDB stack increase in address_space.rs when gdbstub feature is enabled and update commit message.

roy-hopkins:
Rebased on main.

roy-hopkins:
Another update push. Changes this time:

Rebased on main which introduced a couple of changes that required reworks in this PR:
  * Fixes to Send/Sync for `Spinlock` highlighted some issues in the lock and ref counting primitives used in the RunQueue and TaskList. These have been resolved. See individual commit updates below.
  * VirtualMemory changes. Although this had little impact with the previous implementation, the PR has been updated to take advantage of VMKernelStack. This is handled in a single patch which has been submitted as a separate PR but forms the base for this patch series. See #136 .
 
Individual patch updates:

* task, percpu: Use VMR to manage the task pagetable and VMKernelStack to allocate the task kernel stack
    New commit to use new virtual memory management code. Duplicated in #136.
* task/schedule: Implement a global task list:
    Update locking and reference counting primitives used to protect `TaskNode`  to make safe for `Send`.  Impl Sync for TaskNode. Code was previously using `Rc` which was unsafe in the `TaskNode` context. Switched to `Arc`. Used `RWLock` to protect the `Box` containing the `Task` to synchronise access.
* task/schedule: Add a task runqueue for managing per CPU task switches:
   Assign APIC id to runqueue when created. Update locking as above.
* percpu: Assign apic_id in new():
    New commit
* cpu/percpu: Add scheduler instance to each vCPU
    Add `RWLock` locking to runqueue.
* task/schedule: Add schedule() to switch between tasks
    Allocate tasks to current CPU in schedule() instead of create_task().

src/task/schedule.rs
```diff
@@ -0,0 +1,316 @@
...
+        .expect("Task termination handler called when there is no current task");
+    TASKLIST.lock().terminate(task_node.clone());
+}
+
+pub fn schedule() {
+    this_cpu_mut().allocate_tasks();
+
+    let (next_task, current_task) = this_cpu().runqueue().lock_write().schedule();
+    if let Some(next_task) = next_task {
+        unsafe { (*next_task).set_current(current_task) };
```
Freax13, line 306: How is it guaranteed the `next_task` is not aliased by another vCPU taking the task's lock?

src/cpu/percpu.rs
```diff
@@ -191,19 +215,22 @@ pub struct PerCpu {
...
     pub vrange_2m: VirtualRange,
+
+    /// Task list that has been assigned for scheduling on this CPU
+    runqueue: RWLock<RunQueue>,
 }
 
 impl Default for PerCpu {
     fn default() -> Self {
-        Self::new()
+        Self::new(0)
```
Freax13, line 225: This `Default` implementation is unused. Have you considered just removing it?

Freax13 (commented):
Sorry for the late review.

roy-hopkins:
@Freax13 - many thanks for the review. I've hopefully addressed your comments with the following changes to individual commits:

* percpu: Assign apic_id in new
Removed 'default' as suggested.

* task/tasks: Add an optional hook that can be called after task switch
New commit to help support holding locks over a task switch.

* task/schedule: Implement a global task list
Switch to using an `AtomicLink`. Incorporate suggested change for get_task().

* locking/rwlock: Add ability to lock/unlock directly without a guard
New commit to allow scope of `RWLock` to be manually managed.

* task/schedule: Add a task runqueue for managing per CPU task switches
Switch to `RBTreeAtomicLink`. 
Fix comment typo.
The biggest change to this commit is to support holding the previous and next task `RWLock`s over the task list. This is achieved by locking them in `schedule()` then unlocking them in a new task switch hook.

* task/schedule: Add schedule() to switch between tasks
Hold the runqueue `RWLock` until after the task switch is complete.

src/task/schedule.rs
```diff
@@ -0,0 +1,316 @@
...
+    tree_link: RBTreeLink,
+    list_link: Link,
+    pub task: RWLock<Box<Task>>,
+}
+
+// SAFETY: Send + Sync is required for Arc<TaskNode> to implement Send. The `task`
+// member is Send + Sync but the intrusive_collection links are only Send. The only
+// access to these is via the intrusive_adapter! generated code which does not use
+// them concurrently across threads.
+unsafe impl Sync for TaskNode {}
```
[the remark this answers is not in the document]
roy-hopkins: Good spot, although unintentionally I don't think the race condition can occur as the TASKLIST lock is held whenever a task node is added to a runqueue, meaning there is no possibility for another thread to jump in between the check for `t.allocation.is_none()` and the task being added to the queue.

I think it makes sense to switch over to using those atomic link types, although that would convert any race condition from unpredictability to a lockup of the thread.

Ignoring the TASKLIST lock, I've fixed the potential race condition with a second check of `t.allocation` when the task lock is held. The insertion/removal from the tree must remain in sync with `t.allocation` in all code paths.

src/task/schedule.rs
roy-hopkins: So, `next_task`, `prev_task` and the percpu runqueue all had potential for aliasing issues due to the fact that the locks were released before the task switch occurred.

I've made a fairly significant change to the locking to extend `RWLock` to allow the scope to be manually managed rather than purely through a scope-based variable. This allows the runqueue to take the lock over the previous and next tasks and persist that lock, tracked by a structure member through to the completion of the switch which is handled by a new hook.

The percpu runqueue is now also locked in the same way, with the lock being taken at the start of the context switch and released after the switch has completed.

src/task/tasks.rs
```diff
@@ -315,19 +296,32 @@ impl Task {
...
         let mut pt = (*new_task).page_table.lock();
-        pt.copy_entry(&this_cpu().get_pgtable(), PGTABLE_LVL3_IDX_PERCPU);
+        this_cpu().populate_page_table(&mut pt);
         pt.cr3_value().bits() as u64
     }
 }
 
+#[allow(unused)]
+#[no_mangle]
+extern "C" fn on_switch(new_task: *mut Task) {
```
Freax13, line 317:
```suggestion
extern "C" fn on_switch(new_task: &mut Task) {
```

That way no unsafe is required inside the function.

src/task/schedule.rs
```diff
@@ -0,0 +1,372 @@
...
+        .expect("Task termination handler called when there is no current task");
+    TASKLIST.lock().terminate(task_node.clone());
+}
+
+pub fn schedule() {
+    this_cpu_mut().allocate_tasks();
+
+    // The runqueue lock needs to be held until after the task switch has ocurred.
+    // The lock will either be released in schedule() if no task switch is required
+    // or in task_switch_hook().
```
Freax13, line 354: Why does the lock have to be held until the task switch?

src/locking/rwlock.rs
```diff
@@ -156,4 +156,32 @@ impl<T: Debug> RWLock<T> {
...
+            {
+                break;
+            }
+            core::hint::spin_loop();
+        }
+
+        // Now locked for write - wait until all readers finished
+        let val: u64 = self.wait_for_readers();
+        assert!(val == compose_val(0, 1));
+        unsafe { &mut *self.data.get() }
```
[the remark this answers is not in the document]
roy-hopkins: That shows my inexperience with rust!

src/locking/rwlock.rs
```diff
@@ -156,4 +156,32 @@ impl<T: Debug> RWLock<T> {
             data: unsafe { &mut *self.data.get() },
         }
     }
+
+    pub fn lock_write_direct(&self) -> *mut T {
```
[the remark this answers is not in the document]
roy-hopkins: Nice! I've taken your implementation.

PR #69, opened 2023-08-02
Implement the attestation report feature

cclaudio:
This PR implements the attestation report feature discussed in the issue #67 

`make test` will test the main functions used to encrypt and decrypt messages.

The branch [attestation-report-test](https://github.com/cclaudio/svsm/tree/attestation-report-test) can be used to test the `get_report()` and `get_ext_report()` functions. It contains one extra patch that calls them at boot time and print some information in the log.

PR #70, opened 2023-08-08
address_space: add STACK_PAGES_GDB constant

stefano-garzarella:
Some new features may require increasing the stack size: https://github.com/coconut-svsm/svsm/pull/69#discussion_r1282900603

To avoid problems when the gdb stub is enabled, it is best to define STACK_PAGES as a constant plus extra pages depending on the features (e.g. STACK_PAGES_GDB).

This way we have a single point to modify for the base of STACK_PAGES.

No functional changes.

### DIFF
--- a/src/mm/address_space.rs
+++ b/src/mm/address_space.rs
@@ -85,10 +85,11 @@ pub const SIZE_LEVEL0: usize = 1usize << ((9 * 0) + 12);
 // Stack definitions
 // The GDB stub requires a larger stack.
 #[cfg(feature = "enable-gdb")]
-pub const STACK_PAGES: usize = 16;
+pub const STACK_PAGES_GDB: usize = 12;
 #[cfg(not(feature = "enable-gdb"))]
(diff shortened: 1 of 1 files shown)


joergroedel:
Hmm, this makes me think whether we should go towards growable kernel stacks for the SVSM. On the other hand this means taking page-faults and page allocations during kernel operations (requires a lock-less way to allocate single-order pages).

stefano-garzarella:
@joergroedel if we add modules and functionality statically, I don't know if it could add more complexity.

For what cases would it be useful?

stefano-garzarella:
Okay, I see. I'll take a closer look in the coming weeks (next week I'll be off).
Since I'm fairly new to these topics, any kind of advice, help, is more than welcome :-)

joergroedel:
Take your time, this is out of scope for this PR. And I am not even sure yet it is a good idea to have dynamically growing stacks in the kernel.

PR #73, opened 2023-08-14
Derive or implement Default for some structs

Zildj1an:
I'm working on unit tests and I need the Default trait more widely available. Derive or implement Default trait for PhysAddr, VirtAddr, PTEntry, PageTable and PTPage. This will help, among other things, to implement said unit tests in the future.


src/mm/pagetable.rs
```diff
@@ -121,9 +121,15 @@ bitflags! {
...
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct PTEntry(PhysAddr);
 
+impl Default for PTEntry {
+    fn default() -> Self {
+        PTEntry(PhysAddr::new(0))
+    }
+}
+
```
00xc, line 132: I think this could be derived as well, since you added Default for PhysAddr.

00xc:
Left a couple of comments, otherwise this looks good.

Zildj1an:
This is updated and ready for your review, @00xc

src/cpu/registers.rs
```diff
@@ -24,6 +24,8 @@ pub struct X86GeneralRegs {
     pub rax: usize,
 }
 
+impl Copy for X86GeneralRegs {}
+
```
00xc, line 28: This can also be derived.

src/cpu/registers.rs
```diff
@@ -36,10 +38,13 @@ pub struct X86SegmentRegs {
...
+#[derive(Default, Debug, Clone)]
 pub struct X86InterruptFrame {
     pub rip: usize,
     pub cs: usize,
     pub flags: usize,
     pub rsp: usize,
     pub ss: usize,
 }
+
+impl Copy for X86InterruptFrame {}
```
00xc, line 50: Same here, could be derived

src/elf/mod.rs
```diff
@@ -17,8 +17,9 @@ use core::fmt;
 use core::matches;
 use core::mem;
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub enum ElfError {
+    #[default]
```
00xc, line 22: Not sure how I feel about this. Does it make sense to have a default `ElfError`? If we had an `Other` variant this would probably be fine but in this case I'm not sure.

src/fs/api.rs
```diff
@@ -18,8 +18,9 @@ use packit::PackItError;
 const MAX_FILENAME_LENGTH: usize = 64;
 pub type FileName = FixedString<MAX_FILENAME_LENGTH>;
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, Default)]
 pub enum FsError {
+    #[default]
     Inval,
```
00xc, line 24: This one I think is fine since EINVAL is usually very generic.

src/elf/mod.rs
00xc: To avoid littering the PR with comments, I'll add that I think this applies too for `FwCfgError` and `PageStateChangeOp` - i.e. I don't know if a default makes much sense. If this is essential for the tests you're writing then keep the derives, but otherwise I'd rather not have them on these enums.

Zildj1an: Your concern makes sense, I'm removing this.

Zildj1an:
Thanks @00xc I have addressed your feedback and taken the liberty of adding your Reviewed-By in the commit message

00xc (approved):

Issue #65, opened 2023-07-14
Provide a decent build system for SVSM


joergroedel:
Design and implement a decent build system for the SVSM to simplify the building process.

The key features should be:

- Support build recipes in one file. The format for that file needs to be extensible (JSON?) and needs to specify the build target and the enabled features. In the future also other things like the firmware to include, boot mode, hypervisor target, and so on.
- Needs to take feature dependencies into account
- Ideally a wrapper around _cargo build_

00xc:
I wonder if we could do most if not all of these in `build.rs` so as to have a more standard way of building things. This would also help packaging when we switch to stable Rust.

joergroedel:
Is it possible to specify additional input files for build.rs? If so we can try to do parts of that in build.rs.

Some parts like the final packaging as done in the Makefile today are probably beyond the scope of build.rs.

00xc:
I'm not sure. I haven't researched it thoroughly, but there's something similar in how the bootimage crate is built and integrated into dependent projects:

https://github.com/rust-osdev/bootimage#configuration

We could perhaps do some parts in Cargo.toml

cclaudio:
Handling cargo feature flags in the `build.rs` seems very limited when it comes to testing. although `cargo test` also calls `build.rs`, something like `#[cfg(not(test))]` does not work inside the `build.rs` because the test feature is not set there.

Any problem with Makefile?

joergroedel:
Makefiles do not provide an easy way to enable/disable features for the build process. The number of features settings we have will likely grow and the current way to enable them does not scale well.

aik:
Is there an easy way to configure coconut to let vim walk through compile errors using the "quickfix" feature of vim?
Normally "compiler cargo" would do but svsm is makefile-based so this does not work. What do you folks do?

joergroedel:
The Makefile also calls 'cargo build', so as a workaround you can run cargo build from vim. You just need to specify the binary to build, either stage2 or svsm. For final testing after all errors are fixed you have to invoke make again so get all binaries packaged into the svsm.bin file.

aik:
[...]

PR #76, opened 2023-08-23
scripts: Build the svsm.bin from a container

cclaudio:
I have seem people asking about a script to build the SVSM from a container (Issue #65)as the build process depends on a recent binutils. 

This PR add scripts to build the project from a docker container. This can be useful for multiple situations e.g., build it for a CI, quick build the SVSM, or build it from a fresh environment for debugging. 

$ ./scripts/docker/build.sh

The script will not stop the container created since there is a chance it can be used again in a short term.

### DIFF
--- a/scripts/docker/build.sh
+++ b/scripts/docker/build.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+CURDIR=$(dirname $(realpath "$0"))
+WORKDIR=$(realpath ${CURDIR}/../..)
+DOCKER_FILE=$WORKDIR/scripts/docker/opensuse-rust.docker
+
+IMAGE_NAME=opensuse-rust
+CONTAINER_NAME=coconut-build
+
+# Command line arguments
+ARG_REUSE=0
+
+####
+#### Help function
+####
+Help()
+{
+    echo "Build the SVSM within a container."
+    echo
+    echo "Syntax: $0 [-r|v|h]"
+    echo "options:"
+    echo "r    Reuse the ${CONTAINER_NAME} to build the SVSM"
+    echo "v    Verbose mode"
+    echo "h    Print this help"
+}
+
+####
+#### ParseOptions function
+####
+ParseOptions()
+{
+    while getopts "rvh" option; do
+        case $option in
+            r) # Reuse option
+                ARG_REUSE=1
+                ;;
+            v) # Verbose mode
+                set -x
+                ;;
+            h) # Display help
+                Help
+                exit
+                ;;
+            \?) # Invalid option
+                echo -e "ERR: Invalid option\n"
+                Help
+                exit
+                ;;
+        esac
+    done
+}
+
+####
+#### Reuse container to build the SVSM
+####
+BuildSvsmReuse()
+{
+    CONTAINER_ID=$(docker ps -q -f name=${CONTAINER_NAME})
+
+    # Create and start the the container if it's not running
+    if [ -z ${CONTAINER_ID} ] ; then
+
+        CONTAINER_ID=$(docker ps -q -f name=coconut-build)
+
+        if [ -z ${CONTAINER_ID} ] ; then
+            docker create \
+                -it --name=${CONTAINER_NAME} \
+                --workdir=${WORKDIR} \
+                --user ${USER} \
+                --mount type=bind,source=${WORKDIR},target=${WORKDIR} \
+                ${IMAGE_NAME} \
+                /bin/bash
+        fi
+
+        docker start ${CONTAINER_NAME}
+    fi
+
+    docker exec \
+        -it ${CONTAINER_NAME} \
+        /bin/bash -c "source $HOME/.cargo/env && make clean && make"
+}
+
+####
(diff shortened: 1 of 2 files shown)


scripts/docker/build.sh
```diff
@@ -0,0 +1,38 @@
...
+    CONTAINER_ID=$(docker ps -q -f name=coconut-build)
+
+    if [ -z ${CONTAINER_ID} ] ; then
+        docker create -it --name=${CONTAINER_NAME} --workdir=${WORKDIR} --user ${USER} --mount type=bind,source=${WORKDIR},target=${WORKDIR} ${IMAGE_NAME} /bin/bash
+    fi
+
+    docker start ${CONTAINER_NAME}
+fi
+
+docker exec -it ${CONTAINER_NAME} /bin/bash -c "source $HOME/.cargo/env && make clean && make"
```
stefano-garzarella, line 38: What about replacing with just `docker run`:
```suggestion
docker run --rm -it --workdir=${WORKDIR} --user ${USER} --userns keep-id --mount type=bind,source=${WORKDIR},target=${WORKDIR} ${IMAGE_NAME} /bin/bash -c "source $HOME/.cargo/env && make clean && make"
```

I added ` --userns keep-id` to properly work with `podman-docker`, eventually what about using podman directly?

scripts/docker/opensuse-rust.docker
```diff
@@ -0,0 +1,23 @@
+# Usage example:
+#
+# git clone https://github.com/coconut-svsm/svsm.git
+# cd svsm
+# docker build -t fedora-rust -f scripsts/fedora-rust.docker .
```
stefano-garzarella, line 5: s/fedora/opensuse

scripts/docker/build.sh
cclaudio: I can add your `docker run --rm ` command line as the default action. It would be more in-sync with the idea that containers are usually ephemeral.

Since that command line does not scale in terms of time (the SVSM crates are downloaded and the index is rebuilt on every build), I will add an option `-r` that when set will force the build to reuse an existing container.

Your command line is failing in Docker v20.10.12 though `--userns: invalid user mode`. This might be one of those command line arguments that are not supported in both podman and docker. If we want to support podman, it looks like we would need to extend the script.

scripts/docker/opensuse-rust.docker
cclaudio: I deleted that line and updated the following one. Thanks.

cclaudio:
All changes applied (v2)

scripts/docker/build.sh
stefano-garzarella: Thanks, LGTM.

We can add podman support later, adding a parameter.

scripts/docker/build.sh
```diff
@@ -0,0 +1,127 @@
...
+#### Reuse container to build the SVSM
+####
+BuildSvsmReuse()
+{
+    CONTAINER_ID=$(docker ps -q -f name=${CONTAINER_NAME})
+
+    # Create and start the the container if it's not running
+    if [ -z ${CONTAINER_ID} ] ; then
+
+        CONTAINER_ID=$(docker ps -q -f name=coconut-build)
```
stefano-garzarella, line 63: Should we add `-a` here?
We can also use `${CONTAINER_NAME}`

```suggestion
        CONTAINER_ID=$(docker ps -q -a -f name=${CONTAINER_NAME})
```

osteffenrh:
Just FYI:
Here is a try to make it work with podman.
podman does not need the manual user mapping, but the SELinux stuff needs additional attention:
https://github.com/coconut-svsm/svsm/commit/845cf0a833c33fec8e7de688c53c8ef5fe87deb8
We can try to make it work with podman later - please just go ahead with the docker-only version first :-)

scripts/docker/build.sh
```diff
@@ -0,0 +1,127 @@
...
+
+####
+#### Build the SVSM in the container, but delete the container afterwards
+####
+BuildSvsmDelete()
+{
+    docker run \
+        --rm -it \
+        --workdir=${WORKDIR} \
+        --user ${USER} \
```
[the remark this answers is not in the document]
cclaudio: Good idea, I will run some experiments with that. Thanks.

stefano-garzarella:
@joergroedel there were still some things to fix (https://github.com/coconut-svsm/svsm/pull/76#discussion_r1303946073) in my opinion, but we can patch on top ;-)

joergroedel:
Yeah, I based the decision on your "the current state is fine with me ;-)" comment :)

Anyway, next time I will ask for explicit approval.

stefano-garzarella:
> Yeah, I based the decision on your "the current state is fine with me ;-)" comment :)

argh, sorry for the confusion ;-)

scripts/docker/build.sh
stefano-garzarella: Followup patch sent: https://github.com/coconut-svsm/svsm/pull/83

PR #77, opened 2023-08-24
Add initial Address allocator implementation

joergroedel:
The address allocator uses a first-fit algorithm and an RBTree as its data structure. This is just a start, more features will be needed in the future, but here is the initial version for review.

### DIFF
--- a/src/mm/address_ranges.rs
+++ b/src/mm/address_ranges.rs
@@ -1,0 +1,204 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Joerg Roedel <jroedel@suse.de>
+
+use crate::types::PAGE_SHIFT;
+use crate::utils::{align_down, align_up};
+use core::cmp::max;
+use intrusive_collections::rbtree::{Link, RBTree};
+use intrusive_collections::{intrusive_adapter, KeyAdapter};
+
+extern crate alloc;
+use alloc::boxed::Box;
+
+struct Range<T> {
+    link: Link,
+    start: usize,
+    end: usize,
+    data: T,
+}
+
+impl<T> Range<T> {
+    const fn new(start: usize, end: usize, data: T) -> Self {
+        Self {
+            link: Link::new(),
+            start,
+            end,
+            data,
+        }
+    }
+
+    fn get(&self) -> &T {
+        &self.data
+    }
+}
+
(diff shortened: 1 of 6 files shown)


00xc (changes requested):
As per Carlos Bilbao's RFC on the mailing list, `pub` structs and methods should have some documentation.

src/mm/address_ranges.rs
```diff
@@ -0,0 +1,204 @@
...
+            data,
+        }
+    }
+
+    fn get(&self) -> &T {
+        &self.data
+    }
+}
+
+intrusive_adapter!(RangeAdapater<T> = Box<Range<T>>: Range::<T> { link: Link });
```
00xc, line 38: Small typo, should be `RangeAdapter`

src/mm/address_ranges.rs
```diff
@@ -0,0 +1,204 @@
...
+        let mut start = align_up(self.pfn_low, align);
+        let mut end = start;
+
+        if size == 0 {
+            return None;
+        }
+
+        let mut cursor = self.tree.front_mut();
+        while !cursor.is_null() {
+            let node = cursor.get().unwrap();
```
00xc, line 76: This is equivalent and avoids an unwrap:

```rust
while let Some(node) = cursor.get() {
    ...
}
```

src/mm/address_ranges.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        if end > start && end - start >= size {
+            cursor.insert_before(Box::new(Range::<T>::new(start, start + size, data)));
+            Some(start << PAGE_SHIFT)
+        } else {
+            None
+        }
+    }
+
+    pub fn alloc(&mut self, size: usize, data: T) -> Option<usize> {
```
00xc, line 97: Maybe the return value should be a `VirtAddr`? Same with the input for `free()` and `get()`

src/mm/address_ranges.rs
joergroedel: Fixed, thanks.

src/mm/address_ranges.rs
joergroedel: Yes, this is better, thanks. Changed it.

src/mm/address_ranges.rs
joergroedel: Converted to VirtAddr.

joergroedel:
During further development I realized that a separate generic address allocator is not the best way, because it will cause a lot of issues with locking. Going forward it will be integrated into the data structure that also keeps track of virtual memory mappings and page-tables. So I close this PR for now, the code here will be submitted as part of another PR.

PR #78, opened 2023-08-24
Page tables: Support heap allocation and page-table freeing

joergroedel:
This implements the pieces to free page-tables and support allocation/de-allocation on the heap.

### DIFF
--- a/src/mm/pagetable.rs
+++ b/src/mm/pagetable.rs
@@ -11,8 +11,8 @@ use crate::cpu::features::{cpu_has_nx, cpu_has_pge};
 use crate::cpu::flush_tlb_global_sync;
 use crate::error::SvsmError;
 use crate::locking::{LockGuard, SpinLock};
-use crate::mm::alloc::allocate_zeroed_page;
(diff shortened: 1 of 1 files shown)


src/mm/pagetable.rs
```diff
@@ -325,6 +331,46 @@ impl PageTable {
...
+                PageTable::free_pgtbl_lvl2(l2_page);
+                free_page(phys_to_virt(entry.address()));
+            }
+        }
+    }
+
+    pub fn free(&mut self) {
+        PageTable::free_pgtbl_lvl3(&self.root);
+    }
+
```
00xc, line 379: Shouldn't `free_pgtbl_lvl{1,2,3}` be methods on `PTPage` that take `&self`? `PageTable::free()` would then simply be `self.root.free_lvl3()`

joergroedel: The design will slightly change, most of the mapping/unmapping/freeing code will be moved to a new PageTablePart struct, which only controls page-table levels 0-2. So further changes are needed here, maybe including to move this code out of the impl block and make then general functions.

joergroedel:
Closing this, it will be submitted in a changed version as part of the virtual memory management work.

PR #79, opened 2023-08-25
Fix build error and clippy warnings from rust 1.74-nightly

joergroedel:


src/sev/utils.rs
```diff
@@ -4,6 +4,9 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
+// For bitflags definitions
+#![allow(clippy::bad_bit_mask)]
```
stefano-garzarella, line 8: Is clippy complains about `RMPFlags`?

What about allow it only for that definition?

joergroedel: It complained about RMPFlags, especially the zero masks for VMPL0 and NONE. I tried to make it only for the struct definition, but clippy didn't accept that. If you know a way to do that, feel free to send a PR to update the annotation :)

stefano-garzarella: Ah okay now I see, I'll check if there is a way.

But just to understand better, what is the purpose of `RMPFlags::VMPL0` flag since it is 0?
Where I can find some info about it?

stefano-garzarella: Anyway, looking at https://github.com/bitflags/bitflags/pull/373 it seems it happens only on bitflags v1.*, I'll take a look if we can switch to v2.*

joergroedel: > 
The RMPFlags are used to generate a bitmask for the RMPUPDATE instruction, where the SVSM can define the access permissions for pages, e.g. giving access to VMPL1 or make a page a VMSA page. For targeting VMPL0 permissions, the flag has to be 0 :)

stefano-garzarella: @joergroedel okay I see, thanks for the details!
I opened https://github.com/coconut-svsm/svsm/pull/80 to switch to a new version of bitflags avoiding to suppress this error.

PR #80, opened 2023-08-25
Update `bitflags` crate to 2.4.0

stefano-garzarella:
Commit a94d48d ("SVSM/sev/utils: Suppress clippy::bad_bit_mask errors") was necessary because we use an older version of bitflags. As suggested in this [1] bitflags issue, let's switch to the latest version available. In this way we can avoid to suppress clippy::bad_bit_mask.

Just minor adjustments are needed to support bitflags 2.4.0.

[1] https://github.com/bitflags/bitflags/pull/373

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,4 +46,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-features -- -D warnings -A clippy::bad_bit_mask
+          args: --all-features -- -D warnings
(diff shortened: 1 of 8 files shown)


00xc (approved):

00xc:
Can we also remove the warning suppression in this PR? This would need to be done in the Github Actions yaml and the pre-commit hook.

stefano-garzarella:
> Can we also remove the warning suppression in this PR? This would need to be done in the Github Actions yaml and the pre-commit hook.

Ooops, I completely forgot about them.

Sure, so I'll add a new commit to this PR removing those suppressions!

stefano-garzarella:
v2:
- removed the warning suppression in the Github Actions yaml and the pre-commit hook

00xc:
Looks good to me, thanks!

PR #81, opened 2023-08-29
build: use stable Rust toolchain

### LINKED ISSUE #15: Convert to use stable Rust
The project currently requires rust-nightly to be compiled. This makes it impossible to build the project in an automated build environment like the ones used to compile Linux distributions.

Make the necessary modifications step by step to build the project using Rust stable.

00xc:
Switch to using the stable Rust toolchain instead of nightly. To do this, we must not need to rebuild parts of the core and standard library, which is an unstable feature. In order to do so, switch to a supported build target (`x86_64-unknown-none`) instead of our own (`svsm-target.json`).

Fixes #15


00xc:
Keeping this as a draft until I figure out how to fix tests, as they need to build the standard library.

stefano-garzarella:
In order to successfully build it on my Fedora 38, I needed to run `rustup target add x86_64-unknown-none`, should we add an hint about it in the INSTALL.md?

Related to INSTALL.md I'd also suggest updating the gdb example:

```patch
diff --git a/INSTALL.md b/INSTALL.md
index b79cdbf..8301fca 100644
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -341,7 +341,7 @@ If you have the source code available on the host system then you can add the
 debug symbols and use source-level debugging:
 

-(gdb) symbol-file target/svsm-target/debug/svsm
+(gdb) symbol-file target/x86_64-unknown-none/debug/svsm

```

I'd also suggest updating `script/docker/build.sh` in something like this:
```patch
diff --git a/scripts/docker/opensuse-rust.docker b/scripts/docker/opensuse-rust.docker
index e46cbbb..13fd729 100644
--- a/scripts/docker/opensuse-rust.docker
+++ b/scripts/docker/opensuse-rust.docker
@@ -19,4 +19,4 @@ RUN zypper ref && \
 USER $USER_NAME
 
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > /tmp/rustup-init.sh && \
-    sh /tmp/rustup-init.sh --default-toolchain nightly-x86_64-unknown-linux-gnu --component rust-src -y
+    sh /tmp/rustup-init.sh --default-toolchain stable-x86_64-unknown-linux-gnu --target x86_64-unknown-none -y
(END)
```

Anyway, cool PR ;-)

svsm-target.json
```diff
@@ -1,14 +0,0 @@
...
-    "llvm-target": "x86_64-unknown-none",
-    "data-layout": "e-m:e-i64:64-f80:128-n8:16:32:64-S128",
-    "arch": "x86_64",
-    "target-endian": "little",
-    "target-pointer-width": "64",
-    "target-c-int-width": "32",
-    "os": "none",
-    "executables": true,
-    "linker-flavor": "gcc",
-    "panic-strategy": "abort",
```
stefano-garzarella, line 11: Just a question to let me understand better the rust targets ;-)

How we are now specifying the `panic-strategy`? Is default to `abort` for `x86_64-unknown-none`?
I found something [here](https://doc.rust-lang.org/rustc/platform-support/x86_64-unknown-none.html) for example about the red-zone, but nothing about panic.

00xc: It is the default for the target, correct:
https://github.com/rust-lang/rust/blob/master/compiler/rustc_target/src/spec/x86_64_unknown_none.rs

00xc:
> In order to successfully build it on my Fedora 38, I needed to run `rustup target add x86_64-unknown-none`, should we add an hint about it in the INSTALL.md?
> 
> Related to INSTALL.md I'd also suggest updating the gdb example

Thanks! I'll add these changes to v2

stefano-garzarella:
> Keeping this as a draft until I figure out how to fix tests, as they need to build the standard library.

What about use nightly only for tests?
I mean something like: ` cargo +nightly test ...`

00xc:
> > Keeping this as a draft until I figure out how to fix tests, as they need to build the standard library.
> 
> What about use nightly only for tests? I mean something like: ` cargo +nightly test ...`

That would be my last resource. However, I managed to run tests by doing:

```
RUSTFLAGS="-C code-model=kernel" cargo test --target=x86_64-unknown-linux-gnu
```

So rebuilding the standard library is not needed at all. And in fact, I've just realized that `code-model=kernel` also works for the regular build, and it's the default for the `x86_64-unknown-none` target.

stefano-garzarella:
> > > Keeping this as a draft until I figure out how to fix tests, as they need to build the standard library.
> > 
> > 
> > What about use nightly only for tests? I mean something like: ` cargo +nightly test ...`
> 
> That would be my last resource. However, I managed to run tests by doing:
> 
> ```
> RUSTFLAGS="-C code-model=kernel" cargo test --target=x86_64-unknown-linux-gnu
> ```
> 
> So rebuilding the standard library is not needed at all. And in fact, I've just realized that `code-model=kernel` also works for the regular build, and it's the default for the `x86_64-unknown-none` target.

Cool! So what about adding it in `.cargo/config.toml`:
```patch
diff --git a/.cargo/config.toml b/.cargo/config.toml
index 4e38721..1a006b6 100644
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,3 +6,7 @@ rustflags = [
        "-C", "code-model=large"
 ]
 
+[target.x86_64-unknown-linux-gnu]
+rustflags = [
+       "-C", "code-model=kernel",
+]
```

Then `cargo test --target=x86_64-unknown-linux-gnu` should work.

00xc:
> Cool! So what about adding it in `.cargo/config.toml`:
> 
> ```diff
> diff --git a/.cargo/config.toml b/.cargo/config.toml
> index 4e38721..1a006b6 100644
> --- a/.cargo/config.toml
> +++ b/.cargo/config.toml
> @@ -6,3 +6,7 @@ rustflags = [
>         "-C", "code-model=large"
>  ]
>  
> +[target.x86_64-unknown-linux-gnu]
> +rustflags = [
> +       "-C", "code-model=kernel",
> +]
> ```
> 
> Then `cargo test --target=x86_64-unknown-linux-gnu` should work.

Yep, I was trying this just now. I will add it to v2 as well.

stefano-garzarella:
LGTM!

I'd just suggest the following changes in INSTALL.md:
```patch

diff --git a/INSTALL.md b/INSTALL.md
index 8301fca..af72adf 100644
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -171,8 +171,9 @@ Building the COCONUT-SVSM
 -------------------------
 
 Building the SVSM itself requires:
-- a recent Rust-nightly compiler and build environment installed. Please refer to
+- a recent Rust compiler and build environment installed. Please refer to
   [https://rustup.rs/](https://rustup.rs/) on how to get this environment installed.
+- `x86_64-unknown-none` target toolchain installed (`rustup target add x86_64-unknown-none`)
 - `binutils` >= 2.39
```

00xc:
> LGTM!
> 
> I'd just suggest the following changes in INSTALL.md:
> 
> ```diff
> diff --git a/INSTALL.md b/INSTALL.md
> index 8301fca..af72adf 100644
> --- a/INSTALL.md
> +++ b/INSTALL.md
> @@ -171,8 +171,9 @@ Building the COCONUT-SVSM
>  -------------------------
>  
>  Building the SVSM itself requires:
> -- a recent Rust-nightly compiler and build environment installed. Please refer to
> +- a recent Rust compiler and build environment installed. Please refer to
>    [https://rustup.rs/](https://rustup.rs/) on how to get this environment installed.
> +- `x86_64-unknown-none` target toolchain installed (`rustup target add x86_64-unknown-none`)
>  - `binutils` >= 2.39
> ```

Done, I merged this with the other change we had in `INSTALL.md`.

stefano-garzarella (approved):

osteffenrh:
That's awesome! Thanks!

PR #83, opened 2023-09-01
scripts/docker: fix reuse option

stefano-garzarella:
Using option "-r" the script uses a pre-created container. If the container was already created but it is stopped, the script is failing whit the following error:

  the container name "coconut-build" is already in use by ...

Let's add "-a" option to "docker ps" to list all containers, not only the running ones.
Since we are editing this line, let's also use the CONTAINER_NAME variable for the container name.


cclaudio (approved):
Looks good to me, thanks!

PR #85, opened 2023-09-07
Convert address space constants to VirtAddr

joergroedel:
This PR enhances VirtAddr to take care that the addresses are canonical and converts all address space layout constants to type VirtAddr.


src/svsm.rs
```diff
@@ -309,9 +309,9 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
 }
 
 #[no_mangle]
-pub extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: VirtAddr) {
+pub extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: usize) {
     let launch_info: KernelLaunchInfo = *li;
-    let vb_ptr = vb_addr.as_mut_ptr::<u64>();
+    let vb_ptr = VirtAddr::new(vb_addr).as_mut_ptr::<u64>();
```
00xc, line 314: Hm. `VirtAddr` is `repr(transparent)` so that it works with external ABIs (i.e. it is a zero-cost abstraction), but making it non-trivial defeats this purpose. I'd rather have a `sign_extend()` method on the type (or on the `Address` trait) and use that conversion here. Another idea is to have a sign-extending constructor and a "trivial" constructor.

joergroedel: The Address trait is used for PhysAddr too, which does not need sign_extend(). Further, sign_extend is important for add/sub operations on VirtAddr. If I understand correctly, repr(transparent) only affects the memory layout and ABI conventions of the generated code, not the behavior of certain operations, no?

00xc: > The Address trait is used for PhysAddr too, which does not need sign_extend(). Further, sign_extend is important for add/sub operations on VirtAddr.

Then we could have it as a method on `VirtAddr`.

> If I understand correctly, repr(transparent) only affects the memory layout and ABI conventions of the generated code, not the behavior of certain operations, no?

Yes, but my point is that since it cannot contain arbitrary inner values we can no longer use it in a public ABI, like in this case. Or am I missing something?

joergroedel: > Yes, but my point is that since it cannot contain arbitrary inner values we can no longer use it in a public ABI, like in this case. Or am I missing something?

You mean use a function taking VirtAddr from C or assembly code? No, that would be problematic from here on. But I think that does not matter much beyond the startup code, no?

00xc: Yes. I was also thinking about the future syscall interface for loadable CPL3 modules, where we might want to take `VirtAddr`s explicitly. We can always convert to a `VirtAddr` explicitly like in this commit, but it adds some boilerplate.

joergroedel: VirtAddr will not be used in the syscall interface. For that we will rely on repr(C) structs with built-in data types.

The future of VirtAddr is a bit more complicated, as we likely need several of them for different parts of the address space to get the compiler check if a pointer is valid in the given context.

00xc (approved):

PR #86, opened 2023-09-07
SVSM/scripts: Remove unstable rustfmt parameters from pre-commit

joergroedel:
After the switch to stable rust the rustfmt checks triggered by the pre-commit script fail because rustfmt is invoked with parameters not available in stable rust. Remove these parameters to make the checks succeed again.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -25,7 +25,7 @@ RET=0
 for file in `git diff --name-only --staged`; do
     ext=${file##*.}
     if [ "$ext" == "rs" ]; then
-        rustfmt --unstable-features --check --edition 2021 --skip-children $file > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


stefano-garzarella (approved):

00xc:
> Not for this PR, but what about using the same of GH workflow that should be `cargo fmt --all -- --check`?
> 
> Anyway this change LGTM!

I think the idea behind not doing this is so that the script would only check staged files. In other words, it's done that way so that format errors in other places do not affect the would-be commit.

stefano-garzarella:
> > Not for this PR, but what about using the same of GH workflow that should be `cargo fmt --all -- --check`?
> > Anyway this change LGTM!
> 
> I think the idea behind not doing this is so that the script would only check staged files. In other words, it's done that way so that format errors in other places do not affect the would-be commit.

Okay, makes sense to me, thanks!

Issue #74, opened 2023-08-22
Document COCONUT-SVSM


Zildj1an:
We should make an affirmative effort to document Coconut-SVSM internal components. I think we should take advantage of Rustdoc. List of the components to document:

- [x] acpi
- [ ] cpu
- [ ] debug
- [x] elf
- [x] fs
- [x] locking
- [ ] mm
  * Allocation done
  * Memory regions done 
- [ ] protocols
- [ ] sev
- [ ] task
- [ ] utils
- [x] documentation guidelines

We also want to document unsafe blocks, see issue #228.

In addition, doctests can provide helpful usage examples, especially for interfaces where we expect several users (e.g. ELF). Including such code examples would be beneficial once basic documentation is in place.

PR #89, opened 2023-09-08
acpi: Document ACPI tables and ELF files handling

Zildj1an:
Document ACPI tables structs and code, including function usage examples. This is related to issue #74 titled "Document COCONUT-SVSM". Also document ELF files handling


cclaudio:
When applicable, I think the two things below can be useful for people that need to work on this file:
- a comment at the top describing what spec we used in the file (e.g. Name, version)
- a comment in the structure indicating the table number in the spec it refers to

Zildj1an:
Regarding @cclaudio comment, what do you think @joergroedel ? Should we link to any specific ACPI documentation?

Zildj1an:
Thanks @joergroedel in that case I'm satisfied with the current state of the commit. I fixed a merge issue and that's the reason why we have a merge commit without my signature.

Zildj1an:
I just updated this pull request, now we also document ELF files handling (see [new commit](https://github.com/coconut-svsm/svsm/pull/89/commits/5c2d2b9b5bbc07d50a692533342a6e7b31983dfc)) I will not continue documenting until this is merged to avoid bottleneck

00xc:
> Thanks @joergroedel in that case I'm satisfied with the current state of the commit. I fixed a merge issue and that's the reason why we have a merge commit without my signature.

I think this could be fixed by rebasing instead, no? We don't have guidelines on merges vs rebases for the project so this is fine for me, just mentioning it.

src/elf/mod.rs
```diff
@@ -250,18 +367,39 @@ impl convert::TryFrom<(Elf64Off, Elf64Xword)> for Elf64FileRange {
...
+    ///
+    /// Returns an `ElfError` if there are issues parsing the ELF file.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use your_module::Elf64File;
+    ///
+    /// let elf_file = Elf64File::read(&elf_file_buf).unwrap();
+    /// ``
```
00xc, line 402: There's a missing closing backtick here. `cargo doc` will warn about it:

```
warning: could not parse code block as Rust code
   --> src/elf/mod.rs:398:9
    |
398 |       /// ```rust
    |  _________^
399 | |     /// use your_module::Elf64File;
400 | |     ///
401 | |     /// let elf_file = Elf64File::read(&elf_file_buf).unwrap();
402 | |     /// ``
    | |__________^
    |
    = note: error from rustc: unknown start of token: `
    = note: `#[warn(rustdoc::invalid_rust_codeblocks)]` on by default
```

00xc:
The doctests for this PR are broken at the moment. Moreover, existing doctests in the main branch are also broken, so I suggest we fix those before merging these. I've opened #93 for this purpose.

You can run the doctests via `cargo test --target=x86_64-unknown-linux-gnu --doc`.

Zildj1an:
Thanks @00xc I will reopen this PR without the issues mentioned.

Issue #90, opened 2023-09-10
regression with bitflags 2.4


crobinso:
Trying to boot coconut-svsm git main (3c23e268aefc) throws a KVM error:

```
+ sudo /usr/bin/qemu-kvm -enable-kvm -cpu EPYC-v4 -machine q35,confidential-guest-support=sev0,memory-backend=ram1,kvm-type=protected -object memory-backend-memfd-private,id=ram1,size=8G,share=true -object sev-snp-guest,id=sev0,cbitpos=51,reduced-phys-bits=1,svsm=on -smp 8 -no-reboot -drive if=pflash,format=raw,unit=0,file=/usr/share/edk2/ovmf-coconutsvsm/OVMF_CODE.coconutsvsm.fd,readonly=on -drive if=pflash,format=raw,unit=1,file=/usr/share/edk2/ovmf-coconutsvsm/OVMF_VARS.coconutsvsm.fd,snapshot=on -drive if=pflash,format=raw,unit=2,file=./src/coconut-svsm/svsm.bin,readonly=on -netdev user,id=vmnic -device e1000,netdev=vmnic,romfile= -drive file=crobinso-snp.qcow2,if=none,id=disk0,snapshot=off -device virtio-scsi-pci,id=scsi0,disable-legacy=on,iommu_platform=on -device scsi-hd,drive=disk0,bootindex=0 -nographic
[sudo] password for crobinso: 
qemu-kvm: warning: Setting vm_type: protected
qemu-kvm: warning: Setting vm_type: protected
qemu-kvm: warning: UPM mode enabled, discard=(null)
qemu-kvm: warning: creating ROM device with private memory.
qemu-kvm: warning: creating ROM device with private memory.
qemu-kvm: warning: creating ROM device with private memory.
Flash mapping at 0xffc84000 size: 0x37c000
Flash mapping at 0xffb3f000 size: 0xc1000
[Stage2] COCONUT Secure Virtual Machine Service Module (SVSM) Stage 2 Loader
[Stage2] Order-00: total pages:    12 free pages:     0
[Stage2] Order-01: total pages:     1 free pages:     0
[Stage2] Order-02: total pages:     1 free pages:     1
[Stage2] Order-03: total pages:     1 free pages:     1
[Stage2] Order-04: total pages:     1 free pages:     1
[Stage2] Order-05: total pages:     1 free pages:     1
[Stage2] Total memory: 296KiB free memory: 240KiB
[...]

PR #92, opened 2023-09-11
SVSM: Increase boot-stack size

joergroedel:
The 8 KiB of boot stack are not enough anymore after the switch to stable rust. Especially in debug builds the stack usage is higher. The maximum value I measured so far was 9608 bytes, which leads to early boot crashes. Increase the boot stack size to 16 KiB to be on the safe side.

Fixes: Github issue #90

### DIFF
--- a/src/svsm.rs
+++ b/src/svsm.rs
@@ -87,7 +87,7 @@ global_asm!(
 
         .align 4096
     bsp_stack:
-        .fill 8192, 1, 0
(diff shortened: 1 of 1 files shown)


stefano-garzarella (approved):
Tested successfully here!

crobinso (approved):
Fixes things for me. Thanks!

joergroedel:
Thanks for testing. Merged this now.

PR #94, opened 2023-09-13
tests: fix doctests and run in Makefile target

### LINKED ISSUE #93: Fix doctests
Doctests are a Rust feature where code tests can be inlined into the documentation itself. That way, documentation examples are also ran as integration tests. For example:

```rust
/// Sums two integers.
///
/// ```rust
/// let a = 1;
/// let b = 2;
/// assert_eq!(my_sum(a, b), 3);
/// ```
fn my_sum(a: u32, b: u32) -> u32 {
    ...
}
```

These tests can be ran via `cargo test --target=x86_64-unknown-linux-gnu --doc`, which at the moment we do NOT do via our Makefile `test` target. Using `--tests` instead of `--doc` will also work.

Currently, we have some doctests for `ImmutAfterInitCell` and `ImmutAfterInitRef`, which are currently failing. Moreover, more doctests will be added while implementing the tasks for #74. Thus, we must:

* Fix `ImmutAfterInit*` doctests.
* Run doctests via `make test`

Check the doctest documentation here:
https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

00xc:
Our custom bare-metal allocator (`SvsmAllocator`) does not work during tests, as they run as regular userspace binaries. Thus, we must not set it as the global Rust allocator. To do this, we need a compile-time directive that lets us know we are building doctests and not the regular binary (`cfg(doctests)`), but sadly it is not properly set during test compilation (see rust-lang/rust#45599, rust-lang/rust#67295 and coconut-svsm/svsm#93).

In order to bypass this, set the compile time flag ourselves via the `RUSTFLAGS` environment variable so that tests work. Also add the new invocation to the `test` target in the Makefile.

Closes #93

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ all: svsm.bin
 
 test:
 	cargo test --target=x86_64-unknown-linux-gnu
+	RUSTFLAGS="--cfg doctest" cargo test --target=x86_64-unknown-linux-gnu --doc
 
 utils/gen_meta: utils/gen_meta.c
 	cc -O3 -Wall -o $@ $<
--- a/src/mm/alloc.rs
+++ b/src/mm/alloc.rs
@@ -1282,7 +1282,7 @@ unsafe impl GlobalAlloc for SvsmAllocator {
     }
 }
 
-#[cfg_attr(not(test), global_allocator)]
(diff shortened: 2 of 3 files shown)


roy-hopkins (approved):
LGTM

PR #97, opened 2023-09-15
Implement Virtual Memory Manager

joergroedel:
Introduce data structures for virtual memory management and use it for the PerCpu mappings in the SVSM.

TODO: Documentation is currently missing.


src/mm/vm/mapping/phys_mem.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn map(&self, offset: usize) -> Option<PhysAddr> {
+        if offset < self.size {
+            Some((self.base + offset).page_align())
+        } else {
+            None
+        }
+    }
+
+    fn pt_flags(&self) -> PTEntryFlags {
+        PTEntryFlags::NX
```
roy-hopkins, line 43: We're going to need to support executable memory when loading modules.

src/mm/vm/range.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        for part in parts.iter() {
+            pgtbl.populate_pgtbl_part(part);
+        }
+    }
+
+    pub fn initialize(&mut self) -> Result<(), SvsmError> {
+        let start = VirtAddr::from(self.start_pfn << PAGE_SHIFT);
+        let end = VirtAddr::from(self.end_pfn << PAGE_SHIFT);
+        assert!(start.is_aligned(VMR_GRANULE) && end.is_aligned(VMR_GRANULE));
```
roy-hopkins, line 79: Probably ought to also check for `end >= (start + VMR_GRANULE)`. `alloc_page_tables()` arithmetic operations can overflow if end < start.

roy-hopkins (commented):
Looking good.

src/mm/vm/mapping/kernel_stack.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[derive(Default, Debug)]
+pub struct VMKernelStack {
+    alloc: RawAllocMapping,
+    guard_pages: usize,
+}
+
+impl VMKernelStack {
+    const fn guard_pages() -> usize {
+        STACK_PAGES.next_power_of_two() / 2
```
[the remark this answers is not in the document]
joergroedel: No particular reason, just to make the guard-size the same as the stack-size and to keep the sizes power-of-two aligned for easier math.

src/mm/vm/mapping/kernel_stack.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    pub fn top_of_stack(base: VirtAddr) -> VirtAddr {
+        let pages = Self::guard_pages() + STACK_PAGES;
+        base + (pages * PAGE_SIZE)
+    }
+
+    pub fn new() -> Result<Self, SvsmError> {
+        let mut stack = VMKernelStack {
+            alloc: RawAllocMapping::new(STACK_SIZE),
```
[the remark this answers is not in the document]
joergroedel: Makes sense. I added a new_size() constructor which takes size as parameter.

src/mm/vm/mapping/kernel_stack.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    pub fn alloc_pages(&mut self) -> Result<(), SvsmError> {
+        self.alloc.alloc_pages()
+    }
+}
+
+impl VirtualMapping for VMKernelStack {
+    fn mapping_size(&self) -> usize {
+        (2 * STACK_PAGES.next_power_of_two()) << PAGE_SHIFT
```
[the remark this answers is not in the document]
joergroedel: Well, yes, but since self.guard_pages is (STACK_PAGES / 2) it should not matter much. I re-worked this function in favor of the variable stack size now.

src/mm/vm/mapping/phys_mem.rs
joergroedel: Yes, but not when mapping physical addresses directly. For mapping modules there will be a VMFileMapping type or something, which supports mapping from RamFS files and support executable, r/w and ro offset mappings.

src/mm/vm/range.rs
```diff
@@ -0,0 +1,257 @@
...
+pub struct VMR {
+    start_pfn: usize,
+    end_pfn: usize,
+    tree: RWLock<RBTree<VMMAdapter>>,
+    pgtbl_parts: RWLock<Vec<PageTablePart>>,
+    pt_flags: PTEntryFlags,
+}
+
+impl VMR {
+    pub fn new(start: VirtAddr, end: VirtAddr, flags: PTEntryFlags) -> Self {
```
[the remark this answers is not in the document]
joergroedel: Yes, it will be PTEntryFlags::empty() or a combination of PTEntryFlags::GLOBAL and PTEntryFlags::USER. Those two flags are set at the VMR level.

src/mm/vm/range.rs
joergroedel: Yes, that makes sense. Added a check for start < end to the assert.

src/mm/vm/range.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    fn unmap_vmm(&self, vmm: &VMM) {
+        let (rstart, _) = self.virt_range();
+        let (vmm_start, vmm_end) = vmm.range();
+        let mut pgtbl_parts = self.pgtbl_parts.lock_write();
+        let mut offset: usize = 0;
+
+        while vmm_start + offset < vmm_end {
+            let idx = PageTable::index::<3>(VirtAddr::from(vmm_start - rstart));
+            if pgtbl_parts[idx].unmap_4k(vmm_start + offset).is_some() {
```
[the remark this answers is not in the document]
joergroedel: Yes, that is missing, thanks for pointing that out.

src/mm/pagetable.rs
```diff
@@ -745,3 +759,218 @@ impl DerefMut for PageTableRef {
...
+impl RawPageTablePart {
+    fn entry_to_page(entry: PTEntry) -> Option<&'static mut PTPage> {
+        let flags = entry.flags();
+        if !flags.contains(PTEntryFlags::PRESENT) || flags.contains(PTEntryFlags::HUGE) {
+            return None;
+        }
+
+        let address = phys_to_virt(entry.address());
+        Some(unsafe { &mut *address.as_mut_ptr::<PTPage>() })
+    }
```
Freax13, line 778: The unsafe at the bottom seems dubious. How is it guaranteed that now other references to the page exist?

src/mm/vm/mapping/api.rs
```diff
@@ -0,0 +1,145 @@
...
+    // Use a RefCell here to check borrowing rules at runtime. This should
+    // yield no errors as link and range are protected by the VMR RWLock.
+    range: RefCell<Range>,
+
+    pt_flags: PTEntryFlags,
+    mapping: RWLock<Rc<dyn VirtualMapping>>,
+}
+
+unsafe impl Send for VMM {}
+unsafe impl Sync for VMM {}
```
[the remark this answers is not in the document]
joergroedel: Yes, you are right. But this is a bit tricky. All of this works only because VMM is only ever instantiated as a node in a VMR rbtree, and as such protected by the VMR rbtree lock. That is also the reason for the RefCell, as the VMR code needs to access it all the time (and these values are protected by the VMR lock) and taking a per-VMM lock every time is not feasible.

This all certainly needs clearer annotations in the code. I have to check again why the Send and Sync annotations were needed.

src/mm/pagetable.rs
joergroedel: We only do page-table page-sharing at the top-level of the page-table. This is implemented via PageTablePart. Any sub-page allocated in the PT radix tree is only ever referenced by the same PageTablePart it was allocated for.

This allows to treat a PageTablePart as a whole and reference-count it. Once the last reference to a PageTablePart is gone, the whole sub-tree can be safely freed.

I know that the page-table code still lacks proper lifetime annotations, that is also on the list of things to fix.

src/mm/vm/mapping/api.rs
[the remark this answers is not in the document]
Freax13: > Yes, you are right. But this is a bit tricky. All of this works only because VMM is only ever instantiated as a node in a VMR rbtree, and as such protected by the VMR rbtree lock. That is also the reason for the RefCell, as the VMR code needs to access it all the time (and these values are protected by the VMR lock) and taking a per-VMM lock every time is not feasible.

If the VMM is already protected by another lock, swapping a RefCell for a spin lock is a no-op in terms of performance, because the lock won't be contended anyways, right? Am I missing something?
Similarly, is there are reason that we can't replace the use of `Rc` with `Arc`?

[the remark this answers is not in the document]
joergroedel: A spin lock is not free, even if it is always free to take. It requires at least two atomic memory writes. I will have a look again on why it was needed to mark VMM Sync and Send.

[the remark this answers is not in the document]
Freax13: > A spin lock is not free, even if it is always free to take. It requires at least two atomic memory writes.

Yes, but so does a `RefCell` (the writes aren't atomic in a `RefCell`, but on x86 atomic writes are just normal writes).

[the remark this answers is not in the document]
joergroedel: > Yes, but so does a `RefCell` (the writes aren't atomic in a `RefCell`, but on x86 atomic writes are just normal writes).

Turned out that with VMMs handled entirely within VMRange the RefCell is also not needed anymore, so I removed it too. During development I planned to actually hand out references of VMMs to callers, but that showed not to be a good idea.

Makefile
```diff
@@ -18,7 +18,7 @@ all: svsm.bin
 
 test:
 	cargo test --target=x86_64-unknown-linux-gnu
-	RUSTFLAGS="--cfg doctest" cargo test --target=x86_64-unknown-linux-gnu --doc
+#	RUSTFLAGS="--cfg doctest" cargo test --target=x86_64-unknown-linux-gnu --doc
```
roy-hopkins, line 21: This can be re-enabled now #119 has been merged.

joergroedel: Yes, I already dropped the top-commit of that branch :)

src/mm/vm/mapping/kernel_stack.rs
```diff
@@ -0,0 +1,110 @@
...
+    ///
+    /// Initialized stack on success, Err(SvsmError::Mem) on error
+    pub fn new() -> Result<Self, SvsmError> {
+        VMKernelStack::new_size(STACK_SIZE)
+    }
+
+    fn alloc_pages(&mut self) -> Result<(), SvsmError> {
+        self.alloc.alloc_pages()
+    }
+}
```
roy-hopkins, line 81: I'm just looking at rebasing the task code against this and I think we're going to need a way to populate content within the stack pages using a temporary virtual address before the stack is mapped into a tasks's virtual memory space. 

When a task is first created, a `VMKernelStack` will be created for the task kernel stack. This will initially contain a task frame which includes the entry point of the task as expected by `switch_context`. So, `create_task()` will need to access the new task's stack from the context of another task in order to create this frame.

src/mm/vm/mapping/api.rs
```diff
@@ -0,0 +1,218 @@
...
+struct Range {
+    start: usize,
+    end: usize,
+}
+
+impl Range {
+    pub const fn new(start: usize, end: usize) -> Self {
+        Range { start, end }
+    }
+}
```
Freax13, line 91: Have you considered using the [`Range`](https://doc.rust-lang.org/core/ops/struct.Range.html) type in the standard library instead?

src/mm/vm/mapping/api.rs
```diff
@@ -0,0 +1,218 @@
...
+        node.range.start
+    }
+}
+
+impl VMM {
+    /// Create a new VMM instance with at a given address and backing struct
+    ///
+    /// # Arguments
+    ///
+    /// * `start` - Virtual start address to store in the mapping
```
Freax13, line 128: `start` is a virtual page number, not an address.

joergroedel: Thanks, this is fixed now.

src/mm/vm/mapping/kernel_stack.rs
joergroedel: Right, I need to think about the best way to do that. The options that come to mind are:

1. Add the mapping to our own address space first and populate the stack. Then remove it, take the returned `Box<VMM>` and add it to the target address space.
2. Just make use of the fact that VMMs store the underlying mapping in an Rc and create multiple mappings from the same backend. We can deploy it in the creating tasks address space for populating the stack and in the target address space for usage.

Thinking more about it, option 2 is probably the best one.

joergroedel: The only issue is that `RWLock<>` protecting the backend mapping from being modified. This is per `VMM` and with option 2 we would have two `VMM`s pointing to the same backend mapping. This is unsafe, the locking needs to be part of the backend mapping (or another wrapper data type).

joergroedel: Okay, I pushed a new version which solves these problems. There is a new `Mapping` type now which handles the locking and the `RBTree` now stores `Arc` instead of `Box`. I needed to add a `'static` annotation to `Mapping` which I am not sure is correct, please have a look at that.

The `Arc<Mapping>` can be used to add the same mapping into multiple VMRanges so that stacks can be pre-populated.

src/cpu/percpu.rs
```diff
@@ -183,14 +184,23 @@ pub struct PerCpu {
     guest_vmsa: SpinLock<GuestVmsaRef>,
     reset_ip: u64,
 
+    /// PerCpu Virtual Memory Range
+    vm_range: VMR,
```
roy-hopkins, line 188: This is private but the `Tasks` structure needs access to this in order to update the task pagetable when performing a task switch and also to add a range when configuring a new task stack.

src/mm/vm/mapping/kernel_stack.rs
joergroedel: Okay, thanks for testing this out. The accessibility problem of the `PerCpu.vm_range` can be solved in the task PRs. Probably the best way is to provide a public API like `percpu_create_mapping()` or something like that. Since it is a temporary mapping even a guard-struct might be useful.

PR #98, opened 2023-09-18
types: explicitly type page size parameters

00xc:
* Explicitly type the page size parameters for `pvalidate()`, `GHCB::page_state_change()` and related functions.

* Explicitly type the state parameter for `pvalidate()`.

This way, the following:

```rust
pvalidate(vaddr, false, true)?;
```

Turns into:

```rust
pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid)?;
```

And this:

```rust
    this_cpu_mut()
        .ghcb()
        .page_state_change(pstart, pend, false, PageStateChangeOp::PscPrivate)
```

turns into:

```rust
    this_cpu_mut()
        .ghcb()
        .page_state_change(
            pstart,
            pend,
            PageSize::Regular,
            PageStateChangeOp::PscPrivate,
        )
```

### DIFF
--- a/src/sev/utils.rs
+++ b/src/sev/utils.rs
@@ -50,7 +50,7 @@ impl fmt::Display for SevSnpError {
     }
 }
 
-fn pvalidate_range_4k(start: VirtAddr, end: VirtAddr, valid: bool) -> Result<(), SvsmError> {
+fn pvalidate_range_4k(start: VirtAddr, end: VirtAddr, valid: PvalidateOp) -> Result<(), SvsmError> {
     for addr in (start.bits()..end.bits())
         .step_by(PAGE_SIZE)
         .map(VirtAddr::from)
@@ -61,7 +61,11 @@ fn pvalidate_range_4k(start: VirtAddr, end: VirtAddr, valid: bool) -> Result<(),
     Ok(())
 }
 
-pub fn pvalidate_range(start: VirtAddr, end: VirtAddr, valid: bool) -> Result<(), SvsmError> {
+pub fn pvalidate_range(
+    start: VirtAddr,
+    end: VirtAddr,
+    valid: PvalidateOp,
+) -> Result<(), SvsmError> {
     let mut addr = start;
 
     while addr < end {
@@ -84,7 +88,15 @@ pub fn pvalidate_range(start: VirtAddr, end: VirtAddr, valid: bool) -> Result<()
     Ok(())
 }
 
-pub fn pvalidate(vaddr: VirtAddr, size: PageSize, valid: bool) -> Result<(), SvsmError> {
+/// The desired state of the page passed to PVALIDATE.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(u64)]
+pub enum PvalidateOp {
+    Invalid = 0,
+    Valid = 1,
+}
+
+pub fn pvalidate(vaddr: VirtAddr, size: PageSize, valid: PvalidateOp) -> Result<(), SvsmError> {
     let rax = vaddr.bits();
     let rcx: u64 = match size {
         PageSize::Regular => 0,
--- a/src/sev/utils.rs
+++ b/src/sev/utils.rs
@@ -6,7 +6,7 @@
 
 use crate::address::{Address, VirtAddr};
 use crate::error::SvsmError;
-use crate::types::{GUEST_VMPL, PAGE_SIZE, PAGE_SIZE_2M};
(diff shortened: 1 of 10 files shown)


cclaudio (approved):
This is helpful, thanks for the PR. 
It looks good to me.

src/sev/utils.rs
```diff
@@ -68,25 +68,35 @@ pub fn pvalidate_range(start: VirtAddr, end: VirtAddr, valid: bool) -> Result<()
...
     Ok(())
 }
 
-pub fn pvalidate(vaddr: VirtAddr, huge_page: bool, valid: bool) -> Result<(), SvsmError> {
+/// The size of the page passed to PVALIDATE.
+#[derive(Clone, Copy, Debug)]
+#[repr(u64)]
+pub enum PvalidateSize {
+    /// Regular page (4K)
+    Page = 0,
```
stefano-garzarella, line 96: I find a bit confusing the name, what about `Regular` and `Huge` ?
Or `RegularPage` and `HugePage`?

src/sev/utils.rs
```diff
@@ -68,25 +68,35 @@ pub fn pvalidate_range(start: VirtAddr, end: VirtAddr, valid: bool) -> Result<()
...
+    // Huge page (2M)
```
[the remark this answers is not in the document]
00xc: Oops, good catch, I'll update the PR.

src/sev/utils.rs
```diff
@@ -68,25 +68,35 @@ pub fn pvalidate_range(start: VirtAddr, end: VirtAddr, valid: bool) -> Result<()
...
+            pvalidate(addr, PvalidateSize::Page, valid)?;
             addr = addr + PAGE_SIZE;
         }
     }
 
     Ok(())
 }
 
-pub fn pvalidate(vaddr: VirtAddr, huge_page: bool, valid: bool) -> Result<(), SvsmError> {
+/// The size of the page passed to PVALIDATE.
```
stefano-garzarella, line 91: Should we mention in the documentation that the enum values are the one used for the PVALIDATE instructions?

Maybe it's implied and we can avoid it, it just occurred to me while looking at the code because we didn't add a match to choose the value.

00xc: Not sure I understand. Do you mean the `repr` values? I don't see any reason for a caller to use them.

src/sev/utils.rs
00xc: Updated to use `RegularPage` and `HugePage`.

src/sev/utils.rs
00xc: I've removed this so that the enum is explicitly converted to the expected register value for the `PVALIDATE` instruction within the appropriate functions.

00xc:
I've updated this PR so that there is a single `PageSize` enum that can be used in multiple places (`pvalidate()` and `GHCB::page_state_change()` are the main candidates for now).

`PvalidateOp` remains unchanged.

src/types.rs
```diff
@@ -11,6 +11,21 @@ pub const PAGE_SHIFT_2M: usize = 21;
 pub const PAGE_SIZE: usize = 1 << PAGE_SHIFT;
 pub const PAGE_SIZE_2M: usize = PAGE_SIZE * 512;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PageSize {
+    Regular,
+    Huge,
+}
+
+impl From<PageSize> for usize {
```
stefano-garzarella, line 20: What is the advantage of implementing `From` for `usize` compared to embed the values in the `enum`?

I mean something like this:
```
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
#[repr(usize)]
pub enum PageSize {
    Regular = PAGE_SIZE,
    Huge = PAGE_SIZE_2M,
}
```

I'm asking because I'm worried about starting to use `as usize` wrongly with the `PageSize` enum.

00xc: Not sure if I understand. By implementing `From<T>` you do not enable `as T`, so right now as it is you cannot do that worrying conversion. This would only be possible with `repr(usize)`.

In general, not tying the enum variants with specific values allows the compiler to optimize the discriminant as it sees fit. In this case I don't think it would make much a difference, but the conversion seems more explicit this way:

```rust
let page_size = PageSize::Huge;
let size_bytes = usize::from(page_size);
// or also:
let size_bytes: usize = page_size.into();
```

It also makes it harder to accidentally change the conversion values down the line, as you mentioned when we were discussing `PvalidateSize`.

stefano-garzarella: @00xc thanks for the details! It's clearer to me now, and I agree it's better that way!

PR #99, opened 2023-09-18
elf: Add unit tests

Zildj1an:
Add unit tests to test correct behavior of ELF handling.

### DIFF
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -17,7 +17,7 @@ use core::fmt;
 use core::matches;
 use core::mem;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum ElfError {
     FileTooShort,
 
@@ -250,7 +250,7 @@ impl convert::TryFrom<(Elf64Off, Elf64Xword)> for Elf64FileRange {
     }
 }
 
-#[derive(Default, Debug)]
+#[derive(Default, Debug, PartialEq)]
 pub struct Elf64File<'a> {
     elf_file_buf: &'a [u8],
     elf_hdr: Elf64Hdr,
@@ -647,7 +647,7 @@ impl<'a> Elf64File<'a> {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
 pub struct Elf64Hdr {
     #[allow(unused)]
     e_ident: [Elf64char; 16],
@@ -958,7 +958,7 @@ impl Elf64Shdr {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
 struct Elf64LoadSegments {
     segments: Vec<(Elf64AddrRange, Elf64Half)>,
 }
@@ -1026,7 +1026,7 @@ impl Elf64LoadSegments {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct Elf64DynamicRelocTable {
     base_vaddr: Elf64Addr, // DT_RELA / DT_REL
     size: Elf64Xword,      // DT_RELASZ / DT_RELSZ
@@ -1044,7 +1044,7 @@ impl Elf64DynamicRelocTable {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct Elf64DynamicSymtab {
     base_vaddr: Elf64Addr, // DT_SYMTAB
     entsize: Elf64Xword,   // DT_SYMENT
@@ -1058,7 +1058,7 @@ impl Elf64DynamicSymtab {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct Elf64Dynamic {
     // No DT_REL representation: "The AMD64 ABI architectures uses only
     // Elf64_Rela relocation entries [...]".
@@ -1241,7 +1241,7 @@ impl<'a> Iterator for Elf64ImageLoadSegmentIterator<'a> {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
 struct Elf64Strtab<'a> {
     strtab_buf: &'a [u8],
 }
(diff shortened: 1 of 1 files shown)


src/elf/mod.rs
```diff
@@ -1605,3 +1605,198 @@ impl<'a, RP: Elf64RelocProcessor> Iterator for Elf64AppliedRelaIterator<'a, RP>
...
+
+        // Use the Elf64File::read method to create an Elf64File instance
+        match Elf64File::read(&byte_data) {
+            Ok(_elf_file) => {
+                assert!(false);
+            }
+            Err(err) => {
+                assert!(matches!(err, crate::elf::ElfError::InvalidPhdrSize));
+            }
+        }
```
00xc, line 1742: This can probably be collapsed to a single statement (instead of having an `assert!(false)`) to something like:

```rust
let res = Elf64File::read(&byte_data);
assert!(matches!(res, Err(crate::elf::ElfError::InvalidPhdrSize)))
```

You could also use `assert_eq!()` instead of `assert!(matches!())` if you derive `Eq` for `ElfError`.

Zildj1an: Thank you for the review, @00xc. I've updated my commit to simplify the logic by consolidating it into a single assert. To use `assert_eq!()`, we would need to implement `PartialEq` for `ElfError`. But, this would also require implementing `PartialEq` in several other related structures, e.g. `Elf64File`, `Elf64Strtab`, `Elf64Dynamic`, and more. I've kept the changes focused on the PR's scope.

src/elf/mod.rs
```diff
@@ -1605,3 +1605,192 @@ impl<'a, RP: Elf64RelocProcessor> Iterator for Elf64AppliedRelaIterator<'a, RP>
...
+            0x00, 0x00, 0x00, 0xff, 0x15, 0x53, 0x2f, 0x00, 0x00, 0xf4, 0x66, 0x2e, 0x0f, 0x1f,
+            0x84, 0x00, 0x00, 0x00, 0x00, 0x48, 0x8d, 0x3d, 0x79, 0x2f, 0x00, 0x00, 0x48, 0x8d,
+            0x05, 0x72, 0x2f, 0x00, 0x00, 0x48, 0x39, 0xf8, 0x74, 0x15, 0x48, 0x8b, 0x05, 0x36,
+            0x2f, 0x00, 0x00, 0x48, 0x85, 0xc0, 0x74, 0x09, 0xff, 0xe0, 0x0f, 0x1f, 0x80, 0x00,
+            0x00, 0x00, 0x00, 0xc3,
+        ];
+
+        // Use the Elf64File::read method to create an Elf64File instance
+        let res = Elf64File::read(&byte_data);
+        assert!(matches!(res, Err(crate::elf::ElfError::InvalidPhdrSize)));
```
cclaudio, line 1736: If you use `assert_eq` here, the test case would have a better error message.  I was able to do that deriving some structures from PartialEq.

```diff
 diff --git a/src/elf/mod.rs b/src/elf/mod.rs
index f795560f0963..ecec8543c06f 100644
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -17,7 +17,7 @@ use core::fmt;
 use core::matches;
 use core::mem;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum ElfError {
     FileTooShort,
 
@@ -250,7 +250,7 @@ impl convert::TryFrom<(Elf64Off, Elf64Xword)> for Elf64FileRange {
     }
 }
 
-#[derive(Default, Debug)]
+#[derive(Default, Debug, PartialEq)]
 pub struct Elf64File<'a> {
     elf_file_buf: &'a [u8],
     elf_hdr: Elf64Hdr,
@@ -647,7 +647,7 @@ impl<'a> Elf64File<'a> {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
 pub struct Elf64Hdr {
     #[allow(unused)]
     e_ident: [Elf64char; 16],
@@ -958,7 +958,7 @@ impl Elf64Shdr {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
 struct Elf64LoadSegments {
     segments: Vec<(Elf64AddrRange, Elf64Half)>,
 }
@@ -1026,7 +1026,7 @@ impl Elf64LoadSegments {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct Elf64DynamicRelocTable {
     base_vaddr: Elf64Addr, // DT_RELA / DT_REL
     size: Elf64Xword,      // DT_RELASZ / DT_RELSZ
@@ -1044,7 +1044,7 @@ impl Elf64DynamicRelocTable {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct Elf64DynamicSymtab {
     base_vaddr: Elf64Addr, // DT_SYMTAB
     entsize: Elf64Xword,   // DT_SYMENT
@@ -1058,7 +1058,7 @@ impl Elf64DynamicSymtab {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 struct Elf64Dynamic {
     // No DT_REL representation: "The AMD64 ABI architectures uses only
     // Elf64_Rela relocation entries [...]".
@@ -1241,7 +1241,7 @@ impl<'a> Iterator for Elf64ImageLoadSegmentIterator<'a> {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq)]
 struct Elf64Strtab<'a> {
     strtab_buf: &'a [u8],
 }
@@ -1733,10 +1733,13 @@ mod tests {
 
         // Use the Elf64File::read method to create an Elf64File instance
         let res = Elf64File::read(&byte_data);
-        assert!(matches!(res, Err(crate::elf::ElfError::InvalidPhdrSize)));
+        assert_eq!(res.err(), Some(crate::elf::ElfError::InvalidPhdrSize));
```

src/elf/mod.rs
```diff
@@ -1605,3 +1605,192 @@ impl<'a, RP: Elf64RelocProcessor> Iterator for Elf64AppliedRelaIterator<'a, RP>
...
+
+        // Construct an Elf64Hdr instance from the byte data
+        let elf_hdr = Elf64Hdr::read(&byte_data).expect("Failed to read ELF header");
```
cclaudio, line 1739: If we hit this expect, will the remaining tests not be executed? If so, we could have something like this
```diff
         // Construct an Elf64Hdr instance from the byte data
-        let elf_hdr = Elf64Hdr::read(&byte_data).expect("Failed to read ELF header");
+        let elf_hdr = Elf64Hdr::read(&byte_data);
+       assert!(elf_hdr.is_ok());
+
+       let elf_hdr = elf_hdr.unwrap();
```

cclaudio (commented):
I have only a few comments

src/elf/mod.rs
Zildj1an: I told @00xc, that I preferred to focus this PR on pure documentation, but since this seems to be the preferred approach, I will update commit

Zildj1an:
Thanks for your comments @cclaudio , I have updated my commit accordingly.

00xc (approved):

Zildj1an:
I will reopen PR soon. Needed some housekeeping

PR #102, opened 2023-09-25
sev/ghcb: explicitly type size parameter in `GHCB::page_state_change()`

00xc:
Use an enum instead of a boolean to make the page size parameter explicit. This improves readability.

this:

```rust
    this_cpu_mut()
        .ghcb()
        .page_state_change(pstart, pend, false, PageStateChangeOp::PscPrivate)
```

turns into:

```rust
    this_cpu_mut()
        .ghcb()
        .page_state_change(
            pstart,
            pend,
            PageStateChangeSize::RegularPage,
            PageStateChangeOp::PscPrivate,
        )
```

### DIFF
--- a/src/fw_meta.rs
+++ b/src/fw_meta.rs
@@ -11,7 +11,7 @@ use crate::cpu::percpu::this_cpu_mut;
 use crate::error::SvsmError;
 use crate::mm::PerCPUPageMappingGuard;
 use crate::mm::SIZE_1G;
-use crate::sev::ghcb::PageStateChangeOp;
+use crate::sev::ghcb::{PageStateChangeOp, PageStateChangeSize};
 use crate::sev::{pvalidate, rmp_adjust, RMPFlags};
 use crate::types::PAGE_SIZE;
 use crate::utils::{overlap, zero_mem_region};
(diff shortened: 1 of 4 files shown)


stefano-garzarella (approved):
Thanks for these cleanups!

LGTM, just a small comment that maybe we can address in another PR: since we also added an enum in https://github.com/coconut-svsm/svsm/pull/98 to distinguish between RegularPage and HugePage, might it make sense to start defining a generic enum in `crate::types` and use it as much as possible?

Perhaps we will still need the one in Pvalidate since it directly maps the values to be used in the instruction, but for example in this PR we only use it as an enum and may need it somewhere else.

00xc:
> Thanks for these cleanups!
> 
> LGTM, just a small comment that maybe we can address in another PR: since we also added an enum in #98 to distinguish between RegularPage and HugePage, might it make sense to start defining a generic enum in `crate::types` and use it as much as possible?
> 
> Perhaps we will still need the one in Pvalidate since it directly maps the values to be used in the instruction, but for example in this PR we only use it as an enum and may need it somewhere else.

Yeah this make sense. Closing this PR in favor of #98, where I'll use the new `PageSize` enum both for `pvalidate()` and `GHCB::page_state_change()`.

PR #103, opened 2023-09-27
run clippy on tests on the CI and fix several warnings

stefano-garzarella:
Running `cargo clippy --tests --target=x86_64-unknown-linux-gnu --all-features` highlighted several things that we can improve in our tests code.

Only tests are affected, so no change to code behavior.

We need to add a second call to clippy for tests in the CI because we need to use a different target (i.e. x86_64-unknown-linux-gnu) to avoid build failure.

I squashed all the tests changes in a single commit since they are tests, but if you prefer I can split them in several commits.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -48,3 +48,9 @@ jobs:
         with:
           command: clippy
           args: --all-features -- -D warnings
+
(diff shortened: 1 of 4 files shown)


00xc (approved):
LGTM

PR #105, opened 2023-09-28
locking: Add unit tests

Zildj1an:
Add unit tests to test correct behavior of both spinlock and rwlock.

### DIFF
--- a/src/locking/rwlock.rs
+++ b/src/locking/rwlock.rs
@@ -155,3 +155,46 @@ impl<T: Debug> RWLock<T> {
         }
     }
 }
+
+mod tests {
+
+    #[test]
+    fn test_lock_rw() {
+        use crate::locking::*;
+        let rwlock = RWLock::new(42);
+
+        // Acquire a read lock and check the initial value
+        let read_guard = rwlock.lock_read();
+        assert_eq!(*read_guard, 42);
+
+        drop(read_guard);
+
+        let read_guard2 = rwlock.lock_read();
+        assert_eq!(*read_guard2, 42);
+
+        // Create another RWLock instance for modification
+        let rwlock_modify = RWLock::new(0);
+
+        let mut write_guard = rwlock_modify.lock_write();
+        *write_guard = 99;
+        assert_eq!(*write_guard, 99);
+
+        drop(write_guard);
+
+        let read_guard = rwlock.lock_read();
+        assert_eq!(*read_guard, 42);
+
+        // Let's test two concurrent readers on a new RWLock instance
+        let rwlock_concurrent = RWLock::new(123);
+
+        let read_guard1 = rwlock_concurrent.lock_read();
+        let read_guard2 = rwlock_concurrent.lock_read();
+
+        // Assert that both readers can access the same value (123)
+        assert_eq!(*read_guard1, 123);
+        assert_eq!(*read_guard2, 123);
+
+        drop(read_guard1);
+        drop(read_guard2);
+    }
+}
--- a/src/locking/spinlock.rs
+++ b/src/locking/spinlock.rs
@@ -89,3 +89,23 @@ impl<T: Debug> SpinLock<T> {
         None
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_spin_lock() {
+        let spin_lock = SpinLock::new(0);
+
+        let mut guard = spin_lock.lock();
+        *guard += 1;
+
+        // Ensure the locked data is updated.
+        assert_eq!(*guard, 1);
+
+        // Try to lock again; it should fail and return None.
+        let try_lock_result = spin_lock.try_lock();
+        assert!(try_lock_result.is_none());
+    }
+}


src/locking/rwlock.rs
```diff
@@ -155,3 +155,46 @@ impl<T: Debug> RWLock<T> {
         }
     }
 }
+
+mod tests {
+
+    #[test]
+    fn test_lock_rw() {
```
Freax13, line 162: Note that this test fails when run with miri:
```
$ MIRIFLAGS="-Zmiri-permissive-provenance" cargo +nightly miri test --target=x86_64-unknown-linux-gnu test_lock_rw
[...]
test locking::rwlock::tests::test_lock_rw ... error: Undefined Behavior: trying to retag from <137787> for SharedReadOnly permission at alloc48741[0x8], but that tag does not exist in the borrow stack for this location
   --> src/locking/rwlock.rs:27:9
    |
27  |         self.data
    |         ^^^^^^^^^
    |         |
    |         trying to retag from <137787> for SharedReadOnly permission at alloc48741[0x8], but that tag does not exist in the borrow stack for this location
    |         this error occurs as part of retag at alloc48741[0x8..0xc]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <137787> was created by a Unique retag at offsets [0x8..0xc]
   --> src/locking/rwlock.rs:190:27
    |
190 |         let read_guard1 = rwlock_concurrent.lock_read();
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: <137787> was later invalidated at offsets [0x8..0xc] by a Unique retag
   --> src/locking/rwlock.rs:127:28
    |
127 |             data: unsafe { &mut *self.data.get() },
    |                            ^^^^^^^^^^^^^^^^^^^^^
    = note: BACKTRACE (of the first span):
    = note: inside `<locking::rwlock::ReadLockGuard<'_, i32> as core::ops::Deref>::deref` at src/locking/rwlock.rs:27:9: 27:18
note: inside `locking::rwlock::tests::test_lock_rw`
   --> src/locking/rwlock.rs:194:20
    |
194 |         assert_eq!(*read_guard1, 123);
    |                    ^^^^^^^^^^^^
note: inside closure
   --> src/locking/rwlock.rs:163:23
    |
162 |     #[test]
    |     ------- in this procedural macro expansion
163 |     fn test_lock_rw() {
    |                       ^
    = note: this error originates in the attribute macro `test` (in Nightly builds, run with -Z macro-backtrace for more info)

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
```

00xc: > Note that this test fails when run with miri:
> ...

Could you check if this patch fixes it? I believe a `ReadLockGuard` should only have an immutable reference to the underlying `T`:

```diff
diff --git a/src/locking/rwlock.rs b/src/locking/rwlock.rs
index 517862e..ad954fe 100644
--- a/src/locking/rwlock.rs
+++ b/src/locking/rwlock.rs
@@ -12,7 +12,7 @@ use core::sync::atomic::{AtomicU64, Ordering};
 #[derive(Debug)]
 pub struct ReadLockGuard<'a, T: Debug> {
     rwlock: &'a AtomicU64,
-    data: &'a mut T,
+    data: &'a T,
 }
 
 impl<'a, T: Debug> Drop for ReadLockGuard<'a, T> {
@@ -124,7 +124,7 @@ impl<T: Debug> RWLock<T> {
 
         ReadLockGuard {
             rwlock: &self.rwlock,
-            data: unsafe { &mut *self.data.get() },
+            data: unsafe { &*self.data.get() },
         }
     }
```

On this topic, would it make sense to run tests under Miri in our Github Actions?

Freax13: > > Note that this test fails when run with miri:
> > ...
> 
> Could you check if this patch fixes it? I believe a `ReadLockGuard` should only have an immutable reference to the underlying `T`:
> 
> ```diff
> diff --git a/src/locking/rwlock.rs b/src/locking/rwlock.rs
> index 517862e..ad954fe 100644
> --- a/src/locking/rwlock.rs
> +++ b/src/locking/rwlock.rs
> @@ -12,7 +12,7 @@ use core::sync::atomic::{AtomicU64, Ordering};
>  #[derive(Debug)]
>  pub struct ReadLockGuard<'a, T: Debug> {
>      rwlock: &'a AtomicU64,
> -    data: &'a mut T,
> +    data: &'a T,
>  }
>  
>  impl<'a, T: Debug> Drop for ReadLockGuard<'a, T> {
> @@ -124,7 +124,7 @@ impl<T: Debug> RWLock<T> {
>  
>          ReadLockGuard {
>              rwlock: &self.rwlock,
> -            data: unsafe { &mut *self.data.get() },
> +            data: unsafe { &*self.data.get() },
>          }
>      }
> ```
> 

The patch looks good to me. Your reasoning is correct. Multiple mutable references are not allowed.

> On this topic, would it make sense to run tests under Miri in our Github Actions?

Miri is quite resource hungry, peaking at more than 5GB of memory for these tests on my system (I'm not sure if this is normal or if this could be caused by memory tests leaking memory). IIRC, the default runners on githubs CI don't have that much RAM. Besides that, it probably makes sense.

00xc: > The patch looks good to me. Your reasoning is correct. Multiple mutable references are not allowed.

Fix in #106.

> > On this topic, would it make sense to run tests under Miri in our Github Actions?
> 
> Miri is quite resource hungry, peaking at more than 5GB of memory for these tests on my system (I'm not sure if this is normal or if this could be caused by memory tests leaking memory). IIRC, the default runners on githubs CI don't have that much RAM. Besides that, it probably makes sense.

Apparently the runners have 7GB RAM:

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

It will make tests really slow though. I'll keep tabs on this and I'll try to run them in my branch, let's see what happens.

src/locking/spinlock.rs
```diff
@@ -89,3 +89,28 @@ impl<T: Debug> SpinLock<T> {
...
+        let mut guard = spin_lock.lock();
+        *guard += 1;
+
+        // Ensure the locked data is updated.
+        assert_eq!(*guard, 1);
+
+        let spin_lock = SpinLock::new(0);
+
+        let mut guard_failed = spin_lock.lock();
+        *guard_failed += 1;
```
joergroedel, line 110: I don't get what this tests, you are defining a new spinlock, overwriting the old name. Can you explain what this tests and what guard_failed is for?

Zildj1an:
I will reopen PR soon. Needed some housekeeping

PR #109, opened 2023-09-30
miscellaneous fixes

Freax13:


src/cpu/percpu.rs
```diff
@@ -46,27 +45,20 @@ impl PerCpuInfo {
...
-// We use an UnsafeCell to allow for a static with interior
-// mutability. Normally, we would need to guarantee synchronization
-// on the backing datatype, but this is not needed because writes to
-// the structure only occur at initialization, from CPU 0, and reads
-// should only occur after all writes are done.
 #[derive(Debug)]
 pub struct PerCpuAreas {
-    areas: UnsafeCell<Vec<PerCpuInfo>>,
+    areas: RWLock<Vec<PerCpuInfo>>,
 }
```
00xc, line 51: `RWLock` introduces unnecessary overhead in my opinion, so I'd say this change is not needed, as the comment explains. Safety is already documented and limited to private functions.

Freax13: > `RWLock` introduces unnecessary overhead in my opinion

The get and push functions are very rarely called, so IMHO overhead should not be a big concern here.

> so I'd say this change is not needed, as the comment explains. Safety is already documented and limited to private functions.

Is the safety comment correct though? AFAICT the secondary APs are launched at the same time that `PerCpuAreas` is initialized and the BSP even waits for them to become online. Once the APs come online they start serving requests from the guest that can cause read accesses to `PerCpuAreas` while the BSP might still be adding areas. Is there a barrier I'm missing that prevents this?

joergroedel: > Is the safety comment correct though? AFAICT the secondary APs are launched at the same time that `PerCpuAreas` is initialized and the BSP even waits for them to become online. Once the APs come online they start serving requests from the guest that can cause read accesses to `PerCpuAreas` while the BSP might still be adding areas. Is there a barrier I'm missing that prevents this?

There is nothing running at a higher VMPL when the PerCpu areas are added. That happens when everything is set up and all APs have been booted. Then the SVSM will launch the firmware, and from that point on we can get requests on other APs.

joergroedel: Another way to explain this is: When the APs are brought up, the code will not setup guest VMSA or CAA areas. Without those the code in `request_loop()` will directly go into `halt()` once the function is entered. The VMSAs and CAAs for the APs are setup by the guest VMPL and that has not been launched yet.

Freax13: Alright that makes sense, thanks for the explanation! I'll remove this patch.

joergroedel (commented):
Thanks for all the fixes, I have commented on some of them where I have objections or need more explanations. The others are fine, feel free to submit them in a separate PR to reduce your rebasing overhead.

The undisputed changes are:
```
a150f011f203 sev/ghcb: remove `GHCBIOPort`
66a9990907be io: add `Sync` requirement to `IOPort`
87e76e1ab39a console: don't use raw pointer in `Console`
0145cb149d57 serial: add `Sync` requirement to `Terminal`
fc9f1f756965 pagetable: implement `Send` for `PageTableRef`
ff36fedbe646 mm/validate: implement `Send` for `ValidBitmap`
3d285e31988e locking: fix `Sync` bounds + impl `Send` for SpinLock
c6d3e59cca8a console: use static instead of static mut
a2f15cc309f4 mm/alloc: use static for `ALLOCATOR`
ea7d3441025a cpu/gdt: specify layout
1db43453dba1 cpu/gdt: get rid of GDT_DESC global
714ed933d07f fs: remove `Send`/`Sync` impls in favor of bounds
```

src/locking/rwlock.rs
```diff
@@ -5,62 +5,62 @@
...
 #[derive(Debug)]
-pub struct RWLock<T: Debug> {
+pub struct RWLock<T> {
     rwlock: AtomicU64,
     data: UnsafeCell<T>,
 }
 
-unsafe impl<T: Debug> Sync for RWLock<T> {}
+unsafe impl<T: Send> Send for RWLock<T> {}
+unsafe impl<T: Send + Sync> Sync for RWLock<T> {}
```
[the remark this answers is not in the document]
Freax13: > Hmm, I don't buy the justification for this. In my thinking, the purpose of `RWLock` is to add `Sync` to any type by:
> 
>   1. Serializing writes wrt. other writes and readers (there is no reader active when write-lock is aquired)
> 
>   2. Allowing multiple readers at the same time by using the fact that immutable values are always `Sync`

Not all types allow shared accesses from multiple threads. [`Rc`](https://doc.rust-lang.org/std/rc/struct.Rc.html#impl-Sync-for-Rc%3CT,+A%3E) can't be accessed because its reference count isn't atomic, so it can't be accessed from multiple threads. [`RefCell`](https://doc.rust-lang.org/core/cell/struct.RefCell.html#impl-Sync-for-RefCell%3CT%3E) can't be accessed from multiple threads because the way it's tracking borrows is not thread-safe (i.e. doesn't use atomics).

I'm not sure what you mean by 

> immutable values are always `Sync`

but I want to point out that interior mutability exists (e.g. in `RefCell`), so mutations can also occur through immutable references. 

> Can you elaborate a bit more on the justification for this change?

The `Send` bound on the `Send` impl is required because another thread will be able to obtain a mutable reference to the contained value after the `RWLock` has been sent to that other thread.
The `Send` bound on the `Sync` impl is required because another thread is able to obtain a mutable reference to the contained value from a shared reference by write locking the thread.
The `Sync` bound on the `Sync` impl is required because multiple threads are able to create shared references to the contained value from a shared reference by read-locking the value at the same time.

FWIW these are also the bounds used by other implementations e.g. [std](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#impl-Send-for-RwLock%3CT%3E), [tokio](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html#impl-Send-for-RwLock%3CT%3E) or [spin](https://docs.rs/spin/latest/spin/rwlock/struct.RwLock.html#impl-Send-for-RwLock%3CT,+R%3E).

src/fs/filesystem.rs
```diff
@@ -129,20 +125,17 @@ impl SvsmFs {
     }
 }
 
-static mut FS_ROOT: SvsmFs = SvsmFs::new();
+static FS_ROOT: RWLock<SvsmFs> = RWLock::new(SvsmFs::new());
```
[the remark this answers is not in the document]
Freax13: > Is there another reason for the lock besides removing the `mut`?

Is there a reason to keep the `static mut`? `static mut` is kind of a huge foot-gun, so IMHO it should be avoided if possible. If you don't agree, I'd also be fine with dropping this commit.

> This change has a significant impact on the scalability of the FS code, so if there is no other reason I'd rather live the the `mut` than introducing a lock.

How so? The lock will never be contented, so I don't expect a significant slow-down.
Arguably the atomic updates from multiple threads are not completely free even if there is no contention because the CPU will still have to lock the cache lines to perform the updates to the lock's read counter, but then again the same can be said about the reference counter of the `Arc` in `SvsmFs` that contains the root directory.

src/locking/rwlock.rs
```diff
@@ -5,62 +5,62 @@
 // Author: Joerg Roedel <jroedel@suse.de>
 
 use core::cell::UnsafeCell;
-use core::fmt::Debug;
 use core::ops::{Deref, DerefMut};
 use core::sync::atomic::{AtomicU64, Ordering};
 
 #[derive(Debug)]
-pub struct ReadLockGuard<'a, T: Debug> {
+pub struct ReadLockGuard<'a, T> {
```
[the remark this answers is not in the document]
Freax13: > There was some reasoning behind adding those bounds. @00xc might know more, probably because it will make it easier to add a Debug trait to structs containing a lock.

Note that `RWLock<T>` and `SpinLock<T>` can implement `Debug` (and still do after my changes) even if `T` doesn't implement `Debug` because they don't actually print the contained value as doing so would require taking the lock.

Freax13:
I rebased onto the latest master branch (and dropped some of the commits that were already merged).

Issue #107, opened 2023-09-28
mm/alloc: undefined behavior in `SvsmAllocator::dealloc()`


00xc:
[Miri](/rust-lang/miri) reports undefined behavior when running the `mm::alloc::test_slab_alloc_free_many` test:

```console
$ MIRIFLAGS="-Zmiri-permissive-provenance" cargo +nightly miri test --target=x86_64-unknown-linux-gnu test_slab_alloc_free_many
Preparing a sysroot for Miri (target: x86_64-unknown-linux-gnu)... done
    Finished test [unoptimized + debuginfo] target(s) in 0.01s
     Running unittests src/lib.rs (target/miri/x86_64-unknown-linux-gnu/debug/deps/svsm-4dadd712fd0a3e32)

running 1 test
test mm::alloc::test_slab_alloc_free_many ... error: Undefined Behavior: trying to retag from <wildcard> for SharedReadWrite permission at alloc174[0x0], but no exposed tags have suitable permission in the borrow stack for this location
    --> src/mm/alloc.rs:1276:17
     |
1276 |                 (*slab).deallocate(virt_addr);
     |                 ^^^^^^^
     |                 |
     |                 trying to retag from <wildcard> for SharedReadWrite permission at alloc174[0x0], but no exposed tags have suitable permission in the borrow stack for this location
     |                 this error occurs as part of two-phase retag at alloc174[0x0..0x40]
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
     = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
     = note: BACKTRACE:
     = note: inside `<mm::alloc::SvsmAllocator as core::alloc::GlobalAlloc>::dealloc` at src/mm/alloc.rs:1276:17: 1276:24
note: inside `mm::alloc::test_slab_alloc_free_many`
    --> src/mm/alloc.rs:1569:26
     |
1569 |                 unsafe { ALLOCATOR.dealloc(*p, layout) };
     |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside closure
    --> src/mm/alloc.rs:1539:32
     |
1537 | #[test]
     | ------- in this procedural macro expansion
1538 | // Allocate and free a couple of objects for each slab size.
1539 | fn test_slab_alloc_free_many() {
     |                                ^
[...]

PR #110, opened 2023-09-30
mm/alloc: fix UB in shrink_slab

Freax13:
This pr fixes the issues described in https://github.com/coconut-svsm/svsm/issues/107#issuecomment-1741849531.


00xc:
Can we get an explanation of what the bug was in the commit text, for documentation purposes? I would also suggest linking to the issue via something like:

```
Fixes: #107 
Signed-off-by: ...
```

00xc (approved):
LGTM

PR #111, opened 2023-10-02
docs: Document ELF and ACPI tables

Zildj1an:
Document ELF and ACPI tables code and structures.
This is related to issue #74 Document COCONUT-SVSM.

### DIFF
--- a/src/acpi/tables.rs
+++ b/src/acpi/tables.rs
@@ -13,17 +13,34 @@ use alloc::vec::Vec;
 use core::mem;
 use log;
 
+/// ACPI Root System Description Pointer (RSDP)
+/// used by ACPI programming interface
 #[derive(Debug, Default)]
 #[repr(C, packed)]
 struct RSDPDesc {
+    /// Signature must contain "RSD PTR"
     sig: [u8; 8],
+    /// Checksum to add to all other bytes
     chksum: u8,
+    /// OEM-supplied string
     oem_id: [u8; 6],
+    /// Revision of the ACPI
     rev: u8,
+    /// Physical address of the RSDT
     rsdt_addr: u32,
 }
 
 impl RSDPDesc {
+    /// Create an RSPDesc instance from FwCfg
+    ///
+    /// # Arguments
+    ///
+    /// - `fw_cfg`: A reference to the FwCfg instance.
+    ///
+    /// # Returns
+    ///
+    /// A [`Result`] containing the [`RSDPDesc`] if successful, or an [`SvsmError`] on failure.
+    ///
     fn from_fwcfg(fw_cfg: &FwCfg) -> Result<Self, SvsmError> {
         let mut buf = mem::MaybeUninit::<Self>::uninit();
         let file = fw_cfg.file_selector("etc/acpi/rsdp")?;
@@ -46,20 +63,32 @@ impl RSDPDesc {
 
 #[derive(Copy, Clone, Debug, Default)]
 #[repr(C, packed)]
+/// Raw header of an ACPI table. It corresponds to the beginning
+/// portion of ACPI tables, before any specific table data
 struct RawACPITableHeader {
+    /// Signature specificies the type of ACPI table
     sig: [u8; 4],
+    /// Length of the table
     len: u32,
+    /// Revision (signature field)
     rev: u8,
+    /// Checksum for data integrity
     chksum: u8,
+    /// OEM-supplied string to identify OEM
     oem_id: [u8; 6],
+    /// OEM-supplied string to identify tables
     oem_table_id: [u8; 8],
+    /// OEM-supplied version number
     oem_rev: u32,
+    /// ID for compiler
     compiler_id: [u8; 4],
+    /// Revision of compiler used to create the table
     compiler_rev: u32,
 }
 
 #[derive(Debug, Default)]
 #[allow(dead_code)]
+/// Higher level representation of the raw ACPI table header
 struct ACPITableHeader {
     sig: [u8; 4],
     len: u32,
@@ -73,6 +102,17 @@ struct ACPITableHeader {
 }
 
 impl ACPITableHeader {
+    /// Create a new [`ACPITableHeader`] from a raw [`RawACPITableHeader`].
+    ///
+    /// This constructor converts a raw ACPI table header into a higher-level [`ACPITableHeader`].
+    ///
+    /// # Arguments
+    ///
+    /// * `raw` - A [`RawACPITableHeader`] containing the raw header data.
+    ///
+    /// # Returns
+    ///
+    /// A new [`ACPITableHeader`] instance.
     const fn new(raw: RawACPITableHeader) -> Self {
         Self {
             sig: raw.sig,
@@ -87,6 +127,7 @@ impl ACPITableHeader {
         }
     }
 
+    /// Print a human-readable summary of the ACPI table header's fields
     #[allow(dead_code)]
     fn print_summary(&self) {
         let sig = FixedString::from(self.sig);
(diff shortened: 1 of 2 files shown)


Zildj1an:
@00xc this is the new PR, as you can see I have included the discussed changes as well as code examples for ELF. We need to ignore the Rust check in this case because it can't use my_elf_parser which ELF's doc code examples depends on.

00xc:
> @00xc this is the new PR, as you can see I have included the discussed changes as well as code examples for ELF. We need to ignore the Rust check in this case because it can't use my_elf_parser which ELF's doc code examples depends on.

Um, no, we just need to provide valid code snippets.

For example, the doctest for `Elf64AddrRange::len()` is failing because the example not only has an invalid import, but also is invalid code - `Elf64Addr` it is an alias, not a type constructor. This fixes it:

```diff
diff --git a/src/elf/mod.rs b/src/elf/mod.rs
index 379ddca..bf04804 100644
--- a/src/elf/mod.rs
+++ b/src/elf/mod.rs
@@ -216,11 +216,11 @@ impl Elf64AddrRange {
     /// # Examples
     ///
     /// ```rust
-    /// use my_elf_parser::{Elf64AddrRange, Elf64Addr};
+    /// use svsm::elf::{Elf64AddrRange, Elf64Addr};
     ///
     /// let range = Elf64AddrRange {
-    ///     vaddr_begin: Elf64Addr(0x1000),
-    ///     vaddr_end: Elf64Addr(0x1100),
+    ///     vaddr_begin: 0x1000,
+    ///     vaddr_end: 0x1100,
     /// };
     ///
     /// assert_eq!(range.len(), 0x100);
```

Here's another example: the patch below fixes the `load_acpi_cpu_info()` doctest. You will also notice the generated docs will only include the relevant parts of the code, since we use `#`.

```diff

diff --git a/src/acpi/tables.rs b/src/acpi/tables.rs
index 9d8b8d8..74660ec 100644
--- a/src/acpi/tables.rs
+++ b/src/acpi/tables.rs
@@ -450,10 +450,22 @@ pub struct ACPICPUInfo {
 /// # Example
 ///
 /// ```
-/// use svsm::acpi::load_acpi_cpu_info;
+/// use svsm::acpi::tables::load_acpi_cpu_info;
 /// use svsm::fw_cfg::FwCfg;
+/// use svsm::io::IOPort;
 ///
-/// let fw_cfg = FwCfg::new(/* initialize your FwCfg interface */);
+/// struct MyIo;
+///
+/// impl IOPort for MyIo {
+/// #     fn outb(&self, _port: u16, _value: u8) {}
+/// #     fn outw(&self, _port: u16, _value: u16) {}
+/// #     fn inb(&self, _port: u16) -> u8 { 0 }
+/// #     fn inw(&self, _port: u16) -> u16 { 0 }
+///     // your implementation
+/// }
+///
+/// let io = MyIo;
+/// let fw_cfg = FwCfg::new(&io);
 /// match load_acpi_cpu_info(&fw_cfg) {
 ///     Ok(cpu_info) => {
 ///         for info in cpu_info {
```

src/acpi/tables.rs
```diff
@@ -13,17 +13,34 @@ use alloc::vec::Vec;
...
+    /// Physical address of the RSDT
     rsdt_addr: u32,
 }
 
 impl RSDPDesc {
+    /// Create an RSPDesc instance from FwCfg
+    ///
+    /// # Arguments
+    ///
+    /// - [`fw_cfg`]: A reference to the FwCfg instance.
```
00xc, line 38: This use of square brackets does not do anything, sadly. As far as I know, they only work for types, not variable names.

00xc: There's a few more, just using this one as an example to avoid spamming.

00xc: I just noticed, apparently run `cargo doc` will already tell you about this:

```console
$ cargo doc
warning: unresolved link to `fw_cfg`
   --> src/acpi/tables.rs:438:9
    |
438 | /// * [`fw_cfg`]: A reference to the Firmware Configuration (FwCfg) interface for accessing ACPI tables.
    |         ^^^^^^ no item named `fw_cfg` in scope
    |
    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
```

00xc:
Happy to help! Doctests look good, but there's still a few uses of square brackets on variable names, we should get rid of those.

src/elf/mod.rs
```diff
@@ -764,18 +1164,29 @@ impl Elf64Hdr {
...
     pub p_memsz: Elf64Xword,
+    /// Alignment of the segment in memory and in the file
     pub p_align: Elf64Xword,
 }
 
+// Attributes of an ELF64 program header, to specify whether
+// the segment is readable, writable, and/or executable
 bitflags! {
     #[derive(Debug)]
     pub struct Elf64PhdrFlags : Elf64Word {
```
00xc, line 1192: These need to be `///` and be inside the `bitflags!` macro in order to show up in the generated docs.

src/elf/mod.rs
```diff
@@ -814,6 +1237,12 @@ impl Elf64Phdr {
         }
     }
 
+    /// Verifies the integrity and validity of the program header.
+    ///
+    /// # Returns
+    ///
+    /// Returns [`Ok(())`] if the program header is valid; otherwise, an [`Err`]
```
00xc, line 1244: This should just be `Ok` without the parenthesis. This won't show up in the generated docs because it is a private function, so it is just a nitpick.

00xc: Just realized `Ok<()>` would also work.

src/elf/mod.rs
```diff
@@ -599,6 +919,28 @@ impl<'a> Elf64File<'a> {
...
+    /// * `image_load_addr` - The virtual address where the ELF image is loaded in memory.
+    ///
+    /// # Returns
+    ///
+    /// - [`Ok(Some(iterator))`]: If relocations are successfully applied, returns an iterator
+    ///   over the applied relocations.
+    /// - [`Ok(None)`]: If no relocations are present or an error occurs during processing,
+    ///   returns [`None`].
+    /// - [`Err(ElfError)`]: If an error occurs while processing relocations, returns an
+    ///   [`ElfError`].
```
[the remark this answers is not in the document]
00xc: There's a few more, just using this one as an example. I mostly care if they are in `pub` interfaces, as they won't display properly in the generated docs.

Zildj1an:
@00xc I have updated the PR following your feedback :+1:

src/elf/mod.rs
```diff
@@ -599,6 +919,28 @@ impl<'a> Elf64File<'a> {
...
+    /// loaded in memory.
+    ///
+    /// # Arguments
+    ///
+    /// * `rela_proc` - A relocation processor implementing the [`Elf64RelocProcessor`] trait.
+    /// * `image_load_addr` - The virtual address where the ELF image is loaded in memory.
+    ///
+    /// # Returns
+    ///
+    /// - [`Ok<Some(iterator)>`]: If relocations are successfully applied, returns an iterator
```
00xc, line 937: This one still does not render properly

src/elf/mod.rs
```diff
@@ -1532,22 +2313,37 @@ impl<'a, RP: Elf64RelocProcessor> Elf64AppliedRelaIterator<'a, RP> {
...
+    /// and returns the result.
+    ///
+    /// If there are no more relocations to process, [`None`] is returned to signal
+    /// the end of the iterator.
+    ///
+    /// # Returns
+    ///
+    /// - [`Some(Ok(None))`]: If the relocation entry indicates no operation (type == 0).
+    /// - [`Some(Ok(Some(reloc_op)))`]: If a relocation operation is successfully applied.
+    /// - [`Some(Err(ElfError))`]: If an error occurs during relocation processing.
```
00xc, line 2326: These ones as well

src/elf/mod.rs
```diff
@@ -1215,21 +1815,31 @@ pub struct Elf64ImageLoadSegmentIterator<'a> {
 impl<'a> Iterator for Elf64ImageLoadSegmentIterator<'a> {
     type Item = Elf64ImageLoadSegment<'a>;
 
+    /// Advances the iterator to the next ELF64 image load segment and returns it.
+    ///
+    /// # Returns
+    ///
+    /// - [`Some(Elf64ImageLoadSegment)`] if there are more segments to iterate over.
```
00xc, line 1822: Also this one.

00xc:
Okay this looks really good now, but I just noticed there are still some warnings with `cargo doc --document-private-items` (just realized this flag exists). They should be pretty straightforward to fix.

Zildj1an:
You're correct @00xc . I have fixed those warnings :)

00xc (approved):

Issue #34, opened 2023-05-10
Implement fuzzing via `cargo fuzz`


00xc:
Identify interesting interfaces to fuzz, from a security perspective, and implement fuzzing test cases for them. These should ideally include at least data coming from the hypervisor, guest kernel and VMPL0 userspace.

joergroedel:
Thanks for proposing this, fuzzing would be great to have. In addition, stress-tests are also something worth looking into.

PR #113, opened 2023-10-02
fuzzing: initial fuzzing implementation

00xc:
This PR adds:
* Basic fuzzing infrastructure via `cargo-fuzz`.
* Harnesses for the following interfaces:
    * Firmware metadata (`fw_meta`)
    * ACPI table parsing.
    * Filesystem operations.
* Relevant documentation.
* Updates to CI to run on the new pieces of code.

Works towards #34


FUZZING.md
```diff
@@ -0,0 +1,79 @@
...
+=======
+
+The COCONUT SVSM includes several fuzzing harnesses to find bugs in
+security-critical interfaces. These do not currently provide coverage
+of all interesting interfaces, so contributions are welcome.
+
+At the moment, fuzzing is done through
+[cargo-fuzz](/rust-fuzz/cargo-fuzz). You may find a complete tutorial
+for this tool on the
+[rust-fuzz website](https://rust-fuzz.github.io/book/cargo-fuzz.html).
```
joergroedel, line 11: Can you please add a section describing the dependencies and how install them on openSUSE? For me it was `cargo-fuzz`, `clang` and `lld`.

src/acpi/tables.rs
```diff
@@ -125,6 +125,10 @@ impl ACPITable {
         let content = ptr.get(..size).ok_or(SvsmError::Acpi)?;
 
         let mut buf = Vec::<u8>::new();
+        #[cfg(fuzzing)]
+        if size > 0x1000 {
+            return Err(SvsmError::Mem);
+        }
```
joergroedel, line 131: Why is that needed?

src/acpi/tables.rs
```diff
@@ -176,6 +180,10 @@ impl ACPITableBuffer {
         let size = file.size() as usize;
 
         let mut buf = Vec::<u8>::new();
+        #[cfg(fuzzing)]
+        if size > 0x1000 {
+            return Err(SvsmError::Mem);
+        }
```
joergroedel, line 186: Same here, why is that needed?

src/acpi/tables.rs
```diff
@@ -286,9 +294,10 @@ pub fn load_acpi_cpu_info(fw_cfg: &FwCfg) -> Result<Vec<ACPICPUInfo>, SvsmError>
             .content_ptr::<RawMADTEntryHeader>(offset)
             .ok_or(SvsmError::Acpi)?;
         let (madt_type, entry_len) = unsafe { ((*entry_ptr).entry_type, (*entry_ptr).entry_len) };
+        let entry_len = usize::from(entry_len);
 
         match madt_type {
-            0 => {
+            0 if entry_len == mem::size_of::<RawMADTEntryLocalApic>() => {
```
joergroedel, line 300: This looks like a fix for an issue found by fuzzing. Can you please submit all such fixes in a separate PR, please? We can merge them independently of the fuzzer itself.

src/fw_cfg.rs
```diff
@@ -124,6 +126,11 @@ impl<'a> FwCfg<'a> {
         self.select(FW_CFG_FILE_DIR);
         let n: u32 = self.read_be();
 
+        #[cfg(fuzzing)]
+        if n > 0x1000 {
+            return Err(SvsmError::FwCfg(FwCfgError::TooManyFiles));
+        }
+
```
joergroedel, line 133: This does not need to be limited to fuzzing. I think we can safely assume a maximum supported buffer size for the fw_cfg file-list.

joergroedel (changes requested):
I made a few comments, but besides those this looks good overall. Thanks for working on fuzzing, it already uncovered some bugs.

src/acpi/tables.rs
00xc: Because the size is a hypervisor-controlled value. Since we are fuzzing from the perspective of a malicious hypervisor, we would potentially be allocating thousands of bytes on each fuzzing iteration, which would slow down the fuzzer greatly with little to no additional coverage.

src/fw_cfg.rs
00xc: I was not sure this would make sense in regular builds, as the [only size limited imposed by qemu](https://www.qemu.org/docs/master/specs/fw_cfg.html#file-directory-key-0x0019-fw-cfg-file-dir) is `u32::MAX`. However I would agree that it would not make much sense to have such amount of files. Should we just set the limit it to 4096 for all builds?

src/acpi/tables.rs
00xc: Opened #125

FUZZING.md
00xc: Done in the new version

src/fw_cfg.rs
00xc: Okay, so this one is about the number of files in a directory. There's two others:
* The size of the `etc/acpi/tables` file.
* The size of an ACPI table, obtained from a field in the header (`RawACPITableHeader`).

Should we limit these as well for all builds? Same limit values?

joergroedel: > Okay, so this one is about the number of files in a directory. There's two others:
> 
>     * The size of the `etc/acpi/tables` file.

For that we can probably set a limit to 128KiB, that is the current size. Maybe also 256KiB, we can't allocate more via the allocator anyway.

>     * The size of an ACPI table, obtained from a field in the header (`RawACPITableHeader`).

Here we can just check that the size + offset of the table fully fits into buffer read from the `etc/acpi/tables` file.

00xc: - File amount limited to 4096.
- `etc/acpi/tables` file size limited to 128 KiB.
- ACPI table size not limited, as we first try to get a slice to the buffer according to its announced length.

00xc:
Also made `parse_fw_meta_data()` unsafe, as the caller must guarantee that the provided `VirtAddr` is valid and points to a region of the adequate size (4k).

00xc:
Last version: rebased on main and fixed the ACPI harness to accommodate the new `Sync` bound for the `IOPort` trait.

fuzz/fuzz_targets/fs.rs
```diff
@@ -0,0 +1,230 @@
...
+    Truncate(usize, usize),
+}
+
+fn get_idx<T>(v: &[T], idx: usize) -> Option<usize> {
+    if !v.is_empty() {
+        Some(idx % v.len())
+    } else {
+        None
+    }
+}
```
Freax13, line 51:
```suggestion
fn get_idx<T>(v: &[T], idx: usize) -> Option<usize> {
    idx.checked_rem(v.len())
}
```

fuzz/fuzz_targets/fs.rs
```diff
@@ -0,0 +1,230 @@
...
+
+/// A handle for a file that also holds its original name
+#[derive(Debug)]
+struct Handle {
+    fd: FileHandle,
+    name: Rc<str>,
+}
```
Freax13, line 58:
```suggestion
/// A handle for a file that also holds its original name
#[derive(Debug)]
struct Handle<'a> {
    fd: FileHandle,
    name: &'a str,
}
```

Allocating into an `Rc` is not needed here.

fuzz/fuzz_targets/fs.rs
```diff
@@ -0,0 +1,230 @@
...
+    MkdirNamed(usize),
+    ListDir(&'a str),
+    ListDirNamed(usize),
+    Read(usize, usize),
+    Write(usize, usize),
+    Seek(usize, usize),
+    Truncate(usize, usize),
+}
+
+fn get_idx<T>(v: &[T], idx: usize) -> Option<usize> {
```
Freax13, line 45: All callers except one then go on to use the returned index to index into the slice. Have you considered just returning a `Option<&T>` instead of `Option<usize>`?

src/fw_meta.rs
```diff
@@ -245,15 +244,17 @@ fn find_table<'a>(uuid: &Uuid, mem: &'a [u8]) -> Option<&'a [u8]> {
...
 }
 
-pub fn parse_fw_meta_data() -> Result<SevFWMetaData, SvsmError> {
+/// Parse the firmware metadata located at the memory page starting at `vstart`.
+///
+/// # Safety
+///
+/// The caller must ensure that the provided virtual address points to a
+/// page-sized mapped memory region
+pub unsafe fn parse_fw_meta_data(vstart: VirtAddr) -> Result<SevFWMetaData, SvsmError> {
```
Freax13, line 253: Should this function maybe just take a `&RawMetaBuffer` or `&[u8]`?

Freax13 (changes requested):

fuzz/fuzz_targets/fs.rs
00xc: Fixed in v2.

fuzz/fuzz_targets/fs.rs
00xc: Fixed in v2.

src/fw_meta.rs
00xc: `RawMetaBuffer` is private, so I would avoid that. A `&[u8]` would be better I think, that way we can check the length. With that I think we can avoid the `unsafe`, since `RawMetaBuffer` is `#[repr(C, packed)]`, which means there are no alignment requirements.

Freax13: > `RawMetaBuffer` is private, so I would avoid that. A `&[u8]` would be better I think, that way we can check the length. With that I think we can avoid the `unsafe`, since `RawMetaBuffer` is `#[repr(C, packed)]`, which means there are no alignment requirements.

Sure, sounds good to me. 

I know y'all don't like adding external dependencies, but [`bytemuck`](https://docs.rs/bytemuck/latest/bytemuck/) is a really good crate that, among other things, makes it possible to easily and safely [convert byte slices to Rust types](https://docs.rs/bytemuck/latest/bytemuck/fn.from_bytes.html).

00xc: Yeah there's also [zerocopy](/google/zerocopy), I've used it for the [PackIt](/coconut-svsm/packit) crate. I'm not sure how they compare to one another, but for now we can live without them I think. Definitely something to consider if we get more parsing-intensive code in the future.

src/fw_meta.rs
```diff
@@ -245,16 +244,18 @@ fn find_table<'a>(uuid: &Uuid, mem: &'a [u8]) -> Option<&'a [u8]> {
...
-    // representations.
-    let raw_meta = unsafe { &*vstart.as_ptr::<RawMetaBuffer>() };
+
+    if mem.len() != size_of::<RawMetaBuffer>() {
+        return Err(SvsmError::Firmware);
+    }
+
+    // Safety: `RawMetaBuffer` has no invalid representations and is
+    // `repr(C, packed)`, which means there are no alignment requirements.
+    // We have also verified that the size of the slice matches.
```
Freax13, line 257: Great job on the safety comments!

Freax13 (approved):
Looks good to me!

src/fw_meta.rs
Freax13: > Yeah there's also [zerocopy](/google/zerocopy), I've used it for the [PackIt](/coconut-svsm/packit) crate. I'm not sure how they compare to one another, 

I haven't used zerocopy, but AFAICT bytemuck can do more than zerocopy. They are pretty similar though.

> but for now we can live without them I think. Definitely something to consider if we get more parsing-intensive code in the future.

Fair enough. Definitively shouldn't be part of this pr anyways since it's unrelated to fuzzing.

src/fw_meta.rs
00xc: Thanks!

00xc:
Dropped `fuzzing: limit hypervisor-controlled allocation sizes` in favor of #140. With that this PR has no unrelated commits.

joergroedel (approved):
Looks good now.

Issue #115, opened 2023-10-02
acpi/tables: undefined behavior via `core::slice::from_raw_parts()`


00xc:
I found the following UB through fuzzing + Miri.

In `ACPITableBuffer::load_tables()` a `&[u32]` is set up via `core::slice::from_raw_parts()`:

https://github.com/coconut-svsm/svsm/blob/ad4ed6cb4a78380a2b8c66b08b04fcc024fa56d3/src/acpi/tables.rs#L200-L206

However, this function requires for the pointer to be aligned to the size of `T` (4 in this case), which is not guaranteed.

`rdst.content()` returns a slice of a `Vec`:

https://github.com/coconut-svsm/svsm/blob/ad4ed6cb4a78380a2b8c66b08b04fcc024fa56d3/src/acpi/tables.rs#L142-L146

Which was previously allocated with a hypervisor-controlled size:

https://github.com/coconut-svsm/svsm/blob/ad4ed6cb4a78380a2b8c66b08b04fcc024fa56d3/src/acpi/tables.rs#L127-L130

PR #116, opened 2023-10-03
acpi/tables: fix undefined behavior due to pointer misalignment

00xc:
Fix an undefined behavior that would occur when the alignment requirements for `core::slice::from_raw_parts()` would not be held. The pointer passed to this function must be aligned so that each member of the slice is itself aligned, which could not be guaranteed.

Instead of creating a new slice of u32's, iterate over the individual bytes in chunks of 4 and parse each u32 individually.

Fixes: #115
Fixes: 62f5c6fe "acpi: return ACPITable content as byte slice instead of raw pointer"

### DIFF
--- a/src/acpi/tables.rs
+++ b/src/acpi/tables.rs
@@ -199,14 +199,11 @@ impl ACPITableBuffer {
 
         let rsdt = self.acpi_table_from_offset(desc.rsdt_addr as usize)?;
         let content = rsdt.content().ok_or(SvsmError::Acpi)?;
-        let offsets = unsafe {
-            let ptr = content.as_ptr().cast::<u32>();
-            let size = content.len() / mem::size_of::<u32>();
-            core::slice::from_raw_parts(ptr, size)
-        };
+        let offsets = content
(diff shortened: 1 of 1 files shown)


joergroedel:
This fails to boot here with

```
[SVSM] ERROR: Panic: CPU[0] panicked at 'Failed to load ACPI tables: Acpi', src/svsm.rs:435:44
[SVSM] ---BACKTRACE---:
[SVSM]   [0xffffff8000006101]
[SVSM] ---END---
```

joergroedel:
Backtrace points to `svsm_main()`, so not very helpful.

00xc:
> This fails to boot here with
> 
> ```
> [SVSM] ERROR: Panic: CPU[0] panicked at 'Failed to load ACPI tables: Acpi', src/svsm.rs:435:44
> [SVSM] ---BACKTRACE---:
> [SVSM]   [0xffffff8000006101]
> [SVSM] ---END---
> ```

Had the wrong endianness for some reason, it should properly boot now.

joergroedel:
Thanks, boots now.

PR #119, opened 2023-10-05
test:  Use target OS to determine use of bare-metal allocator

roy-hopkins:
Our custom bare-metal allocator (`SvsmAllocator`) cannot be used in regular userspace binaries so needs to be disabled for unit and doc testing. The previous fix for documentation tests used the `doctest` cfg to conditionally use the allocator, passing the configuration option via RUSTFLAGS. Unfortunately, this caused compilation issues with dependencies that make use of #[cfg(doctest)] such as memoffset which is used by intrusive-collections.

Instead, this change makes use of the fact that the tests run on an OS platform whereas the SVSM is built for a target OS of 'none'. The custom allocator is enabled only when `target_os` = "none". This means the specific docs test invocation can be removed from the test Makefile target as the doc tests will be run alongside the unit tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ test = false
 
 [lib]
 test = true
-doctest = false
+doctest = true
 
 [dependencies]
 bitflags = "2.4"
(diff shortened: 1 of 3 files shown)


00xc (approved):
Yeah this makes much more sense.

PR #120, opened 2023-10-08
run tests inside SVSM

Freax13:
This PR makes it possible to run tests inside the SVSM. Simply setting the `TEST` environment variable before running `make` will produce an SVSM build that runs tests instead of running the guest.

This is implemented using the nightly `custom_test_frameworks` feature to specify our own tests runner. This nightly feature is only required when building tests for running inside the SVSM, non-test builds and normal test execution can continue to use a stable toolchain. The `#[test]` macro expects a crate named `test` to exist, so a small, no_std compatible subset of the `test` crate has been copied.

Some  tests currently fail when run inside the SVSM (AFAICT mostly for benign reasons related to setup of the allocator and file system). Those tests are currently disabled using the `ignore` attribute. Fixing those tests is probably best left to separate PRs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,9 @@ utils/gen_meta: utils/gen_meta.c
 utils/print-meta: utils/print-meta.c
 	cc -O3 -Wall -o $@ $<
 
+utils/cbit: utils/cbit.c
+	cc -O3 -Wall -o $@ $<
+
 stage1/meta.bin: utils/gen_meta utils/print-meta
 	./utils/gen_meta $@
 
--- a/Makefile
+++ b/Makefile
@@ -57,4 +57,4 @@ clean:
 	cargo clean
 	rm -f stage1/stage2.bin svsm.bin stage1/meta.bin stage1/kernel.elf stage1/stage1 stage1/svsm-fs.bin ${STAGE1_OBJS} utils/gen_meta utils/print-meta
 
-.PHONY: stage1/stage2.bin stage1/kernel.elf svsm.bin clean stage1/svsm-fs.bin
+.PHONY: stage1/stage2.bin stage1/kernel.elf svsm.bin clean stage1/svsm-fs.bin test
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ FS_FILE ?= none
 
 STAGE1_OBJS = stage1/stage1.o stage1/reset.o
 
-all: svsm.bin
+all: stage1/kernel.elf svsm.bin
 
 test:
 	cargo test --target=x86_64-unknown-linux-gnu
@@ -42,7 +42,7 @@ ifneq ($(FS_FILE), none)
 endif
 	touch stage1/svsm-fs.bin
 
-stage1/stage1.o: stage1/stage1.S stage1/stage2.bin stage1/kernel.elf stage1/svsm-fs.bin
+stage1/stage1.o: stage1/stage1.S stage1/stage2.bin stage1/svsm-fs.bin
 	cc -c -o $@ stage1/stage1.S
 
 stage1/reset.o:  stage1/reset.S stage1/meta.bin
--- a/Makefile
+++ b/Makefile
@@ -22,28 +22,8 @@ all: stage1/kernel.elf svsm.bin
 test:
 	cargo test --target=x86_64-unknown-linux-gnu
 
-test-in-svsm: stage1/test-kernel.elf svsm.bin
-ifndef QEMU
-	echo "Set QEMU environment variable to QEMU installation path" && exit 1
-endif
-ifndef OVMF
-	echo "Set OVMFenvironment variable to a folder containing OVMF_CODE.fd and OVMF_VARS.fd" && exit 1
-endif
-	$(QEMU)/qemu-system-x86_64 \
-		-enable-kvm \
-		-cpu EPYC-v4 \
-		-machine q35,confidential-guest-support=sev0,memory-backend=ram1,kvm-type=protected \
-		-object memory-backend-memfd-private,id=ram1,size=1G,share=true \
-		-object sev-snp-guest,id=sev0,cbitpos=$(C_BIT_POS),reduced-phys-bits=1,svsm=on \
-		-smp 8 \
-		-no-reboot \
-		-drive if=pflash,format=raw,unit=0,file=$(OVMF)/OVMF_CODE.fd,readonly=on \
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -9,7 +9,17 @@ TARGET_PATH="debug"
 endif
 
 STAGE2_ELF = "target/x86_64-unknown-none/${TARGET_PATH}/stage2"
+ifdef TEST
+KERNEL_ELF = "target/x86_64-unknown-none/${TARGET_PATH}/deps/svsm-0651abaef6a489bb"
```
roy-hopkins, line 13: I'm not sure the metadata hash can be hardcoded here. I think it can change based on compiler version and other parameters.

Freax13: I think you're right in that there is no guarantee that the hash will be stable. Cargo doesn't seem to have way to fix the path: https://github.com/rust-lang/cargo/issues/1924). Anecdotally, over the course of development for this pr, I've been using multiple nightly versions and the hash remained the same.

roy-hopkins: I spotted in when I was trying it out - I was playing around with my own build script that deploys to our test machine and I ended up making the hash value change. Unfortunately I don't know what I did to cause it to change and it reverted back to the original when I tried to reproduce it.

Freax13: Addressed in ace2c559346b63932e5f3323bfaf9ae9f2d423e3

Makefile
```diff
@@ -2,22 +2,28 @@ FEATURES ?= "default"
...
-KERNEL_ELF = "target/x86_64-unknown-none/${TARGET_PATH}/deps/svsm-0651abaef6a489bb"
+KERNEL_ELF = target/x86_64-unknown-none/${TARGET_PATH}/svsm-test
 CARGO_ENV_KERNEL = LINK_TEST=1
 CARGO_TOOLCHAIN_KERNEL = +nightly
 CARGO_ARGS_KERNEL = --tests
+CARGO_VERB_KERNEL = test
+# Define a test runner that simply copies the binary to a static location for
+# further processing. We do this because there's no good way to statically know
+# the test binary location.
+CARGO_ARGS_KERNEL += --config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ${KERNEL_ELF}"]'
```
roy-hopkins, line 21: That's a great solution!

joergroedel:
Thanks for doing this, it looks very promising! Somehow I have issues getting the test binary to build, it can't find the `core` crate. Any hint on how I can fix this on my side?

Freax13:
> Thanks for doing this, it looks very promising! Somehow I have issues getting the test binary to build, it can't find the `core` crate. Any hint on how I can fix this on my side?

Sounds like the x86_64-unknown-none target on the nightly toolchain isn't installed. Try
```
rustup target add --toolchain nightly x86_64-unknown-none
```

src/testing.rs
```diff
@@ -0,0 +1,33 @@
...
+            continue;
+        }
+
+        info!("test {} ...", test_case.desc.name.0);
+        (test_case.testfn.0)();
+    }
+
+    info!("All tests passed!");
+
+    request_termination_msr();
```
joergroedel, line 32: I wonder if we can use the qemu exit device here?

joergroedel (commented):
In general this looks great. A few things I'd like to see changed:

1. We can safely assume that the test SVSM binary will run in QEMU, so I'd like to use the QEMU exit device to get exit qemu with a given exit-code. This way running the in-kernel tests can be more easily automated.
2. Please add a Makefile target to build and run the in-kernel tests.

Is there an easy way to limit certain tests to a single test-runner? So that we run some tests only in user-space and others only with the in-kernel runner?

src/testing.rs
Freax13: Turns out we can: d8b17373c7f68cb92ff84a1079b4d79d4d049e5f

Freax13:
> In general this looks great. A few things I'd like to see changed:
> 
>   1. We can safely assume that the test SVSM binary will run in QEMU, so I'd like to use the QEMU exit device to get exit qemu with a given exit-code. This way running the in-kernel tests can be more easily automated.

Good idea! Implemented in 0ef34e232046c4c825fcd0f8402898e431212d71.
 
>   2. Please add a Makefile target to build and run the in-kernel tests.

Implemented in 773b3b2f66a8f8ba1eb62da161ae7e4c1f30641a. I added a `test-in-svsm` target. This targets determines the QEMU and OVMF installation paths from environment variables (and errors out if they are missing) to run the tests.

> Is there an easy way to limit certain tests to a single test-runner? So that we run some tests only in user-space and others only with the in-kernel runner?

We can use `cfg_attr(target_os = "none", ignore)` to prevent tests from being run inside the SVSM and vice versa.

joergroedel:
> We can use `cfg_attr(target_os = "none", ignore)` to prevent tests from being run inside the SVSM and vice versa.

A more descriptive way of specifying this would be great. Can this work with defining specific features and making tests depend on them?

Freax13:
> > We can use `cfg_attr(target_os = "none", ignore)` to prevent tests from being run inside the SVSM and vice versa.
> 
> A more descriptive way of specifying this would be great. Can this work with defining specific features and making tests depend on them?

Yes, we could either use a feature or just a raw `cfg(some_feature_name)`.

Freax13:
> > > We can use `cfg_attr(target_os = "none", ignore)` to prevent tests from being run inside the SVSM and vice versa.
> > 
> > 
> > A more descriptive way of specifying this would be great. Can this work with defining specific features and making tests depend on them?
> 
> Yes, we could either use a feature or just a raw `cfg(some_feature_name)`.

I added a cfg option `test_in_svsm`. See https://github.com/coconut-svsm/svsm/pull/120/commits/d368d2386d5c9c79e449d435874634fbfcfe271f on how this is used.

Freax13:
I just rebased onto the latest master branch.

joergroedel:
Made a few updates and stored them in [this branch](https://github.com/joergroedel/svsm/tree/custom-test-framework). Can you please look at my changes and include them into this PR?

Freax13:
> Made a few updates and stored them in [this branch](https://github.com/joergroedel/svsm/tree/custom-test-framework). Can you please look at my changes and include them into this PR?

Looks good to me, thanks!

Freax13:
I also rebased onto the latest master branch so resolve some conflicts.

joergroedel (approved):
Thanks, looks good now.

PR #121, opened 2023-10-09
INSTALL.md: Document new Linux kernel and QEMU branches

joergroedel:
The branches with the Linux kernel support for SVSM (guest and host side) have been updated to use the latest upstream patch-sets. Same for QEMU, which was updated to version 8.0.0.

Update the documentation to point to the updated branches.


roy-hopkins (approved):
LGTM

PR #122, opened 2023-10-09
SVSM: Execute Guest OS at VMPL-2

joergroedel:
Move the guest OS to VMPL-2, keeping VMPL-1 for future usage to the SVSM and VMPL-3 for future usage to the OS.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -37,7 +37,10 @@ pub const SVSM_CS_FLAGS: u16 = 0x29b;
 pub const SVSM_DS_FLAGS: u16 = 0xc93;
 pub const SVSM_TR_FLAGS: u16 = 0x89;
 
-pub const GUEST_VMPL: usize = 1;
+/// VMPL level the guest OS will be executed at.
+/// Keep VMPL 1 for the SVSM and execute the OS at VMPL-2. This leaves VMPL-3
+/// free for the OS to use in the future.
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -37,7 +37,10 @@ pub const SVSM_CS_FLAGS: u16 = 0x29b;
 pub const SVSM_DS_FLAGS: u16 = 0xc93;
 pub const SVSM_TR_FLAGS: u16 = 0x89;
 
-pub const GUEST_VMPL: usize = 1;
+/// VMPL level the guest OS will be exectued at.
```
roy-hopkins, line 40: typo: executed

roy-hopkins (approved):

tlendacky:
Why not choose VMPL-2 so that the you still have an extra VMPL for SVSM and you could then have an extra VMPL for the OS? Do you see a use for 3 VMPL levels in the SVSM?

00xc (approved):

joergroedel:
I chose VMPL-3 for now as the SVSM can only handle one OS VMPL level right now. So running the OS at VMPL-3 makes sure the OS does not try to setup any other VMPL. But we can change to VMPL-2 once that changes and once there is a use-case for the OS. The SVSM does currently not have a use-case for the other VMPLs.

PR #126, opened 2023-10-12
mm: stack: Add unit tests

Zildj1an:
Add basic unit tests for the stack. This is a start to adding unit tests to all memory components.

### DIFF
--- a/src/mm/stack.rs
+++ b/src/mm/stack.rs
@@ -125,3 +125,27 @@ pub fn free_stack(stack: VirtAddr) {
 
     STACK_ALLOC.lock().dealloc(stack);
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::mm::stack::*;
+
+    #[test]
+    fn test_allocate_and_free_stack() {
+        /*
+         * For offline testing purposes, we can't
+         * really map physical memory.
+         */
+        let stack_res = STACK_ALLOC.lock().alloc();
+        let stack = stack_res.unwrap();
+        let base_pointer = stack_base_pointer(stack);
+
+        assert!(stack >= SVSM_SHARED_STACK_BASE);
+        assert!(stack < SVSM_SHARED_STACK_END);
+
+        let bits = stack.bits();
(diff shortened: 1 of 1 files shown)


src/mm/stack.rs
```diff
@@ -125,3 +125,40 @@ pub fn free_stack(stack: VirtAddr) {
...
+         */
+        let stack_res = STACK_ALLOC.lock().alloc();
+        let stack = stack_res.unwrap();
+        let base_pointer = stack_base_pointer(stack);
+
+        assert!(stack >= SVSM_SHARED_STACK_BASE);
+        assert!(stack < SVSM_SHARED_STACK_END);
+
+        let bits = stack.bits();
+        assert_eq!(0xFFFFFFC000000000, bits);
```
joergroedel, line 147: The constant here looks like it can be expressed based on `SVSM_SHARED_STACK_BASE`

src/mm/stack.rs
```diff
@@ -125,3 +125,40 @@ pub fn free_stack(stack: VirtAddr) {
...
+         * To calculate the base address we use stack.bits()
+         * plus STACK_SIZE
+         */
+        let expected_base = if cfg!(feature = "enable-gdb") {
+            // 12 pages (GDB enabled)
+            VirtAddr::new(0xFFFFFFC000006000)
+        } else {
+            // 0 pages for STACK_PAGES_GDB (GDB disabled)
+            VirtAddr::new(0xFFFFFFC000004000)
+        };
```
joergroedel, line 159: And the expected base can be easier calculated using one of the STACK_SIZE or STACK_PAGES constants. Otherwise this test will always need adoption when we change the stack layout.

joergroedel (changes requested):
Thanks for adding this test. I left a few comments.

Zildj1an:
> Thanks for adding this test. I left a few comments.

Thanks for the feedback @joergroedel I have updated the test

joergroedel (approved):
Looks good now. Will merge once I am back at my test machine.

PR #128, opened 2023-10-13
mm/alloc: bounds-check page order before indexing

00xc:
Bounds check the order parameter to `MemoryRegion::refill_page_list()` before indexing, as an invalid order could cause an out of bounds access, resulting in a panic. The previous bounds check was too late, and also had an off-by-one bug.

This function is reachable from other modules via `allocate_pages()`, which naively forwards the order parameter.

Fixes: bee9fe779bf0 ("SVSM: Implement higher-order allocations")


p4zuu:
Nice catch. LGTM

00xc:
Tests with Miri also pass.

PR #129, opened 2023-10-13
tests/mm: use RAII guard to drop test root memory

00xc:
Replace `setup_test_root_mem()` and `destroy_test_root_mem()` by a new guard-like struct that, when acquired, grabs a global lock and initializes the test memory, and when dropped it destroys the memory before releasing the lock.

This simplifies some tests by removing the need to manually destroy the global memory. Moreover, guards on `ROOT_MEM` do not need to be dropped manually either, since they are always acquired after the memory lock, and Rust drops objects in reverse order of acquisition.

### DIFF
--- a/src/fs/filesystem.rs
+++ b/src/fs/filesystem.rs
@@ -273,11 +273,11 @@ pub fn seek(fh: &FileHandle, pos: usize) {
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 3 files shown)


00xc:
Tests with Miri pass.

roy-hopkins (approved):

PR #131, opened 2023-10-16
fw_meta: do not trust firmware metadata memory regions

### LINKED ISSUE #114: SVSM trusts firmware metadata addresses
The SVSM kernel parses out hypervisor-provided firmware metadata into the following structure:

```rust
pub struct SevFWMetaData {
    pub reset_ip: Option<PhysAddr>,
    pub cpuid_page: Option<PhysAddr>,
    pub secrets_page: Option<PhysAddr>,
    pub caa_page: Option<PhysAddr>,
    pub valid_mem: Vec<SevPreValidMem>,
}
```

The regions in `valid_mem` are blindly `PVALIDATE`'d in `validate_fw_memory()`, which can lead to remapping attacks if the hypervisor provides regions covering already validated memory.

00xc:
As described in #114, the SVSM `PVALIDATE`s and uses hypervisor-provided memory regions without any checks, which could lead to the SVSM kernel doubly validating its own memory.

In order to fix this, introduce a new type, `MemoryRegion`, to serve as a common type to perform these kinds of checks. While we are at it, use this new type everywhere possible instead of having per-module types with essentially the same functionality.

While we are at it, cleanup code around the use of memory regions.

Fixes #114.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -55,7 +55,7 @@ pub trait Address:
         self.is_aligned(PAGE_SIZE)
     }
 
-    fn checked_offset(&self, off: InnerAddr) -> Option<Self> {
+    fn checked_add(&self, off: InnerAddr) -> Option<Self> {
         self.bits().checked_add(off).map(|addr| addr.into())
     }
 
@@ -262,7 +262,7 @@ impl ops::Add<InnerAddr> for VirtAddr {
 }
 
 impl Address for VirtAddr {
-    fn checked_offset(&self, off: InnerAddr) -> Option<Self> {
+    fn checked_add(&self, off: InnerAddr) -> Option<Self> {
         self.bits()
             .checked_add(off)
             .map(|addr| sign_extend(addr).into())
--- a/src/address.rs
+++ b/src/address.rs
@@ -63,6 +63,10 @@ pub trait Address:
         self.bits().checked_sub(off).map(|addr| addr.into())
     }
 
+    fn saturating_add(&self, off: InnerAddr) -> Self {
+        Self::from(self.bits().saturating_add(off))
+    }
+
     fn page_offset(&self) -> usize {
         self.bits() & (PAGE_SIZE - 1)
     }
--- a/src/fw_cfg.rs
+++ b/src/fw_cfg.rs
@@ -6,6 +6,7 @@
 
 extern crate alloc;
 
+use crate::address::{Address, PhysAddr};
 use crate::error::SvsmError;
 use crate::mm::pagetable::max_phys_addr;
 
@@ -154,14 +155,17 @@ impl<'a> FwCfg<'a> {
     }
 
     fn read_memory_region(&self) -> MemoryRegion {
-        let start: u64 = self.read_le();
-        let size: u64 = self.read_le();
-        let end = start.saturating_add(size);
+        let start = PhysAddr::from(self.read_le::<u64>());
+        let size = self.read_le::<u64>();
+        let end = start.saturating_add(size as usize);
 
         assert!(start <= max_phys_addr(), "{start:#018x} is out of range");
         assert!(end <= max_phys_addr(), "{end:#018x} is out of range");
 
-        MemoryRegion { start, end }
+        MemoryRegion {
+            start: start.into(),
+            end: end.into(),
+        }
     }
 
     pub fn get_memory_regions(&self) -> Result<Vec<MemoryRegion>, SvsmError> {
--- a/src/fw_cfg.rs
+++ b/src/fw_cfg.rs
@@ -9,6 +9,7 @@ extern crate alloc;
 use crate::address::{Address, PhysAddr};
 use crate::error::SvsmError;
 use crate::mm::pagetable::max_phys_addr;
+use crate::utils::MemoryRegion;
 
 use super::io::IOPort;
 use super::string::FixedString;
@@ -61,20 +62,6 @@ impl FwCfgFile {
     }
     pub fn selector(&self) -> u16 {
         self.selector
-    }
-}
-
-#[derive(Clone, Copy, Debug)]
-pub struct MemoryRegion {
-    pub start: u64,
-    pub end: u64,
-}
-
-impl MemoryRegion {
-    /// Returns `true` if the region overlaps with another region with given
-    /// start and end.
-    pub fn overlaps(&self, start: u64, end: u64) -> bool {
-        self.start < end && start < self.end
     }
 }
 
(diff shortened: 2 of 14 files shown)


00xc:
@joergroedel would it also make sense to check if the regions are already validated in the `ValidBitmap`?

src/utils/memory_region.rs
```diff
@@ -0,0 +1,205 @@
...
+// Author: Carlos López <carlos.lopez@suse.com>
+
+use crate::address::Address;
+use crate::types::PageSize;
+
+/// An abstraction over a memory region, expressed in terms of physical
+/// ([`PhysAddr`](crate::address::PhysAddr)) or virtual
+/// ([`VirtAddr`](crate::address::VirtAddr)) addresses.
+#[derive(Clone, Copy, Debug)]
+pub struct MemoryRegion<A> {
```
Freax13, line 14: Is there a use case that requires supporting `start > end`? If not, can we make it an invariant of this struct that this is never the case? AFAICT `overlaps` and `contiguous` already assume this.

src/address.rs
```diff
@@ -55,7 +55,7 @@ pub trait Address:
         self.is_aligned(PAGE_SIZE)
     }
 
-    fn checked_offset(&self, off: InnerAddr) -> Option<Self> {
+    fn checked_add(&self, off: InnerAddr) -> Option<Self> {
```
Freax13, line 58: FWIW this also matches the function names of pointers in the standard library: [`pointer::add`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add) takes a `usize` (like the function here), whereas [`pointer::offset`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset) takes an `isize`.

src/mm/memory.rs
```diff
@@ -40,37 +39,31 @@ pub fn init_memory_map(fwcfg: &FwCfg, launch_info: &KernelLaunchInfo) -> Result<
...
+                MemoryRegion::from_addresses(region_after_start, region_after_end),
             );
             i += 1;
         }
     }
 
     log::info!("Guest Memory Regions:");
     for r in regions.iter() {
-        log::info!("  {:018x}-{:018x}", r.start, r.end);
+        log::info!("  {:018x}-{:018x}", r.start(), r.end());
```
Freax13, line 66: Should we maybe just implement `Debug` or `Display` for `MemoryRange`?

src/stage2.rs
```diff
@@ -157,8 +157,8 @@ pub extern "C" fn stage2_main(launch_info: &Stage1LaunchInfo) {
 
     log::info!("COCONUT Secure Virtual Machine Service Module (SVSM) Stage 2 Loader");
 
-    let kernel_region_phys_start = PhysAddr::from(r.start);
-    let kernel_region_phys_end = PhysAddr::from(r.end);
+    let kernel_region_phys_start = r.start();
+    let kernel_region_phys_end = r.end();
```
Freax13, line 161: Maybe outside of the scope of this pr, but it seems to me that maybe `init_valid_bitmap_alloc` should take a `MemoryRange`.

src/fw_cfg.rs
```diff
@@ -154,14 +155,17 @@ impl<'a> FwCfg<'a> {
     }
 
     fn read_memory_region(&self) -> MemoryRegion {
-        let start: u64 = self.read_le();
-        let size: u64 = self.read_le();
-        let end = start.saturating_add(size);
+        let start = PhysAddr::from(self.read_le::<u64>());
+        let size = self.read_le::<u64>();
+        let end = start.saturating_add(size as usize);
```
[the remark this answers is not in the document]
00xc: Start and end now need to be `PhysAddr`s in order to be compared with the return value of `max_phys_addr()`

src/utils/memory_region.rs
00xc: Yes, so the idea was to enforce this always by only having a constructor that takes a base and a length, but it made code in some places harder to read, so now it is in fact possible to construct an invalid region where `end` < `start` (`start` == `end` should be valid though) via `MemoryRegion::from_addresses()`.

Not sure what the best way to proceed is here. Maybe having `from_addresses()` always perform the check and return an `Option`?

src/utils/memory_region.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// # use svsm::address::VirtAddr;
+    /// # use svsm::types::{PAGE_SIZE, PageSize};
+    /// # use svsm::utils::MemoryRegion;
+    /// let region = MemoryRegion::new(VirtAddr::from(0xffffff0000u64), PAGE_SIZE * 2);
+    /// let mut iter = region.iter_pages(PageSize::Regular);
+    /// assert_eq!(iter.next(), Some(VirtAddr::from(0xffffff0000u64)));
+    /// assert_eq!(iter.next(), Some(VirtAddr::from(0xffffff1000u64)));
+    /// assert_eq!(iter.next(), None);
+    /// ```
+    pub fn iter_pages(&self, size: PageSize) -> impl Iterator<Item = A> {
```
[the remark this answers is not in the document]
00xc: Documented this is in the next version.

src/utils/memory_region.rs
Freax13: > Not sure what the best way to proceed is here. Maybe having `from_addresses()` always perform the check and return an `Option`?

Checking sounds reasonable. We could either return an `Option`  or we could also just assert that `start <= end`.

Freax13: Comparing might be difficult in a const context though, so maybe returning an `Option` is easier.

00xc: Another option I might go for is simply taking the minimum of the two addresses as the start and the maximum as the end. That way we avoid introducing a bunch of unwraps.

00xc: Ugh, nevermind. As you mentioned, comparison in const context does not work easily.

src/stage2.rs
00xc: Added this to the new version of the PR.

src/mm/memory.rs
00xc: `Debug` is derived, `Display` might be useful, but perhaps not all callers want this format.

src/utils/memory_region.rs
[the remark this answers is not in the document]
00xc: I thought about this but I didn't think it added much clarity

joergroedel:
Merged manually, thanks.

PR #132, opened 2023-10-16
github/workflows: Add CI test for documentation

Zildj1an:
Check documentation didn't break on pull requests, as discussed in pull request #124 (this test depends on merging first `make doc`).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,16 @@
+name: Docs
+
+on: [pull_request]
+
+jobs:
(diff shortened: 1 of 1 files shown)


00xc:
Maybe we should have this in #124

00xc:
Also, I don't know if this will work as it is, since we need cargo to be installed. This is better added to the `rust.yml` workflow.

Zildj1an:
@00xc agreed, will combine with #124 and use `rust.yml`

PR #134, opened 2023-10-17
mm/validate: fully migrate and clear bitmap & other cleanups

00xc:
* Remove useless `pub` on `ValidBitmap` methods.
* Fix: if the number of pages in the bitmap was not a multiple of 64, the remainder would not be copied over to the new bitmap during migration, which could lead to tracking validated pages as invalid.
* Use an `usize` to index into the bitmap, as offsets are always positive.
* Fully clear the bitmap after initialization.

### DIFF
--- a/src/mm/validate.rs
+++ b/src/mm/validate.rs
@@ -135,14 +135,9 @@ impl ValidBitmap {
     }
 
     fn clear_all(&mut self) {
-        let (mut i, bit) = self.index(self.region.end());
-        if bit != 0 {
-            i += 1;
-        }
-        let index: usize = i.try_into().unwrap();
-
+        let len = self.bitmap_len();
         unsafe {
-            ptr::write_bytes(self.bitmap, 0, index);
+            ptr::write_bytes(self.bitmap, 0, len);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/mm/validate.rs
```diff
@@ -152,11 +152,20 @@ impl ValidBitmap {
...
 
-    fn migrate(&mut self, new_bitmap: *mut u64) {
-        let (count, _) = self.index(self.pend);
+    /// The number of u64's in the bitmap
+    fn bitmap_len(&self) -> usize {
+        let num_pages = self.pend.bits().saturating_sub(self.pbase.bits()) / PAGE_SIZE;
+        let mut num_elems = num_pages / u64::BITS as usize;
+        if num_elems % u64::BITS as usize != 0 {
+            num_elems += 1;
+        }
```
Freax13, line 156: Is the check correct? I'd expect it to be `num_pages % u64::BITS as usize != 0`. Note that you may be able to use [`usize::div_ceil`](https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil) here (which was stabilized very recently).

00xc: Yep, nice catch, fixed in the new version with `div_ceil()`

Freax13 (approved):

00xc:
@joergroedel I think this is ready to be merged

joergroedel:
@00xc This breaks `make test-in-svsm`. Can you please have a look?

joergroedel:
Bisection points at

```
commit 9e5d903b2d940bf2e81cfbf305582af7e35591fd (HEAD)
Author: Carlos López <carlos.lopez@suse.com>
Date:   Tue Oct 17 14:35:44 2023 +0200

    mm/validate: use usize instead of isize to index into the bitmap
    
    The offsets are always positive from the start of the bitmap, so use
    an unsigned type.
    
    Signed-off-by: Carlos López <carlos.lopez@suse.com>
```

00xc:
> 9e5d903b2d940bf2e81cfbf305582af7e35591fd

Only happens in debug builds, so I suspect some debug assertion is being triggered.

00xc:
I'm working on a bigger rework of the bitmap to increase safety.

00xc:
Turns out my assumption on 498b599221f0fa7b9f7eed073ebcfbea1f5690aa was wrong, things should work now. I'll push my rework of the bitmap to a separate PR.

PR #136, opened 2023-10-18
task, percpu: Use VMR to manage the task pagetable and VMKernelStack to allocate the task kernel stack

roy-hopkins:
Now that virtual memory ranges and page table entries can be managed using the new `VMR` and associated structs, update the Task structure implementation to manage the task pagetable using these new capabilities.

The task structure now includes a `VMR` instance that defines the virtual address range for the task. The kernel stack for the task is created using `VMKernelStack` and added to the task VMR. This is applied to the task pagetable during creation.

`PerCpu` already includes a per cpu VMR. This needs to be applied to the task when it is context switched onto the CPU. Currently this is done directly by copying the PerCpu level 3 index. This patch adss the function `populate_page_table()` to PerCpu which applies the PerCpu VMR to the pagetable instead.

### DIFF
--- a/src/cpu/percpu.rs
+++ b/src/cpu/percpu.rs
@@ -16,7 +16,7 @@ use crate::locking::{LockGuard, RWLock, SpinLock};
 use crate::mm::alloc::{allocate_page, allocate_zeroed_page};
 use crate::mm::pagetable::{get_init_pgtable_locked, PTEntryFlags, PageTable, PageTableRef};
 use crate::mm::virtualrange::VirtualRange;
-use crate::mm::vm::{Mapping, VMKernelStack, VMPhysMem, VMReserved, VMR};
+use crate::mm::vm::{Mapping, VMKernelStack, VMPhysMem, VMRMapping, VMReserved, VMR};
 use crate::mm::{
     virt_to_phys, SVSM_PERCPU_BASE, SVSM_PERCPU_CAA_BASE, SVSM_PERCPU_END,
     SVSM_PERCPU_TEMP_BASE_2M, SVSM_PERCPU_TEMP_BASE_4K, SVSM_PERCPU_TEMP_END_2M,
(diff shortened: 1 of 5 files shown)


roy-hopkins:
You can see this being used in #64

src/task/tasks.rs
```diff
@@ -183,12 +170,12 @@ pub struct TaskContext {
...
     pub rsp: u64,
 
-    /// Information about the task stack
-    pub stack: TaskStack,
-
     /// Page table that is loaded when the task is scheduled
     pub page_table: SpinLock<PageTableRef>,
 
+    /// Task virtual memory range
+    vm_range: VMR,
```
joergroedel, line 177: This needs a more specific name, as it covers the kernel VMR. We will get another VMR for user-space in the same structure. Maybe `vm_kernel_range`?

src/cpu/percpu.rs
```diff
@@ -171,6 +171,29 @@ impl GuestVmsaRef {
...
+}
+
+impl Drop for PerCpuMapping {
+    fn drop(&mut self) {
+        this_cpu_mut()
+            .vm_range
+            .remove(self.0)
+            .expect("Failed to remove PerCpuMapping managed range");
+    }
+}
```
joergroedel, line 195: Maybe we can look into making this more generic to work with any VMR instance. This requires taking a reference to the VMR within the struct and lifetime annotations. But this is just a note, no requirement for this PR.

joergroedel (changes requested):
Looks good overall, thanks. I requested one change and made one suggestion for the mapping guard (but it is fine as-is for now).

src/task/tasks.rs
roy-hopkins: Renamed as suggested.

src/cpu/percpu.rs
roy-hopkins: I didn't really like that struct sitting there in percpu.rs so as suggested I've added the generic function using a couple of new commits. Please take a look.

joergroedel (approved):
Looks good now, thanks!

PR #143, opened 2023-10-26

Zildj1an:


00xc (commented):
Maybe the tests for each of the submodules should be in a separate commit?

Zildj1an:
> Maybe the tests for each of the submodules should be in a separate commit?

You got it @00xc, done :)

src/mm/address_space.rs
```diff
@@ -157,5 +157,68 @@ pub const SVSM_PERTASK_BASE: VirtAddr = virt_from_idx(PGTABLE_LVL3_IDX_PERTASK);
...
+
+    #[test]
+    #[cfg(not(target_os = "none"))]
+    fn test_phys_to_virt() {
+        let paddr = PhysAddr::new(0x4500);
+        let vaddr = phys_to_virt(paddr);
+
+        assert_eq!(vaddr, VirtAddr::new(0x4500));
+    }
+}
```
00xc, line 224: I think this will be problematic when running tests inside the SVSM. If a test runs before `test_init_kernel_mapping_info()`, it will access an uninitialized `KERNEL_MAPPING` (via `phys_to_virt()` or `virt_to_phys()`), leading to undefined behavior.

This does not happen when running tests regularly because `virt_to_phys()` and `phys_to_virt()` do not access `KERNEL_MAPPING` when `#[cfg(not(target_os = "none"))]` evaluates to true.

00xc: Actually I'm wrong but it's still not OK. Running tests inside the SVSM will initialize `KERNEL_MAPPING` (`svsm_start() -> mapping_info_init() -> init_kernel_mapping_info()`). With debug assertions enabled, this will panic in `KERNEL_MAPPING.init()` due to double initialization. With debug assertions disabled my guess is this will corrupt memory or panic in `virt_to_phys()` / `phys_to_virt()` at some point.

joergroedel: This needs to be sorted out before it can be merged.

Zildj1an: Fixing in a new PR

Zildj1an:
Closing, look at https://github.com/coconut-svsm/svsm/pull/170

PR #145, opened 2023-10-27
fs: Add VMFileMapping to map ramfs file contents to virtual memory

roy-hopkins:
In order to be able to provide the ability to load and map ELF files into a task virtual memory space, we need the ability to take RamFS files and create virtual memory ranges to map the file physical pages to virtual memory addresses.

This PR uses the new `VirtualRange` trait and `VMR` ranges to implement a range that takes ownership of a file handle and allows a view of the file to be created as a virtual memory range and added to any `VMR` such as within the current task or the PerCpu range.

The goal of the implementation is to keep the file system contents as immutable, providing read-only access to the backing pages and copy-on-write access for writable access.

Multiple `VMFileMapping` ranges can be created over a single file by opening multiple file handles on the same file.

### DIFF
--- a/src/mm/vm/mapping/file_mapping.rs
+++ b/src/mm/vm/mapping/file_mapping.rs
@@ -6,11 +6,16 @@
 
 extern crate alloc;
 
+use core::slice::from_raw_parts_mut;
+
+use alloc::sync::Arc;
 use alloc::vec::Vec;
 
-use super::{RawAllocMapping, VirtualMapping};
+use super::{Mapping, RawAllocMapping, VMPageFaultResolution, VMPhysMem, VirtualMapping};
+use crate::address::Address;
 use crate::error::SvsmError;
 use crate::fs::FileHandle;
+use crate::mm::vm::VMR;
 use crate::mm::PageRef;
 use crate::mm::{pagetable::PageTable, PAGE_SIZE};
 use crate::types::PAGE_SHIFT;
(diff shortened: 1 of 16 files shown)


src/mm/vm/mapping/file_mapping.rs
```diff
@@ -0,0 +1,199 @@
...
+            let write_addr = write_copy.map(offset);
+            if write_addr.is_some() {
+                write_addr
+            } else {
+                self.file.mapping(self.offset + offset)
+            }
+        } else {
+            self.file.mapping(self.offset + offset)
+        }
+    }
```
Freax13, line 143: AFAICT this function and none of the function it calling end up calling `get_file_page`, so there reference count for the page is never increased. Am I missing something here?

src/mm/vm/mapping/rawalloc.rs
```diff
@@ -86,4 +106,22 @@ impl RawAllocMapping {
...
+    /// # Returns
+    ///
+    /// 'true' if the page containing the offset has been allocated
+    /// otherwise 'false'.
+    pub fn present(&self, offset: usize) -> bool {
+        let pfn = offset >> PAGE_SHIFT;
+        self.pages
+            .get(pfn)
+            .and_then(|r| r.as_ref().map(|_| ()))
+            .is_some()
```
Freax13, line 125:
```suggestion
        self.pages.get(pfn).and_then(|r| r.as_ref()).is_some()
```

src/mm/vm/mapping/file_mapping.rs
roy-hopkins: I was relying on the fact that the `FileHandle` is owned by this struct which in turn ultimately holds a reference to the pages themselves. However on reflection, this ties the implementation of the filesystem API and ramfs too closely to this struct so I've now modified this to hold a `PageRef` clone for each of the pages in the region.

roy-hopkins:
Added unit tests plus a couple of fixes for the issues they uncovered.

joergroedel:
@00xc Can you please have a look at the Copy and Debug traits for MM code? I temporarily disabled the deny statements to make this PR work in the upstream branch. Thanks!

PR #148, opened 2023-11-02
fuzzing: add allocator harnesses

00xc:
Add two fuzzing harnesses. One explicitly targets page allocation, while the other one targets the higher level `SvsmAllocator` API.

### DIFF
--- a/fuzz/fuzz_targets/alloc.rs
+++ b/fuzz/fuzz_targets/alloc.rs
@@ -1,0 +1,158 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2023 SUSE LLC
+//
+// Author: Carlos López <carlos.lopez@suse.com>
+
+#![no_main]
+
+use arbitrary::Arbitrary;
+use core::alloc::{GlobalAlloc, Layout, LayoutError};
+use core::num::NonZeroUsize;
+use libfuzzer_sys::fuzz_target;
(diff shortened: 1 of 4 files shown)


fuzz/fuzz_targets/alloc.rs
```diff
@@ -0,0 +1,154 @@
...
+                    continue;
+                };
+
+                let ptr = unsafe { heap.realloc(ptr, layout, new_layout) };
+                if !ptr.is_null() {
+                    ptrs.push((ptr, new_layout));
+                }
+            }
+        }
+    }
```
vsntk18, line 153: Did you forget to free the memory chunks that might still be left in the _ptrs_ vector after exiting for loop?

00xc: Added it, thanks

fuzz/fuzz_targets/page_alloc.rs
```diff
@@ -0,0 +1,164 @@
...
+                    pages.push(page);
+                }
+            }
+            Action::AllocateZeroed => {
+                if let Ok(page) = allocate_zeroed_page() {
+                    pages.push(page);
+                    inited.insert(page);
+                }
+            }
+            Action::AllocateFile => {
```
vsntk18, line 111: _get_file_page_() and _put_file_page()_ can also be called for a wider coverage?

vsntk18: Alright.  got it.

PR #150, opened 2023-11-03
Add support for running tests inside stage2

AdamCDunlap:
In the same way that support was added for running tests inside the svsm kernel, add tests that run inside stage2. This is useful for stage2-specific componenets such as the stage 2 #VC handler.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,7 +32,8 @@ jobs:
       # ld to work, so build all the objects without performing the
       # final linking step.
       - name: Build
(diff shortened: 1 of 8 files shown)


joergroedel:
Please fix the failing test case. This also breaks the test-in-svsm target.

AdamCDunlap:
Thanks for taking a look!

> Please fix the failing test case.

I updated the github action to build the stage2.bin target since it's required to do that before building stage1. I think it'd be better if this was not necessary, but this is how it was done when the test-kernel target was added and I'm not sure how to do it better.

> This also breaks the test-in-svsm target.

Fixed in the same way. I missed that because it only breaks if you make clean first.

Also rebased and fixed merge conflicts

PR #153, opened 2023-11-09
utils: Add unit tests for util.rs

Zildj1an:
Add unit tests to check correct behavior of auxiliary functions in util.rs.

### DIFF
--- a/src/utils/util.rs
+++ b/src/utils/util.rs
@@ -46,3 +46,46 @@ pub fn zero_mem_region(start: VirtAddr, end: VirtAddr) {
     // Zero region
     unsafe { start.as_mut_ptr::<u8>().write_bytes(0, size) }
 }
+
+#[cfg(test)]
+mod tests {
+
+    use crate::utils::util::*;
+
+    #[test]
+    fn test_mem_utils() {
+        // Align up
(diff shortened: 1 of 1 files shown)


src/utils/util.rs
```diff
@@ -46,3 +46,46 @@ pub fn zero_mem_region(start: VirtAddr, end: VirtAddr) {
...
+        assert!(overlap(1, 5, 3, 6));
+        assert!(overlap(0, 10, 5, 15));
+        assert!(!overlap(1, 5, 6, 8));
+    }
+
+    #[test]
+    fn test_zero_mem_region() {
+        let mut data: [u8; 10] = [1; 10];
+        let start = VirtAddr::new(&mut data[0] as *mut u8 as usize);
+        let end = start + 10;
```
00xc, line 83:
```suggestion
        let start = VirtAddr::from(data.as_mut_ptr());
        let end = start + core::mem::size_of_val(&data);
```

00xc (changes requested):
Looks good overall, just one suggestion below.

src/utils/util.rs
joergroedel: @Zildj1an Please fix this with a follow-on PR.

PR #157, opened 2023-11-22
cpu/vc,idt: drop debug support in stage2

p4zuu:
The #VC handler added in stage 2 breaks the gdb-enabled builds due to a too big stage 2 binary.

Breakpoints handling in stage 2 is breaking, and useless so we can drop the handler function calls so that gdbstub code is not included in stage 2 binary.


roy-hopkins (approved):
LGTM - I tested the changes in my environment and this fixed the issue, still allowing the gdbstub to work where expected.

PR #159, opened 2023-11-22
debug: remove handle_db_exception

p4zuu:
handle_db_exception was only used in the SVSM #VC handler. This function was calling handle_debug_exception under the hood.

We can remove handle_db_exception and integreate its logic into handle_debug_exception so that it remains only 1 debug exception handling function.

### DIFF
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -15,7 +15,7 @@ pub mod svsm_gdbstub {
 
     use crate::address::{Address, VirtAddr};
     use crate::cpu::control_regs::read_cr3;
-    use crate::cpu::idt::common::{X86ExceptionContext, BP_VECTOR};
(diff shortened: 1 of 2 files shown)


src/debug/gdbstub.rs
```diff
@@ -72,15 +72,12 @@ pub mod svsm_gdbstub {
     pub fn handle_debug_exception(ctx: &mut X86ExceptionContext, exception: usize) {
         let tp = match exception {
             BP_VECTOR => ExceptionType::SwBreakpoint,
+            VC_VECTOR => ExceptionType::Debug,
             _ => ExceptionType::PageFault,
         };
```
00xc, line 77: While you are cleaning things up, could you implement `From<usize> for ExceptionType` and do this match case there?

PR #166, opened 2023-11-24
Implement From<usize> for ExceptionType

p4zuu:
Suggested by @00xc in #159

### DIFF
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -716,7 +716,5 @@ pub mod svsm_gdbstub {
 
     pub fn handle_debug_exception(_ctx: &mut X86ExceptionContext, _exception: usize) {}
 
-    pub fn handle_db_exception(_ctx: &mut X86ExceptionContext) {}
-
     pub fn debug_break() {}
 }
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -62,23 +62,25 @@ pub mod svsm_gdbstub {
         Ok(())
     }
 
-    #[derive(PartialEq, Eq)]
+    #[derive(PartialEq, Eq, Debug)]
     enum ExceptionType {
         Debug,
         SwBreakpoint,
         PageFault,
     }
 
-    pub fn handle_debug_exception(ctx: &mut X86ExceptionContext, exception: usize) {
-        let tp = match exception {
-            BP_VECTOR => ExceptionType::SwBreakpoint,
-            VC_VECTOR => ExceptionType::Debug,
-            _ => ExceptionType::PageFault,
-        };
-        handle_exception(ctx, tp);
+    impl From<usize> for ExceptionType {
+        fn from(value: usize) -> Self {
+            match value {
+                BP_VECTOR => ExceptionType::SwBreakpoint,
+                VC_VECTOR => ExceptionType::Debug,
+                _ => ExceptionType::PageFault,
+            }
+        }
     }
 
-    fn handle_exception(ctx: &mut X86ExceptionContext, exception_type: ExceptionType) {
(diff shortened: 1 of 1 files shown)


src/debug/gdbstub.rs
```diff
@@ -62,20 +62,25 @@ pub mod svsm_gdbstub {
...
+    }
+
     pub fn handle_debug_exception(ctx: &mut X86ExceptionContext, exception: usize) {
-        let tp = match exception {
-            BP_VECTOR => ExceptionType::SwBreakpoint,
-            VC_VECTOR => ExceptionType::Debug,
-            _ => ExceptionType::PageFault,
-        };
-        handle_exception(ctx, tp);
+        handle_exception(ctx, ExceptionType::from(exception));
```
00xc, line 83: And now you can inline this function in the callers in my opinion, since it's only one line :)

p4zuu: You mean convert the usize to ExceptionType in the callers and move `handle_exception()`'s body to `handle_debug_exception()` ?

00xc: Yep, replace `handle_debug_exception()` calls with `handle_exception(ctx, ExceptionType::from(...))`. If you think it adds any clarity, of course.

p4zuu: Since `handle_debug_exception()` is the current public API, I would prefer keeping this function name for the callers (`handle_debug_exception(ctx, ExceptionType::from(...)`) and move `handle_exception()` body in `handle_debug_exception()` :)

00xc: Yep that also makes sense

p4zuu: So actually, doing this breaks what the module is intended for:
```rust
// For release builds this module should not be compiled into the
// binary. See the bottom of this file for placeholders that are
// used when the gdb stub is disabled.
```
Calling `Exception::from()` in the callers breaks the sentence above, since `ExceptionType` will have to be included in release builds. We could move `ExceptionType` into another module that doesn't depend on the build type, but I don't know if it's worth it

00xc: Okay so just inline `handle_exception()` into `handle_debug_exception()` as you suggested, no? That would keep the public API consistent.

00xc: So something like:

```diff
diff --git a/src/debug/gdbstub.rs b/src/debug/gdbstub.rs
index 181ee0b..8c5e84c 100644
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -69,16 +69,18 @@ pub mod svsm_gdbstub {
         PageFault,
     }
 
-    pub fn handle_debug_exception(ctx: &mut X86ExceptionContext, exception: usize) {
-        let tp = match exception {
-            BP_VECTOR => ExceptionType::SwBreakpoint,
-            VC_VECTOR => ExceptionType::Debug,
-            _ => ExceptionType::PageFault,
-        };
-        handle_exception(ctx, tp);
+       impl From<usize> for ExceptionType {
+        fn from(value: usize) -> Self {
+            match value {
+                BP_VECTOR => ExceptionType::SwBreakpoint,
+                VC_VECTOR => ExceptionType::Debug,
+                _ => ExceptionType::PageFault,
+            }
+        }
     }
 
-    fn handle_exception(ctx: &mut X86ExceptionContext, exception_type: ExceptionType) {
+    pub fn handle_debug_exception(ctx: &mut X86ExceptionContext, exception: usize) {
+        let exception_type = ExceptionType::from(exception);
         let id = this_cpu().runqueue().lock_read().current_task_id();
         let mut task_ctx = TaskContext {
             regs: X86GeneralRegs {
```

p4zuu: Looks good to me :)

00xc (approved):

PR #168, opened 2023-11-29
Control startup behavior based on launch configuration

msft-jlange:
Defines a new structure that encapsulates launch configuration into a configuration object and controls the behavior of the launch based on the encapsulated configuration

### DIFF
--- a/src/svsm.rs
+++ b/src/svsm.rs
@@ -15,8 +15,8 @@ use svsm::fw_meta::{parse_fw_meta_data, print_fw_meta, validate_fw_memory, SevFW
 use core::arch::global_asm;
 use core::panic::PanicInfo;
 use core::slice;
-use svsm::acpi::tables::load_acpi_cpu_info;
 use svsm::address::{PhysAddr, VirtAddr};
+use svsm::config::SvsmConfig;
 use svsm::console::{init_console, install_console_logger, WRITER};
 use svsm::cpu::control_regs::{cr0_init, cr4_init};
 use svsm::cpu::cpuid::{dump_cpuid_table, register_cpuid_table, SnpCpuidTable};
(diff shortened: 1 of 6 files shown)


src/svsm.rs
```diff
@@ -449,31 +449,43 @@ pub extern "C" fn svsm_main() {
...
-    let fw_meta = map_and_parse_fw_meta()
-        .unwrap_or_else(|e| panic!("Failed to parse FW SEV meta-data: {:#?}", e));
+    let fw_metadata: Option<SevFWMetaData>;
+    if config.should_launch_fw() {
+        fw_metadata = Some(
+            map_and_parse_fw_meta()
+            .unwrap_or_else(|e| panic!("Failed to parse FW SEV meta-data: {:#?}", e)));
+    } else {
+        fw_metadata = None;
+    }
```
00xc, line 459: The `fw_metadata` can be constructed directly:

```suggestion
    let fw_metadata = if config.should_launch_fw() {
        Some(
            map_and_parse_fw_meta()
                .unwrap_or_else(|e| panic!("Failed to parse FW SEV meta-data: {:#?}", e)),
        )
    } else {
        None
    };
```

Or even better:

```suggestion
    let fw_metadata = config.should_launch_fw().then(|| {
        map_and_parse_fw_meta()
            .unwrap_or_else(|e| panic!("Failed to parse FW SEV meta-data: {:#?}", e))
    });
```

00xc (commented):
Overall looks good to me, just left one suggestion below. Also, please sign your commits via `git commit -s` and run your changes through `cargo fmt` and `cargo clippy`. CI should have warned you about these but it's disabled because it's your first contribution.

msft-jlange:
Thanks for the feedback!  I’m going to go with the first of your two proposed options, since the essence of any good code is readability, and the second option uses esoteric Rust syntax that I personally find much less readable (plus the second option will be much more difficult to integrate with future changes that integrate a different option provider).

-Jon

00xc (commented):
I'm still seeing some clippy warnings (`cargo clippy`). Would you mind fixing those as well? Thanks!

msft-jlange:
Strange - I did run cargo clippy and it didn't give me any errors.  And now that you've prompted me to run it again, it did generate warnings.  I've fixed them and updated the PR.  Thanks for being patient with me!

-Jon

PR #169, opened 2023-11-29
mm: replace recently unstable bitmap_len()

Zildj1an:
The usage of the unstable library feature 'int_roundings' would trigger a compilation error (See Rust issue #88581). Replace it with a standard integer division.

This is the error:

```
error[E0658]: use of unstable library feature 'int_roundings'
   --> src/mm/validate.rs:151:19
    |
151 |         num_pages.div_ceil(u64::BITS as usize)
    |                   ^^^^^^^^
    |
    = note: see issue #88581 <https://github.com/rust-lang/rust/issues/88581> for more information

```

Fixes: 2b0162d6 (mm/validate: migrate all pages in bitmap)

### DIFF
--- a/src/mm/validate.rs
+++ b/src/mm/validate.rs
@@ -148,7 +148,12 @@ impl ValidBitmap {
     /// The number of u64's in the bitmap
     fn bitmap_len(&self) -> usize {
         let num_pages = self.region.len() / PAGE_SIZE;
-        num_pages.div_ceil(u64::BITS as usize)
+        let additional_u64 = if self.region.len() % PAGE_SIZE != 0 {
+            1
+        } else {
+            0
+        };
+        num_pages + additional_u64
     }
 
     fn migrate(&mut self, new_bitmap: *mut u64) {


00xc:
`div_ceil` should be stable since Rust 1.73. Have you updated your toolchain?

Zildj1an:
> `div_ceil` should be stable since Rust 1.73. Have you updated your toolchain?

You're absolutely correct! Doing `rustup update` solved it. I have removed the Fixes tag and updated the commit message; even though this works for latest toolchains I think we should still apply this change, given that we don't really need this (recently unstable) utility to perform the operation at hand.

src/mm/validate.rs
```diff
@@ -148,7 +148,12 @@ impl ValidBitmap {
     /// The number of u64's in the bitmap
     fn bitmap_len(&self) -> usize {
         let num_pages = self.region.len() / PAGE_SIZE;
-        num_pages.div_ceil(u64::BITS as usize)
+        let additional_u64 = if self.region.len() % PAGE_SIZE != 0 {
+            1
+        } else {
+            0
+        };
+        num_pages + additional_u64
```
stefano-garzarella, line 156: Not a strong opinion, but eventually we can avoid `additional_u64` and do something like this:
```suggestion
        if self.region.len() % PAGE_SIZE != 0 {
            num_pages + 1
        } else {
            num_pages
        }
```

PR #170, opened 2023-11-29

Zildj1an:

### DIFF
--- a/src/mm/address_space.rs
+++ b/src/mm/address_space.rs
@@ -153,3 +153,81 @@ pub const SVSM_PERTASK_END: VirtAddr = SVSM_PERTASK_BASE.const_add(SIZE_LEVEL3);
 
 /// Kernel stack for a task
 pub const SVSM_PERTASK_STACK_BASE: VirtAddr = SVSM_PERTASK_BASE;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::locking::SpinLock;
+
+    static KERNEL_MAPPING_TEST: ImmutAfterInitCell<KernelMapping> = ImmutAfterInitCell::uninit();
+    static INITIALIZED: SpinLock<bool> = SpinLock::new(false);
+
+    #[test]
+    #[cfg_attr(test_in_svsm, ignore = "Offline testing")]
+    fn init_km_testing() {
+        let mut initialized = INITIALIZED.lock();
+        if *initialized {
+            return;
+        }
+        KERNEL_MAPPING_TEST
+            .init(&KernelMapping {
+                virt_start: VirtAddr::new(0x1000),
+                virt_end: VirtAddr::new(0x2000),
+                phys_start: PhysAddr::new(0x3000),
+            })
+            .unwrap();
+        *initialized = true;
+    }
+
+    #[test]
+    #[cfg_attr(test_in_svsm, ignore = "Offline testing")]
+    fn test_init_kernel_mapping_info() {
+        init_km_testing();
+
+        let km = &KERNEL_MAPPING_TEST;
+
+        assert_eq!(km.virt_start, VirtAddr::new(0x1000));
+        assert_eq!(km.virt_end, VirtAddr::new(0x2000));
+        assert_eq!(km.phys_start, PhysAddr::new(0x3000));
+    }
+
+    #[test]
+    #[cfg(target_os = "none")]
+    #[cfg_attr(test_in_svsm, ignore = "Offline testing")]
+    fn test_virt_to_phys() {
+        let vaddr = VirtAddr::new(0x1500);
+        let paddr = virt_to_phys(vaddr);
+
+        assert_eq!(paddr, PhysAddr::new(0x4500));
+    }
+
+    #[test]
+    #[cfg(not(target_os = "none"))]
+    #[cfg_attr(test_in_svsm, ignore = "Offline testing")]
+    fn test_virt_to_phys() {
+        let vaddr = VirtAddr::new(0x1500);
+        let paddr = virt_to_phys(vaddr);
+
+        assert_eq!(paddr, PhysAddr::new(0x1500));
+    }
+
+    #[test]
+    #[cfg(target_os = "none")]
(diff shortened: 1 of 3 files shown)


src/mm/guestmem.rs
```diff
@@ -222,3 +222,31 @@ impl<T: Sized + Copy> GuestPtr<T> {
         unsafe { GuestPtr::from_ptr(self.ptr.offset(count)) }
     }
 }
+
+mod tests {
+
+    #[test]
+    fn test_read_u8_valid_address() {
```
00xc, line 229: This ends up using inline assembly, so Miri cannot handle it.

```suggestion
    #[test]
    #[cfg_attr(miri, ignore)]
    fn test_read_u8_valid_address() {
```

src/mm/guestmem.rs
```diff
@@ -222,3 +222,31 @@ impl<T: Sized + Copy> GuestPtr<T> {
...
+        use crate::mm::guestmem::*;
+        // Create a region to read from
+        let test_buffer: [u8; 6] = [0; 6];
+        let test_address = &test_buffer as *const u8 as usize;
+
+        let result = read_u8(VirtAddr::new(test_address)).unwrap();
```
00xc, line 235: Nit: instead of using "as" casting, you can simply do:

```suggestion
        let test_address = VirtAddr::from(test_buffer.as_ptr());

        let result = read_u8(test_address).unwrap();
```

src/mm/guestmem.rs
```diff
@@ -222,3 +222,31 @@ impl<T: Sized + Copy> GuestPtr<T> {
...
+
+        assert_eq!(result, test_buffer[0]);
+    }
+
+    #[test]
+    fn test_write_u8_valid_address() {
```
00xc, line 241:
```suggestion
    #[test]
    #[cfg_attr(miri, ignore)]
    fn test_write_u8_valid_address() {
```

src/mm/guestmem.rs
```diff
@@ -222,3 +222,31 @@ impl<T: Sized + Copy> GuestPtr<T> {
...
+        use crate::mm::guestmem::*;
+        // Create a mutable region we can write into
+        let mut test_buffer: [u8; 6] = [0; 6];
+        let test_address = &mut test_buffer as *mut u8 as usize;
+        let data_to_write = 0x42;
+
+        write_u8(VirtAddr::new(test_address), data_to_write).unwrap();
```
00xc, line 248: Nit: same here, instead of 'as' casting:

```suggestion
        let test_address = VirtAddr::from(test_buffer.as_mut_ptr());
        let data_to_write = 0x42;

        write_u8(test_address, data_to_write).unwrap();
```

src/mm/address_space.rs
```diff
@@ -153,3 +153,67 @@ pub const SVSM_PERTASK_END: VirtAddr = SVSM_PERTASK_BASE.const_add(SIZE_LEVEL3);
...
+    }
+
+    #[test]
+    #[cfg(target_os = "none")]
+    fn test_virt_to_phys() {
+        let vaddr = VirtAddr::new(0x1500);
+        let paddr = virt_to_phys(vaddr);
+
+        assert_eq!(paddr, PhysAddr::new(0x4500));
+    }
```
00xc, line 191: This test fails inside the SVSM when calling `phys_to_virt()` in this panic:

https://github.com/coconut-svsm/svsm/blob/be10d1b5ef8f4365434238e916623f14b59359d5/src/mm/address_space.rs#L43-L47

I suspect some of the other tests might also fail.

00xc (changes requested):
At least one test is failing inside the SVSM, so we need that fixed at the minimum. I would also really want the Miri addition to go in, as I'm running Miri tests nightly.

Zildj1an:
> At least one test is failing inside the SVSM, so we need that fixed at the minimum. I would also really want the Miri addition to go in, as I'm running Miri tests nightly.

I foresee the complexity of these and future tests growing considerably if we want all to be able to run inside the SVSM. My intention when I wrote and added these tests was to run them offline. For example, I need to either implement new functions that use `KERNEL_MAPPING_TEST` instead, or I have to consider the ramifications of using `KERNEL_MAPPING` uninitialized, etc. So I'm wondering, Is there a way we can leave some tests for offline testing only?

00xc:
> I foresee the complexity of these and future tests growing considerably if we want all to be able to run inside the SVSM. My intention when I wrote and added these tests was to run them offline. For example, I need to either implement new functions that use `KERNEL_MAPPING_TEST` instead, or I have to consider the ramifications of using `KERNEL_MAPPING` uninitialized, etc. So I'm wondering, Is there a way we can leave some tests for offline testing only?

Yes, there are already some tests that are ignored inside the SVSM, for example in `mm/alloc.rs`. You can do that with:

```rust
#[test]
#[cfg_attr(test_in_svsm, ignore = "<REASON>")]
fn my_test() {}
```

I think you can also omit the reason.

src/mm/guestmem.rs
Zildj1an: Not sure if this is really needed anymore if we do `#[cfg_attr(test_in_svsm, ignore = "<REASON>")]` but including just in case

src/mm/guestmem.rs
Zildj1an: Good! thanks

Zildj1an:
> > I foresee the complexity of these and future tests growing considerably if we want all to be able to run inside the SVSM. My intention when I wrote and added these tests was to run them offline. For example, I need to either implement new functions that use `KERNEL_MAPPING_TEST` instead, or I have to consider the ramifications of using `KERNEL_MAPPING` uninitialized, etc. So I'm wondering, Is there a way we can leave some tests for offline testing only?
> 
> Yes, there are already some tests that are ignored inside the SVSM, for example in `mm/alloc.rs`. You can do that with:
> 
> ```rust
> #[test]
> #[cfg_attr(test_in_svsm, ignore = "<REASON>")]
> fn my_test() {}
> ```
> 
> I think you can also omit the reason.

Thanks, I'm doing this, opening new clean PR

PR #172, opened 2023-11-30

Zildj1an:

### DIFF
--- a/src/mm/memory.rs
+++ b/src/mm/memory.rs
@@ -102,3 +102,25 @@ pub fn writable_phys_addr(paddr: PhysAddr) -> bool {
 
     valid_phys_address(paddr)
 }
+
+#[cfg(test)]
+#[cfg_attr(test_in_svsm, ignore = "Offline testing")]
+mod tests {
+    use super::*;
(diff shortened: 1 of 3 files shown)


src/mm/memory.rs
```diff
@@ -102,3 +102,25 @@ pub fn writable_phys_addr(paddr: PhysAddr) -> bool {
 
     valid_phys_address(paddr)
 }
+
+#[cfg(test)]
+#[cfg_attr(test_in_svsm, ignore = "Offline testing")]
```
00xc, line 107: Please remove this. Everything else looks good.

```
warning: `#[ignore]` only has an effect on functions
   --> src/mm/memory.rs:107:26
    |
107 | #[cfg_attr(test_in_svsm, ignore = "Offline testing")]
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_attributes)]` on by default
```

Zildj1an: Sounds good!

00xc (approved):

PR #177, opened 2023-12-08
fuzz: add bitmap_allocator harness.

vsntk18:
Added a fuzz harness for bitmap_allocator.  This might not be interesting from a security perspective, but more  from stress testing perspective.

### DIFF
--- a/src/utils/bitmap_allocator.rs
+++ b/src/utils/bitmap_allocator.rs
@@ -18,6 +18,10 @@ pub trait BitmapAllocator {
     fn empty(&self) -> bool;
     fn capacity(&self) -> usize;
     fn used(&self) -> usize;
+    #[cfg(fuzzing)]
+    fn max_align(&self) -> usize {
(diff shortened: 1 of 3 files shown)


src/utils/bitmap_allocator.rs
```diff
@@ -171,6 +186,7 @@ impl<T: BitmapAllocator + Debug> BitmapAllocator for BitmapAllocatorTree<T> {
 fn alloc_aligned(ba: &mut impl BitmapAllocator, entries: usize, align: usize) -> Option<usize> {
     // Iterate through the bitmap checking on each alignment boundary
     // for a free range of the requested size
+    assert!((align as u32) < ba.capacity().ilog2());
```
joergroedel, line 189: I am wondering whether this should be an assert or a `return None`?

vsntk18: Right.   Might as well return a _None_ value.

vsntk18: updated PR with this small change.

PR #180, opened 2023-12-11
Acquire the choice of debug serial port from the IGVM parameter block

msft-jlange:
Permit the IGVM configuration to specify a debug output port other than COM1.  This is specified at the time the IGVM file is built and is not selectable by the host loader.

### DIFF
--- a/src/svsm.rs
+++ b/src/svsm.rs
@@ -42,7 +42,6 @@ use svsm::mm::virtualrange::virt_log_usage;
 use svsm::mm::{init_kernel_mapping_info, PerCPUPageMappingGuard, SIZE_1G};
 use svsm::requests::{request_loop, update_mappings};
 use svsm::serial::SerialPort;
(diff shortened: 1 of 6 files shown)


src/svsm.rs
```diff
@@ -334,6 +330,10 @@ pub extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: usize) {
     load_gdt();
     early_idt_init();
 
+    // Capture the debug serial port befure the launch info disappears from
```
stefano-garzarella, line 333: s/befure/before

stefano-garzarella (approved):
LGTM, there is just a little typo in a comment.

src/svsm.rs
msft-jlange: Thanks for catching that!  This has already merged but I'll fix it with my next change.

PR #181, opened 2023-12-13
IGVM: Add support for launching OVMF firmware when using an IGVM file for guest configuration

roy-hopkins:
This changes the IGVM parameter block to replace the flag that indicates whether to boot OVMF with two parameters that define the GPA range of the OVMF region. 

The same constraints on OVMF location are currently still true - the OVMF pages must be aligned up to the 4GB boundary therefore the IGVM builder must ensure this alignment for OVMF to be booted correctly.

### DIFF
--- a/igvm_params/src/lib.rs
+++ b/igvm_params/src/lib.rs
@@ -55,11 +55,16 @@ pub struct IgvmParamBlock {
     /// The port number of the serial port to use for debugging.
     pub debug_serial_port: u16,
 
-    /// A flag indicating whether the kernel should proceed with the flow
-    /// to launch guest firmware once kernel initialization is complete.
-    pub launch_fw: u8,
+    _reserved: u16,
+    /// The guest physical address of the start of the guest firmware. The
+    /// permissions on the pages in the firmware range are adjusted to the guest
+    /// VMPL. If this field is zero then no firmware is launched after
+    /// initialization is complete.
+    pub fw_start: u32,
 
-    _reserved: u8,
+    /// The number of pages of guest firmware. If the firmware size is zero then
+    /// no firmware is launched after initialization is complete.
+    pub fw_size: u32,
 
     /// The amount of space that must be reserved at the base of the kernel
     /// memory region (e.g. for VMSA contents).
--- a/igvm_params/src/lib.rs
+++ b/igvm_params/src/lib.rs
@@ -56,6 +56,7 @@ pub struct IgvmParamBlock {
     pub debug_serial_port: u16,
 
     _reserved: u16,
+
     /// The guest physical address of the start of the guest firmware. The
     /// permissions on the pages in the firmware range are adjusted to the guest
     /// VMPL. If this field is zero then no firmware is launched after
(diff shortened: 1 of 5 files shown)


igvm_params/src/lib.rs
msft-jlange, line 1: Your changes break 64-bit alignment of the kernel base field.  Please reinsert padding to keep that field aligned.

src/svsm.rs
```diff
@@ -473,8 +470,7 @@ pub extern "C" fn svsm_main() {
         if let Err(e) = copy_tables_to_fw(fw_meta) {
```
msft-jlange, line 470: Your PR moves the location of the firmware to the IGVM file (which is good) but there is still a hardcoded assumption that the firmware wants CPUID/secrets information in a particular place (see map_and_parse_fw_meta).  Shouldn't this be abstracted in the IGVM as well so that the SVSM is not making assumptions about OVMF-specific layout?

roy-hopkins: Yes. I mentioned in the commit and the description of the PR that although the firmware is in the IGVM file the builder must currently align the firmware to 4GB. This PR allows me to get back to the point where we can boot firmware from SVSM when using an IGVM file. At this stage I think it is ok to assume that OVMF places it's configuration at the top of its GPA range as this is well defined.

The future plan is to allow for IGVM to specify a different location for the firmware, and at that point we would need to describe the layout via parameters to SVSM. But this would require changes to OVMF which I think will be the subject of future work.

Do you think this approach is acceptable?

msft-jlange: Sorry, I meant something slightly different.  The current implementation of `map_and_parse_fw_meta` has hard-coded logic which assumes that the firmware metadata must be at 4 GB less 4 KB.  That may be great for OVMF today but won't be great for any other firmware that might be packaged into an IGVM file.  Since you're making IGVM flexible enough to specify the base/size of the firmware, it seems like you should also have the IGVM file include the location of the metadata information (if it exists at all) so that `map_and_parse_fw_meta` can do its work to match what the IGVM specifies.  Note that none of this has anything to with changing OVMF; it's just about making the SVSM self-consistent in the way it looks for FW-specific data (i.e., dictate all references from the IGVM and not from hard-coded constants).

I can always add that in a future PR if this is too much of a distraction from completing your current PR.  It shouldn't be much work regardless of which one of us does it.

roy-hopkins: As far as I'm aware, the format of the metadata itself is fairly unique to OVMF, including the location of where to find it. But I see your point - it is possible for the IGVM file to locate it elsewhere even if OVMF won't boot in that case. The fw range check in SVSM currently ensures it aligns with 4GB so it won't allow the firmware to be located anywhere else, but that will change and it does make sense to add the new field to describe the GPA of the firmware metadata page. I'll add it to the parameters.

msft-jlange: This is better, but I foresee a version where we have firmware but where the firmware doesn't use metadata in this way.  In that case, I would expect the IGVM parameters to specify a base/size of the firmware but would specify a metadata address of zero.  However, your new version of the code appears to treat that as an error, because `get_fw_metadata` is called unconditionally and would return an error.  Can this be changed so that firmware that does not express a metadata location can just skip the step of preparing the metadata instead of generating an error.

roy-hopkins: Ok, done. Firmware metadata is now optional and firmware can be launched even if metadata is not provided.

msft-jlange: Thanks!

msft-jlange (approved):

src/fw_cfg.rs
```diff
@@ -217,7 +217,7 @@ impl<'a> FwCfg<'a> {
     // This needs to be &mut self to prevent iterator invalidation, where the caller
     // could do fw_cfg.select() while iterating. Having a mutable reference prevents
     // other references.
-    pub fn iter_flash_regions(&mut self) -> impl Iterator<Item = MemoryRegion<PhysAddr>> + '_ {
+    pub fn iter_flash_regions(&self) -> impl Iterator<Item = MemoryRegion<PhysAddr>> + '_ {
```
00xc, line 220: This change ignores the comment above the function. Not taking `&mut self` allows potential misbehavior. The comment should be updated or the change reverted.

PR #189, opened 2023-12-19
Enter the guest VMSA from a consistent place

msft-jlange:
Ensure that every entry to the guest VMPL occurs from the same point in the execution loop.  This will become important in the future when REFLECT_VC is enabled, and will be critical on non-SNP platforms that do not use simple GHCB calls for guest entry.

### DIFF
--- a/src/requests.rs
+++ b/src/requests.rs
@@ -72,13 +72,35 @@ fn request_loop_once(
 
 pub fn request_loop() {
     loop {
-        if update_mappings().is_err() {
-            log::debug!("No VMSA or CAA! Halting");
-            halt();
-            continue;
-        }
+        // Determine whether the guest is runnable.  If not, halt and wait for
+        // the guest to execute.  When halting, assume that the hypervisor
+        // will schedule the guest VMPL on its own.
+        let vmsa = if update_mappings().is_ok() {
+            let vmsa = this_cpu_mut().guest_vmsa();
 
-        let vmsa = this_cpu_mut().guest_vmsa();
+            // Make VMSA runnable again by setting EFER.SVME
+            vmsa.enable();
+
+            flush_tlb_global_sync();
+
+            this_cpu_mut()
+                .ghcb()
+                .run_vmpl(GUEST_VMPL as u64)
+                .expect("Failed to run guest VMPL");
+
+            vmsa
+        } else {
+            loop {
+                log::debug!("No VMSA or CAA! Halting");
+                halt();
+
+                if update_mappings().is_ok() {
+                    break;
+                }
+            }
+
+            this_cpu_mut().guest_vmsa()
+        };
 
         // Clear EFER.SVME in guest VMSA
         vmsa.disable();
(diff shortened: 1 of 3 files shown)


src/requests.rs
```diff
@@ -114,18 +124,5 @@ pub fn request_loop() {
...
         // Write back results
         params.write_back(vmsa);
-
-        // Make VMSA runable again by setting EFER.SVME
-        vmsa.enable();
-
-        flush_tlb_global_sync();
-
-        // Check if mappings still valid
-        if update_mappings().is_ok() {
```
joergroedel, line 124: I don't see the call to update_mappings() moved, why is it safe to remove it? Also, I think the commit message should explain the reasoning behind this change.

joergroedel: Right, just checked that again too. Thanks for the explanation.

joergroedel:
I did some testing again with this change and run into a boot issue with the Linux kernel. At boot, it dies with this message:

```
[    0.887221][    T2] ------------[ cut here ]------------^M
[    0.888472][    T2] Failed to validate address ffff888100a00000 ret=0xffffffea (-22)^M
[    0.888477][    T2] WARNING: CPU: 4 PID: 2 at arch/x86/kernel/sev-shared.c:1314 pvalidate_pages+0x1f3/0x430^M
```

Right after that the kernel calls the GHCB Terminate call to kill the VM. Debugging showed that the most likely reason for this is that the SVSM does not clear the `pending` bit in the CAA.

I have not yet figured out how that can happen, probably Linux and the SVSM disagree about the actual CAA location. Any ideas? It still boots fine without this change.

joergroedel:
I think I tracked it down. The problem is that there is no request processing between the `halt()` call and the call to `run_vmpl()`. Here are the steps:

1. SVSM boots up AP; AP has no guest VMSA and goes to HLT
2. SVSM on BSP launches firmware and that boots through to the Linux kernel
3. Linux kernel sets a VMSA for the AP via SVSM calls on BSP
4. Hypervisor starts executing the AP in guest VMPL (note that VMPL-0 is still at the HLT instruction)
5. AP does an SVSM call
6. SVSM on AP comes out of HLT and goes straight back to guest VMPL, without processing the request
7. Linux finds the pending bit still set and panics.

msft-jlange:
> I think I tracked it down. The problem is that there is no request processing between the `halt()` call and the call to `run_vmpl()`. Here are the steps:
> 
> 1. SVSM boots up AP; AP has no guest VMSA and goes to HLT
> 2. SVSM on BSP launches firmware and that boots through to the Linux kernel
> 3. Linux kernel sets a VMSA for the AP via SVSM calls on BSP
> 4. Hypervisor starts executing the AP in guest VMPL (note that VMPL-0 is still at the HLT instruction)
> 5. AP does an SVSM call
> 6. SVSM on AP comes out of HLT and goes straight back to guest VMPL, without processing the request
> 7. Linux finds the pending bit still set and panics.

Thanks for debugging this!  I understand the issue and should be able to provide a fix soon.

PR #195, opened 2023-12-20
Fix issues with IGVM boot

msft-jlange:
- Move the kernel ELF/fileystem so they do not collide with low memory ROMs
- Invalidate the kernel ELF/filesystem memory so it can be reused by OVMF

### DIFF
--- a/src/svsm_paging.rs
+++ b/src/svsm_paging.rs
@@ -15,7 +15,8 @@ use crate::mm::pagetable::{set_init_pgtable, PTEntryFlags, PageTable, PageTableR
 use crate::mm::PerCPUPageMappingGuard;
 use crate::sev::ghcb::PageStateChangeOp;
 use crate::sev::{pvalidate, PvalidateOp};
-use crate::types::{PageSize, PAGE_SIZE};
+use crate::types::PageSize;
+use crate::utils::MemoryRegion;
 use bootlib::kernel_launch::KernelLaunchInfo;
 
 struct IgvmParamInfo<'a> {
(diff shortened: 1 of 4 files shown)


src/svsm_paging.rs
```diff
@@ -94,29 +95,64 @@ pub fn init_page_table(launch_info: &KernelLaunchInfo, kernel_elf: &elf::Elf64Fi
...
+            kernel_elf_size.try_into().unwrap(),
+        );
+        invalidate_boot_memory_region(config, kernel_elf_region)?;
+
+        let kernel_fs_size = launch_info.kernel_fs_end - launch_info.kernel_fs_start;
+        let kernel_fs_region = MemoryRegion::new(
+            PhysAddr::new(launch_info.kernel_fs_start.try_into().unwrap()),
+            kernel_fs_size.try_into().unwrap(),
+        );
+        invalidate_boot_memory_region(config, kernel_fs_region)?;
```
roy-hopkins, line 154: This calls `invalidate_boot_memory_region()` even if the filesystem range is empty. If the config specifies that a page state change is required then this will result in a call to `ghcb().page_state_change()`. Maybe this should only call `invalidate_boot_memory_region()` if `kernel_fs_size` is > 0?

roy-hopkins (commented):
I tested this on my system with IGVM on QEMU and the SVSM launches without issue.

src/svsm_paging.rs
msft-jlange: Good catch!  I tested this on a system that doesn't require page state changes so I missed this.  I've update the PR to fix this.

roy-hopkins (approved):
Tested on QEMU with and without IGVM.

PR #197, opened 2024-01-02
Supply firmware metadata information via the IGVM parameter block

msft-jlange:
This change permits the IGVM file to specify firmware metadata that would normally be read out of QEMU firmware metadata.  This permits the IGVM file to specify metadata without external reference, and permits SVSM configuration of firmware metadata on platforms other than QEMU.

### DIFF
--- a/bootlib/src/igvm_params.rs
+++ b/bootlib/src/igvm_params.rs
@@ -42,7 +42,7 @@ pub struct IgvmParamBlock {
     /// The guest physical address of the CPUID page.
     pub cpuid_page: u32,
 
-    /// The guest physical address of the secrets page.
+    /// The guest physical address of the SVSM secrets page.
     pub secrets_page: u32,
 
     /// The port number of the serial port to use for debugging.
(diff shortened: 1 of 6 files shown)


roy-hopkins (approved):
Looks good and tested ok with non-IGVM configuration under QEMU.

With this change though, we now have three different ways to specify the firmware metadata:

1. When no IGVM configuration is provided, via OVMF metadata at a fixed address.
2. Via OVMF metadata located using an address passed in the IGVM configuration.
3. Via explicit configuration in the IGVM configuration.

It seems to make more sense to me to _only_ allow metadata to be provided using the third option. However, until we have a QEMU that fully implements IGVM we still need to maintain the first option. 

I think we should implement the parsing of the OVMF metadata in the IGVM builder and always use option 3 when using an IGVM file. This will require a couple more fields in the parameter block but I don't see that as a problem.

If we decide to do this then it can be done in a separate PR so we can get this one merged.

PR #199, opened 2024-01-04
IGVM: Embed firmware in IGVM file and parse OVMF metadata into IGVM parameter block

roy-hopkins:
This PR adds support for embedding firmware, particularly OVMF into an IGVM file using the IGVM builder. 

The Makefile has been updated to allow the OVMF file to be specified by the `FW_FILE` environment variable. This needs to be set to the `OVMF.fd` file (and not the separate `OVMF_CODE.fd` and `OVMF_VARS.fd` files). When provided, the firmware binary is parsed by the IGVM builder during the build to extract metadata which is then populated into the IGVM parameter block, passed to SVSM.

By providing the metadata in the IGVM parameter block, SVSM does not need to parse the OVMF metadata when using an IGVM file. The SVSM has been updated to reflect this by removing the `fw_metadata` IGVM parameter that specified the OVMF metadata page. In the future, if IGVM is solely used to configure the SVSM then the OVMF metadata parsing can be removed entirely from SVSM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,13 @@ KERNEL_ELF = "target/x86_64-unknown-none/${TARGET_PATH}/svsm"
 TEST_KERNEL_ELF = target/x86_64-unknown-none/${TARGET_PATH}/svsm-test
 FS_FILE ?= none
 
+FW_FILE ?= none
+ifneq ($(FW_FILE), none)
+BUILD_FW = --firmware ${FW_FILE}
+else
+BUILD_FW = 
+endif
+
 C_BIT_POS ?= 51
 
 STAGE1_OBJS = stage1/stage1.o stage1/reset.o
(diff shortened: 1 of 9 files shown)


igvmbld/ovmfmeta.c
```diff
@@ -0,0 +1,238 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2024 SUSE LLC
+//
+// Author: Joerg Roedel <jroedel@suse.de>
+// Author: Roy Hopkins <roy.hopkins@suse.com>
```
roy-hopkins, line 6: This code is mostly a copy of `utils/print-meta.c`, hence keeping Joerg as the author.

Makefile
```diff
@@ -19,6 +19,13 @@ KERNEL_ELF = "target/x86_64-unknown-none/${TARGET_PATH}/svsm"
 TEST_KERNEL_ELF = target/x86_64-unknown-none/${TARGET_PATH}/svsm-test
 FS_FILE ?= none
 
+FW_FILE ?= none
+ifneq ($(FW_FILE), none)
+BUILD_FW = --firmware ${FW_FILE} --fwtop 100000000
```
msft-jlange, line 24: It seems a little unusual to specify a hex address without the 0x prefix, but it's probably not significant enough to merit updating the parsing logic to recognize the number base.

igvmbld/igvmbld.c
```diff
@@ -688,6 +674,65 @@ int generate_igvm_file(const char *filename)
...
+        fprintf(stderr, "Failed to read firmware file: \"%s\"\n", fw_filename);
+        fclose(fp);
+        return 1;
+    }
+    fw_size = ftell(fp);
+    fclose(fp);
+
+    // The command line should have provided either the firmware base address
+    // or the firmware top address but not both.
+    if ((fw_base == 0) && (fw_top == 0))
```
msft-jlange, line 707: This is going to conflict with Hyper-V requirements.  Is there any way to make this check, and the subsequent address validation, specific to QEMU configurations?

igvmbld/igvmbld.c
```diff
@@ -984,6 +1086,24 @@ int main(int argc, const char *argv[])
...
+        if (!construct_file_data_object(fw_filename, fw_base))
+        {
+            return 1;
+        }
+
+        // If the firmware file is an OVMF binary then we can extract the OVMF
+        // metadata from it. If the firmware is not OVMF then this function has
+        // no effect.
+        parse_ovmf_metadata(fw_filename, igvm_parameter_block);
+        print_fw_metadata(igvm_parameter_block);
```
msft-jlange, line 1104: Perhaps gate this behind a -v or --verbose argument?

roy-hopkins:
@msft-jlange, on reflection following our email conversation it is probably best to hardcode the firmware base calculation for QEMU and just return an error at the moment if `--firmware` is specified for Hyper-V as I don't know how the base address will be calculated in this case. This means I can remove `--fwtop` and `--fwbase`.

I've pushed a new version with this change.

igvmbld/igvmbld.c
roy-hopkins: See comment below. I've now simplified this code by removing the configurable parameters.

msft-jlange (approved):

PR #207, opened 2024-01-10
Ensure all boot memory is invalidated when QEMU guest is configured with IGVM

roy-hopkins:
When booting using an IGVM configuration, the current IGVM builder places some of the prevalidated pages outside the first 640K of low memory. These need to be invalidated prior to booting the firmware. 

This PR moves some of the regions in the IGVM builder and detects and invalidates the relevant sections in the SVSM when using IGVM.

Note that this PR is based on a branch which already includes the commit under separate review #206. Please ignore commit d6b606cfb1e283d3973b357f450a3a6a30fa4a41 when reviewing this PR.

### DIFF
--- a/bootlib/src/kernel_launch.rs
+++ b/bootlib/src/kernel_launch.rs
@@ -20,6 +20,8 @@ pub struct KernelLaunchInfo {
     pub kernel_fs_end: u64,
     pub cpuid_page: u64,
     pub secrets_page: u64,
+    pub stage2_igvm_params_phys_addr: u64,
+    pub stage2_igvm_params_size: u64,
     pub igvm_params_phys_addr: u64,
     pub igvm_params_virt_addr: u64,
     pub debug_serial_port: u16,
--- a/igvmbld/igvmbld.c
+++ b/igvmbld/igvmbld.c
@@ -78,9 +78,14 @@ static uint32_t crc32b_finish() {
    return ~_crc;
 }
 
-void construct_parameter_page(uint32_t address, ParameterPageIndex index)
+void construct_parameter_page(uint32_t address, ParameterPageIndex index, const char *description)
 {
     PARAM_PAGE *param_page;
+
+    if (is_verbose)
+    {
+        printf("%08X-%08X: \"%s\" parameter page\n", address, address + 0x1000, description);
+    }
 
     param_page = malloc(sizeof(PARAM_PAGE));
     param_page->address = address;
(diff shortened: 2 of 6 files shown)


bootlib/src/kernel_launch.rs
```diff
@@ -20,6 +20,8 @@ pub struct KernelLaunchInfo {
     pub kernel_fs_end: u64,
     pub cpuid_page: u64,
     pub secrets_page: u64,
+    pub stage2_igvm_params_phys_addr: u64,
```
msft-jlange, line 23: I think it would be cleaner if we avoid adding fields to the kernel launch info.  That should be possible by changing `SvsmConfig::invalidate_boot_data()` to return a vector of regions to be invalidated instead of a bool.  It could take the kernel launch info as its input and then it could add whatever regions are appropriate based on the boot configuration.  For `FirmwareConfig`, this would be an empty vector, and for `IgvmConfig`, it would be the kernel image, the filesystem, and the IGVM parameter block (which `IgvmParams` knows how to calculate without additional input).

msft-jlange: I take it back - I misunderstood the intention of the invalidation because there are actually two copies of the IGVM parameters, the copy manifested in the IGVM file and the copy that is made into the kernel region during stage 2.  You're talking about invalidating the first copy, which is in low memory (I somehow missed that fact).  The live `IgvmConfig` object doesn't know where that is, so it would have to be passed via the kernel lanch info as you have done.

However, in a perfect world, we could avoid doing this copy at all and instead just manifest the IGVM parameters at the base of the kernel region so there would be no copy that requires invalidation.  For the Hyper-V version of IGVM, that should be easy, because we know where the kernel region is at IGVM build time and we can easily put data there.  However, I'm now remembering why we didn't do this for QEMU, which is that stage 2 initially only has access to the 32-bit address space and the kernel region is above that, so placing the IGVM parameters directly into the kernel region would render them inaccessible to stage2.  I think we've said that in the long run (once OVMF can tolerate the SVSM living within the memory map) we will move the kernel region back down to the 32-bit address space and then we can put the IGVM paramters there directly on all platforms - and then none of this copying or invalidation would be necessary.

My preference, then, would be to modify this logic so that builds that can place the IGVM parameters directly into kernel memory will do so and avoid the copy and the invalidation, and builds that cannot will retain the copy and invalidation.  It probably makes the most sense for me to do that work on top of your change rather than asking you to refactor unless you feel like you really want to take the whole thing on now.

msft-jlange:
I'm unclear on why both your tests and my tests passed without requiring this change, and I'd like to understand that before merging this PR just in case there is some other hidden issue that we need to understand and that will be covered up after this change goes in.

msft-jlange (approved):

PR #210, opened 2024-01-11
scripts/check-signed-off: check that commits do not have an empty body

00xc:
Check that all new commits have a commit body besides the summary line and git trailers.

Unfortunately git does not have a direct option to show the body (i.e. everything after the summary line) without the trailers, so we have to filter them out manually.

### DIFF
--- a/scripts/check-signed-off.sh
+++ b/scripts/check-signed-off.sh
@@ -4,6 +4,9 @@
 # Copyright (c) 2023 SUSE LLC
 #
 # Author: Carlos López <carlos.lopez@suse.com>
+#
+# Simple script to check that a commit has a Signed-off-by trailer and a
+# nonempty commit body.
 
 matches_any() {
 	needle=$1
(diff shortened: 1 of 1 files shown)


scripts/check-signed-off.sh
```diff
@@ -18,6 +21,21 @@ matches_any() {
...
 
+# Check that the body for the given commit is not empty
+nonempty_body() {
+	body=$(git show --no-patch --format="%b" "$1" | sed '/^ *$/d')
+	trailers=$(git show --no-patch --format="%(trailers:only)" "$1")
+
+	# Remove trailers
+	for trailer in "${trailers[@]}"; do
+		body=$(echo "$body" | grep -F -v "$trailer")
+	done
```
joergroedel, line 32: Instead of removing the trailer, can we just count the number of lines for body and trailers and fail if they are equal?

00xc: Switched to this approach in the new version, thanks

PR #213, opened 2024-01-11
Introduce InsnBuffer and other improvements

p4zuu:
- Introduce InsnBuffer for similar Instructions fields.
-  Add unit tests for x86 instructions deocding
- Add documentation for svsm::cpu::insn

### DIFF
--- a/src/cpu/insn.rs
+++ b/src/cpu/insn.rs
@@ -53,7 +53,7 @@ where
 
 #[derive(Default, Debug, Copy, Clone, PartialEq)]
 pub struct Instruction {
-    pub prefixes: InsnBuffer<MAX_INSN_FIELD_SIZE>,
+    pub prefixes: Option<InsnBuffer<MAX_INSN_FIELD_SIZE>>,
     pub insn_bytes: InsnBuffer<MAX_INSN_SIZE>,
     pub opcode: InsnBuffer<MAX_INSN_FIELD_SIZE>,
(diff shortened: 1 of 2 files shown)


src/cpu/insn.rs
```diff
@@ -9,39 +9,74 @@ extern crate alloc;
...
+    pub nb_bytes: usize,
 }
 
-#[derive(Default, Debug, Copy, Clone)]
-pub struct InstructionField {
-    pub bytes: [u8; MAX_INSN_FIELD_SIZE],
-    pub nb_bytes: usize,
+impl<const N: usize> InsnBuffer<N>
+where
+    [u8; N]: Default + Index<usize>,
```
00xc, line 35: I think the `Index<usize>` bound is not necessary, as it is always met by any array of size `N`. `Default` is necessary because it is only implemented for certain values of `N` due to some toolchain limitations AFAIK.

joergroedel:
Hi Thomas,

In general this looks good to me.

What is you long-term plan with the decoder, do you plan to turn it into a full x86-64 instruction decoder at some point?

src/cpu/insn.rs
p4zuu: Indeed, thanks. Probably a leftover from my previous tries

joergroedel:
Alright, at some point we need a full decoder, but there is still some time.

PR #214, opened 2024-01-11
Remove references to SEV_INFO OVMF metadata since it is never used

msft-jlange:
The data from the SEV_INFO metadata field is never consumed anywhere, so
there is no reason to collect any data from it.

### DIFF
--- a/bootlib/src/igvm_params.rs
+++ b/bootlib/src/igvm_params.rs
@@ -52,7 +52,7 @@ pub struct IgvmParamBlockFwInfo {
     /// memory and will not be loaded into the ROM range.
     pub in_low_memory: u8,
 
-    _reserved: [u8; 3],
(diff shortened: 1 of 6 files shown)


roy-hopkins (approved):
I'm a little concerned that we _should_ be doing something with `reset_ip` in SVSM but we are not. I'll see if I can work out the intention behind why it was included in the first place. However, the removal in this PR has no functional effect due to the fact that it isn't used so I'm happy for this to be merged.

Tested working by myself with QEMU using both IGVM and non-IGVM configurations.

msft-jlange:
In that case we should probably wait a day or two before merging so you can spend a little time investigating.  If the investgiation takes too long or is too much of a distraction right now, we can merge this PR and it will be easy to revert later if we find a need for SEV_INFO.

PR #216, opened 2024-01-16
gdbstub: Fix order of task list when first BP occurs after tasks created

roy-hopkins:
A remote GDB connects to the SVSM GDB stub after SVSM hits the first breakpoint. At this point the remote GDB does not know which task is halted or anything about the current list of tasks in SVSM. Therefore after requesting the tasklist it assumes the first task (thread) in the list is the one that is halted. See 'qfThreadInfo' documentation at https://sourceware.org/gdb/current/onlinedocs/gdb.html/General-Query-Packets.html for details.

The order of the list of tasks provided by the GDB stub is based on iterating the task list. Therefore the first task reported by SVSM may not be the task that is currently halted.

This patch queries the current task and makes sure it is reported first every time the task list is requested.

### DIFF
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -539,31 +539,22 @@ pub mod svsm_gdbstub {
         ) -> Result<(), Self::Error> {
             let mut tl = TASKLIST.lock();
 
-            let mut any_scheduled = false;
-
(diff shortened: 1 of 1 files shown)


osteffenrh:
That fixes it for me. Thanks a lot!

PR #218, opened 2024-01-19
Rework Task and Scheduler Code

joergroedel:
Here are changes to the Task and Scheduler implemenation of COCONUT-SVSM. The highlights are:

* Move from a time-slice based to a simpler round-robin scheduler
* Remove affinity and allocation code - relying fully on WaitQueues to move tasks between CPUs
* Merge TaskNode and Task structs into one
* Rework locking for struct Task

Building on the Task/Scheduler changes this PR also introduces a simple WaitQueue and a BLOCKING task state, which allows to put tasks to sleep and wake them up again.

Finally the last patch in the PR uses all of that to move core protocol request processing into a separate kernel task.

### DIFF
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -24,7 +24,7 @@ pub mod svsm_gdbstub {
     use crate::mm::PerCPUPageMappingGuard;
     use crate::serial::{SerialPort, Terminal};
     use crate::svsm_console::SVSMIOPort;
-    use crate::task::{is_current_task, TaskContext, TaskState, INITIAL_TASK_ID, TASKLIST};
+    use crate::task::{is_current_task, TaskContext, INITIAL_TASK_ID, TASKLIST};
     use core::arch::asm;
     use core::fmt;
     use core::sync::atomic::{AtomicBool, Ordering};
@@ -572,9 +572,10 @@ pub mod svsm_gdbstub {
             let str = match tl.get_task(tid.get() as u32) {
                 Some(t) => {
                     let t = t.task.lock_read();
-                    match t.state {
-                        TaskState::RUNNING => "Running".as_bytes(),
-                        TaskState::TERMINATED => "Terminated".as_bytes(),
+                    if t.is_running() {
+                        "Running".as_bytes()
+                    } else {
+                        "Terminated".as_bytes()
                     }
                 }
                 None => "Stopped".as_bytes(),
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -16,7 +16,7 @@ pub mod svsm_gdbstub {
     use crate::address::{Address, VirtAddr};
     use crate::cpu::control_regs::read_cr3;
     use crate::cpu::idt::common::{X86ExceptionContext, BP_VECTOR, VC_VECTOR};
-    use crate::cpu::percpu::{current_task, this_cpu};
+    use crate::cpu::percpu::this_cpu;
     use crate::cpu::X86GeneralRegs;
     use crate::error::SvsmError;
     use crate::locking::{LockGuard, SpinLock};
@@ -104,8 +104,6 @@ pub mod svsm_gdbstub {
             flags: ctx.frame.flags as u64,
             ret_addr: ctx.frame.rip as u64,
         };
-
-        current_task().task.lock_write().rsp = &task_ctx as *const TaskContext as u64;
 
         // Locking the GDB state for the duration of the stop will cause any other
         // APs that hit a breakpoint to busy-wait until the current CPU releases
@@ -195,9 +193,9 @@ pub mod svsm_gdbstub {
             } else {
                 let tl = TASKLIST.lock();
                 let cr3 = read_cr3();
-                let task_node = tl.get_task(id);
-                if let Some(task_node) = task_node {
-                    task_node.task.lock_write().page_table.lock().load();
+                let task = tl.get_task(id);
+                if let Some(task) = task {
+                    task.page_table.lock().load();
                     cr3.bits()
                 } else {
                     0
@@ -446,11 +444,10 @@ pub mod svsm_gdbstub {
                 }
             } else {
                 let task = TASKLIST.lock().get_task(tid.get() as u32);
-                if let Some(task_node) = task {
+                if let Some(task) = task {
                     // The registers are stored in the top of the task stack as part of the
                     // saved context. We need to switch to the task pagetable to access them.
                     let _task_context = GdbTaskContext::switch_to_task(tid.get() as u32);
-                    let task = task_node.task.lock_read();
                     unsafe {
                         *regs = X86_64CoreRegs::from(&*(task.rsp as *const TaskContext));
                     };
@@ -548,7 +545,7 @@ pub mod svsm_gdbstub {
 
                 let mut cursor = tl.list().front_mut();
                 while cursor.get().is_some() {
-                    let this_task = cursor.get().unwrap().task.lock_read().get_task_id();
+                    let this_task = cursor.get().unwrap().get_task_id();
                     if this_task != current_task {
                         thread_is_active(Tid::new(this_task as usize).unwrap());
                     }
@@ -570,9 +567,8 @@ pub mod svsm_gdbstub {
             // Get the current task from the stopped CPU so we can mark it as stopped
             let tl = TASKLIST.lock();
             let str = match tl.get_task(tid.get() as u32) {
-                Some(t) => {
-                    let t = t.task.lock_read();
-                    if t.is_running() {
+                Some(task) => {
+                    if task.is_running() {
                         "Running".as_bytes()
                     } else {
                         "Terminated".as_bytes()
--- a/src/debug/gdbstub.rs
+++ b/src/debug/gdbstub.rs
@@ -548,7 +548,7 @@ pub mod svsm_gdbstub {
 
                 let mut cursor = tl.list().front_mut();
                 while cursor.get().is_some() {
-                    let this_task = cursor.get().unwrap().task.lock_read().id;
+                    let this_task = cursor.get().unwrap().task.lock_read().get_task_id();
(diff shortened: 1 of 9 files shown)


src/mm/pagetable.rs
```diff
@@ -104,6 +104,7 @@ fn set_c_bit(paddr: PhysAddr) -> PhysAddr {
 bitflags! {
     #[derive(Copy, Clone, Debug)]
     pub struct PTEntryFlags: u64 {
+        const NONE      = 0;
```
00xc, line 107: The `bitflag` docs explicitly discourage flags with no bits set. Instead we should use `PTEntryFlags::empty()`

joergroedel: Right, thanks. Will fix it.

roy-hopkins (commented):
I've done a first pass of reviewing the code and performed some testing. A few observations:

1. As you're aware, the gdbstub doesn't work entirely correctly. I've left a comment about one change of behaviour but it also deadlocks when performing cross-thread operations. There's also the FIXME in the code.
2. I've noticed some strange scheduler behaviour that seems to starve tasks. This is with some test code I previously wrote which may be the cause of the problem. I'll investigate further and add a comment with the findings.

src/task/schedule.rs
```diff
@@ -421,6 +428,49 @@ pub unsafe fn current_task_terminated() {
     TASKLIST.lock().terminate(task_node.clone());
 }
 
+fn task_pointer(taskptr: TaskPointer) -> *mut Task {
+    let mut guard = taskptr.task.lock_write();
+    let task = guard.as_mut();
+
+    task as *mut Task
+}
```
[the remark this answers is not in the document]
joergroedel: Agreed. Since the return value is only needed for the task switch, it might be even better to return a u64 or VirtAddr. That would avoid the mutable references. On the other hand task switching is unsafe anyway, so making this function unsafe does not hurt much.

src/task/schedule.rs
```diff
@@ -421,6 +428,49 @@ pub unsafe fn current_task_terminated() {
...
+    let next = task_pointer(this_cpu().runqueue().lock_write().schedule_init());
+    switch_to(null_mut(), next);
+}
+
+pub fn schedule_rr() {
+    let work = this_cpu().runqueue().lock_write().schedule_prepare();
+
+    // !!! Runqueue lock must be release here !!!
+    if work.is_some() {
+        let (current, next) = work.unwrap();
```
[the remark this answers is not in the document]
joergroedel: Yes, you are right, thanks for catching this. The `current´ and ´next´ variables are consumed by `task_pointer()` later, but the references are still live until switch_to() returns. I will fix this.

src/debug/gdbstub.rs
```diff
@@ -575,17 +573,7 @@ pub mod svsm_gdbstub {
...
-                                if this_cpu().get_apic_id() == allocation {
-                                    "Stopped".as_bytes()
-                                } else {
-                                    "Running".as_bytes()
-                                }
-                            } else {
-                                "Stopped".as_bytes()
-                            }
-                        }
+                        TaskState::RUNNING => "Running".as_bytes(),
```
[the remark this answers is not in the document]
joergroedel: The question is what is meant by "assigned". In the model implemented here only tasks which are in `RUNNING` state are assigned to a CPU, by either being in the CPUs run queue or being the `current_task`. Both can be evaluated by searching the local CPUs scheduling state. To optimise things the `cpu` field could be made an indicator. That requires some changes in the code, because the `cpu` field is not updated before the task is actually scheduled (so in the run queue the information may be outdated).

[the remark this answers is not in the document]
joergroedel: Yes, I agree that reporting the task state is the best option. Tasks which are not in `RUNNING` state do not have a CPU assigned and can theoretically run anywhere when woken up.
The underlying idea is that tasks run on the CPU where they were woken up. If they wait on a CPU-specific event source (like guest requests from a particular CPU), they will always run on that CPU. If they wait on a CPU-independent event source (e.g. TPM requests) they may next run on any CPU.

src/task/schedule.rs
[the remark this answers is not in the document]
joergroedel: Fixed, SAFETY comments still missing.

src/task/schedule.rs
```diff
@@ -74,12 +47,11 @@ impl RunQueue {
...
-    fn handle_task(&mut self, taskptr: TaskPointer) {
-        let task = taskptr.task.lock_read();
+    fn handle_task(&mut self, task: TaskPointer) {
         if task.is_running() && !task.is_idle_task() {
-            self.run_list.push_back(taskptr.clone());
+            self.run_list.push_back(task.clone());
         } else if task.is_terminated() {
-            self.terminated_task = Some(taskptr.clone());
+            self.terminated_task = Some(task.clone());
         }
```
00xc: We don't need these `task.clone()` calls since `task` is already passed by value.

src/debug/gdbstub.rs
```diff
@@ -446,20 +445,16 @@ pub mod svsm_gdbstub {
...
                     *regs = X86_64CoreRegs::from(context);
                 }
+            } else if let Some(task) = TASKLIST.lock().get_task(tid.get() as u32) {
+                // The registers are stored in the top of the task stack as part of the
+                // saved context. We need to switch to the task pagetable to access them.
+                let _task_context = GdbTaskContext::switch_to_task(tid.get() as u32);
+                unsafe {
+                    *regs = X86_64CoreRegs::from(&*(task.rsp as *const TaskContext));
+                };
+                regs.regs[7] = task.rsp;
```
roy-hopkins, line 455: The previous code deliberately assigned the task `Option` to a variable before checking if it was `Some` or `None` as otherwise the `TASKLIST` lock is held until the if scope is exited, causing a deadlock in `switch_to_task()`. This is what was causing the thread list to deadlock as GDB queries the register state of each task in turn.

```suggestion
            } else {
                let task = TASKLIST.lock().get_task(tid.get() as u32);
                if let Some(task) = task {
                    // The registers are stored in the top of the task stack as part of the
                    // saved context. We need to switch to the task pagetable to access them.
                    let _task_context = GdbTaskContext::switch_to_task(tid.get() as u32);
                    unsafe {
                        *regs = X86_64CoreRegs::from(&*(task.rsp as *const TaskContext));
                    };
                    regs.regs[7] = task.rsp;
```

src/debug/gdbstub.rs
```diff
@@ -105,9 +105,8 @@ pub mod svsm_gdbstub {
             ret_addr: ctx.frame.rip as u64,
         };
 
-        if let Some(task_node) = this_cpu_mut().runqueue().lock_read().get_task(id) {
-            task_node.task.lock_write().rsp = &task_ctx as *const TaskContext as u64;
-        }
+        // FIXME: current_task() return value is not mutable
+        //current_task().rsp = &task_ctx as *const TaskContext as u64;
```
roy-hopkins, line 109: I've taken a look at the code and this change doesn't seem to break the switching of threads in gdb. `task.rsp` is only used for reading registers in the non-current thread whereas the disabled code sets `task.rsp` for the current thread only. I think it can safely be removed.

src/task/schedule.rs
joergroedel: Right, thanks. Fixed now.

src/debug/gdbstub.rs
joergroedel: Thanks for checking. These two lines are removed now.

src/debug/gdbstub.rs
joergroedel: Thanks for fixing this up. I merged this change into the patch `SVSM/task: Move task scheduling state into separate struct`

src/task/schedule.rs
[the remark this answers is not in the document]
joergroedel: SAFETY comment added.

src/debug/gdbstub.rs
```diff
@@ -550,7 +545,7 @@ pub mod svsm_gdbstub {
 
                 let mut cursor = tl.list().front_mut();
                 while cursor.get().is_some() {
-                    let this_task = cursor.get().unwrap().task.lock_read().id;
+                    let this_task = cursor.get().unwrap().get_task_id();
                     if this_task != current_task {
                         thread_is_active(Tid::new(this_task as usize).unwrap());
                     }
```
roy-hopkins, line 551: The idle task for each CPU is not added to `TASKLIST` which means this loop misses all of the idle tasks. Is there any way we can include the idle tasks here?

joergroedel: Thanks, I missed that. Changed the second patch to put idle tasks on the task-list. This should resolve the problem.

src/debug/gdbstub.rs
```diff
@@ -572,21 +567,11 @@ pub mod svsm_gdbstub {
...
-                            } else {
-                                "Stopped".as_bytes()
-                            }
-                        }
-                        TaskState::TERMINATED => "Terminated".as_bytes(),
+                Some(task) => {
+                    if task.is_running() {
+                        "Running".as_bytes()
+                    } else {
+                        "Terminated".as_bytes()
```
roy-hopkins, line 574: Now the idle tasks appear in the task list it has highlighted one last issue where blocked tasks are marked as terminated:

```suggestion
                Some(task) => {
                    if task.is_running() {
                        "Running".as_bytes()
                    } else {
                        "Terminated".as_bytes()                Some(task) => {
                    if task.is_running() {
                        "Running".as_bytes()
                    } else if task.is_terminated() {
                        "Terminated".as_bytes()
                    } else {
                        "Blocked".as_bytes()
```

joergroedel: Alright, thanks. I added this hunk to the commit adding the BLOCKED task state.

roy-hopkins (approved):
As well as reviewing the code, I've tested your branch by creating a number of threads using some test code whilst connected to gdb using the gdbstub. All working well.

PR #219, opened 2024-01-21
Fix guest VMSA borrowing lifetimes and mutability requirements

msft-jlange:
Tie guest VMSA reference lifetime to the VMSA reference lock scope.
Remove requirement for mutability on PerCpu reference to obtain the guest VMSA, since the guest VMSA reference is protected by a spin lock and does not require static borrow checking.

### DIFF
--- a/src/cpu/percpu.rs
+++ b/src/cpu/percpu.rs
@@ -173,6 +173,11 @@ impl GuestVmsaRef {
     pub fn caa_phys(&self) -> Option<PhysAddr> {
         self.caa
     }
+
+    pub fn vmsa(&mut self) -> &mut VMSA {
+        assert!(self.vmsa.is_some());
+        unsafe { SVSM_PERCPU_VMSA_BASE.as_mut_ptr::<VMSA>().as_mut().unwrap() }
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


joergroedel:
This panics here when trying to boot a Linux kernel:

```
[SVSM] ERROR: Panic: CPU[0] panicked at src/cpu/percpu.rs:179:9:
assertion failed: self.gen_in_use == self.generation
```

I have not completely tracked it down yet, somewhere a VMSA reference is taken before update_mappings() is called.

joergroedel:
The most likely flow of events is:

1. Guest sends REMAP_CAA request
2. SVSM updates CAA physical address and increases `generation` counter to map new CAA on next request loop entry
3. When writing back request results, SVSM takes a reference to the VMSA via `this_cpu().guest_vmsa_ref()`
4. In this function the assertion fails, because the `generation` field does not match `gen_in_use`

msft-jlange:
Yes, this is kind of messy because GuestVmsaRef refers to the VMSA and CAA, not just the VMSA.  Ideally, we could separate these into two different lock scopes so that it's possible to update one without interfering with the other.  I think we are likely to eliminate this locking mechanism in the future once it's possible to schedule VMSA/CAA mapping requests onto the target vCPU, at which time we will replace GuestVmsaRef with a different reference management mechanism altogether.  Therefore, I think it makes more sense now just to get rid of this assert, and we can make everything more robust once we refactor the VMSA/CAA borrow checking to eliminate the use of spin locks.

PR #222, opened 2024-01-24
Keep missing_debug_implementations conditional

deeglaze:
The debug definitions to satisfy deny(missing_debug_implementation) are behind the enable-stacktrack feature, so the denial should also be behind this feature.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,7 @@
 
 #![no_std]
 #![deny(missing_copy_implementations)]
-#![deny(missing_debug_implementations)]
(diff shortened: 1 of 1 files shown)


00xc:
We should probably just derive `Debug` for the missing structs instead, this can lead to missing Debug derives if one disables the `enable-stacktrace` feature

00xc:
Turns out the struct missing the derive was not needed in the first place, so it can be deleted altogether. Build should be fixed with #223, so `deny(missing_debug_implementation)` can be kept unconditionally.

PR #232, opened 2024-01-29
Enforce dynamic borrow checking of the current CPU object

msft-jlange:


src/cpu/percpu.rs
```diff
@@ -458,8 +458,13 @@ impl PerCpu {
...
-    pub fn ghcb(&mut self) -> &'static mut GHCB {
-        unsafe { self.ghcb.as_mut().unwrap() }
+    /// # Safety
+    ///
+    /// This routine performs no checks on the validity of the current GHCB
+    /// and should only be used by callers that can verify that the GHCB is
+    /// properly usable.
+    pub unsafe fn ghcb_unsafe(&mut self) -> *mut GHCB {
+        self.ghcb
     }
```
00xc, line 486: Getting a pointer is not an unsafe operation in Rust. The unsafety is in using the pointer, like creating a reference (`&*ptr`) or dereferencing it (`*ptr`).

For example, see slices' `as_mut_ptr()` method in the standard library, which is not marked as unsafe:
https://doc.rust-lang.org/std/primitive.slice.html#method.as_mut_ptr

src/cpu/ghcb.rs
```diff
@@ -0,0 +1,38 @@
...
+    fn deref_mut(&mut self) -> &'static mut GHCB {
+        unsafe { &mut *self.ghcb }
+    }
+}
+
+pub fn current_ghcb() -> GHCBRef {
+    // FIXME - Add borrow checking to GHCB references.
+    unsafe {
+        let cpu_ptr = this_cpu_mut();
+        let cpu = &mut *cpu_ptr;
```
00xc, line 34: What is this line doing exactly? `cpu_ptr` is already a `&mut PerCpu`.

src/cpu/percpu.rs
```diff
@@ -607,12 +622,98 @@ impl PerCpu {
...
+}
+
+pub fn this_cpu_mut() -> CpuRefMut {
+    unsafe {
+        let cpu = SVSM_PERCPU_BASE.as_mut_ptr::<PerCpu>();
+        #[cfg(debug_assertions)]
+        {
+            // Prohibit a mutable borrow if any other borrow exists.
+            assert!((*cpu).borrow_count.0 == 0);
+            assert!((*cpu).borrow_count.1 == 0);
```
joergroedel, line 693: These assertions need to be warning or error messages for now, at least until we fixed all cases that trigger them. Currently this breaks booting in debug builds.

joergroedel (changes requested):
Hi Jon,

Thanks for working on making the PerCPU borrowing safe. I tested the changes and found a couple issues which need fixing first:
* `make test-in-svsm` is broken, a change in src/testing.rs is needed
* Debug builds do not boot because they trigger the debug assertions. See my inline comment
* Found a build issue with release builds, switch_context needs to be made global to fix it.

src/cpu/percpu.rs
```diff
@@ -612,6 +612,11 @@ impl PerCpu {
     }
 }
 
+pub fn get_current_apic_id() -> u32 {
+    let cpu = this_cpu();
+    (*cpu).get_apic_id()
+}
```
00xc: I don't see how this solves anything vs `this_cpu().get_apic_id()`. We are still going through an immutable borrow borrow and the borrow is temporary in both cases. If we *really* want to avoid creating a reference we would need to do a `ptr::read()` from `SVSM_PERCPU_BASE` plus a certain offset.

src/cpu/percpu.rs
```diff
@@ -613,16 +624,97 @@ impl PerCpu {
...
+
+pub fn this_cpu() -> CpuRef {
+    unsafe {
+        let cpu = SVSM_PERCPU_BASE.as_mut_ptr::<PerCpu>();
+        #[cfg(debug_assertions)]
+        {
+            // Prohibit a shared borrow if a mutable borrow exists.
+            assert!((*cpu).borrow_count.1 == 0);
+            (*cpu).borrow_count.0 += 1;
+        }
```
00xc, line 681: Instead of doing all of this accounting manually perhaps `CpuRef` should be something like a `RefCell<*mut PerCpu>` when `debug_assertions` is enabled.

joergroedel:
I fixed up the three issues mentioned in my last comment and pushed an updated branch [here](https://github.com/joergroedel/svsm/tree/cpu_borrow)

src/cpu/percpu.rs
msft-jlange: The intention is to fix all issues before the PR merges, so these can be asserts from the start.  I didn't see any issues when I booted my limited scenario, though I am aware of at least two mistakes that could contribute to your problems.  Perhaps once those are fixed you can test again and perhaps help me identify where the remaining borrow conflicts are so they can be fixed as part of this PR?

src/cpu/percpu.rs
msft-jlange: This is a mistake - it should be a call to `this_cpu_unsafe()`.  The intention here is to avoid a borrow conflict when obtaining the APIC ID, because it is safe to obtain the APIC ID (which is always constant) from the CPU even if there is an outstanding borrow on the CPU itself (essentially implementing borrow splitting).

msft-jlange: It seems you may have been looking the wrong version of the PR that I may have pushed inadvertently - the current draft uses `this_cpu_unsafe()`.

src/cpu/percpu.rs
msft-jlange: The reason this is marked unsafe is because it does not perform any borrow checking with respect to the current CPU, and therefore callers must be on notice that they cannot safely use this call directly.  I suppose your position might be that because it returns a pointer, it is already inherently unsafe and therefore does not need to be marked as such.  We certainly do not want any callers calling `PerCpu::ghcb()` directly other than `current_ghcb()` because `current_ghcb()` is the only function that will know enough to handle the borrow checks properly (I have work underway to add GHCB borrow checking to this function).  The question is whether the best way to achieve that is to decorate `PerCpu::ghcb()` as unsafe or whether we can simply assume that any callers are smart enough to understand what's going on here.

src/cpu/ghcb.rs
msft-jlange: The current version of the PR (commit b08e81c7fdca5d1e342faf2ac1be56030a0dc833) doesn't have this code; it uses `this_cpu_unsafe()`.  I can clean up the individual commits so that the code sequence you identified above doesn't appear in any of them, but the final version will rely on `this_cpu_unsafe()`.

src/cpu/percpu.rs
msft-jlange: I tested again and found the borrow conflict - not sure why my earlier testing didn't reveal this.  The borrow conflict is now fixed.

msft-jlange:
> I fixed up the three issues mentioned in my last comment and pushed an updated branch [here](https://github.com/joergroedel/svsm/tree/cpu_borrow)

I've incorporated your fixes for the release and test issues into the PR.  I had already found and fixed the borrow conflict.  Thanks for the help!

src/cpu/percpu.rs
00xc: > * Certain paths need to be able to bypass the borrow check, even in debug builds, because they can safely peer into the PerCpu to extract data that is not subject to borrow conflicts (the APIC ID and the GHCB are two examples).

This can be done through the raw pointer if needed. Also, using an immutable reference to the PerCpu struct that is aliased by a mutable reference is UB, even if the APIC ID is constant.

> * (...) We could return `CpuRef` which is defined to hold one or the other depending on the configuration, but that involves deref/drop logic similar to what's implemented in this PR.  It's not clear that such an overhaul actually results in an improvement.

Yes, this is what I meant. It's an improvement in the sense that it uses a battle-tested standard library type and removes the need for us to re-implement dynamic borrow checking.

src/cpu/percpu.rs
00xc: Yes, but the correct use of the `unsafe` keyword is for APIs which may be unsound under certain conditions. `ghcb_unsafe()` cannot create UB, just like this can't:

```rust
let val: u8 = 45;
let ptr: *const u8 = &val as *const u8;
```

In order to convey this to callers I would simply remove `Deref` / `DerefMut` and add an explicitly unsafe API that returns a `&GHCB` / `&mut GHCB` from `GHCBRef`. Hiding these unsafe operations behind the traits sets us up to do the unsafety decoration backwards.

msft-jlange: The function `current_ghcb()` is designed to be safe, because it is where borrow checking and GHCB nesting management will ultimately live.  The intention here is to make `PerCpu::ghcb_unsafe()` unsafe to force all callers to go through `current_ghcb()` instead.  The intention with `GHCBRef` is to provide a transparent mechanism to provide a validated reference to the current GHCB.  Eliminating `GHCBRef` and exposing the raw pointer to callers would require all of them to be unsafe, which is certainly what we don't want.

Note that GHCB borrow checking and nesting is not part of this PR but will come in the future.  Part of the intention here was to establish the `current_ghcb()` usage pattern now so that future nesting management can be added transparently in the future.

It sounds like your point is that `PerCpu::ghcb_unsafe()` already implies the fact that it should not be used directly because it returns a pointer and not a reference.  I can clean up the unsafe decoration if that is better.

src/cpu/percpu.rs
msft-jlange: > Also, using an immutable reference to the PerCpu struct that is aliased by a mutable reference is UB, even if the APIC ID is constant.

Fair point, and something we should clean up, but that problem exists countless times over in the current code - which is exactly why we are sorely in need of borrow checking on the current CPU.  It seems to me that it would best to solve this in phases, where we implement basic borrow checking first and then get rid of overloaded borrows to constant fields like APIC ID.

> > * (...) We could return `CpuRef` which is defined to hold one or the other depending on the configuration, but that involves deref/drop logic similar to what's implemented in this PR.  It's not clear that such an overhaul actually results in an improvement.
> Yes, this is what I meant. It's an improvement in the sense that it uses a battle-tested standard library type and removes the need for us to re-implement dynamic borrow checking.

A reasonable plan, but I don't know enough Rust to know how to create entirely different implementations between debug and non-debug code.  Again, this seems like another great candidate for phasing, where we get basic borrow checking in place to prevent the creation of new problems, and then improve the borrow checking implementation in a future PR.

joergroedel:
Merged manually, thanks Jon.

PR #235, opened 2024-01-30
Implement an IGVM builder in rust

roy-hopkins:
Coconut SVSM currently packages the and optionally the OVMF firmware inside an IGVM file using a builder tool that is invoked from the `Makefile` for both QEMU and Hyper-V. The current implementation of the builder tool is written in C. The intention was always to replace this with a rust implementation of a builder.

This PR implements an IGVM builder in rust equivalent to the C implementation, and is intended to provide a drop-in replacement.

The commits in the PR are engineered to demonstrate the fact that both builders generate the same output, which generally follows this sequence:

1. Implement the new rust builder (4edabe71d00171ad158632a302ab552e314420c9, 12219e06ae48dc57172e9ce810ed17b394fb2b2b and 9d978f7a2860d6ce912547ceb263475c308b3b9c)
2. Sort the output of directives in both builders to make it easy to compare outputs (766c80136246e8b20a7e61385d9a3aa49978a7e5, 083880bf21f71b0fe2b9f79eb428619c8b4c1423)
3. Remove the C implementation of the builder (ca85d7ca4eef085f73f3532ba01122b02cde8a72)

This approach meant that I could generate a sorted output from both the C and rust builders simultaneously during a build, dump the output using the `dump_igvm` tool implemented in this PR: https://github.com/microsoft/igvm/pull/3 then perform a `diff` to ensure the outputs are consistent for the following configurations:

1. QEMU without OVMF
2. QEMU with OVMF
3. Hyper-V without firmware
4. Hyper-V with IGVM based firmware

Once I determined the outputs were identical, the final commit was added to remove the C builder.


msft-jlange:
Booting with real Hyper-V firmware works identically between the new and old versions of the file builder - no regressions here!

PR #238, opened 2024-01-31
Makefile: respect `CARGO_ARGS` when building the igvmbuilder

00xc:
When building the igvmbuilder binary, respect `CARGO_ARGS`, which potentially includes among other things the `--release` flag. This was causing an error when building in release mode, as the resulting binary was being placed in the debug build path, but the Makefile was expecting it to be placed in the release path:

```
/bin/sh: line 1: target/x86_64-unknown-linux-gnu/release/igvmbuilder: No such file or directory
make: *** [Makefile:44: bin/coconut-qemu.igvm] Error 127
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 FEATURES ?= "default"
-CARGO_ARGS = --features ${FEATURES}
+SVSM_ARGS = --features ${FEATURES}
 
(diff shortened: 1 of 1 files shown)


roy-hopkins (approved):
LGTM. Thanks for fixing this.

PR #241, opened 2024-02-01
Copy IGVM builder to bin directory

msft-jlange:
To support external workflows that are not incorporated into the makefile, publish the IGVM file builder to the bin directory so it can be consumed by those workflows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,19 +31,25 @@ C_BIT_POS ?= 51
 STAGE1_OBJS = stage1/stage1.o stage1/reset.o
 IGVM_FILES = bin/coconut-qemu.igvm bin/coconut-hyperv.igvm
 IGVMBUILDER = "target/x86_64-unknown-linux-gnu/${TARGET_PATH}/igvmbuilder"
+IGVMBIN = bin/igvmbld
 
 all: stage1/kernel.elf svsm.bin igvm
 
-igvm: $(IGVM_FILES)
+igvm: $(IGVM_FILES) $(IGVMBIN)
+
+$(IGVMBIN): $(IGVMBUILDER)
+	mkdir -v -p bin
+	cp -f $(IGVMBUILDER) $@
 
 $(IGVMBUILDER):
-	mkdir -v -p bin
 	CARGO_TARGET_DIR=target cargo build ${CARGO_ARGS} --target=x86_64-unknown-linux-gnu --manifest-path igvmbuilder/Cargo.toml
 
 bin/coconut-qemu.igvm: $(IGVMBUILDER) stage1/kernel.elf stage1/stage2.bin
+	mkdir -v -p bin
 	$(IGVMBUILDER) --sort --output $@ --stage2 stage1/stage2.bin --kernel stage1/kernel.elf ${BUILD_FW} qemu
 
 bin/coconut-hyperv.igvm: $(IGVMBUILDER) stage1/kernel.elf stage1/stage2.bin
+	mkdir -v -p bin
 	$(IGVMBUILDER) --sort --output $@ --stage2 stage1/stage2.bin --kernel stage1/kernel.elf --comport 3 hyper-v
 
 test:


Makefile
```diff
@@ -31,10 +31,14 @@ C_BIT_POS ?= 51
 STAGE1_OBJS = stage1/stage1.o stage1/reset.o
 IGVM_FILES = bin/coconut-qemu.igvm bin/coconut-hyperv.igvm
 IGVMBUILDER = "target/x86_64-unknown-linux-gnu/${TARGET_PATH}/igvmbuilder"
+IGVMBIN = bin/igvmbld
```
stefano-garzarella, line 34: Is there a particular reason for renaming it?

I mean copying it to bin is okay IMO, but I'd leave it with the same name, so there's no confusion.

Makefile
```diff
@@ -31,10 +31,14 @@ C_BIT_POS ?= 51
...
 all: stage1/kernel.elf svsm.bin igvm
 
-igvm: $(IGVM_FILES)
+igvm: $(IGVM_FILES) $(IGVMBIN)
+
+$(IGVMBIN): $(IGVMBUILDER)
+	cp -f $(IGVMBUILDER) $@
 
 $(IGVMBUILDER):
 	mkdir -v -p bin
```
stefano-garzarella, line 44: Should we move this line on the target we are adding in this PR (line 40)?

Makefile
msft-jlange: I kept the old name because the external workflows assume the name "igvmbld".  The new tool has a new name so it could coexist with the old one during development, and I'd argue that now that the tool conversion is complete, it should change its name back to the old name for the sake of continuity.  That would require more discussion, but at least by having the copy step keep the old name, we can have continuity for external workflows, which was the whole idea.

Makefile
msft-jlange: We could, but keeping it as a separate target keeps the dependencies obvious.  This logic explicitly calls for a file to be placed into the bin directory, and if the file already in the bin directory is out of date, it will copy a new one (rebuilding if necessary).  Having the copy be a side-effect of building the tool runs the risk of not completing the build because there is no dependency check on the bin output.

Makefile
stefano-garzarella: Ah I see, I missed that part ;-)

Makefile
stefano-garzarella: I'm fine with separate targets. I meant, why we are creating the bin directory here, if it is only used by the `$(IGVMBIN)` target?

msft-jlange: In order to support external workflows, all binaries that can be consumed in the production of a COCONUT-SVSM binary need to be available in a place that is not target-dependent (i.e. "make" and "make RELEASE=1" must both put their artifcats into the same place).  That includes the IGVM file builder, as well as binaries like stage2 and the kernel.  Ideally, stage2 and the kernel would also be copied to the bin directory so there is a single point of distribution, but right now, they are copied to the stage1 directory (because of the way the stage1 build process works).  That's good enough for now at least.  When the day comes that we eliminate stage1, I believe we will move the distribution point for stage2 and the kernel into the bin directory, and then everything will get more consistent.

msft-jlange: Oh - and I see your point now about the creation of the bin directory.  I agree that should be relocated to a place where it is actually used.  I'll plan to revise the PR to make that change.

Makefile
```diff
@@ -31,19 +31,26 @@ C_BIT_POS ?= 51
...
 
-igvm: $(IGVM_FILES)
+igvm: $(IGVM_FILES) $(IGVMBIN)
+
+$(IGVMBIN): $(IGVMBUILDER)
+	mkdir -v -p bin
+	cp -f $(IGVMBUILDER) $@
 
 $(IGVMBUILDER):
 	mkdir -v -p bin
```
stefano-garzarella, line 45: So we can remove this now, right?

Makefile
```diff
@@ -31,19 +31,26 @@ C_BIT_POS ?= 51
...
-	CARGO_TARGET_DIR=target cargo build ${CARGO_ARGS} --target=x86_64-unknown-linux-gnu --manifest-path igvmbuilder/Cargo.toml
+	CARGO_TARGET_DIR=target cargo build --target=x86_64-unknown-linux-gnu --manifest-path igvmbuilder/Cargo.toml
```
stefano-garzarella, line 46: Is this an intentional change?

msft-jlange: No, this was a bad merge with the PRs that completed earlier today.  I thought I grabbed the current code but apparently not.  Good catch!

Makefile
msft-jlange: No, because then the tool will not be copied to the bin directory, which is the whole point of the PR.  Based on my earlier comments about explicit dependencies in the makefile, do you have a better suggestion for how to ensure the binary is copied while keeping it as a distinct dependency?

msft-jlange: Sorry I misunderstood your point.  I completely agree and will remove the extra line.

PR #245, opened 2024-02-02
repo: Move all rust crates into a single workspace

chris-oo:
This change moves all rust crates into a single workspace, which helps organize and manage code more easily throughout the repo.

It also helps make rust-analyzer work better, as now all crates in the workspace are correctly recognized and loaded by rust-analyzer. Note that in order to help rust-analyzer correctly recognize fuzzing only methods, #[cfg(fuzzing)] was changed to an explicit feature. Otherwise, rust-analyzer errors out about missing methods which is a poor developer experience.

Some default workspace lints are added to the root Cargo.toml along with additional options noted that currently are too noisy and require more fixing.

Future changes to further split the large kernel crate into smaller crates can now easily be done with the repo having a single workspace. The first change following should be to split out the stage2 binary target into its own crate.


Cargo.toml
```diff
@@ -1,40 +1,66 @@
...
+packit = { git = "https://github.com/coconut-svsm/packit", version = "0.1.1" }
 
-[features]
-default = []
-enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
+[workspace.lints.rust]
+future_incompatible = "deny"
+nonstandard_style = "deny"
+# TODO: Many lints fire from this, fix then enable.
+# rust_2018_idioms = "deny"
```
00xc, line 44: These will be fixed with #248

chris-oo: awesome

kernel/src/fs/filesystem.rs
```diff
@@ -178,7 +178,7 @@ impl SvsmFs {
         self.root = Some(root.clone());
     }
 
-    #[cfg(all(any(test, fuzzing), not(test_in_svsm)))]
+    #[cfg(all(any(test, feature = "fuzzing-hooks"), not(test_in_svsm)))]
```
stefano-garzarella, line 181: Not a strong opinion, but I would have made this change in a separate preparation commit.

PR #255, opened 2024-02-08
Improve igvmbuilder error reporting

msft-jlange:
Errors involving any of the input files or output should report which file access failed and should return the error so it can be decoded and reported.

### DIFF
--- a/igvmbuilder/src/igvm_builder.rs
+++ b/igvmbuilder/src/igvm_builder.rs
@@ -120,13 +120,22 @@ impl IgvmBuilder {
         }
 
         let file = IgvmFile::new(IgvmRevision::V1, self.platforms, vec![], self.directives)
-            .expect("Failed to create output file");
+            .map_err(|e| {
+                eprintln!("Failed to create output file");
+                e
+            })?;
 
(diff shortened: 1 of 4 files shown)


igvmbuilder/src/igvm_builder.rs
```diff
@@ -353,7 +362,10 @@ impl IgvmBuilder {
         gpa_start: u64,
     ) -> Result<(), Box<dyn Error>> {
         let mut gpa = gpa_start;
-        let mut in_file = File::open(path).expect("Could not open input file");
+        let mut in_file = File::open(path).map_err(|e| {
+            eprintln!("Could not open input file{}", path);
```
roy-hopkins, line 366: Missing space between in `file{}`.

roy-hopkins (commented):
Looks good. Just a formatting error in one of the error messages to fix.

igvmbuilder/src/igvm_builder.rs
joergroedel: Fixed that up manually. Thanks.

joergroedel:
Applied manually. Closing.

PR #256, opened 2024-02-08
Move all generated binary targets to bin dir

msft-jlange:
Besides being more consistent, this improves the process of performing a clean make ("make clean") because all generated targets are in one place and can be removed more consistently.

The "make clean" target contines to refer to the old files in the stage1 directory so that executing "make clean" after picking up this change will correctly remove the old files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,29 +38,29 @@ IGVM_FILES = bin/coconut-qemu.igvm bin/coconut-hyperv.igvm
 IGVMBUILDER = "target/x86_64-unknown-linux-gnu/${TARGET_PATH}/igvmbuilder"
 IGVMBIN = bin/igvmbld
 
-all: svsm.bin igvm
+all: bin/svsm.bin igvm
 
 igvm: $(IGVM_FILES) $(IGVMBIN)
 
-$(IGVMBIN): $(IGVMBUILDER)
+bin:
 	mkdir -v -p bin
+
+$(IGVMBIN): $(IGVMBUILDER) bin
 	cp -f $(IGVMBUILDER) $@
 
 $(IGVMBUILDER):
 	cargo build ${CARGO_ARGS} --target=x86_64-unknown-linux-gnu -p igvmbuilder
 
-bin/coconut-qemu.igvm: $(IGVMBUILDER) stage1/svsm-kernel.elf stage1/stage2.bin
-	mkdir -v -p bin
-	$(IGVMBUILDER) --sort --output $@ --stage2 stage1/stage2.bin --kernel stage1/svsm-kernel.elf ${BUILD_FW} qemu
+bin/coconut-qemu.igvm: $(IGVMBUILDER) bin/svsm-kernel.elf bin/stage2.bin
+	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf ${BUILD_FW} qemu
 
-bin/coconut-hyperv.igvm: $(IGVMBUILDER) stage1/svsm-kernel.elf stage1/stage2.bin
-	mkdir -v -p bin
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -71,54 +69,61 @@ utils/print-meta: utils/print-meta.c
...
 	cc -O3 -Wall -o $@ $<
 
-stage1/meta.bin: utils/gen_meta utils/print-meta
+bin/meta.bin: utils/gen_meta utils/print-meta
 	./utils/gen_meta $@
 
-stage1/stage2.bin:
+bin/stage2.bin:
 	cargo build ${CARGO_ARGS} ${SVSM_ARGS} --bin stage2
+	mkdir -v -p bin
```
stefano-garzarella, line 77: Since we're touching this code and adding more calls to `mkdir -v -p bin`, what about adding a dedicated target, so we can avoid duplicating/calling it multiple times?

I mean something like this:

```
bin:
	mkdir -v -p bin

...

bin/stage2.bin: bin
...
bin/svsm-kernel.elf: bin
...
etc.
```

Makefile
```diff
@@ -71,54 +69,61 @@ utils/print-meta: utils/print-meta.c
...
 	objcopy -O binary $< $@
 
-svsm-test.bin: stage1/stage1-test
+svsm-test.bin: bin/stage1-test
 	objcopy -O binary $< $@
 
 clean:
 	cargo clean
-	rm -f stage1/stage2.bin svsm.bin stage1/meta.bin stage1/kernel.elf stage1/stage1 stage1/svsm-fs.bin ${STAGE1_OBJS} utils/gen_meta utils/print-meta
+	rm -f stage1/*.o stage1/*.bin stage1/*.elf
```
stefano-garzarella, line 125: I think we can remove this line, since `*.elf` and `*.bin` files are now in `bin/`, and `*.o` are covered in the next line by `${STAGE1_OBJS}`

```suggestion
```

Makefile
```diff
@@ -71,54 +69,61 @@ utils/print-meta: utils/print-meta.c
...
-stage1/stage1-test: ${STAGE1_TEST_OBJS}
+bin/stage1-test: ${STAGE1_TEST_OBJS}
 	$(CC) -o $@ $(STAGE1_TEST_OBJS) -nostdlib -Wl,--build-id=none -Wl,-Tstage1/stage1.lds -no-pie
 
-svsm.bin: stage1/stage1
+svsm.bin: bin/stage1
 	objcopy -O binary $< $@
 
-svsm-test.bin: stage1/stage1-test
+svsm-test.bin: bin/stage1-test
```
stefano-garzarella, line 120: Should we move these two to bin as well, or since they are the main artifacts do we leave them in the main folder?

Makefile
msft-jlange: Good idea!

Makefile
msft-jlange: As I wrote in the PR comments,
>The "make clean" target contines to refer to the old files in the stage1 directory so that executing "make clean" after picking up this change will correctly remove the old files.

We want to make sure that the first time somebody runs "make clean" after picking up this PR, the old binary files will be removed and not left behind forever.

stefano-garzarella: Ah sorry, I had missed that comment. I honestly wouldn't worry about it too much, but I get the point, no objection ;-)

Makefile
msft-jlange: The intention is that all generated binary outputs should go into `bin`, and the `stage1` dir should just be for sources or for intermediate objects generated by the C compiler (even those intermediate objects should go somewhere else, but that's a task for a future PR).

stefano-garzarella: So should we move also `svsm.bin` and `svsm-test.bin` in `bin/`?

msft-jlange: I misunderstood what you were getting at.  I've moved those two targets as well.

msft-jlange:
> Last thing, we should update `Documentation/INSTALL.md`

Done!

stefano-garzarella (approved):

PR #257, opened 2024-02-09
Add OFFLINE build option.

deeglaze:
Our extra locked down internal build environment has no network access, so we need this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,11 @@ else ifeq ($(V), 2)
 CARGO_ARGS += -vv
 endif
 
+ifeq ($(OFFLINE), true)
+CARGO_ARGS += --locked --offline
+endif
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,6 +14,11 @@ else ifeq ($(V), 2)
 CARGO_ARGS += -vv
 endif
 
+ifeq ($(OFFLINE), true)
+CARGO_ARGS += --locked --offline
+endif
```
00xc, line 19: Maybe we could use `ifdef` so it's more consistent with other options (e.g. `RELEASE`).

joergroedel: Made that change when applying.

joergroedel:
Applied manually. Closing.

PR #258, opened 2024-02-09
Makefile: add clippy target

msft-jlange:
Following the change to a cargo workspace, running "cargo clippy" no longer completes successfully due to the variety of configurations in the workspace.  Now, executing "make clippy" completes the same operations as the CI flow, enabling local validation of clippy prior to submitting a PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,10 @@ svsm.bin: stage1/stage1
 svsm-test.bin: stage1/stage1-test
 	objcopy -O binary $< $@
 
+clippy:
+	cargo clippy --workspace --exclude igvmbuilder --exclude svsm-fuzz --all-features -- -D warnings
+	cargo clippy --workspace --all-features --exclude svsm --target=x86_64-unknown-linux-gnu -- -D warnings
+
 clean:
 	cargo clean
 	rm -f stage1/stage2.bin svsm.bin stage1/meta.bin stage1/kernel.elf stage1/stage1 stage1/svsm-fs.bin ${STAGE1_OBJS} utils/gen_meta utils/print-meta


Makefile
```diff
@@ -120,6 +120,10 @@ svsm.bin: stage1/stage1
 svsm-test.bin: stage1/stage1-test
 	objcopy -O binary $< $@
 
+clippy:
```
joergroedel, line 123: `clippy` also needs to be added to the `.PHONY` target.

00xc:
I would suggest adding the changes in #260 as well

Makefile
msft-jlange: My understanding is that `.PHONY` is only required for targets that collide with file names, as this tells `make` to proceed with the target even if it already exists.  This is needed, for example, for files that are generated by `cargo` because their dependencies are not expressible in a makefile, and thus `make` needs to execute the target rule regardless of whether it exists or whether it has tracked dependencies.  However, for targets that are truly phony, like `all` and `clean`, there is no file with the matching target name, and therefore `make` will always execute the target rule because it will determine that the target does not exist and must be executed.  By this logic, there is no reason to include `clippy` in the `.PHONY` list unless we expect to have a file at the root of the project called `clippy`.

I would argue that `clean` is incorrectly included in `.PHONY` today because it also does not correspond to a filename.  We have an `all` target in our makefile which is not included in `.PHONY`, which is as it should be.  We also do include `test` in `.PHONY` because there is a directory by the same name, and thus the `.PHONY` declaration is required to disambiguate.

msft-jlange:
> I would suggest adding the changes in #260 as well

Is it possible to have rust.yml invoke `make clippy`?  Then we could have the clippy arguments live in one place and not have to keep two different files in sync.

Makefile
joergroedel: This is almost right, `.PHONY` is for targets that do not generate a file with the same name. This of course implies that the target commands will always be executed when no file is generated, but more robust is to add these targets to `.PHONY` to make it clear that it is not expected that these targets generate a file.
The ELF files in there are a different story, though. They are there because `make` can not figure out the dependencies and needs to rely on `cargo` for that.

joergroedel:
> > I would suggest adding the changes in #260 as well
> 
> Is it possible to have rust.yml invoke `make clippy`? Then we could have the clippy arguments live in one place and not have to keep two different files in sync.

Good suggestion, let's look into this with a separate PR.

joergroedel:
I applied the patch manually and added `clippy` to the `.PHONY` target.

PR #262, opened 2024-02-12
Cargo.toml: extend SVSM kernel lints to whole workspace

00xc:
Move required lints in `kernel/src/lib.rs` to Cargo.toml so they apply to all crates in the workspace.


00xc:
@joergroedel manual apply failed again. Closing.

PR #265, opened 2024-02-12
igvmbuilder: avoid unsafe blocks

00xc:
Use zerocopy's `AsBytes` trait to get a byte representation of certain types where it is convenient. This requires deriving such trait in the bootlib crate.

### DIFF
--- a/bootlib/Cargo.toml
+++ b/bootlib/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+zerocopy.workspace = true
 
 [lints]
 workspace = true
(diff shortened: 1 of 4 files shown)


bootlib/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+zerocopy.workspace = true
```
roy-hopkins, line 7: This effectively adds a dependency on zerocopy to the kernel as it shares the bootlib package. Is that ok?

00xc: Yes, it was already a dependency via the packit crate.

PR #266, opened 2024-02-13
Rework Entry code to use separate entry points and handling functions per exception vector

joergroedel:
This PR changes the entry code of the SVSM kernel to implement separate entry points for each vector in assembly and call vector-specific handling functions. With these changes the vector dispatch in generic_idt_handler() is not needed anymore and the old entry code is removed.


kernel/src/cpu/idt/svsm.rs
```diff
@@ -11,29 +11,71 @@ use super::super::tss::IST_DF;
...
+    idt_mut().set_entry(PF_VECTOR, IdtEntry::entry_with_handler(asm_entry_pf));
+    idt_mut().set_entry(MF_VECTOR, IdtEntry::entry_with_handler(asm_entry_mf));
+    idt_mut().set_entry(AC_VECTOR, IdtEntry::entry_with_handler(asm_entry_ac));
+    idt_mut().set_entry(MCE_VECTOR, IdtEntry::entry_with_handler(asm_entry_mce));
+    idt_mut().set_entry(XF_VECTOR, IdtEntry::entry_with_handler(asm_entry_xf));
+    idt_mut().set_entry(CP_VECTOR, IdtEntry::entry_with_handler(asm_entry_cp));
+    idt_mut().set_entry(HV_VECTOR, IdtEntry::entry_with_handler(asm_entry_hv));
+    idt_mut().set_entry(VC_VECTOR, IdtEntry::entry_with_handler(asm_entry_vc));
+    idt_mut().set_entry(SX_VECTOR, IdtEntry::entry_with_handler(asm_entry_sx));
+    idt_mut().load();
```
00xc: `idt_mut()` acquires a lock, so perhaps we should acquire it just once:

```rust
let idt = idt_mut();
idt.set_entry(DE_VECTOR, IdtEntry::entry_with_handler(asm_entry_de));
// ...
idt.load();
```

kernel/src/cpu/idt/common.rs
```diff
@@ -92,6 +92,16 @@ impl IdtEntry {
...
 
+    pub fn entry_with_handler(handler: unsafe extern "C" fn()) -> Self {
+        let target = VirtAddr::from(handler as *const () as u64);
+        IdtEntry::create(target, SVSM_CS, 0)
+    }
+
+    pub fn ist_entry_with_handler(handler: unsafe extern "C" fn(), ist: u8) -> Self {
+        let target = VirtAddr::from(handler as *const () as u64);
+        IdtEntry::create(target, SVSM_CS, ist)
+    }
```
00xc: `VirtAddr` implements `From<*const T>` so these can just be:

```rust
        let target = VirtAddr::from(handler as *const ());
```

p4zuu:
Looks good to me to add IST support upon this. I would need to add a bit of code between `push regs` and `call	ex_handler_\handler`, but  I guess I can add a new IST entry in `entry.S`  that will be loaded in `init_ist_vectors()` (instead of using the `default_entry_no_ist`).

kernel/src/cpu/idt/svsm.rs
joergroedel: Right, changed that, thanks.

kernel/src/cpu/idt/common.rs
joergroedel: Thanks, removed the unnecessary casts.

joergroedel:
> Looks good to me to add IST support upon this. I would need to add a bit of code between `push regs` and `call ex_handler_\handler`, but I guess I can add a new IST entry in `entry.S` that will be loaded in `init_ist_vectors()` (instead of using the `default_entry_no_ist`). I think it will be much cleaner to add IST entries with these changes than with the current code.

Yes, the way to do this is to create new macro similar to `default_entry_no_ist` for VC and use it to create the VC entry point.

00xc (approved):

p4zuu (approved):

PR #269, opened 2024-02-14
gdbstub: take task ID into account when writing registers and other improvements

00xc:
Add some safety comments around pointer manipulation, fix a bug where the task ID was not taken into account when writing registers and other minor improvements.

### DIFF
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -186,6 +186,7 @@ pub mod svsm_gdbstub {
     }
 
     impl GdbTaskContext {
+        #[must_use = "The task switch will have no effect if the context is dropped"]
         fn switch_to_task(id: u32) -> Self {
             let cr3 = if is_current_task(id) {
                 0
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -341,15 +341,22 @@ pub mod svsm_gdbstub {
     }
 
     struct GdbStubTarget {
-        ctx: usize,
+        ctx: *mut TaskContext,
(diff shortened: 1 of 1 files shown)


kernel/src/debug/gdbstub.rs
```diff
@@ -362,8 +365,24 @@ pub mod svsm_gdbstub {
...
+        fn ctx_mut(&mut self) -> Option<&mut TaskContext> {
+            // SAFETY: this is a pointer to the exception context on the
+            // stack, so it is not aliased from a different task. We trust
+            // the debug exception handler to pass a well-aligned pointer
+            // pointing to valid memory.
+            unsafe { self.ctx.as_mut() }
+        }
+
+        fn set_regs(&mut self, ctx: &mut TaskContext) {
+            self.ctx = core::ptr::from_mut(ctx)
```
roy-hopkins, line 385: Just a note - with rustc 1.74.1 you get this error:

```
error[E0658]: use of unstable library feature 'ptr_from_ref'
   --> kernel/src/debug/gdbstub.rs:385:24
    |
385 |             self.ctx = core::ptr::from_mut(ctx)
    |                        ^^^^^^^^^^^^^^^^^^^
    |
```

Updating to 1.76.0 fixed it.

roy-hopkins (approved):
I've tested this by attaching the debugger, switching threads and attempting to read/write registers. Everything works as expected, including the error message that is displayed when trying to modify registers on the non-current task.

Many thanks for these improvements.

kernel/src/debug/gdbstub.rs
00xc: Yes, it recently got stabilized. We never had a discussion around the minimum supported version for the project, so I assume this is fine for the time being.

PR #271, opened 2024-02-14
Make #VC handling IST

p4zuu:
Making #VC handler working on IST stack is cleaner. However, it requires to properly handle side-effects, like the handling of nested #VC.

This PR:
 * makes #VC and #DB IST
 * add a few methods to easily find if a given address is on #VC IST stack memory region
 * handle nested IST #VC
 * makes a few comments in the VC code Rust idiomatic.

A few interrogations I would like to solve before merging:
 * please double-check that the [addresses](https://github.com/coconut-svsm/svsm/blob/f90083200b6576bc8934c4265ae0db88cfdfbe2c/kernel/src/mm/address_space.rs#L132) created for IST stacks are correct
 * is making #DB IST stricly necessary to make #VC IST?

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -5,13 +5,18 @@
 // Author: Joerg Roedel <jroedel@suse.de>
 
 use super::idt::common::X86ExceptionContext;
+use crate::address::VirtAddr;
 use crate::cpu::cpuid::{cpuid_table_raw, CpuidLeaf};
 use crate::cpu::ghcb::current_ghcb;
 use crate::cpu::insn::{insn_fetch, Instruction};
+use crate::cpu::percpu::this_cpu;
+use crate::cpu::registers::from_user;
 use crate::debug::gdbstub::svsm_gdbstub::handle_debug_exception;
 use crate::error::SvsmError;
 use crate::sev::ghcb::{GHCBIOSize, GHCB};
-use core::fmt;
+use crate::utils::align_down;
+use core::mem::size_of;
+use core::{fmt, ptr};
 
 pub const SVM_EXIT_EXCP_BASE: usize = 0x40;
 pub const SVM_EXIT_LAST_EXCP: usize = 0x5f;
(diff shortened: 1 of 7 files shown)


msft-jlange:
What is the rationale for using IST dispatch?  I believe we are on a path not to use SYSCALL due to the complexities it will introduce with TDX Partitioning, so there will be no RIP/RSP modification gap on user/kernel transitions.  There is no use of the GS segment in the kernel, so there is no SWAPGS gap on user/kernel transitions.  As a result, there should never be a window in which dispatch on the current stack is not possible.  Given these statements, is there any benefit for using IST dispatch for any of these exceptions?

Even if #VC dispatch is considered vulnerable to stack-based delivery challenges, why move #DB to an IST stack?  The considerations that compel other operating systems to use IST dispatch for #DB do not apply to the COCONUT-SVSM kernel.

I believe we will want to avoid IST usage as much as we can because of the challenges that IST dispatch poses to reentrant delivery, so we should be able to clearly articulate a valuable reason that is relevant within the COCONUT-SVSM architecture.

00xc:
> What is the rationale for using IST dispatch? I believe we are on a path not to use SYSCALL due to the complexities it will introduce with TDX Partitioning (...)

Perhaps this is a dumb question, but what's the alternative here for transitioning to CPL-0? I guess `int 0x80`-style interrupts?

msft-jlange:
> Perhaps this is a dumb question, but what's the alternative here for transitioning to CPL-0? I guess `int 0x80`-style interrupts?

Yes, the INT N instruction.  I expect that the performance gap between software interrupt dispatch and SYSCALL transitions has narrowed significantly since the SYSCALL instruction was first designed 20 years ago.  I can try to dig up numbers.

kernel/src/cpu/vc.rs
```diff
@@ -257,3 +256,42 @@ fn vc_decode_insn(ctx: &mut X86ExceptionContext) -> Result<Instruction, SvsmErro
...
+///
+/// [`VirtAddr`]: the new stack address. This address is directly moved to RSP in the caller.
+/// Therefore, the new address should absolutely be valid.
+#[no_mangle]
+extern "C" fn vc_switch_off_ist(ctx: &X86ExceptionContext) -> VirtAddr {
+    // Detect user #VC or nested IST #VC
+    let mut new_rsp = if from_user(ctx) || vc_on_ist_stack(ctx) {
+        this_cpu().current_stack.end()
+    } else {
+        VirtAddr::from(ctx.frame.rsp)
```
msft-jlange, line 277: This will not be correct in a SYSCALL gap.  If the #VC is delivered immediately after control transfers to kernel mode to the SYSCALL entry point, the CS on the exception frame will indicate a kernel mode CS, but the RSP in the exception frame will be the user-mode RSP since the SYSCALL entry point will not yet have been able to switch to the kernel-mode stack.  It would be a serious problem if this routine were to attempt to switch off of the IST stack back onto the user-mode stack.

Because the untrusted host can choose to remove the page backing the SYSCALL entry point and replace it with a different page, it is certainly possible for the first instruction at the SYSCALL entry point to raise #VC due to a page-not-validated error, so this must be anticipated.

A similar problem exists immediately prior to SYSRET.  A carefully timed interrupt by the untrusted host could cause the code page to disappear after the user RSP is reloaded but before SYSRET executes, and the resulting #VC will again appear to come from kernel mode on the user stack.  This code must also anticipate that possibility.  Of course, both of these cases are security attacks which could reasonably result in a panic, but that panic must occur before an inadvertent switch to the user stack.

p4zuu: > This will not be correct in a SYSCALL gap. If the #VC is delivered immediately after control transfers to kernel mode to the SYSCALL entry point, the CS on the exception frame will indicate a kernel mode CS, but the RSP in the exception frame will be the user-mode RSP since the SYSCALL entry point will not yet have been able to switch to the kernel-mode stack. It would be a serious problem if this routine were to attempt to switch off of the IST stack back onto the user-mode stack.

Indeed. I made the choice not to try to detect the SYSCALL gap here at the beginning of the `vc_switch_off_ist()` function since we still don't support SYSCALL handling, but that's something I had in mind and would have worked on when we have proper SYSCALL handling. I can try to detect and handle the syscall gap (of course if we go for the SYSCALL option instead of the software interrupt way you mentioned).

> A similar problem exists immediately prior to SYSRET. A carefully timed interrupt by the untrusted host could cause the code page to disappear after the user RSP is reloaded but before SYSRET executes, and the resulting #VC will again appear to come from kernel mode on the user stack. This code must also anticipate that possibility. Of course, both of these cases are security attacks which could reasonably result in a panic, but that panic must occur before an inadvertent switch to the user stack.

Indeed again, I was less aware of the same issue in SYSRET. Many thanks for the precision.
I guess spotting that the exception context's RIP is at the very early SYSCALL handler stage (or very end of SYSRET handler) would be a solution. Do you see any better way? And does relocating #VC handler's stack to current task's stack if we detect the SYSCALL gap look the proper way to you? Or would you find that panic would be safer?

p4zuu:
The current syscall handling developing plan goes to software interrupt instead of `syscall`/`sysret`. I think we can safely close this.

PR #273, opened 2024-02-16
console: Initialize serial port

AdamCDunlap:
Previously SVSM did not call the `init()` of serial port, thus did not set up the serial port. It worked fine with QEMU because QEMU's default serial port config complies with SVSM's usage. Google's hypervisor has different defaults, so it is necessary to initialize it.

Fix the problem by actually calling the `init()` at proper location.

Original patch by: Zixuan Wang <zxwang42@gmail.com>
Signed-off-by: Adam Dunlap <acdunlap@google.com>

### DIFF
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -46,6 +46,7 @@ pub mod svsm_gdbstub {
 
     pub fn gdbstub_start() -> Result<(), u64> {
         unsafe {
+            GDB_SERIAL.init();
             let mut target = GdbStubTarget::new();
             let gdb = GdbStubBuilder::new(GdbStubConnection::new())
                 .with_packet_buffer(&mut PACKET_BUFFER)
--- a/kernel/src/stage2.rs
+++ b/kernel/src/stage2.rs
@@ -114,6 +114,7 @@ fn setup_env(config: &SvsmConfig<'_>) {
             port: config.debug_serial_port(),
         })
         .expect("console serial output already configured");
+    (*CONSOLE_SERIAL).init();
 
     WRITER.lock().set(&*CONSOLE_SERIAL);
     init_console();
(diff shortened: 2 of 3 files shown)


kernel/src/console.rs
```diff
@@ -47,12 +47,16 @@ impl fmt::Debug for Console {
...
 
-pub static WRITER: SpinLock<Console> = SpinLock::new(Console {
+static WRITER: SpinLock<Console> = SpinLock::new(Console {
     writer: Some(&DEFAULT_SERIAL_PORT),
 });
 static CONSOLE_INITIALIZED: ImmutAfterInitCell<bool> = ImmutAfterInitCell::new(false);
 
-pub fn init_console() {
+pub fn init_console(writer: &'static SerialPort<'_>) {
+    if !*CONSOLE_INITIALIZED {
```
stefano-garzarella, line 56: I think we are changing the behavior.
Before this patch, the caller invoked `WRITER.lock().set()` without checking `CONSOLE_INITIALIZED`, so each time changed the serial port used by our console, here instead only the first invocation sets it, so I'm not sure if it's the behavior we wanted.

It is true that we call it always with the same serial port, so it should be okay, but I think here we are mixing the initialization of the serial port, with the initialization of our console. IMHO we should keep them separate.

AdamCDunlap: I didn't think about the case where you might want to change the console midway through execution, but that makes sense. I changed this PR to be much simpler and just call init() after constructing the SerialPort struct.

I thought about changing the existing SerialPort::new function to call init() and then disabling the "manual" construction of the SerialPort struct, but that doesn't work with how gdbstub creates the SerialPort.

stefano-garzarella: > I thought about changing the existing SerialPort::new function to call init() and then disabling the "manual" construction of the SerialPort struct, but that doesn't work with how gdbstub creates the SerialPort.

I like this idea. Maybe we should refactor a bit of the code, so I'd do in another PR.

stefano-garzarella (approved):

PR #275, opened 2024-02-19
repo: apply cargo fmt to doctests

00xc:
Using cargo fmt in doctests is a nightly feature, so we don't use it in CI. Thus, apply changes manually.


00xc:
> LGTM, but what about adding a job in our CI with the nightly toolchain to enable this? I can send a PR for that.

Done

PR #279, opened 2024-02-27
bootlib: Align on a single definition of the stage 2 stack

msft-jlange:
This change eliminates the separate definitions for the stage 2 stack parameters that were previously held in stage 2 and the IGVM builder and consolidates them into a single structure definition in the bootlib crate.


roy-hopkins (commented):
Looks good to me so far. All working with QEMU.

00xc (approved):

PR #287, opened 2024-03-08
Change build to use vendored dependencies

deeglaze:
Allows for a container-based build with no network access to build all of SVSM.

This PR as generated by running `cargo vendor` and adding the output changes to .cargo/config.toml.


deeglaze:
The build container label has to be known statically, so we would need to
have an extra released container image that is our toolchain container with
all svsm dependencies vendored. That can be done, but is not ideal for
transparency to have to use a specialized build container when it’s easier
to build trust in a single toolchain container label, like
gcc+rust-bullseye. If we’re going to go to the level of deeply inspecting
the toolchain container, we’re going beyond the typical slsa provenance
model.

Less of a problem is that we’d need to have a nontrivial build script that
edits the .cargo/config.toml file. That build script can be committed to
this repo though.

deeglaze:
I'll check with our Assured Open Source Software team whether this is the appropriate way to hermeticize Rust builds or if we need to create a crates.io and github.com mirror in our internal git host to avoid multiple versions of libraries scattered about.

joergroedel:
> I'll check with our Assured Open Source Software team whether this is the appropriate way to hermeticize Rust builds or if we need to create a crates.io and github.com mirror in our internal git host to avoid multiple versions of libraries scattered about.

The Cargo.lock file contains the git hashes for all dependencies. Is that not enough to verify their soundness?

deeglaze:
Git hashes are not source code. We need to have all source available internally on servers we control instead of relying on internet fetches that can then be verified against a lock file. That means we either mirror all transitive dependencies and create a private crates.io mirror to reference them, or SVSM vendors all its dependencies itself. That's a horizontal mandate for managed open source.

deeglaze:
I checked with the Assured Open Source Software team and we do not have to follow the one-version rule for open source dependencies. Vendoring is a much better option that setting up multiple mirrors.

joergroedel:
As discussed in our meeting on March 27th, 2024, we will try other workflows for SLSA-compliant builds. Closing this PR for now.

PR #288, opened 2024-03-08
sev/ghcb Add #VC handler tests

AdamCDunlap:
Adds tests that make sure a \#VC exception is raised while executing various instructions and that the #VC handler properly handles them via the ghcb. Instructions tested are cpuid and port IO.

Looking for feedback on the following:

- Is the raw-cpuid dependency OK, or would it be better to just use inline asm or something else?
- What's a good location to put these? I chose ghcb.rs, but they might make more sense somewhere else.
- I have tests for other \#VC handler cases such as ones for msrs, dr7, wbinvd, rdtsc, mmio, and other IOIO instructions, but they fail since those cases aren't handled in the svsm handler. I could add these as \#\[should_panic\] if we think they might be useful later.
- I did not test these with qemu since it doesn't work on hardware I have. The IOIO tests rely on a "testdev" device that echos characters sent to a specific port. I'm pretty sure qemu has a compatible device, but it might need to be explicitly enabled.

@p4zuu please take a look.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -260,8 +260,10 @@ fn vc_decoding_needed(error_code: usize) -> bool {
 
 #[cfg(test)]
 mod tests {
+    use crate::cpu::msr::{rdtsc, rdtscp, read_msr, write_msr, RdtscpOut};
     use crate::cpu::percpu::this_cpu_mut;
     use crate::sev::ghcb::GHCB;
+    use crate::sev::utils::{get_dr7, raw_vmmcall, set_dr7};
     use core::arch::asm;
     use core::arch::x86_64::__cpuid_count;
 
@@ -325,9 +327,20 @@ mod tests {
             ret
         }
     }
+    fn inb_from_testdev_echo() -> u8 {
+        unsafe {
+            let ret: u8;
+            asm!("inb $0xe0, %al", out("al") ret, options(att_syntax));
+            ret
+        }
+    }
 
     fn outb(port: u16, value: u8) {
         unsafe { asm!("outb %al, %dx", in("al") value, in("dx") port, options(att_syntax)) }
+    }
+
+    fn outb_to_testdev_echo(value: u8) {
+        unsafe { asm!("outb %al, $0xe0", in("al") value, options(att_syntax)) }
     }
 
     fn inw(port: u16) -> u16 {
@@ -337,9 +350,20 @@ mod tests {
             ret
         }
     }
+    fn inw_from_testdev_echo() -> u16 {
+        unsafe {
+            let ret: u16;
+            asm!("inw $0xe0, %ax", out("ax") ret, options(att_syntax));
+            ret
+        }
+    }
 
     fn outw(port: u16, value: u16) {
         unsafe { asm!("outw %ax, %dx", in("ax") value, in("dx") port, options(att_syntax)) }
+    }
+
+    fn outw_to_testdev_echo(value: u16) {
+        unsafe { asm!("outw %ax, $0xe0", in("ax") value, options(att_syntax)) }
     }
 
     fn inl(port: u16) -> u32 {
@@ -349,9 +373,26 @@ mod tests {
             ret
         }
     }
+    fn inl_from_testdev_echo() -> u32 {
+        unsafe {
+            let ret: u32;
+            asm!("inl $0xe0, %eax", out("eax") ret, options(att_syntax));
+            ret
+        }
+    }
 
     fn outl(port: u16, value: u32) {
         unsafe { asm!("outl %eax, %dx", in("eax") value, in("dx") port, options(att_syntax)) }
+    }
+
+    fn outl_to_testdev_echo(value: u32) {
+        unsafe { asm!("outl %eax, $0xe0", in("eax") value, options(att_syntax)) }
+    }
+
+    fn rep_outsw(port: u16, data: &[u16]) {
+        unsafe {
+            asm!("rep outsw", in("dx") port, in("rsi") data.as_ptr(), in("rcx") data.len(), options(att_syntax))
+        }
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


joergroedel:
One of the new tests triggers an assertion:

```
[SVSM] test sev::ghcb::test::test_port_io_16 ...
[SVSM] ERROR: Panic: CPU[0] panicked at kernel/src/sev/ghcb.rs:664:9:
  left: 17185
 right: 65535
```

Is that a known issue?

p4zuu:
> One of the new tests triggers an assertion:
> 
> ```
> [SVSM] test sev::ghcb::test::test_port_io_16 ...
> [SVSM] ERROR: Panic: CPU[0] panicked at kernel/src/sev/ghcb.rs:664:9:
> assertion `left == right` failed
>   left: 17185
>  right: 65535
> ```
> 
> Is that a known issue?

The current CPUID #VC handler actually doesn't use the GHCB. It directly looks into the registered CPUID table. That's maybe the reason, the GHCB is not modified while handling a CPUID.

p4zuu:
> * Is the raw-cpuid dependency OK, or would it be better to just use inline asm or something else?

If it's not too much work, I would actually prefer inlining asm here, just to reduce the supply-chain attack surface.
Since this is for testing only, I think we only need a minimal implementation.

> * What's a good location to put these? I chose ghcb.rs, but they might make more sense somewhere else.

I am unsure too. Actually the CPUID handling of the current #VC handler is not touching the GHCB since it directly reads into the registered CPUID table (instead of calling the HV), so that might be an argument to move this somewhere else. 
Moreover, the purpose of this is to verify the #VC handler's behaviors, so I would simply put it into `kernel/src/cpu/vc.rs`.

> * I have tests for other #VC handler cases such as ones for msrs, dr7, wbinvd, rdtsc, mmio, and other IOIO instructions, but they fail since those cases aren't handled in the svsm handler. I could add these as #[should_panic] if we think they might be useful later.

Yes, even if we currently don't support those, I agree with adding them as `#[should_panic]`.

> * I did not test these with qemu since it doesn't work on hardware I have. The IOIO tests rely on a "testdev" device that echos characters sent to a specific port. I'm pretty sure qemu has a compatible device, but it might need to be explicitly enabled.

I can try taking a look at this since I'm using qemu :)

00xc:
> > * Is the raw-cpuid dependency OK, or would it be better to just use inline asm or something else?
> 
> If it's not too much work, I would actually prefer inlining asm here, just to reduce the supply-chain attack surface. Since this is for testing only, I think we only need a minimal implementation.

Rust provides a CPUID function for convenience:
https://doc.rust-lang.org/beta/core/arch/x86_64/fn.__cpuid.html

> > * What's a good location to put these? I chose ghcb.rs, but they might make more sense somewhere else.
> 
> I am unsure too. Actually the CPUID handling of the current #VC handler is not touching the GHCB since it directly reads into the registered CPUID table (instead of calling the HV), so that might be an argument to move this somewhere else. Moreover, the purpose of this is to verify the #VC handler's behaviors, so I would simply put it into `kernel/src/cpu/vc.rs`.

I agree with this.

> > * I have tests for other #VC handler cases such as ones for msrs, dr7, wbinvd, rdtsc, mmio, and other IOIO instructions, but they fail since those cases aren't handled in the svsm handler. I could add these as #[should_panic] if we think they might be useful later.
> 
> Yes, even if we currently don't support those, I agree with adding them as `#[should_panic]`.

Perhaps something like `#[ignore = "not yet supported"]` would be better. `#[must_panic]` is not really supported inside the SVSM and these tests won't run outside.

AdamCDunlap:
> One of the new tests triggers an assertion:

This doesn't happen with the google hypervisor, so I'm assuming it's an issue with the qemu testdev device. I'll try to figure out if it intentionally doesn't support 16-bit echos, but if so then I may just need to disable this test or modify qemu. Since I don't have a way of testing with qemu, it'd be nice to get some help here.

> If it's not too much work, I would actually prefer inlining asm here, just to reduce the supply-chain attack surface.

Done

>  I would simply put it into kernel/src/cpu/vc.rs.

Done

> Yes, even if we currently don't support those, I agree with adding them as #[should_panic].

Added a bunch more tests. Unfortunately, these tests are untested, but should be a decent starting point once support for those operations are added in the \#VC handler.

roy-hopkins:
> > One of the new tests triggers an assertion:
> 
> This doesn't happen with the google hypervisor, so I'm assuming it's an issue with the qemu testdev device. I'll try to figure out if it intentionally doesn't support 16-bit echos, but if so then I may just need to disable this test or modify qemu. Since I don't have a way of testing with qemu, it'd be nice to get some help here.

I've investigated this using QEMU. You need to add the `pc-testdev` device to the QEMU configuration for the test port to be active. Applying this patch to your branch fixes the tests for me:

```
index b638303..9ebab2f 100755
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -14,6 +14,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 C_BIT_POS=`$SCRIPT_DIR/../utils/cbit`
 DEBUG_SERIAL=""
 QEMU_EXIT_DEVICE=""
+QEMU_TEST_IO_DEVICE=""
 
 while [[ $# -gt 0 ]]; do
   case $1 in
@@ -38,6 +39,7 @@ while [[ $# -gt 0 ]]; do
       ;;
     --unit-tests)
       QEMU_EXIT_DEVICE="-device isa-debug-exit,iobase=0xf4,iosize=0x04"
+      QEMU_TEST_IO_DEVICE="-device pc-testdev"
       shift
       ;;
     -*|--*)
@@ -110,6 +112,7 @@ $SUDO_CMD \
     -monitor none \
     -serial stdio \
     $DEBUG_SERIAL \
-    $QEMU_EXIT_DEVICE
+    $QEMU_EXIT_DEVICE \
+    $QEMU_TEST_IO_DEVICE
 
 stty intr ^C
```

kernel/src/cpu/vc.rs
```diff
@@ -257,3 +257,369 @@ fn vc_decode_insn(ctx: &X86ExceptionContext) -> Result<Instruction, SvsmError> {
...
+        let edx: u32;
+        asm!("rdmsr", out("eax") eax, out("edx") edx, in("ecx") ecx);
+        ((edx as u64) << 32) | (eax as u64)
+    }
+
+    unsafe fn wrmsr(ecx: u32, value: u64) {
+        let eax = value as u32;
+        let edx = (value >> 32) as u32;
+        asm!("wrmsr", in("eax") eax, in("edx") edx, in("ecx") ecx);
+    }
```
00xc, line 571: We already have `read_msr()` / `write_msr()` in `kernel/src/cpu/msr.rs`, could you reuse those?

kernel/src/cpu/vc.rs
```diff
@@ -257,3 +257,369 @@ fn vc_decode_insn(ctx: &X86ExceptionContext) -> Result<Instruction, SvsmError> {
...
+        unsafe {
+            // The count param is 1 to only write one ghcb's worth of data
+            core::ptr::write_bytes(ghcb, GHCB_FILL_TEST_VALUE, 1);
+        }
+    }
+    fn verify_ghcb_was_altered() {
+        let ghcb = this_cpu_mut().ghcb_unsafe();
+        let ghcb_bytes: &[u8; core::mem::size_of::<GHCB>()] = unsafe { &*(ghcb as *const _) };
+        let changed_byte = ghcb_bytes.iter().find(|&&v| v != GHCB_FILL_TEST_VALUE);
+        assert!(changed_byte.is_some());
```
00xc, line 306: This is creating an intermediate reference to the GHCB. I'd favor just using `slice::from_raw_parts()`, which is also less convoluted in my opinion. Also, you can test that at least one byte changed more easily with `Iterator::any()`:

```suggestion
        let ghcb_bytes =
            unsafe { core::slice::from_raw_parts(ghcb.cast::<u8>(), core::mem::size_of::<GHCB>()) };
        assert!(ghcb_bytes.iter().any(|v| *v != GHCB_FILL_TEST_VALUE));
```

00xc (commented):
Just a couple of small comments from my side

AdamCDunlap:
> You need to add the pc-testdev device to the QEMU configuration for the test port to be active. 

Thanks! I applied that commit to this PR

> We already have read_msr() / write_msr() in kernel/src/cpu/msr.rs, could you reuse those?

Whoops, didn't notice them. Changed to use those.

> I'd favor just using slice::from_raw_parts(), which is also less convoluted in my opinion. Also, you can test that at least one byte changed more easily with Iterator::any():

Thanks, I applied that change.

00xc (approved):

p4zuu:
LGTM too, thanks Adam!

PR #292, opened 2024-03-13
Fix Hyper-V IGVM without firmware

msft-jlange:
In the Hyper-V configuration, the IGVM builder places the kernel image just above the firmware image.  If there is no firmware image, this results in the kernel image being placed at address zero.  This overlaps with where stage 2 is loaded, and that results in corruption of the kernel image and/or the IGVM parameter block depending on the extent of the collision.  Moving the kernel to a minimum address of 1 MB ensures that there is no collision.

### DIFF
--- a/igvmbuilder/src/gpa_map.rs
+++ b/igvmbuilder/src/gpa_map.rs
@@ -111,8 +111,15 @@ impl GpaMap {
                 1 << 20
             }
             Hypervisor::HyperV => {
-                // Load the kernel image after the firmware.
-                firmware_range.get_end()
+                // Load the kernel image after the firmware, but now lower than
+                // 1 MB.
+                let firmware_end = firmware_range.get_end();
+                let addr_1mb = 1 << 20;
+                if firmware_end < addr_1mb {
+                    addr_1mb
+                } else {
+                    firmware_end
+                }
             }
         };
         let kernel_elf = GpaRange::new(kernel_address, kernel_elf_len as u64)?;


igvmbuilder/src/gpa_map.rs
```diff
@@ -111,8 +111,15 @@ impl GpaMap {
...
-                firmware_range.get_end()
+                // Load the kernel image after the firmware, but now lower than
+                // 1 MB.
+                let firmware_end = firmware_range.get_end();
+                let addr_1mb = 1 << 20;
+                if firmware_end < addr_1mb {
+                    addr_1mb
+                } else {
+                    firmware_end
+                }
```
00xc, line 122: Or just:

```rust
let firmware_end = firmware_range.get_end().max(1 << 20);
```

Anyway this LGTM.

00xc (approved):

igvmbuilder/src/gpa_map.rs
00xc: Feel free to mark this as resolved, it's just a nitpick.

roy-hopkins (approved):
LGTM too.

igvmbuilder/src/gpa_map.rs
msft-jlange: Thanks for the tip - that was a Rust-ism that I was not familiar with.  I'd like to stick with the existing version for now just because it's slightly more readable for those who are not deeply familiar with all of the available Rust tricks.

PR #293, opened 2024-03-13
Prevent use of ImmutAfterInit in the multithreaded environment

msft-jlange:
The initialization of `ImmutAfterInitCell` objects (and, by extension, `ImmutAfterInitRef` objects) is designed specifically only to safe when the system is still in a single-threaded environment, because the intialization checks are not thread-safe.  This change ensures that no attempt is made to initialize or reinitialize any of these objects once the system has started multiple processors.  These checks are debug-only, like the checks on initialization state.

### DIFF
--- a/kernel/src/console.rs
+++ b/kernel/src/console.rs
@@ -52,7 +52,9 @@ pub static WRITER: SpinLock<Console> = SpinLock::new(Console {
 static CONSOLE_INITIALIZED: ImmutAfterInitCell<bool> = ImmutAfterInitCell::new(false);
 
 pub fn init_console() {
-    CONSOLE_INITIALIZED.reinit(&true);
+    CONSOLE_INITIALIZED
+        .reinit(&true)
+        .expect("could not reinit CONSOLE_INITIALIZED");
 }
 
 #[doc(hidden)]
--- a/kernel/src/cpu/smp.rs
+++ b/kernel/src/cpu/smp.rs
@@ -10,6 +10,7 @@ use crate::cpu::percpu::{this_cpu_mut, PerCpu};
 use crate::cpu::vmsa::init_svsm_vmsa;
 use crate::requests::{request_loop, request_processing_main};
 use crate::task::{create_kernel_task, schedule_init, TASK_FLAG_SHARE_PT};
+use crate::utils::immut_after_init::immut_after_init_set_multithreaded;
 
 fn start_cpu(apic_id: u32, vtom: u64) {
     unsafe {
@@ -44,6 +45,7 @@ fn start_cpu(apic_id: u32, vtom: u64) {
 }
 
 pub fn start_secondary_cpus(cpus: &[ACPICPUInfo], vtom: u64) {
+    immut_after_init_set_multithreaded();
     let mut count: usize = 0;
     for c in cpus.iter().filter(|c| c.apic_id != 0 && c.enabled) {
         log::info!("Launching AP with APIC-ID {}", c.apic_id);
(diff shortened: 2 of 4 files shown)


00xc:
Should we perhaps use `Ordering::SeqCst` when loading/storing into `MULTI_THREADED`? It should have no impact before bringing up the secondary CPUs anyway.

msft-jlange:
> Should we perhaps use `Ordering::SeqCst` when loading/storing into `MULTI_THREADED`? It should have no impact before bringing up the secondary CPUs anyway.

Since this is just for debug checking, the ordering applied to these atomics is irrelevant.  Ordering rules simply define which values are observable before/after which other values in a weakly ordered memory system.  The whole point of this change is not to guarantee consistency of data, but simply to detect whether we have passed the point of multi-threaded init.  In other words, since `ImmutAfterInitCell::init()` is expressly designed NOT to be thread-safe, no changes to ordering rules could improve thread safety of its use, so sticking with relaxed ordering is the most sensible.

00xc (approved):

joergroedel:
For some reason this triggers a boot failure on my side when bringing up secondary CPUs:
```
[SVSM] Launching AP with APIC-ID 21
[SVSM] Launching request-processing task on CPU 20
[SVSM] ERROR: Panic: CPU[0] panicked at kernel/src/cpu/smp.rs:38:14:
Failed to launch secondary CPU: Ghcb(VmgexitError(2, 5))
[SVSM] ---BACKTRACE---:
[SVSM]   [0xffffff800000c8f1]
[SVSM]   [0xffffff800000cb15]
[SVSM]   [0xffffff8000003738]
[SVSM]   Invalid frame
[SVSM] ---END---
```

KVM complains about a 2M-aligned VMSA:

```
kvm_amd: kvm [36813]: vcpu0, guest rIP: 0x0 vmgexit: AP VMSA address [8000c00000] from guest is unsafe as it is 2M aligned
```

Not sure yet how that can happen, as the VMSA allocation code makes sure the VMSA is not 2M-aligned.

00xc:
> For some reason this triggers a boot failure on my side when bringing up secondary CPUs:
> 
> ```
> [SVSM] Launching AP with APIC-ID 21
> [SVSM] Launching request-processing task on CPU 20
> [SVSM] ERROR: Panic: CPU[0] panicked at kernel/src/cpu/smp.rs:38:14:
> Failed to launch secondary CPU: Ghcb(VmgexitError(2, 5))
> [SVSM] ---BACKTRACE---:
> [SVSM]   [0xffffff800000c8f1]
> [SVSM]   [0xffffff800000cb15]
> [SVSM]   [0xffffff8000003738]
> [SVSM]   Invalid frame
> [SVSM] ---END---
> ```
> 
> KVM complains about a 2M-aligned VMSA:
> 
> ```
> kvm_amd: kvm [36813]: vcpu0, guest rIP: 0x0 vmgexit: AP VMSA address [8000c00000] from guest is unsafe as it is 2M aligned
> ```
> 
> Not sure yet how that can happen, as the VMSA allocation code makes sure the VMSA is not 2M-aligned.

That's odd, I can boot a guest kernel fine with and without this PR.

joergroedel:
> That's odd, I can boot a guest kernel fine with and without this PR.

I seems not to be related to this PR. I found an issue where sometimes there is an off-by-one-page bug between the virtual and physical addresses where the heap starts. That causes the offset to not be 2M aligned anymore and the SVSM might allocate a 2m aligned virtual address which is 2m aligned physical. The bug originates somewhere in stage2, I am still investigating.

joergroedel:
The issue has been fixed by PR #301

PR #294, opened 2024-03-14
Ensure IGVM output is compatible with QEMU and calculate launch measurement

roy-hopkins:
This PR introduces a new tool `igvmmeasure` that parses an IGVM file and calculates the launch digest which should match the launch measurement included as part of the attestation report from the platform.

By introducing this tool, it created the opportunity to verify that the directives within the IGVM file produced by the build process are properly sent as initial guest state, exactly as described and in the correct order. It was found that there were some discrepancies between the current IGVM file and the QEMU loader/KVM launch process that resulted in a mismatch between the expected and actual digest.

The first two commits in this PR fix these issues which both relate to the VMSA then proceed to introduce the measurement tool and introduce it into the build. An option has also been added to the tool to catch these inconsistencies in future changes to the IGVM file, causing the build to abort if detected.

Note that in order to obtain a matching measurement, fixes are required on top of the existing coconut/qemu. I will be raising a new PR with these changes shortly.

### DIFF
--- a/igvmbuilder/src/igvm_builder.rs
+++ b/igvmbuilder/src/igvm_builder.rs
@@ -71,11 +71,15 @@ impl IgvmBuilder {
         directive_a: &IgvmDirectiveHeader,
         directive_b: &IgvmDirectiveHeader,
     ) -> Ordering {
-        let IgvmDirectiveHeader::PageData { gpa: a, .. } = directive_a else {
-            panic!("Attempted to compare non-page directive");
+        let a = match directive_a {
+            IgvmDirectiveHeader::PageData { gpa, .. }
+            | IgvmDirectiveHeader::SnpVpContext { gpa, .. } => gpa,
+            _ => panic!("Attempted to compare non-page/vp directive"),
         };
-        let IgvmDirectiveHeader::PageData { gpa: b, .. } = directive_b else {
-            panic!("Attempted to compare non-page directive");
+        let b = match directive_b {
+            IgvmDirectiveHeader::PageData { gpa, .. }
+            | IgvmDirectiveHeader::SnpVpContext { gpa, .. } => gpa,
+            _ => panic!("Attempted to compare non-page/vp directive"),
         };
         a.cmp(b)
     }
@@ -268,7 +272,7 @@ impl IgvmBuilder {
             },
         ));
 
-        // Place the VMSA at the base of the kernel region.
+        // Add the VMSA.
         self.directives.push(construct_vmsa(
             self.gpa_map.vmsa.get_start(),
             param_block.vtom,
@@ -414,5 +418,6 @@ impl IgvmBuilder {
 
     fn filter_pages(directive: &IgvmDirectiveHeader) -> bool {
         matches!(directive, IgvmDirectiveHeader::PageData { .. })
+            || matches!(directive, IgvmDirectiveHeader::SnpVpContext { .. })
     }
 }
--- a/igvmmeasure/src/igvm_measure.rs
+++ b/igvmmeasure/src/igvm_measure.rs
@@ -1,0 +1,233 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2024 SUSE LLC
+//
+// Author: Roy Hopkins <roy.hopkins@suse.com>
+
+use std::error::Error;
+use std::fs;
+
+use igvm::snp_defs::SevVmsa;
+use igvm::{IgvmDirectiveHeader, IgvmFile, IgvmPlatformHeader};
+use igvm_defs::{IgvmPageDataFlags, IgvmPageDataType, IgvmPlatformType, PAGE_SIZE_4K};
+use zerocopy::AsBytes;
+
+use crate::cmd_options::CmdOptions;
+use crate::page_info::PageInfo;
+
+#[derive(PartialEq)]
+enum SnpPageType {
+    None,
+    Normal,
+    Unmeasured,
+    Zero,
+    Secrets,
+    CpuId,
+    Vmsa,
+}
+
+impl std::fmt::Display for SnpPageType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                SnpPageType::None => "None",
+                SnpPageType::Normal => "Normal",
+                SnpPageType::Unmeasured => "Unmeasured",
+                SnpPageType::Zero => "Zero",
+                SnpPageType::Secrets => "Secrets",
+                SnpPageType::CpuId => "Cpuid",
+                SnpPageType::Vmsa => "VMSA",
+            }
+        )
+    }
+}
+
+pub struct IgvmMeasure<'a> {
+    options: &'a CmdOptions,
+    digest: [u8; 48],
+    last_page_type: SnpPageType,
+    last_gpa: u64,
+    last_next_gpa: u64,
+    last_len: u64,
+    compatibility_mask: u32,
+}
+
+const PAGE_SIZE_2M: u64 = 2 * 1024 * 1024;
+
+impl<'a> IgvmMeasure<'a> {
+    pub fn new(options: &'a CmdOptions) -> Self {
+        Self {
+            options,
+            digest: [0u8; 48],
+            last_page_type: SnpPageType::None,
+            last_gpa: 0,
+            last_next_gpa: 0,
+            last_len: 0,
+            compatibility_mask: 0,
+        }
+    }
+
+    fn find_compatibility_mask(&mut self, igvm: &IgvmFile) -> Result<(), Box<dyn Error>> {
+        for platform in igvm.platforms() {
+            let IgvmPlatformHeader::SupportedPlatform(platform) = platform;
+            match platform.platform_type {
+                IgvmPlatformType::SEV_SNP => {
+                    self.compatibility_mask = platform.compatibility_mask;
+                    return Ok(());
+                }
+                _ => continue,
+            }
+        }
+        Err("IGVM file is not compatible with the specified platform.".into())
+    }
+
+    pub fn measure(&mut self) -> Result<[u8; 48], Box<dyn Error>> {
+        let igvm_buffer = fs::read(&self.options.igvm_file).map_err(|e| {
+            eprintln!("Failed to open firmware file {}", self.options.igvm_file);
+            e
+        })?;
+        let igvm = IgvmFile::new_from_binary(igvm_buffer.as_bytes(), None)?;
+
+        self.find_compatibility_mask(&igvm)?;
+
+        for directive in igvm.directives() {
+            match directive {
+                IgvmDirectiveHeader::PageData {
+                    gpa,
+                    compatibility_mask,
+                    flags,
+                    data_type,
+                    data,
+                } => {
+                    if (*compatibility_mask & self.compatibility_mask) != 0 {
+                        self.measure_page(*gpa, flags, *data_type, data)?;
+                    }
+                }
+                IgvmDirectiveHeader::ParameterInsert(param) => {
+                    if (param.compatibility_mask & self.compatibility_mask) != 0 {
+                        self.measure_page(
+                            param.gpa,
+                            &IgvmPageDataFlags::new().with_unmeasured(true),
+                            IgvmPageDataType::NORMAL,
+                            &[],
+                        )?
+                    }
+                }
+                IgvmDirectiveHeader::SnpVpContext {
+                    gpa,
+                    compatibility_mask,
+                    vp_index: _vp_index,
+                    vmsa,
+                } => {
+                    if (*compatibility_mask & self.compatibility_mask) != 0 {
+                        self.measure_vmsa(*gpa, vmsa);
+                    }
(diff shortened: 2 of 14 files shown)


roy-hopkins:
@msft-jlange Please can you check if the changes I've made here affect Hyper-V: https://github.com/coconut-svsm/svsm/pull/294/commits/43ebc3b4d9370664ae14a897e31997c355d8e6d2

I can make these hypervisor specific if necessary.

igvmbuilder/src/igvm_builder.rs
```diff
@@ -71,11 +71,19 @@ impl IgvmBuilder {
...
         };
-        let IgvmDirectiveHeader::PageData { gpa: b, .. } = directive_b else {
-            panic!("Attempted to compare non-page directive");
+        let b = if let IgvmDirectiveHeader::PageData { gpa, .. } = directive_b {
+            gpa
+        } else if let IgvmDirectiveHeader::SnpVpContext { gpa, .. } = directive_b {
+            gpa
+        } else {
+            panic!("Attempted to compare non-page/vp directive");
         };
```
00xc, line 87: Same thing, a match here is a bit more clear

igvmbuilder/src/igvm_builder.rs
```diff
@@ -71,11 +71,19 @@ impl IgvmBuilder {
...
     ) -> Ordering {
-        let IgvmDirectiveHeader::PageData { gpa: a, .. } = directive_a else {
-            panic!("Attempted to compare non-page directive");
+        let a = if let IgvmDirectiveHeader::PageData { gpa, .. } = directive_a {
+            gpa
+        } else if let IgvmDirectiveHeader::SnpVpContext { gpa, .. } = directive_a {
+            gpa
+        } else {
+            panic!("Attempted to compare non-page/vp directive");
         };
```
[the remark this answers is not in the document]
roy-hopkins: Much neater. Thanks.

igvmmeasure/src/igvm_measure.rs
```diff
@@ -0,0 +1,279 @@
...
+                            be configured to match the value set by KVM."
+                    .into());
+            }
+            if !vmsa.sev_features.debug_swap() {
+                return Err("KVM check failure: DEBUG_SWAP(0x20) is not set in \
+                            sev_features in the VMSA file. This feature is \
+                            automatically applied by KVM during the initial \
+                            measurement so it must be specified in the VMSA in \
+                            the IGVM file in order to match."
+                    .into());
```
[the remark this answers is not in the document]
roy-hopkins: I had to implement std::fmt::Debug manually as well otherwise just the enum type was displayed.

igvmbuilder/src/igvm_builder.rs
roy-hopkins: Duplicate comment?

roy-hopkins:
@osteffenrh - I heard on the OC3 talk yesterday that you are planning to update https://github.com/virtee/sev-snp-measure to work with IGVM. You might want to take a look at this PR to see if it meets your needs.

osteffenrh:
> @osteffenrh - I heard on the OC3 talk yesterday that you are planning to update https://github.com/virtee/sev-snp-measure to work with IGVM. You might want to take a look at this PR to see if it meets your needs.

@roy-hopkins, Maybe that is no longer needed, now that you are contributing this PR here. :-D
We pre-recoded the talk a while back...

msft-jlange:
> @msft-jlange Please can you check if the changes I've made here affect Hyper-V: [43ebc3b](https://github.com/coconut-svsm/svsm/commit/43ebc3b4d9370664ae14a897e31997c355d8e6d2)

We have test environments in which we need to be able to boot Coconut without executing any guest requests, and your changes to obtain an attestation report break that environment.  Can you please add some IGVM parameter (ideally triggered from the command line) which can be used to suppress the guest request?

I'm aware of other future use cases for Coconut where boot-time guest requests will prove problematic, so I think we want to avoid establishing a pattern where guest requests are invoked unless we know that the execution environment is one that demands their use.  Encoding this in the IGVM parameters would serve to capture that.

Also, I'm not entirely clear on why you want to capture the VMPL 0 measurement via attestation.  If you generate an SNP_ID_BLOCK as part of the IGVM builder, then you can prevent the VM from even launching if its measurement is incorrect.  At some point, it would be helpful to understand what problem you're trying to solve by performing this early attestation and we can see if there is an equivalent way to solve the problem that doesn't require capturing an attestation report.

roy-hopkins:
> Also, I'm not entirely clear on why you want to capture the VMPL 0 measurement via attestation. If you generate an SNP_ID_BLOCK as part of the IGVM builder, then you can prevent the VM from even launching if its measurement is incorrect. At some point, it would be helpful to understand what problem you're trying to solve by performing this early attestation and we can see if there is an equivalent way to solve the problem that doesn't require capturing an attestation report.

The generation of the attestation report at this stage is purely informational and was added after discussions around making progress towards generating SNP_ID_BLOCKs. It serves as a rehearsal for possible future functionality where the SVSM will use the attestation report to request access to keys via the host.

At this stage I don't think there is any need to complicate this by making it conditional and I'd be happy to drop this commit from the PR.

jepio:
@roy-hopkins: how about adding a `igvmmeasure/.cargo/config.toml`:
```
[build]
target = "x86_64-unknown-linux-gnu"
```
so that `cargo build` can be directly executed from the `igvmmeasure` directory?

malt3:
There seems to be something off here. 
I just wrote my own tool that should perform the same job. So far, I was only able to successfully extract the `ld` value from the id block:

https://github.com/microsoft/igvm/blob/ac37f5e5658262f3e377f3f6019c9e631a32f644/igvm/src/lib.rs#L706

I have an IGVM file I can boot, that has a different value for the launch measurement in the id block than what igvmmeasure returns.

After launching the VM, reports I get from the VM also show the measurement value from the id block.

I think this tool is measuring something incorrectly.

This is an IGVM file I got from an Azure AKS node for testing purposes:

igvmmeasure/src/igvm_measure.rs
```diff
@@ -0,0 +1,333 @@
...
+    ) {
+        let page_info = match data_type {
+            IgvmPageDataType::NORMAL => {
+                if flags.unmeasured() {
+                    self.log_page(SnpPageType::Unmeasured, gpa, PAGE_SIZE_4K);
+                    Some(PageInfo::new_unmeasured_page(self.digest, gpa))
+                } else if data.is_empty() {
+                    self.log_page(SnpPageType::Zero, gpa, PAGE_SIZE_4K);
+                    Some(PageInfo::new_zero_page(self.digest, gpa))
+                } else {
```
jepio, line 274: this is not a legal transformation, we need to insert this as an actual page of zeros:
```suggestion
                } else if data.is_empty() {
                    self.log_page(SnpPageType::Normal, gpa, PAGE_SIZE_4K);
                    Some(PageInfo::new_normal_page(self.digest, gpa, &self.zero_page))
                } else {
```
and then you would cache `zero_page: vec![0u8; PAGE_SIZE_4K as usize]` somewhere.

malt3:
Would it be a good idea to print a warning (or even return with a non-zero exit code) if the `ld` value does not match with the result of measuring the actual pages?
It feels to me like this might catch other issues at some point.

igvmmeasure/src/igvm_measure.rs
roy-hopkins: The IGVM loader in QEMU performs the exact same transformation, converting empty pages to `SnpPageType::Zero` so in this case the measurement performed by the firmware is updated in accordance with the same rules as the measure tool and we get matching measurements for the COCONUT IGVM files and QEMU.

The SNP Firmware ABI specification describes the behaviour of the firmware for `PAGE_TYPE_ZERO` as equivalent to inserting a page of zeros - the resulting guest memory is the same, just the measurement is affected. I would assume that there is a performance benefit in using `PAGE_TYPE_ZERO` resulting in faster guest startup but I'm unsure.

So this transformation is legal, its just that the virtualisation stack you are using is interpreting IGVM directives differently to QEMU.

roy-hopkins:
By chance at the same time as @jepio's observation about zero pages above, I managed to get the output of `igvmmeasure` to match the ID block LD value by experimenting with measuring a page of zeros instead of using `PAGE_TYPE_ZERO`. 

As far as I am concerned, the QEMU behaviour is correct and we shouldn't be populating a page of zeros into guest memory when we encounter empty page data in the IGVM file, but should instead use the SNP page type of `PAGE_TYPE_ZERO`. However, in order to support the virtualisation stack you are using I can add an option to the measure tool to change the behaviour when measuring these pages.

roy-hopkins:
> Would it be a good idea to print a warning (or even return with a non-zero exit code) if the `ld` value does not match with the result of measuring the actual pages? It feels to me like this might catch other issues at some point.

Yes, that does make sense. The COCONUT IGVM builder does not generate an ID block yet but it will at some point, so I'll add a check that it matches if it is present.

igvmmeasure/src/igvm_measure.rs
jepio: I get your point that the outcome is equivalent - but is it worth diverging? The igvm file does say "PAGE_TYPE_NORMAL" so optimizing that into a "PAGE_TYPE_ZERO" feels like a loose interpretation of the file format. I don't know if TDX supports the equivalent of "PAGE_TYPE_ZERO" either.

At this stage we could totally still update Qemu igvm support to match what cloud-hypervisor and other igvm stacks do, it won't be possible to change later.

malt3:
> However, in order to support the virtualisation stack you are using I can add an option to the measure tool to change the behaviour when measuring these pages.

I would appreciate that! I personally can’t comment on ABI compliance (just a user of the platform). @jepio is probably in a better position to comment on that.

The virtualization stack is cloud-hypervisor with the hyperv backend on Azure with nested virt (the Azure AKS CoCo offering)

igvmmeasure/src/igvm_measure.rs
roy-hopkins: I was following the IGVM specification itself, documented [here under the `file_offset` field](https://docs.rs/igvm_defs/0.1.6/igvm_defs/struct.IGVM_VHS_PAGE_DATA.html#fields). 

> If the file offset is zero, then a page of zeroes is to be loaded. If the page contents are to be measured, and the underlying platform supports measured zero pages as a native page type, then the loader must request this page type; otherwise, the loader must generate a page of zeroes to load as the page contents.

jepio: Yikes, you're right. Looks like cloud-hypervisor got it wrong when they copied the mshv stack. That ship may have sailed though, what do you think @msft-jlange?

msft-jlange: Roy's interpretation of the spec is definitely what the spec was intended to do.  However, it appears that other IGVM loaders do not follow that interpretation of the spec.  For example, the Hyper-V loader will supply 4 KB of zeroes with PAGE_TYPE_NORMAL when it sees a zero file offset.  Since this behavior is already in production, it is not practical to declare that the production behavior is buggy because it will cause incompatibilities with IGVM files already in production.  I will plan to change the IGVM spec and I will introduce some sort of new flag information to indicate when a page of zeroes *must* be supplied as PAGE_TYPE_ZERO.

kernel/src/greq/services.rs
```diff
@@ -104,3 +107,31 @@ pub fn get_regular_report(buffer: &mut [u8]) -> Result<usize, SvsmReqError> {
...
+///     * [`SvsmReqError`]
+pub fn log_attestation() -> Result<(), SvsmReqError> {
+    let vaddr = allocate_zeroed_page().unwrap();
+    let buffer = unsafe { from_raw_parts_mut(vaddr.as_mut_ptr::<u8>(), PAGE_SIZE) };
+
+    let result = match get_regular_report(buffer) {
+        Ok(response_len) => {
+            assert_eq!(response_len, size_of::<SnpReportResponse>());
+            let response = SnpReportResponse::try_from_as_ref(buffer).unwrap();
+            let report = response.report().expect("Invalid attestation report");
```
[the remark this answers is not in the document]
roy-hopkins: The commit containing this has now been removed from this PR.

igvmmeasure/src/igvm_measure.rs
roy-hopkins: In anticipation of the changes to the IGVM specification, I've added an option to the `igvmmeasure` tool to allow the behaviour of handling of zero pages to be changed via a command-line option. This now defaults to using the normal page type rather than the zero page type.

The QEMU branch currently used by COCONUT uses the native zero page type so `Makefile` sets this new option to measure using zero pages. I'll update QEMU then raise a new PR to remove the option (or modify this one if it has not yet merged).

roy-hopkins:
Changes in the latest push:

* Dropped the commit that obtains an attestation report in SVSM kernel startup and logs it to the console.
* Added igvmmeasure/.cargo/config.toml as per @jepio's suggestion.
* Added `native-zero` option to determine how to handle pages that contain only zeros.
* Added check for matching measurement in ID block if present in the IGVM file, and an option to disable the check if required.

This hopefully resolves all the current comments on the PR.

I see a rebase on main is required so I'll push this shortly.

PR #295, opened 2024-03-15
cpu/vc: handle faults when fecthing instructions

p4zuu:
Faults can happen in the #VC handler if the exception context RIP (more generally the [RIP; RIP+15] array) doesn't belong to a mapped page.

To handle these faults, we can rely on GuestPtr::read() that uses the exception table if a fault occurs.

This touches the #VC handler, so I would wait for #288 to be merged to correctly test this PR.

### DIFF
--- a/kernel/src/cpu/insn.rs
+++ b/kernel/src/cpu/insn.rs
@@ -184,24 +184,6 @@ impl Instruction {
     }
 }
 
-/// Copy the instruction bytes where rip points to.
-///
-/// # Arguments
-///
-/// rip: instruction pointer as [`*const u8`].
-///
-/// # Returns
-///
-/// [`[u8; MAX_INSN_SIZE]`]: the 15-byte buffer where rip points to.
-///
-/// # Safety
-///
-///  The caller should validate that `rip` is set to a valid address
-///  and that the next [`MAX_INSN_SIZE`] bytes are within valid memory.
-pub unsafe fn insn_fetch(rip: *const u8) -> [u8; MAX_INSN_SIZE] {
-    rip.cast::<[u8; MAX_INSN_SIZE]>().read()
-}
-
 #[cfg(test)]
 mod tests {
     use super::{InsnBuffer, Instruction, MAX_INSN_SIZE};
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -248,4 +248,25 @@ mod tests {
 
         assert_eq!(test_buffer[0], data_to_write);
     }
+
+    #[test]
+    #[cfg_attr(miri, ignore = "inline assembly")]
+    fn test_read_15_bytes_valid_address() {
+        let test_buffer = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
+        let test_addr = VirtAddr::from(test_buffer.as_ptr());
+        let ptr: GuestPtr<[u8; 15]> = GuestPtr::new(test_addr);
+        let result = ptr.read().unwrap();
+
+        assert_eq!(result, test_buffer);
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore = "inline assembly")]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_read_invalid_address() {
+        let ptr: GuestPtr<u8> = GuestPtr::new(VirtAddr::new(0xDEAD_BEEF));
+
+        let err = ptr.read();
+        assert!(err.is_err());
+    }
 }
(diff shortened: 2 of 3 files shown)


kernel/src/mm/guestmem.rs
```diff
@@ -248,4 +248,15 @@ mod tests {
...
+    #[test]
+    #[cfg_attr(miri, ignore = "inline assembly")]
+    fn test_read_15_bytes_valid_address() {
+        let test_buffer = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
+        let test_addr = VirtAddr::from(test_buffer.as_ptr());
+        let ptr: GuestPtr<[u8; 15]> = GuestPtr::new(test_addr);
+        let result = ptr.read().unwrap();
+
+        assert_eq!(result, test_buffer);
+    }
```
00xc, line 261: This looks good. While you are at it, you could probably add another test that reads invalid memory and that only runs inside the SVSM.

kernel/src/cpu/insn.rs
```diff
@@ -198,8 +199,10 @@ impl Instruction {
 ///
 ///  The caller should validate that `rip` is set to a valid address
 ///  and that the next [`MAX_INSN_SIZE`] bytes are within valid memory.
-pub unsafe fn insn_fetch(rip: *const u8) -> [u8; MAX_INSN_SIZE] {
-    rip.cast::<[u8; MAX_INSN_SIZE]>().read()
+pub unsafe fn insn_fetch(
+    rip: GuestPtr<[u8; MAX_INSN_SIZE]>,
+) -> Result<[u8; MAX_INSN_SIZE], SvsmError> {
+    rip.read()
 }
```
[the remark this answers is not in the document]
p4zuu: You mean adding an `#[inline]` attribute or get rid of the `insn_fetch()` additional layer? I chose to leave it to make it clear that this is the instruction fetching part, but calling `read()` directly is probably clear enough :)

kernel/src/mm/guestmem.rs
p4zuu: Yes good idea!

kernel/src/cpu/insn.rs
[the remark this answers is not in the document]
00xc: Yes I meant removing the function directly. You could just add a comment that explains what that line does if you feel it's not clear enough.

kernel/src/mm/guestmem.rs
p4zuu: I added 2 new tests, but `make test-in-svsm` is currently broken on my environment.

kernel/src/mm/guestmem.rs
```diff
@@ -259,4 +262,28 @@ mod tests {
...
+    fn test_read_invalid_region() {
+        let _test_root_mem = TestRootMem::setup(PAGE_SIZE);
+        let page = allocate_page().unwrap();
+
+        // ptr is starting on a mapped page and ends on unmapped memory
+        let ptr: GuestPtr<[u8; 0x10]> = GuestPtr::new(page + (PAGE_SIZE - 8));
+
+        let err = ptr.read();
+        assert!(err.is_err());
+    }
```
00xc, line 288: This test fails - the memory beyond the allocated page is not lent by the allocator, but it's mapped as part of its memory region. I suggest you simply remove this test.

kernel/src/mm/guestmem.rs
```diff
@@ -248,4 +251,39 @@ mod tests {
...
+    #[test]
+    #[cfg_attr(miri, ignore = "inline assembly")]
+    fn test_read_15_bytes_valid_address() {
+        let test_buffer = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
+        let test_addr = VirtAddr::from(test_buffer.as_ptr());
+        let ptr: GuestPtr<[u8; 15]> = GuestPtr::new(test_addr);
+        let result = ptr.read().unwrap();
+
+        assert_eq!(result, test_buffer);
+    }
```
00xc, line 264: Could you move this into the new commit adding tests?

00xc (changes requested):
Just a couple of comments, other than that this LGTM.

kernel/src/mm/guestmem.rs
p4zuu: I wanted to test the case where we fetch memory straddling on both mapped and unmapped memory, but that's maybe far-fetched.

00xc: You could perhaps use a `PerCPUMappingGuard` to map a previously unmapped region to do that. Otherwise I'd say just remove the test.

00xc (approved):

kernel/src/mm/guestmem.rs
p4zuu: Done

kernel/src/mm/guestmem.rs
p4zuu: I removed it :)

PR #298, opened 2024-03-19
Documentation: update QEMU v8.2.0 parameters and fix launch_guest.sh example

stefano-garzarella:
In the QEMU's `svsm-igvm` branch that we recommend using, we moved to
QEMU v8.2.0 which wants slightly different parameters as we already
have in the `scripts/launch_guest.sh` script.
    
So let's also update INSTALL.md with these new parameters and also fix
`launch_guest.sh` example adding the missing `.sh` extension.


roy-hopkins (approved):
LGTM

PR #300, opened 2024-03-19
cpu/vc: improve error handling

00xc:
Remove calls to `Result::expect()` in the VC code and return errors so that they can be handled by callers.

`VcError` is already implemented, so use the already existing type. Introduce `VcError::new()` to improve ergonomics when creating a new error.


p4zuu (approved):
LGTM. Thanks Carlos!

PR #303, opened 2024-03-20
Move Elf code to crate

joergroedel:
Move the ELF parser code to a new crate and split up the code into modules.


Cargo.toml
```diff
@@ -6,7 +6,9 @@ members = [
     # binary targets
     "kernel",
     # fuzzing
-    "fuzz"
+    "fuzz",
+    # ELF loader
+    "elf"
```
osteffenrh, line 11: trailing commas are OK in TOML

elf/src/error.rs
```diff
@@ -0,0 +1,181 @@
+// SPDX-License-Identifier: (GPL-2.0-or-later OR MIT)
+//
+// Copyright (c) 2023-2024 SUSE LLC
+//
+// Author: Nicolai Stange <nstange@suse.de>
+//
+// vim: ts=4 sw=4 et
```
osteffenrh, line 7: Do we really need editor specific formatting options in code files?

There are things like [editorconfig](https://editorconfig.org/) that allows setting this easily for the whole project (supported by neovim out of the box)

On the other hand, don't we run the rust formatter on everything automatically anyway?

Cargo.toml
joergroedel: Okay, added a comma, thanks.

PR #309, opened 2024-03-26
cpu/vc: improve instruction decoder + handle more `IN` / `OUT` instructions + handle `RDMSR` / `WRMSR` + add SVSM discovery support

00xc:
* Simplify the instruction decoder and its API.
* Decode remaining `IN` / `OUT` instruction cases and enable related #VC tests.
* Handle `SVM_EXIT_MSR` in the #VC handler and enable related #VC tests.
* Support SVSM discovery via MSR for post-boot services.
* Support SVSM discovery via CPUID for post-boot services.

This works towards a more complete implementation of #14

### DIFF
--- a/fuzz/fuzz_targets/insn.rs
+++ b/fuzz/fuzz_targets/insn.rs
@@ -11,7 +11,7 @@ fuzz_target!(|input: &[u8]| -> Corpus {
     let mut data = [0u8; MAX_INSN_SIZE];
     data.copy_from_slice(input);
 
-    let mut insn = Instruction::new(data);
+    let insn = Instruction::new(data);
     let _ = core::hint::black_box(insn.decode());
 
     Corpus::Keep
--- a/kernel/src/cpu/insn.rs
+++ b/kernel/src/cpu/insn.rs
@@ -65,12 +65,18 @@ impl DecodedInsn {
     pub const fn size(&self) -> usize {
         match self {
             Self::Cpuid => 2,
-            Self::Inb(..) => 1,
-            Self::Inw(..) => 2,
-            Self::Inl(..) => 1,
-            Self::Outb(..) => 1,
-            Self::Outw(..) => 2,
-            Self::Outl(..) => 1,
+            Self::Inb(Operand::Reg(..)) => 1,
+            Self::Inw(Operand::Reg(..)) => 2,
+            Self::Inl(Operand::Reg(..)) => 1,
+            Self::Outb(Operand::Reg(..)) => 1,
+            Self::Outw(Operand::Reg(..)) => 2,
+            Self::Outl(Operand::Reg(..)) => 1,
+            Self::Inb(Operand::Imm(..)) => 2,
+            Self::Inw(Operand::Imm(..)) => 3,
+            Self::Inl(Operand::Imm(..)) => 2,
+            Self::Outb(Operand::Imm(..)) => 2,
+            Self::Outw(Operand::Imm(..)) => 3,
+            Self::Outl(Operand::Imm(..)) => 2,
         }
     }
 }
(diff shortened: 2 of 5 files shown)


00xc:
@AdamCDunlap have you tested `test_sev_snp_enablement_msr()` in qemu? It fails for me.

00xc:
@roy-hopkins added your review to verify that the gdb stub still works since there are some changes around `X86_TRAP_DB`.

AdamCDunlap:
> @AdamCDunlap have you tested `test_sev_snp_enablement_msr()` in qemu? It fails for me.

I have not, sorry. I would expect that one to work, though. I'm on vacation this week but I could check things out when I get back on Monday.

p4zuu:
According to the SVSM specs, the #VC handler is also supposed to set the eax[28] bit when handling a CPUID(EAX=8000_001F). I think this is not implemented neither. A quick change of the cpuid #VC in-svsm test fails for me.
While you are at it, could you please adapt the CPUID handler and the related test to check the eax[28] bit of a CPUID(EAX=8000_001F) result? :)

00xc:
@roy-hopkins could you take a brief look at the gdbstub-related changes?
@p4zuu could you approve the review if everything looks good?
@AdamCDunlap any chance you had time to take a look at the test I mentioned above?

roy-hopkins (approved):
Changes look good to me.
Tested the GDB stub from using this branch and it all still functions correctly.

p4zuu (approved):
Except the small comments I already left, this looks good to me

00xc:
Addressed Thomas' comments and rebased on main.

AdamCDunlap:
I checked out test_sev_snp_enablement_msr and it looks like it was written incorrectly. Here's a diff that fixes it and this passes for me:

```
diff --git a/kernel/src/cpu/vc.rs b/kernel/src/cpu/vc.rs
index e265f6f..0d8f3e2 100644
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -508,13 +508,12 @@ mod tests {
     }

     #[test]
-    // #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
-    #[ignore = "Currently unhandled by #VC handler"]
     fn test_sev_snp_enablement_msr() {
-        const MSR_SEV_STATUS: u32 = 0b10;
-        const MSR_SEV_STATUS_SEV_SNP_ENABLED: u64 = 0b100;
+        const MSR_SEV_STATUS: u32 = 0xc0010131;
+        const MSR_SEV_STATUS_SEV_SNP_ENABLED: u64 = 0b10;

-        let sev_status = verify_ghcb_gets_altered(|| read_msr(MSR_SEV_STATUS));
+        let sev_status = read_msr(MSR_SEV_STATUS);
         assert_ne!(sev_status & MSR_SEV_STATUS_SEV_SNP_ENABLED, 0);
     }
```

00xc:
> I checked out test_sev_snp_enablement_msr and it looks like it was written incorrectly. 

Ah yes, I did notice the `MSR_SEV_STATUS` value being wrong. However, I'm not sure why the GHCB is not being modified. In fact, the `rdmsr` is not even generating a `#VC` exception as far as I can tell, which it should. Am I missing something?

AdamCDunlap:
>  However, I'm not sure why the GHCB is not being modified.

I spent a while trying to debug this but I didn't get too far. I spent a while trying to figure out why adding print statements made the test pass before realizing that the print statements generate \#VCs themselves which do modify the ghcb (I was searching for UB due to the unsafe). Our internal version of this test does not check that the ghcb was modified, either, which is why I was OK leaving it out.

I wasn't completely sure that the ghcb used in verify_ghcb_gets_altered is the same as the ghcb used in the vc handler. I was trying to print and compare their addresses and I thought they were different, but I got sidetracked and thought it wasn't too important.

> In fact, the rdmsr is not even generating a #VC exception as far as I can tell, which it should

I added print statements in and it looks like the correct \#VC was generated.

AdamCDunlap:
> > In fact, the rdmsr is not even generating a #VC exception as far as I can tell, which it should
> 
> I added print statements in and it looks like the correct #VC was generated.

Actually I took a closer look and you're right, the \#VC isn't generated for readmsr. I'm not sure why that is. Maybe only some msrs are trapped by \#VC? Would need to take a closer look at the spec.

00xc:
> > > In fact, the rdmsr is not even generating a #VC exception as far as I can tell, which it should
> > 
> > 
> > I added print statements in and it looks like the correct #VC was generated.
> 
> Actually I took a closer look and you're right, the #VC isn't generated for readmsr. I'm not sure why that is. Maybe only some msrs are trapped by #VC? Would need to take a closer look at the spec.

I took a brief look and could not find anything, other than a slight hint in this sentence (AMD APM Rev. 3.40, section 15.34.10):

> The SEV_STATUS MSR is read-only and accesses to the SEV_STATUS MSR cannot be intercepted by the hypervisor.

So perhaps, and this is a wild guess, reading this MSR never causes a #VC since there is no point? After all the whole idea of #VC is to set up the necessary state to enable some hypervisor involvement. Maybe @tlendacky can provide a better explanation.

00xc:
> I checked out test_sev_snp_enablement_msr and it looks like it was written incorrectly. Here's a diff that fixes it and this passes for me:
> 
> ```
> diff --git a/kernel/src/cpu/vc.rs b/kernel/src/cpu/vc.rs
> index e265f6f..0d8f3e2 100644
> --- a/kernel/src/cpu/vc.rs
> +++ b/kernel/src/cpu/vc.rs
> @@ -508,13 +508,12 @@ mod tests {
>      }
> 
>      #[test]
> -    // #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
> -    #[ignore = "Currently unhandled by #VC handler"]
> +    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
>      fn test_sev_snp_enablement_msr() {
> -        const MSR_SEV_STATUS: u32 = 0b10;
> -        const MSR_SEV_STATUS_SEV_SNP_ENABLED: u64 = 0b100;
> +        const MSR_SEV_STATUS: u32 = 0xc0010131;
> +        const MSR_SEV_STATUS_SEV_SNP_ENABLED: u64 = 0b10;
> 
> -        let sev_status = verify_ghcb_gets_altered(|| read_msr(MSR_SEV_STATUS));
> +        let sev_status = read_msr(MSR_SEV_STATUS);
>          assert_ne!(sev_status & MSR_SEV_STATUS_SEV_SNP_ENABLED, 0);
>      }
> ```

In the meantime I pushed this change. I added you in the `Suggested-by` tag of the commit.

00xc:
Forgot to mention this before - the changes in this PR should prevent the WeSee attack (via "cpu/insn: improve instruction decoding"):
https://ahoi-attacks.github.io/wesee/

tlendacky:
> So perhaps, and this is a wild guess, reading this MSR never causes a #VC since there is no point? After all the whole idea of #VC is to set up the necessary state to enable some hypervisor involvement. Maybe @tlendacky can provide a better explanation.

The SEV_STATUS MSR is a non-interceptable MSR. Since it is non-interceptable it does not generate a #VC, it just executes.

PR #312, opened 2024-04-01
kernel/cpu: overhaul per-CPU borrow checking

msft-jlange:
This PR changes the implementation of CPU borrow checking in two important ways.  First, it uses `RefCell` instead of a custom implementation of borrow checking - this is valuable because it relies on core common code.  Second, borrow checking is now performed in all builds, not just debug builds - this is important because the memory safety guarantees made by Rust are intended to apply in all cases, in all code paths, not just those code paths that were validated using debug builds.

The contents of `PerCpu` have been refactored to enforce meaningful borrow checking according to a specific set of principles.
- In general, *all* references to a structure must enforce borrow checking rules to prevent conflicts between shared and mutable references.
- When references are synthesized from global constants (and thus the compiler cannot correctly check borrowed references), borrow checking must be performed dynamically.
- Borrow checking can safely be avoided if, and only if, references to a structure can only be issued as shared references, because no borrow conflict can arise when no mutable references are possible.
- If an object cannot adhere to these borrow checking rules, because of some runtime condition that could require access to object fields in violation of the borrow rules, references to the object must be made using raw pointers.  This requires the use of unsafe code by any logic that has to reach in, and should thus prompt careful justification that any non-compliant reference is actually correct and appropriate.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -233,7 +233,7 @@ impl PerCpuShared {
 
 #[derive(Debug)]
 pub struct PerCpu {
-    pub shared: &'static PerCpuShared,
+    shared: *const PerCpuShared,
     online: AtomicBool,
     apic_id: u32,
     pgtbl: SpinLock<PageTableRef>,
@@ -268,7 +268,7 @@ pub struct PerCpu {
 }
 
 impl PerCpu {
-    fn new(apic_id: u32, shared: &'static PerCpuShared) -> Self {
+    fn new(apic_id: u32, shared: *const PerCpuShared) -> Self {
         PerCpu {
             shared,
             online: AtomicBool::new(false),
@@ -311,11 +311,15 @@ impl PerCpu {
 
             let percpu = vaddr.as_mut_ptr::<PerCpu>();
 
-            (*percpu) = PerCpu::new(apic_id, &*percpu_shared);
+            (*percpu) = PerCpu::new(apic_id, percpu_shared);
 
             PERCPU_AREAS.push(PerCpuInfo::new(apic_id, shared_vaddr));
             Ok(percpu)
         }
+    }
+
+    pub fn shared(&self) -> &'static PerCpuShared {
+        unsafe { &*self.shared }
     }
 
     pub fn set_online(&mut self) {
@@ -542,7 +546,7 @@ impl PerCpu {
     }
 
     pub fn guest_vmsa_ref(&self) -> LockGuard<'_, GuestVmsaRef> {
-        self.shared.guest_vmsa.lock()
+        self.shared().guest_vmsa.lock()
     }
 
     pub fn alloc_guest_vmsa(&mut self) -> Result<(), SvsmError> {
@@ -552,7 +556,7 @@ impl PerCpu {
         let vmsa = vmsa_mut_ref_from_vaddr(vaddr);
         init_guest_vmsa(vmsa, self.reset_ip);
 
-        self.shared.update_guest_vmsa(paddr);
+        self.shared().update_guest_vmsa(paddr);
 
         Ok(())
     }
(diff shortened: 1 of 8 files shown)


joergroedel:
There is Linux kernel boot failure with these changes, I am currently invesigating:

```
[    0.997422][    T1] smpboot: Max logical packages: 1
[    0.998545][    T1] smpboot: Total of 24 processors activated (127199.90 BogoMIPS)
[SVSM] ERROR: Panic: CPU[19] panicked at kernel/src/cpu/percpu.rs:181:9:
assertion failed: self.vmsa.is_some()
[SVSM] ---BACKTRACE---:
[SVSM]   [0xffffff8000132043]
[SVSM]   [0xffffff800016fc3f]
[SVSM]   Invalid frame
[SVSM] ---END---
```

joergroedel:
It looks like there is a pointer offset of 32 bytes between when the VMSAs are set and when they are fetched. The printed addresses are of GuestVmsaRef structures:

```
[SVSM] Setting VMSA and CAA for ptr 0xffffff8000b97000
[SVSM] Set new VMSA and CAA for CPU 23
...
[SVSM] Getting VMSA ptr 0xffffff8000b97020
[SVSM] ERROR: Panic: CPU[23] panicked at kernel/src/cpu/percpu.rs:185:9:
assertion failed: self.vmsa.is_some()
[SVSM] ---BACKTRACE---:
[SVSM]   [0xffffff800013285a]
[SVSM]   [0xffffff800017066f]
[SVSM]   Invalid frame
[SVSM] ---END---
```

kernel/src/cpu/percpu.rs
```diff
@@ -255,75 +324,51 @@ pub struct PerCpu {
...
 
-            let shared_vaddr = vaddr + private_size;
-            let percpu_shared = shared_vaddr.as_mut_ptr::<PerCpuShared>();
-            (*percpu_shared) = PerCpuShared::new();
-
-            let percpu = vaddr.as_mut_ptr::<PerCpu>();
-
-            (*percpu) = PerCpu::new(apic_id, &*percpu_shared);
+            let percpu_unsafe = vaddr.as_mut_ptr::<PerCpuUnsafe>();
+            (*percpu_unsafe) = PerCpuUnsafe::new(apic_id, percpu_unsafe);
```
joergroedel, line 358: This creates a confusing imbalance. `alloc()` is an associated method of `struct PerCpu`, but does allocate a `struct PerCpuUnsafe` and returns an inner member of it.
So while at it, the allocation of `PerCpuUnsafe` data structures should be moved into a generic function.

joergroedel, line 358: This creates a confusing imbalance. `alloc()` is an associated method of `struct PerCpu`, but does allocate a struct `PerCpuUnsafe` and returns an inner member of it.
So while at it, the allocation of `PerCpuUnsafe` data structures should be moved into a generic function.

joergroedel: This change also needs a corresponding change in `PerCpuAreas::get()`, like mentioned in my other comment on the PR.

joergroedel (changes requested):

msft-jlange:
> Found the problem: The percpu area now holds a PerCpuUnsafe instance, but the `PerCpuAreas::get()` method casts the pointer into a `PerCpuShared` instance. Rewriting `get()` like below fixes the issue:
> [...]
> In general the storing of a `VirtAddr` in `PerCpuInfo` was a bad idea to begin with, it made this mistake slip through undetected. The percpu pointer should be stored in a type-safe way going forward.

Thanks for taking the time to debug this!  I've changed the `PerCpuInfo` to store a reference instead of a pointer or address, so we should now have the type safety we want.

kernel/src/cpu/percpu.rs
msft-jlange: I've restructured the `alloc` methods of `PerCpu` and `PerCpuUnsafe` which should eliminate the confusion.  Callers will still call `PerCpu::alloc()` but there is a more rational chain of allocation and construction to build up the various components.

joergroedel (approved):
Looks good now. Testing also went well.

PR #314, opened 2024-04-03
Support execution in user-mode

joergroedel:
Here are changes to allow the SVSM to load ELF binaries from the RAM file-system and execute them in user-mode.

The changes itself are not very useful yet, as there are no syscalls implemented besides some needed for testing, but the PR contains all the heavy-lifting needed to:

- VMM changes to populate and manage a user address space
- Opening and parsing of ELF files
- Creating user tasks and set up the correct stack layout to get it into user-mode once the task is scheduled
- Segment setup and IDT/entry code changes needed for user-mode

The changes in this PR are the groundwork for running SVSM modules in user-space.

### DIFF
--- a/kernel/src/address.rs
+++ b/kernel/src/address.rs
@@ -212,6 +212,11 @@ impl VirtAddr {
         Self(sign_extend(addr))
     }
 
+    /// Returns the index into page-table pages of given levels.
+    pub const fn to_pgtbl_idx<const L: usize>(&self) -> usize {
+        (self.0 >> (12 + L * 9)) & 0x1ffusize
+    }
+
     #[inline]
     pub fn as_ptr<T>(&self) -> *const T {
         self.0 as *const T
--- a/kernel/src/address.rs
+++ b/kernel/src/address.rs
@@ -8,6 +8,8 @@ use crate::types::{PAGE_SHIFT, PAGE_SIZE};
 use core::fmt;
 use core::ops;
 
+use core::slice;
+
 // The backing type to represent an address;
 type InnerAddr = usize;
 
@@ -250,6 +252,24 @@ impl VirtAddr {
 
     pub const fn const_add(&self, offset: usize) -> Self {
         VirtAddr::new(self.0 + offset)
+    }
+
+    /// Converts the `VirtAddr` to a slice of a given type
+    ///
+    /// # Arguments:
+    ///
+    /// * `len` - Number of elements of type `T` in the slice
+    ///
+    /// # Returns
+    ///
+    /// Slice with `len` elements of type `T`
+    ///
+    /// # Safety
+    ///
+    /// All Safety requirements from [`core::slice::from_raw_parts`] for the
+    /// data pointed to by the `VirtAddr` apply here as well.
+    pub unsafe fn to_slice<T>(&self, len: usize) -> &[T] {
+        slice::from_raw_parts::<T>(self.as_ptr::<T>(), len)
     }
 }
 
--- a/kernel/src/cpu/idt/entry.S
+++ b/kernel/src/cpu/idt/entry.S
@@ -145,3 +145,6 @@ default_entry_no_ist	name=vc		handler=vmm_communication	error_code=1	vector=29
 
 // #SX Security Exception (Vector 30)
 default_entry_no_ist	name=sx		handler=panic			error_code=1	vector=30
+
+// INT 0x80 system call handler
+default_entry_no_ist	name=int80	handler=system_call		error_code=0	vector=0x80
--- a/kernel/src/cpu/idt/entry.S
+++ b/kernel/src/cpu/idt/entry.S
@@ -58,9 +58,18 @@ asm_entry_\name:
 	jmp	default_return
 .endm
 
+.globl default_return
 default_return:
+	testb	$3, 18*8(%rsp) // Check CS in exception frame
+	jnz return_user
 	pop_regs
+default_iret:
 	iretq
+
+return_user:
+	// Put user-mode specific return code here
+	pop_regs
+	jmp default_iret
 
 // #DE Divide-by-Zero-Error Exception (Vector 0)
 default_entry_no_ist	name=de		handler=panic			error_code=0	vector=0
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -11,13 +11,14 @@ use super::super::tss::IST_DF;
 use super::super::vc::handle_vc_exception;
 use super::common::PF_ERROR_WRITE;
 use super::common::{
-    idt_mut, IdtEntry, AC_VECTOR, BP_VECTOR, BR_VECTOR, CP_VECTOR, DB_VECTOR, DE_VECTOR, DF_VECTOR,
-    GP_VECTOR, HV_VECTOR, MCE_VECTOR, MF_VECTOR, NMI_VECTOR, NM_VECTOR, NP_VECTOR, OF_VECTOR,
-    PF_VECTOR, SS_VECTOR, SX_VECTOR, TS_VECTOR, UD_VECTOR, VC_VECTOR, XF_VECTOR,
+    idt_mut, user_mode, IdtEntry, AC_VECTOR, BP_VECTOR, BR_VECTOR, CP_VECTOR, DB_VECTOR, DE_VECTOR,
+    DF_VECTOR, GP_VECTOR, HV_VECTOR, MCE_VECTOR, MF_VECTOR, NMI_VECTOR, NM_VECTOR, NP_VECTOR,
+    OF_VECTOR, PF_VECTOR, SS_VECTOR, SX_VECTOR, TS_VECTOR, UD_VECTOR, VC_VECTOR, XF_VECTOR,
 };
 use crate::address::VirtAddr;
 use crate::cpu::X86ExceptionContext;
 use crate::debug::gdbstub::svsm_gdbstub::handle_debug_exception;
+use crate::task::terminate;
 use core::arch::global_asm;
 
 use crate::syscall::*;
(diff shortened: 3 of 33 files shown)


kernel/src/address.rs
```diff
@@ -8,6 +8,9 @@ use crate::types::{PAGE_SHIFT, PAGE_SIZE};
 use core::fmt;
 use core::ops;
 
+extern crate alloc;
+use alloc::slice;
```
00xc, line 12: Should probably use `core::slice` instead.

kernel/src/address.rs
```diff
@@ -251,6 +254,25 @@ impl VirtAddr {
...
+    ///
+    /// # Returns
+    ///
+    /// Slice with `len` elements of type `T`
+    ///
+    /// # Safety
+    ///
+    /// The caller must assure that the virtual mapping at `VirtAddr` covers
+    /// the whole slice and that the data at the virtual address contains valid
+    /// instances of type `T`.
```
00xc, line 276: And that alignment is met. See the requirements in `core::slice::from_raw_parts`

kernel/src/address.rs
```diff
@@ -213,6 +213,10 @@ impl VirtAddr {
         Self(sign_extend(addr))
     }
 
+    pub const fn to_idx(&self) -> usize {
```
00xc, line 216: The method name does not explain much, I would make it more verbose or at least add some documentation. Could we also make it generic over the page table level, like `PageTable::index()`?

kernel/src/cpu/idt/svsm.rs
```diff
@@ -166,7 +169,18 @@ extern "C" fn ex_handler_vmm_communication(ctx: &mut X86ExceptionContext) {
...
+    if input.is_err() {
+        ctx.regs.rax = !0;
+        return;
+    }
+
+    ctx.regs.rax = match input.unwrap() {
+        SYS_HELLO => sys_hello(),
+        SYS_EXIT => sys_exit(),
+        _ => !0,
+    };
```
00xc:
```suggestion
    let Ok(input) = u64::try_from(ctx.regs.rax) else {
        ctx.regs.rax = !0;
        return;
    };
    ctx.regs.rax = match input {
        SYS_HELLO => sys_hello(),
        SYS_EXIT => sys_exit(),
        _ => !0,
    };
```

kernel/src/mm/alloc.rs
```diff
@@ -917,6 +917,19 @@ impl PageRef {
     pub fn phys_addr(&self) -> PhysAddr {
         self.phys_addr
     }
+
+    pub fn try_copy_page(&self) -> Result<Self, SvsmError> {
+        let virt_addr = allocate_file_page()?;
+        unsafe {
+            let src = self.virt_addr.as_ptr::<[u8; PAGE_SIZE]>();
+            let dst = virt_addr.as_mut_ptr::<[u8; PAGE_SIZE]>();
+            dst.write(*src);
```
00xc, line 926: `copy_to_nonoverlapping()` seems more appropriate and cannot create an intermediate copy in debug builds.

kernel/src/cpu/idt/svsm.rs
```diff
@@ -153,18 +197,51 @@ extern "C" fn ex_handler_hypervisor_injection(_ctx: &mut X86ExceptionContext) {
 
 // VMM Communication handler
 #[no_mangle]
-extern "C" fn ex_handler_vmm_communication(ctx: &mut X86ExceptionContext) {
-    handle_vc_exception(ctx).expect("Failed to handle #VC");
+extern "C" fn ex_handler_vmm_communication(ctxt: &mut X86ExceptionContext) {
+    let rip = ctxt.frame.rip;
+    let err = ctxt.error_code;
+
+    if handle_vc_exception(ctxt).is_err() {
```
00xc, line 204: We could also retrieve the specific error from `handle_vc_exception()` and display it.

kernel/src/mm/pagetable.rs
```diff
@@ -969,11 +970,30 @@ impl PageTablePart {
...
+    pub fn alloc(&mut self) {
+        self.get_or_init_mut();
+    }
+
+    fn get_or_init_mut(&mut self) -> &mut RawPageTablePart {
+        if self.raw.is_none() {
+            self.raw = Some(Box::<RawPageTablePart>::default());
+        }
+        self.raw.as_mut().unwrap()
+    }
```
00xc, line 987:
```suggestion
    fn get_or_init_mut(&mut self) -> &mut RawPageTablePart {
        self.raw.get_or_insert_with(|| Box::default())
    }
```

kernel/src/cpu/idt/entry.S
```diff
@@ -136,3 +145,6 @@ default_entry_no_ist	name=vc		handler=vmm_communication	error_code=1	vector=29
 
 // #SX Security Exception (Vector 30)
 default_entry_no_ist	name=sx		handler=panic			error_code=1	vector=30
+
+// INT 0x80 system call handler
+default_entry_no_ist	name=int80	handler=system_call		error_code=0	vector=0x80
```
cxdong, line 150: The TDX and SEV have disabled int80 in linux kernel with commits around this ( https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/arch/x86/coco/tdx?h=v6.7&id=b82a8dbd3d2f4563156f7150c6f2ecab6e960b30), which looks like to prevent with the Ahoi (https://ahoi-attacks.github.io/) attacks. It appears that enabling SEV's interrupt restriction can prevent host to inject 0x80 to SVSM to mitigate the Ahoi attack. But still would like to know if there is any consideration of not using SYSCALL/SYSRET instruction?

00xc: See @msft-jlange's comment regarding the downsides of using `SYSCALL` here:
https://github.com/coconut-svsm/svsm/pull/271#issuecomment-1944289580

kernel/src/mm/alloc.rs
p4zuu: What about a `GuestPtr::<u8, PAGE_SIZE>` here?
It's again not perfectly appropriate since it's supposed to be "Guest" related, but that would remove the unsafe and `write()` also returns a `Result`. However I don't know the performance difference with `copy_to_nonoverlapping()`

kernel/src/cpu/idt/entry.S
msft-jlange: SYSCALL forces the use of IST which introduces huge amounts of complexity around event delivery.  If SYSCALL is the only reason for that complexity, then it isn't really worth it, since the performance difference between interrupt dispatch and SYSCALL dispatch is minimal, particularly given the number of system calls that are normally expected during SVSM execution.  The use of restricted injection will completely mitigate the described attack, and COCONUT-SVSM is already planning to use restricted injection (it's not used yet because QEMU doesn't yet support its use fully).  Linux doesn't really have the option of using restricted injection because it's a significant change to an existing code base, but COCONUT-SVSM is being designed from the beginning to anticipate the use of restricted injection (plus it has a significantly different set of requirements for the use of interrupts) so we will not have the same challenges.

kernel/src/address.rs
joergroedel: Good point. I renamed the method to `to_pgtbl_idx()` and made it a generic like `PageTable::index()`. In fact, I also re-implemented the latter to use the former.

kernel/src/mm/alloc.rs
joergroedel: Concurrent modifications can be a problem, but most often are not. The data in file-pages is semantically the same as data behind a file descriptor in a normal Linux Rust application.

kernel/src/address.rs
joergroedel: Changed to `core::slice`.

kernel/src/mm/alloc.rs
00xc: Yes, we already agreed in the past that preventing race conditions there should be done by an external mechanism. However my point about `copy_to_nonoverlapping()` still stands.

joergroedel: Switched to `ptr::copy_nonoverlapping()`.

kernel/src/cpu/idt/svsm.rs
joergroedel: Done, thanks.

kernel/src/mm/pagetable.rs
joergroedel: That is a lot better, changed the method to your suggestion.

kernel/src/cpu/idt/svsm.rs
```diff
@@ -153,18 +197,51 @@ extern "C" fn ex_handler_hypervisor_injection(_ctx: &mut X86ExceptionContext) {
...
+    if input.is_err() {
+        ctxt.regs.rax = !0;
+        return;
+    }
+
+    ctxt.regs.rax = match input.unwrap() {
+        SYS_HELLO => sys_hello(),
+        SYS_EXIT => sys_exit(),
+        _ => !0,
+    };
```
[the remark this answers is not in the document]
joergroedel: That is better, updated the function.

kernel/src/cpu/idt/entry.S
joergroedel: Besides what @msft-jlange wrote, the reason for currently using int80 is to save some work for making certain exception handlers use IST. We can leave it this way if no other compelling reasons to use SYSCALLs are brought up. E.g., we can discuss whether the ability to inject arbitrary vectors on TDX would be a reason for using SYSCALL.

cxdong: Yes, I initially worry about TDX as currently there is no interrupt filtering feature for L1. As long as svsm can distinguish if the int80 comes from the user space or the host VMM, svsm itself can filter out the arbitrary injection by the host VMM. And what @msft-jlange and you mentioned also makes sense to me. Thanks a lot!

msft-jlange: With TDX, it is possible for an interrupt handler to determine whether it was invoked as a software interrupt or a virtual hardware interrupt, so the int80 handler will know whether it's being attacked by the host.

cxdong: Yes. For TDX, SVSM can use the same way as what the Linux kernel did, which checks the LAPIC ISR to see if vector 0x80 is set or not, to determine if it is a software interrupt or a virtual hardware interrupt injected by host VMM. This is can be done after introducing the LAPIC driver in SVSM.

joergroedel: Okay, this discussion seems to be resolved. If there are any other points to add, please open an issue and continue discussion there.

Issue #313, opened 2024-04-03
`make RELEASE=1` fails after vTPM merge


joergroedel:
I run into build errors with release builds using the latest main branch. The error message is:

```
error: linking with `ld` failed: exit status: 1
  |
[...]

PR #316, opened 2024-04-05
Makefile: Exclude default features when building stage2

roy-hopkins:
The svsm module in kernel/ builds both the stage2 and kernel binaries. This can be configured with enable-gdb and mstpm features, both of which only apply to the kernel binary and not stage2. However the current Makefile enables the same features set for both binary builds.

The linker would normally remove the unused code generated for the features but, for mstpm the linker leaves a large amount of dead code in stage2. This causes the link to fail as stage2 does not link against libmstpm.

This commit changes the Makefile to remove all default features when building stage 2. In order to do this, it must run the cargo build directly on the kernel Cargo.toml.

This addresses https://github.com/coconut-svsm/svsm/issues/313

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ bin/meta.bin: utils/gen_meta utils/print-meta
 	./utils/gen_meta $@
 
 bin/stage2.bin: bin
-	cargo build ${CARGO_ARGS} ${SVSM_ARGS} --bin stage2
+	cargo build --manifest-path kernel/Cargo.toml ${CARGO_ARGS} --no-default-features --bin stage2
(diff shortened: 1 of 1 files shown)


cclaudio:
This would fix the build but the guest boot still crashes with an exception 6 (invalid opcode) when compiled with RELEASE=1. I traced that down to the openssl being optimized to call a `pxor` instruction. I have a PR that also fix the build and "mitigates" the exception 6. Let me submit it as well for discussion

cclaudio:
Actually, your patch looks good to me. I will submit a PR with only the SVSM boot crash. Thanks!

cclaudio (approved):

PR #320, opened 2024-04-12
Fix race conditions in SVSM core protocol handling

joergroedel:
There are a couple of race conditions in how the SVSM core protocol handling is implemented. These issues can be used by an attacker to leak data out of the SVSM or escalate privileges of lower VMPLs.


kernel/src/protocols/core.rs
```diff
@@ -44,12 +44,23 @@ struct PValidateRequest {
...
 
-fn core_create_vcpu_error_restore(vaddr: VirtAddr) {
-    if let Err(err) = rmp_clear_guest_vmsa(vaddr) {
-        log::error!("Failed to restore page permissions: {:#?}", err);
+fn core_create_vcpu_error_restore(paddr: Option<PhysAddr>, vaddr: Option<VirtAddr>) {
+    for v in vaddr.iter() {
+        if let Err(err) = rmp_clear_guest_vmsa(*v) {
+            log::error!("Failed to restore page permissions: {:#?}", err);
+        }
     }
```
00xc, line 62: Why not `if let Some(v) = vaddr { ... }`? Seems more clear. Same thing below with `paddr`.

kernel/src/protocols/core.rs
```diff
@@ -44,12 +44,23 @@ struct PValidateRequest {
...
+
     // In case mappings have been changed
     flush_tlb_global_sync();
+
+    for p in paddr.iter() {
+        // SAFETY: this can only fail if another CPU unregisters our
+        // unused VMSA. This is not possible, since unregistration of
+        // an unused VMSA only happens in the error path for this function,
+        // with a physical address that only this CPU managed to register.
```
00xc, line 71: This comment should be updated. This is only called in the error path for `core_create_vcpu()`.

kernel/src/protocols/core.rs
```diff
@@ -86,6 +86,11 @@ fn core_create_vcpu(params: &RequestParams) -> Result<(), SvsmReqError> {
         return Err(SvsmReqError::invalid_address());
     }
 
+    // Check whether VMSA page and CAA region overlap
+    if overlap(paddr, paddr + PAGE_SIZE - 1, pcaa, pcaa + 7) {
```
00xc, line 100: Maybe this is clearer:
```suggestion
    if overlap(paddr, paddr + PAGE_SIZE - 1, pcaa, pcaa + size_of::<SvsmCaa> - 1) {
```

roy-hopkins, line 100: According to the SVSM specification revision 1.00 the CAA is "guaranteed to be aligned to a 4 KB boundary" and "the remainder of the page may be used by the guest for memory-based parameter passing". 

The check above checks for an 8 byte alignment and this code only checks if the 8 byte CAA structure overlaps.

See https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/58019.pdf,  section 5.

kernel/src/protocols/core.rs
```diff
@@ -293,6 +293,9 @@ fn core_pvalidate_one(entry: u64, flush: &mut bool) -> Result<(), SvsmReqError>
         // down the #NPF loops.
         //
         if writable_phys_addr(paddr) {
+            // FIXME: This checks leaves a window open for the attack described
```
roy-hopkins, line 296: "This check leaves" (remove the s).

kernel/src/protocols/core.rs
joergroedel: Updated for vaddr and paddr.

kernel/src/protocols/core.rs
joergroedel: Updated.

kernel/src/protocols/core.rs
joergroedel: Makes sense, changed that.

kernel/src/protocols/core.rs
joergroedel: Removed the `s`

PR #323, opened 2024-04-17
Automatically publish documentation to github pages

roy-hopkins:
## Background

In order to make it easier to write and consume documentation for COCONUT-SVSM it is desirable to have a process that automatically takes documentation added to the main branch via PRs to be published to a hosted site where the documentation can be viewed. Github pages provide a good hosting environment where publishing pages can be triggered on changes to the main branch using github actions.

## What this PR introduces

This PR adds configuration and makefile targets for building a documentation website using [mkdocs](https://www.mkdocs.org/). 

The new Makefile `docsite` target creates a site folder that contains rendered html then copies in the rustdoc output for SVSM (not including dependencies). This can then be published to a web server.

A new github action is also introduced that is run each time the main branch is pushed. This builds the `docsite` target then force pushes the output to a new branch named `gh-pages`. Github can be configured to automatically publish this branch to the project github pages (which will be at https://coconut-svsm.github.io/svsm by default when published).

## How to publish docs using these changes

Once this is merged and Github pages has been configured to publish the `gh-pages` branch, contributors can update documentation by simply adding markdown files to `Documentation/docs` and updating the navigation in `Documentation/mkdocs.yml`. Once the PR containing their changes is merged, the updated documentation will be published.

The documentation can also be previewed locally with:

```
$ make docsite-serve
```

## Configuring github pages

To configure the SVSM repository to publish the pages:

1. Go to the settings for the repository.
2. Select "Pages" from the menu on the left.
3. Under "Build and Deployment", select "Deploy from a branch" for the source.
4. Select `gh-pages` as the branch and deploy from the root.

## Example

You can see what the result of merging this PR looks like with the github pages from my fork: https://roy-hopkins.github.io/svsm/


Makefile
```diff
@@ -84,6 +87,15 @@ test-in-svsm: utils/cbit bin/coconut-test-qemu.igvm
 doc:
 	cargo doc -p svsm --open --all-features --document-private-items
 
+docsite:
+	mkdir -p ${DOC_SITE}
+	cargo doc -p svsm --all-features --document-private-items --no-deps
```
00xc, line 92: For the public docs I think we could omit `--document-private-items`? Or maybe leave it as an option.

roy-hopkins: I did wonder that, but the documentation is aimed more towards people developing internals for the SVSM and not consuming public APIs so thought it best to leave the private items there in this case. I'm happy to exclude them though if people think that's best.

joergroedel: Documenting the private items makes sense, I agree that the documentation is aimed for COCONUT developers and not API users.

PR #324, opened 2024-04-17
test: remove `memoffset` dependency

00xc:
As of Rust 1.77, the `core::mem::offset_of!()` macro is stable, which means we no longer need to use the `memoffset` crate in tests. The new  macro is also more efficient as it does not have excessive stack usage, which means we can enable the tests that make use of this macro inside the SVSM.

While we are at it, remove an unnecessary check in `scripts/test-in-svsm.sh`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -833,7 +833,6 @@ dependencies = [
  "intrusive-collections",
  "libmstpm",
  "log",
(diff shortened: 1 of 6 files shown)


roy-hopkins (approved):
LGTM

PR #326, opened 2024-04-19
mm/alloc: remove allocator wrappers and fix a soundness issue

00xc:
* Remove `mem_allocate()`, `mem_reallocate()` and `mem_deallocate()`, as they simply duplicate the functions in the alloc library (`alloc::alloc::{alloc, realloc, dealloc}`).
* Hold the root memory lock for the whole duration of `layout_from_ptr()`. Otherwise the pointer in the slab page info could be invalidated concurrently.

### DIFF
--- a/kernel/src/mm/alloc.rs
+++ b/kernel/src/mm/alloc.rs
@@ -1615,28 +1615,6 @@ pub fn root_mem_init(pstart: PhysAddr, vstart: VirtAddr, page_count: usize) {
 /// [`TestRootMem::setup()`].
 static TEST_ROOT_MEM_LOCK: SpinLock<()> = SpinLock::new(());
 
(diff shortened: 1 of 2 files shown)


kernel/src/mm/alloc.rs
```diff
@@ -1653,10 +1631,9 @@ pub fn layout_from_size(size: usize) -> Layout {
 pub fn layout_from_ptr(ptr: *mut u8) -> Option<Layout> {
     let va = VirtAddr::from(ptr);
 
-    let Ok(pfn) = ROOT_MEM.lock().get_pfn(va) else {
-        return None;
-    };
-    let info = ROOT_MEM.lock().read_page_info(pfn);
```
p4zuu, line 1659: Double-fetch bug indeed, nice catch

p4zuu (approved):

PR #327, opened 2024-04-23
cpu/vc: implement `RDTSC` and `RDSTCP` #VC handlers

00xc:
* Add GHCB getters and setters that check or set the valid bitmap (now that there's a stable `offset_of!()` macro).
* Fix potential problem with `GHCB::rdmsr_regs()`.
* Implement `RDTSC` #VC handler and enable relevant in-SVSM test.
* Fix `rdtscp()`.
* Implement `RDTSCP` #VC handler and enable relevant in-SVSM test.
* Add a test for the GHCB layout.

This works towards a more complete implementation of #14 .

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -21,6 +21,7 @@ use core::fmt;
 
 pub const SVM_EXIT_EXCP_BASE: usize = 0x40;
 pub const SVM_EXIT_LAST_EXCP: usize = 0x5f;
+pub const SVM_EXIT_RDTSC: usize = 0x6e;
 pub const SVM_EXIT_CPUID: usize = 0x72;
 pub const SVM_EXIT_IOIO: usize = 0x7b;
 pub const SVM_EXIT_MSR: usize = 0x7c;
@@ -98,8 +99,8 @@ pub fn stage2_handle_vc_exception(ctx: &mut X86ExceptionContext) -> Result<(), S
     let err = ctx.error_code;
 
     // To handle NAE events, we're supposed to reset the VALID_BITMAP field of the GHCB.
-    // This is currently only relevant for IOIO handling. This field is currently reset in
-    // the ioio_{in,ou} methods but it would be better to move the reset out of the different
+    // This is currently only relevant for IOIO and RDTSC handling. This field is currently reset in
+    // the relevant GHCB methods but it would be better to move the reset out of the different
     // handlers.
     let mut ghcb = current_ghcb();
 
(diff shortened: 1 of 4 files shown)


kernel/src/sev/ghcb.rs
```diff
@@ -70,6 +52,27 @@ const PSC_FLAG_HUGE: u64 = 1 << PSC_FLAG_HUGE_SHIFT;
 
 const GHCB_BUFFER_SIZE: usize = 0x7f0;
 
+macro_rules! ghcb_getter {
+    ($name:ident, $field:ident,$t:ty) => {
+        #[allow(unused)]
+        fn $name(&self) -> Result<$t, GhcbError> {
+            self.is_valid(offset_of!(Self, $field))
```
p4zuu, line 59: Do we always need to check if the related bit in `VALID_BITMAP` is set?
Without asking right now for other getters and setters that don't check the valid bit, I like the Linux kernel naming `ghcb_get_##field##` and `ghcb_get_##field##_if_valid`.
When reading a call to one of the new `ghcb_getter` here, I would find it slightly clearer to have `if_valid` suffix at the end of the method

kernel/src/sev/ghcb.rs
```diff
@@ -250,91 +298,26 @@ impl GHCB {
...
-        if !self.is_valid(OFF_SW_EXIT_INFO_1) {
-            return Err(GhcbError::VmgexitInvalid);
-        }
-
-        if self.sw_exit_info_1 != 0 {
-            return Err(GhcbError::VmgexitError(
-                self.sw_exit_info_1,
-                self.sw_exit_info_2,
-            ));
+        let sw_exit_info_1 = self.get_exit_info_1()?;
```
p4zuu, line 339: Just an example of my comment about `ghcb_getter`s, I would find this clearer to have `self.get_exit_info_1_if_valid()?`, but it's maybe nitpicking :)

kernel/src/sev/ghcb.rs
00xc: > Do we always need to check if the related bit in VALID_BITMAP is set?

In that case we can just access the field directly - the methods are not `pub` anyway so this is only intended for internal use.

> Without asking right now for other getters and setters that don't check the valid bit, I like the Linux kernel naming ghcb_get_##field## and ghcb_get_##field##_if_valid.
When reading a call to one of the new ghcb_getter here, I would find it slightly clearer to have if_valid suffix at the end of the method

I initially decided not to do this because because `set_rax()`, `set_rbx()`, etc. already existed (it's just now they're implemented though a macro), so renaming them to e.g. `set_rax_valid()` would make this PR more intrusive by changing all their uses. But sure, for clarity's sake I'll add `_valid` to the new methods.

kernel/src/cpu/vc.rs
```diff
@@ -97,9 +99,10 @@ pub fn stage2_handle_vc_exception_no_ghcb(ctx: &mut X86ExceptionContext) -> Resu
 pub fn stage2_handle_vc_exception(ctx: &mut X86ExceptionContext) -> Result<(), SvsmError> {
     let err = ctx.error_code;
 
-    // To handle NAE events, we're supposed to reset the VALID_BITMAP field of the GHCB.
-    // This is currently only relevant for IOIO handling. This field is currently reset in
-    // the ioio_{in,ou} methods but it would be better to move the reset out of the different
+    // To handle NAE events, we're supposed to reset the VALID_BITMAP field of
```
p4zuu, line 102: Would it be a better to call `ghcb.clear()` here now? I mean in the `*handle_vc_exception()` functions. That would factorize the `self.clear()` calls in the GHCB handlers (eg. in the new `rdtsc_regs()` and `rdtsc_regs()`). I don't really know which solution is better

p4zuu: This makes sense to me. You can close the conversation :)

joergroedel:
Well done! I especially like the new macros for GHCB getters/setters and the test-case checking the GHCB offsets.

PR #332, opened 2024-04-25
mm/alloc: track slab item sizes at compile time

00xc:
* Store the slab item size instead of the slab pointer in the page info. This simplifies `layout_from_ptr()` and the initialization of several structures.
* Track slab item sizes at compile time via const generics. This removes several runtime fields in a couple of structures and simplifies initialization further.

### DIFF
--- a/fuzz/fuzz_targets/page_alloc.rs
+++ b/fuzz/fuzz_targets/page_alloc.rs
@@ -33,7 +33,7 @@ enum Action {
     /// Allocate a regular page
     Allocate,
     /// Allocate a slab page
-    AllocateSlab,
+    AllocateSlab(u16),
     /// Allocate pages of higher order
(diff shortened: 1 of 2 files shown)


00xc:
I'm currently running Miri tests to make sure nothing is broken with the new changes.

00xc:
> I'm currently running Miri tests to make sure nothing is broken with the new changes.

Miri ran cleanly

joergroedel:
The changes break `make test-in-svsm`, the test binary hangs after stage2.

00xc:
> The changes break `make test-in-svsm`, the test binary hangs after stage2.

Looking into it

00xc:
Tests and regular boot both work for me now, both on debug and release builds. @joergroedel could you double check?

PR #338, opened 2024-05-03
kernel: Support restricted injection

msft-jlange:
This PR adds the required logic to support restricted injection of interrupts via the #HV protocol.  No interrupt handlers are defined in this PR.

### DIFF
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -153,7 +153,7 @@ extern "C" fn ex_handler_general_protection(ctxt: &mut X86ExceptionContext) {
 
 // Page-Fault handler
 #[no_mangle]
-extern "C" fn ex_handler_page_fault(ctxt: &mut X86ExceptionContext) {
+extern "C" fn ex_handler_page_fault(ctxt: &mut X86ExceptionContext, vector: usize) {
     let cr2 = read_cr2();
     let rip = ctxt.frame.rip;
     let err = ctxt.error_code;
@@ -178,7 +178,7 @@ extern "C" fn ex_handler_page_fault(ctxt: &mut X86ExceptionContext) {
         .is_err()
         && !handle_exception_table(ctxt)
     {
-        handle_debug_exception(ctxt, ctxt.vector);
+        handle_debug_exception(ctxt, vector);
         panic!(
             "Unhandled Page-Fault at RIP {:#018x} CR2: {:#018x} error code: {:#018x}",
             rip, cr2, err
@@ -197,11 +197,11 @@ extern "C" fn ex_handler_hypervisor_injection(_ctxt: &mut X86ExceptionContext) {
 
 // VMM Communication handler
 #[no_mangle]
-extern "C" fn ex_handler_vmm_communication(ctxt: &mut X86ExceptionContext) {
+extern "C" fn ex_handler_vmm_communication(ctxt: &mut X86ExceptionContext, vector: usize) {
     let rip = ctxt.frame.rip;
     let code = ctxt.error_code;
 
-    if let Err(err) = handle_vc_exception(ctxt) {
+    if let Err(err) = handle_vc_exception(ctxt, vector) {
         log::error!("#VC handling error: {:?}", err);
         if user_mode(ctxt) {
             log::error!("Failed to handle #VC from user-mode at RIP {:#018x} code: {:#018x} - Terminating task", rip, code);
@@ -235,15 +235,14 @@ extern "C" fn ex_handler_system_call(ctxt: &mut X86ExceptionContext) {
 }
 
 #[no_mangle]
-pub extern "C" fn ex_handler_panic(ctx: &mut X86ExceptionContext) {
-    let vec = ctx.vector;
+pub extern "C" fn ex_handler_panic(ctx: &mut X86ExceptionContext, vector: usize) {
     let rip = ctx.frame.rip;
     let err = ctx.error_code;
     let rsp = ctx.frame.rsp;
     let ss = ctx.frame.ss;
     panic!(
         "Unhandled exception {} RIP {:#018x} error code: {:#018x} RSP: {:#018x} SS: {:#x}",
-        vec, rip, err, rsp, ss
+        vector, rip, err, rsp, ss
     );
 }
 
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -16,8 +16,10 @@ use super::common::{
     OF_VECTOR, PF_VECTOR, SS_VECTOR, SX_VECTOR, TS_VECTOR, UD_VECTOR, VC_VECTOR, XF_VECTOR,
 };
 use crate::address::VirtAddr;
+use crate::cpu::percpu::this_cpu_unsafe;
 use crate::cpu::X86ExceptionContext;
 use crate::debug::gdbstub::svsm_gdbstub::handle_debug_exception;
+use crate::platform::SVSM_PLATFORM;
 use crate::task::{is_task_fault, terminate};
 
 use core::arch::global_asm;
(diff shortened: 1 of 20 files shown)


kernel/src/cpu/idt/svsm.rs
```diff
@@ -94,6 +99,20 @@ pub fn early_idt_init() {
...
+    // holds the actual pointer is constant across all CPUs, so that is the
+    // pointer that is actually captured.  The address that is captured is
+    // stored as a usize instead of a typed value, because the declarations
+    // required for type safety here are cumbersome, and the assembly code
+    // that uses the value is not type safe in any case, so enforcing type
+    // safety on the pointer would offer no meaningful value.
+    unsafe {
+        let global_ref = ptr::addr_of!(HV_DOORBELL_ADDR) as usize;
+        *(global_ref as *mut usize) = (*this_cpu_unsafe()).hv_doorbell_addr();
+    }
```
00xc, line 115: We can just do this if we set `HV_DOORBELL_ADDR` to be `static mut`.

```suggestion
    unsafe {
        let global_ref = ptr::addr_of_mut!(HV_DOORBELL_ADDR);
        *global_ref = (*this_cpu_unsafe()).hv_doorbell_addr();
    }
```

kernel/src/cpu/percpu.rs
```diff
@@ -445,6 +458,32 @@ impl PerCpu {
...
+
+        unsafe {
+            let cpu_unsafe = self.cpu_unsafe as *mut PerCpuUnsafe;
+            (*cpu_unsafe).hv_doorbell = paddr.as_mut_ptr::<HVDoorbell>();
+        }
+
+        Ok(())
+    }
+
+    pub fn configure_hv_doorbell(&self) -> Result<(), SvsmError> {
```
cxdong, line 477: Wondering if this is something feasible to be implemented in the platform abstraction layer?

kernel/src/cpu/idt/svsm.rs
```diff
@@ -94,6 +99,20 @@ pub fn early_idt_init() {
...
+    // different, but the address of the field in the PerCpu structure that
+    // holds the actual pointer is constant across all CPUs, so that is the
+    // pointer that is actually captured.  The address that is captured is
+    // stored as a usize instead of a typed value, because the declarations
+    // required for type safety here are cumbersome, and the assembly code
+    // that uses the value is not type safe in any case, so enforcing type
+    // safety on the pointer would offer no meaningful value.
+    unsafe {
+        let global_ref = ptr::addr_of_mut!(HV_DOORBELL_ADDR);
+        *global_ref = (*this_cpu_unsafe()).hv_doorbell_addr();
```
00xc, line 114: Now that I think about it, is there any reason why we're not just doing this?

```rust
    HV_DOORBELL_ADDR = (*this_cpu_unsafe()).hv_doorbell_addr();
```

For some reason I thought `HV_DOORBELL_ADDR` was an extern symbol, but it just being a global we can just directly assign.

kernel/src/cpu/percpu.rs
msft-jlange: Yes, at some point.  However, some aspects of this code will be needed for multiple platforms to support presentation of interrupts to lower privileged guest contexts, so the abstraction requires a bit more factoring.  Also, the core platform abstraction has not yet evolved to fully define how initialization of the PerCpu and related structures should be defined.  Consequently, I'd prefer to leave this as it is for now and work on the refactoring as the platform abstraction evolves.

kernel/src/cpu/idt/svsm.rs
msft-jlange: I had assumed that would be illegal because assigning to a mutable static is incompatible with safe multi-threaded code.  However, it seems that the compiler will happily allow that inside of an `unsafe` block, so I'll just do it that way.  Thanks for the suggestion!

00xc: Yes, it is unsafe, in the same way that reading or writing to a pointer is not thread-safe because the pointer may be aliased from a different thread.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -33,4 +44,46 @@ impl HVDoorbell {
...
+        }
+
+        // Consume interrupts as long as they are available.
+        loop {
+            // Consume any interrupt that may be present.
+            let vector = self.vector.swap(0, Ordering::Relaxed);
+            if vector == 0 {
+                break;
+            }
+            common_isr_handler(vector as usize);
```
roy-hopkins, line 75: Reading through the GHCB specification and particularly the host pseudocode in section 5.5.1, it looks like this loop may not work as expected.

Firstly, `no_further_signal` is cleared above before reading the vector and exchanging it with zero. This means that if a further NMI is pending then another #HV can be issued which can cause this code to re-enter. I'm not sure if that's a problem or not.

Secondly, if I understand this correctly, as soon as the vector is swapped with 0 then the next vector that the host will populate in the doorbell page will result in a new #HV exception as both `PendingEvent.Vector==0` and `PendingEvent.NoFurtherSignal==0`. This will mean that it will be handled in a re-entered #HV handler meaning that the second iteration of this loop will always result in `vector == 0`.

Finally, could the host potentially change the flags between reading them above and reading the vector? If so, should the vector and flags be read together atomically here and the flags checked again?

msft-jlange: > Firstly, `no_further_signal` is cleared above before reading the vector and exchanging it with zero. This means that if a further NMI is pending then another #HV can be issued which can cause this code to re-enter. I'm not sure if that's a problem or not.
> 
> Secondly, if I understand this correctly, as soon as the vector is swapped with 0 then the next vector that the host will populate in the doorbell page will result in a new #HV exception as both `PendingEvent.Vector==0` and `PendingEvent.NoFurtherSignal==0`. This will mean that it will be handled in a re-entered #HV handler meaning that the second iteration of this loop will always result in `vector == 0`.

The #HV handler is an interrupt gate, which means that it executes with EFLAGS.IF=0.  The first thing the #HV handler itself does is to check to see whether it was invoked with EFLAGS.IF=0, and if so, it will return immediately (after checking to see whether it was in the return-to-VMPL tail or the return-from-trap tail).  That means that any recursive #HV will essentially be ignored until such time as it can be handled.  Because the vector processing loop here runs with interrupts disabled, any recursive #HV will simply be ignored.

> Finally, could the host potentially change the flags between reading them above and reading the vector? If so, should the vector and flags be read together atomically here and the flags checked again?

I don't see any issue here.  If the host clears `no_further_signal` on its own, then it is violating the protocol and we can't do anything useful about it.  If the host sets `no_further_signal` while the loop is executing, then it will be caught during the IRET flow out of the #HV handler, which would just cause the loop to run again.  It might be slightly more optimal to clear `no_further_signal` every time through the loop but there is no correctness issue by failing to do so.

joergroedel:
Hey @msft-jlange, can you please rebase these changes to the latest main branch? I merged it locally and it booted fine, so I will merge it after the update.

There is one follow-on request: Can you please submit an additional PR (or include it when updating this one) adding a comment with pseudo-code describing the flow of the HV handling assembly? That will make it easier to understand the code when someone needs to touch it in the future.

msft-jlange:
> Hey @msft-jlange, can you please rebase these changes to the latest main branch? I merged it locally and it booted fine, so I will merge it after the update.

Done.  Hopefully the merge didn't break anything.

> There is one follow-on request: Can you please submit an additional PR (or include it when updating this one) adding a comment with pseudo-code describing the flow of the HV handling assembly? That will make it easier to understand the code when someone needs to touch it in the future.

I've expanded the comments as part of this PR.  Hopefully they add the clarity you are looking for.  It's not pseudocode, but it is a much more detailed explanation of what the code is trying to do.

joergroedel:
Thanks @msft-jlange. With the comments it becomes a lot easier to maintain. Testing on my side was also successful.

MelodyHuibo:
> Thanks @msft-jlange. With the comments it becomes a lot easier to maintain. Testing on my side was also successful.

Hi @joergroedel , Could you share the details that how did you do the testing for it?

joergroedel:
> Hi @joergroedel , Could you share the details that how did you do the testing for it?

I did not test the functionality itself, as support in KVM is still missing. I tested it in my environment without restricted injection support and made sure there are no regressions.

PR #341, opened 2024-05-10
fuzz: do not use default features when building the SVSM

00xc:
Do not use the default features when building the SVSM for fuzzing, as linking `libmstpm` breaks the fuzzers.


00xc:
Without this PR fuzzing harnesses do not work, as they either hang or crash when starting up.

PR #342, opened 2024-05-13
svsm: remove `default-test` feature & fix leftover nightly clippy warning

00xc:
* Remove the `default-test` feature: it is not really needed, as we can detect if we're building tests via the "test" cfg.
* Remove a nightly clippy warning in the gdbstub code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 FEATURES ?= "default"
 SVSM_ARGS = --features ${FEATURES}
 
-FEATURES_TEST ?= "default-test"
-SVSM_ARGS_TEST = --no-default-features --features ${FEATURES_TEST}
+SVSM_ARGS_TEST = --no-default-features
+ifdef FEATURES_TEST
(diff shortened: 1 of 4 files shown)


cclaudio (approved):
Looks good to me, thanks.

Makefile
```diff
@@ -1,8 +1,10 @@
 FEATURES ?= "default"
 SVSM_ARGS = --features ${FEATURES}
 
-FEATURES_TEST ?= "default-test"
-SVSM_ARGS_TEST = --no-default-features --features ${FEATURES_TEST}
+SVSM_ARGS_TEST = --no-default-features
+ifdef FEATURES_TEST
+	SVSM_ARGS_TEST := ${SVSM_ARGS_TEST} --features ${FEATURES_TEST}
```
stefano-garzarella, line 6: What about using `+=` to concatenate the new part?

```suggestion
	SVSM_ARGS_TEST += --features ${FEATURES_TEST}
```

stefano-garzarella (approved):

PR #343, opened 2024-05-14
[RFC] virtio-blk storage via mmio using virtio-drivers (Qemu)

osteffenrh:
Add support for virtio-blk storage devices using the virtio-mmio transport.

The [virtio-drivers](https://github.com/rcore-os/virtio-drivers/) crate from the rcore-os project is used. It required some [modifications to support custom MMIO access functions](https://github.com/osteffenrh/virtio-drivers/pull/1).
Crate License: MIT.

This PR requires a [patched version of Qemu](https://github.com/coconut-svsm/qemu/pull/13) , adding virtio-mmio slots to the Q35 machine
model.

## ToDo
For persistent storage to be usable we still need to:
- [ ] Copy virtio-drivers into the repo
- [ ] Set up some basic Fuzzing
- [ ] Improve memory management:
  - Allow buffers > one page
  - Allocate buffers in a smarter/more compact way?
- [ ] Add block device layer to Coconut and integrate vitio-blk into it
- [ ] Add encryption layer
- [ ] Add a simple file system on top of the block layer. Choices?
- [ ] Block access to the MMIO ranges for `VMPL!=0`
- [ ] Supply MMIO base address via IGVM file

### Building & Preparations

Build the patched Qemu as usual with IGVM support.
```
git clone https://github.com/osteffenrh/qemu.git --branch virtio-pr --single-branch --depth=1
cd qemu
$ ./configure --prefix=$HOME/bin/qemu-svsm/ --target-list=x86_64-softmmu --enable-igvm
ninja -C build/
make install
```

Build Coconut as usual.

Create an empty disk image for Coconut to use
```
truncate -s16M mmio.raw
```

### Launching Qemu

Add
```
-machine x-svsm-virtio-mmio=on
```
and
```
-global virtio-mmio.force-legacy=false \
-drive file="mmio.raw",format=raw,if=none,id=mmio \
-device virtio-blk-device,drive=mmio
````
to your Qemu command line.

Full example:
```
$HOME/bin/qemu-svsm/bin/qemu-system-x86_64 \
  -enable-kvm \
  -cpu EPYC-v4 \
  -smp 4 \
  -machine q35,confidential-guest-support=sev0,memory-backend=mem0 \
  -machine x-svsm-virtio-mmio=on \
  -object memory-backend-memfd,id=mem0,size=4G,share=true,prealloc=false,reserve=false \
  -object sev-snp-guest,id=sev0,cbitpos=51,reduced-phys-bits=1,init-flags=1,igvm-file="bin/coconut-qemu.igvm" \
  -no-reboot \
  -drive file="disk.qcow2",if=none,id=disk0,format=qcow2,snapshot=on \
  -device virtio-scsi-pci,id=scsi0,disable-legacy=on,iommu_platform=true \
  -device scsi-hd,drive=disk0 \
  -nographic \
  -serial stdio \
  -monitor none \
  -global virtio-mmio.force-legacy=false \
  -drive file="mmio.raw",format=raw,if=none,id=mmio \
  -device virtio-blk-device,drive=mmio
```

A short write-read test will be run on the virtio-blk device, followed by a
simple write speed benchmark. This requires external time measurement.

Example output (timestamps added externally):
```
   0.0011s   [SVSM] Virtio test trace
   0.0004s   [SVSM] Detected virtio MMIO device with vendor id 0x554D4551, device type Block, version Modern
   0.0006s   [SVSM] config: 0xfef03100
   0.0007s   [SVSM] found a block device of size 131072KB
   0.0173s   [SVSM] Write+Read Test Start
   0.0002s   [SVSM] Write+Read Test End
   0.0005s   [SVSM] Write Benchmark Start
  65.2324s   [SVSM] virtio-blk start write. Block size = 512
   0.0005s   [SVSM] virtio-blk end write. Block size = 512, Total bytes = 134217728
   8.1706s   [SVSM] virtio-blk start write. Block size = 4096
   0.0003s   [SVSM] virtio-blk end write. Block size = 4096, Total bytes = 134217728
   0.0008s   [SVSM] Write Benchmark End
   0.0005s   [SVSM] Virtio test end
```

## Write Speed

Write speed depends on the block size used.
From the output above we get:
- 512Byte writes: 1.96MB/s
- 4096Byte writes: 15.7MB/s

### DIFF
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -96,6 +96,8 @@ enum GHCBExitCode {
     IOIO = 0x7b,
     MSR = 0x7c,
     RDTSCP = 0x87,
+    MMIO_READ = 0x80000001,
+    MMIO_WRITE = 0x80000002,
     SNP_PSC = 0x8000_0010,
     GUEST_REQUEST = 0x8000_0011,
     GUEST_EXT_REQUEST = 0x8000_0012,
@@ -435,6 +437,28 @@ impl GHCB {
         Ok(())
     }
 
+    fn read_buffer_as<T>(&self, offset: usize) -> Result<T, GhcbError>
+    where
+        T: Sized + Copy,
+    {
+        offset
+            .checked_add(mem::size_of::<T>())
+            .filter(|end| *end <= GHCB_BUFFER_SIZE)
+            .ok_or(GhcbError::InvalidOffset)?;
+
+        // SAFETY: we have verified that offset is within bounds and does not
+        // overflow
+        let src = unsafe { self.buffer.as_ptr().add(offset) };
+
+        if src.align_offset(mem::align_of::<T>()) != 0 {
+            return Err(GhcbError::InvalidOffset);
+        }
+
+        // SAFETY: we have verified the pointer is aligned, as well as within
+        // bounds.
+        unsafe { Ok(src.cast::<T>().read_volatile()) }
+    }
+
     fn write_buffer<T>(&self, data: &T, offset: usize) -> Result<(), GhcbError>
     where
         T: Copy,
@@ -558,6 +582,33 @@ impl GHCB {
         let exit_info_2: u64 = vmsa_gpa.into();
         self.set_rax_valid(sev_features);
         self.vmgexit(GHCBExitCode::AP_CREATE, exit_info_1, exit_info_2)?;
+        Ok(())
+    }
+
+    pub fn mmio_read<T: Sized + Copy>(&self, pa: PhysAddr) -> Result<T, SvsmError> {
+        self.clear();
+        let buffer_va = VirtAddr::from(self.buffer.as_ptr());
+        let buffer_pa = u64::from(virt_to_phys(buffer_va));
+        self.set_sw_scratch_valid(buffer_pa);
+        self.vmgexit(
+            GHCBExitCode::MMIO_READ,
+            u64::from(pa),
+            mem::size_of::<T>() as u64,
+        )?;
+        Ok(self.read_buffer_as::<T>(0)?)
+    }
+
+    pub fn mmio_write<T: Copy + Sized>(&self, pa: PhysAddr, value: &T) -> Result<(), SvsmError> {
+        self.clear();
+        self.write_buffer(value, 0)?;
+        let buffer_va = VirtAddr::from(self.buffer.as_ptr());
+        let buffer_pa = u64::from(virt_to_phys(buffer_va));
+        self.set_sw_scratch_valid(buffer_pa);
+        self.vmgexit(
+            GHCBExitCode::MMIO_WRITE,
+            u64::from(pa),
+            mem::size_of::<T>() as u64,
+        )?;
         Ok(())
     }
 
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -446,13 +446,13 @@ impl GHCB {
 
         // SAFETY: we have verified that the offset is within bounds and does
         // not overflow
-        let dst = unsafe { self.buffer.as_ptr().cast::<u8>().add(offset) };
+        let dst = unsafe { self.buffer.as_ptr().cast::<u8>().add(offset).cast::<T>() };
         if dst.align_offset(mem::align_of::<T>()) != 0 {
             return Err(GhcbError::InvalidOffset);
         }
 
-        // SAFETY: we have verified the pointer is aligned and within bounds.
-        unsafe { dst.cast::<T>().copy_from_nonoverlapping(data, 1) }
+        unsafe { dst.write_volatile(*data) };
+
         Ok(())
     }
 
--- a/kernel/src/virtio/mod.rs
+++ b/kernel/src/virtio/mod.rs
@@ -1,0 +1,218 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2024 Red Hat, Inc.
+//
+// Author: Oliver Steffen <osteffen@redhat.com>
+
+use core::ptr::{addr_of, NonNull};
+
+use virtio_drivers::{
+    device::blk::{VirtIOBlk, SECTOR_SIZE},
+    transport::{
+        mmio::{MmioTransport, VirtIOHeader},
+        DeviceType, Transport,
+    },
+};
+
+use crate::{
+    address::{PhysAddr, VirtAddr},
+    cpu::{self, percpu::this_cpu},
+    mm::{alloc::*, page_visibility::*, *},
+};
+
+struct SvsmHal;
+
+/// Implementation of virtio-drivers MMIO hardware abstraction for AMD SEV-SNP
+/// in the Coconut-SVSM context. Due to missing #VC handler for MMIO, use ghcb exits
+/// instead.
+unsafe impl virtio_drivers::Hal for SvsmHal {
+    fn dma_alloc(
+        pages: usize,
+        _direction: virtio_drivers::BufferDirection,
+    ) -> (virtio_drivers::PhysAddr, NonNull<u8>) {
+        // TODO: allow more than one page.
+        //       This currently works, becasue in "modern" virtio mode the crate only allocates
+        //       one page at a time.
+        assert!(pages == 1);
+
+        let mem = allocate_zeroed_page().expect("Error allocating page");
+        make_page_shared(mem).expect("Error making page shared");
+
+        (virt_to_phys(mem).into(), unsafe {
+            NonNull::<u8>::new_unchecked(mem.as_mut_ptr())
+        })
+    }
+
+    unsafe fn dma_dealloc(
+        _paddr: virtio_drivers::PhysAddr,
+        vaddr: NonNull<u8>,
+        pages: usize,
+    ) -> i32 {
+        //TODO: allow more than one page
+        assert!(pages == 1);
+
+        make_page_private(vaddr.as_ptr().into()).expect("Error making page private");
+        free_page(vaddr.as_ptr().into());
+
+        0
+    }
+
+    unsafe fn mmio_phys_to_virt(paddr: virtio_drivers::PhysAddr, _size: usize) -> NonNull<u8> {
+        NonNull::new(phys_to_virt(paddr.into()).as_mut_ptr())
+            .expect("Error getting VirtAddr from PhysAddr")
+    }
+
+    unsafe fn share(
+        buffer: NonNull<[u8]>,
+        direction: virtio_drivers::BufferDirection,
+    ) -> virtio_drivers::PhysAddr {
+        // TODO: allow more than one page
+        assert!(buffer.len() <= PAGE_SIZE);
+
+        let mem = allocate_zeroed_page().expect("Error allocating zeroed page");
+        let phys = virt_to_phys(mem);
+
+        make_page_shared(mem).expect("Error making page shared");
+
+        if direction == virtio_drivers::BufferDirection::DriverToDevice {
+            unsafe {
+                let src = buffer.as_ptr().cast::<u8>();
+                let dst = mem.as_mut_ptr::<u8>();
+                core::ptr::copy_nonoverlapping(src, dst, buffer.len());
+            }
+        }
+
+        phys.into()
+    }
+
+    unsafe fn unshare(
+        paddr: virtio_drivers::PhysAddr,
+        buffer: NonNull<[u8]>,
+        direction: virtio_drivers::BufferDirection,
+    ) {
+        assert!(buffer.len() <= PAGE_SIZE);
+
+        let vaddr = phys_to_virt(paddr.into());
+
+        if direction == virtio_drivers::BufferDirection::DeviceToDriver {
+            unsafe {
+                let dst = buffer.as_ptr().cast::<u8>();
+                let src = vaddr.as_mut_ptr::<u8>();
+                core::ptr::copy_nonoverlapping(src, dst, buffer.len());
+            }
+        }
+        make_page_private(vaddr).expect("Error making page private");
+
+        free_page(phys_to_virt(paddr.into()));
+    }
+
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by joergroedel** [comment]
Wow, cool! We might also think about using IRQs once the COCONUT kernel has the infrastructure for it.

**Comment by 00xc** [comment]
> Add support for virtio-blk storage devices using the virtio-mmio transport. MMIO accesses are done via explicit vmgexits. Interrupts are not used (-> polling).
> 
> The [virtio-drivers](https://github.com/rcore-os/virtio-drivers/) crate from the rcore-os project is used. It required some [modifications to support custom MMIO access functions](https://github.com/osteffenrh/virtio-drivers/pull/1). Crate License: MIT.

I haven't gone too deep in the codebase, but can't this be done by implementing [`Transport`](https://docs.rs/virtio-drivers/0.7.2/virtio_drivers/transport/trait.Transport.html) for a new [`MmioTransport`](https://docs.rs/virtio-drivers/0.7.2/virtio_drivers/transport/mmio/struct.MmioTransport.html)-like type?

**Comment by osteffenrh** [comment]
> > Add support for virtio-blk storage devices using the virtio-mmio transport. MMIO accesses are done via explicit vmgexits. Interrupts are not used (-> polling).
> > The [virtio-drivers](https://github.com/rcore-os/virtio-drivers/) crate from the rcore-os project is used. It required some [modifications to support custom MMIO access functions](https://github.com/osteffenrh/virtio-drivers/pull/1). Crate License: MIT.
> 
> I haven't gone too deep in the codebase, but can't this be done by implementing [`Transport`](https://docs.rs/virtio-drivers/0.7.2/virtio_drivers/transport/trait.Transport.html) for a new [`MmioTransport`](https://docs.rs/virtio-drivers/0.7.2/virtio_drivers/transport/mmio/struct.MmioTransport.html)-like type?

Unfortunately not, because the devices (virtio-blk for example) use mmio accesses directly too (to the configuration space), without going through the transport.

**Comment by osteffenrh** [comment]
To be clear, there is a MMIO transport in the virtio-drivers crate. But it simply uses volatile reads and writes by default. My change here was to add mmio access functions to the hardware abstraction trait, to be able to use explicit vmgextis instead of "plain" volatile accesses.
Since we don't have a `#VC` handler yet that supports MMIO, that seemed like the way to go.

**Comment by kraxel** [comment]
> Wow, cool! We might also think about using IRQs once the COCONUT kernel has the infrastructure for it.

Not convinced this makes sense, so my qemu patch doesn't assign interrupts to these virtio-mmio transport slots.
Do we want run virtio devices asynchronously?  Is the complexity worth the performance benefit?

When using interrupts:  How would that look like?  Would SVSM (partly) emulate lapic and ioapic then, to make sure the guest os can't tamper with the IRQ lines owned by SVSM?

File: kernel/src/sev/ghcb.rs [hunk 1]
```diff
@@ -384,6 +386,27 @@ impl GHCB {
...
+        unsafe {
+            let src = self
+                .buffer
+                .as_mut_ptr()
+                .cast::<u8>()
+                .offset(offset)
+                .cast::<T>();
+            Ok(*src)
+        }
+    }
```
00xc, line 408: We do not need to use `isize` here. I understand this was perhaps done to mirror `write_buffer()` below, but we can do the right thing here (and the caller), i.e., use `usize`, and thus [`add()`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add-1) instead of `offset()`

Also, there are some safety issues:

* We need to check that `offset + size` does not overflow (otherwise `add()` / `offset()` [might cause UB](https://doc.rust-lang.org/std/primitive.pointer.html#safety-13). 
* We need to check that `src` has the alignment requirements for `T`. We should either add a check *or* use [`read_unaligned()`](https://doc.rust-lang.org/std/primitive.pointer.html#method.read_unaligned). Otherwise this might cause UB.

Something like this:

```suggestion
    fn read_buffer_as<T>(&mut self, offset: usize) -> Result<T, GhcbError>
    where
        T: Sized + Copy,
    {
        offset
            .checked_add(mem::size_of::<T>())
            .filter(|end| *end <= GHCB_BUFFER_SIZE)
            .ok_or(GhcbError::InvalidOffset)?;

        // SAFETY: we have verified that offset is within bounds and does not
        // overflow
        let src = unsafe { self.buffer.as_mut_ptr().add(offset) };

        if src.align_offset(mem::align_of::<T>()) != 0 {
            return Err(GhcbError::InvalidOffset);
        }
        // SAFETY: we have verified the pointer is aligned, as well as within
        // bounds.
        unsafe { Ok(src.cast::<T>().read()) }
    }
```

I believe these additional checks should not cause a noticeable performance hit. As a last resort, we could make this function (as well as `write_buffer()`) unsafe and have the caller ensure the safety requirements are met.

kernel/src/virtio/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+                core::ptr::copy_nonoverlapping(src, dst, buffer.len());
+            }
+        }
+        make_page_private(vaddr).expect("Error making page private");
+
+        free_page(phys_to_virt(paddr.into()));
+    }
+
+    unsafe fn mmio_read<T: Sized + Copy>(src: &T) -> T {
+        let paddr = PhysAddr::from((src as *const T) as u64);
```
00xc, line 110: Not sure I understand this, why are we not going through `virt_to_phys()`?

kernel/src/virtio/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        cpu::ghcb::current_ghcb()
+            .mmio_read::<T>(paddr)
+            .expect("GHCB MMIO Read failed")
+    }
+
+    unsafe fn mmio_write<T: Sized>(dst: &mut T, v: T) {
+        let paddr = PhysAddr::from((dst as *mut T) as u64);
```
00xc, line 118: Same here, why are we converting a regular reference to a `PhysAddr` directly?

kernel/src/virtio/mod.rs
osteffenrh: Since we can't access mmio locations directly and have to go through the ghcb mmio functions anyway, the mmio area is not mapped.

Instead, the physical MMIO base address is cast to the `VirtIOHeader` struct. This only works because the struct members are never actually accessed directly. It only serves as a fancy way to calculate the correct offsets of the individual registers. (This is just how the crate does it).

It should be possible to map the mmio range, cast that as `VirtIOHeader` and then translate back to the paddr when doing the mmio calls.

kernel/src/sev/ghcb.rs
00xc: This sounds like UB. Can you try with `read_volatile()` instead? The compiler probably cannot infer that the buffer changes after a `VMGEXIT`.

osteffenrh: Tried it. Still the same: hangs.

Also very strange: The last thing I see is "Validating memory", which happens way before
the virtio code (which is the only place calling this) runs.

00xc: One last try: could you update `write_buffer` so that it uses `write_volatile()` as well? E.g. something like:

```diff
diff --git a/kernel/src/sev/ghcb.rs b/kernel/src/sev/ghcb.rs
index 9a5fe44..682d2cf 100644
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -386,7 +386,7 @@ impl GHCB {
 
     fn write_buffer<T>(&mut self, data: &T, offset: isize) -> Result<(), GhcbError>
     where
-        T: Sized,
+        T: Copy,
     {
         let size: isize = mem::size_of::<T>() as isize;
 
@@ -401,9 +401,8 @@ impl GHCB {
                 .cast::<u8>()
                 .offset(offset)
                 .cast::<T>();
-            let src = data as *const T;
 
-            ptr::copy_nonoverlapping(src, dst, 1);
+            dst.write_volatile(*data);
         }
 
         Ok(())
```

Perhaps with both of them being volatile the compiler can make the right assumptions.

kernel/src/virtio/mod.rs
00xc: For a temporary mapping use `PerCPUPageMappingGuard`. For a permanent mapping you would probably need to call into `this_cpu().get_pgtable().map_region(..)`.

kernel/src/sev/ghcb.rs
osteffenrh: This is strange. I retried the approaches mentioned above (read, read_volatile, read_volatile + write_volatile, and the plain "cast" line) and now they all work (virtio-blk tests code runs).
Before Coconut would hang reliably on all approaches except the one I mentioned. :man_shrugging: 

So, shall we use `read_volatile` and `write_volatile` to be safe? Or what would you recommend?

00xc: Have you tried both debug and release modes? Anyhow, yes, we should probably use the volatile versions to be safe.

osteffenrh: I tried both, debug and release.

Using volatile accesses now.

kernel/src/virtio/mod.rs
osteffenrh: Mapping the MMIO range works also now:
```diff
diff --git a/kernel/src/virtio/mod.rs b/kernel/src/virtio/mod.rs
index 0f9d8b8..63da985 100644
--- a/kernel/src/virtio/mod.rs
+++ b/kernel/src/virtio/mod.rs
@@ -4,7 +4,7 @@
 //
 // Author: Oliver Steffen <osteffen@redhat.com>
 
-use core::ptr::NonNull;
+use core::ptr::{addr_of, NonNull};
 
 use virtio_drivers::{
     device::blk::{VirtIOBlk, SECTOR_SIZE},
@@ -15,9 +15,7 @@ use virtio_drivers::{
 };
 
 use crate::{
-    address::PhysAddr,
-    cpu,
-    mm::{alloc::*, page_visibility::*, *},
+    address::{PhysAddr, VirtAddr}, cpu::{self, percpu::this_cpu}, mm::{alloc::*, page_visibility::*, *}
 };
 
 struct SvsmHal;
@@ -107,7 +105,7 @@ unsafe impl virtio_drivers::Hal for SvsmHal {
     }
 
     unsafe fn mmio_read<T: Sized + Copy>(src: &T) -> T {
-        let paddr = PhysAddr::from((src as *const T) as u64);
+        let paddr = this_cpu().get_pgtable().phys_addr(VirtAddr::from(addr_of!(*src))).unwrap();
 
         cpu::ghcb::current_ghcb()
             .mmio_read::<T>(paddr)
@@ -115,7 +113,7 @@ unsafe impl virtio_drivers::Hal for SvsmHal {
     }
 
     unsafe fn mmio_write<T: Sized + Copy>(dst: &mut T, v: T) {
-        let paddr = PhysAddr::from((dst as *mut T) as u64);
+        let paddr = this_cpu().get_pgtable().phys_addr(VirtAddr::from(addr_of!(*dst))).unwrap();
 
         cpu::ghcb::current_ghcb()
             .mmio_write::<T>(paddr, &v)
@@ -127,8 +125,12 @@ unsafe impl virtio_drivers::Hal for SvsmHal {
 pub fn test_mmio() {
     static MMIO_BASE: u64 = 0xfef03000; // Hard-coded in Qemu
 
+    let paddr = PhysAddr::from(MMIO_BASE);
+    let mem = PerCPUPageMappingGuard::create_4k(paddr).expect("Error mapping MMIO region");
+
+    log::info!("mapped MMIO range {:016x} to vaddr {:016x}", MMIO_BASE, mem.virt_addr());
     // Test code below taken from virtio-drivers aarch64 example.
-    let header = NonNull::new(MMIO_BASE as *mut VirtIOHeader).unwrap();
+    let header = NonNull::new(mem.virt_addr().as_mut_ptr() as *mut VirtIOHeader).unwrap();
     match unsafe { MmioTransport::<SvsmHal>::new(header) } {
         Err(e) => log::warn!(
             "Error creating VirtIO MMIO transport at {:016x}: {}",
```

kernel/src/virtio/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+/// instead.
+unsafe impl virtio_drivers::Hal for SvsmHal {
+    fn dma_alloc(
+        pages: usize,
+        _direction: virtio_drivers::BufferDirection,
+    ) -> (virtio_drivers::PhysAddr, NonNull<u8>) {
+        // TODO: allow more than one page.
+        //       This currently works, becasue in "modern" virtio mode the crate only allocates
+        //       one page at a time.
+        assert!(pages == 1);
```
00xc, line 36: I think just using `allocate_pages()` with some extra logic to zero them and make them all shared should work. right?

osteffenrh:
Rebased onto f23151f6b4a1a1252c36e3b3a5ec1e0d1ac3bdcc from Sept 16.
Going to current main branch requires more work due to changes to `ghcb::write_buffer()`
(now uses traits from zerocopy).

This now uses the latest virtio-drivers crate, see https://github.com/osteffenrh/virtio-drivers/pull/1

Freax13:
> Going to current main branch requires more work due to changes to `ghcb::write_buffer()` (now uses traits from zerocopy).

Feel free to let me know if you need any help with that.

osteffenrh:
> > Going to current main branch requires more work due to changes to `ghcb::write_buffer()` (now uses traits from zerocopy).
> 
> Feel free to let me know if you need any help with that.

Here is some discussion on how this could be done: https://github.com/google/zerocopy/issues/1919

kernel/src/sev/ghcb.rs
```diff
@@ -384,9 +386,31 @@ impl GHCB {
         Ok(())
     }
 
+    fn read_buffer_as<T>(&mut self, offset: usize) -> Result<T, GhcbError>
+    where
+        T: Sized + Copy,
```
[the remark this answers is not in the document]
Freax13: Furthermore, `Copy` implies `Sized` so this is doubly-redundant :)

osteffenrh:
Closing this in favor of https://github.com/coconut-svsm/svsm/pull/542

PR #345, opened 2024-05-16
Highlight build issue with std and no_std in the same workspace

stefano-garzarella:
PLEASE DON'T MERGE!

This commit is just to highlight the problem I am having. I have minimized the changes to only discuss the problem.

In this case I added a dependency (rdrand) that supports std or no_std with the (std) feature enabled by default. Here I am disabling it for use in the kernel without std.

Dependencies
```
- kernel
  -> rdrand (default-features = false, so "std" feature disabled)
      -> rand_core (default-features = false, so "std" feature disabled)
          -> getrandom (default-features = false, so it should be disabled)
- igvmmeasure
  -> p384 (default-features = true)
      -> elliptic-curve
          -> rand_core
              -> getrandom
```
But when building `svsm`, cargo seems to use an interesention of the features used in the workspace, as it goes to compile `getrandom` which should be disabled, more expecting std.

```
$ make
cargo build --manifest-path kernel/Cargo.toml  --no-default-features --bin stage2
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.00s
objcopy -O binary "target/x86_64-unknown-none/debug/stage2" bin/stage2.bin
touch bin/svsm-fs.bin
cargo build  --features "default" --bin svsm
   Compiling getrandom v0.2.12
   Compiling zerocopy v0.7.32
   Compiling zerocopy v0.6.6
   Compiling crypto-common v0.1.6
   Compiling inout v0.1.3
error[E0463]: can't find crate for `std`
 --> /home/stefano/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.12/src/error_impls.rs:2:1
  |
2 | extern crate std;
  | ^^^^^^^^^^^^^^^^^ can't find crate
  |
  = note: the `x86_64-unknown-none` target may not support the standard library

error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> /home/stefano/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.12/src/lib.rs:296:9
    |
296 | /         compile_error!("target is not supported, for more information see: \
297 | |                         https://docs.rs/getrandom/#unsupported-targets");
    | |________________________________________________________________________^

   Compiling universal-hash v0.5.1
   Compiling aead v0.5.2
error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /home/stefano/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.12/src/lib.rs:347:9
    |
347 |         imp::getrandom_inner(dest)?;
    |         ^^^ use of undeclared crate or module `imp`

Some errors have detailed explanations: E0433, E0463.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `getrandom` (lib) due to 3 previous errors
warning: build failed, waiting for other jobs to finish...
make: *** [Makefile:118: bin/svsm-kernel.elf] Error 101
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -746,6 +746,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8edc89eaa583cf6bc4c6ef16a219f0a60d342ca3bf0eae793560038ac8af1795"
 
 [[package]]
+name = "rdrand"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d92195228612ac8eed47adbc2ed0f04e513a4ccb98175b6f2bd04d963b533655"
+dependencies = [
+ "rand_core",
+]
+
+[[package]]
 name = "rfc6979"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -835,6 +844,7 @@ dependencies = [
  "libmstpm",
  "log",
  "packit",
+ "rdrand",
  "syscall",
  "test",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ p384 = { version = "0.13.0" }
 uuid = "1.6.1"
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.7.32", features = ["derive"] }
+rdrand = { version = "0.8.3", default-features = false }
 
 # other repos
 packit = { git = "https://github.com/coconut-svsm/packit", version = "0.1.1" }
--- a/igvmbuilder/Cargo.toml
+++ b/igvmbuilder/Cargo.toml
@@ -3,7 +3,9 @@ name = "igvmbuilder"
 version = "0.1.0"
 edition = "2021"
 
-[dependencies]
+# specify target to avoid feature unification with SVSM
+# see https://doc.rust-lang.org/cargo/reference/features.html#feature-unification
+[target.'cfg(all(target_os = "linux"))'.dependencies]
 bootlib.workspace = true
 
 clap = { workspace = true, default-features = true, features = ["derive"] }
--- a/igvmmeasure/Cargo.toml
+++ b/igvmmeasure/Cargo.toml
@@ -3,7 +3,9 @@ name = "igvmmeasure"
 version = "0.1.0"
 edition = "2021"
 
-[dependencies]
+# specify target to avoid feature unification with SVSM
+# see https://doc.rust-lang.org/cargo/reference/features.html#feature-unification
+[target.'cfg(all(target_os = "linux"))'.dependencies]
 clap = { workspace = true, default-features = true, features = ["derive"] }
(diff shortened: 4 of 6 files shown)


stefano-garzarella:
@00xc to continue the discussion we were having in the call yesterday.

If I disable the default features of p384 in `igvmmeasure/Cargo.toml`, specifically use `alloc` instead of `std`, it seems to go (there's just the problem that the trait `std::error::Error` is not implemented), but in `igvmmeasure` we should leave the option of using `std` since it is a tool.
​
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 6af59e2..43d7c84 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ igvm = { version = "0.1.9", default-features = false}
 intrusive-collections = "0.9.6"
 libfuzzer-sys = "0.4"
 log = "0.4.17"
-p384 = { version = "0.13.0" }
+p384 = { version = "0.13.0", default-features = false }
 uuid = "1.6.1"
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.7.32", features = ["derive"] }
diff --git a/igvmmeasure/Cargo.toml b/igvmmeasure/Cargo.toml
index 5e2e236..c0900c8 100644
--- a/igvmmeasure/Cargo.toml
+++ b/igvmmeasure/Cargo.toml
@@ -8,7 +8,7 @@ clap = { workspace = true, default-features = true, features = ["derive"] }
 hmac-sha512.workspace = true
 igvm.workspace = true
 igvm_defs.workspace = true
-p384.workspace = true
+p384 = { workspace = true, features = ["arithmetic", "ecdh", "ecdsa", "pem", "alloc"] }
 zerocopy.workspace = true
 
 [lints]
```

Any idea what is wrong?

Could be a limitation of cargo workspace?
If it is the case, maybe we should have 2 different workspaces for svsm and tools

00xc:
> Could be a limitation of cargo workspace?

I was not aware, but apparently yes:
https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

> If it is the case, maybe we should have 2 different workspaces for svsm and tools

I agree

stefano-garzarella:
> > Could be a limitation of cargo workspace?
> 
> I was not aware, but apparently yes: https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

Thanks for the link.
Following it, I found this https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2

Maybe playing with `target` can help us, but I'm not sure if the toolchain can be used for that.

> 
> > If it is the case, maybe we should have 2 different workspaces for svsm and tools
> 
> I agree

Okay, I'll open an issue to track this and work on a solution.

stefano-garzarella:
@00xc using `[target.'cfg(all(target_os = "linux"))'.dependencies]` in the igvm tools seems to fix this issue. I'll open another PR for that commit, just pushed here for testing.

PR #348, opened 2024-05-17
Introduce safe page allocation abstractions

00xc:
* Introduce a `PageBox<T>` type. It works like `Box<T>`, releasing memory on drop, but it only has a subset of the functionality and uses memory obtained directly from the page allocator. This is useful for some types that need to reside on a page of their own.
* Introduce a `VmsaPage` abstraction based on the `PageBox` type.
* Introduce a `HVDoorbellPage` abstraction based on the `PageBox` type.
* Introduce a `GhcbPage` abstraction based on the `PageBox` type.
* Update the `PerCpu` code to use the `PageBox` type.
* Update the pagetable code to use the `PageBox` type.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -15,12 +15,11 @@ use crate::cpu::vmsa::{init_guest_vmsa, init_svsm_vmsa};
 use crate::cpu::LocalApic;
 use crate::error::SvsmError;
 use crate::locking::{LockGuard, RWLock, SpinLock};
-use crate::mm::alloc::allocate_zeroed_page;
 use crate::mm::pagetable::{get_init_pgtable_locked, PTEntryFlags, PageTableRef};
 use crate::mm::virtualrange::VirtualRange;
 use crate::mm::vm::{Mapping, VMKernelStack, VMPhysMem, VMRMapping, VMReserved, VMR};
 use crate::mm::{
-    virt_to_phys, SVSM_PERCPU_BASE, SVSM_PERCPU_CAA_BASE, SVSM_PERCPU_END,
+    virt_to_phys, PageBox, SVSM_PERCPU_BASE, SVSM_PERCPU_CAA_BASE, SVSM_PERCPU_END,
     SVSM_PERCPU_TEMP_BASE_2M, SVSM_PERCPU_TEMP_BASE_4K, SVSM_PERCPU_TEMP_END_2M,
     SVSM_PERCPU_TEMP_END_4K, SVSM_PERCPU_VMSA_BASE, SVSM_STACKS_INIT_TASK, SVSM_STACK_IST_DF_BASE,
 };
@@ -343,14 +342,10 @@ impl PerCpu {
     /// Creates a new default [`PerCpu`] struct, allocates it via the page
     /// allocator and adds it to the global per-cpu area list.
     pub fn alloc(apic_id: u32) -> Result<&'static Self, SvsmError> {
-        let vaddr = allocate_zeroed_page()?;
-        let percpu_ptr = vaddr.as_mut_ptr::<Self>();
-        unsafe {
-            (*percpu_ptr) = Self::new(apic_id);
-            let percpu = &*percpu_ptr;
-            PERCPU_AREAS.push(PerCpuInfo::new(apic_id, &percpu.shared));
-            Ok(percpu)
-        }
+        let page = PageBox::try_new(Self::new(apic_id))?;
+        let percpu = PageBox::leak(page);
+        unsafe { PERCPU_AREAS.push(PerCpuInfo::new(apic_id, &percpu.shared)) };
+        Ok(percpu)
     }
 
     pub fn shared(&self) -> &PerCpuShared {
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -15,7 +15,7 @@ use crate::cpu::vmsa::{init_guest_vmsa, init_svsm_vmsa};
 use crate::cpu::LocalApic;
 use crate::error::SvsmError;
 use crate::locking::{LockGuard, RWLock, SpinLock};
-use crate::mm::alloc::{allocate_zeroed_page, free_page};
+use crate::mm::alloc::allocate_zeroed_page;
 use crate::mm::pagetable::{get_init_pgtable_locked, PTEntryFlags, PageTableRef};
 use crate::mm::virtualrange::VirtualRange;
 use crate::mm::vm::{Mapping, VMKernelStack, VMPhysMem, VMRMapping, VMReserved, VMR};
@@ -26,7 +26,7 @@ use crate::mm::{
 };
 use crate::platform::{SvsmPlatform, SVSM_PLATFORM};
 use crate::sev::ghcb::{GhcbPage, GHCB};
-use crate::sev::hv_doorbell::HVDoorbell;
+use crate::sev::hv_doorbell::{HVDoorbell, HVDoorbellPage};
 use crate::sev::msr_protocol::{hypervisor_ghcb_features, GHCBHvFeatures};
 use crate::sev::utils::RMPFlags;
 use crate::sev::vmsa::{VMSAControl, VmsaPage};
@@ -449,17 +449,8 @@ impl PerCpu {
     }
 
     fn setup_hv_doorbell(&self) -> Result<(), SvsmError> {
-        let vaddr = allocate_zeroed_page()?;
-        let ghcb = current_ghcb();
-        if let Err(e) = HVDoorbell::init(vaddr, ghcb) {
-            free_page(vaddr);
-            return Err(e);
-        }
-        // SAFETY: the page contents have been allocated on valid memory and
-        // initialized. The HVDoorbell type's alignment requirements are met
-        // by the fact that we allocated a whole page. Mutable references to
-        // the page are never created, so this cannot be mutably aliased.
-        let doorbell = unsafe { &*vaddr.as_mut_ptr::<HVDoorbell>() };
+        let page = HVDoorbellPage::new(current_ghcb())?;
+        let doorbell = HVDoorbellPage::leak(page);
         self.hv_doorbell
             .set(doorbell)
             .expect("Attempted to reinitialize the HV doorbell page");
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -11,7 +11,7 @@ use super::tss::{X86Tss, IST_DF};
 use crate::address::{Address, PhysAddr, VirtAddr};
 use crate::cpu::idt::common::INT_INJ_VECTOR;
 use crate::cpu::tss::TSS_LIMIT;
-use crate::cpu::vmsa::{init_guest_vmsa, init_svsm_vmsa, vmsa_mut_ref_from_vaddr};
+use crate::cpu::vmsa::{init_guest_vmsa, init_svsm_vmsa};
 use crate::cpu::LocalApic;
 use crate::error::SvsmError;
 use crate::locking::{LockGuard, RWLock, SpinLock};
@@ -29,7 +29,7 @@ use crate::sev::ghcb::GHCB;
 use crate::sev::hv_doorbell::HVDoorbell;
 use crate::sev::msr_protocol::{hypervisor_ghcb_features, GHCBHvFeatures};
 use crate::sev::utils::RMPFlags;
-use crate::sev::vmsa::{allocate_new_vmsa, VMSAControl};
+use crate::sev::vmsa::{VMSAControl, VmsaPage};
 use crate::task::{schedule, schedule_task, RunQueue, Task, TaskPointer, WaitQueue};
 use crate::types::{PAGE_SHIFT, PAGE_SHIFT_2M, PAGE_SIZE, PAGE_SIZE_2M, SVSM_TR_FLAGS, SVSM_TSS};
 use crate::utils::MemoryRegion;
@@ -288,7 +288,7 @@ pub struct PerCpu {
 
     pgtbl: RefCell<PageTableRef>,
     tss: Cell<X86Tss>,
-    svsm_vmsa: OnceCell<&'static VMSA>,
+    svsm_vmsa: OnceCell<VmsaPage>,
     reset_ip: Cell<u64>,
     /// PerCpu Virtual Memory Range
     vm_range: VMR,
@@ -610,16 +610,11 @@ impl PerCpu {
             return Err(SvsmError::Mem);
         }
 
-        let vaddr = allocate_new_vmsa(RMPFlags::GUEST_VMPL)?;
-        let paddr = virt_to_phys(vaddr);
-
-        // SAFETY: we have exclusive access to this memory, as we just
-        // allocated it. allocate_new_vmsa() takes care of allocating
-        // memory of the right size and alignment for a VMSA.
-        let vmsa = unsafe { &mut *vaddr.as_mut_ptr::<VMSA>() };
+        let mut vmsa = VmsaPage::new(RMPFlags::GUEST_VMPL)?;
+        let paddr = vmsa.paddr();
 
         // Initialize VMSA
-        init_svsm_vmsa(vmsa, vtom);
+        init_svsm_vmsa(&mut vmsa, vtom);
         vmsa.tr = self.vmsa_tr_segment();
         vmsa.rip = start_rip;
         vmsa.rsp = self.get_top_of_stack().into();
@@ -664,13 +659,13 @@ impl PerCpu {
             ghcb.configure_interrupt_injection(INT_INJ_VECTOR)?;
         }
 
-        let vaddr = allocate_new_vmsa(RMPFlags::GUEST_VMPL)?;
-        let paddr = virt_to_phys(vaddr);
+        let mut vmsa = VmsaPage::new(RMPFlags::GUEST_VMPL)?;
+        let paddr = vmsa.paddr();
 
-        let vmsa = vmsa_mut_ref_from_vaddr(vaddr);
-        init_guest_vmsa(vmsa, self.reset_ip.get(), use_alternate_injection);
+        init_guest_vmsa(&mut vmsa, self.reset_ip.get(), use_alternate_injection);
 
         self.shared().update_guest_vmsa(paddr);
+        let _ = VmsaPage::leak(vmsa);
 
         Ok(())
     }
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -25,7 +25,7 @@ use crate::mm::{
     SVSM_PERCPU_TEMP_END_4K, SVSM_PERCPU_VMSA_BASE, SVSM_STACKS_INIT_TASK, SVSM_STACK_IST_DF_BASE,
 };
 use crate::platform::{SvsmPlatform, SVSM_PLATFORM};
-use crate::sev::ghcb::GHCB;
+use crate::sev::ghcb::{GhcbPage, GHCB};
 use crate::sev::hv_doorbell::HVDoorbell;
 use crate::sev::msr_protocol::{hypervisor_ghcb_features, GHCBHvFeatures};
 use crate::sev::utils::RMPFlags;
@@ -304,7 +304,7 @@ pub struct PerCpu {
     apic: RefCell<Option<LocalApic>>,
 
     /// GHCB page for this CPU.
-    ghcb: Cell<Option<&'static GHCB>>,
+    ghcb: OnceCell<GhcbPage>,
 
     /// `#HV` doorbell page for this CPU.
     hv_doorbell: OnceCell<&'static HVDoorbell>,
@@ -332,7 +332,7 @@ impl PerCpu {
             apic: RefCell::new(None),
 
             shared: PerCpuShared::new(apic_id),
-            ghcb: Cell::new(None),
+            ghcb: OnceCell::new(),
             hv_doorbell: OnceCell::new(),
             init_stack: Cell::new(None),
             ist: IstStacks::new(),
@@ -359,17 +359,14 @@ impl PerCpu {
 
     /// Sets up the CPU-local GHCB page.
     pub fn setup_ghcb(&self) -> Result<(), SvsmError> {
-        let ghcb_page = allocate_zeroed_page()?;
-        if let Err(e) = GHCB::init(ghcb_page) {
-            free_page(ghcb_page);
-            return Err(e);
-        };
-        let ghcb = unsafe { &*ghcb_page.as_ptr() };
-        self.ghcb.set(Some(ghcb));
+        let page = GhcbPage::new()?;
+        self.ghcb
+            .set(page)
+            .expect("Attempted to reinitialize the GHCB");
         Ok(())
     }
 
-    fn ghcb(&self) -> Option<&'static GHCB> {
+    fn ghcb(&self) -> Option<&GhcbPage> {
         self.ghcb.get()
     }
 
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -21,9 +21,11 @@ use crate::sev::utils::raw_vmgexit;
 use crate::types::{Bytes, PageSize, GUEST_VMPL, PAGE_SIZE_2M};
 use crate::utils::MemoryRegion;
 
+use crate::mm::PageBox;
 use core::arch::global_asm;
 use core::cell::Cell;
 use core::mem::{self, offset_of};
+use core::ops::{Deref, DerefMut};
 use core::ptr;
 
 use super::msr_protocol::{invalidate_page_msr, register_ghcb_gpa_msr, validate_page_msr};
@@ -126,6 +128,56 @@ impl TryFrom<Bytes> for GHCBIOSize {
     }
 }
 
+#[derive(Debug)]
+pub struct GhcbPage(PageBox<GHCB>);
+
+impl GhcbPage {
+    pub fn new() -> Result<Self, SvsmError> {
+        let page = PageBox::try_new_zeroed()?;
+        let vaddr = page.vaddr();
+        let paddr = virt_to_phys(vaddr);
+
+        if sev_snp_enabled() {
+            // Make page invalid
+            pvalidate(vaddr, PageSize::Regular, PvalidateOp::Invalid)?;
+
+            // Let the Hypervisor take the page back
+            invalidate_page_msr(paddr)?;
+
+            // Needs guarding for Stage2 GHCB
+            if valid_bitmap_valid_addr(paddr) {
+                valid_bitmap_clear_valid_4k(paddr);
+            }
+        }
+
+        // Map page unencrypted
+        get_init_pgtable_locked().set_shared_4k(vaddr)?;
+        flush_tlb_global_sync();
+
+        // SAFETY: all zeros is a valid representation for the GHCB.
+        unsafe { Ok(Self(page.assume_init())) }
+    }
+}
+
+impl Drop for GhcbPage {
+    fn drop(&mut self) {
+        self.0.shutdown().expect("Could not shut down GHCB");
+    }
+}
+
+impl Deref for GhcbPage {
+    type Target = GHCB;
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
+    }
+}
+
+impl DerefMut for GhcbPage {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.0.deref_mut()
+    }
+}
+
 #[repr(C)]
 #[derive(Debug, Clone)]
 pub struct GHCB {
@@ -203,30 +255,6 @@ impl GHCB {
 
     ghcb_getter!(get_usage_valid, usage, u32);
     ghcb_setter!(set_usage_valid, usage, u32);
-
-    pub fn init(vaddr: VirtAddr) -> Result<(), SvsmError> {
-        let paddr = virt_to_phys(vaddr);
-
-        if sev_snp_enabled() {
-            // Make page invalid
-            pvalidate(vaddr, PageSize::Regular, PvalidateOp::Invalid)?;
-
-            // Let the Hypervisor take the page back
-            invalidate_page_msr(paddr)?;
-
-            // Needs guarding for Stage2 GHCB
-            if valid_bitmap_valid_addr(paddr) {
-                valid_bitmap_clear_valid_4k(paddr);
-            }
-        }
-
-        // Map page unencrypted
-        get_init_pgtable_locked().set_shared_4k(vaddr)?;
-
-        flush_tlb_global_sync();
-
-        Ok(())
-    }
 
     pub fn rdtscp_regs(&self, regs: &mut X86GeneralRegs) -> Result<(), SvsmError> {
         self.clear();
--- a/kernel/src/sev/hv_doorbell.rs
+++ b/kernel/src/sev/hv_doorbell.rs
@@ -1,15 +1,16 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0 Copyright (c) Microsoft Corporation
 // Author: Jon Lange (jlange@microsoft.com)
 
-use crate::address::VirtAddr;
 use crate::cpu::idt::svsm::common_isr_handler;
 use crate::cpu::percpu::this_cpu;
 use crate::error::SvsmError;
 use crate::mm::page_visibility::{make_page_private, make_page_shared};
-use crate::mm::virt_to_phys;
+use crate::mm::{virt_to_phys, PageBox};
 use crate::sev::ghcb::GHCB;
 
 use bitfield_struct::bitfield;
+use core::mem::ManuallyDrop;
+use core::ops::Deref;
 use core::sync::atomic::{AtomicU32, AtomicU8, Ordering};
 
 #[bitfield(u8)]
@@ -46,6 +47,53 @@ pub struct HVExtIntInfo {
     pub isr: [AtomicU32; 8],
 }
 
+/// An allocation containing the `#HV` doorbell page.
+#[derive(Debug)]
+pub struct HVDoorbellPage(PageBox<HVDoorbell>);
+
+impl HVDoorbellPage {
+    /// Allocates a new HV doorbell page and registers it on the hypervisor
+    /// using the given GHCB.
+    pub fn new(ghcb: &GHCB) -> Result<Self, SvsmError> {
+        // SAFETY: all zeroes is a valid representation for `HVDoorbell`.
+        let page = PageBox::try_new_zeroed()?;
+        let paddr = virt_to_phys(page.vaddr());
+
+        // The #HV doorbell page must be shared before it can be used.
+        make_page_shared(page.vaddr())?;
+
+        // Now Drop will have correct behavior, so construct the new type.
+        // SAFETY: all zeros is a valid representation of the HV doorbell page.
+        let page = unsafe { Self(page.assume_init()) };
+        ghcb.register_hv_doorbell(paddr)?;
+        Ok(page)
+    }
+
+    /// Leaks the page allocation, ensuring it will never be freed.
+    pub fn leak(self) -> &'static HVDoorbell {
+        let mut doorbell = ManuallyDrop::new(self);
+        let ptr = core::ptr::from_mut(&mut doorbell);
+        // SAFETY: this pointer will never be freed because of ManuallyDrop,
+        // so we can create a static mutable reference. We go through a raw
+        // pointer to promote the lifetime to static.
+        unsafe { &mut *ptr }
+    }
+}
+
+impl Deref for HVDoorbellPage {
+    type Target = HVDoorbell;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
+    }
+}
+
+impl Drop for HVDoorbellPage {
+    fn drop(&mut self) {
+        make_page_private(self.0.vaddr()).expect("Failed to restore HV doorbell page visibility");
+    }
+}
+
 #[repr(C)]
 #[derive(Debug)]
 pub struct HVDoorbell {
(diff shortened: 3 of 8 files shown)


msft-jlange:
Is your change compatible with gdbstub?  The gdb support requires borrowing the GHCB in order to step in the debugger, but the debugger may be stepping through code that already has the GHCB borrowed, which would cause debugger actions to panic.  I'm also concerned about the possibility that somebody tries to add logging to code that has an active GHCB borrow, which would cause a similar panic because the console code will attempt to use the GHCB in order to communicate with the console.  While this is safe, it is not exactly desirable to make logging impossible in certain code paths.

I started (but did not complete) a nested GHCB borrow design in which GHCB borrowing had a "level" concept, in which each borrow of the GHCB declared an intended nesting level.  Three levels were defined: normal, console, and debugger.  The premise was that recursive borrows were illegal at a level that was the same or lower than any current borrow, but a recursive borrow was legal at a higher level.  The reference tracker (equivalent of `Ref` or `RefMut`) would contain a separate copy of the GHCB, populated at the time of the recursive borrow, and copied back to the GHCB page at the time the nested borrow is dropped.  This creates a safe recursive borrow for the console and debugger cases described above.

As I said, I never finished that work because it seemed like a lower priority than other core logic, but I could pick it up and try to finish it if necessary.  But I would be really concerned about any intermediate change that breaks debugger/console functionality in the pursuit of borrow safety.

00xc:
> Is your change compatible with gdbstub? The gdb support requires borrowing the GHCB in order to step in the debugger, but the debugger may be stepping through code that already has the GHCB borrowed, which would cause debugger actions to panic. I'm also concerned about the possibility that somebody tries to add logging to code that has an active GHCB borrow, which would cause a similar panic because the console code will attempt to use the GHCB in order to communicate with the console. While this is safe, it is not exactly desirable to make logging impossible in certain code paths.

I have not tested this, but it looks like it would interfere, yes.

> I started (but did not complete) a nested GHCB borrow design in which GHCB borrowing had a "level" concept, in which each borrow of the GHCB declared an intended nesting level. Three levels were defined: normal, console, and debugger. The premise was that recursive borrows were illegal at a level that was the same or lower than any current borrow, but a recursive borrow was legal at a higher level. The reference tracker (equivalent of `Ref` or `RefMut`) would contain a separate copy of the GHCB, populated at the time of the recursive borrow, and copied back to the GHCB page at the time the nested borrow is dropped. This creates a safe recursive borrow for the console and debugger cases described above.

I'm not sure I understand how this would work. If the contents are only copied on drop, then what will we pass to the GHCB method that performs the `VMGEXIT`? All of these methods (e.g. `GHCB::wrmsr()`) take a `&mut self`, but that cannot be done if the reference is dropped. Hope I'm not completely missing the point.

I would have assumed that the new `RefMut`-like type would need to register its GHCB copy via the GHCB MSR on acquisition. Then the VMGEXIT would work on the new copy, and the MSR could be restored on drop (avoiding the need to copy the GHCB page contents back).

> As I said, I never finished that work because it seemed like a lower priority than other core logic, but I could pick it up and try to finish it if necessary. But I would be really concerned about any intermediate change that breaks debugger/console functionality in the pursuit of borrow safety.

I can try to implement something like that if you want.

00xc:
It just dawned on me that you perhaps meant that the new `RefMut`-like type would still operate on the original GHCB page, but it would use its copy as a snapshot. I think that could work, but if I understand correctly, this opens a window for the following:

```rust
ref1 = ghcb_mut(Level::Normal);    // Takes snapshot 0
ref2 = ghcb_mut(Level::Debugger);  // Takes snapshot 1

ref1.some_field = some_value;      // Some modification to the GHCB

drop(ref2);                        // Restores to snapshot 1

ref1.some_call()                   // GHCB has unexpected state, ref1 modification is lost
```

I'm not claiming this is necessarily a very common pattern, but one that may happen, making the abstraction somewhat unsafe.

00xc:
> It just dawned on me that you perhaps meant that the new `RefMut`-like type would still operate on the original GHCB page, but it would use its copy as a snapshot. I think that could work, but if I understand correctly, this opens a window for the following
> ...

It would also have some usability issues. If the new `RefMut`-like type implements `DerefMut<GHCB>`, nothing is stopping us from doing the following, which is unsound:

```rust
let ghcb = ghcb_mut()?;
let some_ref: &mut GHCB = ghcb.deref_mut();
log::info() {
    // call some GHCB method that takes `&mut self`
    ghcb_mut()?.some_call();
}
some_ref.some_call()?;
```

On the other hand, if the new type does not implement `DerefMut` there is no straightforward way to call GHCB methods. We would have to do something like the following to scope the lifetime of the mutable reference:

```rust
let ghcb = ghcb_mut()?;
ghcb.with_mut(|ghcb| ghcb.some_call())?;
```

00xc:
For now I'm tempted to drop the last commit on this PR so this can be merged straight away.

msft-jlange:
> It just dawned on me that you perhaps meant that the new `RefMut`-like type would still operate on the original GHCB page, but it would use its copy as a snapshot. I think that could work, but if I understand correctly, this opens a window for the following:
> ...
> I'm not claiming this is necessarily a very common pattern, but one that may happen, making the abstraction somewhat unsafe.

I take your point, and I don't have any alternative suggestions.  I suspect this is a case where consumers of the GHCB will almost certainly do the right thing and we can accept the risk, because the pattern you are concerned about is one that looks like it is deliberately trying to do something very strange, and therefore is not one we should ever expect to see.  Yes, it's unfortunate if the compiler can't stop this pattern from occurring, but I think we can be OK with that.

msft-jlange:
> It would also have some usability issues. If the new `RefMut`-like type implements `DerefMut<GHCB>`, nothing is stopping us from doing the following, which is unsound:
> ```rust
> let ghcb = ghcb_mut(Level::Normal)?;
> let some_ref: &mut GHCB = ghcb.deref_mut();
> log::info() {
>     // call some GHCB method that takes `&mut self`
>     ghcb_mut(Level::Console)?.some_call();
> }
> some_ref.some_call()?;
> ```

But that's exactly the pattern we need to support: some outer block acquires a mutable reference to the GHCB and some inner block temporarily acquires a nested reference.  In your example, the inner block will drop the nested reference before the outer block attempts to reuse its reference, so the nesting should be accomplished successfully - meaning that the inner block will make a temporary copy in its local `RefMut`, then the drop will put that data back into the GHCB page, and the outer block will continue based on the data that it holds.  What is about this that is unsound?  It should certainly be the case that any contents of the GHCB page will be identical both before and after the inner block so the outer block won't be able to observe that any temporary copies were made.

msft-jlange:
> For now I'm tempted to drop the last commit on this PR so this can be merged straight away.

That's very reasonable, but now I want to raise a separate concern.  The HV doorbell page contains no mutable data, and thus no mutable reference can ever be acquired (we don't enforce that today, but we should).  If no mutable reference can ever be acquired, then there are no constraints on the number of shared references that can be acquired.  Since there is no constraint, it seems like we would not want to implement any dynamic logic to enforce borrow checking because it can never fail (by construction) and the borrow checking would be wasted effort.  Wouldn't it be better if we just introduced a function that returns `&'static HVDoorbell` and avoid the need for borrow checking?

00xc:
> I take your point, and I don't have any alternative suggestions. I suspect this is a case where consumers of the GHCB will almost certainly do the right thing and we can accept the risk, because the pattern you are concerned about is one that looks like it is deliberately trying to do something very strange

Not necessarily. This could happen by attaching and detaching with the debugger IIUC.

> and therefore is not one we should ever expect to see. Yes, it's unfortunate if the compiler can't stop this pattern from occurring, but I think we can be OK with that.

I think we can enforce this at the compiler level by making any subsequent reference's lifetime depend on the previous one (e.g. something like `let r2 = r1.borrow()`). I have implemented a small PoC of this but I have to test it properly, plus it's hard to add into the SVSM because the PerCpu cannot store the GHCB and something that references the GHCB (the `RefMut`-like type), since it makes the `PerCpu` self-referential (I think). Perhaps this can be done with the `Pin` type, but I haven't gotten around to trying it.

> > It would also have some usability issues. If the new `RefMut`-like type implements `DerefMut<GHCB>`, nothing is stopping us from doing the following, which is unsound:
> > ```rust
> > let ghcb = ghcb_mut(Level::Normal)?;
> > let some_ref: &mut GHCB = ghcb.deref_mut();
> > log::info() {
> >     // call some GHCB method that takes `&mut self`
> >     ghcb_mut(Level::Console)?.some_call();
> > }
> > some_ref.some_call()?;
> > ```
> 
> But that's exactly the pattern we need to support: some outer block acquires a mutable reference to the GHCB and some inner block temporarily acquires a nested reference. In your example, the inner block will drop the nested reference before the outer block attempts to reuse its reference, so the nesting should be accomplished successfully - meaning that the inner block will make a temporary copy in its local `RefMut`, then the drop will put that data back into the GHCB page, and the outer block will continue based on the data that it holds. What is about this that is unsound? It should certainly be the case that any contents of the GHCB page will be identical both before and after the inner block so the outer block won't be able to observe that any temporary copies were made.

It is still technically undefined behavior as far as I understand. I don't think in practice it would cause much of a problem in the current state of the compiler, but then again I'm not a compiler expert and it's still not allowed. The following snippet won't build for example:

```rust
fn main() {
    let mut v = 5u32;

    let r1 = &mut v;
    *r1 += 1;
    {
        let r2 = &mut v;
    }
    println!("{}", *r1);
}
```

```
error[E0499]: cannot borrow `v` as mutable more than once at a time
 --> src/main.rs:7:18
  |
4 |     let r1 = &mut v;
  |              ------ first mutable borrow occurs here
...
7 |         let r2 = &mut v;
  |                  ^^^^^^ second mutable borrow occurs here
8 |     }
9 |     println!("{}", *r1);
  |                    --- first borrow later used here
```

If you try going around the compiler with something like an `UnsafeCell` and run the program with Miri you'll see it spills an error as well.

> > For now I'm tempted to drop the last commit on this PR so this can be merged straight away.
> 
> That's very reasonable, but now I want to raise a separate concern. The HV doorbell page contains no mutable data, and thus no mutable reference can ever be acquired (we don't enforce that today, but we should). If no mutable reference can ever be acquired, then there are no constraints on the number of shared references that can be acquired. Since there is no constraint, it seems like we would not want to implement any dynamic logic to enforce borrow checking because it can never fail (by construction) and the borrow checking would be wasted effort. Wouldn't it be better if we just introduced a function that returns `&'static HVDoorbell` and avoid the need for borrow checking?

Yes, that makes sense, I will change `hv_doorbell_unsafe()` so that it returns an immutable reference and I'll update the function name.

kernel/src/cpu/percpu.rs
```diff
@@ -309,25 +283,26 @@ impl PerCpuUnsafe {
...
+    pub fn ghcb_unsafe(&mut self) -> *mut GHCB {
         self.ghcb
+            .as_mut()
+            .map(|g| g.as_mut_ptr())
+            .unwrap_or(ptr::null_mut())
     }
 
-    pub fn hv_doorbell_unsafe(&self) -> *mut HVDoorbell {
-        self.hv_doorbell
+    pub fn hv_doorbell(&self) -> Option<&HVDoorbell> {
```
msft-jlange, line 297: I have two concerns with this approach.

First, this is still a method on `PerCpuUnsafe`, which means that callers will still have to access this via unsafe code (since `this_cpu_unsafe()` returns a pointer).  I think it would be better to define a global function called `current_hv_doorbell()` or similar which hides all of the unsafe and returns a static reference.

Second, it changes the return type to an `Option` which means that all callers must be prepared to handle error cases.  The intention with the use of the doorbell page is that no caller should ever attempt to access the doorbell page until it has been allocated, which means it is preferable to put the unwrap and associated panic into the global function instead of requiring every caller to handle it separately.  I know that's not how it's implemented today but that's the path I've been on with APIC emulation.  It would be unfortunate to require every caller to perform Option checks on their own on a data structure that is expected to be always present.

00xc: > I have two concerns with this approach.
> 
> First, this is still a method on `PerCpuUnsafe`, which means that callers will still have to access this via unsafe code (since `this_cpu_unsafe()` returns a pointer). I think it would be better to define a global function called `current_hv_doorbell()` or similar which hides all of the unsafe and returns a static reference.

Okay, having a standalone `current_hv_doorbell()` sounds good.

> Second, it changes the return type to an `Option` which means that all callers must be prepared to handle error cases. The intention with the use of the doorbell page is that no caller should ever attempt to access the doorbell page until it has been allocated, which means it is preferable to put the unwrap and associated panic into the global function instead of requiring every caller to handle it separately. I know that's not how it's implemented today but that's the path I've been on with APIC emulation. It would be unfortunate to require every caller to perform Option checks on their own on a data structure that is expected to be always present.

How about a `Result` with something like `SvsmError::MissingHvDoorbellPage`? This follows the pattern of other error variants like `SvsmError::MissingSecrets` and `SvsmError::MissingVMSA`, and allows callers to handle the error (even if that  always means panicking).

msft-jlange: > How about a `Result` with something like `SvsmError::MissingHvDoorbellPage`? This follows the pattern of other error variants like `SvsmError::MissingSecrets` and `SvsmError::MissingVMSA`, and allows callers to handle the error (even if that always means panicking).

I don't see that as meaningfully different than the `Option` approach - both of them require the caller to do some sort of unwrap; it's just that one returns an `Err` and the other requires the caller to make a decision.  My point is that there will be no calls to the global accessor that will ever occur at a time that the doorbell page is missing, so there is no case where a caller wants to gracefully handle an error.  If a panic will occur in 100% of the cases, then why not do it in a single function?

BTW, I already have a PR ready to submit that includes a global `current_hv_doorbell()` function because I need it to fix a shortcoming in my last #HV doorbell PR.  Do you want me to submit it and we can refine that one together, or do you need me to wait until yours completes before I can post mine?

00xc: >
> I don't see that as meaningfully different than the Option approach - both of them require the caller to do some sort of unwrap; it's just that one returns an Err and the other requires the caller to make a decision. My point is that there will be no calls to the global accessor that will ever occur at a time that the doorbell page is missing, so there is no case where a caller wants to gracefully handle an error. If a panic will occur in 100% of the cases, then why not do it in a single function?

The main difference is that `unwrap()`ing or `expect()`ing the error will display the error variant in the panic message, which is more informative. I also think that having an interface implicitly panic is less clean.

> BTW, I already have a PR ready to submit that includes a global current_hv_doorbell() function because I need it to fix a shortcoming in my last #HV doorbell PR. Do you want me to submit it and we can refine that one together, or do you need me to wait until yours completes before I can post mine?

Let's get yours merged first since it's smaller, and I'll rebase on top.

00xc:
> > > But that's exactly the pattern we need to support: some outer block acquires a mutable reference to the GHCB and some inner block temporarily acquires a nested reference. In your example, the inner block will drop the nested reference before the outer block attempts to reuse its reference, so the nesting should be accomplished successfully - meaning that the inner block will make a temporary copy in its local `RefMut`, then the drop will put that data back into the GHCB page, and the outer block will continue based on the data that it holds. What is about this that is unsound? It should certainly be the case that any contents of the GHCB page will be identical both before and after the inner block so the outer block won't be able to observe that any temporary copies were made.
> > 
> > 
> > It is still technically undefined behavior as far as I understand. I don't think in practice it would cause much of a problem in the current state of the compiler, but then again I'm not a compiler expert and it's still not allowed. The following snippet won't build for example:
> 
> Thinking through this further, I cannot prove that the compiler cannot introduce badly timed spills which upset the nesting design. This makes me wonder whether a better approach would be to reimplement the `GHCB` object so that all attempts to operate on it use shared references, and the reference to the underlying data is managed internally as an unsafe mutable pointer.

This could also be done by replacing GHCB fields with `Cell`s, since they all implement `Copy` (I think) and `Cell<T>` is guaranteed to have the same layout as `T`.

This would eliminate any possibility that a `GHCB` might be borrowed while any mutable reference is outstanding. It would mean that the `GHCB` implementation would have to track its own nesting state (which likely would require that all calls specify a nesting level to ensure that the requested operation is safe to attempt). Internal state could be stored as atomics for ergonomics, and access to the mutable pointers could be gated on whether the atomic accesses that push/pop nesting state can permit the operation to proceed. This seems to me like it would require fewer contortions.

Yes, or `Cell` could be used too instead of atomics. It's what I used in my PoC.

kernel/src/cpu/percpu.rs
00xc: I've left this essentially unchanged, leaving the panic in the global accessor you introduced. We can discuss what to do here in a separate PR, especially regarding the static string issue you mentioned during yesterday's SVSM call.

00xc:
We continued the nested reference discussion via email. I think all review comments are addressed, so this should be good to go.

00xc:
This seems to fail booting in debug mode, I'll take a look. Please don't merge yet.

00xc:
So this is very messy. As-is, the SVSM works with this PR. The thing is, once I introduce the nested protection the SVSM panics early. This is because the new `NestedRef` type keeps a reference on the PerCpu, which itself is also reentrant and uses a `RefCell`, which causes it to panic when a mutable borrow is attempted. So this will require a greater rework.

kernel/src/sev/vmsa.rs
```diff
@@ -5,45 +5,85 @@
...
+
+impl VmsaPage {
+    /// Allocates a new VMSA for the given VPML.
+    pub fn new(vmpl: RMPFlags) -> Result<Self, SvsmError> {
+        assert!(vmpl.bits() < (VMPL_MAX as u64));
+
+        let mut page = PageBox::try_new_zeroed()?;
+        // Make sure the VMSA page is not 2M aligned. Some hardware generations
+        // can't handle this properly.
+        while page.vaddr().is_aligned(PAGE_SIZE_2M) {
```
msft-jlange, line 33: This can result in an infinite loop as the allocator could keep flip-flopping between the same two 2M-aligned pages.  The original code would allocate two contiguous pages in the case that the first allocation was 2M-aligned, because allocating two contiguous pages is guaranteed to produce at least one that is not 2M-aligned.  The old code would also have the unfortunate behavior that when allocating a pair of pages, one of them would be leaked, which is not strictly necessary as I believe the page allocator can free partial allocations.

We should make sure that the new code has compatible behavior.  Perhaps we need a flag to `PageBox::try_new()` (and its callers) to indicate that the allocation must not be 2M-aligned, and then `PageBox::try_new()` can internally do the dance required to produce (and not leak) an unaligned page.  Alternatively, a distinct function like `PageBox::try_new_unaligned()` could be implemented, and since its use is so rare, we could require callers to take responsibility for initializing the contents rather than having to implement a whole suite of unaligned functions.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -46,6 +47,59 @@ pub struct HVExtIntInfo {
...
+    /// using the given GHCB.
+    pub fn new(ghcb: &GHCB) -> Result<Self, SvsmError> {
+        // SAFETY: all zeroes is a valid representation for `HVDoorbell`.
+        let page = PageBox::try_new_zeroed()?;
+        let paddr = virt_to_phys(page.vaddr());
+
+        // The #HV doorbell page must be shared before it can be used.
+        make_page_shared(page.vaddr())?;
+
+        // Now Drop will have correct behavior, so construct the new type.
```
msft-jlange, line 65: `Drop` will not have the correct behavior once the page has been made shared, because the implementation of `PageBox::drop()` is not aware that the page has become shared and thus cannot make the page private before it is returned to the free pool.  This will cause fatal errors if the page is attempted to be used again in a private context.

I think we're going to need to introduce `PageBox::make_page_shared(&self)` and a matching function to make the page private, and to have the `PageBox` structure hold a flag indicating whether the page was made shared so that it can be made private again if necessary before being freed.

kernel/src/sev/ghcb.rs
```diff
@@ -126,6 +128,56 @@ impl TryFrom<Bytes> for GHCBIOSize {
...
+
+impl GhcbPage {
+    pub fn new() -> Result<Self, SvsmError> {
+        let page = PageBox::try_new_zeroed()?;
+        let vaddr = page.vaddr();
+        let paddr = virt_to_phys(vaddr);
+
+        if sev_snp_enabled() {
+            // Make page invalid
+            pvalidate(vaddr, PageSize::Regular, PvalidateOp::Invalid)?;
```
msft-jlange, line 142: This should use `make_page_shared()` in the same way that the HV doorbell page makes the page shared.  The old code was different, but this is our opportunity to align the two implementations - especially since we will want to have the same shared/private tracking for the GHCB page that we do for the HV doorbell page.

tlendacky: Is this running on the vCPU that the GHCB page is going to be used on? If so, then make_page_shared() will attempt to use the very same page that is trying to be made shared. I imagine that is the reason for the MSR protocol usage just below this.

msft-jlange: That's a good point, but if we design `PageBox` to track the shared/private state of the page as I have suggested, then we really want to avoid having two different ways for the state to change.  I would suggest changing `make_page_shared()` to use the MSR protocol instead of the GHCB protocol, since it solves the concern you raise, it has no important performance or behavior difference compared to the GHCB protocol, and it permits the unification of code paths as I am recommending.

kernel/src/sev/hv_doorbell.rs
00xc: Not sure what you mean. If `make_page_shared` fails `drop()` will not be called because the type has not been constructed. Once `Self` is constructed a few lines below, any failure *will* cause `drop()` to run. So the behavior is expected. Also, the `drop()` ordering makes the implementation safe, as it will first run for `HVDoorbellPage` (marking the page as private), and then for `PageBox` (returning the page to the allocator). Rust guarantees this ordering.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -46,6 +47,59 @@ pub struct HVExtIntInfo {
...
+
+impl Deref for HVDoorbellPage {
+    type Target = HVDoorbell;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
+    }
+}
+
+impl DerefMut for HVDoorbellPage {
```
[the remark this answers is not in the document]
00xc: Yep, I agree, I'll remove this.

kernel/src/sev/ghcb.rs
00xc: I was also thinking of introducing a new abstraction for page state on top of `PageBox`, in order to keep the type as simple as possible (i.e. only concerned with interacting with the page allocator and nothing more).

In any case, for now I think we should probably add a new variant for `make_page_shared()` that uses the MSR protocol, so the caller is able to choose, and use this new variant in the `GhcbPage` code. What do you think?

kernel/src/sev/vmsa.rs
00xc: Freeing the second page in the order-1 allocation will also free the first one as far as I can tell. For example, running the following test:

```diff
diff --git a/kernel/src/mm/alloc.rs b/kernel/src/mm/alloc.rs
index 4a0fbb96..7404d7c2 100644
--- a/kernel/src/mm/alloc.rs
+++ b/kernel/src/mm/alloc.rs
@@ -1972,6 +1972,27 @@ fn test_slab_page_slab_for_self() {
     assert!(SLAB_PAGE_SLAB.lock().common.free_pages < 2);
 }
 
+#[test]
+fn test_partial_free() {
+    extern crate std;
+    let _test_mem = TestRootMem::setup(DEFAULT_TEST_MEMORY_SIZE);
+    let mut root_mem = ROOT_MEM.lock();
+
+    let p1 = root_mem.allocate_pages(1).unwrap();
+    let pfn1 = root_mem.get_pfn(p1).unwrap();
+
+    let p2 = p1 + PAGE_SIZE;
+    let pfn2 = root_mem.get_pfn(p2).unwrap();
+
+    root_mem.free_page(p2);
+
+    let info1 = root_mem.read_page_info(pfn1);
+    std::println!("info1 = {info1:?}");
+
+    let info2 = root_mem.read_page_info(pfn2);
+    std::println!("info2 = {info2:?}");
+}
+
 #[test]
 #[cfg_attr(test_in_svsm, ignore = "FIXME")]
 /// Allocate enough objects to hit an OOM situation and verify null gets
```

Will produce the following output:

```
info1 = Free(FreeInfo { next_page: 0, order: 1 })
info2 = Compound(CompoundInfo { order: 1 })
```

So I think we can implement `PageBox::try_new_unaligned()` using that trick, and the `Drop` handler should just work fine. Does that sound right?

kernel/src/sev/hv_doorbell.rs
msft-jlange: I'm saying something different: if `make_page_shared()` succeeds, then any subsequent `drop()` will free the page without restoring it to the private state, causing issues later.  Once a page has successfully become shared, then we need a guarantee that it is made private again before it can be returned to the free list, and it seems like this is something that must be tracked so `drop()` can accomplish that.

00xc: Again, no, because once `make_page_shared()` succeeds,  we next create the actual `HvDoorbellPage` right below, whose `drop()` will restore page visibility before the page is freed.

kernel/src/sev/ghcb.rs
msft-jlange: > I wouldn't recommend using the MSR protocol once a GHCB is established unless you save and restore the MSR value whenever a VMGEXIT is issued. For example, you issue an MSR protocol request and exit to the hypervisor. Upon return from the hypervisor, an NMI is injected. If the NMI handling was to perform anything that requires a VMGEXIT using the GHCB, you will now lose the MSR protocol return value because the GHCB support is writing the GHCB address into the GHCB MSR as part of issuing its VMGEXIT.

The existing GHCB logic will set the GHCB MSR every time it issues VMGEXIT.  Note that we also use the MSR form of VMPL switch today because that is necessary in order to make #HV interact correctly with VMPL switch flows.  That obligates us to keep the MSR management correct.

I do not anticipate needing NMI handling.  As long as the SVSM uses restricted injection, NMI injection is impossible (and we don't handle it anyway because we don't expect any NMI sources).  Once the SVSM uses secure AVIC, we can set AllowNmi=0 and NMI injection will remain impossible.  And if the SVSM uses legacy (unrestricted injection), all bets are off anyway.

kernel/src/sev/vmsa.rs
msft-jlange: I had forgotten that the allocator requires the entire allocation to be freed at once.  In a perfect world, we'd change the allocator to support partial free operations.  If the "trick" you're referring to is to allocate a pair of pages and to have the reference/pointer/vaddr accessors calculate which half is being used, then I agree that we can implement `try_new_unaligned()` this way.

kernel/src/sev/hv_doorbell.rs
msft-jlange: The correctness of what you describe is contingent on `HVDoorbellPage` implementing `Drop` correctly.  The same is true of `GHCB`, which does not call `make_page_private()` in its `drop()` method despite making the page shared when it acquires it.  If the callers of `PageBox` do not implement `Drop` such that the private/shared management is performed correctly, then subsequent problems will be observed much later by other consumers of `PageBox`, which is not a great outcome.  As you've said to me many times before, true safety comes from a construction that does not depend on consumers of an object treating that object correctly, and I assert that true safety here would require that it be impossible to return any page to the free list if it is not in the expected state (i.e. private).

An ideal outcome would be to eliminate the `make_page_shared()` and `make_page_private()` functions altogether and make them methods on `PageBox` so that there is no opportunity for any consumer to change the state of a page such that it cannot be safely restored.  I do not believe there are any legitimate uses for making pages shared and private outside of allocations made through the page allocator.  We don't have to do that in this PR, but we really need to seriously consider it.

kernel/src/sev/ghcb.rs
tlendacky: > The existing GHCB logic will set the GHCB MSR every time it issues VMGEXIT. Note that we also use the MSR form of VMPL switch today because that is necessary in order to make #HV interact correctly with VMPL switch flows. That obligates us to keep the MSR management correct.

Why is it necessary to use the MSR protocol? There should be no issue with using the NAE event form / GHCB method to switch VMPLs. And once a GHCB is established, that is what should be used.

> 
> I do not anticipate needing NMI handling. As long as the SVSM uses restricted injection, NMI injection is impossible (and we don't handle it anyway because we don't expect any NMI sources). Once the SVSM uses secure AVIC, we can set AllowNmi=0 and NMI injection will remain impossible. And if the SVSM uses legacy (unrestricted injection), all bets are off anyway.

Even with Restricted Injection, a #HV can be injected, and any VMGEXIT within that path will cause issues.

kernel/src/sev/hv_doorbell.rs
msft-jlange: Putting the shared/private logic into a separate type would be a good step.  The goal should be to ensure that the actual consumers of the pages (GHCB, HV doorbell) have no responsibility for the correctness of state management because there are too many places for bugs to exist.  Having a single type outside of `PageBox` that takes this responsibility would work well.

kernel/src/sev/ghcb.rs
msft-jlange: > Why is it necessary to use the MSR protocol? There should be no issue with using the NAE event form / GHCB method to switch VMPLs. And once a GHCB is established, that is what should be used.

Receipt of #HV must cancel any VMPL switch or else pending interrupts will be lost.  In order for this to work, the flow between the point in time that the VMPL switch code checks for pending #HV events and the point in time that the VMGEXIT occurs must be straight-line, with no function calls - and must also necessarily be in assembly.  Performing memory-based exits is significantly more complex.

Note that the code that I'm describing is already checked in so you can review it.  Refer to `switch_to_vmpl_unsafe()` and review its interactions with `asm_entry_hv`.

> Even with Restricted Injection, a #HV can be injected, and any VMGEXIT within that path will cause issues.

The #HV handler specifically understands this case and avoids any VMGEXITs within that window to ensure that there is no collision.

kernel/src/sev/vmsa.rs
00xc: Not even calculate which half is used, but simply store the second page in the internal pointer during initialization and "forget" about this fact. The Drop handler will pass the second page to `free_page()`, but as seen above, this should work fine anyway.

kernel/src/sev/hv_doorbell.rs
00xc: If it is okay with you, let's mark this conversation as resolved and I'll open a follow-up PR with the changes you propose.

msft-jlange: Agreed.

kernel/src/sev/vmsa.rs
msft-jlange: That seems a little fragile to me in case we change the allocator in the future to permit partial freeing, or to panic if there is an attempt to free a pointer that was not directly handed out.  A robust allocator should always require free operations to pass the same pointer obtained by the allocation operation, and the fact that ours doesn't require that is a quirk that I don't think we want to take a dependency on.  It would be better if you could design `PageBox` to remember whether the offset occurred so that it can compensate for it at free time.  That should only require a `bool` so I would think that could be done cheaply.

00xc: That would effectively double the size of the type (which is supposed to be pointer-like) for a very specific case that will likely never occur again, as it is due to a hardware bug. What about this?

```rust
struct VmsaPage(PageBox<[VMSA; 2]>);

// impl VmsaPage { ... }

impl Deref for VmsaPage {
    type Target = VMSA;

    #[inline]
    fn deref(&self) -> &Self::Target {
        self.0[1].deref()
    }
}
```

kernel/src/sev/ghcb.rs
00xc: @tlendacky I take it you have no objections to using the MSR protocol unconditionally then? Otherwise we can discuss this change in a new PR, as it seems to fall out of scope of this one.

kernel/src/sev/vmsa.rs
00xc: > I agree with your concern about structure sizing. However, I'm not certain that `VMSA` is precisely sized to a single page

It fits within a single page as of the latest version of the AMD manual. There are several references to the "VMSA page" in the AMD manual as well.

> nor that `PageBox` can (or should) deal with a `<T>` that is larger than one page.

It can, it uses the size of `T` to request a page allocation of the adequate order to fit it.

> Given how unusual the alignment requirement is, and how unlikely it is that we will ever need to free an allocated VMSA, I wonder whether it would be better for `VMSA` not to be based on `PageBox` but to stick to raw allocations, but still to implement the same `vaddr()` and `deref()` methods for safety, and just not to implement `Drop` at all so that the underlying allocation is just leaked (since we're never realistically going to free a VMSA without shutting down).

The idea behind this PR is to eventually make the low-level calls private, so that only the safe abstraction is used outside the mm module, so this would be a sub-optimal solution in my opinion.

msft-jlange: > > I agree with your concern about structure sizing. However, I'm not certain that `VMSA` is precisely sized to a single page
> 
> It fits within a single page as of the latest version of the AMD manual. There are several references to the "VMSA page" in the AMD manual as well.

Sorry, what I meant was that I didn't think the SVSM definition of `VMSA` was declared to be exactly 0x1000 bytes - I thought our definition was somewhat less.  I'm concerned that the definition of `VMSA` might be small enough that `[VMSA; 2]` might also fit into a single page, which would defeat the attempt to use a duple to cause a two-page allocation.

> The idea behind this PR is to eventually make the low-level calls private, so that only the safe abstraction is used outside the mm module, so this would be a sub-optimal solution in my opinion.

I understand your point.  I'm also starting to question whether the alignment checks performed on the `VMSA` allocation are actually sufficient to avoid the hardware problem, since the use of 4 KB backing pages means that the GPA alignment checks prove nothing about physical alignment - and if that is true, then perhaps all of this alignment checking should just be removed.  So whatever we do, we should plan the design so that removal of alignment hacks like we're discussing here doesn't leave us with any regrets if/when we do remove such logic.

00xc: > Sorry, what I meant was that I didn't think the SVSM definition of `VMSA` was declared to be exactly 0x1000 bytes - I thought our definition was somewhat less. I'm concerned that the definition of `VMSA` might be small enough that `[VMSA; 2]` might also fit into a single page, which would defeat the attempt to use a duple to cause a two-page allocation.

Size is correct in our definition because we add the needed padding, so each VMSA would fall into a separate page:

https://github.com/coconut-svsm/svsm/blob/4eb69fc3acccfa04f1cf95a902a8091b7a8175a0/cpuarch/src/vmsa.rs#L225
 
> I understand your point. I'm also starting to question whether the alignment checks performed on the `VMSA` allocation are actually sufficient to avoid the hardware problem, since the use of 4 KB backing pages means that the GPA alignment checks prove nothing about physical alignment - and if that is true, then perhaps all of this alignment checking should just be removed. So whatever we do, we should plan the design so that removal of alignment hacks like we're discussing here doesn't leave us with any regrets if/when we do remove such logic.

I think the alignment issue happens on GVAs, since that fix was introduced due to a crash/malfunction observed at some point in the past, and it seems to have been fixed.

kernel/src/sev/ghcb.rs
00xc: For now I'm inclined to leave this as-is, and we can evaluate this change in a separate PR, as it seems to fall out of the scope for `PageBox`.

kernel/src/sev/vmsa.rs
msft-jlange: > I think the alignment issue happens on GVAs, since that fix was introduced due to a crash/malfunction observed at some point in the past, and it seems to have been fixed.

The alignment issue happens on system physical addresses.  The current code makes an assumption that GVA alignment matches SPA alignment, which is only guaranteed to be true if both the GVA/GPA mapping uses a 2 MB entry, and if the GPA/SPA mapping also uses a 2 MB entry.  Our code can only guarantee the former but the latter is up to the discretion of the host, which means that anything we attempt to guarantee alignment will not guarantee success.  The best fix would be for the host to detect when a VMSA is aligned to a 2 MB boundary (which it can do because the host is required to know all VMSA addresses), and to rearrange the GPA/SPA mappings to ensure that there is no alignment conflict.  Obviously any host change is out of scope for any PR, but if the host were to make such a change someday, then the alignment hacks in our code would be irrelevant.

I'm not trying to say that your alignment proposal is not a good one, only that it may need to be removed someday, so you should have confidence that your `PageBox` implementation will not require significant rework if it is removed.

Another thing to consider is that we have discussed removing the direct physical/virtual map for the heap, to be replaced with an on-demand virtual mapping.  When that happens, the allocation logic may need to change since allocation of multiple physical pages may produce free pages with identical alignment characteristics.  I cannot anticipate how the physical page allocator would work in such a design, but we need to anticipate the possibility and be comfortable with the idea that allocation of multiple pages through `PageBox` might no longer solve the GPA alignment problem at that time.

I raise these points not to suggest that any changes to your PR are needed, but so that you consider them before you decide whether you think the PR is good as it is.

kernel/src/sev/ghcb.rs
msft-jlange: We can make the change in a separate PR, but I stand behind the proposal of adding shared/private state changes as a method of `PageBox` for safety (meaning that the change is within the scope of `PageBox`, even if it is beyond what you want to accomplish in this PR).

kernel/src/sev/vmsa.rs
00xc: > > I think the alignment issue happens on GVAs, since that fix was introduced due to a crash/malfunction observed at some point in the past, and it seems to have been fixed.
> 
> The alignment issue happens on system physical addresses. The current code makes an assumption that GVA alignment matches SPA alignment, which is only guaranteed to be true if both the GVA/GPA mapping uses a 2 MB entry, and if the GPA/SPA mapping also uses a 2 MB entry. Our code can only guarantee the former but the latter is up to the discretion of the host, which means that anything we attempt to guarantee alignment will not guarantee success. The best fix would be for the host to detect when a VMSA is aligned to a 2 MB boundary (which it can do because the host is required to know all VMSA addresses), and to rearrange the GPA/SPA mappings to ensure that there is no alignment conflict. Obviously any host change is out of scope for any PR, but if the host were to make such a change someday, then the alignment hacks in our code would be irrelevant.
> 
> I'm not trying to say that your alignment proposal is not a good one, only that it may need to be removed someday, so you should have confidence that your `PageBox` implementation will not require significant rework if it is removed.

With my latest version it would just be a matter of removing the extra logic. I'm fine with coming back and fixing the issues where necessary, the purpose of this PR is not to have the definitive API but to have a **safe** abstraction that can be iteratively improved.  

> Another thing to consider is that we have discussed removing the direct physical/virtual map for the heap, to be replaced with an on-demand virtual mapping. When that happens, the allocation logic may need to change since allocation of multiple physical pages may produce free pages with identical alignment characteristics. I cannot anticipate how the physical page allocator would work in such a design, but we need to anticipate the possibility and be comfortable with the idea that allocation of multiple pages through `PageBox` might no longer solve the GPA alignment problem at that time.
> 
> I raise these points not to suggest that any changes to your PR are needed, but so that you consider them before you decide whether you think the PR is good as it is.

Yes, this also makes sense. We can adjust the type when this happens.

00xc:
Added some extra documentation in the latest version of the PR. I will merge this by the end of the day if nobody has any objections.

msft-jlange (approved):

PR #354, opened 2024-05-21
platform/snp: Perform lazy #HV doorbell EOI when able

msft-jlange:
The #HV doorbell protocol defines a "lazy" EOI model that permits a VM to indicate an intention to perform an EOI without requiring an immediate exit to the host.  This change modifies the SNP EOI flow to take advantage of lazy EOI when possible.

### DIFF
--- a/kernel/src/platform/snp.rs
+++ b/kernel/src/platform/snp.rs
@@ -11,6 +11,7 @@ use crate::cpu::percpu::PerCpu;
 use crate::error::SvsmError;
 use crate::io::IOPort;
 use crate::platform::{PageEncryptionMasks, PageStateChangeOp, SvsmPlatform};
+use crate::sev::hv_doorbell::current_hv_doorbell;
 use crate::sev::msr_protocol::verify_ghcb_version;
 use crate::sev::status::vtom_enabled;
 use crate::sev::{pvalidate_range, sev_status_init, sev_status_verify, PvalidateOp};
@@ -115,8 +116,12 @@ impl SvsmPlatform for SnpPlatform {
     }
 
     fn eoi(&self) {
-        // 0x80E is the X2APIC EOI MSR.
-        // Errors here cannot be handled but should not be grounds for panic.
-        let _ = current_ghcb().wrmsr(0x80E, 0);
+        // Issue an explicit EOI unless no explicit EOI is required.
+        if !current_hv_doorbell().no_eoi_required() {
+            // 0x80B is the X2APIC EOI MSR.
+            // Errors here cannot be handled but should not be grounds for
+            // panic.
+            let _ = current_ghcb().wrmsr(0x80B, 0);
+        }
     }
 }
--- a/kernel/src/sev/hv_doorbell.rs
+++ b/kernel/src/sev/hv_doorbell.rs
@@ -3,6 +3,7 @@
 
 use crate::address::VirtAddr;
 use crate::cpu::idt::svsm::common_isr_handler;
+use crate::cpu::percpu::this_cpu_unsafe;
 use crate::error::SvsmError;
 use crate::mm::page_visibility::{make_page_private, make_page_shared};
 use crate::mm::virt_to_phys;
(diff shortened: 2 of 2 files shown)


kernel/src/platform/snp.rs
```diff
@@ -115,8 +116,12 @@ impl SvsmPlatform for SnpPlatform {
...
     fn eoi(&self) {
-        // 0x80E is the X2APIC EOI MSR.
-        // Errors here cannot be handled but should not be grounds for panic.
-        let _ = current_ghcb().wrmsr(0x80E, 0);
+        // Issue an explicit EOI unless no explicit EOI is required.
+        if !current_hv_doorbell().no_eoi_required() {
+            // 0x80E is the X2APIC EOI MSR.
+            // Errors here cannot be handled but should not be grounds for
+            // panic.
+            let _ = current_ghcb().wrmsr(0x80E, 0);
```
vijaydhanraj, line 124: Currently, code base has different implementations of write MSR like `write_msr()` and `wrmsr()`. Would implementing `wrmsr` as a platform abstracted trait function be better?

kernel/src/platform/snp.rs
```diff
@@ -115,8 +116,12 @@ impl SvsmPlatform for SnpPlatform {
     }
 
     fn eoi(&self) {
-        // 0x80E is the X2APIC EOI MSR.
-        // Errors here cannot be handled but should not be grounds for panic.
-        let _ = current_ghcb().wrmsr(0x80E, 0);
+        // Issue an explicit EOI unless no explicit EOI is required.
+        if !current_hv_doorbell().no_eoi_required() {
+            // 0x80E is the X2APIC EOI MSR.
```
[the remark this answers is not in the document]
msft-jlange: Good catch!  This error existed in the previous version of the code but I should fix it here.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -75,6 +76,43 @@ impl HVDoorbell {
...
+                no_eoi_required & !1,
+                Ordering::Relaxed,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => break,
+                Err(new) => no_eoi_required = new,
+            }
+        }
+
+        // If the flag was successfully cleareed, then no explicit EOI is
```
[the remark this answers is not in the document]
msft-jlange: Fixed.

kernel/src/platform/snp.rs
msft-jlange: `write_msr()` is a wrapper for the `WRMSR` instruction.  Here, `wrmsr()` is a method on the GHCB, which performs an MSR exit to the host via the GHCB protocol.  Since this code is the implementation of the SNP-specific platform abstraction, it is appropriate to use SNP-specific methods here.  It would be wasteful for this code to execute an instruction that raises a #VC exception which has to be emulated when this platform-specific code could just perform the GHCB operation directly.

vijaydhanraj: Understand, thanks. I was thinking something like below so that each platform can implement its own abstraction. For example, SNP specific platforms could perform GHCB operation directly and TDP platforms could do `wrmsr` instruction handling. But this could be a different PR if needed.
```
<pseudo code>
+pub fn wrmsr(msr_index: u32, value: u64) {
+    let platform =  SVSM_PLATFORM.as_dyn_ref();
+    platform.msr_write(msr_index, value)
+}
+
+pub fn rdmsr(msr_index: u32) {
+    let platform =  SVSM_PLATFORM.as_dyn_ref();
+    platform.msr_read(msr_index)
+} 
```

msft-jlange: The current direction across the code base generally is that the ISA Is the abstraction, and we will rely on #VC/#VE dispatch to handle platform-specific logic.  When the call comes from platform-specific logic instead of common code, it can go directly to a platform-specific handler, as is done here.

We can debate whether this is the right direction, but that goes far beyond the scope of this PR and should be discussed as a separate issue.

vijaydhanraj: Thanks for the clarification.

PR #355, opened 2024-05-22
Interface proposal for the split TD partitioning VMM

cxdong:
This draft PR is opened to collect review comments and discussions, but not to merge.

The TD partitioning VMM was implemented fully in the coconut-svsm kernel mode, which runs the coconut-svsm as a L1 VMM in TD and a Linux OS as a L2 guest. The code is published[ here](https://github.com/intel-staging/td-partitioning-svsm/tree/svsm-tdp).

Following the development plan, most parts of this TD partitioning VMM should be moved to the user mode for security reasons, while keeping the necessary parts (e.g., executing TDCALL instruction) left in the svsm kernel. We made a document, available for preview [ here](https://github.com/cxdong/svsm/blob/interface-doc/Documentation/docs/developer/design/tdx/INTERFACES-OF-SPLIT-TDP-VMM.md), which discusses the high-level architecture of the split TD partitioning VMM and outlines the key VMM interfaces between user mode and kernel mode. In short, the proposal is to use the common file based syscalls ABI, e.g., open, ioctl, mmap as the VMM interfaces, and clone syscall to create multiple vCPU threads which shares the same user mode address space.

Before moving forward, we would like to use this PR to collect the review comments from the community about this proposal and possibility of extending it for SEV user mode VMM, as well as any alternative proposals that the community prefers. We would greatly appreciate it if you could take some time to review and share your thoughts.


cxdong:
The markdown doc is converted to a [ pdf file](https://github.com/cxdong/svsm/blob/interface-doc/Documentation/docs/developer/design/tdx/INTERFACES-OF-SPLIT-TDP-VMM.pdf), which is now uploaded to this PR.

cxdong:
Hi @joergroedel, this document is updated with describing a basic vcpu run loop implementation in the user mode VMM to enter guest and handle simple vmexit event, with utilizing the syscalls outlined in the [X86-64 Syscall API/ABI Document](https://mail.8bytes.org/pipermail/svsm-devel/2024-June/000334.html). And [here](https://github.com/cxdong/svsm/blob/interface-doc/Documentation/docs/developer/design/tdx/USER-MODE-VMM.pdf) is the pdf version. Would you please let us know if this is on the right direction or not? Thanks!

peterfang:
Closing as the design is now outdated. The community is actively working towards a cross-platform VMM design.

PR #356, opened 2024-05-22
igvm: initial support for TD Partitioning

peterfang:
This is the first set of patches to begin supporting TD Partitioning in the IGVM framework. Patches for TDX enabling will follow.
No changes to the IGVM spec are proposed yet. The goal is to allow TDP enabling to hit the ground running first.
The main work in this patchset is to separate TDP from SEV platforms.

To generate an IGVM file for a TDP platform as part of the build process, run:

`$ make PLATFORM=tdp FW_FILE=/path/to/OVMF.fd`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,15 +69,15 @@ $(IGVMMEASURE):
 	cargo build ${CARGO_ARGS} --target=x86_64-unknown-linux-gnu -p igvmmeasure
 
 bin/coconut-qemu.igvm: $(IGVMBUILDER) $(IGVMMEASURE) bin/svsm-kernel.elf bin/stage2.bin ${FS_BIN}
-	$(IGVMBUILDER) --sort --policy 0x30000 --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --filesystem ${FS_BIN} ${BUILD_FW} qemu
+	$(IGVMBUILDER) --sort --policy 0x30000 --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --filesystem ${FS_BIN} ${BUILD_FW} qemu --snp
 	$(IGVMMEASURE) --check-kvm $@ measure
 
 bin/coconut-hyperv.igvm: $(IGVMBUILDER)  $(IGVMMEASURE) bin/svsm-kernel.elf bin/stage2.bin
-	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --comport 3 hyper-v --native
+	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --comport 3 hyper-v --native --snp
 	$(IGVMMEASURE) $@ measure
 
 bin/coconut-test-qemu.igvm: $(IGVMBUILDER)  $(IGVMMEASURE) bin/test-kernel.elf bin/stage2.bin
-	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/test-kernel.elf qemu
+	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/test-kernel.elf qemu --snp
 	$(IGVMMEASURE) $@ measure
 
 test:
--- a/Makefile
+++ b/Makefile
@@ -69,16 +69,16 @@ $(IGVMBUILDER):
 $(IGVMMEASURE):
 	cargo build ${CARGO_ARGS} --target=x86_64-unknown-linux-gnu -p igvmmeasure
 
-bin/coconut-qemu.igvm: $(IGVMBUILDER) $(IGVMMEASURE) bin/svsm-kernel.elf bin/stage2.bin ${FS_BIN}
-	$(IGVMBUILDER) --sort --policy 0x30000 --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --filesystem ${FS_BIN} ${BUILD_FW} qemu --snp
+bin/coconut-qemu.igvm: $(IGVMBUILDER) $(IGVMMEASURE) bin/stage1-trampoline.bin bin/svsm-kernel.elf bin/stage2.bin ${FS_BIN}
+	$(IGVMBUILDER) --sort --policy 0x30000 --output $@ --tdx-stage1 bin/stage1-trampoline.bin --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --filesystem ${FS_BIN} ${BUILD_FW} qemu --snp --tdp
 	$(IGVMMEASURE) --check-kvm $@ measure
 
-bin/coconut-hyperv.igvm: $(IGVMBUILDER)  $(IGVMMEASURE) bin/svsm-kernel.elf bin/stage2.bin
-	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --comport 3 hyper-v --native --snp
+bin/coconut-hyperv.igvm: $(IGVMBUILDER) $(IGVMMEASURE) bin/stage1-trampoline.bin bin/svsm-kernel.elf bin/stage2.bin
+	$(IGVMBUILDER) --sort --output $@ --tdx-stage1 bin/stage1-trampoline.bin --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --comport 3 hyper-v --native --snp --tdp
 	$(IGVMMEASURE) $@ measure
 
-bin/coconut-test-qemu.igvm: $(IGVMBUILDER)  $(IGVMMEASURE) bin/test-kernel.elf bin/stage2.bin
-	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/test-kernel.elf qemu --snp
+bin/coconut-test-qemu.igvm: $(IGVMBUILDER) $(IGVMMEASURE) bin/stage1-trampoline.bin bin/test-kernel.elf bin/stage2.bin
+	$(IGVMBUILDER) --sort --output $@ --tdx-stage1 bin/stage1-trampoline.bin --stage2 bin/stage2.bin --kernel bin/test-kernel.elf qemu --snp --tdp
 	$(IGVMMEASURE) $@ measure
 
 test:
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ clippy:
 clean:
 	cargo clean
 	rm -f stage1/*.o stage1/*.bin stage1/*.elf
-	rm -f ${STAGE1_OBJS} utils/gen_meta utils/print-meta
+	rm -f utils/gen_meta utils/print-meta
 	rm -rf bin
 
 distclean: clean
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ C_BIT_POS ?= 51
 
 STAGE1_OBJS = stage1/stage1.o stage1/reset.o
 STAGE1_TEST_OBJS = stage1/stage1-test.o stage1/reset.o
+STAGE1_TRAMPOLINE_OBJS = stage1/stage1-trampoline.o stage1/reset.o
 IGVM_FILES = bin/coconut-qemu.igvm bin/coconut-hyperv.igvm
 IGVMBUILDER = "target/x86_64-unknown-linux-gnu/${TARGET_PATH}/igvmbuilder"
 IGVMBIN = bin/igvmbld
@@ -107,7 +108,7 @@ utils/print-meta: utils/print-meta.c
 utils/cbit: utils/cbit.c
 	cc -O3 -Wall -o $@ $<
 
-bin/meta.bin: utils/gen_meta utils/print-meta
+bin/meta.bin: utils/gen_meta utils/print-meta bin
 	./utils/gen_meta $@
 
 bin/stage2.bin: bin
@@ -130,13 +131,16 @@ endif
 
 stage1/stage1.o: stage1/stage1.S bin/stage2.bin bin/svsm-fs.bin bin/svsm-kernel.elf bin
 	ln -sf svsm-kernel.elf bin/kernel.elf
-	cc -c -o $@ stage1/stage1.S
+	cc -c -DLOAD_STAGE2 -o $@ $<
 	rm -f bin/kernel.elf
 
 stage1/stage1-test.o: stage1/stage1.S bin/stage2.bin bin/svsm-fs.bin bin/test-kernel.elf bin
 	ln -sf test-kernel.elf bin/kernel.elf
-	cc -c -o $@ stage1/stage1.S
+	cc -c -DLOAD_STAGE2 -o $@ $<
 	rm -f bin/kernel.elf
+
+stage1/stage1-trampoline.o: stage1/stage1.S
+	cc -c -o $@ $<
 
 stage1/reset.o:  stage1/reset.S bin/meta.bin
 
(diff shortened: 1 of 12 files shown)


bootlib/src/platform.rs
```diff
@@ -10,19 +10,22 @@
 pub enum SvsmPlatformType {
     Native = 0,
     Snp = 1,
+    Tdp = 2,
```
deeglaze, line 13: What's Tdp? Not meant to be Tdx?

msft-jlange:
This PR seems to require a specific choice of platform: either SNP or TDP.  The IGVM format is designed to be multi-platform (this is how the NATIVE support in the IGVM builder is currently implemented).  It would be much, much better if platform selection could be a combination instead of a choice.  I would argue that the default `make igvm` should build an IGVM file that contains both TDP and SNP support in a single file, and should build all of the components needed by both platforms, and that command-line arguments to `make` should only be used if there is a desire to suppress one platform or another.

igvmbuilder/src/cmd_options.rs
```diff
@@ -24,6 +29,10 @@ pub struct CmdOptions {
     #[arg(short, long)]
     pub firmware: Option<String>,
 
+    /// Platform to generate IGVM file for
```
msft-jlange, line 32: This makes it impossible to build a multi-platform IGVM file, as only one platform can be selected here.  It would be much better to have --snp and --tdx as options that are not mutually exclusive, just like --native is already an optional platform target argument that creates multi-platform files.  It would be reasonable to make every platform optional, and to generate an error message only if no platforms were selected.

kernel/src/platform/tdp.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    fn setup_percpu_current(&self, _cpu: &mut PerCpu) -> Result<(), SvsmError> {
+        Err(SvsmError::Tdx)
+    }
+
+    fn get_page_encryption_masks(&self, _vtom: usize) -> PageEncryptionMasks {
+        PageEncryptionMasks {
+            private_pte_mask: 0,
+            shared_pte_mask: 0,
```
msft-jlange, line 50: Shouldn't this include the shared GPA bit 47?  I understand this code isn't actually used for anything yet, but we might as well get the obvious stuff right from the beginning.

bootlib/src/platform.rs
peterfang: TDP refers to TD Partitioning which is part of the TDX technology stack. Since SNP refers to a specific variation of the SEV technology stack, I used Tdp here instead of Tdx.

kernel/src/platform/tdp.rs
peterfang: It can be either bit 47 or bit 51 depending on the output of TDG.VP.INFO (GPAW). This requires a TDCALL library to be imported first. That said, I could make these values a bit more meaningful if that makes it less confusing.

igvmbuilder/src/cmd_options.rs
peterfang: Makes sense. I'll rework this.

peterfang:
> This PR seems to require a specific choice of platform: either SNP or TDP. The IGVM format is designed to be multi-platform (this is how the NATIVE support in the IGVM builder is currently implemented). It would be much, much better if platform selection could be a combination instead of a choice. I would argue that the default `make igvm` should build an IGVM file that contains both TDP and SNP support in a single file, and should build all of the components needed by both platforms, and that command-line arguments to `make` should only be used if there is a desire to suppress one platform or another.



> This PR seems to require a specific choice of platform: either SNP or TDP. The IGVM format is designed to be multi-platform (this is how the NATIVE support in the IGVM builder is currently implemented). It would be much, much better if platform selection could be a combination instead of a choice. I would argue that the default `make igvm` should build an IGVM file that contains both TDP and SNP support in a single file, and should build all of the components needed by both platforms, and that command-line arguments to `make` should only be used if there is a desire to suppress one platform or another.

Thanks for sharing how you envision IGVM multi-platform support in the future. This is very helpful. I'll do some reworking and update this PR soon.

kernel/src/platform/tdp.rs
msft-jlange: The shared GPA position will most likely need to be known long before we can make use of any common library code.  However, it is made available on the BSP as the reset value of RBX.  It would be better if we could pass that out of stage 1 and into stage 2 and the kernel launch info so that it is available from the beginning of execution.

peterfang: The earliest point (to my knowledge) where we would need the shared bit is when we set up the shared page for L0->L1 interrupts. There should be very few cases in SVSM itself where the shared bit would be needed before the L2 guest launches.
Thanks for reminding me about RBX. I'll incorporate it as part of my next revision.

peterfang:
Hi @msft-jlange the PR has been updated. Please take a look. Thanks!

igvmbuilder/src/igvm_builder.rs
```diff
@@ -62,9 +61,9 @@ impl IgvmBuilder {
         // revision.
         let mut use_igvm_v2 = false;
 
-        // SNP is always included in the compatibility mask regardless of the
+        // SNP/TDP is always included in the compatibility mask regardless of the
```
msft-jlange, line 64: I think we'd like to avoid forcing the inclusion of any specific platform.  I think a better approach would be to default to no platforms and require specific command-line options for every platform that should be included.  The IGVM builder should generate an error if no platforms are selected.  Then the Makefile can include both SNP and TDP as a default build target.  This at least leaves open the possibility of building single-platform files in workflows that require it.

Makefile
```diff
@@ -36,16 +36,30 @@ else
 BUILD_FW =
 endif
 
+PLATFORM ?= sev-snp
+ifeq ($(PLATFORM), tdp)
+BUILD_PLATFORM = --stage1 bin/stage1-trampoline.bin
```
msft-jlange, line 41: Since your change intends to include TDP by default in the generated IGVM file, then shouldn't this also be built by default without requiring a specific `PLATFORM=` argument?

Makefile
```diff
@@ -68,16 +82,16 @@ $(IGVMBUILDER):
...
 
-bin/coconut-qemu.igvm: $(IGVMBUILDER) $(IGVMMEASURE) bin/svsm-kernel.elf bin/stage2.bin ${FS_BIN}
-	$(IGVMBUILDER) --sort --policy 0x30000 --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --filesystem ${FS_BIN} ${BUILD_FW} qemu
+bin/coconut-qemu.igvm: $(IGVMBUILDER) $(IGVMMEASURE) $(IGVM_QEMU_BINS)
+	$(IGVMBUILDER) --sort --policy 0x30000 --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --filesystem ${FS_BIN} ${BUILD_FW} ${BUILD_PLATFORM} qemu
 	$(IGVMMEASURE) --check-kvm $@ measure
 
-bin/coconut-hyperv.igvm: $(IGVMBUILDER)  $(IGVMMEASURE) bin/svsm-kernel.elf bin/stage2.bin
+bin/coconut-hyperv.igvm: $(IGVMBUILDER) $(IGVMMEASURE) $(IGVM_HYPERV_BINS)
 	$(IGVMBUILDER) --sort --output $@ --stage2 bin/stage2.bin --kernel bin/svsm-kernel.elf --comport 3 hyper-v --native
```
msft-jlange, line 90: Stage 1 needs to be included in the Hyper-V command line as well as the QEMU command line.

msft-jlange:
Something is wrong with the way stage1 is included, at least in the Hyper-V version of the IGVM file.  The bytes at the reset vector are `e9 0d fe 66 2e 0f 1f 84 00 00 00 00 00 0f 1f 00`.  But because TDX starts in 32-bit protected mode, that first instruction decodes as `jmp 0x2E66FE02` which is clearly not correct.

igvmbuilder/src/igvm_builder.rs
```diff
@@ -391,6 +393,11 @@ impl IgvmBuilder {
             )?;
         }
 
+        // Add optional stage 1 binary.
+        if let Some(stage1) = &self.options.stage1 {
+            self.add_data_pages_from_file(&stage1.clone(), self.gpa_map.stage1_image.get_start())?;
```
msft-jlange, line 398: The stage 1 image must only be included for TDP platforms, not for SNP platforms.  Since the desire is to create a multi-platform IGVM file, the stage 1 TDP image will be included on the command line even if SNP is available as a target platform.  But including stage 1 for SNP will cause the IGVM file to fail to load, because the loader may not treat the reset page as valid memory for SNP VMs.  Therefore, the compatibility mask for the page data for stage 1 must be set to the TDP compatibility mask only.

Makefile
peterfang: Sure, I'll add it to my next revision

peterfang:
> Something is wrong with the way stage1 is included, at least in the Hyper-V version of the IGVM file. The bytes at the reset vector are `e9 0d fe 66 2e 0f 1f 84 00 00 00 00 00 0f 1f 00`. But because TDX starts in 32-bit protected mode, that first instruction decodes as `jmp 0x2E66FE02` which is clearly not correct.

This actually doesn't work in this patchset because stage1 has not been changed yet. In my next PR I'll start adding TDX support in stage1 & stage2 and stage1 will then support 32-bit protected mode at the reset vector.

igvmbuilder/src/igvm_builder.rs
```diff
@@ -253,6 +244,17 @@ impl IgvmBuilder {
...
             ));
         }
+        if COMPATIBILITY_MASK.contains(TDP_COMPATIBILITY_MASK) {
+            self.platforms.push(IgvmPlatformHeader::SupportedPlatform(
+                IGVM_VHS_SUPPORTED_PLATFORM {
+                    compatibility_mask: TDP_COMPATIBILITY_MASK,
+                    highest_vtl: 2,
+                    platform_type: IgvmPlatformType::TDX,
+                    platform_version: 1,
+                    shared_gpa_boundary: param_block.vtom,
```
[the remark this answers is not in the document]
msft-jlange: What you say is only true for the architecture you've chosen for TDP.  It is possible, for example, to build an IGVM file that has no stage 2, but simply loads CR3 with a pre-constructed set of page tables, and obviously such a binary would only work correctly if the shared GPA bit is exactly where those pre-constructed page tables expected it to be.  By performing a check on the IGVM file, the loader can reject files that are known to be incompatible instead of waiting for the TD to crash/panic on its own.  In fact, this is exactly how the Hyper-V IGVM loader works.

We can consider changing the IGVM standard to define a shared GPA of zero to mean that the TD is taking its own responsibility for managing the shared GPA bit, and thus the loader has no obligation to consider compatibility concerns.  But as long as the shared GPA value is non-zero, it must be consistent with what the loader is prepared to support.

igvmbuilder/src/igvm_builder.rs
peterfang: Great thanks for catching this - I'll update this

Makefile
msft-jlange: See my other comment about inclusion of stage 1 only for the TDP platform.  SNP only requires a stage 1 when it is booted directly by QEMU; IGVM files never include a stage 1 for SNP because the IGVM files contain VMSA contents that permit starting up at some place other than the reset vector.

The goal for your first PR should be to construct a multi-platform IGVM file that boots correctly on SNP and is constructed correctly for TDX (the contents of stage1 can be saved for another PR).  That should be achievable if you restrict the stage 1 page data to TDP only.

igvmbuilder/src/igvm_builder.rs
peterfang: OK. I can add support for --platform=xxx,yyy and convert them to compatibility masks.

msft-jlange: The NATIVE platform established a platform of having a separate argument for each platform, like `--native`.  Perhaps we can start with defining `--snp` and `--tdp` as arguments that imply platforms.  In my mind, this is going to be a little bit easier for tooling workflows than having a single `--platform` argument that has to be specially constructed based on which platforms are selected, because it lends itself to Makefile constructs like `$(TDP_PLATFORM_ARG)` which can be either empty or populated based on whether TDP is selected for inclusion.

igvmbuilder/src/igvm_builder.rs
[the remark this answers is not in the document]
peterfang: Thanks for the input. Here's what I think we can do:
- igvmbuilder does a sanity check to make sure VTOM is sane for TDX (must indicate GPAW is 48 or 52, and 0 means no preference)
- QEMU takes `shared_gpa_boundary`, if non-zero, as a hint for the guest's physical address size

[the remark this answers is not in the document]
msft-jlange: The more I think about it, the more I think we should fill in zero, and define zero to mean that the loader should not perform any compatibility check (for the TDX platform type).  This minimizes the amount of friction that will be present in the build pipelines as well as the loaders.

igvmbuilder/src/igvm_builder.rs
peterfang: Makes sense. Will add them as separate arguments.

peterfang:
Hi @msft-jlange I just updated the PR. Please take a look. Thanks.

igvmbuilder/src/cmd_options.rs
```diff
@@ -8,6 +8,10 @@ use clap::{Parser, ValueEnum};
 
 #[derive(Parser, Debug)]
 pub struct CmdOptions {
+    /// Optional stage 1 binary file
+    #[arg(long)]
+    pub stage1: Option<String>,
```
msft-jlange, line 13: Now that this file has been made specific to TDX, I wonder if we should rename this parameter to `--tdx-stage1` or something similar.  I can envision that in the future, there may be another platform that requires use of a stage 1, which must be a different binary than the TDX stage 1, and we don't want any collision to arise.

msft-jlange:
> Hi @msft-jlange I just updated the PR. Please take a look. Thanks.

This is in great shape now!  My only outstanding comment is the one about the stage 1 file argument name.

igvmbuilder/src/cmd_options.rs
peterfang: Sure, I can make this change

msft-jlange (approved):

msft-jlange:
> @msft-jlange the comment has been addressed in the last two commits and the others remain the same as the previous version.

@peterfang, I've submitted a PR review approving the changes.  You will have to mark all of the conversations as Resolved before it will be possible to merge.

igvmbuilder/src/igvm_builder.rs
```diff
@@ -32,8 +32,9 @@ use crate::stage2_stack::Stage2Stack;
 use crate::vmsa::{construct_start_context, construct_vmsa};
 use crate::GpaMap;
 
-pub const SNP_COMPATIBILITY_MASK: u32 = 1;
-pub const NATIVE_COMPATIBILITY_MASK: u32 = 2;
+pub const SNP_COMPATIBILITY_MASK: u32 = 1u32 << 0;
+pub const NATIVE_COMPATIBILITY_MASK: u32 = 1u32 << 1;
```
stefano-garzarella, line 34: nit: I would have made these changes in a separate patch

Makefile
```diff
@@ -158,7 +168,7 @@ clippy:
 clean:
 	cargo clean
 	rm -f stage1/*.o stage1/*.bin stage1/*.elf
-	rm -f ${STAGE1_OBJS} utils/gen_meta utils/print-meta
+	rm -f utils/gen_meta utils/print-meta
```
stefano-garzarella, line 171: Why we don't need to remove `${STAGE1_OBJS}` anymore?

stefano-garzarella: Oh, maybe because the line above already does. Maybe better to do the cleanup in a separate commit.

stefano-garzarella (approved):
I left minor comments, but the end result LGTM.

I tested on SEV-SNP the igvm file, even having the measurement printed from SVSM and it matches the pre-calculated one, so no regression. (I need to add a test on this).

joergroedel:
Hey @peterfang,

Can you please rebase this to the latest upstream branch? There is a conflict and a few missing pieces required by the recent APIC support changes.

Thanks!

peterfang:
> Hey @peterfang,
> 
> Can you please rebase this to the latest upstream branch? There is a conflict and a few missing pieces required by the recent APIC support changes.
> 
> Thanks!

Thanks @joergroedel ! Will rebase as well as incorporate the latest feedback.

igvmbuilder/src/igvm_builder.rs
peterfang: Thanks @stefano-garzarella, I made a separate patch for this.

Makefile
peterfang: A separate patch has been created for this cleanup.

peterfang:
> > Hey @peterfang,
> > Can you please rebase this to the latest upstream branch? There is a conflict and a few missing pieces required by the recent APIC support changes.
> > Thanks!
> 
> Thanks @joergroedel ! Will rebase as well as incorporate the latest feedback.

This PR has been updated based on the latest upstream branch.

PR #362, opened 2024-05-27
stage2: refactor stage2

00xc:
`stage2_main()` was very long and plagued with spread out panics in different places. Split out functionality into separate functions, each returning an error, and panic at the top level.

While we are at it, clean up the kernel layout logic, using the `MemoryRegion` type and updating the loaded region at each stage. Add documentation on each step for clarity.

Finally, add a missing check in the new `prepare_heap()` function to verify that we did not exceed the reserved kernel region size.


00xc:
Rebased on main

PR #364, opened 2024-05-27
sev/ghcb: fix `GHCB::write_buffer()` soundness issues and clean up `GHCBExitCode`

00xc:
* `GHCB::write_buffer()` had several soundness issues, so fix them:
    * Computing the destination pointer could overflow.
    * The resulting pointer could be misaligned.
    * The specified  `T` needed to be `Copy`.
* There is no reason to use associated constants for `GHCBExitCode` values, so simply use enum variants with defined discriminants.

### DIFF
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -118,18 +118,19 @@ impl From<GhcbError> for SvsmError {
     }
 }
 
-enum GHCBExitCode {}
-
-impl GHCBExitCode {
-    pub const RDTSC: u64 = 0x6e;
-    pub const IOIO: u64 = 0x7b;
-    pub const MSR: u64 = 0x7c;
-    pub const RDTSCP: u64 = 0x87;
-    pub const SNP_PSC: u64 = 0x8000_0010;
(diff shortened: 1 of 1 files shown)


kernel/src/sev/ghcb.rs
```diff
@@ -384,28 +384,24 @@ impl GHCB {
...
-                .cast::<u8>()
-                .offset(offset)
-                .cast::<T>();
-            let src = data as *const T;
-
-            ptr::copy_nonoverlapping(src, dst, 1);
-        }
-
+        // SAFETY: we have verified the pointer is aligned and within bounds.
+        unsafe { dst.cast::<T>().copy_from_nonoverlapping(data, 1) }
```
p4zuu, line 405: Would it make sense to also check the alignment of `data`? In other words, could the caller trick this to give something not T-aligned?

p4zuu (approved):

kernel/src/sev/ghcb.rs
p4zuu: Perfect. You can mark this as resolved :)

Issue #349, opened 2024-05-20
Pre-calculated launch measurement doesn't match the attestation report one


stefano-garzarella:
As I mentioned in the last `SVSM call` I have some issues with the pre-calculated launch measurement

I restarted with a clean build, I'm using the following components:
```
igvm: commit 9b27f1b173cbbb6558663c83b62cbd0398e25853 (tag: igvm_defs-v0.2.1, tag: igvm-v0.2.1)
edk2: commit d965a1bd3d331e4b422526368e832f3d327fd0da (HEAD -> svsm, origin/svsm)
qemu: commit 260df2b36be5e31e1ffe6b97e7c846f1d8cd3c52 (HEAD -> svsm-igvm, origin/svsm-igvm)
svsm: commit fc982434231cd1835efc16f6e101afe8023d3dee (origin/main, origin/HEAD, main)
```

On SVSM I applied this patch (you can find the branch here: https://github.com/stefano-garzarella/svsm/tree/get_report):
```diff
diff --git a/kernel/src/greq/pld_report.rs b/kernel/src/greq/pld_report.rs
index dc8ca69..29924c7 100644
--- a/kernel/src/greq/pld_report.rs
+++ b/kernel/src/greq/pld_report.rs
@@ -73,7 +73,7 @@ pub struct SnpReportResponse {
     /// Reserved
     _reserved: [u8; 24],
     /// The attestation report generated by firmware
-    report: AttestationReport,
+    pub report: AttestationReport,
 }
 
 /// Supported values for SnpReportResponse.status
@@ -164,7 +164,7 @@ pub struct AttestationReport {
     /// Guest-provided data
     report_data: [u8; 64],
     /// The measurement calculated at launch
-    measurement: [u8; 48],
+    pub measurement: [u8; 48],
     /// Data provided by the hypervisor at launch
     host_data: [u8; 32],
     /// SHA-384 digest of the ID public key that signed the ID block
diff --git a/kernel/src/svsm.rs b/kernel/src/svsm.rs
index 486d43d..dc85ea5 100755
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -57,6 +57,12 @@ use svsm::vtpm::vtpm_init;
 
 use svsm::mm::validate::{init_valid_bitmap_ptr, migrate_valid_bitmap};
 
+extern crate alloc;
+
+use alloc::{format, vec};
+use svsm::greq::pld_report::SnpReportResponse;
[...]

PR #365, opened 2024-05-27
igvm tools: fix pre-calculated launch measurement with QEMU/KVM

stefano-garzarella:
The latest versions of KVM patches (not yet upstream) set `debug_swap` to false by default. Also being a debug feature, we don't expect it to be enabled in production, so let's disable it to avoid difference between the launch measurement at runtime and the pre-calculated one.

The latest patch adds a new option to check debugging features like SNP debug_swap, but I'm not sure it's needed.

Any comments are welcome!

Tested with one patch on top to print the launch measurement (https://github.com/stefano-garzarella/svsm/tree/get_report_fixed):

```
$ make
...
"target/x86_64-unknown-linux-gnu/debug/igvmmeasure" --check-kvm bin/coconut-qemu.igvm measure

===============================================================================================================
igvmmeasure 'bin/coconut-qemu.igvm'
Launch Digest: B520B306FE8F57D3B14EA1F755A48A9F2E77642A41167B4AC0183AF8BD1BB629E2C3FCB224892E92110100C178EA148C
===============================================================================================================

$ ./scripts/launch_guest.sh --qemu /path/to/qemu-system-x86_64
...
[SVSM] VTPM: Microsoft TPM 2.0 initialized
[SVSM] [CPU 0] Virtual memory pages used: 0 * 4K, 0 * 2M
[SVSM] VMSA PA: 0x8000f34000
[SVSM] Launching Firmware
[SVSM] Launching request-processing task on CPU 0
[SVSM] ERROR: Panic: CPU[0] panicked at kernel/src/svsm.rs:485:9:
SNP Launch Measurement: b520b306fe8f57d3b14ea1f755a48a9f2e77642a41167b4ac0183af8bd1bb629e2c3fcb224892e92110100c178ea148c
```

Fixes: https://github.com/coconut-svsm/svsm/issues/349

### DIFF
--- a/igvmbuilder/src/cmd_options.rs
+++ b/igvmbuilder/src/cmd_options.rs
@@ -51,6 +51,14 @@ pub struct CmdOptions {
     /// Include NATIVE platform target
     #[arg(long, default_value_t = false)]
     pub native: bool,
+
+    /// Enable debug features (e.g. SNP debug_swap)
+    #[arg(short, long, default_value_t = false)]
+    pub debug: bool,
+
+    /// Extra SEV features to be enabled in the VMSA (multiple values can be provided separated by ',')
+    #[arg(long, value_delimiter = ',')]
+    pub sev_features: Vec<SevExtraFeatures>,
 }
 
 impl CmdOptions {
@@ -72,4 +80,13 @@ pub enum Hypervisor {
 
     /// Build an IGVM file compatible with Hyper-V
     HyperV,
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum, Debug)]
+pub enum SevExtraFeatures {
+    ReflectVc,
+    AlternateInjection,
+    DebugSwap,
+    PreventHostIBS,
+    SNPBTBIsolation,
 }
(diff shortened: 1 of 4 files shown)


stefano-garzarella:
@roy-hopkins I'm not sure if `debug_swap` should be always enabled for HyperV. WDYT?

tlendacky:
> The latest versions of KVM patches (not yet upstream) set `debug_swap` to false by default. Also being a debug feature, we don't expect it to be enabled in production, so let's disable it to avoid difference between the launch measurement at runtime and the pre-calculated one.

Note, it isn't a debug feature in the sense of development debugging of SNP. It is for securely using the debug registers within a guest. See section 15.36.13 of AMD APM Volume 2.

stefano-garzarella:
> > The latest versions of KVM patches (not yet upstream) set `debug_swap` to false by default. Also being a debug feature, we don't expect it to be enabled in production, so let's disable it to avoid difference between the launch measurement at runtime and the pre-calculated one.
> 
> Note, it isn't a debug feature in the sense of development debugging of SNP. It is for securely using the debug registers within a guest. See section 15.36.13 of AMD APM Volume 2.

Oh, thanks for point that out.

How do you suggest handling it?
I'm thinking about add an option to igvmbuilder to control the SNP features, WDYT?

stefano-garzarella:
> > I'm thinking about add an option to igvmbuilder to control the SNP features, WDYT?
> 
> I think that is a good idea. Then you know exactly what SEV_FEATURES values you have and can than therefore generate the proper measurement.

Yeah, thanks for your feedback, I'll re-work this PR to introduce that option.

stefano-garzarella:
v2:
- rewrote comments a bit to avoid confusion with SNP debug swap feature [Tom]
- replaced `--debug` with a more generic `--sev-features` option

Issue #359, opened 2024-05-23
Unsound code patterns


p4zuu:
Here are some unsound code patterns possibly remaining after @Freax13's initial issue #104. 

- [ ] [PageTableRef](https://github.com/coconut-svsm/svsm/blob/53a5bd3be8cb4e32ece3bcc221a431cac2bb246c/kernel/src/mm/pagetable.rs#L695) doesn't have a lifetime associated with it, so there's no guarantee that the contained `PageTable` pointer is still live when `PageTableRef::deref(_mut)` dereferences the pointer.
- [ ] [PageRef::new](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/mm/alloc.rs#L701) should be unsafe. The [AsMut<[u8; PAGE_SIZE]>](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/mm/alloc.rs#L720-L725) impl is dubious given that PageRef implements [Clone](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/mm/alloc.rs#L727-L735)
- [ ] [this_cpu_mut](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/cpu/percpu.rs#L498) is unsound as it allows creating multiple mutable references pointing to the same location.
- [ ] [PerCpu::ghcb](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/cpu/percpu.rs#L346) shouldn't give out mutable references with a static lifetime. (@00xc currently working on it)
- [ ] [zero_mem_region](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/utils/util.rs#L53) should be unsafe.
- [ ] [VmsaRef](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/cpu/percpu.rs#L87-L107) allows turning an arbitrary address into a mutable reference. (Will be fixed by #348)
- [ ] Some of the methods of [GuestPtr](https://github.com/coconut-svsm/svsm/blob/739fa7f423f9a19c0615e22d199e24493a0efe84/src/mm/guestmem.rs#L174-L223) should be unsafe.
[...]

PR #366, opened 2024-05-27
cpu: fix IDT and GDT lifetimes on load methods

00xc:
When running the LIDT or LGDT instructions, we pass addresses pointing to architecture-specific tables that the CPU will use during operation. These tables must reside on valid memory during CPU operation to avoid malfunction.

In practical terms, this means the GDT and IDT entries must have a `static` lifetime. This will avoid at compile time the following construct:

```rust
{
    let idt = IDT::new();
    idt.load();
}
// the IDT entries are no longer valid here
```

We could enforce this at the type level by requiring the `load()` method on these structs to take a `&'static lifetime`. However, since we use locking, we cannot get such reference, only a reference that lives as long as a lock guard we get from the lock protecting the struct. However, this does not matter, as we can check that the guard points to a struct with static lifetime (`Guard<'static, T>`).

Thus, rewrite the methods on these structs that have these requirements such that they take a guard pointing to a `'static` IDT or GDT.

### DIFF
--- a/kernel/src/cpu/idt/common.rs
+++ b/kernel/src/cpu/idt/common.rs
@@ -176,8 +176,18 @@ impl IDT {
 
         self
     }
+}
 
-    pub fn load(&self) -> &Self {
+impl Default for IDT {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl WriteLockGuard<'static, IDT> {
+    /// Load an IDT. Its lifetime must be static so that its entries are
+    /// always available to the CPU.
+    pub fn load(&self) {
         let desc: IdtDesc = IdtDesc {
             size: (IDT_ENTRIES * 16) as u16,
             address: VirtAddr::from(self.entries.as_ptr()),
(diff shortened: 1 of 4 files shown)


p4zuu:
This is partially fixing #359.

kernel/src/cpu/idt/common.rs
```diff
@@ -186,20 +196,14 @@ impl IDT {
...
         }
-
-        self
     }
+}
 
+impl ReadLockGuard<'static, IDT> {
     pub fn base_limit(&self) -> (u64, u32) {
-        let base = (self as *const IDT) as u64;
+        let base: *const IDT = core::ptr::from_ref(self);
```
Freax13, line 204:
```suggestion
        let base: *const IDT = self;
```

Unless your intention is to be explicit, `core::ptr::from_ref` isn't required for converting a reference to its pointer type.

00xc: Yes, the idea was to be explicit in the conversion from `&ReadLockGuard<'static, IDT>` to `&IDT` to `*const IDT`.

PR #368, opened 2024-05-28
svsm: implement SVSM Alternate Injection protocol for APIC emulation

msft-jlange:
This change includes an APIC emulator and all of the logic required to enable Alternate Injection, including the extensions to the GHCB protocol to accept interrupts from the hypervisor, as well as an SVSM protocol to enable guest interaction with APIC state.

This change does not include support for APIC timers nor INIT/SIPI.  APIC timers will come in a future change.  INIT/SIPI will likely not be implemented for SNP because processor startup is already controlled through the VCPU creation elements of the SVSM Core Protocol.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -9,6 +9,7 @@ use crate::platform::guest_cpu::GuestCpuState;
 #[derive(Clone, Copy, Debug, Default)]
 pub struct LocalApic {
     irr: [u32; 8],
+    _allowed_irr: [u32; 8],
     isr_stack_index: usize,
     isr_stack: [u8; 16],
     update_required: bool,
@@ -20,6 +21,7 @@ impl LocalApic {
     pub fn new() -> Self {
         LocalApic {
             irr: [0; 8],
+            _allowed_irr: [0; 8],
             isr_stack_index: 0,
             isr_stack: [0; 16],
             update_required: false,
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -5,12 +5,15 @@
 // Author: Jon Lange (jlange@microsoft.com)
 
 use crate::address::VirtAddr;
-use crate::cpu::percpu::PerCpuShared;
+use crate::cpu::percpu::{this_cpu, PerCpuShared};
 use crate::mm::GuestPtr;
 use crate::platform::guest_cpu::GuestCpuState;
 use crate::requests::SvsmCaa;
+use crate::sev::hv_doorbell::HVExtIntStatus;
+use crate::types::GUEST_VMPL;
 
 use bitfield_struct::bitfield;
+use core::sync::atomic::Ordering;
 
 const APIC_REGISTER_APIC_ID: u64 = 0x802;
 const APIC_REGISTER_TPR: u64 = 0x808;
@@ -250,6 +253,10 @@ impl LocalApic {
         cpu_state: &mut T,
         caa_addr: Option<VirtAddr>,
     ) {
+        // Make sure any interrupts being presented by the host have been
+        // consumed.
+        self.consume_host_interrupts();
+
         if self.update_required {
             // Make sure that all previously delivered interrupts have been
             // processed before attempting to process any more.
@@ -440,6 +447,7 @@ impl LocalApic {
             _ => Err(ApicError::ApicError),
         }
     }
+
     pub fn configure_vector(&mut self, vector: u8, allowed: bool) {
         let index = (vector >> 5) as usize;
         let mask = 1 << (vector & 31);
@@ -447,6 +455,104 @@ impl LocalApic {
             self.allowed_irr[index] |= mask;
         } else {
             self.allowed_irr[index] &= !mask;
+        }
+    }
+
+    fn signal_one_host_interrupt(&mut self, vector: u8) {
+        let index = (vector >> 5) as usize;
+        let mask = 1 << (vector & 31);
+        if (self.allowed_irr[index] & mask) != 0 {
+            self.post_interrupt(vector);
+        }
+    }
+
+    fn signal_several_interrupts(&mut self, group: usize, mut bits: u32) {
+        let vector = (group as u8) << 5;
+        while bits != 0 {
+            let index = 31 - bits.leading_zeros();
+            bits &= !(1 << index);
+            self.post_interrupt(vector + index as u8);
+        }
+    }
+
+    pub fn consume_host_interrupts(&mut self) {
+        let hv_doorbell = this_cpu().hv_doorbell().unwrap();
+        let vmpl_event_mask = hv_doorbell.per_vmpl_events.swap(0, Ordering::Relaxed);
+        // Ignore events other than for the guest VMPL.
+        if vmpl_event_mask & (1 << (GUEST_VMPL - 1)) != 0 {
+            let descriptor = &hv_doorbell.per_vmpl[GUEST_VMPL - 1];
+
+            // First consume any level-sensitive vector that is present.
+            let mut flags = HVExtIntStatus::from(descriptor.status.load(Ordering::Relaxed));
+            if flags.level_sensitive() {
+                // Consume the correct vector atomically.
+                loop {
+                    let mut new_flags = flags;
+                    new_flags.set_pending_vector(0);
+                    new_flags.set_level_sensitive(false);
+                    if let Err(fail_flags) = descriptor.status.compare_exchange(
+                        flags.into(),
+                        new_flags.into(),
+                        Ordering::Relaxed,
+                        Ordering::Relaxed,
+                    ) {
+                        flags = fail_flags.into();
+                    } else {
+                        //flags = new_flags;
+                        break;
+                    }
+                }
+                // FIXME - signal the flags.vector as a level-sensitive
+                // interrupt.
+                panic!("No support yet for host-presented level-sensitive interrupts");
+            }
+
+            // If a single vector is present, then signal it, otherwise
+            // process the entire IRR.
+            if flags.multiple_vectors() {
+                // Clear the multiple vectors flag first so that additional
+                // interrupts are presented via the 8-bit vector.  This must
+                // be done before the IRR is scanned so that if additional
+                // vectors are presented later, the multiple vectors flag
+                // will be set again.
+                let multiple_vectors_mask: u32 =
+                    HVExtIntStatus::new().with_multiple_vectors(true).into();
+                descriptor
+                    .status
+                    .fetch_and(!multiple_vectors_mask, Ordering::Relaxed);
+
+                // Handle the special case of vector 31.
+                if flags.vector_31() {
+                    descriptor
+                        .status
+                        .fetch_and(!(1u32 << 31), Ordering::Relaxed);
+                    self.signal_one_host_interrupt(31);
+                }
+
+                for i in 1..8 {
+                    let bits = descriptor.irr[i - 1].swap(0, Ordering::Relaxed);
+                    self.signal_several_interrupts(i, bits & self.allowed_irr[i]);
+                }
+            } else if flags.pending_vector() != 0 {
+                // Atomically consume this interrupt.  If it cannot be consumed
+                // atomically, then it must be because some other interrupt
+                // has been presented, and that can be consumed in another
+                // pass.
+                let mut new_flags = flags;
+                new_flags.set_pending_vector(0);
+                if descriptor
+                    .status
+                    .compare_exchange(
+                        flags.into(),
+                        new_flags.into(),
+                        Ordering::Relaxed,
+                        Ordering::Relaxed,
+                    )
+                    .is_ok()
+                {
+                    self.signal_one_host_interrupt(flags.pending_vector());
+                }
+            }
         }
     }
 }
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -1,0 +1,168 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) Microsoft Corporation
+//
+// Author: Jon Lange (jlange@microsoft.com)
+
+use crate::platform::guest_cpu::GuestCpuState;
+
+#[derive(Clone, Copy, Debug, Default)]
+pub struct LocalApic {
+    irr: [u32; 8],
+    isr_stack_index: usize,
+    isr_stack: [u8; 16],
+    update_required: bool,
+    interrupt_delivered: bool,
+    interrupt_queued: bool,
+}
+
+impl LocalApic {
+    pub fn new() -> Self {
+        LocalApic {
+            irr: [0; 8],
+            isr_stack_index: 0,
+            isr_stack: [0; 16],
+            update_required: false,
+            interrupt_delivered: false,
+            interrupt_queued: false,
+        }
+    }
+
+    fn scan_irr(&self) -> u8 {
+        // Scan to find the highest pending IRR vector.
+        for (i, irr) in self.irr.into_iter().enumerate().rev() {
+            if irr != 0 {
+                let bit_index = 31 - irr.leading_zeros();
+                let vector = (i as u32) * 32 + bit_index;
+                return vector.try_into().unwrap();
+            }
+        }
+        0
+    }
+
+    fn remove_irr(&mut self, irq: u8) {
+        self.irr[irq as usize >> 5] &= !(1 << (irq & 31));
+    }
+
+    fn insert_irr(&mut self, irq: u8) {
+        self.irr[irq as usize >> 5] |= 1 << (irq & 31);
+    }
+
+    fn rewind_pending_interrupt(&mut self, irq: u8) {
+        let new_index = self.isr_stack_index.checked_sub(1).unwrap();
+        assert!(self.isr_stack.get(new_index) == Some(&irq));
+        self.insert_irr(irq);
+        self.isr_stack_index = new_index;
+        self.update_required = true;
+    }
+
+    pub fn check_delivered_interrupts<T: GuestCpuState>(&mut self, cpu_state: &mut T) {
+        // Check to see if a previously delivered interrupt is still pending.
+        // If so, move it back to the IRR.
+        if self.interrupt_delivered {
+            let irq = cpu_state.check_and_clear_pending_interrupt_event();
+            if irq != 0 {
+                self.rewind_pending_interrupt(irq);
+            }
+            self.interrupt_delivered = false;
+        }
+
+        // Check to see if a previously queued interrupt is still pending.
+        // If so, move it back to the IRR.
+        if self.interrupt_queued {
+            let irq = cpu_state.check_and_clear_pending_virtual_interrupt();
+            if irq != 0 {
+                self.rewind_pending_interrupt(irq);
+            }
+            self.interrupt_queued = false;
+        }
+    }
+
+    fn get_ppr_with_tpr(&self, tpr: u8) -> u8 {
+        // Determine the priority of the current in-service interrupt, if any.
+        let ppr = if self.isr_stack_index != 0 {
+            self.isr_stack[self.isr_stack_index]
+        } else {
+            0
+        };
+
+        // The PPR is the higher of the in-service interrupt priority and the
+        // task priority.
+        if (ppr >> 4) > (tpr >> 4) {
+            ppr
+        } else {
+            tpr
+        }
+    }
+
+    fn get_ppr<T: GuestCpuState>(&self, cpu_state: &T) -> u8 {
+        self.get_ppr_with_tpr(cpu_state.get_tpr())
+    }
+
+    fn deliver_interrupt_immediately<T: GuestCpuState>(
+        &mut self,
+        irq: u8,
+        cpu_state: &mut T,
+    ) -> bool {
+        if !cpu_state.interrupts_enabled() || cpu_state.in_intr_shadow() {
+            false
+        } else {
+            // This interrupt can only be delivered if it is a higher priority
+            // than the processor's current priority.
+            let ppr = self.get_ppr(cpu_state);
+            if (irq >> 4) <= (ppr >> 4) {
+                false
+            } else {
+                cpu_state.try_deliver_interrupt_immediately(irq)
+            }
+        }
+    }
+
+    pub fn present_interrupts<T: GuestCpuState>(&mut self, cpu_state: &mut T) {
+        if self.update_required {
+            // Make sure that all previously delivered interrupts have been
+            // processed before attempting to process any more.
+            self.check_delivered_interrupts(cpu_state);
+
+            let irq = self.scan_irr();
+            let current_priority = if self.isr_stack_index != 0 {
+                self.isr_stack[self.isr_stack_index - 1]
+            } else {
+                0
+            };
+
+            // This interrupt is a candidate for delivery only if its priority
+            // exceeds the priority of the highest priority interrupt currently
+            // in service.  This check does not consider TPR, because an
+            // interrupt lower in priority than TPR must be queued for delivery
+            // as soon as TPR is lowered.
+            if (irq & 0xF0) > (current_priority & 0xF0) {
+                // Determine whether this interrupt can be injected
+                // immediately.  If not, queue it for delivery when possible.
+                if self.deliver_interrupt_immediately(irq, cpu_state) {
+                    self.interrupt_delivered = true;
+                } else {
+                    cpu_state.queue_interrupt(irq);
+                    self.interrupt_queued = true;
+                }
+
+                // Mark this interrupt in-service.  It will be recalled if
+                // the ISR is examined again before the interrupt is actually
+                // delivered.
+                self.remove_irr(irq);
+                self.isr_stack[self.isr_stack_index] = irq;
+                self.isr_stack_index += 1;
+            }
+            self.update_required = false;
+        }
+    }
+
+    pub fn perform_eoi(&mut self) {
+        // Pop any in-service interrupt from the stack, and schedule the APIC
+        // for reevaluation.
+        if self.isr_stack_index != 0 {
+            self.isr_stack_index -= 1;
+            self.update_required = true;
+        }
+    }
+}
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -5,7 +5,7 @@
 // Author: Jon Lange (jlange@microsoft.com)
 
 use crate::address::VirtAddr;
-use crate::cpu::percpu::{this_cpu, PerCpuShared};
+use crate::cpu::percpu::{current_ghcb, this_cpu, PerCpuShared};
 use crate::mm::GuestPtr;
 use crate::platform::guest_cpu::GuestCpuState;
 use crate::requests::SvsmCaa;
@@ -21,6 +21,8 @@ const APIC_REGISTER_PPR: u64 = 0x80A;
 const APIC_REGISTER_EOI: u64 = 0x80B;
 const APIC_REGISTER_ISR_0: u64 = 0x810;
 const APIC_REGISTER_ISR_7: u64 = 0x817;
+const APIC_REGISTER_TMR_0: u64 = 0x818;
+const APIC_REGISTER_TMR_7: u64 = 0x81F;
 const APIC_REGISTER_IRR_0: u64 = 0x820;
 const APIC_REGISTER_IRR_7: u64 = 0x827;
 const APIC_REGISTER_ICR: u64 = 0x830;
@@ -104,6 +106,8 @@ pub struct LocalApic {
     allowed_irr: [u32; 8],
     isr_stack_index: usize,
     isr_stack: [u8; 16],
+    tmr: [u32; 8],
+    host_tmr: [u32; 8],
     update_required: bool,
     interrupt_delivered: bool,
     interrupt_queued: bool,
@@ -117,6 +121,8 @@ impl LocalApic {
             allowed_irr: [0; 8],
             isr_stack_index: 0,
             isr_stack: [0; 16],
+            tmr: [0; 8],
+            host_tmr: [0; 8],
             update_required: false,
             interrupt_delivered: false,
             interrupt_queued: false,
@@ -136,18 +142,22 @@ impl LocalApic {
         0
     }
 
-    fn remove_irr(&mut self, irq: u8) {
-        self.irr[irq as usize >> 5] &= !(1 << (irq & 31));
+    fn remove_vector_register(register: &mut [u32; 8], irq: u8) {
+        register[irq as usize >> 5] &= !(1 << (irq & 31));
     }
 
-    fn insert_irr(&mut self, irq: u8) {
-        self.irr[irq as usize >> 5] |= 1 << (irq & 31);
+    fn insert_vector_register(register: &mut [u32; 8], irq: u8) {
+        register[irq as usize >> 5] |= 1 << (irq & 31);
+    }
+
+    fn test_vector_register(register: &[u32; 8], irq: u8) -> bool {
+        (register[irq as usize >> 5] & 1 << (irq & 31)) != 0
     }
 
     fn rewind_pending_interrupt(&mut self, irq: u8) {
         let new_index = self.isr_stack_index.checked_sub(1).unwrap();
         assert!(self.isr_stack.get(new_index) == Some(&irq));
-        self.insert_irr(irq);
+        Self::insert_vector_register(&mut self.irr, irq);
         self.isr_stack_index = new_index;
         self.update_required = true;
     }
@@ -304,12 +314,14 @@ impl LocalApic {
                 // Mark this interrupt in-service.  It will be recalled if
                 // the ISR is examined again before the interrupt is actually
                 // delivered.
-                self.remove_irr(irq);
+                Self::remove_vector_register(&mut self.irr, irq);
                 self.isr_stack[self.isr_stack_index] = irq;
                 self.isr_stack_index += 1;
 
-                // Configure a lazy EOI if possible.
-                if try_lazy_eoi {
+                // Configure a lazy EOI if possible.  Lazy EOI is not possible
+                // for level-sensitive interrupts, because an explicit EOI
+                // is required to acknowledge the interrupt at the source.
+                if try_lazy_eoi && !Self::test_vector_register(&self.tmr, irq) {
                     // A lazy EOI is possible only if there is no other
                     // interrupt pending.  If another interrupt is pending,
                     // then an explicit EOI will be required to prompt
@@ -332,11 +344,29 @@ impl LocalApic {
         }
     }
 
+    fn perform_host_eoi(vector: u8) {
+        // Errors from the host are not expected and cannot be meaningfully
+        // handled, so simply ignore them.
+        let _r = current_ghcb().specific_eoi(vector, GUEST_VMPL.try_into().unwrap());
+        assert!(_r.is_ok());
+    }
+
     pub fn perform_eoi(&mut self) {
         // Pop any in-service interrupt from the stack, and schedule the APIC
         // for reevaluation.
         if self.isr_stack_index != 0 {
             self.isr_stack_index -= 1;
+            let vector = self.isr_stack[self.isr_stack_index];
+            if Self::test_vector_register(&self.tmr, vector) {
+                if Self::test_vector_register(&self.host_tmr, vector) {
+                    Self::perform_host_eoi(vector);
+                    Self::remove_vector_register(&mut self.host_tmr, vector);
+                } else {
+                    // FIXME: should do something with locally generated
+                    // level-sensitive interrupts.
+                }
+                Self::remove_vector_register(&mut self.tmr, vector);
+            }
             self.update_required = true;
             self.lazy_eoi_pending = false;
         }
@@ -352,10 +382,13 @@ impl LocalApic {
         value
     }
 
-    fn post_interrupt(&mut self, irq: u8) {
+    fn post_interrupt(&mut self, irq: u8, level_sensitive: bool) {
         // Set the appropriate bit in the IRR.  Once set, signal that interrupt
         // processing is required before returning to the guest.
-        self.insert_irr(irq);
+        Self::insert_vector_register(&mut self.irr, irq);
+        if level_sensitive {
+            Self::insert_vector_register(&mut self.tmr, irq);
+        }
         self.update_required = true;
     }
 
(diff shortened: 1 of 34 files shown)


kernel/src/cpu/apic.rs
MelodyHuibo, line 1: A small typo in line 269 "an" > "and"

kernel/src/cpu/apic.rs
```diff
@@ -306,12 +317,14 @@ impl LocalApic {
...
+                Self::remove_vector_register(&mut self.irr, irq);
                 self.isr_stack[self.isr_stack_index] = irq;
                 self.isr_stack_index += 1;
 
-                // Configure a lazy EOI if possible.
-                if try_lazy_eoi {
+                // Configure a lazy EOI if possible.  Lazy EOI is not possible
+                // for level-sensitive interrupts, because an explicit EOI
+                // is required to acknowledge the interrupt at the source.
+                if try_lazy_eoi && Self::test_vector_register(&self.tmr, irq) {
```
vijaydhanraj, line 356: Should it should be `!`Self::test_vector_register(&self.tmr, irq) right? As lazy EOI is not possible for level-sensitive interrupts.

kernel/src/cpu/apic.rs
```diff
@@ -227,17 +279,47 @@ impl LocalApic {
...
                 self.isr_stack_index += 1;
+
+                // Configure a lazy EOI if possible.
+                if try_lazy_eoi {
+                    // A lazy EOI is possible only if there is no other
+                    // interrupt pending.  If another interrupt is pending,
+                    // then an explicit EOI will be required to prompt
+                    // delivery of the next interrupt.
+                    if self.scan_irr() == 0 {
+                        self.lazy_eoi_pending = true;
```
vijaydhanraj, line 362: Should this setting ` self.lazy_eoi_pending = true` be removed as it is set again after checking from calling area? Also, comment may need to be updated?

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,851 @@
...
+        // Scan to find the highest pending IRR vector.
+        for i in (0..7).rev() {
+            if self.irr[i] != 0 {
+                let bit_index = 31 - self.irr[i].leading_zeros();
+                let vector = (i as u32) * 32 + bit_index;
+                return vector.try_into().unwrap();
+            }
+        }
+        0
+    }
```
00xc, line 148:
```suggestion
    /// Scan to find the highest pending IRR vector.
    fn scan_irr(&self) -> u8 {
        for (i, irr) in self.irr.into_iter().enumerate().rev() {   
            if irr != 0 {
                let bit_index = 31 - irr.leading_zeros();
                let vector = (i as u32) * 32 + bit_index;
                return vector.try_into().unwrap();
            }
        }
        0
    }
```

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,851 @@
...
+    fn test_vector_register(register: &[u32; 8], irq: u8) -> bool {
+        (register[irq as usize >> 5] & 1 << (irq & 31)) != 0
+    }
+
+    fn rewind_pending_interrupt(&mut self, irq: u8) {
+        assert!(self.isr_stack_index != 0);
+        assert!(self.isr_stack[self.isr_stack_index - 1] == irq);
+        Self::insert_vector_register(&mut self.irr, irq);
+        self.isr_stack_index -= 1;
+        self.update_required = true;
```
00xc, line 167: Using `checked_sub()` (and using a safe array accessor) is probably more readable:

```suggestion
        let stack_index = self.isr_stack_index.checked_sub(1).unwrap();
        assert!(self.isr_stack.get(stack_index) == Some(irq));
        Self::insert_vector_register(&mut self.irr, irq);
        self.isr_stack_index = stack_index;
        self.update_required = true;
```

Although I'd suggest returning an error.

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,851 @@
...
+            let calling_area = GuestPtr::<SvsmCaa>::new(virt_addr);
+            // Ignore errors here, since nothing can be done if an error
+            // occurs.
+            if let Ok(caa) = calling_area.read() {
+                let _ = calling_area.write(caa.update_no_eoi_required(0));
+            }
+            Some(calling_area)
+        } else {
+            None
+        }
```
00xc, line 248: If `caa_addr` is None you can just return early with `?`:

```suggestion
        let virt_addr = caa_addr?;
        let calling_area = GuestPtr::<SvsmCaa>::new(virt_addr);
        // Ignore errors here, since nothing can be done if an error occurs.
        if let Ok(caa) = calling_area.read() {
            let _ = calling_area.write(caa.update_no_eoi_required(0));
        }
        Some(calling_area)
```

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,851 @@
...
+            }
+        }
+    }
+
+    pub fn consume_pending_ipis(&mut self, cpu_shared: &PerCpuShared) {
+        // Scan the IPI IRR vector an transfer any pending IPIs into the local
+        // IRR vector.
+        for i in 0..8 {
+            self.irr[i] |= cpu_shared.ipi_irr_vector(i);
+        }
```
00xc, line 275:
```suggestion
        for (i, irr) in self.irr.iter_mut().enumerate() {
            *irr |= cpu_shared.ipi_irr_vector(i);
        }
```

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,851 @@
...
+    }
+
+    fn get_isr(&self, index: usize) -> u32 {
+        let mut value = 0;
+        for i in 0..self.isr_stack_index {
+            if (usize::from(self.isr_stack[i] >> 5)) == index {
+                value |= 1 << (self.isr_stack[i] & 0x1F)
+            }
+        }
```
00xc, line 410:
```suggestion
        for irq in self.isr_stack.into_iter().take(self.isr_stack_index) {
            if (usize::from(irq >> 5)) == index {
                value |= 1 << (irq & 0x1F)
            }
        }
```

00xc (commented):
I could not go over everything, but I left some comments below. Also, can we get some tests for `LocalApic`?

kernel/src/cpu/apic.rs
```diff
@@ -734,4 +734,78 @@ impl LocalApic {
...
+        // vector field when multiple vectors are present (except for the case
+        // of level-sensitive interrupts).
+        let flags = HVExtIntStatus::from(descriptor.status.load(Ordering::Relaxed));
+        if flags.pending_vector() >= 31 && !flags.level_sensitive() {
+            Self::insert_vector_register(&mut self.irr, flags.pending_vector());
+        }
+
+        // Copy vector 31 if required, and then insert all of the additional
+        // IRR fields into the host IRR.
+        if self.irr[0] & 0x8000 != 0 {
```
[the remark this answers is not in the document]
msft-jlange: Thanks for pointing out the incorrect bit position.

Vector 31 is a valid interrupt vector.  The documentation that I see implies that the lowest allowed interrupt vector in the interrupt architecture is vector 16 (which is associated with priority 1).  Regardless of what the documentation says, it can be observed on physical machines that it possible to send IPIs on vector 31.  Additionally, the Intel TDX 1.5 Base Specification states ```A malicious host VMM or a device may post any virtual interrupt vector in the range 255:31 at any time``` (section 11.9.4) which clarifies the fact that the TDX module will permit the host to post vector 31 via the PID.  Because vector 31 is treated as a legitimate interrupt vector, it must be handled.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -21,13 +21,40 @@ pub struct HVDoorbellFlags {
...
+    rsvd_30_28: u32,
+    pub vector_31: bool,
+}
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct HVExtIntInfo {
+    pub status: AtomicU32,
+    pub irr: [AtomicU32; 7],
+    pub isr: [AtomicU32; 8],
```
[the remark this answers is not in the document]
msft-jlange: See the section "Disable Alternate Injection": ```Reverting to hypervisor APIC emulation also requires 
knowledge of which interrupt(s), if any, have been placed into service. Therefore, the 32 bytes that 
follow the extended interrupt vector associated with the target VMPL are defined as an ISR vector, 
indicating which edge-triggered interrupts are already in service (the hypervisor is already required 
to keep track of which level-sensitive interrupts are currently in service).```

kernel/src/sev/hv_doorbell.rs
```diff
@@ -21,13 +21,40 @@ pub struct HVDoorbellFlags {
...
+
 #[repr(C)]
 #[derive(Debug)]
 pub struct HVDoorbell {
     pub vector: AtomicU8,
     pub flags: AtomicU8,
     pub no_eoi_required: AtomicU8,
-    reserved: u8,
+    pub per_vmpl_events: AtomicU8,
+    reserved_31_4: [u8; 60],
```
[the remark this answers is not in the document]
msft-jlange: Yes.

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,166 @@
...
+        assert!(self.isr_stack[self.isr_stack_index - 1] == irq);
+        self.insert_irr(irq);
+        self.isr_stack_index -= 1;
+        self.update_required = true;
+    }
+
+    pub fn check_delivered_interrupts<T: GuestCpuState>(&mut self, cpu_state: &mut T) {
+        // Check to see if a previously delivered interrupt is still pending.
+        // If so, move it back to the IRR.
+        if self.interrupt_delivered {
```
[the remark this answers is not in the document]
msft-jlange: Good catch - this should be in `deliver_interrupt_immediately()`.

kernel/src/cpu/apic.rs
msft-jlange: Yes.

kernel/src/cpu/apic.rs
msft-jlange: Yes, this should be cleaned up.

kernel/src/platform/snp.rs
```diff
@@ -117,6 +129,76 @@ impl SvsmPlatform for SnpPlatform {
...
+                Err(val) => current = val,
+            }
+        }
+
+        Ok(())
+    }
+
+    fn post_irq(&self, icr: u64) -> Result<(), SvsmError> {
+        current_ghcb().hv_ipi(icr)?;
+        Ok(())
```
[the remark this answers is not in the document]
msft-jlange: This is a point of style that our code base is not at all consistent about.  Personally, I prefer the way that I wrote it, because then if the code is modified later to introduce more logic after the `hv_ipi` call, then everything will continue to work correctly without because the error return will still happen at the correct point.  Besides my personal preference, part of the reason I did it this way was because I was following patterns elsewhere in the code base.  If our position is that your proposed version is the preferred pattern, then we should make that change throughout the code base so that pattern to follow is abundantly clear.

Based on this comment, what do you think is the preferable approach for this specific change?

kernel/src/platform/snp.rs
```diff
@@ -117,6 +129,76 @@ impl SvsmPlatform for SnpPlatform {
         pvalidate_range(region, PvalidateOp::Invalid)
     }
 
+    fn configure_alternate_injection(&mut self, alt_inj_requested: bool) -> Result<(), SvsmError> {
+        // If alternate injection was requested, then it must be supported by
+        // the hypervisor.
+        if alt_inj_requested
+            && !hypervisor_ghcb_features().contains(GHCBHvFeatures::SEV_SNP_EXT_INTERRUPTS)
+        {
+            return Err(NotSupported);
```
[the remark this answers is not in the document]
msft-jlange: I'm confused.  The function is defined to return a `Result<SvsmError>` and `SvsmError` is already imported - so why bother to qualify the error here with the `SvsmError` prefix?  Seems like it reads more simply without it.

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,851 @@
...
+    }
+    const fn from_bits(value: u64) -> Self {
+        match value {
+            3 => Self::AllButSelf,
+            2 => Self::AllWithSelf,
+            1 => Self::OnlySelf,
+            _ => Self::Dest,
+        }
+    }
+}
```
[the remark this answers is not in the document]
msft-jlange: This looks like a nice improvement, but the `bitfield_struct` create explicitly relies on `into_bits()` and `from_bits()` for field accesses (refer to the documentation of that crate).  I tried implementing the change you described, but the compiler gives me errors such as the following:
```
   |
25 | enum IcrDestFmt {
   | --------------- variant or associated item `from_bits` not found for this enum
...
75 | #[bitfield(u64)]
   | ^^^^^^^^^^^^^^^^ variant or associated item not found in `IcrDestFmt`
```

So I think we have to stick with what I have in the current version of the code.

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,851 @@
...
+        match value {
+            7 => Self::ExtInt,
+            6 => Self::Sipi,
+            5 => Self::Init,
+            4 => Self::Nmi,
+            0 => Self::Fixed,
+            _ => Self::Unknown,
+        }
+    }
+}
```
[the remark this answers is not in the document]
msft-jlange: Same problem here, thanks to `bitfield_struct` dependency on `into_bits()` and `from_bits()`.

kernel/src/cpu/apic.rs
msft-jlange: An error here is not appropriate, because the state of the ISR stack is not subject to runtime conditions, i.e. its safety does not depend on set of interrupts in service.  The architecture of APIC interrupt priorities guarantees that it is never possible to have more then 15 interrupts in service at a time.  If any of the assumptions here are violated, it is indicative of a fatal bug in APIC emulation, which is certainly not an error that any caller can be expected to deal with.

Using `checked_sub()` seems advisable because it will panic in case of a bug-induced error instead of an underflow that would cause some other subtle problem elsewhere, so I will make that change.

I don't agree that the `get(i)` notation is more readable than `[i]` notation, particularly in cases where the index is guaranteed to exist.  Particularly since this is only present in a debug assertion, it's not particularly helpful to guard against unexpected bounds violations - in fact, I would prefer to see a panic in such a case to help illustrate that the assertion has failed.

kernel/src/cpu/apic.rs
msft-jlange: This is a clear improvement.

kernel/src/cpu/apic.rs
msft-jlange: Agreed.

kernel/src/cpu/apic.rs
msft-jlange: Fixed.

kernel/src/cpu/apic.rs
msft-jlange: This is a good change.

kernel/src/sev/hv_doorbell.rs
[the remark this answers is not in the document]
vijaydhanraj: ah, I missed this. Thanks for the clarification, it makes sense.

kernel/src/cpu/apic.rs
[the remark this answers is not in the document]
vijaydhanraj: I was checking from the AMD and Intel SDM but if we can observe on physical machines that it possible to send IPIs on vector 31, then it is fair enough to say they are valid. Thanks for confirming.

kernel/src/cpu/apic.rs
```diff
@@ -449,4 +457,102 @@ impl LocalApic {
             self.allowed_irr[index] &= !mask;
         }
     }
+
+    fn signal_one_host_interrupt(&mut self, vector: u8) {
+        let index = (vector >> 5) as usize;
+        let mask = 1 << (vector & 31);
```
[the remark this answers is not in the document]
msft-jlange: The compiler will guarantee that there is no width confusion.

kernel/src/cpu/apic.rs
```diff
@@ -165,4 +251,117 @@ impl LocalApic {
...
+                    self.post_interrupt((value & 0xFF) as u8);
+                    Ok(())
+                }
+            }
+            _ => Err(ApicError::ApicError),
+        }
+    }
+    pub fn configure_vector(&mut self, vector: u8, allowed: bool) {
+        let index = (vector >> 5) as usize;
+        let mask = 1 << (vector & 31);
```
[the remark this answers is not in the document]
msft-jlange: The compiler will guarantee that there is no width confusion.

kernel/src/cpu/apic.rs
```diff
@@ -4,12 +4,98 @@
...
+    pub vector: u8,
+    #[bits(3)]
+    pub message_type: IcrMessageType,
+    pub destination_mode: bool,
+    pub delivery_status: bool,
+    rsvd_13: bool,
+    pub assert: bool,
+    pub trigger_mode: bool,
+    #[bits(2)]
+    pub remote_read_status: usize,
```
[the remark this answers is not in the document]
msft-jlange: For X2APIC, they are technically reserved, though the bit positions for XAPIC have a meaning.  Including that meaning in the definition is harmless.

kernel/src/cpu/apic.rs
```diff
@@ -393,6 +410,128 @@ impl LocalApic {
...
 
+    fn send_logical_ipi(&mut self, icr: ApicIcr) -> bool {
+        let vector = icr.vector();
+        let mut signal = false;
+
+        // Check whether the current CPU matches the destination.
+        let destination = icr.destination();
+        let apic_id = get_current_apic_id();
+        if Self::logical_destination_match(destination, apic_id) {
+            self.post_interrupt(vector, false);
```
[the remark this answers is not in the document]
msft-jlange: No, because in logical format, the logical target set is a bitmask and can include multiple targets.  Matching on the current CPU does not guarantee that there is not also a match on other CPUs.

kernel/src/cpu/apic.rs
```diff
@@ -167,4 +253,117 @@ impl LocalApic {
...
+
+        match register {
+            APIC_REGISTER_APIC_ID => Ok(u64::from(cpu_shared.apic_id())),
+            APIC_REGISTER_IRR_0..=APIC_REGISTER_IRR_7 => {
+                let offset = register - APIC_REGISTER_IRR_0;
+                let index: usize = offset.try_into().unwrap();
+                Ok(self.irr[index] as u64)
+            }
+            APIC_REGISTER_ISR_0..=APIC_REGISTER_ISR_7 => {
+                let offset = register - APIC_REGISTER_IRR_0;
```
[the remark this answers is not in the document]
msft-jlange: Good catch!

kernel/src/sev/hv_doorbell.rs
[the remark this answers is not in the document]
MelodyHuibo: In the alternate Injection spec, Bytes 64-97 (here should be a typo in spec as 95) is for VMPL! and 128-159 for VMPL2, 192-223 for VMPL3, so each of them should be 32 bytes, but  HVExtIntInfo for each VMPL is defined as 64 bytes here, should it be 32 bytes?

MelodyHuibo (commented):
In the alternate Injection spec, Bytes 64-97 (here should be a typo in spec as 95) is for VMPL! and 128-159 for VMPL2, 192-223 for VMPL3, so each of them should be 32 bytes, but  HVExtIntInfo for each VMPL is defined as 64 bytes here, should it be 32 bytes?

kernel/src/sev/hv_doorbell.rs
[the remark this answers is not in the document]
msft-jlange: >In the alternate Injection spec, Bytes 64-97 (here should be a typo in spec as 95) is for VMPL! and 128-159 for VMPL2, 192-223 for VMPL3, so each of them should be 32 bytes, but HVExtIntInfo for each VMPL is defined as 64 bytes here, should it be 32 bytes?

See my comment above, which quotes the document: ```Reverting to hypervisor APIC emulation also requires  knowledge of which interrupt(s), if any, have been placed into service. Therefore, the 32 bytes that  follow the extended interrupt vector associated with the target VMPL are defined as an ISR vector,  indicating which edge-triggered interrupts are already in service (the hypervisor is already required to keep track of which level-sensitive interrupts are currently in service).```

What the spec is saying is that the extended interrupt descriptor is 32 bytes (for IRR information), but the 32 bytes following each extended descriptor is also reserved for ISR information to be given back to the host when Alternate Injection is disabled.  So it is appropriate for the SVSM code to define a 64-byte structure which contains 32 bytes of IRR and 32 bytes of IRR.

You are correct that the spec has a typo where it says 97 instead of 95.  I'll fix that.

msft-jlange:
> In the alternate Injection spec, Bytes 64-97 (here should be a typo in spec as 95) is for VMPL! and 128-159 for VMPL2, 192-223 for VMPL3, so each of them should be 32 bytes, but HVExtIntInfo for each VMPL is defined as 64 bytes here, should it be 32 bytes?

I think you left this comment twice.  See my response to your other comment.

MelodyHuibo:
> > In the alternate Injection spec, Bytes 64-97 (here should be a typo in spec as 95) is for VMPL! and 128-159 for VMPL2, 192-223 for VMPL3, so each of them should be 32 bytes, but HVExtIntInfo for each VMPL is defined as 64 bytes here, should it be 32 bytes?
> 
> I think you left this comment twice. See my response to your other comment.

yes, sorry about that. I see your response, thank you for that

kernel/src/platform/snp.rs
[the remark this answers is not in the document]
00xc: The opposite question could be asked: if the type is already imported, why bother adding a redundant import for a single enum variant? Moreover, how is omitting the type name more clear?

kernel/src/platform/snp.rs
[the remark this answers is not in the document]
00xc: It's just a nitpick, but what you wrote would be in C something like;

```c
int ret = some_call();
if (ret < 0)
    return ret;
return 0;
```

When you could just do:

```c
return some_call();
```

If you introduced more code after the line I suggested it would not compile, so no accidental bugs can happen.

I'm not sure to which other parts in the codebase you're referring to, but it'd be better if they used a single line if possible as well, in my humble opinion. Anyhow, this is a very small stylistic choice so feel free to ignore.

kernel/src/cpu/apic.rs
00xc: The reason for using `.get(i)` is that it won't immediately panic if `i` is out of bounds, not readability. I would then suggest using `assert_eq!(self.isr_stack.get(stack_index), Some(irq))` so that the panic message displays both the obtained value (which will be `None` on an out of bounds access or `Some(val)` if `val` is not expected) and the expected value.

kernel/src/platform/snp.rs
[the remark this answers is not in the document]
msft-jlange: In the C example you gave, the construction ensures that the only values that can be returned are zero or a negative number (the value of `ret`).  Positive numbers cannot accidentally come out.  The pattern you quoted above is actually very, very common in Windows code where we use negative numbers to indicate failure and non-negative numbers to indicate qualified success types - but many functions don't want to expose the non-zero success information from their internal implementations, so they just make sure to return zero (STATUS_SUCCESS) as a final result.  Just returning the status value intact from the last operation is often not desired.

Obviously with Rust, the situation is different, because qualified success values come from `Result<T>` and unqualified success values come from `Result<()>`, and the compiler will not let you mix these up.  Plus, as you say, adding code after the last line will cause a compile error and will force the introduction of `?;` syntax as required.  As you say, it's a style thing: do we want to be forced to modify existing lines of code - even if only to introduce `?;` - when extending the logic of a function?  There is no clearly right answer, but I do think we want to migrate towards a common pattern across the entire code base.

From a quick survey, it didn't take me long to find that `PerCpu::map_self` and `PerCpu::initialize_vm_ranges` follow the pattern I'm using here (`?;` followed by `Ok(())`).  Actually, `PerCpu::setup` is an example of exactly the sort of pattern I'm trying to follow, since that logic is likely to be extended over time.  Digging through the history, I see that this function started with a mix of `Result` and `()` functions, and the `Ok(())` was initially present because the preceding line evaluated to `()`.  The most recent change introduced a new operation at the end that returns `Result` and is written using `?;` syntax, so that it could be inserted without having to disturb the `Ok(())` that was there.  Back to the style question - do we want a style where changes to the function should include repeated removals and insertions of `Ok(())` over time based on the evolving logic, or do we want the changes to be more constrained?

Regardless, I think I'll leave this PR as it is until we settle on a standard pattern to be used throughout the code base.

kernel/src/platform/snp.rs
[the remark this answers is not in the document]
msft-jlange: > Moreover, how is omitting the type name more clear?

It is the case that the existing pattern across the code base is to use the error namespace as part of the error result, so I agree that it is more clear to be explicit here.

kernel/src/cpu/apic.rs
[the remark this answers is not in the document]
msft-jlange: Implementing `from_bits()` and `into_bits()` via `From` is problematic because
1. We would like `from_bits()` and `into_bits()` to be `const` so that we can initialize bitfields as constant expressions if necessary.
2. We cannot implement `From::from` as a `const` because functions in trait implementations cannot be `const` (error E0379).
3. Implementing `from_bits()` and `into_bits()` by calling non-`const` functions forces those functions also to be non-`const`, which contradicts (1) above.

Consequently, I think we have to keep `from_bits()` and `into_bits()` as they are.

That leaves open the possibility that we implement `From` in terms of `from_bits()` or `into_bits()` as you have suggested.  However, there are no callers today that expect to convert between these enumerations and u64, which means these `From` implementations would be dead code.  It doesn't seem like we want to promote dead code patterns unless we think we have a stronger desire to include `From` for all enumerated types (which I don't think is our position today).  What would be your rationale for introducing dead code in this case?

kernel/src/cpu/apic.rs
msft-jlange: I see - you're saying that using `get` means that the assertion failure is more readable because it actually displays as an assertion failure instead of just getting a random panic.  I'm not sure it makes a significant difference when debugging, but I do observe that there is no other assert in the code base that contains an expression that is subject to panic.  Therefore, following your suggestion would be the most consistent so we can maintain the pattern of no panics within assertion clauses.

kernel/src/cpu/apic.rs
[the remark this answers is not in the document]
00xc: Dead trait implementations to do not cause any warnings and can be compiled out. We do not necessarily need the implementation now, but it is a good idea in general to provide methods for callers to write (more) convenient code. In my mind it's like adding free derives to types for common traits. Again, feel free to ignore, but I see this as very effort-free.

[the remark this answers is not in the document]
msft-jlange: It may be easy, but I am wondering about what pattern we are trying to establish.  Do we want every enum that can trivially implement `From` to do so, regardless of whether there is any demand?  This certainly does not appear to be the norm in our code base.  I can certainly understand the desire to favor `From` over `from_bits()`/`to_bits()`, but the goofy behavior of `bitfield_struct` has forced us to avoid `From` in that case.  I still hold the position that there is no value in introducing `From` implementations when there is no demand unless we decide that we will do so for every enum in the code base.

msft-jlange:
> Also, can we get some tests for `LocalApic`?

I've spent some time trying to get some tests built, but in practice it is going to be extremely challenging.  The workhorse of `LocalApic` is `LocalApic::present_interrupts()`.  But that function takes a `PerCpuShared` as a parameter, and it is not possible to construct a standalone `PerCpuShared` (its `new` function is not public); it can only be constructed as part of a `PerCpu`, which requires the static per-CPU virtual address map.  In addition, the function `LocalApic::consume_host_interrupts()` - which is central to APIC emulation - expects to be able to call `this_cpu()`, which is not available in the test environment.

It might be possible to create some traits that abstract these away - such that the SVSM environment supplies trait implementations based on real logic and the test environment supplies trait implementations based on test logic - but this would require the introduction of several abstractions that get in the way both of readability and of performance.  Perhaps someday we will decide that this is a worthwhile tradeoff, but in the first PR, I'm reluctant to bite that off.

It looks like there may be some support for building test code into the SVSM environment, as I see a `test_in_svsm` configuration in multiple places, but I am unable to find any documentation for how this can be used for development or testing purposes.

As much as I would like to see test code, I think it's going to have to wait for a future PR.

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,850 @@
...
+            // below.
+            self.lazy_eoi_pending = false;
+            let guest_caa = Self::clear_guest_eoi_pending(caa_addr);
+
+            // This interrupt is a candidate for delivery only if its priority
+            // exceeds the priority of the highest priority interrupt currently
+            // in service.  This check does not consider TPR, because an
+            // interrupt lower in priority than TPR must be queued for delivery
+            // as soon as TPR is lowered.
+            if (irq & 0xF0) > (current_priority & 0xF0) {
```
00xc, line 321: Let's return early here, e.g.

```rust
if (irq & 0xF0) <= (current_priority & 0xF0) {
    return;
}
```

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,850 @@
...
+        // Errors from the host are not expected and cannot be meaningfully
+        // handled, so simply ignore them.
+        let _r = current_ghcb().specific_eoi(vector, GUEST_VMPL.try_into().unwrap());
+        assert!(_r.is_ok());
+    }
+
+    pub fn perform_eoi(&mut self) {
+        // Pop any in-service interrupt from the stack, and schedule the APIC
+        // for reevaluation.
+        if self.isr_stack_index != 0 {
```
00xc, line 385: Let's return early here, e.g.

```rust
if self.isr_stack_index == 0 {
    return;
}
```

kernel/src/cpu/percpu.rs
```diff
@@ -53,6 +57,10 @@ impl PerCpuInfo {
             cpu_shared,
         }
     }
+
+    pub fn unwrap(&self) -> &'static PerCpuShared {
+        self.cpu_shared
+    }
```
00xc, line 63: Let's name this something else, it's too similar to `Option::unwrap()` or `Result::unwrap()`.

kernel/src/cpu/percpu.rs
```diff
@@ -201,18 +214,39 @@ impl GuestVmsaRef {
...
+            ipi_irr: [
+                AtomicU32::new(0),
+                AtomicU32::new(0),
+                AtomicU32::new(0),
+                AtomicU32::new(0),
+                AtomicU32::new(0),
+                AtomicU32::new(0),
+                AtomicU32::new(0),
+                AtomicU32::new(0),
+            ],
```
00xc, line 240:
```suggestion
            ipi_irr: core::array::from_fn(|_| AtomicU32::new(0)),
```

kernel/src/cpu/percpu.rs
```diff
@@ -270,6 +330,8 @@ pub struct PerCpu {
     runqueue: RefCell<RunQueue>,
     /// WaitQueue for request processing
     request_waitqueue: RefCell<WaitQueue>,
+    /// Local APIC state for APIC emulation
+    apic: RefCell<LocalApic>,
```
00xc, line 334: I wonder if we should have it be `RefCell<Option<LocalApic>>`. That way the type itself conveys whether APIC emulation is enabled, so we could get rid of the `apic_emulation` field. This would also force callers to check the `Option`, avoiding future mistakes where the `apic_emulation` field is not checked.

Enabling and disabling would look like this:

```rust
// enable
self.apic.borrow_mut().insert(LocalApic::new());

// disable
self.apic.borrow_mut().take();
```

What do you think?

kernel/src/cpu/percpu.rs
```diff
@@ -426,6 +489,19 @@ impl PerCpu {
...
+        unsafe {
+            let hv_doorbell = self.hv_doorbell.get();
+            if hv_doorbell.is_null() {
+                None
+            } else {
+                // The HV doorbell page can only ever be borrowed shared, never
+                // mutable, and can safely have a static lifetime.
+                Some(&*hv_doorbell)
+            }
+        }
```
00xc, line 502: [`as_ref()`](https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref) already does the null check for us:

```suggestion
        // SAFETY: The HV doorbell page can only ever be borrowed shared, never
        // mutable, and can safely have a static lifetime.
        unsafe {
            self.hv_doorbell.get().as_ref()
        }
```

kernel/src/cpu/percpu.rs
```diff
@@ -614,6 +699,69 @@ impl PerCpu {
...
+            let caa_addr = vmsa_ref.caa_addr();
+            let vmsa = vmsa_ref.vmsa();
+            self.apic
+                .borrow_mut()
+                .disable_apic_emulation(vmsa, caa_addr);
+            drop(vmsa_ref);
+
+            self.apic_emulation.set(false);
+        }
+        Ok(())
```
00xc, line 717: We could return early here, but also with the change I proposed around wrapping `LocalApic` in an `Option` it looks even better:

```rust
pub fn disable_apic_emulation(&self) -> Result<(), SvsmError> {
    let Some(apic) = self.apic.borrow_mut().take() else {
        return Ok(());
    }
    // ...
    Ok(())
}
```

This pattern naturally also happens in some of the other PerCpu methods using the APIC.

kernel/src/cpu/percpu.rs
00xc: It would also make sure that no APIC state is leaked between a disable and a re-enable, which I assume is the intended behavior, but please correct me if that's not the case.

kernel/src/cpu/percpu.rs
```diff
@@ -614,6 +699,69 @@ impl PerCpu {
...
+        self.apic_emulation.get()
+    }
+
+    pub fn read_apic_register(&self, register: u64) -> Result<u64, ApicError> {
+        let mut vmsa_ref = self.guest_vmsa_ref();
+        let caa_addr = vmsa_ref.caa_addr();
+        let vmsa = vmsa_ref.vmsa();
+        self.apic
+            .borrow_mut()
+            .read_register(self.shared(), vmsa, caa_addr, register)
```
00xc, line 749: For example, here we are not performing the emulation enablement check, which looks inconsistent compared to other methods. It's fine, because we are doing it in the request handling function (`apic_read_register()`), and before that in `apic_protocol_request()`, before calling into this method, but you can see why it's an odd pattern: sometimes we check the state within the `PerCpu` methods, and sometimes from outside, and sometimes we are checking it twice! Using an `Option` would make sure we always make the check once and just once.

This also happens a few lines below with `PerCpu::write_apic_register()` and `PerCpu::configure_apic_vector()`, and their respective calls from SVSM request handling functions.

00xc:
Damn, I submitted the review after it was merged. @msft-jlange could you address some of the comments to see if they'd warrant a follow-up PR?

kernel/src/cpu/percpu.rs
msft-jlange: Once disabled, APIC emulation cannot be reenabled so there is no concern about stage leakage.  However, your point about using `Option` to prevent mistakes due to failure to check the enabled state is a good one.

kernel/src/cpu/percpu.rs
msft-jlange: We cannot perform the `take` until all of the operations associated with disabling emulation have completed successfully.  For example, the call to `SvsmPlatform::disable_apic_emulation()` can fail, and if it does, we do not want to disable emulation (note the `?` syntax in use today).  I agree that this should be changed to make use of the `Option<LocalApic>` design you described, but in a way that doesn't disrupt the existing error flows.

kernel/src/cpu/percpu.rs
msft-jlange: The double layers of checking might seem awkward at first glance, but it's because different layers have different concepts of what an error is.  Take `apic_read_register()` as an example: this function needs to determine whether APIC emulation is active so it can decide whether to return `SVSM_ERR_INVALID_REQUEST`, and then if the actual register access is impossible (for example, because a bad register was specified), then it needs to know to return `SVSM_ERR_INVALID_PARAMETER`.  The distinction between these errors is an aspect of the SVSM protocol, not of APIC emulation.  I don't think we want the implementation of `LocalApic` to be required to understand the `SvsmReqError` error type, but we still need a distinction somehow.  I suppose `LocalApic::read_apic_register()` could be defined to return `Option<Result<(), SvsmError>>` with the intention that `None` means there is no APIC emulation and that `Some(r)` indicates the actual result of emulation - but this also seems like an awkward construct.  I'm inclined to leave this double-layering as it is unless you can suggest a way of representing errors that is not inelegant.

00xc: The `Option<Result<..>>` is not really that awkward. I'll submit a PR with these changes.

msft-jlange: I've already made the switch to `Option<LocalApic>` so I'll include `Option<Result<..>>` in my changes.

kernel/src/cpu/apic.rs
```diff
@@ -0,0 +1,850 @@
...
+
+        match register {
+            APIC_REGISTER_TPR => {
+                // TPR must be an 8-bit value.
+                if value > 0xFF {
+                    Err(ApicError::ApicError)
+                } else {
+                    cpu_state.set_tpr((value & 0xFF) as u8);
+                    Ok(())
+                }
```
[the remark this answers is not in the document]
msft-jlange:
```
error[E0599]: no method named `ok_or` found for enum `Result` in the current scope
   --> kernel/src/cpu/apic.rs:628:47
    |
628 | ...alue).ok_or(ApicError::ApicError)?;
    |          ^^^^^ method not found in `Result<u8, TryFromIntError>`
```

I think I can make this work with `map_err()` instead.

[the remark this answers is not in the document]
msft-jlange: I think this reads a little better:
```Rust
                match u8::try_from(value) {
                    Ok(tpr) => {
                        cpu_state.set_tpr(tpr);
                        Ok(())
                    },
                    Err(_) => Err(ApicError::ApicError),
                }
```

msft-jlange:
> Damn, I submitted the review after the PR was merged. @msft-jlange could you address some of the comments to see if they'd warrant a follow-up PR?

These should be addressed in #386.

PR #377, opened 2024-06-05
svsm: drop `VmsaRegProt` support, add missing SEV features and other improvements

00xc:
* Add missing SEV feature definitions to the kernel and igvmbuilder.
* Get SEV features via the platform API
* Check SEV features in the SVSM kernel again, since we reload them.
* Dump CPUID table after logging is enabled.
* Initialize `reg_prot_nonce` if `VmsaRegProt` is enabled.

### DIFF
--- a/kernel/src/sev/status.rs
+++ b/kernel/src/sev/status.rs
@@ -22,8 +22,10 @@ bitflags! {
         const DBGSWP        = 1 << 7;
         const PREV_HOST_IBS = 1 << 8;
         const BTB_ISOLATION = 1 << 9;
+        const VMPL_SSS      = 1 << 10;
         const SECURE_TSC    = 1 << 11;
         const VMSA_REG_PROT = 1 << 16;
+        const SMT_PROT      = 1 << 17;
     }
 }
 
@@ -163,7 +165,8 @@ pub fn sev_status_verify() {
         | SEVStatusFlags::REST_INJ
         | SEVStatusFlags::PREV_HOST_IBS
         | SEVStatusFlags::BTB_ISOLATION
-        | SEVStatusFlags::VMSA_REG_PROT;
+        | SEVStatusFlags::VMSA_REG_PROT
+        | SEVStatusFlags::SMT_PROT;
 
     let status = sev_flags();
     let required_check = status & required;
--- a/kernel/src/sev/status.rs
+++ b/kernel/src/sev/status.rs
@@ -165,7 +165,6 @@ pub fn sev_status_verify() {
         | SEVStatusFlags::REST_INJ
         | SEVStatusFlags::PREV_HOST_IBS
         | SEVStatusFlags::BTB_ISOLATION
-        | SEVStatusFlags::VMSA_REG_PROT
         | SEVStatusFlags::SMT_PROT;
 
     let status = sev_flags();
--- a/igvmbuilder/src/cmd_options.rs
+++ b/igvmbuilder/src/cmd_options.rs
@@ -89,4 +89,8 @@ pub enum SevExtraFeatures {
     DebugSwap,
     PreventHostIBS,
     SNPBTBIsolation,
+    VmplSSS,
+    SecureTscEn,
+    VmsaRegProt,
+    SmtProtection,
 }
--- a/igvmbuilder/src/vmsa.rs
+++ b/igvmbuilder/src/vmsa.rs
@@ -125,6 +125,10 @@ pub fn construct_vmsa(
             SevExtraFeatures::DebugSwap => features.set_debug_swap(true),
             SevExtraFeatures::PreventHostIBS => features.set_prevent_host_ibs(true),
             SevExtraFeatures::SNPBTBIsolation => features.set_snp_btb_isolation(true),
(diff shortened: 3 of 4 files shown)


kernel/src/sev/status.rs
```diff
@@ -163,7 +165,8 @@ pub fn sev_status_verify() {
         | SEVStatusFlags::REST_INJ
         | SEVStatusFlags::PREV_HOST_IBS
         | SEVStatusFlags::BTB_ISOLATION
-        | SEVStatusFlags::VMSA_REG_PROT;
+        | SEVStatusFlags::VMSA_REG_PROT
+        | SEVStatusFlags::SMT_PROT;
```
msft-jlange, line 169: Enabling register protection will cause the guest VMSA contents to be scrambled such that they are unusable without significant extra code, which I don't see as part of this PR.  It is not safe to declare this feature as supported without having such code in the SVSM or else nothing will work.

kernel/src/cpu/vmsa.rs
```diff
@@ -78,7 +78,17 @@ pub fn init_svsm_vmsa(vmsa: &mut VMSA, vtom: u64) {
     vmsa.vmpl = 0;
     vmsa.vtom = vtom;
 
-    vmsa.sev_features = sev_flags().as_sev_features();
+    let sev_status = sev_flags();
+
+    if sev_status.contains(SEVStatusFlags::VMSA_REG_PROT) {
+        let nonce = rdrand::RdRand::new()
```
stefano-garzarella, line 84: IIRC when I tried `rdrand/rdseed` from that crate (some months ago) I had an issue with SVSM, since they use `cpuid` to check if the instruction is supported, but in SVSM we don't support `cpuid`, so I added a `new_unchecked()` API: https://github.com/nagisa/rust_rdrand/pull/21

Is `rdrand::RdRand::new()` working well now?

00xc: I tested it by pulling it out of the if statement and it worked properly.

kernel/src/sev/status.rs
00xc: If I understand the wording in the spec correctly, this is already done by the hardware, no?

> When an Automatic Exit occurs and this feature is enabled, CPU hardware will obfuscate the values of certain registers before they are written to the VMSA. The hardware will then deobfuscate these values on a subsequent VMRUN.

kernel/src/cpu/vmsa.rs
00xc: Perhaps you used it before we had a `#VC` handler for CPUID?

stefano-garzarella: > Perhaps you used it before we had a `#VC` handler for CPUID?

yeah, is definitely that! I had missed it!
Sorry for the noise ;-)

kernel/src/sev/status.rs
msft-jlange: That's exactly true, and it means that the SVSM will observe obfuscated values when it reaches into the guest VMSA because the guest VMSA is not running at the time the SVSM is running.  For example, `request_processing_main` extracts `vmsa.rax` but has no logic to de-obfuscate it, so it will not handle the request properly.

00xc: Ah right, I missed that case. This could be done by adding accessors in `GuestVmsaRef`, but, since I want to rework that interface first for safe borrowing, I'll likely drop the commit from this PR.

00xc: I also don't see how I can test this at the moment since KVM does not set the flag in `vmsa.sev_features`.

00xc: By the way, we were already announcing the feature as supported before, so I'll add a commit to drop it.

msft-jlange: > This could be done by adding accessors in 'GuestVmsaRef`, but, since I want to rework that interface first for safe borrowing, I'll likely drop the commit from this PR.
It takes a lot more than that, because not all VMSA fields are subject to obfuscation.  The processor advertises the set of fields that are obfuscated through a bitmap describing which 64-bit offsets in the page are subject to obfuscation (as I recall, this comes via the secrets page).  So in addition to accessors, it is necessary to add code to parse the bitmap so the accessors know what they're supposed to do.

> By the way, we were already announcing the feature as supported before, so I'll add a commit to drop it.
I missed that fact, but I think it would be great for you to fix it as part of your PR since you're in that code already.

> I also don't see how I can test this at the moment since KVM does not set the flag in `vmsa.sev_features`.
I think we can satisfy ourselves through code review.  Excluding this bit from the list of supported features will be safe.

00xc: > > By the way, we were already announcing the feature as supported before, so I'll add a commit to drop it.
> 
> I missed that fact, but I think it would be great for you to fix it as part of your PR since you're in that code already.

Yes, I'll add it to this PR.

> > I also don't see how I can test this at the moment since KVM does not set the flag in `vmsa.sev_features`.
> 
> I think we can satisfy ourselves through code review. Excluding this bit from the list of supported features will be safe.

Yes, I meant we cannot test that the XOR decoding is working when the feature **is** enabled.

PR #379, opened 2024-06-11
svsm: pass the memory map to guest firmware

msft-jlange:
This change permits the SVSM kernel to adjust the memory map according to its own needs before passing it to the guest firmware.  This is necessary in case the SVSM requires additional memory beyond the SVSM carveout based on runtime state, such as the total amount of memory in the VM.  The memory map is presented in IGVM format, and its location is specified in the IGVM parameters established when the SVSM IGVM file is built.


peterfang:
I can take a look at the OVMF part of this and see if we can come up with a cross-platform solution.

PR #382, opened 2024-06-14
Make read/write methods from GuestPtr unsafe, and copy_{from, to} methods from SecretsPage as well 

p4zuu:
Leaving as draft because I still have doubts.

Providing partial fix for #359 (tasks 7, 8, 9).

My two interrogations are:

1. There are several uses of the following pattern:
```rust
    let guard = PerCPUPageMappingGuard::create_4k(paddr)?; // paddr possibly guest or HV controlled
    let start = guard.virt_addr();
    let guest_page = GuestPtr::<Tt>::new(start + offset);
    let mut request = unsafe { guest_page.read()? };
``` 
As already discussed with @00xc, if `start + offset` is valid but `start + offset + size_of::<T>` ends on an unmapped page or on an adjacent page belonging to another guest or the SVSM kernel (I don't really know if it's possible), `read()` could raise a #PF and possibly return an error and panicking, or reading content from an unauthorized page. If such scenarii are possible, we should either:
- check that `start +  offset + size_of::<T>` doesn't cross the page end
- allow `GuestPtr::read()/write()` for partial read/write
- more general, we should always check that a guest or the HV can't force `GuestPtr::read()/write()` to return an error that will be treated as a fatal error by the SVSM kernel (that would be valid DoS).

2. In some core protocol handlers, a guest provides a physical address (example the [pvalidate](https://github.com/coconut-svsm/svsm/blob/75b83b3e1a5c860a84ccfe0e4b503e7efee5834f/kernel/src/protocols/core.rs#L312) handler), if I understand the code correctly, the guest-provided physical address pointing to the request entries is only verified if it's mapped or not. My understanding is that a guest could make the kernel read at any mapped address to fetch the `PValidateRequest` (including the kernel memory). Moreover, the guest-provided entries suffer from the same issue, the `paddr` to pvalidate are only verified to be mapped. A guest could make the SVSM kernel pvalidate any page. For the core protocol handlers that work with guest-provided paddrs, my understanding is that we should validate that paddrs only belong to the guest's memory region. I left TODOs is some places (not all of them) to points where there is this issue.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -272,7 +272,10 @@ fn vc_decode_insn(ctx: &X86ExceptionContext) -> Result<Option<DecodedInsnCtx>, S
     // rip and rip+15 addresses should belong to a mapped page.
     // To ensure this, we rely on GuestPtr::read() that uses the exception table
     // to handle faults while fetching.
-    let insn_raw = rip.read()?;
+    // SAFETY: we trust the CPU-provided register state to be valid. Thus, RIP
+    // will point to the instruction that caused #VC to be raised, so it can
+    // safely be read.
+    let insn_raw = unsafe { rip.read()? };
 
     let insn = Instruction::new(insn_raw);
     Ok(Some(insn.decode(ctx)?))
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -394,13 +394,22 @@ pub mod svsm_gdbstub {
             // mapping
 
             let Ok(phys) = this_cpu().get_pgtable().phys_addr(addr) else {
-                // The virtual address is not one that SVSM has mapped. Try safely
-                // writing it to the original virtual address
-                return write_u8(addr, value);
+                // The virtual address is not one that SVSM has mapped.
+                // Try safely writing it to the original virtual address
+                // SAFETY: it is up to the user to ensure that the address we
+                // are writing a breakpoint to is valid.
+                return unsafe { write_u8(addr, value) };
             };
 
             let guard = PerCPUPageMappingGuard::create_4k(phys.page_align())?;
-            write_u8(guard.virt_addr() + phys.page_offset(), value)
+            let dst = guard
+                .virt_addr()
+                .checked_add(phys.page_offset())
+                .ok_or(SvsmError::InvalidAddress)?;
+
+            // SAFETY: guard is a new mapped page, non controllable by user.
+            // We also checked that the destination address didn't overflow.
+            unsafe { write_u8(dst, value) }
         }
     }
 
@@ -532,9 +541,11 @@ pub mod svsm_gdbstub {
         ) -> gdbstub::target::TargetResult<(), Self> {
             let start_addr = VirtAddr::from(start_addr);
             for (off, src) in data.iter().enumerate() {
-                if write_u8(start_addr + off, *src).is_err() {
-                    return Err(TargetError::NonFatal);
-                }
+                let dst = start_addr.checked_add(off).ok_or(TargetError::NonFatal)?;
+
+                // SAFETY: We trust the caller of this trait method to provide a valid address.
+                // We only cheked that start_adddr + off didn't overflow.
+                unsafe { write_u8(dst, *src).map_err(|_| TargetError::NonFatal)? }
             }
             Ok(())
         }
--- a/kernel/src/igvm_params.rs
+++ b/kernel/src/igvm_params.rs
@@ -196,29 +196,35 @@ impl IgvmParams<'_> {
                 return Err(SvsmError::Firmware);
             }
 
-            mem_map
-                .offset(i as isize)
-                .write(IGVM_VHS_MEMORY_MAP_ENTRY {
+            mem_map.offset(i as isize);
+            // SAFETY: mem_map_va points to newly mapped memory, whose physical
+            // address is defined in the IGVM config.
+            unsafe {
+                mem_map.write(IGVM_VHS_MEMORY_MAP_ENTRY {
                     starting_gpa_page_number: u64::from(entry.start()) / PAGE_SIZE as u64,
                     number_of_pages: entry.len() as u64 / PAGE_SIZE as u64,
                     entry_type: MemoryMapEntryType::default(),
                     flags: 0,
                     reserved: 0,
                 })?;
+            }
         }
 
         // Write a zero page count into the last entry to terminate the list.
         let index = map.len();
         if index < max_entries {
-            mem_map
-                .offset(index as isize)
-                .write(IGVM_VHS_MEMORY_MAP_ENTRY {
+            mem_map.offset(index as isize);
+            // SAFETY: mem_map_va points to newly mapped memory, whose physical
+            // address is defined in the IGVM config.
+            unsafe {
+                mem_map.write(IGVM_VHS_MEMORY_MAP_ENTRY {
                     starting_gpa_page_number: 0,
                     number_of_pages: 0,
                     entry_type: MemoryMapEntryType::default(),
                     flags: 0,
                     reserved: 0,
                 })?;
+            }
         }
 
         Ok(())
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -39,9 +39,20 @@ pub fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
     }
 }
 
+/// Writes 1 byte at a virtual address.
+///
+/// # Safety
+///
+/// The caller must verify not to corrupt arbitrary memory, as this function
+/// doesn't make any checks in that regard.
+///
+/// # Returns
+///
+/// Returns an error if the specified address is not mapped or is not mapped
+/// with the appropriate write permissions.
 #[allow(dead_code)]
 #[inline]
-pub fn write_u8(v: VirtAddr, val: u8) -> Result<(), SvsmError> {
+pub unsafe fn write_u8(v: VirtAddr, val: u8) -> Result<(), SvsmError> {
     let mut rcx: u64;
 
     unsafe {
@@ -244,7 +255,10 @@ mod tests {
         let test_address = VirtAddr::from(test_buffer.as_mut_ptr());
         let data_to_write = 0x42;
 
-        write_u8(test_address, data_to_write).unwrap();
+        // SAFETY: test_address points to the virtual address of test_buffer.
+        unsafe {
+            write_u8(test_address, data_to_write).unwrap();
+        }
 
         assert_eq!(test_buffer[0], data_to_write);
     }
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -200,8 +200,16 @@ impl<T: Copy> GuestPtr<T> {
         Self { ptr: p }
     }
 
+    /// # Safety
+    ///
+    /// The caller must verify not to read arbitrary memory, as this function
+    /// doesn't make any checks in that regard.
+    ///
+    /// # Returns
+    ///
+    /// Returns an error if the specified address is not mapped.
     #[inline]
-    pub fn read(&self) -> Result<T, SvsmError> {
+    pub unsafe fn read(&self) -> Result<T, SvsmError> {
         let mut buf = MaybeUninit::<T>::uninit();
 
         unsafe {
@@ -210,13 +218,31 @@ impl<T: Copy> GuestPtr<T> {
         }
     }
 
+    /// # Safety
+    ///
+    /// The caller must verify not to corrupt arbitrary memory, as this function
+    /// doesn't make any checks in that regard.
+    ///
+    /// # Returns
+    ///
+    /// Returns an error if the specified address is not mapped or is not mapped
+    /// with the appropriate write permissions.
     #[inline]
-    pub fn write(&self, buf: T) -> Result<(), SvsmError> {
+    pub unsafe fn write(&self, buf: T) -> Result<(), SvsmError> {
         unsafe { do_movsb(&buf, self.ptr) }
     }
 
+    /// # Safety
+    ///
+    /// The caller must verify not to corrupt arbitrary memory, as this function
+    /// doesn't make any checks in that regard.
+    ///
+    /// # Returns
+    ///
+    /// Returns an error if the specified address is not mapped or is not mapped
+    /// with the appropriate write permissions.
     #[inline]
-    pub fn write_ref(&self, buf: &T) -> Result<(), SvsmError> {
+    pub unsafe fn write_ref(&self, buf: &T) -> Result<(), SvsmError> {
         unsafe { do_movsb(buf, self.ptr) }
     }
 
@@ -269,7 +295,8 @@ mod tests {
         let test_buffer = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
         let test_addr = VirtAddr::from(test_buffer.as_ptr());
         let ptr: GuestPtr<[u8; 15]> = GuestPtr::new(test_addr);
-        let result = ptr.read().unwrap();
+        // SAFETY: ptr points to test_buffer's virtual address
+        let result = unsafe { ptr.read().unwrap() };
 
         assert_eq!(result, test_buffer);
     }
@@ -279,8 +306,9 @@ mod tests {
     #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
     fn test_read_invalid_address() {
         let ptr: GuestPtr<u8> = GuestPtr::new(VirtAddr::new(0xDEAD_BEEF));
-
-        let err = ptr.read();
+        // SAFETY: ptr points to an invalid virtual address (0xDEADBEEF is
+        // unmapped). ptr.read() will return an error but this is expected.
+        let err = unsafe { ptr.read() };
         assert!(err.is_err());
     }
 }
--- a/kernel/src/sev/secrets_page.rs
+++ b/kernel/src/sev/secrets_page.rs
@@ -37,7 +37,7 @@ pub struct SecretsPage {
 
 impl SecretsPage {
     pub const fn new() -> Self {
-        SecretsPage {
+        Self {
             version: 0,
             gctxt: 0,
             fms: 0,
@@ -57,7 +57,13 @@ impl SecretsPage {
         }
     }
 
-    pub fn copy_from(&mut self, source: VirtAddr) {
+    /// Copy secrets page's content pointed by a [`VirtAddr`]
+    ///
+    /// # Safety
+    ///
+    /// The caller should verify that `source` points to mapped memory whose
+    /// size is at least the size of the [`SecretsPage`] structure.
+    pub unsafe fn copy_from(&mut self, source: VirtAddr) {
         let from = source.as_ptr::<SecretsPage>();
 
         unsafe {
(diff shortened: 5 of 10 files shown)


00xc:
> Leaving as draft because I still have doubts.
> 
> Providing partial fix for #359 (tasks 7, 8, 9).
> 
> My two interrogations are:
> 
> 1. There are several uses of the following pattern:
> 
> ```rust
>     let guard = PerCPUPageMappingGuard::create_4k(paddr)?; // paddr possibly guest or HV controlled
>     let start = guard.virt_addr();
>     let guest_page = GuestPtr::<Tt>::new(start + offset);
>     let mut request = unsafe { guest_page.read()? };
> ```
> 
> As already discussed with @00xc, if `start + offset` is valid but `start + offset + size_of::<T>` ends on an unmapped page or on an adjacent page belonging to another guest or the SVSM kernel (I don't really know if it's possible), `read()` could raise a #PF and possibly return an error and panicking, or reading content from an unauthorized page. If such scenarii are possible, we should either:
> 
> * check that `start +  offset + size_of::<T>` doesn't cross the page end
> * allow `GuestPtr::read()/write()` for partial read/write
> * more general, we should always check that a guest or the HV can't force `GuestPtr::read()/write()` to return an error that will be treated as a fatal error by the SVSM kernel (that would be valid DoS).

One case we cannot protect from (as we already discussed) is `#VC` handling of kernel code, as we have no way of resetting to a valid instruction pointer.
 
> 2. In some core protocol handlers, a guest provides a physical address (example the [pvalidate](https://github.com/coconut-svsm/svsm/blob/75b83b3e1a5c860a84ccfe0e4b503e7efee5834f/kernel/src/protocols/core.rs#L312) handler), if I understand the code correctly, the guest-provided physical address pointing to the request entries is only verified if it's mapped or not. My understanding is that a guest could make the kernel read at any mapped address to fetch the `PValidateRequest` (including the kernel memory). Moreover, the guest-provided entries suffer from the same issue, the `paddr` to pvalidate are only verified to be mapped. A guest could make the SVSM kernel pvalidate any page. For the core protocol handlers that work with guest-provided paddrs, my understanding is that we should validate that paddrs only belong to the guest's memory region. I left TODOs is some places (not all of them) to points where there is this issue.

As far as I can tell we already guard against this via `valid_phys_address()`, so I think we're safe for now. Although perhaps we should enforce this at the type level.

p4zuu:
> As far as I can tell we already guard against this via `valid_phys_address()`, so I think we're safe for now. Although perhaps we should enforce this at the type level.

You're right, I misread `init_memory_map()`, I thought `MEMORY_MAP` contained SVSM kernel memory regions, but it actually contains guest regions, so we're good.

p4zuu:
Several uses of `GuestPtr::read()/write()` are to update VMSA's calling area. If I understand correctly, VMSA can't be modified by a guest or by the hypervisor, so we can trust its location and its content. With this, I think we can implement 2 safe methods to read and write CAA (and possibly extrapolate to other VMSA fields), that would remove several `unsafe` that I currently added for this PR. Something like:
```rust
pub fn read_caa(va: VirtAddr) -> Result<SvsmCaa, SvsmError> {
    // SAFETY: HV and guests can't change VMSA's content, therefore CAA.
    unsafe { calling_area.read() }
}
```

kernel/src/cpu/vc.rs
```diff
@@ -272,7 +272,11 @@ fn vc_decode_insn(ctx: &X86ExceptionContext) -> Result<Option<DecodedInsnCtx>, S
     // rip and rip+15 addresses should belong to a mapped page.
     // To ensure this, we rely on GuestPtr::read() that uses the exception table
     // to handle faults while fetching.
-    let insn_raw = rip.read()?;
+    // SAFETY: rip can be any mapped address, including userpace. It's technically
+    // not possible for a user to raise a #VC from a different context (different
+    // user or kernelspace), so rip can't belong to a different user or kernel
+    // memory.
```
00xc, line 278: This is a very convoluted explanation. I would just say: we trust the CPU-provided register state to be valid. Thus, RIP will point to the instruction that caused `#VC` to be raised, so it can safely be read.

I think the worst case scenario is that somehow another CPU might unmap the code page where the instruction resides, causing the read to fail. Anyhow, that is not undefined behavior so it is safe.

kernel/src/debug/gdbstub.rs
```diff
@@ -394,13 +394,20 @@ pub mod svsm_gdbstub {
             // mapping
 
             let Ok(phys) = this_cpu().get_pgtable().phys_addr(addr) else {
-                // The virtual address is not one that SVSM has mapped. Try safely
-                // writing it to the original virtual address
-                return write_u8(addr, value);
+                // SAFETY: The virtual address is not one that SVSM has mapped.
+                // Try safely writing it to the original virtual address
+                return unsafe { write_u8(addr, value) };
```
00xc, line 399: This does not explain why it is safe. This should say something like: it is up to the user to ensure that the address we are writing a breakpoint to is valid.

kernel/src/debug/gdbstub.rs
```diff
@@ -532,9 +539,12 @@ pub mod svsm_gdbstub {
...
             for (off, src) in data.iter().enumerate() {
-                if write_u8(start_addr + off, *src).is_err() {
-                    return Err(TargetError::NonFatal);
-                }
+                let dst = start_addr.checked_add(off).ok_or(TargetError::NonFatal)?;
+
+                // SAFETY: we only cheked that start_adddr + off didn't overflow
+                // The validity (ie. attacker controlability) relies on write_addrs
+                // caller.
+                unsafe { write_u8(dst, *src).map_err(|_| TargetError::NonFatal)? }
```
00xc, line 547: I'm having trouble understanding this one. Also, what do we mean by "attacker" here?

kernel/src/igvm_params.rs
```diff
@@ -196,29 +196,35 @@ impl IgvmParams<'_> {
                 return Err(SvsmError::Firmware);
             }
 
-            mem_map
-                .offset(i as isize)
-                .write(IGVM_VHS_MEMORY_MAP_ENTRY {
+            mem_map.offset(i as isize);
+            // SAFETY: mem_map_va points to a new mapped memory, whose address
+            // is defined in the config.
```
00xc, line 201:
```suggestion
            // SAFETY: mem_map_va points to newly mapped memory, whose physical
            // address is defined in the IGVM config.
```

kernel/src/igvm_params.rs
```diff
@@ -196,29 +196,35 @@ impl IgvmParams<'_> {
...
 
         // Write a zero page count into the last entry to terminate the list.
         let index = map.len();
         if index < max_entries {
-            mem_map
-                .offset(index as isize)
-                .write(IGVM_VHS_MEMORY_MAP_ENTRY {
+            mem_map.offset(index as isize);
+            // SAFETY: mem_map_va points to a new mapped memory, whose address
+            // is defined in the config.
```
00xc, line 218: Same as the previous one.

kernel/src/mm/guestmem.rs
```diff
@@ -39,9 +39,21 @@ pub fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
...
 
+/// Writes 1 byte at a virtual address.
+///
+/// # Safety
+///
+/// The caller should verify that the write address (`v`) is valid for write:
+///  - `v` should belong to a mapped memory region
+///  - `v` should only belong to the issuer's memory space (a guest should not be
+///    able to write in the SVSM kernel's memory, or a different guest's
+///    memory).
```
00xc, line 50: * `v` does not need to be a mapped address for this function to be safe. In fact, that is what this function guarantees: that accessing unmapped memory will simply return an error instead of causing an unexpected #PF or #GP.
* `v` does not need to belong to the issuer's address space. In fact we can use this to write to guest memory from the SVSM kernel IIUC. I'd reword that point to say that the caller needs to take care not to corrupt arbitrary memory, as this function won't make any checks in that regard.

kernel/src/mm/guestmem.rs
```diff
@@ -39,9 +39,21 @@ pub fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
...
+///
+/// This function returns an error if it tries to write to an unmapped address,
+/// and that the #PF or #GP handler returns an error.
```
00xc, line 53: This sentence does not make much sense I think, and also talks about exceptions, which are an implementation detail the caller does not need to know about. I'd simply say:

```suggestion
/// # Returns
///
/// Returns an error if the specified address is not mapped or is not mapped
/// with the appropriate write permissions.
```

kernel/src/mm/guestmem.rs
```diff
@@ -189,8 +201,19 @@ impl<T: Copy> GuestPtr<T> {
...
+    ///
+    /// The caller should verify that the read address (`self.ptr`) is valid
+    /// for read:
+    ///  - it should belong to a mapped memory region
+    ///  - it should only belong to the issuer's memory space (a guest should not be
+    ///    able to read in the SVSM kernel's memory, or a different guest's
+    ///    memory).
+    ///
+    /// This function returns an error if it tries to read to an unmapped address,
+    /// and that the #PF or #GP handler returns an error.
```
00xc, line 214: Same as with `write_u8`

kernel/src/mm/guestmem.rs
```diff
@@ -199,13 +222,35 @@ impl<T: Copy> GuestPtr<T> {
...
+    ///
+    /// The caller should verify that the write address (`self.ptr`) is valid
+    /// for write:
+    ///  - it should belong to a mapped memory region
+    ///  - it should only belong to the issuer's memory space (a guest should not be
+    ///    able to write in the SVSM kernel's memory, or a different guest's
+    ///    memory).
+    ///
+    /// This function returns an error if it tries to write to an unmapped address,
+    /// and that the #PF or #GP handler returns an error.
```
00xc, line 235: Same as with `write_u8`

kernel/src/mm/guestmem.rs
```diff
@@ -199,13 +222,35 @@ impl<T: Copy> GuestPtr<T> {
...
```
00xc, line 251: Same as with `write_u8`

kernel/src/mm/guestmem.rs
```diff
@@ -244,7 +289,10 @@ mod tests {
         let test_address = VirtAddr::from(test_buffer.as_mut_ptr());
         let data_to_write = 0x42;
 
-        write_u8(test_address, data_to_write).unwrap();
+        // SAFETY: it's a test.
+        unsafe {
+            write_u8(test_address, data_to_write).unwrap();
+        }
```
00xc, line 295: We don't want tests corrupting memory, so I'd give an actual explanation.

kernel/src/mm/guestmem.rs
```diff
@@ -255,7 +303,8 @@ mod tests {
         let test_buffer = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
         let test_addr = VirtAddr::from(test_buffer.as_ptr());
         let ptr: GuestPtr<[u8; 15]> = GuestPtr::new(test_addr);
-        let result = ptr.read().unwrap();
+        // SAFETY: this is a test
+        let result = unsafe { ptr.read().unwrap() };
```
00xc, line 307: Same here, we need an explanation.

kernel/src/mm/guestmem.rs
```diff
@@ -265,8 +314,8 @@ mod tests {
     #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
     fn test_read_invalid_address() {
         let ptr: GuestPtr<u8> = GuestPtr::new(VirtAddr::new(0xDEAD_BEEF));
-
-        let err = ptr.read();
+        //SAFETY: this is a test
+        let err = unsafe { ptr.read() };
```
00xc, line 318: Same here, we need an explanation.

kernel/src/sev/secrets_page.rs
```diff
@@ -57,15 +57,35 @@ impl SecretsPage {
...
+    /// Copy secrets page's content pointed by a [`VirtAddr`]
+    ///
+    /// # Safety
+    ///
+    /// The caller should verify that both `self` and `source`:
+    /// * are memory-mapped
+    /// * have been allocated with the size of [`SecretsPage`] structure.
+    ///
+    /// The coller should also verify that `self` doesn't point to an
+    /// attacker-controllable address.
```
00xc, line 69: The caller here is not responsible for verifying `self`. We are under the assumption that we are working with a valid type, since `&mut self` is a safe Rust reference. Otherwise every method call that takes `&self` or `&mut self` should be unsafe for that exact same reason.

kernel/src/sev/secrets_page.rs
```diff
@@ -57,15 +57,35 @@ impl SecretsPage {
...
+    ///
+    /// # Safety
+    ///
+    /// The caller should verify that both `self` and `target`:
+    ///  * are memory-mapped
+    ///  * have been allocated with the size of [`SecretsPage`] structure.
+    ///
+    /// The coller should also verify that `target` doesn't point to an
+    /// attacker-controllable address.
+    pub unsafe fn copy_to(&self, target: VirtAddr) {
```
00xc, line 88: Same here `&self` can be assumed to be valid.

kernel/src/svsm.rs
```diff
@@ -138,7 +138,11 @@ fn copy_secrets_page_to_fw(fw_addr: PhysAddr, caa_addr: PhysAddr) -> Result<(),
         u64::from(caa_addr),
     );
 
-    fw_secrets_page.copy_to(start);
+    // SAFETY: start points to a new allocated and zeroed page.
+    // fw_secrets_page is a static memory region.
+    unsafe {
+        fw_secrets_page.copy_to(start);
+    }
```
00xc, line 145: This should be updated regarding my comment above about `&self`.

00xc (changes requested):
Left some suggestions. On top of that, safety comments regarding the CAA should be updated based on my previous comment above.

kernel/src/cpu/vc.rs
p4zuu: Even worse, I guess the page could be remapped to another guest or the SVSM kernel, but as long as we don't leak the bytes we're trying to read , we should be fine.

kernel/src/debug/gdbstub.rs
00xc: Then let's make the function unsafe and describe why in the function docs.

p4zuu:
> No, I don't think that's how it works. First, the guest's VMSA is different from the SVSM's VMSA - the guest is allowed to update its VMSA. Second, the CA is located in a different physical page than the VMSA. When the guest sets the CA's physical address (via `core_create_vcpu()`) or updates it (via `core_remap_ca()`) we have checks in place to verify it is valid. That is why mapping it and reading/writing to it is safe.

That makes sense to me, thanks. I think there is a 3rd way to update the CA: through `prepare_fw_launch()` raised during early setup with IGVM-provided value, so I guess we can safely trust it too, correct?

kernel/src/mm/guestmem.rs
p4zuu: Updated

kernel/src/sev/secrets_page.rs
p4zuu: Updated

kernel/src/debug/gdbstub.rs
p4zuu: Nvm, `write_addrs()` belongs to the `MultiThreadBase` trait of the `gdbstub` crate. I fear there's nothing we can do

joergroedel:
> ```rust
>     let guard = PerCPUPageMappingGuard::create_4k(paddr)?; // paddr possibly guest or HV controlled
>     let start = guard.virt_addr();
>     let guest_page = GuestPtr::<Tt>::new(start + offset);
>     let mut request = unsafe { guest_page.read()? };
> ```

Longer term we should introduce a safe interface around this pattern, which does all the mapping and bounds checking internally and avoids the `unsafe` keyword everywhere in the code. Just not so sure about TLB flushing, maybe we need a collector type as a base which handles that. Something like this:

```rust
{
    // Create TLB handle
    tlb = tlb_gather::new();
    // Create memory handle which can not outlive TLB handle
    mem_handle = tlb.new_mapping(phys_addr, size);
    let mut foo = mem_handle.read();
    let mut bar = mem_handle::offset(1).read();
    drop(mem_handle);
    // mem_handle goes out of scope, unmapping the memory, TLB handle makes sure the virt range is not
    // re-used until it is destroyed

    mem_handle2 = tlb.new_mapping(paddr2, size2)
    // ... use mem_handle2

    // memhandle2 goes out of scope, mapping is removed
    // tlb goes out of scope and flushes the TLB(s).
}
   
```

joergroedel:
But for now it looks good to me, will merge it once @00xc approves.

kernel/src/mm/guestmem.rs
```diff
@@ -189,8 +200,17 @@ impl<T: Copy> GuestPtr<T> {
...
 
+    /// # Safety
+    ///
+    /// The caller must verify not to read arbitrary memory, as this function
+    /// doesn't make any checks in that regard.
+    ///
+    /// # Returns
+    ///
+    /// Returns an error if the specified address is not mapped or is not mapped
+    /// with the appropriate read permissions.
```
00xc, line 211: On x86 there is no way to restrict read permissions on a page, so I think we cannot get `#GP`.

kernel/src/debug/gdbstub.rs
00xc: Then we should have something like "we trust the caller of this trait method to provide a valid address".

00xc:
> But for now it looks good to me, will merge it once @00xc approves.

We still have some unresolved conversations. Once that's done I'll approve.

p4zuu:
> Longer term we should introduce a safe interface around this pattern, which does all the mapping and bounds checking internally and avoids the `unsafe` keyword everywhere in the code. 

I agree, this pattern happens in quite a few places and there is room for improvements.

> Just not so sure about TLB flushing, maybe we need a collector type as a base which handles that.

Couldn't we just simply create new `read()/write()` methods for `PerCPUPageMappingGuard`?

p4zuu:
> That makes sense to me, thanks. I think there is a 3rd way to update the CA: through `prepare_fw_launch()` raised during early setup with IGVM-provided value, so I guess we can safely trust it too, correct?

Checked with Carlos that the CA address updated in `prepare_fw_launch()` is validated in `validate_fw_memory()`.

kernel/src/mm/guestmem.rs
p4zuu: Nice catch. Updated

kernel/src/debug/gdbstub.rs
p4zuu: Updated

kernel/src/mm/guestmem.rs
p4zuu: Applied the suggestion.

00xc:
> > Just not so sure about TLB flushing, maybe we need a collector type as a base which handles that.
> 
> Couldn't we just simply create new `read()/write()` methods for `PerCPUPageMappingGuard`?

It wouldn't solve the fact that we would need to issue a flush for each guard. It also requires making `PerCPUPageMappingGuard` type-aware. Not a bad idea, but it is a more intrusive change.

00xc:
There are some conflicts to resolve, please rebase.

p4zuu:
> There are some conflicts to resolve, please rebase.

Done

00xc (approved):

PR #385, opened 2024-06-18
cpu/percpu: store LocalApic in an Option

00xc:
Instead of having a separate boolean variable indicating whether APIC emulation is enabled, simply store the `LocalApic` type in an `Option`. When APIC emulation is enabled, the field in the `PerCpu` structure will contain a `Some(LocalApic)`, and a `None` otherwise. This enforces at the type level a consistent state of APIC emulation and removes duplicate checks for such state.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -37,7 +37,7 @@ use crate::types::{PAGE_SHIFT, PAGE_SHIFT_2M, PAGE_SIZE, PAGE_SIZE_2M, SVSM_TR_F
 use crate::utils::MemoryRegion;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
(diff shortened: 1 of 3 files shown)


msft-jlange:
> Instead of having a separate boolean variable indicating whether APIC emulation is enabled, simply store the `LocalApic` type in an `Option`. When APIC emulation is enabled, the field in the `PerCpu` structure will contain a `Some(LocalApic)`, and a `None` otherwise. This enforces at the type level a consistent state of APIC emulation and removes duplicate checks for such state.

I already did all of this work as part of the cleanup you requested following my PR.  I was going to submit it along with all of the rest of the fixes you requested - probably tomorrow.  It probably makes more sense for me to do all of the work so I can test it properly.  Do you agree?

00xc:
Sure, I'll close this then.

PR #386, opened 2024-06-18
cpu/apic: Clean up APIC emulation code

msft-jlange:
This change addresses review feedback that was posted on the APIC emulation PR after it merged.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -102,7 +102,10 @@ pub enum ApicError {
     ApicError,
 }
 
-#[derive(Default, Clone, Copy, Debug)]
+// This structure must never be copied because a silent copy will cause APIC
+// state to be lost.
+#[allow(missing_copy_implementations)]
+#[derive(Default, Debug)]
 pub struct LocalApic {
     irr: [u32; 8],
     allowed_irr: [u32; 8],
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -321,52 +321,54 @@ impl LocalApic {
             // in service.  This check does not consider TPR, because an
             // interrupt lower in priority than TPR must be queued for delivery
             // as soon as TPR is lowered.
-            if (irq & 0xF0) > (current_priority & 0xF0) {
-                // Determine whether this interrupt can be injected
-                // immediately.  If not, queue it for delivery when possible.
-                let try_lazy_eoi = if self.deliver_interrupt_immediately(irq, cpu_state) {
-                    self.interrupt_delivered = true;
+            if (irq & 0xF0) <= (current_priority & 0xF0) {
+                return;
+            }
 
-                    // Use of lazy EOI can safely be attempted, because the
-                    // highest priority interrupt in service is unambiguous.
-                    true
-                } else {
-                    cpu_state.queue_interrupt(irq);
-                    self.interrupt_queued = true;
+            // Determine whether this interrupt can be injected
+            // immediately.  If not, queue it for delivery when possible.
+            let try_lazy_eoi = if self.deliver_interrupt_immediately(irq, cpu_state) {
+                self.interrupt_delivered = true;
 
-                    // A lazy EOI can only be attempted if there is no lower
-                    // priority interrupt in service.  If a lower priority
-                    // interrupt is in service, then the lazy EOI handler
-                    // won't know whether the lazy EOI is for the one that
-                    // is already in service or the one that is being queued
-                    // here.
-                    self.isr_stack_index == 0
-                };
+                // Use of lazy EOI can safely be attempted, because the
+                // highest priority interrupt in service is unambiguous.
+                true
+            } else {
+                cpu_state.queue_interrupt(irq);
+                self.interrupt_queued = true;
 
-                // Mark this interrupt in-service.  It will be recalled if
-                // the ISR is examined again before the interrupt is actually
-                // delivered.
-                Self::remove_vector_register(&mut self.irr, irq);
-                self.isr_stack[self.isr_stack_index] = irq;
-                self.isr_stack_index += 1;
+                // A lazy EOI can only be attempted if there is no lower
+                // priority interrupt in service.  If a lower priority
+                // interrupt is in service, then the lazy EOI handler
+                // won't know whether the lazy EOI is for the one that
+                // is already in service or the one that is being queued
+                // here.
+                self.isr_stack_index == 0
+            };
 
-                // Configure a lazy EOI if possible.  Lazy EOI is not possible
-                // for level-sensitive interrupts, because an explicit EOI
-                // is required to acknowledge the interrupt at the source.
-                if try_lazy_eoi && !Self::test_vector_register(&self.tmr, irq) {
-                    // A lazy EOI is possible only if there is no other
-                    // interrupt pending.  If another interrupt is pending,
-                    // then an explicit EOI will be required to prompt
-                    // delivery of the next interrupt.
-                    if self.scan_irr() == 0 {
-                        if let Some(calling_area) = guest_caa {
-                            if let Ok(caa) = calling_area.read() {
-                                if calling_area.write(caa.update_no_eoi_required(1)).is_ok() {
-                                    // Only track a pending lazy EOI if the
-                                    // calling area page could successfully be
-                                    // updated.
-                                    self.lazy_eoi_pending = true;
-                                }
+            // Mark this interrupt in-service.  It will be recalled if
+            // the ISR is examined again before the interrupt is actually
+            // delivered.
+            Self::remove_vector_register(&mut self.irr, irq);
+            self.isr_stack[self.isr_stack_index] = irq;
+            self.isr_stack_index += 1;
+
+            // Configure a lazy EOI if possible.  Lazy EOI is not possible
+            // for level-sensitive interrupts, because an explicit EOI
+            // is required to acknowledge the interrupt at the source.
+            if try_lazy_eoi && !Self::test_vector_register(&self.tmr, irq) {
+                // A lazy EOI is possible only if there is no other
+                // interrupt pending.  If another interrupt is pending,
+                // then an explicit EOI will be required to prompt
+                // delivery of the next interrupt.
+                if self.scan_irr() == 0 {
+                    if let Some(calling_area) = guest_caa {
+                        if let Ok(caa) = calling_area.read() {
+                            if calling_area.write(caa.update_no_eoi_required(1)).is_ok() {
+                                // Only track a pending lazy EOI if the
+                                // calling area page could successfully be
+                                // updated.
+                                self.lazy_eoi_pending = true;
                             }
                         }
                     }
@@ -383,24 +385,29 @@ impl LocalApic {
     }
 
     pub fn perform_eoi(&mut self) {
-        // Pop any in-service interrupt from the stack, and schedule the APIC
-        // for reevaluation.
-        if self.isr_stack_index != 0 {
-            self.isr_stack_index -= 1;
-            let vector = self.isr_stack[self.isr_stack_index];
-            if Self::test_vector_register(&self.tmr, vector) {
-                if Self::test_vector_register(&self.host_tmr, vector) {
-                    Self::perform_host_eoi(vector);
-                    Self::remove_vector_register(&mut self.host_tmr, vector);
-                } else {
-                    // FIXME: should do something with locally generated
-                    // level-sensitive interrupts.
-                }
-                Self::remove_vector_register(&mut self.tmr, vector);
-            }
-            self.update_required = true;
-            self.lazy_eoi_pending = false;
+        // Pop any in-service interrupt from the stack.  If there is no
+        // interrupt in service, then there is nothing to do.
+        if self.isr_stack_index == 0 {
+            return;
         }
+
+        self.isr_stack_index -= 1;
+        let vector = self.isr_stack[self.isr_stack_index];
+        if Self::test_vector_register(&self.tmr, vector) {
+            if Self::test_vector_register(&self.host_tmr, vector) {
+                Self::perform_host_eoi(vector);
+                Self::remove_vector_register(&mut self.host_tmr, vector);
+            } else {
+                // FIXME: should do something with locally generated
+                // level-sensitive interrupts.
+            }
+            Self::remove_vector_register(&mut self.tmr, vector);
+        }
+
+        // Schedule the APIC for reevaluation so any additional pending
+        // interrupt can be processed.
+        self.update_required = true;
+        self.lazy_eoi_pending = false;
     }
 
     fn get_isr(&self, index: usize) -> u32 {
@@ -454,7 +461,7 @@ impl LocalApic {
         // requested destination.  Skip the current CPU, since it was checked
         // above.
         for cpu_ref in PERCPU_AREAS.iter() {
-            let cpu = cpu_ref.unwrap();
+            let cpu = cpu_ref.as_cpu_ref();
             let this_apic_id = cpu.apic_id();
             if (this_apic_id != apic_id)
                 && Self::logical_destination_match(destination, this_apic_id)
@@ -524,7 +531,7 @@ impl LocalApic {
             // current CPU and indicate that an IPI has been requested.
             let apic_id = this_cpu().get_apic_id();
             for cpu_ref in PERCPU_AREAS.iter() {
-                let cpu = cpu_ref.unwrap();
+                let cpu = cpu_ref.as_cpu_ref();
                 if cpu.apic_id() != apic_id {
                     Self::post_ipi_one_target(cpu, icr);
                 }
@@ -628,11 +635,12 @@ impl LocalApic {
         match register {
             APIC_REGISTER_TPR => {
                 // TPR must be an 8-bit value.
-                if value > 0xFF {
-                    Err(ApicError::ApicError)
-                } else {
-                    cpu_state.set_tpr((value & 0xFF) as u8);
-                    Ok(())
+                match u8::try_from(value) {
+                    Ok(tpr) => {
+                        cpu_state.set_tpr(tpr);
+                        Ok(())
+                    }
+                    Err(_) => Err(ApicError::ApicError),
                 }
             }
             APIC_REGISTER_EOI => {
@@ -640,14 +648,13 @@ impl LocalApic {
                 Ok(())
             }
             APIC_REGISTER_ICR => self.handle_icr_write(value),
-            APIC_REGISTER_SELF_IPI => {
-                if value > 0xFF {
-                    Err(ApicError::ApicError)
-                } else {
-                    self.post_interrupt((value & 0xFF) as u8, false);
+            APIC_REGISTER_SELF_IPI => match u8::try_from(value) {
+                Ok(vector) => {
+                    self.post_interrupt(vector, false);
                     Ok(())
                 }
-            }
+                Err(_) => Err(ApicError::ApicError),
+            },
             _ => Err(ApicError::ApicError),
         }
     }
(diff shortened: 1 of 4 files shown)


kernel/src/cpu/percpu.rs
```diff
@@ -700,66 +682,72 @@ impl PerCpu {
...
-            .read_register(self.shared(), vmsa, caa_addr, register)
+    pub fn read_apic_register(&self, register: u64) -> Option<Result<u64, ApicError>> {
+        if let Some(apic) = self.apic.borrow_mut().deref_mut() {
+            let mut vmsa_ref = self.guest_vmsa_ref();
+            let caa_addr = vmsa_ref.caa_addr();
+            let vmsa = vmsa_ref.vmsa();
+            Some(apic.read_register(self.shared(), vmsa, caa_addr, register))
+        } else {
+            None
+        }
```
00xc, line 726:
```suggestion
        let apic = self.apic.borrow_mut().deref_mut()?;
        let mut vmsa_ref = self.guest_vmsa_ref();
        let caa_addr = vmsa_ref.caa_addr();
        let vmsa = vmsa_ref.vmsa();
        Some(apic.read_register(self.shared(), vmsa, caa_addr, register))
```

kernel/src/cpu/percpu.rs
```diff
@@ -700,66 +682,72 @@ impl PerCpu {
...
 
-    pub fn write_apic_register(&self, register: u64, value: u64) -> Result<(), ApicError> {
-        let mut vmsa_ref = self.guest_vmsa_ref();
-        let caa_addr = vmsa_ref.caa_addr();
-        let vmsa = vmsa_ref.vmsa();
-        self.apic
-            .borrow_mut()
-            .write_register(vmsa, caa_addr, register, value)
+    pub fn write_apic_register(&self, register: u64, value: u64) -> Option<Result<(), ApicError>> {
+        if let Some(apic) = self.apic.borrow_mut().deref_mut() {
```
00xc, line 730: Same thing here, let's use `?`

kernel/src/cpu/percpu.rs
```diff
@@ -700,66 +682,72 @@ impl PerCpu {
...
     }
 
-    pub fn configure_apic_vector(&self, vector: u8, allowed: bool) {
-        self.apic.borrow_mut().configure_vector(vector, allowed)
+    pub fn configure_apic_vector(
+        &self,
+        vector: u8,
+        allowed: bool,
+    ) -> Option<Result<(), SvsmError>> {
+        if let Some(apic) = self.apic.borrow_mut().deref_mut() {
```
00xc, line 745: Same thing here, let's use `?`

00xc (commented):
Just a few minor nitpicks below.

On the topic of error handling, I noticed the following in the latest version of the Alternate Injection spec you sent. Under "APIC Emulation Configuration (Call 1)" it says:

> Once Alternate Injection is disabled on the calling vCPU, further calls to the SVSM APIC Protocol will return SVSM_ERR_UNSUPPORTED_PROTOCOL.

But then the rest of the calls after that say:

> If APIC emulation is not active, then SVSM_ERR_INVALID_REQUEST will be returned.

Which in practical terms means, if I understand correctly, that we need to know if APIC emulation has never been enabled (because alternate injection is not used), or active, or disabled after it was active, in order to return the correct error.

We currently (before and after this PR) do not distinguish between the first and the third case. In fact we always return `SVSM_ERR_UNSUPPORTED_PROTOCOL` from `apic_protocol_request()`. If we don't amend the spec, we should probably check `SVSM_PLATFORM.as_dyn_ref().use_alternate_injection()` from `apic_protocol_request()` to see if we return `SVSM_ERR_UNSUPPORTED_PROTOCOL`, and otherwise proceed, returning `SVSM_ERR_INVALID_REQUEST` if the new `Option` contains `None`. What do you think?

kernel/src/cpu/percpu.rs
```diff
@@ -755,11 +749,12 @@ impl PerCpu {
         let vmsa = vmsa_ref.vmsa();
         self.apic
             .borrow_mut()
+            .unwrap()
```
stefano-garzarella: This can panic here now, should we return an error instead?

kernel/src/cpu/percpu.rs
```diff
@@ -755,11 +749,12 @@ impl PerCpu {
...
             .write_register(vmsa, caa_addr, register, value)
     }
 
     pub fn configure_apic_vector(&self, vector: u8, allowed: bool) {
-        self.apic.borrow_mut().configure_vector(vector, allowed)
+        self.apic.borrow_mut().unwrap().configure_vector(vector, allowed)
```
stefano-garzarella: Ditto. IIUC previous implementation allowed this call if apic was enabled or not, not sure if it was right, but at least wasn't panicking. So not sure if now is better to ignore it or add a return error.

kernel/src/protocols/apic.rs
```diff
@@ -53,37 +53,38 @@ fn apic_configure(params: &RequestParams) -> Result<(), SvsmReqError> {
...
-    if !cpu.use_apic_emulation() {
-        return Err(SvsmReqError::invalid_request());
-    }
+
     if params.rcx <= 0x1FF {
         let vector: u8 = (params.rcx & 0xFF) as u8;
         let allowed = (params.rcx & 0x100) != 0;
-        cpu.configure_apic_vector(vector, allowed);
-        Ok(())
+        match cpu.configure_apic_vector(vector, allowed) {
```
stefano-garzarella, line 79: What about returning these errors directly in  `configure_apic_vector()` so we don't need to return the `Option`?

kernel/src/cpu/apic.rs
```diff
@@ -382,22 +384,24 @@ impl LocalApic {
     pub fn perform_eoi(&mut self) {
         // Pop any in-service interrupt from the stack, and schedule the APIC
```
stefano-garzarella, line 385: What about moving this comment after the early return?

kernel/src/cpu/percpu.rs
stefano-garzarella: Oh, I was reviewing this PR per-commit, so I noticed now you fixed in the next commit.
Not a strong opinion, but I'd introduce the error/option before this commit.

stefano-garzarella (commented):
Thanks for the cleanup! I left some minor comments.

kernel/src/cpu/percpu.rs
msft-jlange: Yes, that is better.

kernel/src/cpu/percpu.rs
msft-jlange: Based solely on the first commit, this code should be unreachable when `self.apic` is `None` because the caller will check whether APIC emulation is enabled.  In such a case, this call is technically illegal, in which case panicking is appropriate.  Changing the error paths is an improvement, but even if it were not done, the panic here is reasonable.

kernel/src/protocols/apic.rs
msft-jlange: The format of `params.rcx` is entirely a protocol issue and has nothing to do with how APIC emulation is performed.  Therefore, the decode of the `rcx` parameter should be in the protocol handler and not in the APIC emulation layer.

kernel/src/cpu/percpu.rs
msft-jlange: The compiler disagrees with your proposal:
```
error[E0277]: the `?` operator can only be applied to values that implement `Try`
   --> kernel/src/cpu/percpu.rs:737:20
    |
737 |         let apic = self.apic.borrow_mut().deref_mut()?;
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `?` operator cannot be applied to type `&mut Option<LocalApic>`
    |
    = help: the trait `Try` is not implemented for `&mut Option<LocalApic>`
    = help: the trait `Try` is implemented for `Option<T>`

For more information about this error, try `rustc --explain E0277`.
```

I will experiment to see if I can figure out how to coerce the values correctly.

kernel/src/cpu/percpu.rs
msft-jlange: Same response applies here about the call being illegal if APIC emulation is not enabled.

kernel/src/cpu/percpu.rs
00xc: I think you can omit deref_mut(), or replace it with as_mut(). There's a slightly cleaner way of doing these by using special getters for the APIC (see apic() and apic_mut() in the PR I closed where I did a similar implementation)

msft-jlange: I had to do this to get it to compile:
```Rust
        let mut maybe_apic = self.apic.borrow_mut();
        let apic = maybe_apic.as_mut()?;
```

The separate `maybe_apic` is required because something has to anchor the lifetime of the `RefMut` returned by `borrow_mut()`.

Once I have to go through these contortions, I'm less convinced that the new code is more readable - more concise, maybe, but I find the intent of the previous version to be much more obvious.  But I can make the change anyway.

00xc: Yes, that's why I used `RefMut::filter_map()` in my PR, as it can be used to create an `Option<RefMut<'_, LocalApic>>`

msft-jlange: That approach is clearly much more readable, so I'll take it if you don't mind.

Is the call to `RefMut::filter_map().ok()` subject to panic?  And if `filter_map` can never return `Err` in this usage, is the compiler going to generate wasted code to detect a panic that will never happen?  I'm generally not enthusiastic about an "improvement" to code that helps readability but causes the compiler to emit worse code as a result.

00xc: No, it never panics. filter_map() simply creates a new RefMut from a field U which itself lives another type T that is also wrapped by a RefMut. To do this conversion it uses the provided function or closure, returning Ok(U) if the function returns Some(U), or Err(T) if the function returns None (i.e. if APIC emulation is disabled). I have not looked at the generated code though.

00xc: In other words, it attempts to transform `RefMut<T>` into `RefMut<U>` through a function `f(T) -> Option<U>`. If `f` returns `Some(U)` you get `Ok(RefMut<U>)`, if it returns `None` you get `Err(RefMut<T>)`

msft-jlange: I misunderstood the function of `ok()` so I agree there is no concern here, including with the generated code.

msft-jlange:
> On the topic of error handling, I noticed the following in the latest version of the Alternate Injection spec you sent.

What you highlighted is clearly inconsistent and needs to be fixed.  I will plan to revise the document.

The correct behavior is to return SVSM_ERR_UNSUPPORTED_PROTOCOL once emulation has been disabled on the calling vCPU.  For protocol handling, this check is performed in `apic_protocol_request()`, so that a consistent error can be returned.  Since this is the case, there isn't really any reason to do additional enablement checking in any of the specific protocol handlers because nothing can change between this outermost check and the individual operations.  That means that the change to use `Option<Result<..>>` isn't really necessary - or even useful - because none of those code paths will ever be able to generate `None`.  Consequently, I think it's better for the individual handlers to just assume that APIC emulation is active, and to use `unwrap()` to access the local APIC from the `Option<LocalApic>`.  This will panic if these functions are called unexpectedly when APIC emulation is inactive, but that's a good thing, since we're trying to establish the pattern that they should never be called unless APIC emulation is known to be active.

I think it is preferable to generate SVSM_ERR_UNSUPPORTED_PROTOCOL from a single point rather than requiring each individual function handler to know what error is expected when emulation is inactive.

kernel/src/cpu/percpu.rs
msft-jlange: I appreciate the learning opportunity, but by removing the optionality of the error (as justified in the description of how to handle SVSM_ERR_UNSUPPORTED_PROTOCOL), this is now moot.

00xc (commented):
As it is, this PR still has redundant checking: we check first in the request handler if emulation is enabled or not, and then we do an unwrap() inside some PerCpu methods, which in practice will never panic because of the previous redundant check.

In my opinion we should make it so that it looks like this (simplified):

```rust
fn apic_request() -> Result<(), SvsmReqError> {
    if !SVSM_PLATFORM.as_dyn_ref().use_alternate_injection() {
        // APIC emulation was never enabled
        return Err(SvsmReqError::unsupported_protocol());
    }   
    match request {
        SOME_REQUEST => some_handler()
        ...
    }
}

fn some_handler() -> Result<(), SvsmReqError> {
    this_cpu()
        .some_apic_method()
        // APIC emulation was disabled
        .ok_or_else(SvsmReqError::invalid_request)
        // Failed to service APIC request
        .map_err(|_| SvsmReqError::invalid_parameter())
}
```

This would match the spec and not introduce redundant checks.

The rest of the cleanups LGTM.

00xc:
> I think it is preferable to generate SVSM_ERR_UNSUPPORTED_PROTOCOL from a single point rather than requiring each individual function handler to know what error is expected when emulation is inactive.

The problem is, as it is, we are checking twice whether the `Option` is `None` or not, and that has two meanings: either APIC emulation was never enabled (restricted injection was not enabled) or it was disabled after a request. We need the following to match the spec:

1. If restricted injection is disabled, return `UNSUPPORTED_PROTOCOL`.
2. If APIC is `None`, return `INVALID_REQUEST`.
3. If APIC failed to service request, return `INVALID_PARAMETER`.
4. Otherwise, return `Ok(())`.

msft-jlange:
> The problem is, as it is, we are checking twice whether the `Option` is `None` or not, and that has two meanings: either APIC emulation was never enabled (restricted injection was not enabled) or it was disabled after a request. We need the following to match the spec:

First, the spec is wrong and has been updated (but not yet published).  The correct behavior is to return SVSM_ERR_UNSUPPORTED_PROTOCOL in all cases where APIC emulation is not enabled (either because it was never enabled or because it was disabled at some point).  The spec no longer contemplates the possibility of SVSM_ERR_INVALID_REQUEST, only of SVSM_ERR_INVALID_PARAMETER.

Beyond that, it seems we have a clear difference of opinion.  You are optimizing to minimize the number of times the `Option` is checked.  I am optimizing to minimize the number of places that the code has to be aware of the circumstances in which SVSM_ERR_UNSUPPORTED_PROTOCOL must be detected.  My opinion is that, on balance, the code is more maintainable if the is-enabled check is done once at the top of the protocol handler, so that any future protocol commands do not require awareness of the is-enabled checking, and the work required to check the `Option` a second time is negligible and should not factor into performance/size calculations.  So I'm tilting in favor of a code base that I believe can be more easily maintained.

What we do clearly agree on is that switching to `Option` was a real improvement in safety because there is no longer any ambiguity about whether the APIC state has meaning, and that's something this PR clearly accomplishes.

00xc:
> > The problem is, as it is, we are checking twice whether the `Option` is `None` or not, and that has two meanings: either APIC emulation was never enabled (restricted injection was not enabled) or it was disabled after a request. We need the following to match the spec:
> 
> First, the spec is wrong and has been updated (but not yet published). The correct behavior is to return SVSM_ERR_UNSUPPORTED_PROTOCOL in all cases where APIC emulation is not enabled (either because it was never enabled or because it was disabled at some point). The spec no longer contemplates the possibility of SVSM_ERR_INVALID_REQUEST, only of SVSM_ERR_INVALID_PARAMETER.

Then my point still stands, but we simply omit step 1.

> Beyond that, it seems we have a clear difference of opinion. You are optimizing to minimize the number of times the `Option` is checked. I am optimizing to minimize the number of places that the code has to be aware of the circumstances in which SVSM_ERR_UNSUPPORTED_PROTOCOL must be detected. My opinion is that, on balance, the code is more maintainable if the is-enabled check is done once at the top of the protocol handler, so that any future protocol commands do not require awareness of the is-enabled checking, and the work required to check the `Option` a second time is negligible and should not factor into performance/size calculations. So I'm tilting in favor of a code base that I believe can be more easily maintained.

I simply don't see how doing `is_some()` upfront and then `unwrap()` in a completely separate function, under the assumption that the check happens somewhere else, is more maintainable. It's not about runtime performance, it's about robustness.

You can actually have both goals: remove the first check, let the `PerCpu` methods return `Option<Result<...>>`, and let the top-level APIC handler decide what to do when the `Option` is `None`. This removes redundant checking and has the property you want to optimize for.

> What we do clearly agree on is that switching to `Option` was a real improvement in safety because there is no longer any ambiguity about whether the APIC state has meaning, and that's something this PR clearly accomplishes.

Yes, but it does the equivalent of:

```rust
if some_option.is_some() {
    some_option.unwrap().some_method();
} else {
    Err(some_error)
}
```

Which you can clearly see why is worse. It's not even that, because the `unwrap()` happens on a function on a completely different file, under the assumption that it will never be called because the first check happens somewhere else. This is less maintainable in my humble opinion.

msft-jlange:
> Yes, but it does the equivalent of:
> ```rust
> if some_option.is_some() {
>     some_option.unwrap().some_method();
> } else {
>     Err(some_error)
> }
> ```

In at least one case, it's subtly different than that:
```rust
if some_option.is_some() {
    let params = process_protocol_parameters()?; // error possible on parameter checking but parameter checking should only be attempted if there is an object worth operating on
    some_option.unwrap().some_method(params)
} else {
    Err(some_error)
}
```
(cf. `apic_configure_vector()`, which I now see has yet another redundant check that should be cleaned up.)

Yes, it's possible to flow the `Option<Result<..>>` all the way to the top of the call chain, but that would mean that every new call that's added to the APIC protocol handler would require the author to remember to include the correct logic for flowing the optional result back in the right way.  Contrast that with the approach of having `is_some()` at the top of `apic_protocol_request()` which provides blanket error handling, making it impossible for any future call handler to get it wrong.  I really like code structures that make it impossible for anybody to make mistakes when extending existing code.

msft-jlange:
> I really like code structures that make it impossible for anybody to make mistakes when extending existing code.

As a case in point, I just tried putting together a commit that flowed the `Option<Result<..>>` all the way back up to `apic_protocol_request()` and it took me three tries to get the semantics correct.  I made a mistake in `apic_query_features()` when I failed to check for APIC emulation active before returning `Some(Ok(()))`, and I made a mistake in `apic_configure()` when I failed to check for APIC emulation active before calling through `SVSM_PLATFORM`.  If I, as the person who wrote this code, can make such serious mistakes so easily, then why would we have faith that somebody even less familiar with the APIC code would get new functions correct?  This is exactly what I mean when I talk about structuring the code to make a whole class of mistakes impossible.

00xc:
> > I really like code structures that make it impossible for anybody to make mistakes when extending existing code.
> 
> As a case in point, I just tried putting together a commit that flowed the `Option<Result<..>>` all the way back up to `apic_protocol_request()` and it took me three tries to get the semantics correct. I made a mistake in `apic_query_features()` when I failed to check for APIC emulation active before returning `Some(Ok(()))`, and I made a mistake in `apic_configure()` when I failed to check for APIC emulation active before calling through `SVSM_PLATFORM`. If I, as the person who wrote this code, can make such serious mistakes so easily, then why would we have faith that somebody even less familiar with the APIC code would get new functions correct? This is exactly what I mean when I talk about structuring the code to make a whole class of mistakes impossible.

I'd counter by saying this does make future mistakes possible by calling the functions that do `unwrap()` under the assumptions I detailed above. But okay, let's merge this and we can further improve it if we find a way.

00xc (approved):

00xc:
Maybe for another PR, but it seems these errors are not contemplated in the latest draft of the spec, or at least not with that name:

https://github.com/coconut-svsm/svsm/blob/f196bdcbe56a894d943bdafb8a98705d20a63074/kernel/src/protocols/apic.rs#L26-L27

msft-jlange:
> Maybe for another PR, but it seems these errors are not contemplated in the latest draft of the spec, or at least not with that name:

Good catch - I'll clean that up in this PR.

msft-jlange:
@00xc can you take another look?  I had to update because the previous version was unwrapping `Option<LocalApic>` causing a copy of the APIC, rather than unwrapping `Option<&LocalApic>` which would permit operating on the actual APIC state.  I'm inhibiting `Copy` on `LocalApic` so there are no additional surprises due to inappropriate copies.

00xc:
Please resolve conflicts and I'll give it a final look.

kernel/src/cpu/apic.rs
```diff
@@ -102,7 +102,10 @@ pub enum ApicError {
     ApicError,
 }
 
-#[derive(Default, Clone, Copy, Debug)]
+// This structure must never be copied because a silent copy will cause APIC
+// +// state to be lost.
```
00xc, line 106: Seems something went wrong with formatting here

msft-jlange: Fixed.

PR #387, opened 2024-06-19
cpu/percpu: store safe references to HV doorbell page and SVSM VMSA

00xc:
* Store a safe reference to the HV doorbell in the PerCpu
* Cleanups and add some documentation
* Store a safe reference to the SVSM VMSA in the PerCpu

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -12,8 +12,7 @@ use crate::address::{Address, PhysAddr, VirtAddr};
 use crate::cpu::apic::ApicError;
 use crate::cpu::idt::common::INT_INJ_VECTOR;
 use crate::cpu::tss::TSS_LIMIT;
-use crate::cpu::vmsa::init_guest_vmsa;
-use crate::cpu::vmsa::vmsa_mut_ref_from_vaddr;
+use crate::cpu::vmsa::{init_guest_vmsa, init_svsm_vmsa, vmsa_mut_ref_from_vaddr};
 use crate::cpu::LocalApic;
 use crate::error::SvsmError;
 use crate::locking::{LockGuard, RWLock, SpinLock};
@@ -124,8 +123,13 @@ impl VmsaRef {
         }
     }
 
+    pub fn vmsa(&self) -> &VMSA {
+        let ptr = self.vaddr.as_mut_ptr::<VMSA>();
+        unsafe { ptr.as_mut().unwrap() }
+    }
+
     #[allow(clippy::needless_pass_by_ref_mut)]
-    pub fn vmsa(&mut self) -> &mut VMSA {
+    pub fn vmsa_mut(&mut self) -> &mut VMSA {
         let ptr = self.vaddr.as_mut_ptr::<VMSA>();
         // SAFETY: this function takes &mut self, so only one mutable
         // reference to the underlying VMSA can exist.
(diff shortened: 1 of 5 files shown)


00xc:
Once this is merged I'll rework #348 to remove the remaining unsafe around page allocations

kernel/src/cpu/percpu.rs
```diff
@@ -108,31 +107,6 @@ impl PerCpuAreas {
     }
 }
 
-#[derive(Copy, Clone, Debug, Default)]
-pub struct VmsaRef {
```
msft-jlange, line 112: I'm a little reluctant to lose all of the information associated with the VMSA.  Today, the `guest_owned` field is not used, and the `paddr` is only needed in a couple of specific places where the VMSA is created or bound.  However, in the future, when we support a paravisor mode in addition to the SVSM mode, it will be possible for VMSAs to come both from guest-owned memory and from SVSM-owned memory, and I suspect we will start introducing code that needs to be able to tell the difference.  It's a little harder for me to envision a need for tracking the physical address, since I think it will probably only be needed in the same places that the physical address is used today (for which you have an answer) but if we need to keep a tuple around anyway, then retaining the physical address is a good idea.  We can always remove it if it is not needed when the code base becomes more complete, and that would be a lot easier than removing it now and then discovering later that we need to add it back.

msft-jlange: On closer examination, I see that `GuestVmsaRef` (which refers to guest VMPLs) and `VmsaRef` (which refers to VMPL 0) are different structures, and you're only proposing eliminating the `VmsaRef`.  The SVSM VMSA can never be guest-owned, so the `guest_owned` field is irrelevant, so I have no objection to removing the structure.

kernel/src/cpu/percpu.rs
```diff
@@ -311,13 +285,20 @@ impl PerCpuShared {
...
+/// `shared` field, a reference to which will be stored in [`PERCPU_AREAS`].
 #[derive(Debug)]
 pub struct PerCpu {
+    /// Per-CPU storage that might be accessed from other CPUs.
     shared: PerCpuShared,
 
     pgtbl: RefCell<PageTableRef>,
     tss: Cell<X86Tss>,
-    svsm_vmsa: Cell<Option<VmsaRef>>,
+    svsm_vmsa: Cell<Option<&'static VMSA>>,
```
msft-jlange, line 301: Once the SVSM VMSA has been allocated and initialized, it can never be changed again.  Perhaps `OnceCell` would be more appropriate here to eliminate the possibility that any code tries to change the VMSA.

It is also the case that once the VMSA has been initialized and communicated to the host, it can never even be accessed again by the SVSM, so I'm not sure the `PerCpu` structure even needs to track the pointer at all.  We need to make sure that when the VMSA page is allocated, it is never freed - so we want to effectively leak the allocation after it has been made, but it doesn't have to be tracked after that point.

kernel/src/cpu/percpu.rs
00xc: On top of that, if we ever need to introduce something that looks like `VmsaRef`, it will need to be something different, as the implementation I'm deleting here is completely unsafe.

kernel/src/cpu/percpu.rs
00xc: I'm not sure we gain much from removing the field from the `PerCpu`, especially if in the future we need to have access to the VMSA for some reason. I'll apply your suggestion of using a OnceCell though.

msft-jlange (approved):

PR #389, opened 2024-06-19
cpu/tss: fix TSS layout

00xc:
According to the AMD manual, there is a reserved 64 bit entry between the RSP and IST stack entries. Thus, separate it as such in the `X86Tss` struct. While we are at it, mark it as `packed(4)`, which is the correct alignment, and add a layout test.


kernel/src/cpu/tss.rs
```diff
@@ -8,15 +8,15 @@ use super::gdt::GDTEntry;
 use crate::address::VirtAddr;
 
 // IST offsets
-pub const _IST_INVALID: usize = 0;
-pub const IST_DF: usize = 1;
+pub const IST_DF: usize = 0;
```
joergroedel, line 11: Even if this is `0` now it still means that it needs to be referenced from the IDT by index `1`.

I suggest to leave the `IST_DF` index as is and add a method to `struct X86TSS` do correctly set the indices (allowed values are `0-7`)

joergroedel (changes requested):

kernel/src/cpu/tss.rs
00xc: Done as suggested

PR #390, opened 2024-06-19
igvmmeasure: Implement launch measurement calculation for SEV and SEV-ES

roy-hopkins:
This PR adds implements calculation of the expected launch measurement for IGVM files that target the SEV and SEV-ES platforms.

Although not needed for COCONUT-SVSM itself as this always requires SEV-SNP, these changes are required to support the testing of the QEMU IGVM patch series, which supports all flavours of SEV.

As part of this, the create previously used for the SHA-384 implementation has been replaced with the RustCrypto equivalent. As well as providing SHA-256 implementation as part of the same library (which the previous one did not), this provides consistency in use of RustCrypto with other cryptographic crates we are already using in COCONUT.

### DIFF
--- a/igvmmeasure/src/id_block.rs
+++ b/igvmmeasure/src/id_block.rs
@@ -39,12 +39,14 @@ pub struct SevIdBlockBuilder {
 
 impl SevIdBlockBuilder {
     pub fn build(igvm: &IgvmFile, measure: &IgvmMeasure) -> Result<Self, Box<dyn Error>> {
-        let ld = measure.digest();
         let compatibility_mask = get_compatibility_mask(igvm, IgvmPlatformType::SEV_SNP).ok_or(
             String::from("IGVM file is not compatible with the specified platform."),
         )?;
         let policy = get_policy(igvm, compatibility_mask)
             .ok_or(String::from("IGVM file does not contain a guest policy."))?;
+
+        let mut ld = [0u8; 48];
+        ld.copy_from_slice(measure.digest());
 
         Ok(Self {
             compatibility_mask,
--- a/igvmmeasure/src/igvm_measure.rs
+++ b/igvmmeasure/src/igvm_measure.rs
@@ -8,7 +8,8 @@ use std::error::Error;
 
 use igvm::snp_defs::SevVmsa;
 use igvm::{IgvmDirectiveHeader, IgvmFile};
-use igvm_defs::{IgvmPageDataFlags, IgvmPageDataType, PAGE_SIZE_4K};
+use igvm_defs::{IgvmPageDataFlags, IgvmPageDataType, IgvmPlatformType, PAGE_SIZE_4K};
+use sha2::{Digest, Sha256};
 use zerocopy::AsBytes;
 
 use crate::page_info::PageInfo;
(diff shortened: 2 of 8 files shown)


igvmmeasure/src/id_block.rs
```diff
@@ -39,13 +39,21 @@ pub struct SevIdBlockBuilder {
...
             .ok_or(String::from("IGVM file does not contain a guest policy."))?;
 
+        let mut ld = [0u8; 48];
+        let digest = measure.digest();
+        if digest.len() != ld.len() {
+            return Err("Invalid digest value calculated for ID block".into());
+        }
+        for (i, val) in measure.digest().iter().enumerate() {
+            ld[i] = *val;
+        }
```
stefano-garzarella, line 55: What about using `copy_from_slice`?

```suggestion
        ld.copy_from_slice(&digest);
```

igvmmeasure/src/igvm_measure.rs
```diff
@@ -128,14 +129,17 @@ pub struct IgvmMeasure {
     show_progress: bool,
     check_kvm: bool,
     native_zero: bool,
-    digest: [u8; 48],
+    digest_snp: [u8; 48],
+    digest_es: Sha256,
+    digest: Vec<u8>,
```
stefano-garzarella, line 134: I haven't looked in detail, but could we make it so that we only have `digest: Vec<u8>`?

igvmmeasure/src/id_block.rs
roy-hopkins: Thanks for that. I knew there must be an easier way!

igvmmeasure/src/igvm_measure.rs
roy-hopkins: I struggled with this as `digest_snp` and `digest_es` maintain the state during processing and `digest` contains the final measurement. But as I was writing this comment I realised that the state during processing can be maintained via function parameters. So I've now added `struct IgvmMeasureContext` which is passed between the measurement functions, allowing me to move them from `struct IgvmMeasure`.

igvmmeasure/src/id_block.rs
```diff
@@ -39,13 +39,15 @@ pub struct SevIdBlockBuilder {
...
     pub fn build(igvm: &IgvmFile, measure: &IgvmMeasure) -> Result<Self, Box<dyn Error>> {
-        let ld = measure.digest();
         let compatibility_mask = get_compatibility_mask(igvm, IgvmPlatformType::SEV_SNP).ok_or(
             String::from("IGVM file is not compatible with the specified platform."),
         )?;
         let policy = get_policy(igvm, compatibility_mask)
             .ok_or(String::from("IGVM file does not contain a guest policy."))?;
 
+        let mut ld = [0u8; 48];
+        ld.copy_from_slice(measure.digest());
```
stefano-garzarella, line 49: Should we put back the check on the length as you did in the previous version?

copy_from_slice() should panic if the two slices have different sizes.

stefano-garzarella (commented):
I just left a small comment about `copy_from_slice`, but the rest LGTM!

igvmmeasure/src/id_block.rs
stefano-garzarella: @roy-hopkins aaa, I see now, okay I'm fine with that ;-)

PR #392, opened 2024-06-20
Add test to check the pre-calculated launch measurement

stefano-garzarella:
This PR introduces a test to check that the launch measurement pre-calculated by `igvmmeasure` is the same as that returned by the VMPL0 attestation report at runtime.
To support that, this PR also adds a serial port (COM3) to allow the host to communicate with the test running in the VM. This could be used in the future for other tests as well.

Tested locally with:
```
QEMU=/path/to/coconut/qemu/build/qemu-system-x86_64 make test-in-svsm
```


roy-hopkins:
Just taking a look at this. At the moment I'm getting a test failure due to an incorrect measurement - probably due to my environment.

kernel/src/greq/services.rs
```diff
@@ -104,3 +104,34 @@ pub fn get_regular_report(buffer: &mut [u8]) -> Result<usize, SvsmReqError> {
...
+    use crate::serial::{test::serial_port, Terminal};
+    use alloc::vec;
+
+    #[test]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_snp_launch_measurement() {
+        let sp = serial_port();
+
+        // 0x01: return SEV-SNP pre-calculated launch measurement (48 bytes)
+        sp.put_byte(0x01);
```
00xc, line 122: Can we make an enum out of this value and place it somewhere like `kernel/src/testing.rs`? Let's also add some documentation there referencing the `scripts/test-in-svsm.sh` script. This way it will be easier for other developers to use this infrastructure.

Otherwise I fear this infrastructure will be hard to use by other developers.

kernel/src/greq/services.rs
```diff
@@ -104,3 +104,34 @@ pub fn get_regular_report(buffer: &mut [u8]) -> Result<usize, SvsmReqError> {
...
+        for byte in &mut expected_measurement {
+            *byte = sp.get_byte();
+        }
+
+        let mut buf = vec![0; size_of::<SnpReportResponse>()];
+        let size = get_regular_report(&mut buf).unwrap();
+        assert_eq!(size, buf.len());
+
+        let response = SnpReportResponse::try_from_as_ref(&buf).unwrap();
+        assert!(response.validate().is_ok());
```
00xc, line 134: This will display the error cause:

```suggestion
        response.validate().unwrap();
```

kernel/src/greq/services.rs
stefano-garzarella: > Can we make an enum out of this value and place it somewhere like `kernel/src/testing.rs`? Let's also add some documentation there referencing the `scripts/test-in-svsm.sh` script. This way it will be easier for other developers to use this infrastructure.

Good point!

Should I move also the `test::serial_port()` that I added in `kernel/src/serial.rs` there?
I'd like to put them close together and maybe `kernel/src/testing.rs` is the right place, WDYT?

kernel/src/greq/services.rs
stefano-garzarella: ok, I'll change.

kernel/src/greq/services.rs
00xc: Yep, that is also a good idea in my opinion

stefano-garzarella:
Strage, if I rebase on previous commit (126400b9c2c2cf0f43ecda9b129ee45577f6b493) it works, but on current main (fa92efca6c0c3b6a872db668522385dc00449e93), the test are not executed and the output is just the Stage2:
```
[Stage2] COCONUT Secure Virtual Machine Service Module (SVSM) Stage 2 Loader
[Stage2] Order-02: total pages:     1 free pages:     1
[Stage2] Order-03: total pages:     1 free pages:     1
[Stage2] Order-04: total pages:     0 free pages:     0
[Stage2] Order-05: total pages:     0 free pages:     0
[Stage2] Total memory: 108KiB free memory: 52KiB
[Stage2]   kernel_virtual_base   = 0xffffff8000000000
```

This happen only with the last patch of this PR applied. Any idea?

00xc:
Interesting, when I run in-SVSM tests on your branch I get the following:

```
[SVSM] test greq::pld_report::tests::test_ecdsa_p384_sha384_signature_offsets ...
[SVSM] test greq::pld_report::tests::test_snp_report_request_offsets ...
[SVSM] test greq::pld_report::tests::test_snp_report_response_offsets ...
[SVSM] test greq::services::tests::test_snp_launch_measurement ...
[SVSM] ERROR: Panic: CPU[0] panicked at kernel/src/greq/services.rs:137:9:
  left: [204, 126, 175, 255, 231, 55, 151, 54, 177, 236, 0, 166, 142, 10, 234, 39, 153, 46, 210, 245, 52, 168, 212, 128, 10, 166, 62, 87, 28, 45, 49, 82, 146, 129, 80, 12, 185, 94, 82, 58, 16, 248, 208, 72, 200, 24, 252, 208]
 right: [234, 244, 164, 113, 111, 124, 171, 218, 152, 54, 231, 36, 150, 137, 51, 7, 22, 58, 202, 168, 149, 134, 254, 136, 100, 43, 182, 202, 16, 192, 223, 85, 152, 218, 40, 209, 239, 169, 45, 90, 201, 173, 204, 254, 253, 14, 139, 38]
[SVSM] ---BACKTRACE---:
[SVSM]   [0xffffff80000f17ee]
[SVSM]   [0xffffff8000030d0d]
[SVSM]   [0xffffff800003d33e]
[SVSM]   [0xffffff800009dc61]
[SVSM]   [0xffffff800002b1c1]
[SVSM]   [0xffffff80000313a1]
[SVSM]   [0xffffff80000c8cb5]
[SVSM]   [0xffffff80000c4d5b]
[SVSM]   Invalid frame
[SVSM] ---END---
```

stefano-garzarella:
@00xc just to be sure, commit a9469eab5150a719e603b53b5fc5c2c4ced0ddcb ?

Now that you mention I tried to run the in-SVSM tests in release mode, and it worked (all tests passed included the new one), while in debug I still have https://github.com/coconut-svsm/svsm/pull/392#issuecomment-2221010020

stefano-garzarella:
BTW the mismatch could be related to the host kernel, (e.g. `debug_swap` kernel parameter). Do you have some param enabled or a different kernel in the host? (I'm using commit bc4de28e0cc1e7cf404af311c4052560aba355ba from coconut's linux)

00xc:
> @00xc just to be sure, commit [a9469ea](https://github.com/coconut-svsm/svsm/commit/a9469eab5150a719e603b53b5fc5c2c4ced0ddcb) ?

Yep.

> Now that you mention I tried to run the in-SVSM tests in release mode, and it worked (all tests passed included the new one), while in debug I still have [#392 (comment)](https://github.com/coconut-svsm/svsm/pull/392#issuecomment-2221010020)

It works for me in both modes.

However, I noticed that we do the following when booting with IGVM:

https://github.com/coconut-svsm/svsm/blob/fa92efca6c0c3b6a872db668522385dc00449e93/Makefile#L77

The important part being `--comport 3`, which results in setting I/O port 0x3e8 (COM3) as the debug port. I noticed that you also use this port in your implementation. Could it be that test I/O and console I/O are interfering with each other?

stefano-garzarella:
I just did `rustup update`:
```
   stable-x86_64-unknown-linux-gnu updated - rustc 1.79.0 (129f3b996 2024-06-10) (from rustc 1.78.0 (9b00956e5 2024-04-29))
  nightly-x86_64-unknown-linux-gnu updated - rustc 1.81.0-nightly (0c81f94b9 2024-07-10) (from rustc 1.80.0-nightly (8387315ab 2024-05-14))
```

And now it's working also here, but now I'm not sure if we have some hidden issue.

stefano-garzarella:
> However, I noticed that we do the following when booting with IGVM:
> 
> https://github.com/coconut-svsm/svsm/blob/fa92efca6c0c3b6a872db668522385dc00449e93/Makefile#L77
> 
> The important part being `--comport 3`, which results in setting I/O port 0x3e8 (COM3) as the debug port. I noticed that you also use this port in your implementation. Could it be that test I/O and console I/O are interfering with each other?

ehm, why we are setting  `--comport 3` ?

> 
> Note that using qemu's fw_cfg instead uses port 0x3f8 (COM1).

Without this PR, in `scripts/test-in-svsm.sh` we don't set the backend form COM3, so I'm not sure SVSM is really using it.

Should I use COM4?

stefano-garzarella:
Oh wait, `--comport 3` is only for `bin/coconut-hyperv.igvm`, for QEMU we are using the default (1).

00xc:
> Oh wait, `--comport 3` is only for `bin/coconut-hyperv.igvm`, for QEMU we are using the default (1).

Yes, that's what I meant :)

00xc:
> Should I use COM4?

Yes, pleas try using a different port for tests.

stefano-garzarella:
> > Should I use COM4?
> 
> Yes, please try using a different port for tests.

Okay, I'll do.

About the test failure in your env, can you check this:

> BTW the mismatch could be related to the host kernel, (e.g. `debug_swap` kernel parameter). Do you have some param enabled or a different kernel in the host? (I'm using commit bc4de28e0cc1e7cf404af311c4052560aba355ba from coconut's linux)

stefano-garzarella:
v3:
- replaced COM3 with COM4 to avoid conflicts with hyper-v
- added a new patch to refactor a bit the serial ports in `scripts/launch_guest.sh` 

v2:
- moved API to get the serial port used for tests in `kernel/src/testing.rs`
- added enum for tests and documentation

00xc:
> BTW the mismatch could be related to the host kernel, (e.g. `debug_swap` kernel parameter). Do you have some param enabled or a different kernel in the host? (I'm using commit bc4de28e0cc1e7cf404af311c4052560aba355ba from coconut's linux)

There is no such parameter in the cmdline. I'm not sure which host version we have, as it is an internal machine on which I did not install the kernel. `uname` says "6.8.1-1-coco".

roy-hopkins:
Apologies for the delay in joining this thread - I was away.

> There is no such parameter in the cmdline. I'm not sure which host version we have, as it is an internal machine on which I did not install the kernel. `uname` says "6.8.1-1-coco".

I've just checked on that same machine - the 6.8.1-1-coco kernel has DEBUG_SWAP _disabled_ which should match the IGVM file that is generated but I also see a mismatch in the attestation report. I can confirm however that your latest branch works on my SEV-SNP system for both debug and release builds and the attestation measurement matches.

stefano-garzarella:
v4:
- rebased on 6126750899eceb4ac2e0edda056f1344db986493
- add `$(IGVMMEASUREBIN)` dependency to `test-in-svsm` target in the Makefile

stefano-garzarella:
@00xc about the problem you were seeing on your machine, could it be related to https://github.com/coconut-svsm/svsm/issues/433 ?

Should we merge this PR to discover that issues?

Freax13:
This works for me. Tested on EPYC Milan. I'm using the latest KVM and QEMU patches mentioned in the docs.

joergroedel:
@stefano-garzarella Can you please downgrade a measurement mismatch from a fatal error to a warning? With that the PR is ready to merge and we can move on fixing the causes of the mismatches.

stefano-garzarella:
v5:
- rebased on 1fb6f4f82caacbf53e590ea6989bffb9ad0fdd18
- added `assert_eq_warn!()` macro
- call `assert_eq_warn!()` instead of `assert_eq!()` to avoid test failure till we fix the mismatch

stefano-garzarella:
> I am also running in a measurement mismatch error, `debug_swap` is off on my machine. The machine has a Milan CPU, any hints on what might be wrong?

What kernel is running your host?

> @stefano-garzarella Can you please downgrade a measurement mismatch from a fatal error to a warning? With that the PR is ready to merge and we can move on fixing the causes of the mismatches.

@joergroedel done ;-)

kernel/src/testing.rs
```diff
@@ -1,7 +1,65 @@
...
+                "Assertion warning failed at {}:{}:{}:\nassertion `left == rigth` failed\n left: {:?}\n right: {:?}",
+                file!(),
+                line!(),
+                column!(),
+                $left,
+                $right
+            );
+        }
+    };
+}
```
00xc, line 28: This evaluates `$left` and `$right` twice in the error case. I suggest using temporary variables. There is also a typo in the error message ("rigth").

```suggestion
#[macro_export]
macro_rules! assert_eq_warn {
    ($left:expr, $right:expr) => {
        {
            let left = $left;
            let right = $right;
            if left != right {
                log::warn!(
                    "Assertion warning failed at {}:{}:{}:\nassertion `left == right` failed\n left: {left:?}\n right: {right:?}",
                    file!(),
                    line!(),
                    column!(),
                );
            }
        }
    };
}
```

stefano-garzarella: Applied, thanks!

stefano-garzarella:
@00xc @joergroedel I think I addressed all the comments, can we merge this?

PR #396, opened 2024-06-26
svsm: permit early initialization of console logging

msft-jlange:
On platforms that do not require extensive platform setup (e.g. GHCB) for console output services, initialize the console logger as early as possible to facilitate earlier logging of initialization activity.

### DIFF
--- a/kernel/src/platform/native.rs
+++ b/kernel/src/platform/native.rs
@@ -5,16 +5,21 @@
 // Author: Jon Lange <jlange@microsoft.com>
 
 use crate::address::{PhysAddr, VirtAddr};
+use crate::console::init_console;
 use crate::cpu::cpuid::CpuidResult;
 use crate::cpu::msr::write_msr;
 use crate::cpu::percpu::PerCpu;
 use crate::error::SvsmError;
-use crate::platform::{IOPort, PageEncryptionMasks, PageStateChangeOp, SvsmPlatform};
+use crate::io::IOPort;
+use crate::platform::{PageEncryptionMasks, PageStateChangeOp, SvsmPlatform};
+use crate::serial::SerialPort;
 use crate::svsm_console::NativeIOPort;
 use crate::types::PageSize;
+use crate::utils::immut_after_init::ImmutAfterInitCell;
 use crate::utils::MemoryRegion;
 
 static CONSOLE_IO: NativeIOPort = NativeIOPort::new();
+static CONSOLE_SERIAL: ImmutAfterInitCell<SerialPort<'_>> = ImmutAfterInitCell::uninit();
 
 const APIC_MSR_ICR: u32 = 0x830;
 
@@ -34,8 +39,19 @@ impl Default for NativePlatform {
 }
 
 impl SvsmPlatform for NativePlatform {
-    fn env_setup(&mut self) {}
-    fn env_setup_late(&mut self) {}
+    fn env_setup(&mut self, debug_serial_port: u16) -> Result<(), SvsmError> {
+        // In the native platform, console output does not require the use of
+        // any platform services, so it can be initialized immediately.
+        CONSOLE_SERIAL
+            .init(&SerialPort::new(&CONSOLE_IO, debug_serial_port))
+            .map_err(|_| SvsmError::Console)?;
+        (*CONSOLE_SERIAL).init();
+        init_console(&*CONSOLE_SERIAL).map_err(|_| SvsmError::Console)
+    }
+
+    fn env_setup_late(&mut self, _debug_serial_port: u16) -> Result<(), SvsmError> {
+        Ok(())
+    }
 
     fn setup_percpu(&self, _cpu: &PerCpu) -> Result<(), SvsmError> {
         Ok(())
@@ -58,7 +74,7 @@ impl SvsmPlatform for NativePlatform {
 
     fn setup_guest_host_comm(&mut self, _cpu: &PerCpu, _is_bsp: bool) {}
 
-    fn get_console_io_port(&self) -> &'static dyn IOPort {
+    fn get_io_port(&self) -> &'static dyn IOPort {
         &CONSOLE_IO
     }
 
--- a/kernel/src/platform/snp.rs
+++ b/kernel/src/platform/snp.rs
@@ -5,12 +5,14 @@
 // Author: Jon Lange <jlange@microsoft.com>
 
 use crate::address::{PhysAddr, VirtAddr};
+use crate::console::init_console;
 use crate::cpu::cpuid::cpuid_table;
 use crate::cpu::percpu::{current_ghcb, PerCpu};
 use crate::error::ApicError::Registration;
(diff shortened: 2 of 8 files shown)


kernel/src/stage2.rs
```diff
@@ -101,19 +99,11 @@ fn setup_env(
...
-            platform.get_console_io_port(),
-            config.debug_serial_port(),
-        ))
-        .expect("console serial output already configured");
-    (*CONSOLE_SERIAL).init();
-    init_console(&*CONSOLE_SERIAL).expect("Console writer already initialized");
-
-    // Console is fully working now and any unsupported configuration can be
-    // properly reported.
+    // Complete initializtio of the platform.  After that point, the console
```
00xc, line 102:
```suggestion
    // Complete initialization of the platform. After that point, the console
```

kernel/src/platform/tdp.rs
```diff
@@ -32,9 +38,15 @@ impl Default for TdpPlatform {
...
-
-    fn env_setup_late(&mut self) {}
+    fn env_setup(&mut self, _debug_serial_port: u16) {}
+
+    fn env_setup_late(&mut self, debug_serial_port: u16) {
+        CONSOLE_SERIAL
+            .init(&SerialPort::new(&CONSOLE_IO, debug_serial_port))
+            .expect("console serial output already configured");
+        (*CONSOLE_SERIAL).init();
+        init_console(&*CONSOLE_SERIAL).expect("Console writer already initialized");
```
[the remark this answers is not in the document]
msft-jlange: If this function is to return an error, it should be of type `SvsmError` because the same function in the SNP platform can do multiple things (not just console init) so the error needs to be rich enough to describe the failure.  And if the caller is to panic, the panic really should include the error that is returned to be able to explain what failed.  However, we do not seem to have a default formatter for `SvsmError`.  Is there a simple way to derive one, or is it time to take the plunge and implement a manual formatter for `SvsmError` and all of its embedded types so that we can create appropriate panic messages?  Or, in the interest of binary size, should we just ignore all of this pretty-printing?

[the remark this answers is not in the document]
00xc: I think we can just add a new variant to `SvsmError` to include `ImmutAfterInitError`. Error variants should be descriptive enough, along with the string in the `expect()` call (`expect` will display the error that caused the panic). Either that or we return `ImmutAfterInitResult` for now and change this later.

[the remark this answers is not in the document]
msft-jlange: > `expect` will display the error that caused the panic

This was the part I was missing, because I tried to introduce an `SvsmError` into a formatted panic string and was told that the default formatter was not capable of displaying `SvsmError`.  This makes me wonder how `expect()` can do any better, but perhaps you know some secret that I don't.

In any case, if `expect()` will format correctly, then I think my latest version will do the trick.  I don't mind introducing `SvsmError::Console` which, at least for the time being, will indicate that the failure was due to the `ImmutAfterInit` intialization.  At least in this case, it can distinguish from the `GhcbMsrError` that can occur during SNP platform initialization.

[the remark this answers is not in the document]
00xc: So you will be able to print the `SvsmError` with `{:?}` because of the `Debug` derive implementation, but not with `{}`, because we don't have a `Display` implementation. `expect()` will use the `Debug` implementation anyway, plus I don't think we need anything more elaborate, at least for now.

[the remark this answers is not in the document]
00xc: > In any case, if `expect()` will format correctly, then I think my latest version will do the trick. I don't mind introducing `SvsmError::Console` which, at least for the time being, will indicate that the failure was due to the `ImmutAfterInit` intialization. At least in this case, it can distinguish from the `GhcbMsrError` that can occur during SNP platform initialization.

Sounds like a good idea.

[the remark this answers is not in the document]
msft-jlange: Thanks for the explanation.  In that case, I think the current draft is good.

kernel/src/platform/native.rs
```diff
@@ -34,8 +39,19 @@ impl Default for NativePlatform {
...
 
 impl SvsmPlatform for NativePlatform {
-    fn env_setup(&mut self) {}
-    fn env_setup_late(&mut self) {}
+    fn env_setup(&mut self, debug_serial_port: u16) -> Result<(), SvsmError> {
+        // In the native platform, console output does not require the use of
+        // any platform services, so it can be initialized immediately.
+        CONSOLE_SERIAL
+            .init(&SerialPort::new(&CONSOLE_IO, debug_serial_port))
+            .expect("console serial output already configured");
```
00xc, line 47: Could we also return an error here? The caller will panic anyway and this way we have less implicit panics.

kernel/src/platform/snp.rs
```diff
@@ -43,13 +47,20 @@ impl Default for SnpPlatform {
...
+    fn env_setup(&mut self, _debug_serial_port: u16) -> Result<(), SvsmError> {
         sev_status_init();
+        Ok(())
     }
 
-    fn env_setup_late(&mut self) {
+    fn env_setup_late(&mut self, debug_serial_port: u16) -> Result<(), SvsmError> {
+        CONSOLE_SERIAL
+            .init(&SerialPort::new(&CONSOLE_IO, debug_serial_port))
+            .expect("console serial output already configured");
```
00xc, line 58: Same thing here

kernel/src/platform/tdp.rs
```diff
@@ -32,9 +38,17 @@ impl Default for TdpPlatform {
...
-    fn env_setup(&mut self) {}
+    fn env_setup(&mut self, _debug_serial_port: u16) -> Result<(), SvsmError> {
+        Ok(())
+    }
 
-    fn env_setup_late(&mut self) {}
+    fn env_setup_late(&mut self, debug_serial_port: u16) -> Result<(), SvsmError> {
+        CONSOLE_SERIAL
+            .init(&SerialPort::new(&CONSOLE_IO, debug_serial_port))
+            .expect("console serial output already configured");
```
00xc, line 48: And here

kernel/src/sev/ghcb.rs
```diff
@@ -73,6 +73,8 @@ macro_rules! ghcb_setter {
 
 #[derive(Clone, Copy, Debug)]
 pub enum GhcbError {
+    // Failed to initialize the GHCB
+    Initialization,
```
00xc, line 77: I see we introduced this but we're not using it anywhere.

kernel/src/platform/native.rs
msft-jlange: Agreed.

kernel/src/sev/ghcb.rs
msft-jlange: That was before I noticed that we have `SvsmError::GhcbMsrError`.  I'll remove this.

PR #397, opened 2024-06-28
stage1: add TD Partitioning support

peterfang:
This set of patches enables TDX in stage1 trampoline-only mode. This includes bootstrapping from the reset vector in protected non-paged mode, performing sanity checks, and setting up the ESP register.

### DIFF
--- a/stage1/reset.S
+++ b/stage1/reset.S
@@ -13,6 +13,7 @@
 
 	.code16gcc
 	.section .init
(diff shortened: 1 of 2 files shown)


msft-jlange:
Something is not working correctly for me.  When I run `make igvm`, the IGVM file that is produced does not include any data at GPA `FFFFF000`.  At least, this is true for the Hyper-V IGVM file.

msft-jlange (approved):

PR #400, opened 2024-07-01
greq/msg: simplify `SnpGuestRequestMsgHdr::get_aad_slice()`

00xc:
Now that the `offset_of` macro is stable, simply use it to get the desired offset for the slice. Simplify also the slice creation by simply casting a `*const Self` to a `*const u8` and calling `slice::from_raw_parts` with the adequate size.


cclaudio (approved):
Looks good to me, thanks.

PR #403, opened 2024-07-02
error: subsume ApicError into SvsmError

msft-jlange:
This change removes `ApicError` as a type associated with APIC emulation and defines it as a subordinate type of `SvsmError`, with variants describing the types of errors that may be associated with APIC handling (emulation or otherwise).  This permits unification of error types across different components so they can all use `SvsmError`.

### DIFF
--- a/kernel/src/error.rs
+++ b/kernel/src/error.rs
@@ -28,6 +28,17 @@ use crate::sev::SevSnpError;
 use crate::task::TaskError;
 use elf::ElfError;
 
+/// Errors related to APIC handling.  These may originate from multiple
+/// layers in the system.
+#[derive(Clone, Copy, Debug)]
+pub enum ApicError {
+    /// An error related to APIC emulation.
+    Emulation,
+
+    /// An error related to APIC registration.
+    Registration,
+}
+
 /// A generic error during SVSM operation.
 #[derive(Clone, Copy, Debug)]
 pub enum SvsmError {
@@ -72,11 +83,17 @@ pub enum SvsmError {
     /// The operation is not supported.
     NotSupported,
     /// Generic errors related to APIC emulation.
-    Apic,
+    Apic(ApicError),
 }
 
 impl From<ElfError> for SvsmError {
     fn from(err: ElfError) -> Self {
         Self::Elf(err)
+    }
+}
+
+impl From<ApicError> for SvsmError {
+    fn from(err: ApicError) -> Self {
+        Self::Apic(err)
     }
 }
--- a/kernel/src/protocols/apic.rs
+++ b/kernel/src/protocols/apic.rs
@@ -19,8 +19,6 @@ pub const APIC_PROTOCOL: u32 = 3;
 pub const APIC_PROTOCOL_VERSION_MIN: u32 = 1;
 pub const APIC_PROTOCOL_VERSION_MAX: u32 = 1;
 
-const SVSM_ERR_APIC_CANNOT_REGISTER: u64 = 0;
-
 fn apic_query_features(params: &mut RequestParams) -> Result<(), SvsmReqError> {
     // No features are supported beyond the base feature set.
     params.rcx = 0;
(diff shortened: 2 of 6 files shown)


kernel/src/error.rs
```diff
@@ -28,6 +28,17 @@ use crate::sev::SevSnpError;
 use crate::task::TaskError;
 use elf::ElfError;
 
+/// Errors related to APIC handling.  These may originate from multiple
+/// layers in the system.
+#[derive(Clone, Copy, Debug)]
+pub enum ApicError {
+    /// An error relted to APIC emulation.
```
00xc, line 35:
```suggestion
    /// An error related to APIC emulation.
```

kernel/src/error.rs
```diff
@@ -72,7 +83,7 @@ pub enum SvsmError {
     /// The operation is not supported.
     NotSupported,
     /// Generic errors related to APIC emulation.
-    Apic,
+    Apic(ApicError),
```
00xc, line 86: Let's also implement `From<ApicError> for SvsmError` to follow the convention in the rest of the codebase.

kernel/src/protocols/apic.rs
```diff
@@ -47,7 +45,7 @@ fn apic_configure(params: &RequestParams) -> Result<(), SvsmReqError> {
             // this will not cause any change to the state of the current CPU.
             return match platform.change_apic_registration_state(true) {
                 Ok(_) => Ok(()),
-                Err(_) => Err(SvsmReqError::protocol(SVSM_ERR_APIC_CANNOT_REGISTER)),
+                Err(e) => Err(SvsmReqError::from(e)),
             };
```
00xc, line 49: We can now take advantage of the `From` implementation by using the `?` operator:

```rust
                platform.change_apic_registration_state(true)?;
                return Ok(());
```

kernel/src/protocols/apic.rs
```diff
@@ -71,7 +69,7 @@ fn apic_read_register(params: &mut RequestParams) -> Result<(), SvsmReqError> {
     }
     let value = cpu
         .read_apic_register(params.rcx)
-        .map_err(|_| SvsmReqError::invalid_parameter())?;
+        .map_err(SvsmReqError::from)?;
```
00xc, line 72: Same here, the `?` operator will already use the `From` conversion, so this should just work:

```rust
    let value = cpu.read_apic_register(params.rcx)?;
```

kernel/src/protocols/apic.rs
```diff
@@ -82,7 +80,7 @@ fn apic_write_register(params: &RequestParams) -> Result<(), SvsmReqError> {
         return Err(SvsmReqError::invalid_request());
     }
     cpu.write_apic_register(params.rcx, params.rdx)
-        .map_err(|_| SvsmReqError::invalid_parameter())
+        .map_err(SvsmReqError::from)
```
00xc, line 83: Same here

kernel/src/protocols/errors.rs
```diff
@@ -75,6 +84,7 @@ impl From<SvsmError> for SvsmReqError {
             // to the guest as protocol-specific errors.
             SvsmError::SevSnp(e) => Self::protocol(e.ret()),
             SvsmError::InvalidAddress => Self::invalid_address(),
+            SvsmError::Apic(e) => Self::translate_apic_error(e),
```
00xc, line 87: Any reason we are implementing `translate_apic_error()` instead of just inlining the conversion?

```suggestion
            SvsmError::Apic(e) => match e {
                ApicError::Emulation => Self::invalid_parameter(),
                ApicError::Registration => Self::protocol(SVSM_ERR_APIC_CANNOT_REGISTER),
            }
```

We can also do:

```suggestion
            SvsmError::Apic(ApicError::Emulation) => Self::invalid_parameter(),
            SvsmError::Apic(ApicError::Registration) => {
                Self::protocol(SVSM_ERR_APIC_CANNOT_REGISTER)
            }
```

00xc (commented):
Looks good, just left a few style suggestions below.

kernel/src/protocols/apic.rs
msft-jlange: I didn't realize `from` would be automatically invoked by the compiler to attempt type conversion - that makes things a lot nicer.

kernel/src/protocols/errors.rs
msft-jlange: The thinking was that someday `ApicError` would grow to include new variants and that we didn't want `SvsmReqError::from` to be on a path to grow to an unmanageably large `match` statement.  Keeping different error types in different functions makes it a little easier to figure out how to organize the code as it grows, and the compiler will inline everything anyway (not that error handling needs to be optimized).  But I can go with your suggestion for now since the list is small.

00xc: With my second suggestion, if new `ApicError` variants are added, they will result in the base case, so no further changes would be needed. With the first suggestion a compiler error will be emitted asking for all variants to be considered.

00xc (approved):

Issue #404, opened 2024-07-04
openssl red-zone is corrupted by interrupt handling


jyao1:
Currently, we are using vTPM feature in SVSM, but vTPM sometimes failed with stack crash.

The root cause is that the openssl code enabled `red-zone` feature (See [1]). Take an openssl function that triggers the page fault as example, %rsp is not updated after entering the function BN_ucmp:
```
int BN_ucmp(const BIGNUM *a, const BIGNUM *b)
{
ffffff800002c0e5:	55                   	push   %rbp
ffffff800002c0e6:	48 89 e5             	mov    %rsp,%rbp
ffffff800002c0e9:	48 89 7d c8          	mov    %rdi,-0x38(%rbp)
ffffff800002c0ed:	48 89 75 c0          	mov    %rsi,-0x40(%rbp)
    BN_ULONG t1, t2, *ap, *bp;

    bn_check_top(a);
    bn_check_top(b);

    i = a->top - b->top;
ffffff800002c0f1:	48 8b 45 c8          	mov    -0x38(%rbp),%rax
ffffff800002c0f5:	8b 50 08             	mov    0x8(%rax),%edx
ffffff800002c0f8:	48 8b 45 c0          	mov    -0x40(%rbp),%rax
ffffff800002c0fc:	8b 48 08             	mov    0x8(%rax),%ecx
ffffff800002c0ff:	89 d0                	mov    %edx,%eax
ffffff800002c101:	29 c8                	sub    %ecx,%eax
ffffff800002c103:	89 45 fc             	mov    %eax,-0x4(%rbp)
    if (i != 0)
ffffff800002c106:	83 7d fc 00          	cmpl   $0x0,-0x4(%rbp)
ffffff800002c10a:	74 08                	je     ffffff800002c114 <BN_ucmp+0x2f>
        return i;
ffffff800002c10c:	8b 45 fc             	mov    -0x4(%rbp),%eax
ffffff800002c10f:	e9 8c 00 00 00       	jmp    ffffff800002c1a0 <BN_ucmp+0xbb>
    ap = a->d;
ffffff800002c114:	48 8b 45 c8          	mov    -0x38(%rbp),%rax
ffffff800002c118:	48 8b 00             	mov    (%rax),%rax
ffffff800002c11b:	48 89 45 f0          	mov    %rax,-0x10(%rbp)
    bp = b->d;
ffffff800002c11f:	48 8b 45 c0          	mov    -0x40(%rbp),%rax
[...]

PR #406, opened 2024-07-04
libmstpm: disable red zone

00xc:
As reported by Jiewen Yao, OpenSSL is built using a stack red zone, which causes issues when returning from an interrupt. Disable it in the OpenSSL build flags, as well as in libmstpm and libcrt.

Fixes: #404


00xc:
@jyao1 could you check if this fixes the issue you reported? Thanks!

jyao1:
Thank you!

We also add `-mno-red-zone` to https://github.com/coconut-svsm/svsm/blob/main/libmstpm/Makefile#L106.

00xc:
> Thank you!
> 
> We also add `-mno-red-zone` to https://github.com/coconut-svsm/svsm/blob/main/libmstpm/Makefile#L106.

Added this to libmstpm and also to libcrt

jyao1 (approved):

jyao1:
looks great

PR #410, opened 2024-07-15
SSE and FPU support

vsntk18:
This PR entails following changes:
1. Enable the usage of SSE/FPU instructions in SVSM.
2. xstate management while switching between tasks.
3.  A couple of unit tests check the sse instruction usage and proper xstate management.

I will submit the related exception handling in another PR.


kernel/src/mm/vm/mapping/xsave_area.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::mm::{PageRef, PAGE_SIZE};
+
+#[derive(Debug)]
+pub struct XSaveArea {
+    pg: PageRef,
+}
+
+impl XSaveArea {
+    pub fn new() -> Result<Self, SvsmError> {
+        let pg = allocate_file_page_ref()?;
```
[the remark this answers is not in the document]
joergroedel: A file page is freed automatically when the ref goes away.

joergroedel, line 21: The required size for the xsave area is variable-sized and needs to be calculated. It can extend with new features, so we can not assume that one page is enough.

See AMD Architecture Programmers Manual Vol. 2 Section 11.5.3 for how to calculate the correct size.

joergroedel (changes requested):
This goes into the right direction, thanks for the work! I marked some desired changes inline.

kernel/src/cpu/sse.rs
```diff
@@ -0,0 +1,80 @@
...
+    let res = CpuidResult::get(0x00000001, 0);
+    (res.edx & (1 << CPUID_EDX_SSE1)) != 0
+}
+
+fn legacy_sse_enable() {
+    if legacy_sse_supported() {
+        cr4_osfxsr_enable();
+        cr0_sse_enable();
+    } else {
+        log::info!("Legacy SSE instructions not supported by this hardware");
```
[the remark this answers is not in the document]
vsntk18: ok, initially I was thinking may be SSE/FPU support is optional.

kernel/src/cpu/sse.rs
```diff
@@ -0,0 +1,80 @@
...
+           options(att_syntax))
+    }
+
+    // set bits [0-2] in XCR0 to enable extended SSE
+    xsse_val |= 0x7;
+    unsafe {
+        asm!("xsetbv", in("eax") xsse_val, in("ecx") ecx_val, in("edx") edx_val,
+           options(att_syntax))
+    }
+}
```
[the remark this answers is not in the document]
vsntk18: That's neat.  Thanks for the suggestion.

kernel/src/task/tasks.rs
```diff
@@ -112,10 +113,15 @@ impl TaskSchedState {
...
 }
 
+// Layout of this struct is important. Offset of the members
+// have to be kept the same. Add members to the end if necessary.
 #[repr(C)]
 pub struct Task {
     pub rsp: u64,
 
+    /// XSave area address for the task
+    pub xsa_addr: u64,
```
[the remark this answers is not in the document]
vsntk18: Actually this value is used in the inline assembly, just like _rsp_.

kernel/src/cpu/sse.rs
```diff
@@ -0,0 +1,80 @@
...
+
+fn extended_sse_enable() {
+    if extended_sse_supported() && xsave_supported() {
+        cr4_xsave_enable();
+        xcr0_set();
+        check_xsave_features();
+    } else {
+        log::info!("XSAVE is not supported by this hardware");
```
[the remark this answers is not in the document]
vsntk18: yes, without xsave at least some part of the state needs to be manually saved and restored. I expect xsave to be available on modern cpus.

kernel/src/cpu/sse.rs
```diff
@@ -0,0 +1,80 @@
...
+
+fn xcr0_set() {
+    let mut xsse_val: u32;
+    let ecx_val = 0;
+    let mut edx_val: u32;
+
+    unsafe {
+        asm!("xgetbv", out("eax") xsse_val, in("ecx") ecx_val, out("edx") edx_val,
+           options(att_syntax))
+    }
```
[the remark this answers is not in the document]
vsntk18: hmm,  where else do you expect XCR register value to be used?   This function is called only once during the initialization of the cores.

kernel/src/cpu/sse.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    // set bits [0-2] in XCR0 to enable extended SSE
+    xsse_val |= 0x7;
```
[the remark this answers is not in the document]
vsntk18: ok.

kernel/src/mm/vm/mapping/xsave_area.rs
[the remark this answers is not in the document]
00xc: Right, but as you mention somewhere else, the xsave area can be larger than a page. I suggest using the new PageBox API once #420 is merged.

kernel/src/task/tasks.rs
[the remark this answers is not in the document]
00xc: They should be ABI-compatible for 64 bit systems, but we can leave as-is for consistency.

kernel/src/cpu/sse.rs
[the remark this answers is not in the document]
joergroedel: There is no reason to handle hardware without XSAVE support, as all modern CPUs have that. The feature not being enumerated should case a boot panic.

kernel/src/cpu/sse.rs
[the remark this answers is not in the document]
joergroedel: Okay, in this case you can leave it as is. We can live without caching the value.

kernel/src/mm/vm/mapping/xsave_area.rs
joergroedel: Yes, but also: for the current size of 832 bytes, 4 states would fit into one page. So I think it makes sense to allocate the memory via the global allocator. This handles both cases, small and very big xsave areas.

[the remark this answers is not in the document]
vsntk18: When xsave area spans multiple pages, multiple file pages are allocated. All file pages are deallocated  when they go out of reference. Same technique is used for task stacks as well.

Do you think PageBox is better in this case?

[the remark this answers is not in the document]
00xc: Unless you need reference counting, I suggest using `PageBox`, as the backing memory is also freed on drop. Plus, the bounds of the backing memory is kept track of, meaning memory corruption is less likely.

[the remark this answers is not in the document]
joergroedel: Hmm, given that the XSAVE area is about 800-900 bytes currently, a full page is probably not needed. I think it is better to use a `Box` instead. That will also handle the case of an XSAVE area that is bigger than one page and does not waste as much memory with the current size.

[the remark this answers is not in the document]
vsntk18: > Right, but as you mention somewhere else, the xsave area can be larger than a page. I suggest using the new PageBox API once #420 is merged.

can the memory allocated via PageBox be accessed globally ( both by kernel and tasks)? I got a page fault while accessing this memory to do xsave/xrstore.

vsntk18:
Addressed the review comments.

kernel/src/mm/vm/mapping/xsave_area.rs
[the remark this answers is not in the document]
vsntk18: After a discussion with Joerg,  decided to go with PageBox until we have our own smart pointers in SVSM.

kernel/src/cpu/sse.rs
```diff
@@ -0,0 +1,115 @@
...
+const CPUID_EDX_SSE1: u32 = 25;
+const CPUID_ECX_XSAVE: u32 = 26;
+const CPUID_EAX_XSAVEOPT: u32 = 0;
+const XCR0_X87_ENABLE: u64 = 0x1;
+const XCR0_SSE_ENABLE: u64 = 0x2;
+const XCR0_YMM_ENABLE: u64 = 0x4;
+
+// Check if at least SSE1 is supported
+fn legacy_sse_supported() -> bool {
+    let Some(res) = cpuid_table_raw(1, 0, 0, 0) else {
```
joergroedel, line 21: This code needs to use the `cpuid` instruction directly to be compatible with the native and TD platform targets. I just realized there is no wrapper yet for `cpuid`, so please add one and use it for any FPU-related CPUID queries. This also affects other places in this PR.

kernel/src/cpu/sse.rs
```diff
@@ -0,0 +1,115 @@
...
+    res.ecx
+}
+
+fn extended_sse_enable() {
+    if extended_sse_supported() && xsave_supported() {
+        cr4_xsave_enable();
+        xcr0_set();
+    } else {
+        panic!("XSAVE is not supported by this hardware");
+    }
```
joergroedel, line 76:
```suggestion
    if !extended_sse_supported() {
        panic!("Extended SSE is not supported by this hardware");
    }
    
    if !xsave_supported() {
        panic!("XSAVE is not supported by this hardware");
    }
    
    cr4_xsave_enable();
    xcr0_set();    
```

Make the panic messages a bit more concise.

kernel/src/task/schedule.rs
```diff
@@ -298,6 +300,10 @@ unsafe fn task_pointer(taskptr: TaskPointer) -> *const Task {
 unsafe fn switch_to(prev: *const Task, next: *const Task) {
     let cr3: u64 = unsafe { (*next).page_table.lock().cr3_value().bits() as u64 };
 
+    if !prev.is_null() {
+        sse_save_context(u64::from((*prev).xsa.vaddr()));
+    }
+
```
joergroedel, line 306: This can happen in `schedule()`.

kernel/src/task/schedule.rs
```diff
@@ -361,8 +367,13 @@ pub fn schedule() {
 
     drop(guard);
 
-    // We're now in the context of the new task. If the previous task had terminated
-    // then we can release it's reference here.
+    // We're now in the context of the new task.
+    let cur = current_task();
+    unsafe {
+        sse_restore_context(u64::from(cur.xsa.vaddr()));
+    }
```
joergroedel, line 374: This needs to be moved into the `if let Some((current, next)) = work {` branch. Otherwise the code will load an old FPU context when no context switch is happening.

kernel/src/task/tasks.rs
```diff
@@ -501,8 +525,11 @@ fn setup_new_task() {
     }
 }
 
-extern "C" fn run_kernel_task(entry: extern "C" fn()) {
+extern "C" fn run_kernel_task(entry: extern "C" fn(), xsa_addr: u64) {
     setup_new_task();
+    unsafe {
+        sse_restore_context(xsa_addr);
+    }
```
joergroedel, line 532: This needs to be in `setup_new_task()`.

kernel/src/task/tasks.rs
```diff
@@ -176,7 +184,9 @@ impl Task {
         let vm_kernel_range = VMR::new(SVSM_PERTASK_BASE, SVSM_PERTASK_END, PTEntryFlags::empty());
         vm_kernel_range.initialize()?;
 
-        let (stack, raw_bounds, rsp_offset) = Self::allocate_ktask_stack(cpu, entry)?;
+        let xsa = Self::allocate_xsave_area();
+        let xsa_addr = u64::from(xsa.vaddr()) as usize;
+        let (stack, raw_bounds, rsp_offset) = Self::allocate_ktask_stack(cpu, entry, xsa_addr)?;
```
joergroedel, line 189: The same needs to happen for user tasks.

joergroedel (changes requested):
Thanks for updating the PR! I left a couple of review comments.

kernel/src/cpu/sse.rs
vsntk18: How about CpuidResult::get() which is also being used by platform for ex: kernel/src/platform/native.rs?

kernel/src/cpu/sse.rs
vsntk18: Sure, truncated them a bit.

kernel/src/task/schedule.rs
vsntk18: moved to schedule()

kernel/src/task/schedule.rs
vsntk18: Ah, right!  good point.

kernel/src/task/tasks.rs
vsntk18: done!

kernel/src/task/tasks.rs
vsntk18: added to user tasks as well.

vsntk18:
|Addressed review comments.

vsntk18:
Rebase done.

msft-jlange:
What are the rules for FP/SSE in interrupt handlers?  Is the interrupt dispatch logic required to save/restore FPU state to prevent reentrant corruption?

kernel/src/cpu/sse.rs
joergroedel: Yes, that is the better option. I have seen you updated it that way.

kernel/src/task/schedule.rs
```diff
@@ -353,16 +355,20 @@ pub fn schedule() {
         unsafe {
             let a = task_pointer(current);
             let b = task_pointer(next);
+            sse_save_context(u64::from((*a).xsa.vaddr()));
 
             // Switch tasks
             switch_to(a, b);
+
+            // We're now in the context of the new task.
+            sse_restore_context(u64::from((*a).xsa.vaddr()));
```
joergroedel, line 364: The context needs to be loaded from `*b` instead of `*a`, no?

joergroedel:
> What are the rules for FP/SSE in interrupt handlers? Is the interrupt dispatch logic required to save/restore FPU state to prevent reentrant corruption?

None of this is handled yet in this PR. The plan is that the kernel is in general not allowed to use the FPU at all. There will be calls to enable/disable FPU usage, that would save/restore the state, but will not be possible to be used in a nested way. This means that IRQ handlers are in general not allowed to use the FPU.

kernel/src/task/schedule.rs
joergroedel: Okay, I was wrong, as this part of the function runs once process A is scheduled back in, so restoring the FPU context from `*a` is correct.

joergroedel (approved):
The changes look ready for merging to me. There are still a few rough edges, but some of them need a better kernel initialization infrastructure first. So lets move on with this now. Thanks for taking on and driving this work forward, @vsntk18

PR #412, opened 2024-07-17
cpu/apic: multiple trivial cleanups

00xc:
* Add documentation for several functions.
* Remove the `pub` qualifier from functions that are not used outside the APIC module.
* Clean up formatting in some comments
* Make a few variables immutable.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -118,8 +118,8 @@ pub struct LocalApic {
 }
 
 impl LocalApic {
-    pub fn new() -> Self {
-        LocalApic {
+    pub const fn new() -> Self {
+        Self {
             irr: [0; 8],
             allowed_irr: [0; 8],
             isr_stack_index: 0,
@@ -194,7 +194,7 @@ impl LocalApic {
         }
 
         // If a lazy EOI is pending, then check to see whether an EOI has been
-        // requested by the guest.  Note that if a lazy EOI was dismissed
+        // requested by the guest. Note that if a lazy EOI was dismissed
         // above, the guest lazy EOI flag need not be cleared here, since
         // dismissal of any interrupt above will require reprocessing of
         // interrupt state prior to guest reentry, and that reprocessing will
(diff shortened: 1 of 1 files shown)


msft-jlange:
I'm not sure we want to spend much effort converting the number of spaces after periods in comments.  The VIM editor, for example (which I use) defaults to two spaces after a period when joining two comment lines together.  I don't object to changes to these comments but we could easily wind up wasting a lot of time with comment editing if we try to get strict about requiring specific rules.

00xc:
> I'm not sure we want to spend much effort converting the number of spaces after periods in comments. The VIM editor, for example (which I use) defaults to two spaces after a period when joining two comment lines together. I don't object to changes to these comments but we could easily wind up wasting a lot of time with comment editing if we try to get strict about requiring specific rules.

I just did a quick search & replace, I don't think we should enforce it as a rule.

kernel/src/cpu/apic.rs
```diff
@@ -249,6 +249,8 @@ impl LocalApic {
         Some(calling_area)
     }
 
+    /// Attempts to deliver the specified IRQ into the specified guest CPU
```
msft-jlange, line 252: We should distinguish this from queueing an interrupt.  The key here is that this function attempts to deliver the specified interrupt so it will be observed immediately upon the next guest entry, and will fail if the guest state is such that interrupt delivery would need to be deferred (e.g. due to disabled interrupts or interrupt priority).  I'm not sure how to state this succinctly in a comment, but it would be helpful.

msft-jlange:
@00xc, this has been pending update for a while.  Would you like me to take over the PR and complete it, or did you want to continue to update it yourself?

kernel/src/cpu/apic.rs
00xc: Updated the doc text to include this.

00xc:
> @00xc, this has been pending update for a while. Would you like me to take over the PR and complete it, or did you want to continue to update it yourself?

Should be good to go now

msft-jlange:
I still don't like all of the double- to single-space changes because it introduces a lot of noise into the diff (it makes it harder to see the substantive changes because they're mixed in with a lot of whitespace changes) but the rest of the change looks good.

msft-jlange (approved):

PR #415, opened 2024-07-19
Prepare for update to COCONUT linux host and QEMU 9.0

roy-hopkins:
This PR consists of commits that are required to launch and run tests when hosted on a system with the following kernel and QEMU versions:

Linux-6.10-rc2:
https://github.com/coconut-svsm/linux/pull/6

QEMU 9.0:
https://github.com/coconut-svsm/qemu/pull/15

The intercept behaviour of some MSRs has changed since the previous COCONUT linux version, hence tests needed to be updated. Also, the QEMU command line has changed requiring a change in the launch script. The script is still compatible with early versions of QEMU/kernel.

I've also added an extra commit that fixes failing test-in-svsm cases when run on a Genoa system.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -510,14 +510,12 @@ mod tests {
         assert_eq!(apic_base & APIC_BASE_PHYS_ADDR_MASK, APIC_DEFAULT_PHYS_BASE);
     }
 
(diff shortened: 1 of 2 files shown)


joergroedel:
Obsoleted by efforts to move to kernel 6.11 and QEMU 9.1.

PR #417, opened 2024-07-19
cpu/tlb: do not broadcast per-cpu TLB flushes

msft-jlange:
VM ranges that are specific to a single CPU do not require TLB invalidations to be broadcast to multiple processors.  This is especially important during the early boot phase when no other processors are online and when the infrastructure required to broadcast TLB invalidations may not yet be fully initialized.  The same is true for temporary mappings established in a per-CPU address range.

### DIFF
--- a/kernel/src/cpu/control_regs.rs
+++ b/kernel/src/cpu/control_regs.rs
@@ -24,10 +24,12 @@ pub fn cr4_init() {
 
     cr4.insert(CR4Flags::PSE); // Enable Page Size Extensions
 
-    if cpu_has_pge() {
-        cr4.insert(CR4Flags::PGE); // Enable Global Pages
-    }
+    // All processors that are capable of virtualization will support global
+    // page table entries, so there is no reason to support any processor that
+    // does not enumerate PGE capability.
+    assert!(cpu_has_pge(), "CPU does not support PGE");
 
+    cr4.insert(CR4Flags::PGE); // Enable Global Pages
     write_cr4(cr4);
 }
 
--- a/kernel/src/cpu/control_regs.rs
+++ b/kernel/src/cpu/control_regs.rs
@@ -34,6 +34,7 @@ pub fn cr4_init() {
 }
 
 bitflags! {
+    #[derive(Debug, Clone, Copy)]
     pub struct CR0Flags: u64 {
         const PE = 1 << 0;  // Protection Enabled
         const MP = 1 << 1;  // Monitor Coprocessor
@@ -108,6 +109,7 @@ pub fn write_cr3(cr3: PhysAddr) {
 }
 
 bitflags! {
+    #[derive(Debug, Clone, Copy)]
     pub struct CR4Flags: u64 {
         const VME       = 1 << 0;  // Virtual-8086 Mode Extensions
         const PVI       = 1 << 1;  // Protected-Mode Virtual Interrupts
--- a/kernel/src/cpu/efer.rs
+++ b/kernel/src/cpu/efer.rs
@@ -36,9 +36,11 @@ pub fn write_efer(efer: EFERFlags) {
 pub fn efer_init() {
     let mut efer = read_efer();
 
-    if cpu_has_nx() {
-        efer.insert(EFERFlags::NXE);
-    }
+    // All processors that are capable of virtualization will support
+    // no-execute table entries, so there is no reason to support any processor
+    // that does not enumerate NX capability.
+    assert!(cpu_has_nx(), "CPU does not support NX");
 
+    efer.insert(EFERFlags::NXE);
     write_efer(efer);
 }
--- a/kernel/src/cpu/tlb.rs
+++ b/kernel/src/cpu/tlb.rs
@@ -5,6 +5,8 @@
 // Author: Joerg Roedel <jroedel@suse.de>
 
 use crate::address::{Address, VirtAddr};
+use crate::cpu::control_regs::{read_cr4, write_cr4, CR4Flags};
+
 use core::arch::asm;
 
 const INVLPGB_VALID_VA: u64 = 1u64 << 0;
@@ -48,6 +50,21 @@ pub fn flush_tlb_global() {
 pub fn flush_tlb_global_sync() {
     flush_tlb_global();
     do_tlbsync();
+}
+
+pub fn flush_tlb_global_percpu() {
+    let cr4 = read_cr4();
+    write_cr4(cr4 ^ CR4Flags::PGE);
+    write_cr4(cr4);
+}
+
+pub fn flush_address_percpu(va: VirtAddr) {
+    let va: u64 = va.page_align().bits() as u64;
+    unsafe {
+        asm!("invlpg (%rax)",
+             in("rax") va,
+             options(att_syntax));
+    }
 }
 
 pub fn flush_address(va: VirtAddr) {
--- a/kernel/src/mm/ptguards.rs
+++ b/kernel/src/mm/ptguards.rs
@@ -7,12 +7,12 @@
 use super::pagetable::PTEntryFlags;
 use crate::address::{Address, PhysAddr, VirtAddr};
 use crate::cpu::percpu::this_cpu;
-use crate::cpu::tlb::flush_address_sync;
+use crate::cpu::tlb::flush_address_percpu;
 use crate::error::SvsmError;
 use crate::mm::virtualrange::{
     virt_alloc_range_2m, virt_alloc_range_4k, virt_free_range_2m, virt_free_range_4k,
 };
-use crate::types::{PAGE_SIZE, PAGE_SIZE_2M};
+use crate::types::{PageSize, PAGE_SIZE, PAGE_SIZE_2M};
 
 use crate::utils::MemoryRegion;
 
@@ -99,13 +99,19 @@ impl PerCPUPageMappingGuard {
 
 impl Drop for PerCPUPageMappingGuard {
     fn drop(&mut self) {
-        if self.huge {
+        let size = if self.huge {
             this_cpu().get_pgtable().unmap_region_2m(self.mapping);
             virt_free_range_2m(self.mapping);
+            PageSize::Huge
         } else {
             this_cpu().get_pgtable().unmap_region_4k(self.mapping);
             virt_free_range_4k(self.mapping);
+            PageSize::Regular
+        };
+        // This iterative flush is acceptable for same-CPU mappings because no
+        // broadcast is involved for each iteration.
+        for page in self.mapping.iter_pages(size) {
+            flush_address_percpu(page);
         }
-        flush_address_sync(self.mapping.start());
     }
 }
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -323,7 +323,12 @@ impl PerCpu {
             tss: Cell::new(X86Tss::new()),
             svsm_vmsa: OnceCell::new(),
             reset_ip: Cell::new(0xffff_fff0),
-            vm_range: VMR::new(SVSM_PERCPU_BASE, SVSM_PERCPU_END, PTEntryFlags::GLOBAL),
+            vm_range: {
+                let mut vmr = VMR::new(SVSM_PERCPU_BASE, SVSM_PERCPU_END, PTEntryFlags::GLOBAL);
+                vmr.set_per_cpu(true);
+                vmr
+            },
+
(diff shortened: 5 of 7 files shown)


kernel/src/mm/ptguards.rs
```diff
@@ -106,6 +106,6 @@ impl Drop for PerCPUPageMappingGuard {
             this_cpu().get_pgtable().unmap_region_4k(self.mapping);
             virt_free_range_4k(self.mapping);
         }
-        flush_address_sync(self.mapping.start());
+        flush_address_percpu(self.mapping.start());
```
00xc, line 109: The mapping may be larger than a single page, does this call take that into account?

kernel/src/cpu/tlb.rs
```diff
@@ -50,6 +50,29 @@ pub fn flush_tlb_global_sync() {
...
+        asm!(
+            "movq %cr4, %rax",
+            "movl $0x80, %edx", // CR4_PGE
+            "not %rdx",
+            "andq %rax, %rdx",
+            "movq %rdx, %cr4",
+            "movq %rax, %cr4",
+            options(att_syntax)
+        );
+    }
```
00xc, line 64: Could we do this using `read_cr4()` and `write_cr4()` in `kernel/src/cpu/control_regs.rs`?

kernel/src/mm/ptguards.rs
msft-jlange: The INVLPG instruction does not consider page size, and is guaranteed to remove the TLB entry based on the size of the PTE that created it.  Therefore, there is no reason to pass page size here.

kernel/src/cpu/tlb.rs
msft-jlange: Yes, I'll make that change.

kernel/src/mm/ptguards.rs
00xc: No, I meant the type supports mapping several contiguous 4k pages, but it seems this will only flush the entry for the first one, if I understand correctly.

msft-jlange: The new code has exactly the properties of the code that it's replacing: it invalidates a single page, which could either be a 4 KB page or a 2 MB page.  If your concern is that `PerCPUPageMappingGuard` can map more than a single page, then the existing code is already broken, and this PR doesn't make it any better or worse.  We can define a range-based TLB flush (we will need to at some point) but I think it's preferable to do that outside of this PR, and for this PR to preserve existing behavior, for good or for bad.

00xc: Yes, the code was already broken. This seems like a good opportunity to fix it. I'd say something like the following should do:

```diff
diff --git a/kernel/src/mm/ptguards.rs b/kernel/src/mm/ptguards.rs
index 17f3eb24..6f84fbb5 100644
--- a/kernel/src/mm/ptguards.rs
+++ b/kernel/src/mm/ptguards.rs
@@ -102,10 +102,15 @@ impl Drop for PerCPUPageMappingGuard {
         if self.huge {
             this_cpu().get_pgtable().unmap_region_2m(self.mapping);
             virt_free_range_2m(self.mapping);
+            for page in self.mapping.iter_pages(PageSize::Huge) {
+                flush_address_percpu(page);
+            }
         } else {
             this_cpu().get_pgtable().unmap_region_4k(self.mapping);
             virt_free_range_4k(self.mapping);
+            for page in self.mapping.iter_pages(PageSize::Regular) {
+                flush_address_percpu(page);
+            }
         }
-        flush_address_sync(self.mapping.start());
     }
 }
```

msft-jlange: Since this code is necessarily limited to same-CPU flushes (i.e. not broadcast), then this should be OK.  We would have had major performance problems if we were to do a loop with a broadcast sync after each iteration of the loop, but since that doesn't apply in this case, we can overlook it.

kernel/src/cpu/control_regs.rs
```diff
@@ -24,10 +24,14 @@ pub fn cr4_init() {
...
     cr4.insert(CR4Flags::PSE); // Enable Page Size Extensions
 
-    if cpu_has_pge() {
-        cr4.insert(CR4Flags::PGE); // Enable Global Pages
+    if !cpu_has_pge() {
+        // All processors that are capable of virtualization will support
+        // global page table entries, so there is no reason to support any
+        // processor that does not enumerate PGE capability.
+        panic!("CPU does not support PGE");
     }
```
00xc, line 32: Minor style suggestion:

```suggestion
    // All processors that are capable of virtualization will support
    // global page table entries, so there is no reason to support any
    // processor that does not enumerate PGE capability.
    assert!(cpu_has_pge(), "CPU does not support PGE");
```

kernel/src/cpu/efer.rs
```diff
@@ -36,9 +36,13 @@ pub fn write_efer(efer: EFERFlags) {
...
     let mut efer = read_efer();
 
-    if cpu_has_nx() {
-        efer.insert(EFERFlags::NXE);
+    if !cpu_has_nx() {
+        // All processors that are capable of virtualization will support
+        // no-execute table entries, so there is no reason to support any
+        // processor that does not enumerate NX capability.
+        panic!("CPU does not support NX");
     }
```
00xc, line 44: Again, style suggestion, feel free to ignore:

```suggestion
    // All processors that are capable of virtualization will support
    // no-execute table entries, so there is no reason to support any
    // processor that does not enumerate NX capability.
    assert!(cpu_has_nx(), "CPU does not support NX");
```

kernel/src/cpu/tlb.rs
```diff
@@ -50,6 +52,28 @@ pub fn flush_tlb_global_sync() {
...
+    let cr4 = read_cr4();
+    let cr4_nopge = {
+        // bitflags requires a mut object to add/remove flags, and bitflags
+        // does not implement either Clone or Copy.
+        let mut temp = CR4Flags::from_bits_truncate(cr4.bits());
+        temp.toggle(CR4Flags::PGE);
+        temp
+    };
+    write_cr4(cr4_nopge);
+    write_cr4(cr4);
```
00xc, line 65: We should derive `Copy` (and `Clone`) for `CR4Flags` so that we can do:

```suggestion
    let cr4 = read_cr4();
    write_cr4(cr4 ^ CR4Flags::PGE);
    write_cr4(cr4);
```

00xc (commented):
Left a few minor suggestions, other than that this looks good to me. I'll merge once comments are marked as resolved.

kernel/src/cpu/tlb.rs
00xc: i.e. simply doing:

```diff
diff --git a/kernel/src/cpu/control_regs.rs b/kernel/src/cpu/control_regs.rs
index 741ea5c..73df73e 100644
--- a/kernel/src/cpu/control_regs.rs
+++ b/kernel/src/cpu/control_regs.rs
@@ -110,6 +110,7 @@ pub fn write_cr3(cr3: PhysAddr) {
 }
 
 bitflags! {
+    #[derive(Debug, Clone, Copy)]
     pub struct CR4Flags: u64 {
         const VME       = 1 << 0;  // Virtual-8086 Mode Extensions
         const PVI       = 1 << 1;  // Protected-Mode Virtual Interrupts

```

msft-jlange: That is way, way better.  I didn't realize it would be that simple.

kernel/src/cpu/control_regs.rs
msft-jlange: I changed this to an assertion, but now I'm considering changing it back.  The assertion will only be performed in debug builds, but we want to prevent boot even in non-debug builds.  Doesn't that make `panic` more appropriate?

msft-jlange: Actually, I see that `assert` happens always and `debug_assert` is debug-only, so `assert` is appropriate here.

00xc:
In-SVSM tests fail after the last commit in this PR (e1c168b2b409af34a9d4229ec8500b635036fa11):

```
[SVSM] test greq::msg::tests::encrypt_decrypt_payload ...
[SVSM] ERROR: Panic: CPU[0] KVM: entry failed, hardware error 0xffffffff
EAX=00000020 EBX=00000000 ECX=00000000 EDX=00000000
ESI=00000000 EDI=00000000 EBP=00000000 ESP=00000000
EIP=00000000 EFL=00000002 [-------] CPL=0 II=0 A20=1 SMM=0 HLT=0
ES =0000 00000000 00000000 00000000
CS =0000 00000000 00000000 00000000
SS =0000 00000000 00000000 00000000
DS =0000 00000000 00000000 00000000
FS =0000 00000000 00000000 00000000
GS =0000 00000000 00000000 00000000
LDT=0000 00000000 00000000 00000000
TR =0000 00000000 00000000 00000000
GDT=     0000000000000000 00000000
IDT=     0000000000000000 00000000
CR0=80010031 CR2=0000000000000000 CR3=0000000000000000 CR4=000000f0
DR0=0000000000000000 DR1=0000000000000000 DR2=0000000000000000 DR3=0000000000000000 
DR6=0000000000000000 DR7=0000000000000000
EFER=0000000000000d00
Code=<??> ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ??
```

kernel/src/cpu/control_regs.rs
00xc: Correct, `assert` is always compiled in.

00xc:
> In-SVSM tests fail after the last commit in this PR ([e1c168b](https://github.com/coconut-svsm/svsm/commit/e1c168b2b409af34a9d4229ec8500b635036fa11)):

Happens only on debug builds it seems. It also can happen even earlier, so the test that failed above is not to blame:

```
[SVSM] COCONUT Secure Virtual Machine Service Module
[SVSM] Order-00: total pages:    47 free pages:     0
[SVSM] Order-01: total pages:     1 free pages:     1
[SVSM] Order-02: total pages:     0 free pages:     0
[SVSM] Order-03: total pages:     0 free pages:     0
[SVSM] Order-04: total pages:     0 free pages:     0
[SVSM] Order-05: total pages:   117 free pages:   117
[SVSM] Total memory: 15172KiB free memory: 14984KiB
[SVSM] Boot stack starts        @ 0xffffff800011f000
[SVSM] BSP Runtime stack starts @ 0xffffff000020c000
[SVSM] Guest Memory Regions:
[SVSM]   000000000000000000-000000000080000000
[SVSM]   000000000100000000-000000000280000000
[SVSM] Invalidating boot region 000000000000000000-0000000000000a0000
KVM: entry failed, hardware error 0xffffffff
EAX=0000000a EBX=00000000 ECX=00000000 EDX=00000000
ESI=00000000 EDI=00000000 EBP=00000000 ESP=00000000
EIP=00000000 EFL=00000002 [-------] CPL=0 II=0 A20=1 SMM=0 HLT=0
ES =0000 00000000 00000000 00000000
CS =0000 00000000 00000000 00000000
SS =0000 00000000 00000000 00000000
DS =0000 00000000 00000000 00000000
FS =0000 00000000 00000000 00000000
GS =0000 00000000 00000000 00000000
LDT=0000 00000000 00000000 00000000
TR =0000 00000000 00000000 00000000
GDT=     0000000000000000 00000000
IDT=     0000000000000000 00000000
CR0=80010031 CR2=0000000000000000 CR3=0000000000000000 CR4=000000f0
DR0=0000000000000000 DR1=0000000000000000 DR2=0000000000000000 DR3=0000000000000000 
DR6=0000000000000000 DR7=0000000000000000
EFER=0000000000000d00
Code=<??> ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ??
```

00xc:
Aaaand it went away with the latest nightly toolchain version, so not sure why it was failing. It also only failed when applying the `VMR` changes - applying only the `PerCPUPageMappingGuard` changes made the tests pass.

PR #419, opened 2024-07-23
stage2: add TD Partitioning support

peterfang:
This is the second half of the TDX enabling work in stage2. It enables SVSM running on a TDP platform to boot until the last jump in stage2_main(). It also adds support for serial console.


peterfang:
Will update this PR again after https://github.com/coconut-svsm/svsm/pull/432 since that impacts our stage1 trampoline logic

peterfang:
PR updated. TDP boots stage2 with stage2 moved to 8MB.

kernel/src/platform/mod.rs
```diff
@@ -83,7 +83,11 @@ pub trait SvsmPlatform {
     ) -> Result<(), SvsmError>;
 
     /// Marks a range of pages as valid for use as private pages.
-    fn validate_page_range(&self, region: MemoryRegion<VirtAddr>) -> Result<(), SvsmError>;
+    fn validate_page_range(
+        &self,
+        region: MemoryRegion<VirtAddr>,
+        paddr: PhysAddr,
+    ) -> Result<(), SvsmError>;
```
joergroedel, line 90: This interface change puzzles me a bit. I imposes certain constraints on the method caller (like that the virtual region maps physically contiguous memory).
At least the interface needs to be documented after that change. In general I think it is possible to get the physical address from within the method. Stage2 is identity-mapped and for the COCONUT kernel there is `virt_to_phys()`.

kernel/src/stage2.rs
```diff
@@ -440,6 +440,8 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) {
     // Shut down the GHCB
     shutdown_percpu();
 
+    log::info!("Starting SVSM kernel...");
```
joergroedel, line 443: This needs to happen before `shutdown_percpu()`, otherwise it crashes on SNP. The `shutdown_percpu()` function also shuts down the GHCB on SNP which is required for serial port access, and as a consequence also to print any log messages.

joergroedel (changes requested):
I pointed out two problems which need fixing, other than that it looks good to me. With the `log::info!` issue fixed it boots for me on SNP and testing passes.

kernel/src/stage2.rs
peterfang: Thanks for catching this! Will fix in my update.

kernel/src/platform/mod.rs
peterfang: `virt_to_phys()` is limited in stage2 because `FIXED_MAPPING` is only limited to the first 640KB + the stage2 area, despite the fact that `INIT_PGTABLE` continues to be updated in `map_and_validate()`. Do you think we should try to improve `virt_to_phys()` instead so that it tracks the new page table mappings?

msft-jlange: `validate_page_range()` should really operate on a physical range, because all platforms consider the concept of "validation" as a property of GPAs.  The fact that we use virtual addresses anywhere is an artifact of how `PVALIDATE` works on SNP.  The best approach here would be to have `valid_page_range()` take a physical address range, and then on platforms that require virtual mappings (like SNP), to use `phys_to_virt()` or a temporary per-CPU virtual mapping to complete the request.

msft-jlange: After taking a closer look, I think this is a big hack that really shouldn't be checked in.  Yes, `validate_page_range()` should take a physical address span, but it should not also take a virtual address which might or might not have anything to do with the physical address.  I will take a stab at changing the interface to use physical addresses (with per-CPU temporary mappings on SNP to obtain the virtual addresses required by `PVALIDATE`).

kernel/src/svsm_console.rs
```diff
@@ -52,6 +55,33 @@ impl IOPort for SVSMIOPort {
     }
 }
 
+#[derive(Clone, Copy, Debug, Default)]
+pub struct SvsmTdIOPort {}
```
msft-jlange, line 59: Can we not just change `SVSMIOPort` to use IN and OUT instructions natively, and then have the #VC handler issue GHCB calls and the #VE handler issue GHCI calls?  This seems like it would be much cleaner because it restricts platform-specific functionality to code paths that are already platform-specific (exception handlers) instead of common code (the console code).

kernel/src/cpu/idt/common.rs
```diff
@@ -215,7 +215,7 @@ impl WriteLockGuard<'static, IDT> {
     /// Load an IDT. Its lifetime must be static so that its entries are
     /// always available to the CPU.
     pub fn load(&self) {
-        let desc: IdtDesc = IdtDesc {
+        let desc = IdtDesc {
```
msft-jlange, line 218: What's the reason for this change?  Is this related to TDP?

kernel/src/debug/gdbstub.rs
```diff
@@ -173,7 +173,7 @@ pub mod svsm_gdbstub {
 
     static GDB_INITIALISED: AtomicBool = AtomicBool::new(false);
     static GDB_STATE: SpinLock<Option<SvsmGdbStub<'_>>> = SpinLock::new(None);
-    static GDB_IO: SVSMIOPort = SVSMIOPort::new();
+    static GDB_IO: SvsmSevIOPort = SvsmSevIOPort::new();
```
msft-jlange, line 176: Isn't this going to be problematic when we finally make GDB work on TDP?  Wouldn't it be better to define a platform abstraction now since you are already forced to touch this code?

kernel/src/svsm_console.rs
peterfang: Sounds good to me. I can make that change if it's fine by @joergroedel .

kernel/src/cpu/idt/common.rs
peterfang: I think this bled into this patchset when I created this PR. Would you like me to revert it or create a clean-up patch for it?

kernel/src/debug/gdbstub.rs
peterfang: I agree. I wasn't planning on tackling this in this PR yet but this becomes easier with a common `SVSMIOPort`.

kernel/src/cpu/idt/common.rs
msft-jlange: Personally, I always prefer when the set of changes in a patch is limited to work described in the PR summary - i.e., that each patch be minimal and specific.  Others may feel differently.

kernel/src/debug/gdbstub.rs
msft-jlange: Maybe add a `TODO` comment to point out that this is a known future work item.

kernel/src/platform/mod.rs
peterfang: Sounds good. Will rebase this commit after your changes are merged.

peterfang:
Updated this PR after https://github.com/coconut-svsm/svsm/pull/463 got merged

kernel/src/mm/pagetable.rs
```diff
@@ -611,9 +647,9 @@ impl PageTable {
     /// - `vaddr': The virtual address to transalte.
     ///
     /// # Returns
-    /// Some(PhysAddr) if the virtual address is valid.
+    /// Some(PageFrame) if the virtual address is valid.
     /// None if the virtual address is not valid.
-    pub fn virt_to_phys(vaddr: VirtAddr) -> Option<PhysAddr> {
+    pub fn virt_to_phys(vaddr: VirtAddr) -> Option<PageFrame> {
```
msft-jlange, line 652: For consistency, this function should be renamed `virt_to_frame` because it returns a `PageFrame` and not a `PhysAddr`.

kernel/src/mm/pagetable.rs
```diff
@@ -450,6 +450,42 @@ pub enum Mapping<'a> {
...
+        }
+    }
+
+    fn size(&self) -> usize {
+        let sz = match *self {
+            Self::Size4K(_) => PageSize::Regular,
+            Self::Size2M(_) => PageSize::Huge,
+            Self::Size1G(_) => PageSize::Huge1G,
+        };
+        usize::from(sz)
```
msft-jlange, line 476:
```suggestion
        match *self {
            Self::Size4K(_) => PageSize::Regular,
            Self::Size2M(_) => PageSize::Huge,
            Self::Size1G(_) => PageSize::Huge1G,
        }.into()
```

kernel/src/mm/pagetable.rs
```diff
@@ -611,9 +647,9 @@ impl PageTable {
     /// - `vaddr': The virtual address to transalte.
```
msft-jlange, line 647: Since you're fixing some of my other typos, it would be great if you could change `transalte` to `translate` while you're at it.

kernel/src/types.rs
```diff
@@ -9,20 +9,24 @@ use crate::sev::vmsa::VMPL_MAX;
...
 pub const PAGE_SIZE: usize = 1 << PAGE_SHIFT;
-pub const PAGE_SIZE_2M: usize = PAGE_SIZE * 512;
+pub const PAGE_SIZE_2M: usize = 1 << PAGE_SHIFT_2M;
+pub const PAGE_SIZE_1G: usize = 1 << PAGE_SHIFT_1G;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum PageSize {
     Regular,
     Huge,
+    Huge1G,
```
msft-jlange, line 21: Is this really necessary?  Given that SNP does not support 1 GB pages, I don't believe COCONUT-SVSM will ever attempt to make any use of them, and thus this size will never be meaningful.

kernel/src/platform/tdp.rs
```diff
@@ -88,18 +94,42 @@ impl SvsmPlatform for TdpPlatform {
...
+        region: MemoryRegion<VirtAddr>,
+        op: PageValidateOp,
     ) -> Result<(), SvsmError> {
-        Err(SvsmError::Tdx)
+        match op {
+            PageValidateOp::Validate => {
+                let mut va = region.start();
+                while va < region.end() {
+                    let pa = virt_to_frame(va);
+                    let sz = pa.end() - pa.address();
```
msft-jlange, line 122: The size of the virtual address mapping is not related to the size of the page in the secure EPT, so I don't understand why you are using the size here.

kernel/src/mm/pagetable.rs
peterfang: Good catch. I'll make that change.

kernel/src/mm/pagetable.rs
peterfang: That's more concise. I'll make that change.

kernel/src/mm/pagetable.rs
peterfang: Sure I can fix that as well.

kernel/src/types.rs
peterfang: Got it. I think there's still value in this enum because:
1. `PageTable::virt_to_phys()` checks for `pdpe.huge()` which implies this is a supported case (or we could put `unimplemented!` there to make sure `pdpe.huge()` never happens), and `usize::from(PageSize::Huge1G)` would naturally be the size of the page frame in such case.
2. The TDP spec does support 1G huge pages for L2 SEPTs (managed by the TDX module) and therefore this enum could be useful when managing L2 page aliases, especially if we have a guest with large memory.

msft-jlange: There's a difference between virtual address mappings and physical page sizes.  The `PageFrame` structure that you have defined should define a 1 GB enumeration for the reason you state as (1) above.  But that doesn't mean that the `PageSize` enum (which intends to describe physical pages, not virtual mappings) needs to define a 1 GB size as well.

In general, the definitions within the COCONUT-SVSM code base should exist to support the logic within the COCONUT-SVSM code base, not just to be a collection of what individual target platforms may happen to support.  Because you have defined `PageSize::Huge1GB`, you have been forced to add code that observes that enum and generate errors, which is rather awkward.  It's like one part of the code base says this size is support, and then another part says that it isn't, which does not make the code more readable.  It would be far better to leave the `PageSize` enum alone, and if there is any unique logic that TDP requires for 1 GB pages, then that should live within the TDP platform object.

From a quick observation, it looks like the only place you use `PageSize::Huge1GB` today is in `PageFrame::size()`, and that's only because you translate the virtual mapping to a physical page size before coercing it into a `usize`.  You could rewrite that function to convert from the virtual mapping size directly to a `usize` and never need any change to `PageSize`.  That would have much less impact across the code base and would eliminate the confusion concerns I describe above.

peterfang: I'm okay with removing `PageSize::Huge1GB` but would like to make a few counter points here:

1. Virtual address mappings != physical page sizes but in low-level code they often have a strong correlation. It's intuitive to me to have `PageSize` and `PageFrame` closely defined rather than keep them apart. It feels a bit odd to me if `PageFrame::size()` can't use things like `PageSize` or `PAGE_SIZE_1G` and would have to resort to raw conversion.
2. 1GB huge pages are pretty ubiquitous nowadays and it doesn't really feel awkward to me when there's an error that needs to be thrown on a particular target platform because of them (but it doesn't make code more readable either). Rather than viewing this as something TDP supports and other platforms don't, not supporting it can also be viewed as a restriction on a specific platform. There's nothing special about TDP supporting 1GB pages except for the fact that we allow them in our page tables. I would assume other coco platforms would be supporting them at some point as well. I'm also curious about why `kernel/src/platform/` wouldn't be considered part of the COCONUT-SVSM code base. Handling page aliases and managing L2 guest memory would likely require us to leverage some of the mm code.

That said, I can see that there's currently no mm infrastructure for supporting 1GB huge pages (e.g. in `PageTable`) so adding an enum at this point would lead to lots of redundant error handling. I'm okay with not adding it at this point if we can come up with a way to implement `PageFrame::size()` more semantically than just raw conversion. How about we keep `PAGE_SHIFT_1G` and `PAGE_SIZE_1G`?

msft-jlange: `PageFrame` needs to support whatever `virt_to_frame` can generate.  `virt_to_frame` can generate a 1 GB page because it is architecturally permissible to encode a 1 GB page through large-page PTEs.  I don't think it would be consistent for `virt_to_frame` to generate an error if it sees a 1 GB encoding while the processor would not raise #PF under the same circumstances.  This makes the need for a 1 GB frame size unavoidable.

By "COCONUT-SVSM code base" I mean the kernel proper and all of the common, platform-independent code.  So where I pushed back earlier on the idea of "definitions within the COCONUT-SVSM code base" I meant definitions within common code, like memory management and system-wide definitions, which is the category that includes `PageSize`.  We can reasonably debate where that line should get drawn and what definitions should be placed into common code for the benefit of one or more specific platform implementations, but you've really hit on the key point here: any definition that leads to redundant error handling across the code base is something we should question.

I will also point out that COCONUT-SVSM is intended to support a small (relatively speaking) security module in a virtual machine environment, and it's unlikely that it would ever even have 1 GB of memory allocated to it, so I don't foresee any call for 1 GB page mappings in its lifetime.  This is why there is not only currently no mm infrastructure for 1 GB pages, but why it is unlikely to materialize at any point.

Definitions like `PAGE_SHIFT_1G` and `PAGE_SIZE_1G` correspond to the architectural definition of page table layout, so it makes perfect sense to keep them.  I have the strongest pushback against enum values that force `match` statements to include extraneous and unhelpful error-handling logic, which is why I find `PageSize` the only type of concern.

peterfang: Makes lots of sense to me. I'll remove that enum and update accordingly. Thanks!

msft-jlange (approved):

joergroedel (approved):
Looks good to me now.

PR #420, opened 2024-07-23
mm/pagebox: add dynamically-sized slice support

00xc:
Add support for allocating dynamically-sized slices of values using the page allocator. This essentially works like a regular Vec, except the slice does not support resizing once created.

This is useful for the use case we have in the valid bitmap. Thus, replace the low-level page allocation calls in that code with `PageBox` uses.

### DIFF
--- a/kernel/src/mm/validate.rs
+++ b/kernel/src/mm/validate.rs
@@ -13,7 +13,7 @@ use crate::types::{PAGE_SIZE, PAGE_SIZE_2M};
 use crate::utils::MemoryRegion;
 use core::ptr;
 
-static VALID_BITMAP: SpinLock<ValidBitmap> = SpinLock::new(ValidBitmap::new());
+static VALID_BITMAP: SpinLock<Option<ValidBitmap>> = SpinLock::new(None);
 
 #[inline(always)]
 fn bitmap_alloc_order(region: MemoryRegion<PhysAddr>) -> usize {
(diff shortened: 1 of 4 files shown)


00xc:
Once this is merged we can make the raw page allocation functions private (or directly remove them in some cases).

kernel/src/mm/validate.rs
```diff
@@ -134,84 +138,44 @@ impl ValidBitmap {
         (index, bit)
     }
 
-    fn clear_all(&mut self) {
-        let len = self.bitmap_len();
-        unsafe {
-            ptr::write_bytes(self.bitmap, 0, len);
+    fn migrate(&mut self, mut new: PageBox<[MaybeUninit<u64>]>) {
+        for (dst, src) in new.iter_mut().zip(self.bitmap.iter()) {
```
Freax13, line 142: Here's another way to implement this with a single loop:
```rust
for (dst, src) in new
    .iter_mut()
    .zip(self.bitmap.iter().copied().chain(repeat(0)))
{
    dst.write(src);
}
```

kernel/src/mm/validate.rs
```diff
@@ -192,10 +174,8 @@ impl ValidBitmap {
         assert!(paddr.is_aligned(PAGE_SIZE_2M));
         assert!(self.check_addr(paddr));
 
-        for i in 0..NR_INDEX {
-            unsafe {
-                ptr::write(self.bitmap.add(index + i), val);
-            }
+        for dst in self.bitmap[index..index + NR_INDEX].iter_mut() {
+            *dst = val;
```
Freax13, line 178: This loop (and the one in line 207) can be replaced with
```rust
self.bitmap[index..index + NR_INDEX].fill(val);
```

Freax13 (approved):
LGTM!

00xc:
Applied Tom's suggestions and rebased on main

kernel/src/mm/validate.rs
```diff
@@ -245,8 +202,8 @@ impl ValidBitmap {
             let mask_head = mask >> bit_head_begin << bit_head_begin;
             self.modify_bitmap_word(index_head, mask_head, new_val);
 
-            for index in (index_head + 1)..index_tail {
-                unsafe { ptr::write(self.bitmap.add(index), new_val) };
+            for val in self.bitmap[index_head + 1..index_tail].iter_mut() {
+                *val = new_val;
```
Freax13, line 206: This loop can also be expressed using `.fill(new_value)`.

Freax13 (approved):

PR #423, opened 2024-07-25
Add a target for svsm-test.bin

deeglaze:
We run selftests with separate binaries and would like to use an upstream target to do the objcopy step.


deeglaze:
If y'all would prefer, we can put our build scripts in /scripts/google/... with me and @AdamCDunlap as CODEOWNERS. That would allow us to more seamlessly maintain build transparency.

PR #431, opened 2024-08-01
kernel/task/schedule: Disable interrupt during context switching

cxdong:
The instructions "mov %rdx, %cr3" and "movq (%rdi), %rsp" will switch the page table and stack. An interrupt happening in the middle of the switching can corrupt the stack, as it will push data to the old rsp address but is translated by the new page table. To avoid such stack corruption, disable interrupt before the switching the page table and stack. The interrupt will be restored by the popfq.

### DIFF
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -388,6 +388,9 @@ global_asm!(
         movq    %rsp, (%rsi)
 
     1:
+        // Disable interrupt
+        cli
+
         // Switch to the new task state
         mov     %rdx, %cr3
 


msft-jlange:
The kernel currently executes with interrupts disabled, so this change will have no practical effect.

Once the kernel starts executing with interrupts enabled, there will be many places beyond this place that require careful management of interrupts-enabled state.  That will require more complex constructs for tracking the current interrupt-enabled state, and I expect that this specific CLI instruction will no longer be required (or worse, may actually be in the way).  I think it would be best to avoid making this change until complete interrupt management logic is introduced.

cxdong:
> The kernel currently executes with interrupts disabled, so this change will have no practical effect.

True. I didn't observe any issue when playing the schedule until enabling the interrupts to receive the virtual interrupts injected by the host.

> 
> Once the kernel starts executing with interrupts enabled, there will be many places beyond this place that require careful management of interrupts-enabled state.

Yes, I agree.

> That will require more complex constructs for tracking the current interrupt-enabled state, and I expect that this specific CLI instruction will no longer be required (or worse, may actually be in the way).

Would you please elaborate a bit more about this? My understanding is that if "switch_context" is invoked while interrupts are enabled, we must disable interrupts within "switch_context" to ensure atomic switching of the page table and RSP. Furthermore, if "switch_context" is called with interrupts already disabled, issuing a CLI instruction to guarantee interrupts are disabled within the function appears to be benign. But I suspect I might be overlooking something.

> I think it would be best to avoid making this change until complete interrupt management logic is introduced.

Sure, I am totally fine to close this PR if it can potentially cause any conflict with the ongoing interrupt management developing work. :)

msft-jlange:
> > That will require more complex constructs for tracking the current interrupt-enabled state, and I expect that this specific CLI instruction will no longer be required (or worse, may actually be in the way).
> 
> Would you please elaborate a bit more about this? My understanding is that if "switch_context" is invoked while interrupts are enabled, we must disable interrupts within "switch_context" to ensure atomic switching of the page table and RSP. Furthermore, if "switch_context" is called with interrupts already disabled, issuing a CLI instruction to guarantee interrupts are disabled within the function appears to be benign. But I suspect I might be overlooking something.

The desire is to apply Rust safety principles to interrupt management, so we will define an `InterruptsDisabledGuard` object which will disable interrupts and which is guaranteed to restore the previous state when the object is dropped.  This ensures that it is not possible to write code that inadvertently fails to reenable interrupts.  When this is done, the only place that `CLI` should appear is in the internal implementation of `InterruptsDisabledGuard`.

What I'm describing here is on my short-term list of work, along with enabling use of TPR (which uses a similar guard object mechanism to ensure that TPR is always restored after it is raised).

msft-jlange:
As I look at this more, it appears that making context switch safe (i.e. making the observation of context switches appear to be atomic) means that interrupts will need to be disabled long before entering `switch_context`, so there will need to be additional information that flows through so that the final POPFQ in the context of the new task establishes the correct state.  If that is correct, then no `CLI` is required in `switch_context` because interrupts will already be guaranteed to be disabled.  But I'm still investigating.

cxdong:
> As I look at this more, it appears that making context switch safe (i.e. making the observation of context switches appear to be atomic) means that interrupts will need to be disabled long before entering `switch_context`, so there will need to be additional information that flows through so that the final POPFQ in the context of the new task establishes the correct state. If that is correct, then no `CLI` is required in `switch_context` because interrupts will already be guaranteed to be disabled. But I'm still investigating.

I agree if the interrupt is already guaranteed to be disabled before `switch_context`, then no CLI is needed inside the `switch_context`, but as you mentioned, that will require to restore the proper interrupt state before the new task starts to run. If the interrupt is not explicitly disabled before `switch_context` (task itself can still disable interrupt, but the code in schedule path doesn't explicitly disable interrupt), then seems adding a CLI inside `switch_context` is easier to restore the interrupt state for the new task, just via the POPFQ. Of course, if you do see the reason that interrupt must be disabled long before calling `switch_context`, please let me know. Thanks!

cxdong:
> > I agree if the interrupt is already guaranteed to be disabled before `switch_context`, then no CLI is needed inside the `switch_context`, but as you mentioned, that will require to restore the proper interrupt state before the new task starts to run. If the interrupt is not explicitly disabled before `switch_context` (task itself can still disable interrupt, but the code in schedule path doesn't explicitly disable interrupt), then seems adding a CLI inside `switch_context` is easier to restore the interrupt state for the new task, just via the POPFQ. Of course, if you do see the reason that interrupt must be disabled long before calling `switch_context`, please let me know. Thanks!
> 
> It's going to take me a while to work out the correct implementation here (largely out of a desire to make interrupt management and reentrancy completely safe by Rust safety standards), but I think the direction is going to be that `schedule()` will have to disable interrupts, and will pass the previous interrupt state to `switch_to()`, which will set IF correctly in the value of RFLAGS that it pushes onto the stack.

Thanks for letting me know your implementation plan. So in this direction, the PUSHFQ in `switch_to` can not be used as it pushes the current RFLAGS which always has interrupt state disabled. Instead, `switch_to` push the pass in previous RFLAGS value directly to the stack, right?

> In this way, once the task switch is complete, the final POPFQ will correctly enable interrupts based on the state that task had when it was switched out in the first place. This all means that the CLI you have proposed will not be necessary.

Sure. As the implementation will go with the direction that having interrupt disabled before `switch_to`, I can close this PR as the CLI instruction is not necessary in this case.

cxdong:
Close this PR as Jon is working on another direction which guarantees the interrupt disabled before `switch_to`, so no need to involve a CLI instruction inside `switch_to`

PR #436, opened 2024-08-07
kernel/svsm: Increase boot time stack by 4KB

cxdong:
During early boot of svsm kernel, init page table is created via still using the boot time stack. The page table is allocated by using PageBox which will get a temporary PTPage (4KB) from the stack via the PTPage::default() method. This increases the load on the boot time stack and may cause stack overflow in the next. As the default method requires 4KB stack size, increase the boot time stack by 4KB to satisfy this.

With code increasing in the future, the boot time stack may not be large enough again, so this is not a permanent fix but just works for short term. If it conflicts with any ongoing stack unwind/overflow works, please let me know and I will close it. Thanks!

### DIFF
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -88,7 +88,7 @@ global_asm!(
 
         .align 4096
     bsp_stack:
-        .fill 4*4096, 1, 0
+        .fill 5*4096, 1, 0
     bsp_stack_end:
         "#,
     options(att_syntax)


p4zuu:
Just to be sure, did you already face a crash due to stack overflow?

cxdong:
> Just to be sure, did you already face a crash due to stack overflow?

I didn't face a crash with the upstream main, but did face a crash with my local development branch with additional changes ( IOIO PR, and some other changes for TDX). The crash happens around the init_page_table() when svsm kernel is still using the boot time stack. Increasing the boot time stack with 4k solves this for me. From another point of view, PTPage::default() consumes 1-page stack memory, which seems too heavy for the boot time stack as it totally only has 4-pages, and seems increasing it with additional 1-page would be an insurance for using PTPage::default() with the boot time stack.

p4zuu (approved):

cxdong:
> > I didn't face a crash with the upstream main, but did face a crash with my local development branch with additional changes ( IOIO PR, and some other changes for TDX). The crash happens around the init_page_table() when svsm kernel is still using the boot time stack. Increasing the boot time stack with 4k solves this for me. From another point of view, PTPage::default() consumes 1-page stack memory, which seems too heavy for the boot time stack as it totally only has 4-pages, and seems increasing it with additional 1-page would be an insurance for using PTPage::default() with the boot time stack.
> 
> Thanks for pointing this, that's a good catch. I'm quite surprised that `PageBox::try_new(PageTable::default())` consumes a whole page on the stack (`PageTable::default()` would allocate a `PageTable` on the stack even though it's placed in a `Box`?).

My understanding is that, as PageTable contains PTPage, PageTable::default() will call the default method of PTPage, which consumes 4KB stack by `let entries = [PTEntry::default(); ENTRY_COUNT];`. So PTPage itself is in stack, until it is written to the heap page allocated by PageBox::try_new().

>Maybe @00xc can help on this?
> 
> Overall this looks like a legit temporary solution.

Yes, I agree. Also pointed this out that this PR is not a permanent fix. Please let me know if I should close this PR. Thanks!

joergroedel:
I ran into similar issues when working on the IRQ state code for PR #441. I added a commit there to actually double the boot-stack size, otherwise the test-in-svsm target reliably failed.

We need to investigate why the code-base is getting so stack hungry and work on a plan to fix that, but until then we just double the boot-stack size.

Another feature that would make boot-stack overflow behavior more consistent would be guard pages around it. That is another thing we should look into implementing.

PR #439, opened 2024-08-15
stage1: Fix boot issue where %esi was wrong on stage2 entry

roy-hopkins:
The changes to support TDX from stage1 introduced an exchange of the BSP/AP index between stage1 and stage2. This was not implemented in the non-TDX case resulting in a non-zero %edi register on entry to stage2, resulting in the identity mapped page table not being configured.

This patch clears %esi on entry to stage2.


AdamCDunlap:
I tested this patch and it fixes the boot for us. Thanks!

PR #441, opened 2024-08-16
Introduce IRQ state management API and disable IRQs during task switches.

joergroedel:
This PR introduces helper functions and guards for managing IRQ state during code execution in the COCONUT kernel.

It also introduces IRQ-safe versions on `SpinLock` and `RWLock`.


kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,195 @@
...
+///
+/// Callers need to make sure to pass the correct previous state into
+/// the function to not accidentially re-enable IRQs.
+#[inline(always)]
+pub unsafe fn irqs_restore(state: u64) {
+    asm!("pushq %rax",
+         "popfq",
+         in("rax") state,
+         options(att_syntax));
+}
```
Freax13, line 103: This potentially modifies flags other than IF. I don't know if we currently use any of the other flags (such as AC for SMAP), but it's probably safer to not touch the other flags if we don't need to.

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// And IRQ guard which saves the current IRQ state and disabled interrupts
+/// upon creation. When the guard goes out of scope the previous IRQ state is
+/// restored.
+///
+/// The struct implements the `Default` and `Drop` traits for easy use.
+#[derive(Debug)]
+#[must_use = "if unused previous IRQ state will be immediatly restored"]
+pub struct IrqGuard {
```
Freax13, line 112: This API can cause some problems if multiple IrqGuards with different lifetimes are used.

```rust
let guard1 = IrqGuard::default(); // This disables interrupts.
let guard2 = IrqGuard::default(); // This does nothing because interrupts are already disabled.
drop(guard1); // This re-enables interrupts.
// At this point guard2 is still alive, but interrupts have been enabled again.
drop(guard2); // This disables interrupts again.
// Now interrupts are disabled even though no guards are left.
```

One solution for this problem I've seen before is to have a cpu-local counter to count the number of active IRQ guards. With this approach, IRQs are only disabled when the counter is increased from 0 to 1 and IRQs are only disabled when the counter is decreased from 1 to 0. 

`IrqSafeLocking` suffers from the same issues.

kernel/src/locking/common.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    fn irqs_save(&mut self) {
+        assert!(self.state.is_none());
+        self.state = Some(unsafe { irqs_save() });
+    }
+
+    fn irqs_restore(&mut self) {
+        self.state.map(|s| unsafe { irqs_restore(s) }).take();
+        self.state = None;
```
Freax13, line 63:
```suggestion
        self.state.map(|s| unsafe { irqs_restore(s) });
        self.state = None;
```

or

```suggestion
        if let Some(state) = self.state {
            unsafe {
                irqs_restore(s);
            }
        }
```

kernel/src/locking/rwlock.rs
```diff
@@ -4,59 +4,70 @@
...
 /// A guard that provides read access to the data protected by [`RWLock`]
 #[derive(Debug)]
 #[must_use = "if unused the RWLock will immediately unlock"]
-pub struct ReadLockGuard<'a, T> {
+pub struct RawReadLockGuard<'a, T, I: IrqLocking> {
     /// Reference to the associated `AtomicU64` in the [`RWLock`]
     rwlock: &'a AtomicU64,
     /// Reference to the protected data
     data: &'a T,
+    /// IRQ state before and after critical sectionV
```
Freax13, line 21:
```suggestion
    /// IRQ state before and after critical section
```

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,195 @@
...
+#[must_use = "Unused irqs_enabled() result - meant to be irq_enable()?"]
+pub fn irqs_enabled() -> bool {
+    // SAFETY: The inline assembly just reads the processors RFLAGS register
+    // and does not change any state.
+    let state: u64 = unsafe {
+        let s: u64;
+        asm!("pushfq",
+             "popq %rax",
+             out("rax") s,
+             options(att_syntax));
```
00xc, line 47: We do not need to force using RAX:

```suggestion
        asm!("pushfq",
             "popq {}",
             out(reg) s,
             options(att_syntax, nomem, preserves_flags));
```

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,195 @@
...
+#[inline(always)]
+pub unsafe fn irqs_disable() {
+    asm!("cli", options(att_syntax));
+}
+
+/// Unconditionally enable IRQs
+///
+/// # Safety
+///
+/// Callers need to make sure it is safe to enable IRQs
```
00xc, line 26: This probably needs to be more specific

kernel/src/locking/common.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// Implements the IRQ state handling methods as no-ops. For use it IRQ-unsafe
+/// locks.
+#[derive(Debug, Default)]
+pub struct IrqUnsafeLocking;
+
+impl IrqLocking for IrqUnsafeLocking {
+    fn new() -> Self {
+        IrqUnsafeLocking {}
```
00xc, line 40:
```suggestion
        Self {}
```

kernel/src/locking/common.rs
00xc: The second suggestion does not leave `None` in `state`, it needs `take()` if I'm understanding correctly

kernel/src/locking/rwlock.rs
```diff
@@ -149,9 +165,10 @@ impl<T> RWLock<T> {
     /// let rwlock = RWLock::new(data);
     /// ```
     pub const fn new(data: T) -> Self {
-        RWLock {
+        RawRWLock {
```
00xc, line 168:
```suggestion
        Self {
```

kernel/src/locking/spinlock.rs
```diff
@@ -110,10 +126,11 @@ impl<T> SpinLock<T> {
     /// let spin_lock = SpinLock::new(data);
     /// ```
     pub const fn new(data: T) -> Self {
-        SpinLock {
+        RawSpinLock {
```
00xc, line 129:
```suggestion
        Self {
```

kernel/src/locking/common.rs
00xc: The second suggestion does not leave `None` in `state`, it needs `take()` if I'm understanding correctly

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,195 @@
...
+const EFLAGS_IF: u64 = 1 << 9;
+
+/// Unconditionally disable IRQs
+///
+/// # Safety
+///
+/// Callers need to take care of re-enabling IRQs.
+#[inline(always)]
+pub unsafe fn irqs_disable() {
+    asm!("cli", options(att_syntax));
```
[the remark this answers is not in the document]
Freax13: Adding `preserves_flags` is only safe if we make sure not to modify `CF`, `PF`, `AF`, `ZF`, `SF`, and `OF` (source: https://doc.rust-lang.org/reference/inline-assembly.html).

kernel/src/cpu/irq_state.rs
Freax13: Adding `preserves_flags` is only safe if we make sure not to modify `CF`, `PF`, `AF`, `ZF`, `SF`, and `OF` (source: https://doc.rust-lang.org/reference/inline-assembly.html).

kernel/src/cpu/irq_state.rs
joergroedel: COCONUT does currently not use any other flags than IF. Any use of AC for SMAP should not overlap with IRQ enable/disable logic, maybe we should add a debugging check for that once we start using implementing SMAP.

kernel/src/cpu/irq_state.rs
joergroedel: That is a good point, the suggestion with a percpu counter is the solution here. This is getting to the point where COCONUT should also start implementing preemption checks (like IRQs not disabled during task switch, no pending borrows of percpu).

kernel/src/locking/rwlock.rs
joergroedel: Thanks, fixed.

kernel/src/cpu/irq_state.rs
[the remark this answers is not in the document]
joergroedel: Added, thanks.

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/// Unconditionally enable IRQs
+///
+/// # Safety
+///
+/// Callers need to make sure it is safe to enable IRQs
+#[inline(always)]
+pub unsafe fn irqs_enable() {
+    asm!("sti", options(att_syntax));
```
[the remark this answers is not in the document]
joergroedel: Added, thanks.

kernel/src/cpu/irq_state.rs
joergroedel: Updated, thanks.

kernel/src/locking/common.rs
joergroedel: Fixed, thanks.

kernel/src/locking/spinlock.rs
joergroedel: Fixed.

kernel/src/locking/rwlock.rs
joergroedel: Fixed.

kernel/src/cpu/irq_state.rs
joergroedel: Okay, the new implementation only uses `sti` and `cli` instructions, leaving all other flags untouched. This should be safe going forward.

kernel/src/cpu/irq_state.rs
joergroedel: I implemented this via a percpu-counter now. This caused problems elsewhere, but those should all be solved as well.

kernel/src/locking/common.rs
joergroedel: This code is gone now. There is no state to carry forward, everything is stored in per-cpu now,

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,242 @@
...
+    #[inline(always)]
+    pub unsafe fn enable(&self) {
+        debug_assert!(irqs_disabled());
+
+        self.count.fetch_sub(1, Ordering::Relaxed);
+        let val = self.count.load(Ordering::Relaxed);
+
+        assert!(val >= 0);
+
+        if val == 0 {
```
Freax13, line 130:
```suggestion
        let val = self.count.fetch_sub(1, Ordering::Relaxed);
        self.count.load(Ordering::Relaxed);

        assert!(val > 0);

        if val == 1 {
```

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// And IRQ guard which saves the current IRQ state and disabled interrupts
+/// upon creation. When the guard goes out of scope the previous IRQ state is
+/// restored.
+///
+/// The struct implements the `Default` and `Drop` traits for easy use.
+#[derive(Debug)]
+#[must_use = "if unused previous IRQ state will be immediatly restored"]
+pub struct IrqGuard;
```
Freax13, line 164:
```suggestion
pub struct IrqGuard(());
```

Adding a private unit field makes it impossible to create an instance without calling `IrqGuard::new()` (or `IrqGuard::default()`).

kernel/src/cpu/irq_state.rs
joergroedel: With this implementation the load should be redundant, no?

kernel/src/cpu/irq_state.rs
joergroedel: Thanks, learned something new again :)

kernel/src/cpu/irq_state.rs
Freax13: Oops, yes, of course.

kernel/src/locking/rwlock.rs
```diff
@@ -4,22 +4,27 @@
...
 use core::cell::UnsafeCell;
+use core::marker::PhantomData;
 use core::ops::{Deref, DerefMut};
 use core::sync::atomic::{AtomicU64, Ordering};
 
 /// A guard that provides read access to the data protected by [`RWLock`]
 #[derive(Debug)]
 #[must_use = "if unused the RWLock will immediately unlock"]
-pub struct ReadLockGuard<'a, T> {
+pub struct RawReadLockGuard<'a, T, I: IrqLocking> {
```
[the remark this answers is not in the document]
joergroedel: Actually, this gives me a compile error:
```
error[E0367]: `Drop` impl requires `I: locking::common::IrqLocking` but the struct it is implemented for does not
  --> kernel/src/locking/rwlock.rs:26:12
   |
26 | impl<T, I: IrqLocking> Drop for RawReadLockGuard<'_, T, I> {
   |            ^^^^^^^^^^
   |
note: the implementor must specify the same requirement
  --> kernel/src/locking/rwlock.rs:16:1
   |
16 | pub struct RawReadLockGuard<'a, T, I> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I keep the trait bound in the struct for now.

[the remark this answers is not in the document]
Freax13: Try removing the `RawReadLockGuard` from the `Drop` impl block.

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,242 @@
...
+    // SAFETY: The inline assembly just reads the processors RFLAGS register
+    // and does not change any state.
+    let state: u64 = unsafe {
+        let s: u64;
+        asm!("pushfq",
+             "popq {}",
+             out(reg) s,
+             options(att_syntax, nomem, preserves_flags));
+        s
+    };
```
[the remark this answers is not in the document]
joergroedel: Changed, thanks.

kernel/src/cpu/irq_state.rs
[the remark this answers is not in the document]
joergroedel: AFAIU marker traits like `Send` and `Sync` need to be explicitly set for a type, no? At least I didn't find a way in stable rust to set negative marker traits. Same for the comment below.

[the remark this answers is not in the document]
joergroedel: To be more specific, I tried:

```
impl !Send for IrqGuard {}
```

But that gives me this compile error in stable rust:

```
error[E0658]: negative trait bounds are not yet fully implemented; use marker types for now
   --> kernel/src/cpu/irq_state.rs:164:6
    |
164 | impl !Send for IrqGuard {}
    |      ^^^^^
    |
    = note: see issue #68318 <https://github.com/rust-lang/rust/issues/68318> for more information
```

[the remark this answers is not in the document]
Freax13: You can add a field of type `PhantomData<*const ()`. This will make the type `!Send` & `!Sync`. If you want a type to be `Send`, but not `Sync`, you can use `PhantomData<Cell<()>>`.

kernel/src/cpu/irq_state.rs
joergroedel: Okay, changed it like above just without the load.

kernel/src/cpu/irq_state.rs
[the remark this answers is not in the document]
joergroedel: Thanks. Made `IrqState` and `IrqGuard` both `!Send + `!Sync`. I am not sure what this means for Send and Sync traits for the locking structures, but that should only matter for the IRQ-safe variants anyway.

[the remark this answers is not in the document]
Freax13: > I am not sure what this means for Send and Sync traits for the locking structures, but that should only matter for the IRQ-safe variants anyway.

It shouldn't affect the locks themselves, but the guards for the IRQ-safe variants should now also be `!Send + !Sync` (We don't technically need the `!Sync` bound, but there's also no harm in it).

kernel/src/cpu/irq_state.rs
```diff
@@ -0,0 +1,242 @@
...
+/// This structure keeps track of PerCpu IRQ states. It tracks the original IRQ
+/// state and how deep IRQ-disable calls have been nested. The use of atomics
+/// is necessary for interior mutability and to make state modifications safe
+/// wrt. to IRQs.
+///
+/// The original state needs to be stored to not accidentially enable IRQs in
+/// contexts which have IRQs disabled by other means, e.g. in an exception or
+/// NMI/HV context.
+#[derive(Debug, Default)]
+pub struct IrqState {
```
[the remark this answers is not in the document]
joergroedel: Fixed.

joergroedel:
Pushed a new version which addresses all review comments.

kernel/src/cpu/idt/entry.S
```diff
@@ -213,6 +213,11 @@ handle_as_hv:
 	call 	process_hv_events
 	// fall through to default_return
 
+.globl return_new_task
+return_new_task:
+	call setup_new_task
+	jmp default_return
```
Freax13, line 219:
```suggestion
.globl return_new_task
return_new_task:
	sub rsp, 8
	call setup_new_task
	add rsp, 8
	jmp default_return
```

Make sure that the stack is correctly aligned.

kernel/src/locking/rwlock.rs
```diff
@@ -4,22 +4,26 @@
...
```
Freax13, line 16: See https://github.com/coconut-svsm/svsm/pull/441#discussion_r1724624363

kernel/src/locking/spinlock.rs
```diff
@@ -93,12 +107,14 @@ pub struct SpinLock<T> {
     /// protected data. That is, it allows the data to be accessed/modified
     /// while enforcing the locking mechanism.
     data: UnsafeCell<T>,
+    /// Use generic type I in the struct without consuming space.
+    phantom: PhantomData<I>,
```
Freax13, line 111: This makes the lock `!Send` and `!Sync` if `I` is `!Send` and `!Sync`.

Try this instead:
 ```suggestion
     phantom: PhantomData<fn(I)>,
```

File: kernel/src/cpu/idt/entry.S (hunk 17 above)
Comment (joergroedel) [reply kernel/src/cpu/idt/entry.S:219]: This code path is entered via `RET`, so the stack should be correctly aligned already.

Comment (Freax13) [reply kernel/src/cpu/idt/entry.S:219]: :+1:

File: kernel/src/locking/spinlock.rs (hunk 19 above)
Comment (Freax13) [reply kernel/src/locking/spinlock.rs:111]: No, it doesn't because we have explicit `Send` and `Sync` impls anyway. Please disregard my previous comment.

File: kernel/src/locking/rwlock.rs (hunk 18 above)
Comment (joergroedel) [reply kernel/src/locking/rwlock.rs:16]: Yes, that worked. Thanks.

**Review by Freax13 — APPROVED** [review]

File: kernel/src/locking/spinlock.rs (hunk 19 above)
Comment (joergroedel) [reply kernel/src/locking/spinlock.rs:111]: Okay, changed for `RawSpinLock` and `RawRWLock`.

File: kernel/src/locking/rwlock.rs [hunk 20]
```diff
@@ -28,26 +32,31 @@ impl<T> Drop for ReadLockGuard<'_, T> {
 
 /// Implements the behavior of dereferencing the [`ReadLockGuard`] to
 /// access the protected data.
-impl<T> Deref for ReadLockGuard<'_, T> {
+impl<T, I: IrqLocking> Deref for RawReadLockGuard<'_, T, I> {
```
Freax13, line 35:
```suggestion
impl<T, I> Deref for RawReadLockGuard<'_, T, I> {
```

kernel/src/locking/rwlock.rs
```diff
@@ -56,7 +65,7 @@ impl<T> Drop for WriteLockGuard<'_, T> {
 
 /// Implements the behavior of dereferencing the [`WriteLockGuard`] to
 /// access the protected data.
-impl<T> Deref for WriteLockGuard<'_, T> {
+impl<T, I: IrqLocking> Deref for RawWriteLockGuard<'_, T, I> {
```
Freax13, line 68:
```suggestion
impl<T, I> Deref for RawWriteLockGuard<'_, T, I> {
```

kernel/src/locking/rwlock.rs
```diff
@@ -65,26 +74,31 @@ impl<T> Deref for WriteLockGuard<'_, T> {
 
 /// Implements the behavior of dereferencing the [`WriteLockGuard`] to
 /// access the protected data in a mutable way.
-impl<T> DerefMut for WriteLockGuard<'_, T> {
+impl<T, I: IrqLocking> DerefMut for RawWriteLockGuard<'_, T, I> {
```
Freax13, line 77:
```suggestion
impl<T, I> DerefMut for RawWriteLockGuard<'_, T, I> {
```

kernel/src/locking/rwlock.rs
```diff
@@ -65,26 +74,31 @@ impl<T> Deref for WriteLockGuard<'_, T> {
...
 }
 
+pub type WriteLockGuard<'a, T> = RawWriteLockGuard<'a, T, IrqUnsafeLocking>;
+pub type WriteLockGuardIrqSafe<'a, T> = RawWriteLockGuard<'a, T, IrqSafeLocking>;
+
 /// A simple Read-Write Lock (RWLock) that allows multiple readers or
 /// one exclusive writer.
 #[derive(Debug)]
-pub struct RWLock<T> {
+pub struct RawRWLock<T, I: IrqLocking> {
```
Freax13, line 89:
```suggestion
pub struct RawRWLock<T, I> {
```

kernel/src/locking/rwlock.rs
```diff
@@ -65,26 +74,31 @@ impl<T> Deref for WriteLockGuard<'_, T> {
...
+    /// Silence unused type warning
+    phantom: PhantomData<fn(I)>,
 }
 
 /// Implements the trait `Sync` for the [`RWLock`], allowing safe
 /// concurrent access across threads.
-unsafe impl<T: Send> Send for RWLock<T> {}
-unsafe impl<T: Send + Sync> Sync for RWLock<T> {}
+unsafe impl<T: Send, I: IrqLocking> Send for RawRWLock<T, I> {}
+unsafe impl<T: Send + Sync, I: IrqLocking> Sync for RawRWLock<T, I> {}
```
Freax13, line 101:
```suggestion
unsafe impl<T: Send, I> Send for RawRWLock<T, I> {}
unsafe impl<T: Send + Sync, I> Sync for RawRWLock<T, I> {}
```

kernel/src/locking/spinlock.rs
```diff
@@ -27,13 +29,15 @@ use core::sync::atomic::{AtomicU64, Ordering};
...
+pub struct RawLockGuard<'a, T, I: IrqLocking = IrqUnsafeLocking> {
     holder: &'a AtomicU64,
     data: &'a mut T,
+    #[allow(dead_code)]
+    irq_state: I,
 }
 
 /// Implements the behavior of the [`LockGuard`] when it is dropped
-impl<T> Drop for LockGuard<'_, T> {
+impl<T, I: IrqLocking> Drop for RawLockGuard<'_, T, I> {
```
Freax13, line 40:
```suggestion
impl<T, I> Drop for RawLockGuard<'_, T, I> {
```

kernel/src/locking/spinlock.rs
```diff
@@ -27,13 +29,15 @@ use core::sync::atomic::{AtomicU64, Ordering};
 /// ```
 #[derive(Debug)]
 #[must_use = "if unused the SpinLock will immediately unlock"]
-pub struct LockGuard<'a, T> {
+pub struct RawLockGuard<'a, T, I: IrqLocking = IrqUnsafeLocking> {
```
Freax13, line 32:
```suggestion
pub struct RawLockGuard<'a, T, I = IrqUnsafeLocking> {
```

kernel/src/locking/spinlock.rs
```diff
@@ -42,7 +46,7 @@ impl<T> Drop for LockGuard<'_, T> {
 
 /// Implements the behavior of dereferencing the [`LockGuard`] to
 /// access the protected data.
-impl<T> Deref for LockGuard<'_, T> {
+impl<T, I: IrqLocking> Deref for RawLockGuard<'_, T, I> {
```
Freax13, line 49:
```suggestion
impl<T, I> Deref for RawLockGuard<'_, T, I> {
```

kernel/src/locking/spinlock.rs
```diff
@@ -52,14 +56,24 @@ impl<T> Deref for LockGuard<'_, T> {
 
 /// Implements the behavior of dereferencing the [`LockGuard`] to
 /// access the protected data in a mutable way.
-impl<T> DerefMut for LockGuard<'_, T> {
+impl<T, I: IrqLocking> DerefMut for RawLockGuard<'_, T, I> {
```
Freax13, line 59:
```suggestion
impl<T, I> DerefMut for RawLockGuard<'_, T, I> {
```

kernel/src/locking/spinlock.rs
```diff
@@ -81,7 +95,7 @@ impl<T> DerefMut for LockGuard<'_, T> {
 /// };
 /// ```
 #[derive(Debug, Default)]
-pub struct SpinLock<T> {
+pub struct RawSpinLock<T, I: IrqLocking = IrqUnsafeLocking> {
```
Freax13, line 98:
```suggestion
pub struct RawSpinLock<T, I = IrqUnsafeLocking> {
```

kernel/src/locking/spinlock.rs
```diff
@@ -93,12 +107,14 @@ pub struct SpinLock<T> {
...
     /// while enforcing the locking mechanism.
     data: UnsafeCell<T>,
+    /// Use generic type I in the struct without consuming space.
+    phantom: PhantomData<fn(I)>,
 }
 
-unsafe impl<T: Send> Send for SpinLock<T> {}
-unsafe impl<T: Send> Sync for SpinLock<T> {}
+unsafe impl<T: Send, I: IrqLocking> Send for RawSpinLock<T, I> {}
+unsafe impl<T: Send, I: IrqLocking> Sync for RawSpinLock<T, I> {}
```
Freax13, line 115:
```suggestion
unsafe impl<T: Send, I> Send for RawSpinLock<T, I> {}
unsafe impl<T: Send, I> Sync for RawSpinLock<T, I> {}
```

Freax13 (approved):

Freax13:
@joergroedel If you go into the "Files Changed" tab, there should be a "Add to batch" button next to suggestions. This allows you to add all suggestions to a batch and apply them with a single commit.

joergroedel:
> @joergroedel If you go into the "Files Changed" tab, there should be a "Add to batch" button next to suggestions. This allows you to add all suggestions to a batch and apply them with a single commit.

Thanks, but I need them as single commits to back-merge them into the respective commits of the series. I will of course add Co-authored tags.

Freax13 (approved):

PR #442, opened 2024-08-20
kernel/mm: Reduce stack footprint of page-table freeing

joergroedel:
Turns out that the way the iterators are implemented in RawPageTablePart::free_lvl*() the compiler makes copies of entire pages on the stack, which overflows the 32 KiB stack pretty quickly.

Iterate over references only and significantly reduce stack usage of the page-table freeing code.

### DIFF
--- a/kernel/src/mm/pagetable.rs
+++ b/kernel/src/mm/pagetable.rs
@@ -977,8 +977,8 @@ struct RawPageTablePart {
 impl RawPageTablePart {
     /// Frees a level 1 page table.
     fn free_lvl1(page: &PTPage) {
-        for entry in page.entries {
-            if let Some(page) = PTPage::from_entry(entry) {
+        for entry in page.entries.iter() {
+            if let Some(page) = PTPage::from_entry(*entry) {
                 // SAFETY: the page comes from an entry in the page table,
                 // which we allocated using `PTPage::alloc()`, so this is
                 // safe.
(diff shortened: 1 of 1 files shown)


kernel/src/mm/pagetable.rs
```diff
@@ -977,7 +977,7 @@ struct RawPageTablePart {
 impl RawPageTablePart {
     /// Frees a level 1 page table.
     fn free_lvl1(page: &PTPage) {
-        for entry in page.entries {
+        for &entry in &page.entries {
```
Freax13, line 980:
```suggestion
        for entry in page.entries.iter().copied() {
```
Nit: This could be a little more idiomatic.

Freax13 (approved):

kernel/src/mm/pagetable.rs
00xc: Doesn't this also cause the entry to be copied to the stack? I think omitting the `copied()` should do it

joergroedel: Yes, that is better, thanks. Updated the PR.

Freax13: > Doesn't this also cause the entry to be copied to the stack? I think omitting the `copied()` should do it

Yes, it causes the entry to be copied to the stack, but it doesn't cause all of them to be copied to the stack at the same time like they were previously. A single entry is only 8 bytes, and so that isn't a problem, but previously, we copied the entire page table, which is 4096 bytes, and that is a problem.

We can't emit copying the entry entirely because `PTPage::from_entry` wants a copy, not a reference.

(Also, the single entry will likely be kept in a register instead of on the stack.)

joergroedel: > Doesn't this also cause the entry to be copied to the stack? I think omitting the `copied()` should do it

Right, that reduces the stack usage even further (only 8 bytes, but still :) ). So updated again.

PR #443, opened 2024-08-21
mm: always use `this_cpu().get_pgtable()` instead of `get_init_pgtable_locked()`

Freax13:
`get_init_pgtable_locked()` was problematic because the returned mutable reference was aliasing the mutable reference returned in `this_cpu().get_pgtable()` in stage2.

Removing `get_init_pgtable_locked()` now will probably also make it easier to implement isolated pagetables in the future because everything is already only accessing `this_cpu().get_pgtable()`.


joergroedel (approved):
LGTM.

PR #445, opened 2024-08-22
scripts: add show-stack-sizes.sh

Freax13:
This adds a script to find functions with excessive stack sizes.

Follow-up for https://github.com/coconut-svsm/svsm/discussions/444

### DIFF
--- a/scripts/show-stack-sizes.sh
+++ b/scripts/show-stack-sizes.sh
@@ -1,0 +1,27 @@
+#!/bin/env sh
+# SPDX-License-Identifier: MIT OR Apache-2.0
+#
+# Author: Tom Dohrmann <erbse.13@gmx.de>
+# A script to find functions with excessive stack sizes.
+# Requires yq-go and llvm-readelf (bundled with llvm) to be installed.
+
+# Forcefully enable a nightly toolchain.
+export RUSTUP_TOOLCHAIN=nightly
+
+# Append -Z emit-stack-sizes to the set of rustflags. The RUSTFLAGS environment variable overrides the flags in the config.
+RUSTFLAGS=$(yq '.build.rustflags | join(" ")' .cargo/config.toml)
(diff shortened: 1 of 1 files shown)


scripts/show-stack-sizes.sh
```diff
@@ -0,0 +1,36 @@
...
+# Determine the path to the built binary.
+if [[ -z "${RELEASE}" ]]; then
+    TARGET_PATH=debug
+else
+    TARGET_PATH=release
+fi
+SVSM_PATH=target/x86_64-unknown-none/${TARGET_PATH}/svsm
+
+# Dump the binary into a yaml file.
+obj2yaml target/x86_64-unknown-none/${TARGET_PATH}/svsm > bin/svsm-kernel.yaml
```
joergroedel, line 33: Unfortunately the `obj2yaml` tool is not available on openSUSE :(

But some searching found that `llvm-readelf -C --stack-sizes` will also print the stack sizes. Can the script use this tool instead?

Freax13: Sure, I switched the script over to `llvm-readelf`.

scripts/show-stack-sizes.sh
```diff
@@ -0,0 +1,27 @@
...
+# Determine the path to the built binary.
+if [[ -z "${RELEASE}" ]]; then
+    TARGET_PATH=debug
+else
+    TARGET_PATH=release
+fi
+SVSM_PATH=target/x86_64-unknown-none/${TARGET_PATH}/svsm
+
+# Dump the binary into a yaml file.
+llvm-readelf -C --stack-sizes target/x86_64-unknown-none/${TARGET_PATH}/svsm
```
joergroedel, line 27:
```suggestion
# Print stack frame sizes for all functions, sorted from small to large
llvm-readelf -C --stack-sizes target/x86_64-unknown-none/${TARGET_PATH}/svsm | sort -n
```

PR #447, opened 2024-08-24
igvm: Add the Vanadium hypervisor to igvmbuilder

AdamCDunlap:
The Vanadium hypervisor is used internally by Google. It is similar to qemu in that it uses kvm.


roy-hopkins (approved):
The addition of the new configuration looks good to me, although I might question the need for it if we can make the Qemu and Vanadium configurations match.

PR #451, opened 2024-08-30
soundness improvements around hypervisor-shared memory

Freax13:
This PR improves the soundness of code around hypervisor-shared memory.

~The first patch, 174274d0e8ffb7757b74a42eef1082b05b8ea0a9, is blocked on https://github.com/google/zerocopy/pull/1601. Let me know if you want me to drop that patch if we don't want to wait on a new zerocopy release. I used the following patch to override zerocopy for testing:~
```diff
diff --git a/Cargo.toml b/Cargo.toml
index b7bdb46..d87444d 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,9 @@ zerocopy = { version = "0.7.32", features = ["alloc", "derive"] }
 # other repos
 packit = { git = "https://github.com/coconut-svsm/packit", version = "0.1.1" }
 
+[patch.crates-io]
+zerocopy = { git = "https://github.com/Freax13/zerocopy.git", rev = "68e1cc8" }
+
 [workspace.lints.rust]
 future_incompatible = { level = "deny", priority = 127 }
 nonstandard_style = { level = "deny", priority = 126 }
```

### DIFF
--- a/kernel/src/mm/page_visibility.rs
+++ b/kernel/src/mm/page_visibility.rs
@@ -35,7 +35,7 @@ use zerocopy::{FromBytes, FromZeroes};
 /// Converting the memory at `vaddr` must be safe within Rust's memory model.
 /// Notably any objects at `vaddr` must tolerate unsynchronized writes of any
 /// bit pattern.
-pub unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
+unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
     let platform = SVSM_PLATFORM.as_dyn_ref();
 
     // Revoke page validation before changing page state.
@@ -68,7 +68,7 @@ pub unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
 /// # Arguments
 ///
 /// * `vaddr` - The virtual address of the page to be made private.
-pub unsafe fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
+unsafe fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
     // Update the page tables to map the page as private.
     this_cpu().get_pgtable().set_encrypted_4k(vaddr)?;
     flush_tlb_global_sync();
@@ -147,6 +147,13 @@ impl<T> SharedBox<T> {
                 size_of::<T>(),
             );
         }
+    }
+
+    /// Leak the memory.
+    pub fn leak(self) -> NonNull<T> {
+        let ptr = self.ptr;
+        core::mem::forget(self);
+        ptr
     }
 }
 
--- a/kernel/src/mm/page_visibility.rs
+++ b/kernel/src/mm/page_visibility.rs
@@ -4,17 +4,24 @@
 //
 // Author: Jon Lange (jlange@microsoft.com)
 
+use core::mem::MaybeUninit;
+use core::ptr::NonNull;
+
 use crate::address::VirtAddr;
 use crate::cpu::flush_tlb_global_sync;
+use crate::cpu::mem::{copy_bytes, write_bytes};
 use crate::cpu::percpu::this_cpu;
 use crate::error::SvsmError;
 use crate::mm::validate::{
     valid_bitmap_clear_valid_4k, valid_bitmap_set_valid_4k, valid_bitmap_valid_addr,
 };
-use crate::mm::virt_to_phys;
+use crate::mm::{virt_to_phys, PageBox};
 use crate::platform::{PageStateChangeOp, SVSM_PLATFORM};
+use crate::protocols::errors::SvsmReqError;
 use crate::types::{PageSize, PAGE_SIZE};
 use crate::utils::MemoryRegion;
+
+use zerocopy::{FromBytes, FromZeroes};
 
 /// Makes a virtual page shared by revoking its validation, updating the
 /// page state, and modifying the page tables accordingly.
@@ -22,7 +29,13 @@ use crate::utils::MemoryRegion;
 /// # Arguments
 ///
 /// * `vaddr` - The virtual address of the page to be made shared.
-pub fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
+///
+/// # Safety
+///
+/// Converting the memory at `vaddr` must be safe within Rust's memory model.
+/// Notably any objects at `vaddr` must tolerate unsynchronized writes of any
+/// bit pattern.
+pub unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
     let platform = SVSM_PLATFORM.as_dyn_ref();
 
     // Revoke page validation before changing page state.
@@ -55,7 +68,7 @@ pub fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
 /// # Arguments
 ///
 /// * `vaddr` - The virtual address of the page to be made private.
-pub fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
+pub unsafe fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
     // Update the page tables to map the page as private.
     this_cpu().get_pgtable().set_encrypted_4k(vaddr)?;
     flush_tlb_global_sync();
@@ -77,4 +90,128 @@ pub fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
     }
 
     Ok(())
+}
+
+/// SharedBox is a safe wrapper around memory pages shared with the host.
+pub struct SharedBox<T> {
+    ptr: NonNull<T>,
+}
+
+impl<T> SharedBox<T> {
+    /// Allocate some memory and share it with the host.
+    pub fn try_new_zeroed() -> Result<Self, SvsmError> {
+        let page_box = PageBox::<T>::try_new_zeroed()?;
+        let vaddr = page_box.vaddr();
+
+        let ptr = NonNull::from(PageBox::leak(page_box)).cast::<T>();
+
+        for offset in (0..core::mem::size_of::<T>()).step_by(PAGE_SIZE) {
+            unsafe {
+                make_page_shared(vaddr + offset)?;
+            }
+        }
+
+        Ok(Self { ptr })
+    }
+
+    /// Returns the virtual address of the memory.
+    pub fn addr(&self) -> VirtAddr {
+        VirtAddr::from(self.ptr.as_ptr())
+    }
+
+    /// Read the currently stored value into `out`.
+    pub fn read_into(&self, out: &mut T)
+    where
+        T: FromBytes + Copy,
+    {
+        unsafe {
+            // SAFETY: `self.ptr` is valid. Any bitpattern is valid for `T`.
+            copy_bytes(
+                self.ptr.as_ptr() as usize,
+                out as *const T as usize,
+                size_of::<T>(),
+            );
+        }
+    }
+
+    /// Share `value` with the host.
+    pub fn write_from(&mut self, value: &T)
+    where
+        T: Copy,
+    {
+        unsafe {
+            // SAFETY: `self.ptr` is valid..
+            copy_bytes(
+                value as *const T as usize,
+                self.ptr.as_ptr() as usize,
+                size_of::<T>(),
+            );
+        }
+    }
+}
+
+impl<T, const N: usize> SharedBox<[T; N]> {
+    /// Clear the first `n` elements.
+    pub fn nclear(&mut self, n: usize) -> Result<(), SvsmReqError>
+    where
+        T: FromZeroes,
+    {
+        if n > N {
+            return Err(SvsmReqError::invalid_parameter());
+        }
+
+        unsafe {
+            // SAFETY: `self.ptr` is valid and we did a bounds check on `n`.
+            write_bytes(self.ptr.as_ptr() as usize, size_of::<T>() * n, 0);
+        }
+
+        Ok(())
+    }
+
+    /// Fill up the `outbuf` slice provided with bytes from data
+    pub fn copy_to_slice(&self, outbuf: &mut [T]) -> Result<(), SvsmReqError>
+    where
+        T: FromBytes + Copy,
+    {
+        if outbuf.len() > N {
+            return Err(SvsmReqError::invalid_parameter());
+        }
+
+        let size = core::mem::size_of_val(outbuf);
+        unsafe {
+            // SAFETY: `self.ptr` is valid.
+            copy_bytes(
+                self.ptr.as_ptr() as usize,
+                outbuf.as_mut_ptr() as usize,
+                size,
+            );
+        }
+
+        Ok(())
+    }
+}
+
+unsafe impl<T> Send for SharedBox<T> where T: Send {}
+unsafe impl<T> Sync for SharedBox<T> where T: Sync {}
+
+impl<T> Drop for SharedBox<T> {
+    fn drop(&mut self) {
+        // Re-encrypt the pages.
+        let res = (0..size_of::<Self>())
+            .step_by(PAGE_SIZE)
+            .try_for_each(|offset| unsafe { make_page_private(self.addr() + offset) });
+
+        // If re-encrypting was successful free the memory otherwise leak it.
+        if res.is_ok() {
+            drop(unsafe { PageBox::from_raw(self.ptr.cast::<MaybeUninit<T>>()) });
+        } else {
+            log::error!("failed to set pages to encrypted. Memory leak!");
+        }
+    }
+}
+
+impl<T> core::fmt::Debug for SharedBox<T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("SharedBox").finish_non_exhaustive()
+    }
 }
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -23,7 +23,7 @@ use crate::utils::MemoryRegion;
 use crate::mm::PageBox;
 use core::arch::global_asm;
 use core::mem::{self, offset_of};
-use core::ops::{Deref, DerefMut};
+use core::ops::Deref;
 use core::ptr;
 use core::sync::atomic::{AtomicU16, AtomicU32, AtomicU64, AtomicU8, Ordering};
 
@@ -170,12 +170,6 @@ impl Deref for GhcbPage {
     type Target = GHCB;
     fn deref(&self) -> &Self::Target {
         self.0.deref()
-    }
-}
-
-impl DerefMut for GhcbPage {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        self.0.deref_mut()
     }
 }
 
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -162,7 +162,28 @@ impl GhcbPage {
 
 impl Drop for GhcbPage {
     fn drop(&mut self) {
-        self.0.shutdown().expect("Could not shut down GHCB");
+        let vaddr = self.0.vaddr();
+        let paddr = virt_to_phys(vaddr);
+
+        // Re-encrypt page
+        this_cpu()
+            .get_pgtable()
+            .set_encrypted_4k(vaddr)
+            .expect("Could not re-encrypt page");
+
+        // Unregister GHCB PA
+        register_ghcb_gpa_msr(PhysAddr::null()).expect("Could not unregister GHCB");
+
+        // Ask the hypervisor to change the page back to the private page state.
+        validate_page_msr(paddr).expect("Could not change page state");
+
+        // Make page guest-valid
+        pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid).expect("Could not pvalidate page");
+
+        // Needs guarding for Stage2 GHCB
+        if valid_bitmap_valid_addr(paddr) {
+            valid_bitmap_set_valid_4k(paddr);
+        }
     }
 }
 
@@ -311,30 +332,6 @@ impl GHCB {
 
         // Register GHCB GPA
         Ok(register_ghcb_gpa_msr(paddr)?)
-    }
-
-    pub fn shutdown(&self) -> Result<(), SvsmError> {
-        let vaddr = VirtAddr::from(ptr::from_ref(self));
-        let paddr = virt_to_phys(vaddr);
-
-        // Re-encrypt page
-        this_cpu().get_pgtable().set_encrypted_4k(vaddr)?;
-
-        // Unregister GHCB PA
-        register_ghcb_gpa_msr(PhysAddr::null())?;
-
-        // Make page guest-invalid
-        validate_page_msr(paddr)?;
-
-        // Make page guest-valid
-        pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid)?;
-
-        // Needs guarding for Stage2 GHCB
-        if valid_bitmap_valid_addr(paddr) {
-            valid_bitmap_set_valid_4k(paddr);
-        }
-
-        Ok(())
     }
 
     pub fn clear(&self) {
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -22,16 +22,18 @@ use crate::utils::MemoryRegion;
 
 use crate::mm::PageBox;
 use core::arch::global_asm;
-use core::cell::Cell;
 use core::mem::{self, offset_of};
 use core::ops::{Deref, DerefMut};
 use core::ptr;
+use core::sync::atomic::{AtomicU16, AtomicU32, AtomicU64, AtomicU8, Ordering};
 
 use super::msr_protocol::{invalidate_page_msr, register_ghcb_gpa_msr, validate_page_msr};
 use super::{pvalidate, PvalidateOp};
 
+use zerocopy::AsBytes;
+
 #[repr(C, packed)]
-#[derive(Debug, Default, Clone, Copy)]
+#[derive(Debug, Default, Clone, Copy, AsBytes)]
 pub struct PageStateChangeHeader {
     cur_entry: u16,
     end_entry: u16,
@@ -56,7 +58,7 @@ macro_rules! ghcb_getter {
         #[allow(unused)]
         fn $name(&self) -> Result<$t, GhcbError> {
             self.is_valid(offset_of!(Self, $field))
-                .then(|| self.$field.get())
+                .then(|| self.$field.load(Ordering::Relaxed))
                 .ok_or(GhcbError::VmgexitInvalid)
         }
     };
@@ -66,7 +68,7 @@ macro_rules! ghcb_setter {
     ($name:ident, $field:ident, $t:ty) => {
         #[allow(unused)]
         fn $name(&self, val: $t) {
-            self.$field.set(val);
+            self.$field.store(val, Ordering::Relaxed);
             self.set_valid(offset_of!(Self, $field));
         }
     };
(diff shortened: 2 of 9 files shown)


Freax13:
> Please move that patch to a separate draft-PR, which you can then "undraft" once all blockers are solved.

Done.

Freax13:
Just rebased onto main. I resolved the `TODO`s by switching to the functions in `crate::cpu::mem`.

kernel/src/stage2.rs
```diff
@@ -63,10 +63,9 @@ fn init_percpu(platform: &mut dyn SvsmPlatform) -> Result<(), SvsmError> {
     Ok(())
 }
 
-fn shutdown_percpu() {
-    this_cpu()
-        .shutdown()
-        .expect("Failed to shut down percpu data (including GHCB)");
+unsafe fn shutdown_percpu() {
```
msft-jlange, line 66: Why is this function `unsafe`?  At a minimum, there should be a Safety comment here.  But there is also nothing about the function declaration that suggests that it can only be called from unsafe code.  My understanding of the convention we have been using is that a function should only be declared `unsafe` if it is not possible to call it from safe code due to its parameters or return value (this was the position advocated by @00xc as I always understood it).

kernel/src/cpu/mem.rs
```diff
@@ -37,3 +37,54 @@ pub unsafe fn write_bytes(dst: usize, size: usize, value: u8) {
...
 }
+
+/// Returns whether there are `size` null-bytes at `src`.
+///
+/// # Safety
+///
+/// This function has all the safety requirements of `core::ptr::read` except
+/// that data races are explicitly permitted.
+#[inline(always)]
+pub unsafe fn is_clear(src: usize, size: usize) -> bool {
```
msft-jlange, line 48: This appears to be used only by test code.  Should it therefore be within the test module so it is not accidentally referenced by non-test code?

kernel/src/mm/page_visibility.rs
```diff
@@ -4,25 +4,38 @@
...
 ///
 /// # Arguments
 ///
 /// * `vaddr` - The virtual address of the page to be made shared.
-pub fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
+///
+/// # Safety
+///
+/// Converting the memory at `vaddr` must be safe within Rust's memory model.
+/// Notably any objects at `vaddr` must tolerate unsychronized writes of any
```
msft-jlange, line 36: Typo: `unsychronized` => `unsynchronized`.

kernel/src/sev/ghcb.rs
```diff
@@ -160,7 +162,28 @@ impl GhcbPage {
...
+        let paddr = virt_to_phys(vaddr);
+
+        // Re-encrypt page
+        this_cpu()
+            .get_pgtable()
+            .set_encrypted_4k(vaddr)
+            .expect("Could not re-encrypt page");
+
+        // Unregister GHCB PA
+        register_ghcb_gpa_msr(PhysAddr::null()).expect("Could not unregister GHCB");
```
msft-jlange, line 175: This is not guaranteed to succeed, because the VMM is not required to accept NULL as a valid GHCB page.

kernel/src/sev/ghcb.rs
```diff
@@ -160,7 +162,28 @@ impl GhcbPage {
 
 impl Drop for GhcbPage {
     fn drop(&mut self) {
```
msft-jlange, line 164: Is there any scenario for which dropping a GHCB is not associated with a fatal termination of the SVSM?  The process of restoring a GHCB page to the private state is fragile, and I believe it would be best to avoid any attempt to do so unless we were aware of a valid use case for this.  If we cannot come up with one, then it would be simplest for `GhcbPage::drop()` simply to panic, since we shouldn't get here anyway.

kernel/src/sev/ghcb.rs
```diff
@@ -317,33 +334,10 @@ impl GHCB {
...
-            valid_bitmap_set_valid_4k(paddr);
-        }
-
-        Ok(())
-    }
-
     pub fn clear(&self) {
         // Clear valid bitmap
-        self.valid_bitmap.set([0, 0]);
+        self.valid_bitmap[0].store(0, Ordering::SeqCst);
```
msft-jlange, line 339: The GHCB page is not manipulated across processors, so there is no reason for expensive memory barriers when manipulating its contents.  The only race conditions we might expect are between the SVSM environment and the host running on the same processor; this possibility for interruption means that atomic operations are necessary, but `Ordering::Relaxed` will be sufficient for all operations.  That is true everywhere throughout this file.

kernel/src/mm/page_visibility.rs
```diff
@@ -78,3 +91,139 @@ pub fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
 
     Ok(())
 }
+
+/// SharedBox is a safe wrapper around memory pages shared with the host.
+pub struct SharedBox<T> {
```
msft-jlange, line 96: This structure appears to have nothing to do with GHCB - the `GhcbPage` structure doesn't even make use of it - and thus it should be in its own source file and not stuck in the GHCB sources.  In addition, the notion of `SharedBox` will be used on other confidential platforms (like TDX) and definitely should not be in the SEV sources directory.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -47,51 +46,28 @@ pub struct HVExtIntInfo {
...
-
-        // Now Drop will have correct behavior, so construct the new type.
-        // SAFETY: all zeros is a valid representation of the HV doorbell page.
-        let page = unsafe { Self(page.assume_init()) };
-        ghcb.register_hv_doorbell(paddr)?;
-        Ok(page)
-    }
+/// Allocates a new HV doorbell page and registers it on the hypervisor
+/// using the given GHCB.
+pub fn allocate_hv_doorbell_page(ghcb: &GHCB) -> Result<&'static HVDoorbell, SvsmError> {
```
msft-jlange, line 51: Why does this function not use `SharedBox`?  It seems that `SharedBox` is designed to do exactly the sort of assignment and visibility management expected here, and unlike the GHCB, there is no chicken-and-egg problem because `SharedBox` only requires the existence of a GHCB, not of a doorbell page.

kernel/src/mm/page_visibility.rs
Freax13: Fixed, thanks!

kernel/src/stage2.rs
Freax13: > Why is this function `unsafe`? At a minimum, there should be a Safety comment here.

I added a safety section.

> But there is also nothing about the function declaration that suggests that it can only be called from unsafe code. My understanding of the convention we have been using is that a function should only be declared `unsafe` if it is not possible to call it from safe code due to its parameters or return value (this was the position advocated by @00xc as I always understood it).

No, that's not how `unsafe` is used in Rust: A function must be marked as `unsafe` if (and usually only if) safe code using the function can cause undefined behavior (see the [Rustonomicon](https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html)). Whether or not a function must be marked as unsafe has very little to do with its signature and much more to do with what it actually does (you could make any unsafe function with any signature safe by replacing it's body with `loop {}`). In this case, the caller must ensure that the `PerCpu` instance isn't used after it's been dropped because doing so is UB. This isn't something that the compiler can enforce, so we have to mark the function as unsafe and shift the responsibility of upholding this to the user. This isn't a convention that's specified by us in this project, those are the rules for writing unsafe code in Rust.

kernel/src/sev/ghcb.rs
Freax13: That's true, but this PR didn't introduce this code, it merely moved it around. We were already panicking before this PR. We should fix this in a later PR.

kernel/src/sev/ghcb.rs
Freax13: > Is there any scenario for which dropping a GHCB is not associated with a fatal termination of the SVSM? The process of restoring a GHCB page to the private state is fragile, and I believe it would be best to avoid any attempt to do so unless we were aware of a valid use case for this. If we cannot come up with one, then it would be simplest for `GhcbPage::drop()` simply to panic, since we shouldn't get here anyway.

Yes, calling the `shutdown_percpu` function will cause the `GhcbPage` to be dropped. We do this to release the GHCB used by stage2 shortly before entering the svsm kernel. That said, this code wasn't introduced by this PR, this PR just moved that code around.

If we didn't need this, I'd would totally be on your side, we shouldn't write fragile code when we shouldn't need it in practice.

kernel/src/sev/ghcb.rs
Freax13: Sure, sounds reasonable.

kernel/src/mm/page_visibility.rs
Freax13: I'm a little confused, `SharedBox` isn't declared in a file that has anything to do with the GHCB or SEV specific sources. It's declared in the same file as two other functions that a related to page visibility and that file is already platform independent.

kernel/src/sev/hv_doorbell.rs
Freax13: Good idea! I changed the function to use `SharedBox`.

kernel/src/cpu/mem.rs
Freax13: > This appears to be used only by test code.

This function is also referenced by `SharedBox::is_clear` which itself is used by `send_extended_guest_request`.

kernel/src/stage2.rs
msft-jlange: I agree with your perspective on the use of `unsafe`.  However, I once submitted a PR to this project that declared a function `unsafe` for similar reasons, and received feedback from @00xc that the function should not be declared this way, for the reasons I included in my earlier comment.  I'm happy with your approach now that you've added a safety comment, and will leave it to @00xc to comment on whether this use of `unsafe` is appropriate.

kernel/src/mm/page_visibility.rs
msft-jlange: Sorry, somehow I observed this code to be in kernel/src/sev/ghcb.rs.  The current location is great.

kernel/src/cpu/mem.rs
msft-jlange: If this function is required for production code, then why is it written in assembly?  In the interest of minimizing the use of unsafe code, it seems like we don't want to write anything in assembly unless it has to be done that way, and at a glance it appears that this function could be implemented in Rust instead of asm.  If you have a strong justification for the use of assembly, then can you include that justification in the safety comments?

If performance is your primary concern, then there are many ways this loop could be improved (switching to SCASQ at a minimum, and better yet would be to use XMM).  But if performance is not a primary concern, then avoiding assembly altogether would be best.

kernel/src/sev/ghcb.rs
msft-jlange: That is a reasonable position.

kernel/src/sev/ghcb.rs
msft-jlange: I'll take another look at this flow after your PR merges to see if we can be more robust here, but I accept your point that your PR isn't really changing the logic that exists today.

kernel/src/cpu/mem.rs
Freax13: Performance is not my primary concern.

The problem is that Rust doesn't have atomic volatile read/write ops. I tried to hint at this in the safety section by mentioning that synchronization is not required. Using inline assembly was one of the work-arounds suggested in https://github.com/rust-lang/unsafe-code-guidelines/issues/321.

Freax13: The operations on `AtomicU64` aren't volatile.

msft-jlange: Perhaps you and I are using different definitions of "volatile".  I am assuming "volatile" to mean that the compiler is guaranteed to make only a single access to data, at a known point in time relative to other data accesses (i.e. the access is explicitly ordered).  This behavior is exactly what is implied by `Ordered`, with the distinction that `Ordering::Relaxed` removes the microarchitectural ordering guarantees (the processor is free to reorder reads/writes) but retains the compilation guarantees (the compiler will emit the access in precisely the correct location).  Atomic accesses would be completely broken if the compiler were free to reorder reads and writes.  Do you mean something different by "volatile"?

Freax13: > Perhaps you and I are using different definitions of "volatile". I am assuming "volatile" to mean that the compiler is guaranteed to make only a single access to data, at a known point in time relative to other data accesses (i.e. the access is explicitly ordered). 

I agree with you on that definition so far.

> This behavior is exactly what is implied by `Ordered`, with the distinction that `Ordering::Relaxed` removes the microarchitectural ordering guarantees (the processor is free to reorder reads/writes) but retains the compilation guarantees (the compiler will emit the access in precisely the correct location).

I'm not aware of any such guarantee regarding codegen. To quote the LLVM docs on [Atomics](https://llvm.org/docs/Atomics.html):
> Atomic and volatile in the IR are orthogonal; “volatile” is the C/C++ volatile, which ensures that every volatile load and store happens and is performed in the stated order. A couple examples: if a SequentiallyConsistent store is immediately followed by another SequentiallyConsistent store to the same address, the first store can be erased. This transformation is not allowed for a pair of volatile stores. 

Furthermore, AFAIK, reordering is generally permitted by the as-if rule, except in cases where stricter `Ordering`s are used.

> Atomic accesses would be completely broken if the compiler were free to reorder reads and writes. Do you mean something different by "volatile"?

The different `Ordering` variants exist precisely to make some guarantees about the orderings allowed for reads and writes, but they don't always prevent other optimization such as merges that aren't allowed for volatile accesses.

msft-jlange: So what guarantees are you expecting that you believe volatile will achieve but ordered atomics will not?  You're not accessing I/O space memory, so are you not requiring the compiler to generate code to specifically issue bus cycles for all of your addresses.  As far as I can tell, you just want synchronization guarantees to ensure that the data you observe for zero-checking is actually observed, and that's exactly what atomics are intended to solve.  What are you concerned may happen incorrectly through the use of atomics?

Assembly-language code is difficult to maintain and should be avoided at all costs except when it implements deeply architecture-specific behavior (like stack management, interrupt dispatch, and the like).  Since you're just looking for a way to check for zeroes, it's worth looking long and hard for something that can achieve your goal without resorting to non-portable and difficult assembly code.

Freax13: > So what guarantees are you expecting that you believe volatile will achieve but ordered atomics will not? You're not accessing I/O space memory, so are you not requiring the compiler to generate code to specifically issue bus cycles for all of your addresses. As far as I can tell, you just want synchronization guarantees to ensure that the data you observe for zero-checking is actually observed, and that's exactly what atomics are intended to solve. What are you concerned may happen incorrectly through the use of atomics?

My concern is that the compiler will optimize out accesses. My understanding is that volatile and atomic accesses are different, so I'm wary of using atomics to emulate volatile accesses.

> Assembly-language code is difficult to maintain and should be avoided at all costs except when it implements deeply architecture-specific behavior (like stack management, interrupt dispatch, and the like). Since you're just looking for a way to check for zeroes, it's worth looking long and hard for something that can achieve your goal without resorting to non-portable and difficult assembly code.

:+1:

msft-jlange: > My concern is that the compiler will optimize out accesses. My understanding is that volatile and atomic accesses are different, so I'm wary of using atomics to emulate volatile accesses.

The compiler cannot optimize away accesses that contribute to the value graph.  The goal of this function appears to be to scan a slice to see whether it is all zeroes, and therefore the result depends on every entry in the page.  The compiler cannot possibly generate the correct result if it skips over some of the values.  The only way the compiler could optimize this out is if the result of `is_clear()` is ignored, in which case the entire function call could be skipped - but if you don't care about the result, then why exercise the code?

If you were to restructure the loop so that it reached every byte in the page (in larger groups, like `u64`s when possible) as an atomic, and then did a bitwise-OR of every result of `Atomic::load()`, then every load would be required because it contributes the result.  Then the question becomes whether the compiler attempts to load from the same place multiple times as part of its register optimization, but the compiler is not permitted to do that with atomics because atomics are presumed to have different values on each load, and thus atomics cannot lead to compiler-generated TOCTOU confusion.

Freax13: Can the compiler optimize out the load in this code?
```rust
fn foo(bar: &AtomicU64) -> u64 {
    bar.store(1, Ordering::SeqCst);
    bar.load(Ordering::SeqCst)
}
```

In my understanding of atomics, the compiler is allowed to optimize out the load because it's directly preceded by a write to the same address. This wouldn't be allowed for a volatile access.

Freax13: > [...] in larger groups, like `u64`s when possible [...]

Unfortunately, different-sized atomic accesses to the same data are UB (see the 4th example in the first example code block [here](https://doc.rust-lang.org/std/sync/atomic/)), so we'd always have to use `AtomicU8`. We might be able to work around this by requiring accesses to a `SharedBox` to happen through mutable references.

Freax13: > Can the compiler optimize out the load in this code?
> 
> ```rust
> fn foo(bar: &AtomicU64) -> u64 {
>     bar.store(1, Ordering::SeqCst);
>     bar.load(Ordering::SeqCst)
> }
> ```
> 
> In my understanding of atomics, the compiler is allowed to optimize out the load because it's directly preceded by a write to the same address. This wouldn't be allowed for a volatile access.

Hm, if we place an empty inline assembly block (e.g.  `asm!("/* {} */", in(reg) bar);`) between the store and the load, the compiler wouldn't be allowed to do this optimization cause it must assume that the inline assembly mutates `bar`. @msft-jlange would you be okay with using atomic ops like you suggested and preceding them with an empty inline assembly block (and maybe also following them with an empty inline assembly block for write ops)? While this technically still uses inline assembly, we wouldn't run into portability issues because the inline assembly block is empty (except for the comment). Other than forcing the compiler to to allocate a register for the inline assembly block, this shouldn't affect codegen much.

msft-jlange: > Can the compiler optimize out the load in this code?
> 
> ```rust
> fn foo(bar: &AtomicU64) -> u64 {
>     bar.store(1, Ordering::SeqCst);
>     bar.load(Ordering::SeqCst)
> }
> ```
> 
> In my understanding of atomics, the compiler is allowed to optimize out the load because it's directly preceded by a write to the same address. This wouldn't be allowed for a volatile access.

The compiler can optimize away the load because the result is ignored.  It has nothing to do with whether the addresses are the same or different.  However, in your loop, the result of the atomic load is not ignored; it's used to determine whether the function returns `true` or `false`.  This creates a dependency on the data fetched by the load, and when there is a data dependency, the result cannot be ignored unless the entire data graph can be ignored.

> Unfortunately, different-sized atomic accesses to the same data are UB (see the 4th example in the first example code block [here](https://doc.rust-lang.org/std/sync/atomic/)), so we'd always have to use `AtomicU8`. We might be able to work around this by requiring accesses to a `SharedBox` to happen through mutable references.

It turns out that all modern processors will do this correctly because of the way atomics are typically implemented in the cache protocol.  I think there may be no way to avoid that problem here, because any previous data writes to the page being scanned may have used accesses other than 8-bit accesses, thus causing the size conflict you're worried about.  But in any case, you've pointed out that performance of this loop is not critical and therefore scanning it in one-byte increments won't be a problem.

> Hm, if we place an empty inline assembly block (e.g. `asm!("/* {} */", in(reg) bar);`) between the store and the load, the compiler wouldn't be allowed to do this optimization cause it must assume that the inline assembly mutates `bar`. @msft-jlange would you be okay with using atomic ops like you suggested and preceding them with an empty inline assembly block (and maybe also following them with an empty inline assembly block for write ops)? While this technically still uses inline assembly, we wouldn't run into portability issues because the inline assembly block is empty (except for the comment). Other than forcing the compiler to to allocate a register for the inline assembly block, this shouldn't affect codegen much.

This construction solves my primary concern about portability, so I wouldn't object to it if it were necessary.  However, I'm still trying to understand why you think it's necessary given that your loop seems to require the consumption of all of the data in the page, and thus the compiler cannot optimize anything away.

If we were to require true volatile behavior, then a better approach would be to define a platform-specific `unsafe fn read_byte_volatile(b: *const u8) -> u8` and then call it from a platform-independent loop.

Freax13: > > Can the compiler optimize out the load in this code?
> > ```rust
> > fn foo(bar: &AtomicU64) -> u64 {
> >     bar.store(1, Ordering::SeqCst);
> >     bar.load(Ordering::SeqCst)
> > }
> > ```
> > 
> > In my understanding of atomics, the compiler is allowed to optimize out the load because it's directly preceded by a write to the same address. This wouldn't be allowed for a volatile access.
> 
> The compiler can optimize away the load because the result is ignored. It has nothing to do with whether the addresses are the same or different. However, in your loop, the result of the atomic load is not ignored; it's used to determine whether the function returns `true` or `false`. This creates a dependency on the data fetched by the load, and when there is a data dependency, the result cannot be ignored unless the entire data graph can be ignored.

The value is not being ignored, it's returned, so it's part of the data graph (at least as far as `foo` is concerned). Still, I claim that the load can be optimized out.

> > Unfortunately, different-sized atomic accesses to the same data are UB (see the 4th example in the first example code block [here](https://doc.rust-lang.org/std/sync/atomic/)), so we'd always have to use `AtomicU8`. We might be able to work around this by requiring accesses to a `SharedBox` to happen through mutable references.
> 
> It turns out that all modern processors will do this correctly because of the way atomics are typically implemented in the cache protocol. I think there may be no way to avoid that problem here, because any previous data writes to the page being scanned may have used accesses other than 8-bit accesses, thus causing the size conflict you're worried about. But in any case, you've pointed out that performance of this loop is not critical and therefore scanning it in one-byte increments won't be a problem.

:+1: 

> > Hm, if we place an empty inline assembly block (e.g. `asm!("/* {} */", in(reg) bar);`) between the store and the load, the compiler wouldn't be allowed to do this optimization cause it must assume that the inline assembly mutates `bar`. @msft-jlange would you be okay with using atomic ops like you suggested and preceding them with an empty inline assembly block (and maybe also following them with an empty inline assembly block for write ops)? While this technically still uses inline assembly, we wouldn't run into portability issues because the inline assembly block is empty (except for the comment). Other than forcing the compiler to to allocate a register for the inline assembly block, this shouldn't affect codegen much.
> 
> This construction solves my primary concern about portability, so I wouldn't object to it if it were necessary.

:+1: 

> However, I'm still trying to understand why you think it's necessary given that your loop seems to require the consumption of all of the data in the page, and thus the compiler cannot optimize anything away.
>  
> If we were to require true volatile behavior, then a better approach would be to define a platform-specific `unsafe fn read_byte_volatile(b: *const u8) -> u8` and then call it from a platform-independent loop.

That's sort of what I was trying to achieve here with this function (as well as the two functions above it), except that I implemented operations for memory regions instead of a single byte. If other people are okay with this, I'd also be fine with reimplementing them in terms of `read_byte_volatile` and `write_byte_volatile`.

msft-jlange: > The value is not being ignored, it's returned, so it's part of the data graph (at least as far as `foo` is concerned). Still, I claim that the load can be optimized out.

Sorry, I misread your sample code (I read it as returning `()` instead of returning `u64`).  If the return value is part of the data graph, then how can the compiler omit the load and still produce the correct result?  If the ultimate value of the data graph is ignored, then the entire graph can be discarded, but that doesn't appear to be what we're talking about here.  The point of `volatile` is to ensure that the memory access occurs so that its side-effects are observed (like MMIO), but it doesn't impact any ability to collect values into the data graph, except for the additional property of relaxed load/store ordering (which can also be solved with atomics).

Perhaps I am missing something fundamental about the intention of your loop, which is not surprising given that it is in assembly and its intended behavior is not documented (other than a comment about data races).  Your function appears to be exactly equivalent to this:

```
fn check_zeroed(data: [u8]) -> bool {
    for byte in data.iter() {
        if byte != 0 {
            return false;
        }
    true
}
```

Am I reverse engineering your assembly incorrectly?  Is there something you're trying to accomplish other than checking for zeroes?

>> If we were to require true volatile behavior, then a better approach would be to define a platform-specific `unsafe fn read_byte_volatile(b: *const u8) -> u8` and then call it from a platform-independent loop.
>
> That's sort of what I was trying to achieve here with this function (as well as the two functions above it), except that I implemented operations for memory regions instead of a single byte. If other people are okay with this, I'd also be fine with reimplementing them in terms of `read_byte_volatile` and `write_byte_volatile`.

Given that the intent of the loop is not entirely clear, it would be good to minimize the amount of assembly only to the specific operations that are platform-specific, so that all of the rest of the logic is in Rust.  This also minimizes the possibility that the loop behavior differs from one platform to another.

Freax13: > > The value is not being ignored, it's returned, so it's part of the data graph (at least as far as `foo` is concerned). Still, I claim that the load can be optimized out.
> 
> Sorry, I misread your sample code (I read it as returning `()` instead of returning `u64`). If the return value is part of the data graph, then how can the compiler omit the load and still produce the correct result?

IIUC the compiler is allowed to just return `1` because it knows that the load was immediately preceded by a store to the same address with the value 1. This wouldn't be allowed for a volatile access, but for non-volatile atomics, AFAICT this is legal.

> If the ultimate value of the data graph is ignored, then the entire graph can be discarded, but that doesn't appear to be what we're talking about here. The point of `volatile` is to ensure that the memory access occurs so that its side-effects are observed (like MMIO), but it doesn't impact any ability to collect values into the data graph, except for the additional property of relaxed load/store ordering (which can also be solved with atomics).
> 
> Perhaps I am missing something fundamental about the intention of your loop, which is not surprising given that it is in assembly and its intended behavior is not documented (other than a comment about data races). Your function appears to be exactly equivalent to this:
> 
> ```
> fn check_zeroed(data: [u8]) -> bool {
>     for byte in data.iter() {
>         if byte != 0 {
>             return false;
>         }
>     true
> }
> ```
> 
> Am I reverse engineering your assembly incorrectly? Is there something you're trying to accomplish other than checking for zeroes?

Yes, that is the intended and documented behavior. There's nothing else going on here.

> > > If we were to require true volatile behavior, then a better approach would be to define a platform-specific `unsafe fn read_byte_volatile(b: *const u8) -> u8` and then call it from a platform-independent loop.
> > 
> > 
> > That's sort of what I was trying to achieve here with this function (as well as the two functions above it), except that I implemented operations for memory regions instead of a single byte. If other people are okay with this, I'd also be fine with reimplementing them in terms of `read_byte_volatile` and `write_byte_volatile`.
> 
> Given that the intent of the loop is not entirely clear, it would be good to minimize the amount of assembly only to the specific operations that are platform-specific, so that all of the rest of the logic is in Rust. This also minimizes the possibility that the loop behavior differs from one platform to another.

:+1:

msft-jlange: > IIUC the compiler is allowed to just return `1` because it knows that the load was immediately preceded by a store to the same address with the value 1. This wouldn't be allowed for a volatile access, but for non-volatile atomics, AFAICT this is legal.

Not in this case.  Because atomics have `Sync` and `Send` behavior, the compiler cannot assume that another thread did not modify the data in between, and thus the load must be performed in case the value changed asynchronously.  The whole point of the atomic load is to capture the data as it exists *now*, relative to other threads, so the load can never be omitted based on data elision.

msft-jlange: And note also that `Ordering` applies only to memory fences, controlling the order in which the microarchitecture is allowed to process loads and stores.  It does not apply to any assumptions the compiler can make about multi-thread consistency.

Freax13: > > IIUC the compiler is allowed to just return `1` because it knows that the load was immediately preceded by a store to the same address with the value 1. This wouldn't be allowed for a volatile access, but for non-volatile atomics, AFAICT this is legal.
> 
> Not in this case. Because atomics have `Sync` and `Send` behavior, the compiler cannot assume that another thread did not modify the data in between, and thus the load must be performed in case the value changed asynchronously. The whole point of the atomic load is to capture the data as it exists _now_, relative to other threads, so the load can never be omitted based on data elision.

Can you point to any standard or other source that says "Atomics loads must capture the data as it exists _now_" or in other words "omitting loads after a store is not allowed"? The only thing I was able to find so far is an old article from 2009 written by Herb Sutter that says that this kind of optimization is allowed (see the last example [here](https://www.drdobbs.com/parallel/volatile-vs-volatile/212701484)) though the rules for atomics might have changed since then :-/

> And note also that `Ordering` applies only to memory fences, controlling the order in which the microarchitecture is allowed to process loads and stores. It does not apply to any assumptions the compiler can make about multi-thread consistency.

I don't believe that's true, I believe that the compiler is allowed to do optimizations on atomics op including re-ordering. The C++ docs for [`std::memory_order`](https://en.cppreference.com/w/cpp/atomic/memory_order) mention compiler reordering:
> In particular, this may occur if D is completed before C in thread 2, either due to compiler reordering or at runtime.

(this quote talks about two relaxed atomic operations on the same thread)

msft-jlange: > Can you point to any standard or other source that says "Atomics loads must capture the data as it exists _now_" or in other words "omitting loads after a store is not allowed"?

This appears to be very poorly specified.  The LLVM atomics guide for unordered operations (https://llvm.org/docs/Atomics.html#unordered) is vague - it prevents certain optimizations with respect to stores, and it permits certain load optimizations that do not involve dropping loads, but it doesn't address this particular case as being either allowed or unallowed.  I will note that the codegen portion of the atomics guide (https://llvm.org/docs/Atomics.html#atomics-and-codegen) states that on x86, "all atomic loads generate a `MOV`".  I interpret this to mean that *all* atomic loads result in a memory access, not just those that weren't optimized away, but it appears to be subject to interpretation.

> > And note also that `Ordering` applies only to memory fences, controlling the order in which the microarchitecture is allowed to process loads and stores. It does not apply to any assumptions the compiler can make about multi-thread consistency.
> 
> I don't believe that's true, I believe that the compiler is allowed to do optimizations on atomics op including re-ordering. The C++ docs for [`std::memory_order`](https://en.cppreference.com/w/cpp/atomic/memory_order) mention compiler reordering:

The LLVM guide appears to agree with your point of view, in that ordering requirements are permitted to affect the order of generated code and not just the memory fences.

If you really want to err on the side of safety, and assume the most unfavorable interpretation of the codegen guide, then it would be safe to preface your loop with a memory barrier (which probably doesn't exist as a Rust primitive and probably has to be implemented in platform-specific asm code).  Regardless of how the language can support this, either a function call to asm or an explicit memory barrier will put the compiler on notice that cross-thread synchronization is expected.

Freax13: > If you really want to err on the side of safety, and assume the most unfavorable interpretation of the codegen guide, then it would be safe to preface your loop with a memory barrier (which probably doesn't exist as a Rust primitive and probably has to be implemented in platform-specific asm code). Regardless of how the language can support this, either a function call to asm or an explicit memory barrier will put the compiler on notice that cross-thread synchronization is expected.

There's [`core::sync::atomic::fence`](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html) and [`core::sync::atomic::compiler_fence`](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html). In our situation, we'd want to use `compiler_fence` because we're only interested in preventing the compiler from reordering our accesses. Unfortunately, the precise semantics of `compiler_fence` are [unclear](https://github.com/rust-lang/unsafe-code-guidelines/issues/347).

Freax13: I [removed](https://github.com/coconut-svsm/svsm/pull/451/commits/f05092b28f37e4db1979ec3f2258a34cf7d2e5f0) `is_clear`.

msft-jlange (approved):

PR #452, opened 2024-09-05
kernel/requests: Ensure guest VMSA mapping is updated on switch from VMPL2

roy-hopkins:
Currently, when the guest running at VMPL2 switches back to the SVSM it is assumed that the guest VMSA virtual memory mapping is correct. However, it is possible for the VMPL2 AP to be destroyed and recreated while the AP request loop is blocked on waiting for the guest VMPL to call back to the SVSM. In this case the VMSA mapping is incorrect and the SVSM call fails.

This is fixed by updating the mappings on returning to the SVSM from the guest VMPL.

### DIFF
--- a/kernel/src/requests.rs
+++ b/kernel/src/requests.rs
@@ -174,14 +174,14 @@ pub fn request_loop() {
 
             switch_to_vmpl(GUEST_VMPL as u32);
         } else {
-            loop {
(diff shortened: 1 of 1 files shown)


joergroedel (approved):
LGTM. The logic is actually much nicer than before, thanks!

PR #453, opened 2024-09-06
doc/design: Add design doc for object

cxdong:
The syscalls design philosophy is to provide a unified interface for the user to access system resources, which are represented by object handles. This makes objects a fundamental concept in the COCONUT-SVSM kernel.

This document describes the object and object handle from both the user mode's and the COCONUT-SVSM kernel's point of view.

### DIFF
--- a/Documentation/docs/developer/design/OBJECT.md
+++ b/Documentation/docs/developer/design/OBJECT.md
@@ -1,0 +1,453 @@
+# Background
+
+The syscalls design philosophy is to provide a unified interface for the user to
+access system resources, which are represented by object handles. This makes
+objects a fundamental concept in the COCONUT-SVSM kernel
+
+This document describes the object and object handle from both the user mode's
+and the COCONUT-SVSM kernel's point of view.
+
+# Key Data Structures
+
+## Object
+
+An object represents the type of resource like file, VM, vCPU in the
+COCONUT-SVSM kernel, that can be accessible by the user mode. A trait named Obj
+is defined for such type of resource, which defines the common functionalities
+of the object. The Obj trait is defined with the trait bounds of Send and Sync,
+which means the object implementing Obj trait could be sent to another thread
+and shared between threads safely.
+
+```Rust
+trait Obj: Send + Sync {
+    /// Convert to a virtual machine object if is supported.
+    fn as_vm(&self) -> Option<&VmObj> {
+        None
+    }
+
+    /// Convert to a virtual cpu object if is supported.
+    fn as_vcpu(&self) -> Option<&VcpuObj> {
+        None
+    }
+
+    /// Get a mappable file handle if the object is mappable.
+    fn mappable(&self) -> Option<&FileHandle> {
+        None
+    }
+
+    /// Convert to an object which implements EventObj trait.
+    fn as_event(&self) -> Option<&dyn EventObj> {
+        None
+    }
+    ...
+}
+```
+
+In order for the user mode to access the resources in the COCONUT-SVSM kernel,
+the kernel needs to implement the `Obj` trait to represent the resources as
+objects. Some objects may need to implement multiple methods in the Obj trait
+for multiple purposes. For example, a vcpu object needs to implement `as_vcpu()`
+to represent it as a vcpu object, `mappable()` to provide a file handle for its
+user-mode-mappable backing 4k page, and `as_event()` to represent it as an event
+object which can be used by the WAIT_FOR_EVENT syscall.
+
+```Rust
+struct VcpuObj {
+    id: VmId,
+    ...
+}
+
+/// The trait for the object which can be used as an event.
+trait EventObj {
+    ...
+}
+
+impl EventObj for VcpuObj {
+    ...
+}
+
+impl Obj for VcpuObj {
+    fn as_vcpu(&self) -> Option<&VcpuObj> {
+        Some(self)
+    }
+
+    fn mappable(&self) -> Option<&FileHandle> {
+        Some(&self.run_page_file_handle)
+    }
+
+    fn as_event(&self) -> Option<&dyn EventObj> {
+        Some(self)
+    }
+    ...
+}
+
+```
+
+Objects without special requirements can fall back to the default implementation
+in the Obj trait, which returns None.
+
+## Object Handle
+
+When the user mode is trying to open a particular kernel resource via the
+syscalls, the COCONUT-SVSM kernel creates a corresponding object which
+implements Obj trait, and allocates an object handle with a unique id for that
+object. The object handle is defined as below:
+
+```Rust
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub struct ObjHandle(u32);
+
+impl ObjHandle {
+    /// Create a new object handle with the allocated id
+    pub fn new(id: u32) -> Self {
+        Self(id)
+    }
+}
+
+impl From<u32> for ObjHandle {
+    #[inline]
+    fn from(id: u32) -> Self {
+        Self(id)
+    }
+}
+
+impl From<ObjHandle> for u32 {
+    #[inline]
+    fn from(obj_handle: ObjHandle) -> Self {
+        obj_handle.0
+    }
+}
+```
+
+An `ObjHandle` can be converted to a `u32` id which is returned to the user
+mode, and the subsequent syscalls use this id to access this object. The passed
+id from the syscalls can be converted to an `ObjHandle`, which is used to access
+the object in the COCONUT-SVSM kernel.
+
+### User Mode Object Handle
+
+The object is exposed to the user mode via the object-opening related syscalls,
+which returns the id of the object created by the COCONUT-SVSM kernel. The user
+mode can make use this id to access the corresponding object via other syscalls.
+From the user mode's point of view, the object handle is defined as below:
+
+```Rust
+/// User mode object handle received from syscalls.
+pub struct ObjHandle(u32);
+
+impl ObjHandle {
+    /// Create a new object handle with the id returned from a syscall.
+    pub(crate) fn new(id: u32) -> Self {
+        Self(id)
+    }
+}
+
+impl From<&ObjHandle> for u32 {
+    #[inline]
+    fn from(obj_handle: &ObjHandle) -> Self {
+        obj_handle.0
+    }
+}
+
+impl Drop for ObjHandle {
+    fn drop(&mut self) {
+        // Close the object when drop the ObjHandle.
+        unsafe { syscall1(SYS_CLOSE, self.0.into()) };
+    }
+}
+
+```
+
+The user mode `ObjHandle` doesn't implement Copy/Clone trait, and dropping
+`ObjHandle` will automatically close the underlying object via the syscall.
+
+For a syscall class which is associated with a particular object handle type,
+e.g. VM object handles for VMM subsystem syscalls, VCPU object handles for VCPU
+subsystem syscalls, can be defined as:
+
+```Rust
+pub struct VmObjHandle(ObjHandle);
+```
+
+```Rust
+pub struct VcpuObjHandle(ObjHandle);
+```
+
+# Object Management in COCONUT-SVSM Kernel
+
+To facilitate the user mode using the object, the COCONUT-SVSM kernel should:
+
+- Manage the object's lifecycle properly. The underlying object should be
+  dropped when the user mode closes the object handle via syscalls, or the user
+  mode is terminated without closing.
+
+- Prevent one user mode process from misusing the object handle opened by
+  another. But the object handles are shared among the threads within the same
+  process.
+
+To achieve the above goals, the opened object should be associated with the
+process which creates it. The task structure is extended to hold the created
+objects.
+
+```Rust
+pub struct Task {
+    ...
+
+    /// Objects shared among threads within the same process
+    objs: Arc<RWLock<BTreeMap<ObjHandle, Arc<dyn Obj>>>>,
+}
+```
+
+The objs is a BTreeMap with the object handle id as the key and the Arc<dyn Obj>
+as the value. It is wrapped by an Arc and protected by a RWLock, to make it
+shared among the threads within the same process.
+
+The task structure will provide 3 new functions:
+
+- `add_obj(&self, obj: Arc<dyn Obj>) -> Result<ObjHandle, SvsmError>;` It
+  allocates a unique ObjHandle which is local to the process for the object. The
+  object is added to the BTreeMap with the `ObjHandle` as the key. This method
+  will be used by the syscalls which open an object.
+
+- `remove_obj(&self, id: ObjHandle) -> Result<Arc<dyn Obj>, SvsmError>;` It
+  removes the object from the BTreeMap. This method will be used by the CLOSE
+  syscall to remove the corresponding object from process and drop it.
+
+- `get_obj(&self, id: ObjHandle) -> Result<Arc<dyn Obj>, SvsmError>;` It gets
+  the object from the BTreeMap, which increases the reference counter. This
+  method will be used by the syscalls which access an object.
+
+When a task is terminated while it still has opened objects, these objects will
+be dropped automatically when the `objs` is dropped, if `objs` held the last
+reference to the objects.
+
+# Opening an Object in User Mode
+
+The user mode can open a particular object via syscalls. For example, VM_OPEN
+syscall is used to open a virtual machine object. The COCONUT-SVSM kernel
+provides `obj_add()` function to facilitate opening an object in user mode.
+
+```Rust
+pub fn sys_vm_open(idx: u32) -> Result<u64, i32> {
+    // Get the VmObj
+    let vm_obj = get_vm_obj(idx)?;
+
+    // Add the VmObj to the current process and return the
+    // object handle id to the user mode.
+    obj_add(vm_obj).map_or(Err(EINVAL), |id| Ok(u32::from(id).into()))
+}
+
+```
+
+```Rust
+/// Add an object to the current process and assigns it an `ObjHandle`.
+///
+/// # Arguments
+///
+/// * `obj` - An Arc<dyn Obj> representing the object to be added.
+///
+/// # Returns
+///
+/// * `Result<ObjHandle, SvsmError>` - Returns the object handle of the
+///   added object if successful, or an `SvsmError` on failure.
+///
+/// # Errors
+///
+/// This function will return an error if adding the object to the
+/// current task fails.
+pub fn obj_add(obj: Arc<dyn Obj>) -> Result<ObjHandle, SvsmError> {
+    current_task().add_obj(obj)
+}
+```
+
+```Rust
+impl Task {
+    ...
+
+    /// Adds an object to the current task.
+    ///
+    /// # Arguments
+    ///
+    /// * `obj` - The object to be added.
+    ///
+    /// # Returns
+    ///
+    /// * `Result<ObjHandle, SvsmError>` - Returns the object handle for the object
+    ///   to be added if successful, or an `SvsmError` on failure.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if allocating the object handle fails.
+    pub fn add_obj(&self, obj: Arc<dyn Obj>) -> Result<ObjHandle, SvsmError> {
+        let mut objs = self.objs.lock_write();
+        let last_key = objs
+            .keys()
+            .last()
+            .map_or(Some(0), |k| u32::from(*k).checked_add(1))
+            .ok_or(SvsmError::from(ObjError::InvalidHandle))?;
+        let id = ObjHandle::new(if last_key != objs.len() as u32 {
+            objs.keys()
+                .enumerate()
+                .find(|(i, &key)| *i as u32 != u32::from(key))
+                .unwrap()
+                .0 as u32
+        } else {
+            last_key
+        });
+
+        objs.insert(id, obj);
+
+        Ok(id)
+    }
+}
+```
+
+The `obj_add()` takes the `Arc<dyn Obj>` as an input, which represents the
+particular object to be added, and stores the object in the current task via the
+`add_obj()` method. An `ObjHandle` is allocated by the `add_obj()` which is
+local to the process, and returned to the syscall. It is converted to a `u32`
+and returned to the user mode as the user mode `ObjHandle`.
+
+# Closing an Object in User Mode
+
+The CLOSE syscall can close an object, taking the object handle id as the input
+parameter. The COCONUT-SVSM kernel provides `obj_close()` function to facilitate
+closing an object in the syscall.
+
+```Rust
+pub fn sys_close(obj_id: u32) -> Result<u64, i32> {
+    // Close the object by the object handle id.
+    let _ = obj_close(obj_id.into());
+    Ok(0)
+}
+```
+
+```Rust
+/// Closes an object identified by its unique identifier.
+///
+/// # Arguments
+///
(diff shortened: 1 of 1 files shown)


Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,506 @@
...
+
+```Rust
+trait Obj: Send + Sync {
+    /// Open an object
+    fn open(&self) -> Result<(), SvsmError> {
+        Ok(())
+    }
+
+    /// Close an object
+    fn close(&self) {}
```
Freax13, line 29: Why can't the `Obj` just implement `Drop` instead of implementing this method?

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,506 @@
...
+COCONUT-SVSM kernel, that can be accessible by the user mode. A trait named Obj
+is defined for such type of resource, which defines the common functionalities
+of the object. The Obj trait is defined with the trait bounds of Send and Sync,
+which means the object implementing Obj trait could be sent to another thread
+and shared between threads safely.
+
+```Rust
+trait Obj: Send + Sync {
+    /// Open an object
+    fn open(&self) -> Result<(), SvsmError> {
```
Freax13, line 24: What would this method do? Why can't we do this at the time the object is constructed?

cxdong:
Hi @joergroedel, this is the object framework management documentation, which is in line with our POC work for the init process and user mode VMM (so-called device model). We are eager to gather the early feedback from the community through this draft pull request. Your reviews and comments are highly appreciated. Thank you!

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,506 @@
...
+///
+/// # Errors
+///
+/// This function will return an error if adding the object handle to the
+/// current task fails.
+pub fn obj_open(obj: Arc<dyn Obj>) -> Result<u32, SvsmError> {
+    let id = OBJ_ID_ALLOCATOR.next_id();
+    current_task()
+        .add_obj_handle(Arc::new(ObjHandle::new(id, obj)?))
+        .map(|_| id)
```
Freax13, line 294:
```suggestion
    current_task()
        .add_obj_handle(Arc::new(ObjHandle::new(id, obj)?))?;
    Ok(id)
```

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,506 @@
...
+
+The CLOSE syscall can close an object, taking the object id as the input
+parameter. The COCONUT-SVSM kernel provides `obj_close()` function to facilitate
+closing an object in the syscall.
+
+```Rust
+pub fn sys_close(obj_id: u32) -> Result<u64, i32> {
+    // Close the object by the object id.
+    let _ = obj_close(obj_id);
+    Ok(0)
```
Freax13, line 342: Why are we not reporting errors to userspace here?

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,506 @@
...
+To achieve the above goals, the `ObjHandle` should be associated with the task
+which creates it. The task structure is extended to hold the ObjHandles created
+by this thread.
+
+```Rust
+pub struct Task {
+    ...
+
+    /// Object handles created by this task
+    obj_handles: RWLock<BTreeMap<u32, Arc<ObjHandle>>>,
```
Freax13, line 185: Storing the object id in the map keys and in the map values (inside `ObjHandle`) is redundant. Can we remove it from `ObjHandle` (or just remove `ObjHandle` altogether and just pass around `Arc<Obj>`?

cxdong: Yes, I think we can remove the id from `ObjHandle`. One of the purposes of adding obj_handles in tasks is to make sure all its opened objects would be closed if the task is terminated without closing objects first. So implemented Drop method for ObjHandle, which can close all the objects when dropping the task structure. If just pass around Arc<obj>, then seems needs to explicitly close all objects. From this point of view, prefer to remove the id from the `ObjHandle`, which is changed to `structure ObjHandle(Arc<obj>)`

Documentation/docs/developer/design/OBJECT.md
cxdong: Thanks! This is better.

Documentation/docs/developer/design/OBJECT.md
cxdong: This one is based on our understanding about the SVSM syscall ABI/API spec for the CLOSE syscall, which said "Returns Always returns 0, even if called with an invalid handle.". Probably we misunderstood this. Maybe errors should return to the user mode, unless it is caused by an invalid handle?

Freax13: Ah, nvm, you're completely right, the code matches the spec.

Documentation/docs/developer/design/OBJECT.md
Freax13: > We expect some object is constructed statically, means once it is created, it will never be dropped, so it only needs to be constructed once, but would be opened/closed multiple times by the user mode.

I'm not sure that we should pursue supporting this pattern. IMHO we should tie the lifetime of an `Obj` implementation to the accessibility of the `Obj` in userspace (i.e. the object is created, added to the user's map of object handles and destroyed once the user closes the object). If we want to keep some data around for longer than that, we can store it elsewhere and store a reference in the `Obj`. Explicit `open()` and `close()` methods are generally a red flag in Rust because they imply partial initialization which should generally be avoided. Not having those methods also frees the implementation from having to worry about what will happen if `open()` or `close()` are called more than once (`open(); open()`) or in the wrong order (`close(); open();`). 

> The open method can be used to do special handling when the object is opened. For example, the VM object cannot be re-opened, so it can implement this method to prevent re-opening.

This could probably be implemented outside an `open()` method though? My guess is that whatever code is constructing or recycling the object could also implement access checks, right?

Documentation/docs/developer/design/OBJECT.md
Freax13: > One of the purposes of adding obj_handles in tasks is to make sure all its opened objects would be closed if the task is terminated without closing objects first. So implemented Drop method for ObjHandle, which can close all the objects when dropping the task structure. If just pass around Arc, then seems needs to explicitly close all objects. From this point of view, prefer to remove the id from the `ObjHandle`, which is changed to `structure ObjHandle(Arc<obj>)`

Sure, sounds reasonable. I guess we'll have to see what the outcome of the discussion started in https://github.com/coconut-svsm/svsm/pull/453#discussion_r1746782468 is and whether we want to keep the `close()` method to see if we still need `ObjHandle`.

Documentation/docs/developer/design/OBJECT.md
cxdong: > I'm not sure that we should pursue supporting this pattern. IMHO we should tie the lifetime of an `Obj` implementation to the accessibility of the `Obj` in userspace (i.e. the object is created, added to the user's map of object handles and destroyed once the user closes the object). If we want to keep some data around for longer than that, we can store it elsewhere and store a reference in the `Obj`. Explicit `open()` and `close()` methods are generally a red flag in Rust because they imply partial initialization which should generally be avoided. Not having those methods also frees the implementation from having to worry about what will happen if `open()` or `close()` are called more than once (`open(); open()`) or in the wrong order (`close(); open();`).

I see. Thanks for the explanations! In our POC, the vm object is stored in a global BTreeMap and vcpu objects are stored in per-cpu shared area, and both of them are statically constructed and always keep alive in the memory. I will take a deep look to see how to tie the lifecycle of these objects to the accessibility of the `obj` in userspace like you mentioned.

> This could probably be implemented outside an `open()` method though? My guess is that whatever code is constructing or recycling the object could also implement access checks, right?
Yes, I guess so. I will take a deep look. Thanks!

Documentation/docs/developer/design/OBJECT.md
cxdong: > Sure, sounds reasonable. I guess we'll have to see what the outcome of the discussion started in [#453 (comment)](https://github.com/coconut-svsm/svsm/pull/453#discussion_r1746782468) is and whether we want to keep the `close()` method to see if we still need `ObjHandle`.

Yes, it turns out that we can remove `open()` and `close` method, so seems `ObjHandle` is not needed and can just pass around Arc<Obj> to task.

Additionally, I would appreciate any feedback you might have on the section titled "## Moving an ObjHandle from One Thread to Another"? If the option#2 in this section is valid, then seems kernel needs to maintain a global map for `ObjHandle id` -> `task id` (e.g., BTreeMap<obj_id, task_id>), which can let kernel find out the associated task with the `ObjHandle id` and then do the moving. So we expect a mapping is added when kernel creates a new `ObjHandle`, and a mapping is removed when kernel drops a `ObjHandle`. In this case, it looks like keeping an id in `ObjHandle` can make the mapping removal possible when drops the `ObjHandle` as the `drop()` can find out the mapping via the id field in the `ObjHande`. But for option#1, I think it is not necessary to have such global mapping, thus also no need to have the id in `ObjHandle`.

Freax13: There was some discussion of this in https://mail.8bytes.org/pipermail/svsm-devel/2024-June/000346.html. My understanding is that object handles are tied to a process (i.e. they are not global. The object might be used by multiple processes, but each handle to it is always local to a process). Currently, there is no way to spawn a second process, but when we eventually add an `EXEC` syscall for this, this syscall would also take care of transferring object handles to the child and removing them from the parent.

cxdong: Yes, that is the discussion made me realized that, an object handle may need to be transferred from one thread to another. So in our POC, we have implemented `THREAD_CREATE` syscall which is used to create a vcpu thread with moving the vcpu object handle from the parent thread to this vcpu thread, described in the USER_MODE_VMM.dm of [PR 355](https://github.com/coconut-svsm/svsm/pull/355).

As the `THREAD_CREATE` syscall takes a structure X86Regs as the input which is quite common, I suppose this syscall cannot tell whether an object handle should be transferred to the child thread or even which register points to an object handle, so the object handle transferring may not be not done by the `THREAD_CREATE` syscall (I guess `EXEC` syscall is similar). Because of this, we proposed two options in the end of this document. And in our POC, the option#2 is implemented, which can support transferring an `ObjHandle` from one thread to another when a syscall is trying to get an `ObjHandle` by the id. The implementation is by searching the id from a global BTreeMap<obj_id, task_id> to get the owner task and then move the object handle from the owner task to the current task. Because we expect a <obj_id, task_id> pair can be removed when calls `drop()` for `ObjHandle`, the `ObjHandle` should contain the id so that its `drop()` can find out and remove the corresponding task_id from the BTreeMap.

Then to move forward, the suggestion is to pass around the `Arc<Obj>` to tasks, and because of this, seems the `struct ObjHandle(Arc<Obj>)` is not needed so we can just remove it, is this a right understanding? If so, I will update the document accordingly.

Freax13: > Yes, that is the discussion made me realized that, an object handle may need to be transferred from one thread to another. 

If I understand the spec and the e-mails correctly, object handles don't need to be moved from one *thread*/*task* to another iff they belong to the same *process*. The table of object handles shouldn't be uniquely owned by a *thread*/*task* but by a process (very much how in Linux file descriptors can commonly be accessed from any thread within a process). This is also why `THREAD_CREATE` doesn't have the option to share object handles: They are always shared among threads of the same process. Presumably, `EXEC` will be different in that `EXEC` spawns a new process and that new process has a new object handle table which is why there's an opportunity to remove object handles from a parent process and move them into the child.

Under those assumptions, I don't think there's a need for a global table of object handles.

cxdong: Thank you for pointing this out! I had initially thought that object handles should be uniquely by a thread/task, rather than by the entire process. I will revise the documentation to reflect this new understanding. Your input is greatly valued!

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,506 @@
...
+object which can be used by the WAIT_FOR_EVENT syscall.
+
+```Rust
+struct VcpuObj {
+    id: VmId,
+    ...
+}
+
+/// The trait for the object which can be used as an event.
+triat EventObj {
```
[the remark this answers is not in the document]
cxdong: Thanks for correcting this! Will fix.

Documentation/docs/developer/design/OBJECT.md
cxdong: The `close()` method has been removed in the update, mark this conversation as resolved.

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,408 @@
...
+To achieve the above goals, the opened object should be associated with the
+process which creates it. The task structure is extended to hold the created
+objects.
+
+```Rust
+pub struct Task {
+    ...
+
+    /// Objects created by this task
+    objs: Arc<RWLock<BTreeMap<u32, Arc<dyn Obj>>>>,
```
Freax13, line 165: This comment is no longer accurate.

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,408 @@
...
+    ///
+    /// This function will return an error if the object handle id already exists in the current task.
+    pub fn add_obj(&self, id: u32, obj: Arc<dyn Obj>) -> Result<(), SvsmError> {
+        if let Entry::Vacant(entry) = self.objs.lock_write().entry(id) {
+            entry.insert(obj);
+            Ok(())
+        } else {
+            Err(ObjError::Exist.into())
+        }
+    }
```
Freax13, line 257: Can we move the allocation of the object id into this method? Now that we no longer need to maintain a global id, we can just use any available id in the map. Allocating the id in `add_obj` has the advantage that `add_obj` can no longer fail because the id is already in use.

Documentation/docs/developer/design/OBJECT.md
cxdong: The `struct ObjHandle` for the kernel has been removed, and the BTreeMap directly stores the value of `Arc<dyn Obj>`. The doc is also updated to indicate the objects are shared among the threads within the same process.

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,408 @@
...
+as the value. It is wrapped by an Arc and protected by a RWLock, to make it
+shared among the threads within the same process.
+
+The task structure will provide 3 new functions:
+
+- `add_obj(&self, id: u32, obj: Arc<dyn Obj>) -> Result<(), SvsmError>;` This is
+  to add the object to the BTreeMap with the object handle id as the key. The
+  syscalls which open an object will add Arc<dyn Obj> to the current task.
+
+- `remove_obj(&self, id: u32) -> Result<Arc<dyn Obj>, SvsmError>;` This is to
```
Freax13, line 179: Can we also add a newtype `pub struct ObjectHandle(u32);` (without a `Drop` impl) in the kernel and use it here? Personally, I'd prefer this to passing around untyped integers and I expect us to use this type a lot when implementing syscalls.

Documentation/docs/developer/design/OBJECT.md
cxdong: Is this more accurate? `/// Objects shared among threads within the same process`

Documentation/docs/developer/design/OBJECT.md
cxdong: Yes, I think so. Then, do we expect to use the `ObjHandle(u32)` as the key for the `objs: Arc<RWLock<BTreeMap<ObjHandle(u32), Arc<dyn Obj>>>>`, or just u32 as key is also ok?

Also, to align with `remove_obj()`, the `get_obj()` will also change to use `ObjHandle(u32)` as the input.

Documentation/docs/developer/design/OBJECT.md
cxdong: Sure, the id will be an index in the `objs`, which is local to the process. And as the id is allocated inside the `add_obj()`, looks like `add_obj()` needs to return `ObjHandle(u32)` instead of `Result<(), SvsmError>`. I will update accordingly.

Documentation/docs/developer/design/OBJECT.md
Freax13: Yes :+1:

Documentation/docs/developer/design/OBJECT.md
Freax13: > Yes, I think so. Then, do we expect to use the `ObjHandle(u32)` as the key for the `objs: Arc<RWLock<BTreeMap<ObjHandle(u32), Arc<dyn Obj>>>>`, or just u32 as key is also ok?

Is there a reason to use `u32` as the map key? If not, let's just use the `ObjHandle`, that way `get` and `close` won't need to convert between the two types.

> Also, to align with `remove_obj()`, the `get_obj()` will also change to use `ObjHandle(u32)` as the input.

:+1:

cxdong: > > Yes, I think so. Then, do we expect to use the `ObjHandle(u32)` as the key for the `objs: Arc<RWLock<BTreeMap<ObjHandle(u32), Arc<dyn Obj>>>>`, or just u32 as key is also ok?
> 
> Is there a reason to use `u32` as the map key? If not, let's just use the `ObjHandle`, that way `get` and `close` won't need to convert between the two types.

No particular reason to use `u32`, I also prefer to use `ObjHandle(u32)` as the key. Will update. Thanks!

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,427 @@
...
+- `add_obj(&self, obj: Arc<dyn Obj>) -> ObjHandle;` It allocates a unique
+  ObjHandle which is local to the process for the object. The object is added to
+  the BTreeMap with the `ObjHandle` as the key. This method will be used by the
+  syscalls which open an object.
+
+- `remove_obj(&self, id: &ObjHandle) -> Result<Arc<dyn Obj>, SvsmError>;` It
+  removes the object from the BTreeMap. This method will be used by the CLOSE
+  syscall to remove the corresponding object from process and drop it.
+
+- `get_obj(&self, id: &ObjHandle) -> Result<Arc<dyn Obj>, SvsmError>;` It gets
```
Freax13, line 216:
```suggestion
- `remove_obj(&self, id: ObjHandle) -> Result<Arc<dyn Obj>, SvsmError>;` It
  removes the object from the BTreeMap. This method will be used by the CLOSE
  syscall to remove the corresponding object from process and drop it.

- `get_obj(&self, id: ObjHandle) -> Result<Arc<dyn Obj>, SvsmError>;` It gets
```

`ObjHandle` is `Copy`, so we don't need to pass a reference.

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,427 @@
...
+    /// # Arguments
+    ///
+    /// * `obj` - The object to be added.
+    ///
+    /// # Returns
+    ///
+    /// * `ObjHandle` - Returns the object handle for the object to be added.
+    pub fn add_obj(&self, obj: Arc<dyn Obj>) -> ObjHandle {
+        let mut objs = self.objs.lock_write();
+        let id = ObjHandle::new(objs.len() as u32);
```
Freax13, line 271: There may be an object with id `objs.len()`. Consider the following sequence:
1. An object is added -> `len()` is 0, so id is 0
2. Another object is added -> `len()` is 1, id is 1
3. The object with id 0 is removed.
4. Another object is added -> `len()` is 1, so id is 1

In step 4 a second object with id 1 was created.

Documentation/docs/developer/design/OBJECT.md
```diff
@@ -0,0 +1,427 @@
...
+///
+/// # Arguments
+///
+/// * `obj` - An Arc<dyn Obj> representing the object to be opened.
+///
+/// # Returns
+///
+/// * `ObjHandle` - Returns the object handle of the opened object.
+pub fn obj_open(obj: Arc<dyn Obj>) -> ObjHandle {
+    current_task().add_obj(obj)
```
Freax13, line 252: Let's rename this to `obj_add`. We no longer call any `.open()` methods on the object, so the name doesn't quite fit anymore.

Documentation/docs/developer/design/OBJECT.md
cxdong: May I inquire about any potential downsides to use a reference as the input parameter? My previous consideration was that the implementation inside the `remove_obj` and `get_obj` is finding the object in a BTreeMap which takes a reference as an input, so made the change to use reference to indicate this to the users. But seems this is not necessary :)

Documentation/docs/developer/design/OBJECT.md
cxdong: Sure, makes sense.

Documentation/docs/developer/design/OBJECT.md
cxdong: Thanks for caching this!
How about use `let id = ObjHandle::new(objs.keys().last().map_or(0, |k| u32::from(*k) + 1));`, which always get the largest value among the existing `ObjHandle` in the BTreeMap?

Documentation/docs/developer/design/OBJECT.md
Freax13: Small types that implement `Copy` are usually passed by value. Passing such values by value is usually easier/more convenient for the caller and easier for the compiler to understand (if the compiler doesn't inline get_obj, it's forced to write the value to the stack. If the value is always passed by value, the compiler can put the value into a register and never has to touch the stack regardless of whether get_obj is inlined).

Documentation/docs/developer/design/OBJECT.md
Freax13: > How about use `let id = ObjHandle::new(objs.keys().last().map_or(0, |k| u32::from(*k) + 1));`, which always get the largest value among the existing `ObjHandle` in the BTreeMap?

That doesn't get us the smallest possible object id, but that's fine with me. LGTM :+1:

Documentation/docs/developer/design/OBJECT.md
cxdong: I see. Thanks for the detailed explanations! I will use `ObjHandle` for `remove_obj` and `get_obj`.

Documentation/docs/developer/design/OBJECT.md
Freax13: > Below code will try to get the largest key first, which is most likely to be the case.

That's a clever optimization!

> But if there is any gap in the existing keys, try to do a linear search to get the smallest possible key.

The runtime of this is actually `O(n * log(n))` because each `.get(...)` call is `O(log(n))`.

Here's an implementation I came up with for another project: https://github.com/Freax13/mushroom/blob/4fad16fbcdb6abbd0444d915f3ed114d65eda77b/tee/kernel/src/fs/fd.rs#L146-L154 (you can ignore `min` and `no_file_limit`, we don't need these here). That implementation should be linear if `BTreeMap` is linear (which I'm not 100% certain it is).

Freax13: Ooh, very nice, I like it! I think you can use `.unwrap()` instead of `.map_or(last_key, |(id, _)| id as u32)`.

cxdong: Sure, per this implementation, the find should always return `Some(...)` instead of `None` so we are safe to use `.unwrap()`. I will update the doc accordingly. Thanks!

PR #455, opened 2024-09-12
implement shadow stacks

Freax13:
This PR implements shadow stacks.

Shadow stacks are enabled/disabled at compile time. AFAIK all processors supporting either AMD SEV-SNP or Intel TDX support shadow stacks, so no runtime checks are implemented. This also avoids the pitfall where the hypervisor lies about shadow stack availability to maliciously lower the security of the SVSM. 

KVM intercepts accesses to the MSRs used for shadow stacks, so some host-side [modifications](https://github.com/user-attachments/files/16982253/kvm-shstk.zip) are required to make this work.

I implemented a #CP exception handler to display diagnostic information when a shadow stack-related issue occurs. We might want to disable this exception handler for release builds and cause a triple fault instead. #CP exceptions are likely a sign of something fishy going on and we might want to terminate the guest just to be sure.

Currently, shadow stacks are disabled by default for a couple of reasons:
1. The KVM patches mentioned above are needed to make this work.
2. ~We haven't implemented proper user syscalls (using `syscall` & `sysret`) yet and I don't want to hinder any efforts in that direction by forcing a proper syscall implementation to support shadow stacks out of the box. We should probably get syscalls without shadow stacks working first and once that's merged, we can add shadow stack support later and eventually enable shadow stacks by default.~
3. I haven't tested shadow-stacks together with `enable-gdb` yet, but I suspect this probably needs some additional work.

### DIFF
--- a/kernel/src/cpu/idt/entry.S
+++ b/kernel/src/cpu/idt/entry.S
@@ -29,6 +29,12 @@ HV_DOORBELL_ADDR:
         pushq   %r13
         pushq   %r14
         pushq   %r15
+	pushq   $0
+	# rdsspq is a nop when shadow stacks are not supported. Make sure that
+	# rax is 0 if that's the case.
+	xorl	%eax, %eax
+	rdsspq  %rax
+	pushq	%rax
 .endm
 
 .macro default_entry_no_ist name: req handler:req error_code:req vector:req
@@ -141,7 +147,7 @@ hv_not_vmpl_switch:
 	// required.  This could not be performed before the RIP check because
 	// the previous RIP determines where to find the previous EFLAGS.IF
 	// value on the stack.
-	testl	${IF}, 18*8(%rcx)
+	testl	${IF}, 20*8(%rcx)
 	jz	postpone_hv
 	// Switch to the stack pointer from the previous exception, which
 	// points to the register save area, and continue with #HV
@@ -206,6 +212,12 @@ continue_hv:
         pushq   %r13
         pushq   %r14
         pushq   %r15
+	pushq   $0
+        # rdsspq is a nop when shadow stacks are not supported. Make sure that
+	# rdx is 0 if that's the case.
+	xorl	%edx, %edx
+	rdsspq  %rdx
+        pushq	%rdx
 handle_as_hv:
 	// Load the address of the #HV doorbell page.  The global address
 	// might not yet be configured, and the per-CPU page might also not
@@ -225,13 +237,13 @@ handle_as_hv_with_doorbell:
 default_return:
 	// Ensure that interrupts are disabled before attempting any return.
 	cli
-	testb	$3, 17*8(%rsp) // Check CS in exception frame
+	testb	$3, 19*8(%rsp) // Check CS in exception frame
 	jnz 	return_user
 return_all_paths:
 	// If interrupts were previously available, then check whether any #HV
 	// events are pending.  If so, proceed as if the original trap was
 	// #HV.
-	testl 	${IF}, 18*8(%rsp) // check EFLAGS.IF in exception frame
+	testl 	${IF}, 20*8(%rsp) // check EFLAGS.IF in exception frame
 	jz 	begin_iret_return
 	movq 	HV_DOORBELL_ADDR(%rip), %rdi
 	test 	%rdi, %rdi
@@ -251,23 +263,23 @@ iret_return_window:
 begin_iret_return:
 	// Reload registers without modifying the stack pointer so that if #HV
 	// occurs within this window, the saved registers are still intact.
-	movq 	0*8(%rsp), %r15
-	movq 	1*8(%rsp), %r14
-	movq 	2*8(%rsp), %r13
-	movq 	3*8(%rsp), %r12
-	movq 	4*8(%rsp), %r11
-	movq 	5*8(%rsp), %r10
-	movq 	6*8(%rsp), %r9
-	movq 	7*8(%rsp), %r8
-	movq 	8*8(%rsp), %rbp
-	movq 	9*8(%rsp), %rdi
-	movq 	10*8(%rsp), %rsi
-	movq 	11*8(%rsp), %rdx
-	movq 	12*8(%rsp), %rcx
-	movq 	13*8(%rsp), %rbx
-	movq 	14*8(%rsp), %rax
+	movq 	2*8(%rsp), %r15
+	movq 	3*8(%rsp), %r14
+	movq 	4*8(%rsp), %r13
+	movq 	5*8(%rsp), %r12
+	movq 	6*8(%rsp), %r11
+	movq 	7*8(%rsp), %r10
+	movq 	8*8(%rsp), %r9
+	movq 	9*8(%rsp), %r8
+	movq 	10*8(%rsp), %rbp
+	movq 	11*8(%rsp), %rdi
+	movq 	12*8(%rsp), %rsi
+	movq 	13*8(%rsp), %rdx
+	movq 	14*8(%rsp), %rcx
+	movq 	15*8(%rsp), %rbx
+	movq 	16*8(%rsp), %rax
 
-	addq 	$16*8, %rsp
+	addq 	$18*8, %rsp
 
 default_iret:
 	iretq
--- a/kernel/src/cpu/idt/entry.S
+++ b/kernel/src/cpu/idt/entry.S
@@ -370,7 +370,11 @@ default_entry_no_ist	name=xf		handler=panic			error_code=0	vector=19
 // Vector 20 not defined
 
 // #CP Control-Protection Exception (Vector 21)
+.if CFG_SHADOW_STACKS
+default_entry_no_ist	name=cp		handler=control_protection			error_code=1	vector=21
+.else
 default_entry_no_ist	name=cp		handler=panic			error_code=1	vector=21
+.endif
 
 // Vectors 22-27 not defined
 
--- a/kernel/src/cpu/idt/entry.S
+++ b/kernel/src/cpu/idt/entry.S
@@ -113,6 +113,11 @@ asm_entry_hv:
 	// RIP is in the return window, so update RIP to the cancel point.
 	leaq	switch_vmpl_cancel(%rip), %rbx
 	movq	%rbx, 0x20(%rsp)
+	.if CFG_SHADOW_STACKS
+	// Update RIP on the shadow stack to the cancel point.
+	rdsspq  %rax
+        wrssq	%rbx, 8(%rax)
+	.endif
 	// Defer any further processing until interrupts can be processed.
 	jmp	postpone_hv
 hv_not_vmpl_switch:
@@ -153,6 +158,10 @@ hv_not_vmpl_switch:
 	// points to the register save area, and continue with #HV
 	// processing.
 	movq	%rcx, %rsp
+	// Pop the current stack frame, so that the previous stack frame sits
+	// on top of the shadow stack.
+	movl	$3, %eax
+	incsspq	%rax
 	jmp	handle_as_hv
 
 postpone_hv:
@@ -195,6 +204,12 @@ restart_hv:
 	movq	2*8(%rsp), %rax
 	movq	%rax, -2*8(%rcx)
 	leaq	-4*8(%rcx), %rsp
+        .if CFG_SHADOW_STACKS
+	// Pop the current stack frame, so that the previous stack frame sits
+	// on top of the shadow stack.
+	movl	$3, %eax
+        incsspq	%rax
+	.endif
 	
 continue_hv:
 	// At this point, only the dummy error code and first three registers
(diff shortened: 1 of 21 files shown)


kernel/src/task/schedule.rs
```diff
@@ -431,5 +468,63 @@ global_asm!(
...
+/// the following:
+///
+/// 1. Switch to the shadow stack at the fixed address using `rstorssp`.
+/// 2. Transfer the shadow stack restore token from the shadow stack at the
+///    fixed address to the previous shadow stack by executing `saveprevssp`.
+/// 3. Switch the page tables. This doesn't lead to problems with the shadow
+///    stack because is mapped into both page tables.
+/// 4. Switch to the new shadow stack using `rstorssp`.
+/// 5. Transfer the shadow stack restore token from the new shadow stack back
+///    to the shadow stacks at the fixed address by executing `saveprevssp`.
```
joergroedel, line 514: There is another issue around the task-switch code which needs fixing, and that might help for shadow stacks as well.

The problem currently is that the task-switch assembly is not safe against exceptions (like #HV and #NMI) that can occur at any time. The unsafe part come from the fact that the code can not switch page-tables and stack pointers atomically. The solution for that problem is either using IST stacks for these exceptions, or using a per-cpu task-switch stack.

The preferred solution is a per-cpu task switch stack, which would then also need a shadow stack. I think this would solve this problem as well.

joergroedel (changes requested):
This is impressive, thanks for implementing this feature!

There are some open questions, one around the task-switch handling. I left a comment there. The other question is, whether there is a way to enable/disable this feature at boot-time, instead of compile time.

Finally, I found that it uses the old PageRef interface. Since the new one is merged now, this can be re-based on top of latest HEAD.

kernel/src/task/schedule.rs
Freax13: I didn't consider that IRQ may come in during the context switch. I implemented a CPU-local dedicated stack and shadow stack as you suggested.

Freax13:
> There are some open questions, one around the task-switch handling. I left a comment there. The other question is, whether there is a way to enable/disable this feature at boot-time, instead of compile time.

I added a patch that tries to detect CET support at runtime by enabling `CET` in `CR4` and catching any faults that might occur if not supported. This still doesn't technically detect shadow stack support because AFAICT CPUs could theoretically only support CET lBT (indirect branch tracking) and not CET SHSTK (shadow stacks), but in practice no such CPUs exist (and AFAICT TDX only allows turning all of CET on or off, not individual sub-parts).

> Finally, I found that it uses the old PageRef interface. Since the new one is merged now, this can be re-based on top of latest HEAD.

Done.

msft-jlange:
It doesn't look like this change is compatible with the existing #HV handling code.  For example, there is code in `asm_entry_hv` which specifically checks to see whether a recursive #HV has been delivered while executing an IRET sequence, and if so, it overwrites the old IRET data with the newly pushed IRET data.  The shadow stack equivalent of this code must be written, but it is also inherently unsafe, because it requires popping values from the shadow stack and using the WRSS instruction to write them to a previous location on the shadow stack.  Regardless of what approach we ultimately decide to take here, it is critical to support Restricted Injection correctly, and we can't allow shadow stack support to break it.

Freax13:
> It doesn't look like this change is compatible with the existing #HV handling code. For example, there is code in `asm_entry_hv` which specifically checks to see whether a recursive #HV has been delivered while executing an IRET sequence, and if so, it overwrites the old IRET data with the newly pushed IRET data. The shadow stack equivalent of this code must be written, but it is also inherently unsafe, because it requires popping values from the shadow stack and using the WRSS instruction to write them to a previous location on the shadow stack. Regardless of what approach we ultimately decide to take here, it is critical to support Restricted Injection correctly, and we can't allow shadow stack support to break it.

Interesting, thanks for pointing that out, I'll take a look at that. Do you know if there are KVM patches implementing restricted injection that I can test with?

msft-jlange:
> Do you know if there are KVM patches implementing restricted injection that I can test with?

Unfortunately, there are none that I am aware of.  However, I'm happy to review and test any changes you have in our environment which does support restricted injection.  Let me know if you want a more thorough description of how the existing #HV handler works and interacts with the IRET flow.

msft-jlange:
> AFAIK all processors supporting either AMD SEV-SNP or Intel TDX support shadow stacks

That's not actually true.  The Zen3 architecture supports SEV-SNP but not shadow stacks.

Freax13:
> > AFAIK all processors supporting either AMD SEV-SNP or Intel TDX support shadow stacks
> 
> That's not actually true.  The Zen3 architecture supports SEV-SNP but not shadow stacks.

EPYC Milan (based on Zen 3) supports shadow stacks. I used an EPYC Milan based machine to test these patches.

msft-jlange:
> EPYC Milan (based on Zen 3) supports shadow stacks. I used an EPYC Milan based machine to test these patches.

You are correct; I was misinformed.

Freax13:
> It doesn't look like this change is compatible with the existing #HV handling code. For example, there is code in `asm_entry_hv` which specifically checks to see whether a recursive #HV has been delivered while executing an IRET sequence, and if so, it overwrites the old IRET data with the newly pushed IRET data. The shadow stack equivalent of this code must be written, but it is also inherently unsafe, because it requires popping values from the shadow stack and using the WRSS instruction to write them to a previous location on the shadow stack. Regardless of what approach we ultimately decide to take here, it is critical to support Restricted Injection correctly, and we can't allow shadow stack support to break it.

I added a patch to adjust the shadow stack in the #HV.

> > Do you know if there are KVM patches implementing restricted injection that I can test with?
> 
> Unfortunately, there are none that I am aware of. However, I'm happy to review and test any changes you have in our environment which does support restricted injection. Let me know if you want a more thorough description of how the existing #HV handler works and interacts with the IRET flow.

Could you please test my patches on your environment? I tried testing by manually adding `int $28` to some places, but I still don't have a proper test setup, so I might have missed something. I also came across some bugs/regressions unrelated to shadow stacks, so I opened a separate PR to fix those: #466. You might need those changes to properly test the changes in this PR.

kernel/src/cpu/idt/entry.S
```diff
@@ -182,6 +201,15 @@ restart_hv:
 	movq	2*8(%rsp), %rax
 	movq	%rax, -2*8(%rcx)
 	leaq	-4*8(%rcx), %rsp
+        .if CFG_SHADOW_STACKS
+	// Pop the current stack frame, so that the previous stack frame sits
+	// on top of the shadow stack.
+	cmpb	$0, {IS_CET_SUPPORTED}(%rip)
+        je	2f
+	movq	$3, %rax
```
msft-jlange, line 209: Prefer `movl $3, %eax`.  This instruction is 4 bytes shorter since it moves a 32-bit constant instead of a 64-bit constant, and since `rax` is a 64-bit extension of `eax`, the subsequent instruction will be guaranteed to observe the correct value.

kernel/src/cpu/idt/entry.S
```diff
@@ -100,6 +111,14 @@ asm_entry_hv:
 	// RIP is in the return window, so update RIP to the cancel point.
 	leaq	switch_vmpl_cancel(%rip), %rbx
 	movq	%rbx, 0x20(%rsp)
+	.if CFG_SHADOW_STACKS
+	// Update RIP on the shadow stack to the cancel point.
+	cmpb    $0, {IS_CET_SUPPORTED}(%rip)
+        je 2f
+	rdsspq  %rax
+        wrss    %rbx, 8(%rax)
```
msft-jlange, line 119: Prefer `wrssq` because it makes the operand size more explicit.

kernel/src/cpu/idt/entry.S
```diff
@@ -29,6 +29,17 @@ HV_DOORBELL_ADDR:
         pushq   %r13
         pushq   %r14
         pushq   %r15
+	pushq   $0
+	.if CFG_SHADOW_STACKS
+	cmpb $0, {IS_CET_SUPPORTED}(%rip)
+        je 2f
+	rdsspq  %rax
```
msft-jlange, line 36: RDSSPQ is in the NOP space so it will behave as a NOP if CET is not enabled.  That means you can simplify this sequence to remove the jump by doing this instead:
```
xorq %eax, %eax // prefer the 32-bit form because it eliminates one instruction byte but has the same effect
rdsspq %rax
pushq %rax
```

kernel/src/cpu/idt/entry.S
msft-jlange, line 1: There is a problem with the logic to pop off an IRET frame when the previous RIP is not on the IRETQ instruction.  Look for this comment:
```
        // Switch to the stack pointer from the previous exception, which
        // points to the register save area, and continue with #HV
        // processing.
```
Immediately after that, there is existing code to reload `rsp` to switch to the stack frame that was live at the time the #HV occurred.  There must be equivalent code in your PR to switch the shadow stack pointer to the shadow stack frame that was live at the time the #HV occurred.

Note that this is different from the other place where the stack pointer is changed (where we pop a single IRET frame), which you have already addressed.

msft-jlange:
> Could you please test my patches on your environment? I tried testing by manually adding `int $28` to some places, but I still don't have a proper test setup, so I might have missed something. I also came across some bugs/regressions unrelated to shadow stacks, so I opened a separate PR to fix those: #466. You might need those changes to properly test the changes in this PR.

Other than the specific stack adjustment problem I commented on (the one that's missing), everything appears to work.  The other two shadow stack adjustments are correct.  And I found another problem with the #HV handler that's not addressed by your other PR so I'll submit a fix for that too.  Thanks for prompting me to review this code again!

ereshetova:
> KVM intercepts accesses to the MSRs used for shadow stacks, so some host-side [modifications](https://github.com/user->attachments/files/16982253/kvm-shstk.zip) are required to make this work.

Not very relevant yet, but this is not true for TDX case. The CET/shadow stack related MSRs are handled natively for the guest (given that the feature is enabled for the guest during its creation via config) with no host involvement (wont be secure otherwise). TDX module ensures this. So you should be able to enable this for TDX without any host/kvm support (but enabling TD_PARAMS.XFAM[11] || TD_PARAMS.XFAM[12] - matching XSAVE[11],XSAVE[12] for a given guest).

kernel/src/cpu/idt/entry.S
Freax13: Good catch, fixed.

> Note that this is different from the other place where the stack pointer is changed (where we pop a single IRET frame), which you have already addressed.

We also have to pop a single IRET frame here too though, don't we?

msft-jlange: > We also have to pop a single IRET frame here too though, don't we?

We do, but that is accomplished by loading the previous stack pointer from the top IRET frame.  This has the effect of popping the top IRET frame by restoring RSP to the stack pointer at the time of the exception (at which time it will point to the previous IRET frame).  It is always more correct to reload RSP from the top frame than to pop a fixed amount - same with shadow stack pointers - but that assembly is more complex.  For the time being, your logic will be fine and I can clean up the assembly in the future.

I'll test your most recent changes later in the day.

kernel/src/cpu/idt/entry.S
```diff
@@ -29,6 +29,12 @@ HV_DOORBELL_ADDR:
         pushq   %r13
         pushq   %r14
         pushq   %r15
+	pushq   $0
+	# rdsspq is a nop when shadow stacks are not supported. Make sure that
+	# rax is 0 if that's the case.
+	xorl	%eax, %eax
+	rdsspq  %rax
+	push    %rax
```
msft-jlange, line 37: Prefer `pushq` so the size is explicit.

Freax13:
> There's still one issue. You changed the size of the exception frame, and you've adjusted most of the places that refer to the stack frame accordingly. However, the reference at line 151 was never updated, and it is now inaccurate. So this sequence:
> 
> ```
>         testl   $0x200, 18*8(%rcx)
>         jz      postpone_hv
> ```
> 
> needs to become
> 
> ```
>         testl   $0x200, 20*8(%rcx)
>         jz      postpone_hv
> ```
> 
> If I make that change locally, then everything works as expected (or would work, if all of the other #HV-related PRs were included).

Good catch, should be fixed now! Thank you very much for your review!

msft-jlange (approved):

Freax13:
@joergroedel I think this PR is ready.

msft-jlange:
I just found one additional serious problem.

There are functions that generate page table flags to represent different page permissions.  However, some of these functions will make the PTE dirty without also marking it writable, and these functions will effectively generate shadow stack permissions, which is not correct for those permission types.  Specifically, all methods of `PTEntryFlags` other than `PTEntryFlags::data()` and `PTEntryFlags::task_data()` need to remove the `DIRTY` bit (anything that does not set `WRITABLE` should not set `DIRTY`).  These must be fixed or else the shadow stack logic will not be secure.

Freax13:
> I just found one additional serious problem.
> 
> There are functions that generate page table flags to represent different page permissions. However, some of these functions will make the PTE dirty without also marking it writable, and these functions will effectively generate shadow stack permissions, which is not correct for those permission types. Specifically, all methods of `PTEntryFlags` other than `PTEntryFlags::data()` and `PTEntryFlags::task_data()` need to remove the `DIRTY` bit (anything that does not set `WRITABLE` should not set `DIRTY`). These must be fixed or else the shadow stack logic will not be secure.

Good catch! Should be fixed now.

PR #456, opened 2024-09-14
Implement object management

cxdong:
This is the code implementation for the object management. It includes 3 patches, two patches are for the COCONUT-SVSM kernel, which add the Obj trait, ObjHandle and object management. One patch is for the user mode, which add the equivalent ObjHandle define.

The implementation is made according to the object management design code, which has been reviewed at the [PR 4530](https://github.com/coconut-svsm/svsm/pull/453). Thanks to @Freax13 for the invaluable reviews!

### DIFF
--- a/kernel/src/syscall/obj.rs
+++ b/kernel/src/syscall/obj.rs
@@ -4,6 +4,18 @@
 //
 // Author: Chuanxiao Dong <chuanxiao.dong@intel.com>
 
+extern crate alloc;
+
+use crate::cpu::percpu::current_task;
+use crate::error::SvsmError;
+use alloc::sync::Arc;
+
+#[derive(Clone, Copy, Debug)]
+pub enum ObjError {
+    InvalidHandle,
+    NotFound,
+}
+
 /// An object represents the type of resource like file, VM, vCPU in the
 /// COCONUT-SVSM kernel which can be accessible by the user mode. The Obj
 /// trait is defined for such type of resource, which can be used to define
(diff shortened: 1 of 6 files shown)


kernel/src/syscall/obj.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl From<ObjHandle> for u32 {
+    #[inline]
+    fn from(obj_handle: ObjHandle) -> Self {
+        obj_handle.0
+    }
+}
+
+pub type ObjPointer = Arc<dyn Obj>;
```
Freax13, line 54: I'm not a big fan of the name `ObjPointer`. The word "pointer" reminds me too much of raw pointers and this is not a raw pointer. How about `ObjReference` (or just `Arc<dyn Obj>)`?

Freax13 (approved):

kernel/src/syscall/obj.rs
cxdong: I see. Will update the patches and just use Arc<dyn Obj>. Thanks!

kernel/src/syscall/obj.rs
```diff
@@ -0,0 +1,113 @@
...
+
+extern crate alloc;
+
+use crate::cpu::percpu::current_task;
+use crate::error::SvsmError;
+use alloc::sync::Arc;
+
+#[derive(Clone, Copy, Debug)]
+pub enum ObjError {
+    InvalidId,
```
Freax13, line 15: Let's call this `InvalidHandle` (since we already use `ObjHandle`, not `ObjId`).

Freax13 (approved):

kernel/src/syscall/obj.rs
cxdong: I was just writing the comment about this change introduced in this update. Thanks for the review! Will update the name.

cxdong: Just rename to `InvalidHandle`. Now, as this part of code is a bit different with what the document in the [PR453](https://github.com/coconut-svsm/svsm/pull/453) mentioned, I will update the document as well.

kernel/src/task/tasks.rs
```diff
@@ -139,6 +141,9 @@ pub struct Task {
 
     /// Link to scheduler run queue
     runlist_link: LinkedListAtomicLink,
+
+    /// Objects shared among threads within the same process
+    objs: Arc<RWLock<BTreeMap<ObjHandle, Arc<dyn Obj>>>>,
```
joergroedel, line 146: Why is the outer `Arc` needed? Could as well be a `Box`, no?

Freax13: The `Arc` is needed so that multiple tasks in the same process can share the same object map.

joergroedel: That makes sense, thanks.

PR #458, opened 2024-09-18
tpm: Remove need for libssl build dependency

AdamCDunlap:
Coconut-SVSM vendors its own copy of openssl (via git submodule), but previously it didn't pass in the right variables to libmstpm's configure script to tell it to use it. This meant that libssl-dev must be installed on the build system in order to build Coconut.

This can be bypassed by passing LIBCRYPTO_LIBS and LIBCRYPTO_CFLAGS to configure. Coconut-SVSM was previously passing in LIBCRYPTO_LIBS only, but it would double-expanding the variables meaning that it was actually passing LIBCRYPTO_LIBS="", which was not enough for configure. Fix this, and also pass LIBCRYPTO_CFLAGS so that configure realizes the libcrypto dependency is satisfied without needing it installed on the build system.

Tested by building in a docker container based on debian12 without libssl-dev.

This commit also modifies the documentation and github workers to remove this dependency, but I have no way of directly testing these.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -28,7 +28,7 @@ jobs:
             override: true
 
       - name: Install Microsoft TPM build dependencies
-        run: sudo apt install -y autoconf autoconf-archive pkg-config build-essential automake libssl-dev
+        run: sudo apt install -y autoconf autoconf-archive pkg-config build-essential automake
 
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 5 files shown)


deeglaze:
Hurray!

stefano-garzarella:
@AdamCDunlap cool, thanks for that!

Can we also remove `libopenssl-devel` from `scripts/container/opensuse-rust.docker` ?

AdamCDunlap:
> @AdamCDunlap cool, thanks for that!
> 
> Can we also remove `libopenssl-devel` from `scripts/container/opensuscrpits/conse-rust.docker` ?

Pushed a new version with that change. Thanks for taking a look.

stefano-garzarella (approved):

PR #460, opened 2024-09-19
get rid of unnecessary uses of unsafe

Freax13:
This PR get's rid of some unnecessary uses of unsafe mainly by switching to safe alternatives provided by the zerocopy crate.

### DIFF
--- a/kernel/src/mm/pagebox.rs
+++ b/kernel/src/mm/pagebox.rs
@@ -206,36 +206,28 @@ impl<T: ?Sized> DerefMut for PageBox<T> {
 impl<T: ?Sized> borrow::Borrow<T> for PageBox<T> {
     #[inline]
     fn borrow(&self) -> &T {
-        // SAFETY: this is part of the invariants of this type, as it must
-        // hold a pointer to valid memory for the given `T`.
-        unsafe { self.ptr.as_ref() }
+        self
     }
 }
 
 impl<T: ?Sized> borrow::BorrowMut<T> for PageBox<T> {
     #[inline]
     fn borrow_mut(&mut self) -> &mut T {
-        // SAFETY: this is part of the invariants of this type, as it must
-        // hold a pointer to valid memory for the given `T`.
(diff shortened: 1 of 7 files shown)


kernel/src/mm/pagebox.rs
```diff
@@ -197,36 +197,28 @@ impl<T: ?Sized> DerefMut for PageBox<T> {
 impl<T: ?Sized> borrow::Borrow<T> for PageBox<T> {
     #[inline]
     fn borrow(&self) -> &T {
-        // SAFETY: this is part of the invariants of this type, as it must
-        // hold a pointer to valid memory for the given `T`.
-        unsafe { self.ptr.as_ref() }
+        self
```
msft-jlange, line 200: If this function doesn't require any manipulation of `self` before returning, then why keep the function body at all?  I just tried an experiment where I completely removed the four functions you're touching here, and the compiler is perfectly happy, so it doesn't seem like we get any benefit from keeping functions that contain no logic.

00xc: The reason is that these functions implement traits that make it easier to work with generics. For example any function that takes a generic type `S` such that `S: AsRef<T>` will also accept `PageBox<T>`.

Just because the code is not needed now doesn't mean we should remove it, like the usual derives we have on most types.

msft-jlange: Makes sense - thanks!  I think that means that as we introduce other similar types (such as the introduction of `SharedBox<T>` in #451) then those types that can implicitly support `AsRef<T>` or `borrow` and the like should also implement similar skeletal functions so that they are there if/when needed.

00xc: I haven't looked into that PR in depth, but I'd say that's a good idea, yes.

Freax13: `SharedBox<T>` doesn't allow creating references (shared or mutable) to the pointed-to data, so it doesn't need implementations for these traits.

PR #462, opened 2024-09-19
debug: implement `GdbStubConnection` on top of the platform abstraction

msft-jlange:
The GDB stub connection needs to communicate over the serial port.  Like other serial port accesses, this should be done using the I/O port driver implemented by the platform abstraction instead of assuming the GHCB-based I/O driver.


kernel/src/debug/gdbstub.rs
```diff
@@ -306,25 +303,27 @@ pub mod svsm_gdbstub {
...
-    struct GdbStubConnection;
+    struct GdbStubConnection<'a> {
+        serial_port: SerialPort<'a>,
+    }
 
-    impl GdbStubConnection {
-        const fn new() -> Self {
-            Self {}
+    impl GdbStubConnection<'_> {
+        fn new(platform: &'static dyn SvsmPlatform) -> Self {
```
Freax13, line 311:
```suggestion
    impl<'a> GdbStubConnection<'a> {
        fn new(platform: &'a dyn SvsmPlatform) -> Self {
```

Or just change the lifetime of the `serial_port` field to `'static`.

Freax13 (commented):
Nice!

kernel/src/debug/gdbstub.rs
msft-jlange: Since `GdbStubConnection` is already typed with a lifetime, adding the same lifetime to the `& dyn SvsmPlatform` is an easy change.

roy-hopkins (approved):
I've had a quick look at the code which looks good to me, but I've spent some time actually testing the gdbstub functionality with this branch.

I've checked the following operations:

* Initial break into debugger with `debug_break()` call.
* Single stepping
* Correct symbol locations
* Setting and hitting breakpoints
* Analysis of data and threads.

Everything worked correctly.

PR #463, opened 2024-09-19
platform: move I/O port driver into platform abstraction

msft-jlange:
The `SVSMIOPort` and `NativeIOPort` objects are platform-specific, and thus they should be implemented in the platform abstraction instead of in common code.  Furthermore, the console serial port object is common and need not be duplicated in every platform abstraction.


msft-jlange:
@peterfang This should enable you to implement the correct GHCI-based abstraction for I/O port accesses.  After making a couple of different rounds of changes, it became apparent that the best way to make this work was not to have the console logic use native I/O operations, but to correctly implement a platform-specific version of the `IOPort` object which can be used not only by the console, but also by the #VE and #VC handler.  Assuming the #VE handler calls the common instruction emulator, this should be wired up correctly because the common emulator calls through the platform abstraction to perform the I/O accesses.

Feel free to test on top of this.  However, this cannot merge until #462 merges because #462 changes the gdbstub adapter to use the platform abstraction as well, which is a prerequisite for the refactoring in this PR.

peterfang:
> @peterfang This should enable you to implement the correct GHCI-based abstraction for I/O port accesses. After making a couple of different rounds of changes, it became apparent that the best way to make this work was not to have the console logic use native I/O operations, but to correctly implement a platform-specific version of the `IOPort` object which can be used not only by the console, but also by the #VE and #VC handler. Assuming the #VE handler calls the common instruction emulator, this should be wired up correctly because the common emulator calls through the platform abstraction to perform the I/O accesses.
> 
> Feel free to test on top of this. However, this cannot merge until #462 merges because #462 changes the gdbstub adapter to use the platform abstraction as well, which is a prerequisite for the refactoring in this PR.

Got it. I'll keep track of this PR and make updates to https://github.com/coconut-svsm/svsm/pull/419 accordingly. This will also be incorporated into #VE handling in the future.

peterfang:
@msft-jlange Quick update: I combined all 3 of your PRs (https://github.com/coconut-svsm/svsm/pull/461, https://github.com/coconut-svsm/svsm/pull/462, https://github.com/coconut-svsm/svsm/pull/463) and updated my stage2 changes accordingly. I was able to boot stage2 successfully.

joergroedel:
`kernel/src/testing.rs` needs an update as well, `test-in-svsm` does not compile anymore with these changes.

msft-jlange:
> `kernel/src/testing.rs` needs an update as well, `test-in-svsm` does not compile anymore with these changes.

It compiles now.  Hopefully it works too.

peterfang:
Tested this PR on a TDP platform with https://github.com/coconut-svsm/svsm/pull/419 and it worked as expected.

PR #464, opened 2024-09-20
insn_decode: add decoding and emulation support for mov instruction variants

cxdong:
The Mov instruction is usually used to access MMIO. Add decoding and emulation support for several Mov instruction variants, which includes:
    - 0x88/0x89/0xA3: move general-purpose register to memory address.
    - 0x8A/0x8B/0xA1: move value from memory address to general-purpose register.
    - 0xC6/0XC7: move immediate to memory address.

Test code is added for the corresponding mov instructions.


kernel/src/insn_decode/decode.rs
```diff
@@ -523,6 +583,32 @@ fn ioio_perm<I: InsnMachineCtx>(mctx: &I, port: u16, size: Bytes, io_read: bool)
     }
 }
 
+#[inline]
+fn read_bytereg<I: InsnMachineCtx>(mctx: &I, reg: Register, lhbr: bool) -> u8 {
+    let data = mctx.read_reg(reg);
```
p4zuu, line 588: I guess you can cast to `u8` at the beginning to avoid cast duplicate: 
```suggestion
    let data = mctx.read_reg(reg) as u8;
```

kernel/src/insn_decode/insn.rs
```diff
@@ -275,6 +279,59 @@ pub mod test_utils {
...
+            _shared: bool,
+            size: Bytes,
+        ) -> Result<u64, InsnError> {
+            if pa != core::ptr::addr_of!(self.mmio_reg) as usize {
+                return Ok(0);
+            }
+
+            let data = unsafe { *(pa as *const u64) };
+            match size {
+                Bytes::One => Ok(data as u8 as u64),
```
p4zuu, line 304: Is there a reason why not returning `Ok(data)` here?

kernel/src/insn_decode/insn.rs
```diff
@@ -275,6 +279,59 @@ pub mod test_utils {
...
+            data: u64,
+        ) -> Result<(), InsnError> {
+            if pa != core::ptr::addr_of!(self.mmio_reg) as usize {
+                return Ok(());
+            }
+
+            let addr = pa as *mut u64;
+            unsafe {
+                match size {
+                    Bytes::One => *addr = data as u8 as u64,
```
p4zuu, line 326: Same about `Ok(data)`?

p4zuu (commented):
Left a few comments about types and casting. Otherwise, looks good to me

kernel/src/insn_decode/decode.rs
cxdong: We may need to get the high-byte of data if the lhbr is true in the below if case. If cast data as u8 here, the high-byte is lost. How about change below code as `(if lhbr { data >> 8 } else { data }) as u8`?

kernel/src/insn_decode/insn.rs
cxdong: The consideration is that, for the Bytes::One case, to make sure only the lowest byte of the data has valid value, and the other bytes are cleared as 0. And the same consideration for the Bytes::Two/Four.

kernel/src/insn_decode/insn.rs
cxdong: Looks like should cast the pa to the address pointer according to the size instead of always a *mut u64, otherwise writing an u8/u16/u32 will break the original u64 value stored in this pa. How about change to `Bytes::One => unsafe { *(pa as *mut u8) = data as u8 },`?

kernel/src/insn_decode/decode.rs
p4zuu: Yes, this is better indeed :)

kernel/src/insn_decode/insn.rs
p4zuu: Ah yes you're right, I misread it. This makes sens now. You could also use `Ok(data & 0xff)` etc, but I'm not sure if's better so I think you can leave it as it is.

kernel/src/insn_decode/insn.rs
p4zuu: Misread again sorry, I guess you could again use `Ok(data & 0xff)` or leave it as it is. :)

kernel/src/insn_decode/decode.rs
cxdong: I will update with this change. Thanks!

kernel/src/insn_decode/insn.rs
cxdong: Actually, I think casting pa as a *mut u64 regardless of the size is not a right thing, which can result in overwriting data unexpected if the size is Bytes::Ont/Two/Four, so similar to the read case, I prefer to cast pa to *mut u8/u16/u32/u64 according to the size. Thanks!

p4zuu (approved):

PR #467, opened 2024-09-25
Implement initial userspace support 

vijaydhanraj:
This is a draft PR that adds basic syscall support like exit, close, opendir, readdir and exec as well as init task and dm implementation. The purpose of this PR is to get feedback from the community and ensure the design aligns with principles of user-mode support described in https://github.com/coconut-svsm/svsm/blob/main/Documentation/docs/developer/DEVELOPMENT-PLAN.md as well syscall ABIs https://mail.8bytes.org/pipermail/svsm-devel/2024-June/000334.html.

PS: The changes are built on top of https://github.com/coconut-svsm/svsm/pull/456.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ IGVMBIN = bin/igvmbld
 IGVMMEASURE = "target/x86_64-unknown-linux-gnu/${TARGET_PATH}/igvmmeasure"
 IGVMMEASUREBIN = bin/igvmmeasure
 INITELF = "target/x86_64-unknown-none/${TARGET_PATH}/init"
+DMELF = "target/x86_64-unknown-none/${TARGET_PATH}/dm"
 
 RUSTDOC_OUTPUT = target/x86_64-unknown-none/doc
 DOC_SITE = target/x86_64-unknown-none/site
@@ -140,7 +141,11 @@ bin/init: bin
 	cargo build --manifest-path init/Cargo.toml ${CARGO_ARGS} --bin init
 	objcopy -O elf64-x86-64 --strip-unneeded ${INITELF} $@
 
-user: bin/init
+bin/dm: bin
+	cargo build --manifest-path dm/Cargo.toml ${CARGO_ARGS} --bin dm
+	objcopy -O elf64-x86-64 --strip-unneeded ${DMELF} $@
+
+user: bin/init bin/dm
 
 ${FS_BIN}: bin
 ifneq ($(FS_FILE), none)
@@ -183,7 +188,7 @@ bin/svsm-test.bin: bin/stage1-test
 
 clippy:
 	cargo clippy --workspace --all-features --exclude svsm-fuzz --exclude igvmbuilder --exclude igvmmeasure -- -D warnings
-	cargo clippy --workspace --all-features --exclude svsm-fuzz --exclude svsm --exclude init --target=x86_64-unknown-linux-gnu -- -D warnings
+	cargo clippy --workspace --all-features --exclude svsm-fuzz --exclude svsm --exclude init --exclude dm --target=x86_64-unknown-linux-gnu -- -D warnings
 	RUSTFLAGS="--cfg fuzzing" cargo clippy --package svsm-fuzz --all-features --target=x86_64-unknown-linux-gnu -- -D warnings
 	cargo clippy --workspace --all-features --tests --target=x86_64-unknown-linux-gnu -- -D warnings
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -62,7 +62,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --workspace --all-features --exclude svsm --exclude svsm-fuzz --exclude init --target=x86_64-unknown-linux-gnu -- -D warnings
+          args: --workspace --all-features --exclude svsm --exclude svsm-fuzz --exclude init --exclude dm --target=x86_64-unknown-linux-gnu -- -D warnings
 
       - name: Clippy on svsm-fuzz
         uses: actions-rs/cargo@v1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -344,6 +344,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "dm"
+version = "0.1.0"
+dependencies = [
+ "syscall",
+]
+
+[[package]]
 name = "ecdsa"
 version = "0.16.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,8 @@ members = [
     "syscall",
     # init process
     "init",
+    # device model application
+    "dm",
 ]
 
 
--- a/dm/Cargo.toml
+++ b/dm/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "dm"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "dm"
+path = "src/main.rs"
+test = false
+
+[dependencies]
+syscall = { path = "../syscall" }
+
+[lints]
+workspace = true
--- a/dm/build.rs
+++ b/dm/build.rs
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+fn main() {
+    println!("cargo:rustc-link-arg-bin=dm=-nostdlib");
+    println!("cargo:rustc-link-arg-bin=dm=-no-pie");
+    println!("cargo:rustc-link-arg-bin=dm=-Tdm/dm.lds");
+}
--- a/dm/dm.lds
+++ b/dm/dm.lds
@@ -1,0 +1,37 @@
+/* Temporary WA until p_vaddr alignment issue is fixed in our ELF parser */
+
+OUTPUT_ARCH(i386:x86-64)
+
+PHDRS
+{
+	text PT_LOAD FLAGS(5);		/* Read + Execute */
+	rodata PT_LOAD FLAGS(4);	/* Read-only */
+	data PT_LOAD FLAGS(6);		/* Read + Write */
+	bss PT_LOAD FLAGS(6);		/* Read + Write */
+}
+
+SECTIONS
+{
+	.text : {
+		*(.text)
+		*(.text.*)
+	} :text
+	. = ALIGN(4096);
+	.rodata : {
+		*(.rodata)
+		*(.rodata.*)
+	} :rodata
+	. = ALIGN(4096);
+	.data : {
+		*(.data)
+		*(.data.*)
+	} :data
+	. = ALIGN(4096);
+	.bss : {
+		*(.bss) *(.bss.*)
+		. = ALIGN(4096);
+	} :bss
+	. = ALIGN(4096);
+}
+
+ENTRY(dm_start)
--- a/dm/src/main.rs
+++ b/dm/src/main.rs
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2024 Intel Corporation.
+//
+// Author: Chuanxiao Dong <chuanxiao.dong@intel.com>
+
+#![no_std]
+#![no_main]
+
+use core::panic::PanicInfo;
+use syscall::exit;
+
+fn dm_exit() -> ! {
+    exit(0);
+}
+
+#[no_mangle]
+pub extern "C" fn dm_start() -> ! {
+    dm_exit();
+}
+
+#[panic_handler]
+fn panic(_info: &PanicInfo<'_>) -> ! {
+    dm_exit();
+}


syscall/src/def.rs
```diff
@@ -0,0 +1,58 @@
...
+pub const ENOSYS: i32 = -2;
+pub const ENOMEM: i32 = -3;
+pub const EPERM: i32 = -4;
+pub const EFAULT: i32 = -5;
+pub const EBUSY: i32 = -6;
+pub const ENOTFOUND: i32 = -7;
+pub const ENOTSUPP: i32 = -8;
+
+/// Maximum length of file name in bytes
+pub const F_NAME_SIZE: usize = 256;
```
Freax13, line 31: AFAICT the max size inside the kernel is 64 bytes. Why are we using a different size here?

kernel/src/syscall/class0.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::address::VirtAddr;
+use crate::cpu::percpu::current_task;
+use crate::error::SvsmError;
+use crate::fs::{find_dir, FsError};
+use crate::task::{current_task_terminated, exec_user, schedule, TaskError};
+use core::ffi::{c_char, CStr};
+use syscall::{EINVAL, ENOTFOUND};
+
+pub fn sys_exit(exit_code: u32) -> ! {
+    log::info!("Terminating current task, exit_code {}", exit_code);
```
Freax13, line 17:
```suggestion
    log::info!("Terminating current task, exit_code {exit_code}");
```

init/init.lds
Freax13, line 1: Why do we need this linker script?

Makefile
```diff
@@ -170,7 +182,7 @@ bin/svsm-test.bin: bin/stage1-test
 
 clippy:
 	cargo clippy --workspace --all-features --exclude svsm-fuzz --exclude igvmbuilder --exclude igvmmeasure -- -D warnings
-	cargo clippy --workspace --all-features --exclude svsm-fuzz --exclude svsm --target=x86_64-unknown-linux-gnu -- -D warnings
+	cargo clippy --workspace --all-features --exclude svsm-fuzz --exclude svsm --exclude init --exclude dm --target=x86_64-unknown-linux-gnu -- -D warnings
```
Freax13, line 185: Why are these excluded?

init/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
Freax13, line 1: Cargo won't create a `target` directory here (same goes for `dm`).

init/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    if binfile.is_empty() {
+        exit(0);
+    }
+
+    let mut file = String::from(bin.as_c_str().to_str().unwrap());
+    file.push_str(binfile.as_c_str().to_str().unwrap());
+
+    let file = CString::new(file).expect("Failed to create new string");
+    let root = CString::new("/").expect("Failed to create new string");
```
Freax13, line 61: IMHO these messages don't really say much, we might as well use unwrap here.

Freax13 (commented):
Some preliminary review comments.

syscall/src/class0.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use super::call::{syscall1, syscall3, SysCallError};
+use super::{SYS_EXEC, SYS_EXIT};
+use core::ffi::CStr;
+
+pub fn exit(code: u32) -> ! {
+    // # Safety
+    // SYS_EXIT is supported syscall number by the svsm kernel.
+    unsafe {
+        syscall1(SYS_EXIT, code as u64);
```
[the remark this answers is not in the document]
vijaydhanraj: Sure, will do.

syscall/src/class0.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub fn exec(file: &CStr, root: &CStr, flags: u32) -> Result<Tid, SysCallError> {
+    match unsafe {
+        syscall3(
+            SYS_EXEC,
+            file.as_ptr() as u64,
+            root.as_ptr() as u64,
+            flags as u64,
+        )
+    } as i64
```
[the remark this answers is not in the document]
vijaydhanraj: Agree, I will refactor the syscalls.

syscall/src/class1.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+pub fn opendir(path: &CStr) -> Result<FsObjHandle, SysCallError> {
+    match unsafe { syscall1(SYS_OPENDIR, path.as_ptr() as u64) } as i64 {
+        ret if ret >= 0 => Ok(FsObjHandle(ObjHandle::new(ret as u32))),
+        ret => Err(SysCallError::from(ret as i32)),
+    }
+}
+
+pub fn readdir(fs: &FsObjHandle, dirents: &mut [DirEnt]) -> Result<u64, SysCallError> {
```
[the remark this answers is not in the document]
vijaydhanraj: Yes, will do.

syscall/src/def.rs
vijaydhanraj: The spec says max length of file name in bytes is 256, so have adopted this.

kernel/src/string.rs
```diff
@@ -62,6 +62,16 @@ impl<const N: usize> From<[u8; N]> for FixedString<N> {
     }
 }
 
+impl<const N: usize> From<FixedString<N>> for [u8; N] {
+    fn from(src: FixedString<N>) -> [u8; N] {
+        let mut data = [0u8; N];
+        for (i, d) in data.iter_mut().enumerate().take(src.length()) {
+            *d = src.data[i] as u8;
+        }
```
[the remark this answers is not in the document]
Freax13: I don't know of a good solution to this. Maybe we need to rethink `FixedString`. Until we come up with a proper solution, I'm fine leaving this as is, but we should probably add a `// TODO` here to make sure we don't forget.

kernel/src/syscall/class1.rs
```diff
@@ -0,0 +1,69 @@
...
+            _ => EINVAL,
+        })?;
+        obj_add(Arc::new(FsObj::new_dir(&dir))).map_or(Err(EINVAL), |id| Ok(u32::from(id).into()))
+    }
+}
+
+pub fn sys_readdir(obj_id: u32, dirents: usize, size: usize) -> Result<u64, i32> {
+    let fsobj = obj_get(obj_id.into()).map_err(|_| ENOTFOUND)?;
+    let fsobj = fsobj.as_fs().ok_or(ENOTSUPP)?;
+    let dirents = VirtAddr::from(dirents).as_mut_ptr::<DirEnt>();
```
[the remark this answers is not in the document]
Freax13: 1. There are other syscalls where we need to check the pointers passed in by userspace. Please also update the other syscalls.
2. `GuestPtr` doesn't prevent the user from pointing into kernel memory. Please make sure that userspace doesn't do that.

kernel/src/task/exec.rs
```diff
@@ -60,7 +64,23 @@ pub fn exec_user(binary: &str) -> Result<(), SvsmError> {
         }
 
         if file_offset > 0 {
-            task.mmap_user(virt_start, Some(&fh), file_offset, len, flags)?;
+            let valid_file_size = align_up(
+                seg.file_range.offset_end - seg.file_range.offset_begin,
+                PAGE_SIZE,
+            );
```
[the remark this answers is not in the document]
Freax13: No, my point is that anything after `[p_filesz..p_memsz)` must be zero-filled by the loader. The loader cannot assume that the file contains zeroes after `p_filesz` until the end of the page. If we just align up the size and mmap the file, we might map non-zero values into userspace where zero-values are expected.

init/init.lds
Freax13: That's a bug in our elf parser. AFAIK the ELF spec doesn't not guarantee that `p_vaddr` is page aligned, so the output you showed doesn't violate the spec. I'm fine with leaving this as is for now, adding a `// TODO` comment (or opening an issue), and fixing this in a later PR. We shouldn't need to use a linker script here.

vijaydhanraj: Got it, but my understanding from the spec was that `p_align` gives the value to which the segments are aligned in memory and in the file. So, since the addresses were not aligned by `0x1000` felt the parser did the right thing by flagging this as an error. If needed I will go ahead and create a separate issue.
```
  LOAD      0x000000000000c488 0x000000000040d488 0x000000000040d488 
            0x0000000000000b78 0x0000000000010ca0  RW     **0x1000**
```

kernel/src/syscall/class1.rs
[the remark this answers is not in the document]
vijaydhanraj: Sure, will apply the same logic to other syscalls as well.
Regarding the `GuestPtr` not preventing the user from pointing into kernel memory, I see 2 options. Is there a preference?

1. We could compare the VirtAddr with USER_MEM_* macros like below,
```
    if dirents < USER_MEM_START || dirents >= USER_MEM_END {
        return Err(EINVAL);
    }
```

2. Or we could get the current_task() and get the vm_user_range (probably need to introduce `get_user_vmr`) for the task. And then check if the `VirtAddr` falls withing the range.
```
let vmr_range = current_task().get_user_vmr().map_err(|_| EINVAL)?;
if dirents < vmr_range.0 || dirents >= vmr_range.1 {
    return Err(EINVAL);
} 
```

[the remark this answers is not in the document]
Freax13: > Sure, will apply the same logic to other syscalls as well. Regarding the `GuestPtr` not preventing the user from pointing into kernel memory, I see 2 options. Is there a preference?
> 
>   1. We could compare the VirtAddr with USER_MEM_* macros like below,
> 
> 
> ```
>     if dirents < USER_MEM_START || dirents >= USER_MEM_END {
>         return Err(EINVAL);
>     }
> ```
> 
>   2. Or we could get the current_task() and get the vm_user_range (probably need to introduce `get_user_vmr`) for the task. And then check if the `VirtAddr` falls withing the range.
> 
> 
> ```
> let vmr_range = current_task().get_user_vmr().map_err(|_| EINVAL)?;
> if dirents < vmr_range.0 || dirents >= vmr_range.1 {
>     return Err(EINVAL);
> } 
> ```

The first option seems preferable to me. It's probably worthwhile introducing a new `UserPtr` wrapper type around `GuestPtr` that automatically does those checks for any reads/writes.

kernel/src/task/exec.rs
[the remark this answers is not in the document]
Freax13: > ah I get your point; you mean something like below right?
> 
> ```
> diff --git a/kernel/src/task/exec.rs b/kernel/src/task/exec.rs
> index ea1964d6..bd0e094c 100644
> --- a/kernel/src/task/exec.rs
> +++ b/kernel/src/task/exec.rs
> @@ -13,7 +13,7 @@ use crate::mm::vm::VMFileMappingFlags;
>  use crate::mm::USER_MEM_END;
>  use crate::task::{create_user_task, current_task, schedule, TaskError};
>  use crate::types::PAGE_SIZE;
> -use crate::utils::align_up;
> +use crate::utils::{align_up, zero_mem_region};
>  use alloc::sync::Arc;
>  use elf::{Elf64File, Elf64PhdrFlags};
>  
> @@ -64,13 +64,14 @@ pub fn exec_user(binary: &str, root: Arc<dyn Directory>) -> Result<u32, SvsmErro
>          }
>  
>          if file_offset > 0 {
> -            let valid_file_size = align_up(
> -                seg.file_range.offset_end - seg.file_range.offset_begin,
> -                PAGE_SIZE,
> -            );
> +            let p_filesz = seg.file_range.offset_end - seg.file_range.offset_begin;
> +            let valid_file_size = align_up(p_filesz, PAGE_SIZE);
>  
>              if len > valid_file_size {
>                  task.mmap_user(virt_start, Some(&fh), file_offset, valid_file_size, flags)?;
> +                if p_filesz % PAGE_SIZE != 0 {
> +                    zero_mem_region(virt_start + p_filesz, virt_start + valid_file_size);
> +                }
>                  task.mmap_user(
>                      virt_start + valid_file_size,
>                      None,
> ```

Yup.

> [Update] Tried a quick experiment with above change and it throws an unhandled page-fault error. Will look into this. No issues observed when the memory is not zeroed out.
> 
> ```
> [SVSM] ERROR: Panic: CPU[0] panicked at kernel/src/cpu/idt/svsm.rs:198:9:
> Unhandled Page-Fault at RIP 0xffffff800017034d CR2: 0x000000000000cbb0 error code: 0x0000000000000002
> [SVSM] ---BACKTRACE---:
> [SVSM]   [0xffffff8000170b1e]
> [SVSM]   [0xffffff800017034d]
> [SVSM]   [0xffffff800014d0f6]
> [SVSM]   [0xffffff800012fb54]
> [SVSM]   [0xffffff8000140ab0]
> [SVSM] ---END---
> qemu-system-x86_64: terminating on signal 2
> ```

I think the page tables for `task` aren't active.

> No issues observed when the memory is not zeroed out.

That's bound to happen eventually.

init/init.lds
Freax13: Nope, that's not the case. `p_align` does not specific the alignment for `p_vaddr`, it only ensures that `p_vaddr % p_align == p_offset % p_align`.

kernel/src/task/exec.rs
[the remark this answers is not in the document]
vijaydhanraj: Yes, you are correct, the page tables for the new task aren't active and so the error. Few options to resolve the issue,
1. Switch the page tables to the new task, zeroed the memory and switched back to the kernel task page tables. But this doesn't seem like a clean solution.
2. mmap the pages as anonymous as part of the kernel task itself, clear the range `[p_filesz..p_memsz)` and update the new task's VMR range to point to this physical memory ensuring that ref count is not decremented when the mapping is dropped. (No implementation exists currently; we may need to implement this)
3. Customize the linker script in such a way that `.bss` section has a separate segment. This way we can avoid clearing `[p_filesz..p_memsz)` as we can simply mmap this entire region as an anonymous region.

For this PR I have reverted to approach 3, can you please share your thoughts on how we should go about this?

kernel/src/syscall/class1.rs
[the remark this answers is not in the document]
vijaydhanraj: Created a wrapper `UserPtr` around `GuestPtr` that automatically does those checks for `VirtAddr ` to be within `USER_MEM_START/END` range.

vijaydhanraj:
> Some preliminary review comments.

Thanks @Freax13 for the valuable comments. I have addressed all the earlier comments. Can you please take a look?

Freax13 (changes requested):
I don't have time for a full review right now, I'll do that sometime next week.

Freax13:
Now that #456 has been merged, this PR can be rebased onto `main` and probably be marked as "ready to review" (instead of as a draft).

vijaydhanraj:
> Now that #456 has been merged, this PR can be rebased onto `main` and probably be marked as "ready to review" (instead of as a draft).

Sure, will rebase, address your other comments and mark the PR ready to review.

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +277,73 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+pub struct UserPtr<T: Copy> {
+    guest_ptr: GuestPtr<T>,
+}
+
+impl<T: Copy> UserPtr<T> {
+    #[inline]
+    pub fn new(v: VirtAddr) -> Result<Self, SvsmError> {
+        if v < USER_MEM_START || v >= USER_MEM_END {
+            return Err(SvsmError::InvalidAddress);
+        }
```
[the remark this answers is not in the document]
vijaydhanraj: > We need to consider the size of T when doing the bounds checks.

Yes, you are correct. We need to account for the size of `T`. Thanks for the suggestion, I have implemented likewise.

> That said, I'd prefer if we did those checks in the read and write methods (or in another method called by those methods). 

Yes, make sense.  I have implemented a separate function to do the bounds check and invoke it from the `read` and `write` methods.

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +277,73 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+        Ok(Self {
+            guest_ptr: GuestPtr::new(v),
+        })
+    }
+
+    #[inline]
+    pub fn as_ptr(&self) -> *const T {
+        self.guest_ptr.as_ptr()
+    }
```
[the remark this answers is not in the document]
vijaydhanraj: Agree, have removed this function.

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +277,73 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+
+    #[inline]
+    pub const fn from_ptr(p: *mut T) -> Self {
+        Self {
+            guest_ptr: GuestPtr::from_ptr(p),
+        }
+    }
```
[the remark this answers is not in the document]
vijaydhanraj: Yes, removed this function.

kernel/src/syscall/class0.rs
```diff
@@ -0,0 +1,55 @@
...
+        .map(|v| v.as_ptr())
+        .map_err(|_| EINVAL)?;
+
+    let user_root_str = UserPtr::<c_char>::new(VirtAddr::from(root))
+        .map(|v| v.as_ptr())
+        .map_err(|_| EINVAL)?;
+
+    unsafe {
+        let file_str = CStr::from_ptr(user_file_str).to_str().map_err(|_| EINVAL)?;
+        let root_str = CStr::from_ptr(user_root_str).to_str().map_err(|_| EINVAL)?;
```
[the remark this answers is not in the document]
vijaydhanraj: Thanks for the clarification. I have added a new `read_c_string` method to read from null-terminated c-style strings.

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +281,110 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+}
+
+impl UserPtr<c_char> {
+    /// # Safety
+    ///
+    /// Reads a null-terminated C string from the user space into the provided buffer.
+    /// Returns a `&str` reference to the buffer.
+    pub unsafe fn read_c_string<'a>(&self, buffer: &'a mut Vec<u8>) -> Result<&'a str, SvsmError> {
```
Freax13, line 370: We should limit the size of the string. Can we just use `&mut [u8]` instead of `Vec<u8>`? That would also avoid allocating to the heap.

syscall/src/def.rs
```diff
@@ -0,0 +1,58 @@
...
+
+// Syscall error code number
+pub const EINVAL: i32 = -1;
+pub const ENOSYS: i32 = -2;
+pub const ENOMEM: i32 = -3;
+pub const EPERM: i32 = -4;
+pub const EFAULT: i32 = -5;
+pub const EBUSY: i32 = -6;
+pub const ENOTFOUND: i32 = -7;
+pub const ENOTSUPP: i32 = -8;
```
Freax13, line 28: Why do we need this when we have `SysCallError`? Can't we just use `SysCallError` everywhere?

kernel/src/syscall/class0.rs
```diff
@@ -0,0 +1,58 @@
...
+    panic!("schedule() returned in sys_exit()");
+}
+
+pub fn sys_exec(file: usize, root: usize, _flags: usize) -> Result<u64, i32> {
+    let user_file_ptr = UserPtr::<c_char>::new(VirtAddr::from(file)).map_err(|_| EINVAL)?;
+    let user_root_ptr = UserPtr::<c_char>::new(VirtAddr::from(root)).map_err(|_| EINVAL)?;
+    let mut file_buffer = Vec::new();
+    let mut root_buffer = Vec::new();
+
+    unsafe {
```
Freax13, line 34: Please reduce the size of `unsafe` blocks. Most of the code in there isn't unsafe, but that's difficult to tell when everything is wrapped in unsafe. This also applies to `sys_opendir`.

kernel/src/syscall/class1.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub fn sys_readdir(obj_id: u32, dirents: usize, size: usize) -> Result<u64, i32> {
+    let fsobj = obj_get(obj_id.into()).map_err(|_| ENOTFOUND)?;
+    let fsobj = fsobj.as_fs().ok_or(ENOTSUPP)?;
+    let user_dirents_ptr = UserPtr::<DirEnt>::new(VirtAddr::from(dirents)).map_err(|_| EINVAL)?;
+
+    for i in 0..size {
+        match fsobj.readdir() {
+            Ok(Some((name, dirent))) => {
+                let mut new_entry = DirEnt::default();
```
Freax13, line 43: Can we reduce the nesting here? We should treat `Ok(Some(...))` as the happy case and bail out early in the other cases.

kernel/src/task/exec.rs
```diff
@@ -40,13 +43,16 @@ pub fn exec_user(binary: &str) -> Result<(), SvsmError> {
         VMFileMappingFlags::Read,
     )?;
     let buf = unsafe { vstart.to_slice::<u8>(file_size) };
+    // TODO: Due to a bug in the ELF parser, this function returns an error when p_vaddr is not
+    // aligned. But since spec doesn't guarantee that p_vaddr must be aligned, our ELF parser
+    // needs to be fixed. Current WA is to use a custom linker script to align p_vaddr.
```
Freax13, line 48: Please move this comment to the buggy code, not the call-site. Maybe also add that comment to the linker scripts themselves.

syscall/src/call.rs
```diff
@@ -0,0 +1,95 @@
...
+                    out("rcx") _,
+                    out("r11") _,
+                    options(nostack),
+                );
+
+                if ret > i32::MAX as u64 {
+                    // Convert to negative error code
+                    let error_code = ((ret as i64) - (1 << 32)) as i32;
+                    return Err(SysCallError::from(error_code))
+                }
```
Freax13, line 52:
```suggestion
                if ret > (u64::MAX - u64::from(i32::MAX)) {
                    // Convert to negative error code
                    let error_code = ((ret as i64) - (1 << 32)) as i32;
                    return Err(SysCallError::from(error_code))
                }
```

The previous implementation classified `0x2_0000_0000` as an error, but that's not correct. 
We could also make the range even smaller (e.g. IIRC on Linux only everything from `0xffff_ffff_ffff_fe00` to `0xffff_ffff_ffff_ffff` is used for errors).

syscall/src/class0.rs
```diff
@@ -0,0 +1,34 @@
...
+//
+// Author: Chuanxiao Dong <chuanxiao.dong@intel.com>
+
+use super::call::{syscall1, syscall3, SysCallError};
+use super::{SYS_EXEC, SYS_EXIT};
+use core::ffi::CStr;
+
+pub fn exit(code: u32) -> ! {
+    // # Safety
+    // SYS_EXIT is supported syscall number by the svsm kernel.
```
Freax13, line 13:
```suggestion
    // SAFETY: SYS_EXIT is supported syscall number by the svsm kernel.
```

syscall/src/class0.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Debug)]
+pub struct Tid(u32);
+
+pub fn exec(file: &CStr, root: &CStr, flags: u32) -> Result<Tid, SysCallError> {
+    unsafe {
+        syscall3(
+            SYS_EXEC,
+            file.as_ptr() as u64,
+            root.as_ptr() as u64,
+            flags as u64,
```
Freax13, line 30:
```suggestion
            u64::from(flags),
```

syscall/src/class1.rs
```diff
@@ -0,0 +1,38 @@
...
+impl Obj for FsObjHandle {
+    fn id(&self) -> u32 {
+        u32::from(&self.0)
+    }
+}
+
+pub fn opendir(path: &CStr) -> Result<FsObjHandle, SysCallError> {
+    unsafe {
+        syscall1(SYS_OPENDIR, path.as_ptr() as u64)
+            .map(|ret| Ok(FsObjHandle(ObjHandle::new(ret as u32))))?
```
Freax13, line 24:
```suggestion
            .map(|ret| FsObjHandle(ObjHandle::new(ret as u32)))
```

syscall/src/def.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Maximum length of file name in bytes
+pub const F_NAME_SIZE: usize = 256;
+
+/// Files
+pub const F_TYPE_FILE: u8 = 0;
+
+/// Directories
+pub const F_TYPE_DIR: u8 = 1;
```
Freax13, line 37: Can we make this an enum?

init/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[no_mangle]
+pub extern "C" fn init_start() -> ! {
+    unsafe {
+        HEAP_ALLOCATOR
+            .lock()
+            .init(core::ptr::addr_of!(HEAP) as usize, HEAP_SIZE);
+    }
+
+    let bin = CString::new("/bin/").expect("Failed to create new string");
```
Freax13, line 32:
```suggestion
    let bin = CString::new("/bin/").unwrap();
```

See https://github.com/coconut-svsm/svsm/pull/467#discussion_r1775205323

init/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    let obj = match opendir(&bin) {
+        Ok(obj) => obj,
+        _ => exit(0),
+    };
```
Freax13, line 36:
```suggestion
    let Ok(obj) = opendir(&bin) else {
        exit(0);
    };
```

init/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    'outer: loop {
+        let n = readdir(&obj, &mut dirents).unwrap();
+        for d in dirents.iter().take(n) {
+            if d.file_type == F_TYPE_FILE {
+                binfile = CString::from(CStr::from_bytes_until_nul(&d.file_name).unwrap());
+                break 'outer;
+            }
+        }
+        if n < dirents.len() - 1 {
```
Freax13, line 48:
```suggestion
        if n < dirents.len() {
```

init/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+            break;
+        }
+    }
+
+    if binfile.is_empty() {
+        exit(0);
+    }
```
Freax13, line 55: Please use `Option` instead of checking for `is_empty`.

kernel/src/mm/guestmem.rs
```diff
@@ -195,6 +200,11 @@ impl<T: Copy> GuestPtr<T> {
         }
     }
 
+    #[inline]
+    pub fn as_ptr(&self) -> *const T {
+        self.ptr as *const T
+    }
+
```
Freax13, line 207: We don't need this anymore, do we?

kernel/src/syscall/class1.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub fn sys_opendir(path: usize) -> Result<u64, i32> {
+    let user_path_ptr = UserPtr::<c_char>::new(VirtAddr::from(path)).map_err(|_| EINVAL)?;
+    let mut path_buffer = Vec::new();
+    unsafe {
+        let path = user_path_ptr
+            .read_c_string(&mut path_buffer)
+            .map_err(|_| EINVAL)?;
+        let dir = find_dir(current_task().get_rootdir(), path).map_err(|e| match e {
+            SvsmError::FileSystem(FsError::FileNotFound) => ENOTFOUND,
```
Freax13, line 28: We should implement `From<SvsmError> for SysCallError` and just use that everywhere. This would make error handling much less verbose.

kernel/src/task/tasks.rs
```diff
@@ -267,6 +278,10 @@ impl Task {
         self.id
     }
 
+    pub fn get_rootdir(&self) -> Arc<dyn Directory> {
```
Freax13, line 281:
```suggestion
    pub fn rootdir(&self) -> Arc<dyn Directory> {
```

In Rust, [getters don't have a `get_` prefix](https://rust-lang.github.io/api-guidelines/naming.html?highlight=gett#getter-names-follow-rust-convention-c-getter).

kernel/src/task/exec.rs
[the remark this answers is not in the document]
Freax13: > Yes, you are correct, the page tables for the new task aren't active and so the error. Few options to resolve the issue,
> 
>   1. Switch the page tables to the new task, zeroed the memory and switched back to the kernel task page tables. But this doesn't seem like a clean solution.
> 
>   2. mmap the pages as anonymous as part of the kernel task itself, clear the range `[p_filesz..p_memsz)` and update the new task's VMR range to point to this physical memory ensuring that ref count is not decremented when the mapping is dropped. (No implementation exists currently; we may need to implement this)
> 
>   3. Customize the linker script in such a way that `.bss` section has a separate segment. This way we can avoid clearing `[p_filesz..p_memsz)` as we can simply mmap this entire region as an anonymous region.
> 
> 
> For this PR I have reverted to approach 3, can you please share your thoughts on how we should go about this?

In the long run we should implement this properly, but I'm fine with going with approach 3 for now.

IMO we should implement this in `VMFileMapping::new`. If the size isn't a multiple of the page size, instead of failing, we should make a private copy of the page using `.try_cop_page()` and zero out the bytes after `size`. Each `PageRef` is already mapped into the kernel page tables anyways, so we don't need to create new mappings.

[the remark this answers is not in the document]
vijaydhanraj: Thanks for the suggestion. If it is fine with you, I can help with implementing this fix after this PR is merged.
I have created a github issue https://github.com/coconut-svsm/svsm/issues/483 to capture this as well the ELF parser issue and removing the linker script WA.

syscall/src/call.rs
Freax13: Looks good to me.

syscall/src/def.rs
Freax13: I like your second approach. We wouldn't need to cast in very many places, would we?

init/src/main.rs
```diff
@@ -0,0 +1,71 @@
...
+                binfile = Some(CString::from(
+                    CStr::from_bytes_until_nul(&d.file_name).unwrap(),
+                ));
+                break 'outer;
+            }
+        }
+        if n < dirents.len() {
+            break;
+        }
+    }
```
Freax13, line 52:
```suggestion
    loop {
        let n = readdir(&obj, &mut dirents).unwrap();
        if let Some(d) = dirents.iter().take(n).find(|d| d.file_type == F_TYPE_FILE) {
            binfile = Some(CString::from(
                CStr::from_bytes_until_nul(&d.file_name).unwrap(),
            ));
            break;
        }
        if n < dirents.len() {
            break;
        }
    }
```

kernel/src/cpu/idt/svsm.rs
```diff
@@ -235,10 +235,16 @@ extern "C" fn ex_handler_system_call(ctxt: &mut X86ExceptionContext) {
...
-    };
+        SYS_EXIT => sys_exit(ctxt.regs.rdi as u32),
+        SYS_EXEC => sys_exec(ctxt.regs.rdi, ctxt.regs.rsi, ctxt.regs.r8),
+        SYS_CLOSE => sys_close(ctxt.regs.rdi as u32),
+        SYS_OPENDIR => sys_opendir(ctxt.regs.rdi),
+        SYS_READDIR => sys_readdir(ctxt.regs.rdi as u32, ctxt.regs.rsi, ctxt.regs.r8),
+        _ => Err(SysCallError::EINVAL),
+    }
+    // Intermediate casting to u32 prevents overflow and undefined behavior
+    // when casting the usize back to i32 in the user space.
```
Freax13, line 246: This comment is outdated.

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +277,98 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+    fn check_bounds(&self) -> bool {
+        let v = VirtAddr::from(self.guest_ptr.ptr);
+        if !(USER_MEM_START..USER_MEM_END).contains(&v)
+            || !(USER_MEM_START..USER_MEM_END).contains(&(v + size_of::<T>()))
+        {
+            return false;
+        }
+        true
```
Freax13, line 300:
```suggestion
        (USER_MEM_START..USER_MEM_END).contains(&v)
            && (USER_MEM_START..USER_MEM_END).contains(&(v + size_of::<T>()))
```

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +277,98 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
     }
 }
 
+#[derive(Debug)]
+pub struct UserPtr<T: Copy + FromBytes> {
```
Freax13, line 281: Can we move the `FromBytes` bound to the `read` method? I don't see a reason why all types must implement `FromBytes`. This would also solve the problems you mentioned in https://github.com/coconut-svsm/svsm/pull/467#discussion_r1797504701, wouldn't it?

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +277,98 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+        }
+        true
+    }
+
+    /// # Safety
+    ///
+    /// This function ensures that the address is valid for reading. If the address is not valid,
+    /// it returns an `Err(SvsmError::InvalidAddress)`.
+    #[inline]
+    pub unsafe fn read(&self) -> Result<T, SvsmError> {
```
Freax13, line 308:
```suggestion
    #[inline]
    pub fn read(&self) -> Result<T, SvsmError> {
```

This function (and `read_c_string`, `write`, and `write_ref`) don't need to be unsafe. We have all the checks in place to make sure that nothing bad can happen :)

kernel/src/mm/guestmem.rs
Freax13: Hm, yeah, we probably shouldn't allocate a whole page on the stack. Still, can we enforce a size limit here?

kernel/src/syscall/class0.rs
```diff
@@ -0,0 +1,50 @@
...
+use core::ffi::c_char;
+use syscall::SysCallError;
+
+pub fn sys_exit(exit_code: u32) -> ! {
+    log::info!("Terminating current task, exit_code {exit_code}");
+    unsafe {
+        current_task_terminated();
+    }
+    schedule();
+    panic!("schedule() returned in sys_exit()");
```
Freax13, line 24:
```suggestion
    unreachable!("schedule() returned in sys_exit()");
```

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +277,98 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+        }
+    }
+}
+
+impl UserPtr<c_char> {
+    /// # Safety
+    ///
+    /// Reads a null-terminated C string from the user space into the provided buffer.
+    /// Returns a `&str` reference to the buffer.
+    pub unsafe fn read_c_string<'a>(&self, buffer: &'a mut Vec<u8>) -> Result<&'a str, SvsmError> {
```
Freax13, line 359: Given that we don't seem to be reusing `buffer` anywhere, can we just allocate the memory inside this function and return a `String` instead?

kernel/src/syscall/class1.rs
```diff
@@ -0,0 +1,70 @@
...
+pub fn sys_readdir(obj_id: u32, dirents: usize, size: usize) -> Result<u64, SysCallError> {
+    let fsobj = obj_get(obj_id.into())?;
+    let fsobj = fsobj.as_fs().ok_or(ENOTSUPP)?;
+    let user_dirents_ptr = UserPtr::<DirEnt>::new(VirtAddr::from(dirents));
+
+    for i in 0..size {
+        let (name, dirent) = match fsobj.readdir()? {
+            Some((name, dirent)) => (name, dirent),
+            None => return Ok(i as u64),
+        };
```
Freax13, line 41:
```suggestion
        let Some((name, dirent)) = fsobj.readdir()? else {
            return Ok(i as u64);
        };
```

kernel/src/syscall/class1.rs
```diff
@@ -0,0 +1,70 @@
...
+            if n == last_n || n == fname.len() {
+                *c = b'\0';
+                break;
+            } else {
+                *c = fname[n];
+                if *c == b'\0' {
+                    break;
+                }
+            }
+        }
```
Freax13, line 56:
```suggestion
        new_entry.file_name[..MAX_FILENAME_LENGTH].copy_from_slice(&fname);
```

syscall/src/call.rs
```diff
@@ -0,0 +1,95 @@
...
+//
+// Copyright (c) 2024 Intel Corporation.
+//
+// Author: Chuanxiao Dong <chuanxiao.dong@intel.com>
+
+use super::*;
+use core::arch::asm;
+
+/// Macro to generate system call functions with varying numbers of arguments.
+#[macro_export]
```
Freax13, line 11:
```suggestion
```

kernel/src/mm/guestmem.rs
vijaydhanraj: Yes, agree with your suggestion. This seems to be a better solution. But still allocating from the heap instead of statically allocating due to unknown `path` length.

kernel/src/syscall/class1.rs
vijaydhanraj: Done.

kernel/src/syscall/class1.rs
vijaydhanraj: Much better, thanks!

syscall/src/def.rs
vijaydhanraj: This is no longer an issue, and I have made this to an `enum`.

init/src/main.rs
vijaydhanraj: Done.

kernel/src/mm/guestmem.rs
Freax13: I'd be fine with 4K.

Freax13:
FYI this branch has conflicts with main. You might have to rebase your branch to resolve them.

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +278,86 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+        }
+        unsafe { self.guest_ptr.read() }
+    }
+
+    #[inline]
+    pub fn write(&self, buf: T) -> Result<(), SvsmError> {
+        if !self.check_bounds() {
+            return Err(SvsmError::InvalidAddress);
+        }
+        unsafe { self.guest_ptr.write(buf) }
```
Freax13, line 317:
```suggestion
        self.write_ref(&buf)
```

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +278,86 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+    #[inline]
+    pub fn write_ref(&self, buf: &T) -> Result<(), SvsmError> {
+        if !self.check_bounds() {
+            return Err(SvsmError::InvalidAddress);
+        }
+        unsafe { self.guest_ptr.write_ref(buf) }
+    }
+
+    #[inline]
+    pub const fn cast<N: Copy + FromBytes>(&self) -> UserPtr<N> {
```
Freax13, line 329:
```suggestion
    pub const fn cast<N: Copy>(&self) -> UserPtr<N> {
```

kernel/src/syscall/class1.rs
```diff
@@ -0,0 +1,54 @@
...
+    for i in 0..size {
+        let Some((name, dirent)) = fsobj.readdir()? else {
+            return Ok(i as u64);
+        };
+
+        let mut new_entry = DirEnt::default();
+        let fname = FileNameArray::from(name);
+        new_entry.file_name[..fname.len()].copy_from_slice(&fname);
+
+        if let DirEntry::File(ref f) = dirent {
```
Freax13, line 42:
```suggestion
        if let DirEntry::File(f) = dirent {
```

We don't need this here.

syscall/src/def.rs
```diff
@@ -0,0 +1,49 @@
...
+pub const SYS_READDIR: u64 = CLASS1 + 5;
+
+/// Maximum length of file name in bytes
+pub const F_NAME_SIZE: usize = 256;
+
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum FileType {
+    File = 0,
+    Directory = 1,
```
Freax13, line 27:
```suggestion
    File,
    Directory,
```

kernel/src/mm/guestmem.rs
vijaydhanraj: Can you please confirm if this looks fine?
```
diff --git a/kernel/src/mm/guestmem.rs b/kernel/src/mm/guestmem.rs
index c2b0154c..440ee5e9 100644
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -344,7 +344,7 @@ impl UserPtr<c_char> {
     /// Reads a null-terminated C string from the user space.
     /// Allocates memory for the string and returns a `String`.
     pub fn read_c_string(&self) -> Result<String, SvsmError> {
-        let mut buffer: Vec<u8> = Vec::new();
+        let mut buffer = Vec::with_capacity(4096);
 
         for offset in 0.. {
             let current_ptr = self.offset(offset);
```

Freax13: No, this won't work. `Vec::with_capacity` creates a vector with the given capacity, but it doesn't prevent the vector from growing past this capacity. Instead we can add a check to the for loop body to see if `offset` is greater than 4k.

vijaydhanraj: Thanks, done.

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +278,86 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+            return Err(SvsmError::InvalidAddress);
+        }
+        unsafe { self.guest_ptr.read() }
+    }
+
+    #[inline]
+    pub fn write(&self, buf: T) -> Result<(), SvsmError> {
+        if !self.check_bounds() {
+            return Err(SvsmError::InvalidAddress);
+        }
```
Freax13, line 316:
```suggestion
```

kernel/src/mm/guestmem.rs
```diff
@@ -271,6 +278,86 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+    }
+}
+
+impl UserPtr<c_char> {
+    /// Reads a null-terminated C string from the user space.
+    /// Allocates memory for the string and returns a `String`.
+    pub fn read_c_string(&self) -> Result<String, SvsmError> {
+        let mut buffer = Vec::with_capacity(4096);
+
+        for offset in 0..buffer.capacity() {
```
Freax13, line 349: We should return an error if the string exceeds the capacity. Also, we probably shouldn't default to using 4k as the capacity for the vector. Most strings are probably way smaller than that, the default allocation strategy of vectors should work fine.

vijaydhanraj: I see your point, how about this? Also, introduced a new const for max path len.
```
@@ -344,17 +342,17 @@ impl UserPtr<c_char> {
     /// Reads a null-terminated C string from the user space.
     /// Allocates memory for the string and returns a `String`.
     pub fn read_c_string(&self) -> Result<String, SvsmError> {
-        let mut buffer = Vec::with_capacity(4096);
+        let mut buffer = Vec::new();
 
-        for offset in 0..buffer.capacity() {
+        for offset in 0..PATH_MAX {
             let current_ptr = self.offset(offset as isize);
             let char_result = current_ptr.read()?;
             match char_result {
-                0 => break,
+                0 => return String::from_utf8(buffer).map_err(|_| SvsmError::InvalidUtf8),
                 c => buffer.push(c as u8),
             }
         }
-        String::from_utf8(buffer).map_err(|_| SvsmError::InvalidUtf8)
+        Err(SvsmError::InvalidBytes)
     }
 }
```

Freax13: Looks good to me.

kernel/src/mm/guestmem.rs
vijaydhanraj: Removed the redundant check.

kernel/src/mm/guestmem.rs
vijaydhanraj: Thanks, done.

Freax13 (approved):
Looks good to me, thanks!

joergroedel:
Thanks for that work, it is pretty impressive! Although it is hard to review on all levels due to its size:

```
 .github/workflows/rust.yml     |  2 +-
 Cargo.lock                     | 52 ++++++++++++++++++++++++++++++++++++++++++++++++++++
 Cargo.toml                     |  5 +++++
 Makefile                       | 16 ++++++++++++++--
 dm/Cargo.toml                  | 15 +++++++++++++++
 dm/build.rs                    |  6 ++++++
 dm/dm.lds                      | 37 +++++++++++++++++++++++++++++++++++++
 dm/src/main.rs                 | 25 +++++++++++++++++++++++++
 elf/src/program_header.rs      |  3 +++
 init/Cargo.toml                | 16 ++++++++++++++++
 init/build.rs                  |  6 ++++++
 init/init.lds                  | 37 +++++++++++++++++++++++++++++++++++++
 init/src/main.rs               | 73 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 kernel/src/cpu/idt/svsm.rs     | 12 ++++++++----
 kernel/src/error.rs            | 27 +++++++++++++++++++++++++++
 kernel/src/fs/api.rs           |  1 +
 kernel/src/fs/filesystem.rs    | 60 +++++++++++++++++++++++++++++++++++++++++++++++++-----------
 kernel/src/fs/mod.rs           |  2 ++
 kernel/src/fs/obj.rs           | 72 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 kernel/src/mm/guestmem.rs      | 85 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 kernel/src/string.rs           | 11 +++++++++++
 kernel/src/svsm.rs             | 13 +++++++------
 kernel/src/syscall/class0.rs   | 42 ++++++++++++++++++++++++++++++++++++++++++
 kernel/src/syscall/class1.rs   | 54 ++++++++++++++++++++++++++++++++++++++++++++++++++++++
 kernel/src/syscall/handlers.rs | 21 ---------------------
 kernel/src/syscall/mod.rs      |  6 ++++--
 kernel/src/syscall/obj.rs      | 13 ++++++++++++-
 kernel/src/task/exec.rs        | 17 ++++++++++++-----
 kernel/src/task/schedule.rs    |  8 ++++++--
 kernel/src/task/tasks.rs       | 19 +++++++++++++++++--
 syscall/Cargo.toml             |  1 +
 syscall/src/call.rs            | 93 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 syscall/src/class0.rs          | 33 +++++++++++++++++++++++++++++++++
 syscall/src/class1.rs          | 38 ++++++++++++++++++++++++++++++++++++++
 syscall/src/def.rs             | 52 ++++++++++++++++++++++++++++++++++++++++++++++++++++
 syscall/src/lib.rs             | 10 ++++++++--
 syscall/src/numbers.rs         | 10 ----------
 syscall/src/obj.rs             | 15 +++++++++++++++
 38 files changed, 939 insertions(+), 69 deletions(-)
```

Can you please split this into smaller PRs which we can merge one at a time? Please start with the system call interface and implementation. This PR can be rebased then on what has already been merged.

vijaydhanraj:
Sure, I can split this into smaller PRs. Would splitting commits into the below PRs be helpful?

- **PR#8:** Introduce a simple template for initial device model
    - dm: Add initial device model project
    - init: Extend opendir() and readdir() for dm discovery

- **PR#7**: Add the initial version of init

- **PR#6**: Introduce exec syscall
    - syscall: Implement exec()
    - kernel/syscall: Implement EXEC
    - kernel/task: Return TID in exec_user()

- **PR#5**: Introduce opendir and readdir syscalls
    - syscall/class1: Implement opendir() and readdir()
    - kernel/syscall: Implement OPENDIR and READDIR
    - kernel/fs/filesystem: Add opendir()
    - kernel/fs: Add FsObj
    - kernel/fs/filesystem: Add find_dir()
    - svsm: Initialize FS earlier

- **PR#4**: Introduce exit and open syscalls
    - syscall/obj: Close an ObjHandle when dropped
    - kernel/syscall: Implement close syscall
    - syscall: Introduce exit syscall


- **PR#3**: Introduce syscall interface and error code
    - syscall: Define error code
    - syscall: Add syscall macro

- **PR#2**: Introduce UserPtr wrapper around GuestPtr

- **PR#1**: Few bookkeeping and clean-up commits 
    - kernel/cpu/idt: Remove SYS_HELLO for implementing formal syscalls
    - kernel/syscall: Rename handlers.rs to class0.rs
    - syscall: Rename numbers.rs to def.rs
    - syscall: Add Obj trait

Freax13:
FYI, we merged SMAP support in #473, this PR will have to be adjusted accordingly.

vijaydhanraj:
Hi @joergroedel as suggested, split this PR and added a PR https://github.com/coconut-svsm/svsm/pull/502 to introduce the syscall interface implementation.

vijaydhanraj:
> FYI, we merged SMAP support in #473, this PR will have to be adjusted accordingly.

Hi @Freax13 thanks, I have rebased this PR.

Freax13:
> Hi @Freax13 thanks, I have rebased this PR.

I meant to say that we now need to execute `stac` & `clac` in the `UserPtr` methods.

vijaydhanraj:
Got it, will add the `stac` and `clac` as part of the `read/write_ref` `UserPtr` methods and update the PR.

vijaydhanraj:
@Freax13 I have updated the PR to use `stac` & `clac` in `UserPtr` methods. If the changes look fine, please let me know if this should be spined out as a separate PR?

Freax13:
> @Freax13 I have updated the PR to use `stac` & `clac` in `UserPtr` methods. If the changes look fine, please let me know if this should be spined out as a separate PR?

Personally, I don't have a problem with large PRs, but @joergroedel might object.

From your comment above:
>   * **PR#2**: Introduce UserPtr wrapper around GuestPtr

It probably makes sense to add the changes to this PR.

kernel/src/mm/guestmem.rs
```diff
@@ -269,6 +278,98 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+
+    #[inline]
+    pub fn read(&self) -> Result<T, SvsmError>
+    where
+        T: FromBytes,
+    {
+        let _guard = user_access_begin();
+        if !self.check_bounds() {
+            return Err(SvsmError::InvalidAddress);
+        }
```
Freax13, line 322:
```suggestion
        if !self.check_bounds() {
            return Err(SvsmError::InvalidAddress);
        }
        let _guard = user_access_begin();
```

The same change should be applied for `write_ref`.

kernel/src/mm/guestmem.rs
```diff
@@ -269,6 +278,98 @@ impl<T: Copy> InsnMachineMem for GuestPtr<T> {
...
+impl Drop for UserAccessGuard {
+    fn drop(&mut self) {
+        clac();
+    }
+}
+
+fn user_access_begin() -> UserAccessGuard {
+    stac();
+    UserAccessGuard
+}
```
Freax13, line 292:
```suggestion
struct UserAccessGuard;

impl UserAccessGuard {
    pub fn new() -> Self {
        stac();
        Self
    }
}

impl Drop for UserAccessGuard {
    fn drop(&mut self) {
        clac();
    }
}
```

vijaydhanraj: Done.

kernel/src/mm/guestmem.rs
vijaydhanraj: Since we switched to `new` method, used it here rather than calling the function.

vijaydhanraj:
This PR can be closed once https://github.com/coconut-svsm/svsm/pull/547 is merged. Most of the changes from this PR are already merged, with PR #547 being the last one.

peterfang:
Closing as the foundational user mode work is done. https://github.com/coconut-svsm/svsm/pull/547 will be looked at separately.

PR #473, opened 2024-09-30
Enabling SMEP and SMAP

p4zuu:
Enabling SMEP and SMAP if supported by the CPU.

To enable these, `smep` and `smap` features have to be specifically enabled, but this could be changed in the future since I think all modern CPUs support SMEP and SMAP.

I stole the way to give the CFG_SMAP bool to assembly from #455. This will likely cause a small conflict in  `kernel/src/cpu/idt/svsm.rs:265` when one of the two PR is merged.
This also conflict with the userspace PR since SMAP has to be temporary disabled to fetch user-provided pointers in syscalls arguments.
 
Finally, SMAP requires RFLAGS.AC to be unset to be enabled so I added a `clac` instruction in different entries, including #HV entry for which I don't really know if this is required, and if this could break something in the future. @msft-jlange could you please take a look?

### DIFF
--- a/kernel/src/cpu/idt/common.rs
+++ b/kernel/src/cpu/idt/common.rs
@@ -45,8 +45,6 @@ pub const HV_VECTOR: usize = 28;
 pub const VC_VECTOR: usize = 29;
 pub const SX_VECTOR: usize = 30;
 
-pub const PF_ERROR_WRITE: usize = 2;
-
 pub const INT_INJ_VECTOR: usize = 0x50;
 
 bitflags::bitflags! {
--- a/kernel/src/cpu/x86/smap.S
+++ b/kernel/src/cpu/x86/smap.S
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2024 SUSE LLC
+//
+// Authors: Thomas Leroy <tleroy@suse.de>
+
+.code64
+
+.section .text
+.macro asm_clac
+        .if !CFG_NOSMAP
+        clac
+        .endif
+.endm
+
+.macro asm_stac
+        .if !CFG_NOSMAP
+        stac
+        .endif
+.endm
--- a/kernel/src/cpu/x86/smap.rs
+++ b/kernel/src/cpu/x86/smap.rs
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2024 SUSE LLC
+//
+// Authors: Thomas Leroy <tleroy@suse.de>
+
+use core::arch::asm;
+
+/// Clears RFLAGS.AC to enable SMAP.
+/// This is currently only used when SMAP is supported and enabled.
+/// SMAP protection is effective only if CR4.SMAP is set and if RFLAGS.AC = 0.
+#[inline(always)]
+pub fn clac() {
+    if !cfg!(feature = "nosmap") {
+        unsafe { asm!("clac", options(att_syntax, nomem, nostack, preserves_flags)) }
+    }
+}
+
+/// Sets RFLAGS.AC to disable SMAP.
+/// This is currently only used when SMAP is supported and enabled.
+/// SMAP protection is effective only if CR4.SMAP is set and if RFLAGS.AC = 0.
+#[inline(always)]
+pub fn stac() {
+    if !cfg!(feature = "nosmap") {
+        unsafe { asm!("stac", options(att_syntax, nomem, nostack, preserves_flags)) }
+    }
+}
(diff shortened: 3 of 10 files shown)


msft-jlange:
> Finally, SMAP requires RFLAGS.AC to be unset to be enabled so I added a `clac` instruction in different entries, including #HV entry for which I don't really know if this is required, and if this could break something in the future. @msft-jlange could you please take a look?

CLAC should be cleared in every exception handler so that any event handler that is invoked while executing in kernel mode can transition to EFLAGS.AC=0 while it is executing.  The #HV handler is no exception.  Every event handler that wants to access user-mode code should explicitly be required to execute STAC when user-mode access is expected, and the #HV handler is no different.  You have put CLAC in the right place.

msft-jlange:
Why are the `smap` and `smep` features optional?  Is there a reason we wouldn't want them enabled for all builds?  I worry that making them optional is going to lead to regressions pretty quickly as most testing will occur with SMAP disabled, and that will mask important bugs that are only observed with SMAP enabled.

p4zuu:
> CLAC should be cleared in every exception handler so that any event handler that is invoked while executing in kernel mode can transition to EFLAGS.AC=0 while it is executing. The #HV handler is no exception. Every event handler that wants to access user-mode code should explicitly be required to execute STAC when user-mode access is expected, and the #HV handler is no different. You have put CLAC in the right place.

Perfect, thanks!

kernel/src/cpu/x86/smap.S
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (c) 2024 SUSE LLC
+//
+// Authors: Thomas Leroy <tleroy@suse.de>
+
+.code64
+
+.section .text
+.macro asm_clac
+        .if CFG_SMAP
+        .byte 0x0f, 0x01, 0xca
```
Freax13, line 12: Here and elsewhere:
```suggestion
        clac
```

Is there a reason we need to use directives?

p4zuu:
> Why are the `smap` and `smep` features optional? Is there a reason we wouldn't want them enabled for all builds? I worry that making them optional is going to lead to regressions pretty quickly as most testing will occur with SMAP disabled, and that will mask important bugs that are only observed with SMAP enabled.

I was more looking for feedback here. I had the feeling that it was worth enabling smap and smep by default since all modern CPUs have it, but was a bit unsure. The argument of testing without them rings a bell now and very makes sense to me. I'm switching to enable them by default.

kernel/src/cpu/x86/smap.S
Freax13: My guess is that we did that in `do_tlbsync` because of concerns that the compiler doesn't support the `tlbsync` instruction. The compiler definitively supports `stac`/`clac`, so we should just use that.

kernel/src/cpu/x86/smap.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use core::arch::asm;
+
+/// Clears RFLAGS.AC to enable SMAP.
+/// This is currently only used when SMAP is supported and enabled.
+/// SMAP protection is effective only if CR4.SMAP is set and if RFLAGS.AC = 0.
+#[inline(always)]
+pub fn clac() {
+    if !cfg!(feature = "nosmap") {
+        unsafe { asm!("clac", options(att_syntax, nomem, nostack)) }
```
Freax13, line 15:
```suggestion
        unsafe { asm!("clac", options(att_syntax, nomem, nostack, preserves_flags)) }
```

[Inline assembly](https://doc.rust-lang.org/reference/inline-assembly.html) may touch the `AC` flag if the `preserves_flags` is set.

Freax13 (approved):

kernel/src/cpu/x86/smap.rs
p4zuu: I did paid attention to this but I guess I misunderstood the doc:
> The asm! block does not modify the flags register (defined in the rules below). This allows the compiler to avoid recomputing the condition flags after the asm! block.

Isn't it relevant if the code inside the `asm!()` doesn't modify the flags? Which is obviously not the case for `clac/stac`

Freax13: > Maybe it's actually because `stac`/`clac` changes the AC flag, which is not included in the `preserves_flags` rule: CF, PF, AF, ZF, SF, OF

It's this. The compiler doesn't care about the `AC` flag.

kernel/src/cpu/idt/common.rs
```diff
@@ -45,7 +47,16 @@ pub const HV_VECTOR: usize = 28;
...
+bitflags! {
+    #[repr(transparent)]
+    #[derive(Copy, Clone, Debug, PartialEq)]
+    pub struct PFErrors: usize {
+        const PROT = 1 << 0;
+        const WRITE = 1 << 1;
+        const USER = 1 << 2;
+        const INSTR = 1 << 4;
+    }
+}
```
cxdong, line 60: It appears that `PFError` has the similar error flags with the `PageFaultError` within this file. It might be beneficial to consider merging the two for consistency and to reduce redundancy?

Freax13 (approved):

PR #480, opened 2024-10-10
cpuarch/vmsa: fix soundness issue with GuestVMExit

p4zuu:
Previously, it was possible for a guest to exit with an exit code undefined in the GuestVMExit enum, leading to undefined behavior.

As suggested by @Freax13 in #359, we can replace this by a tuple struct, allowing unknown values.

### DIFF
--- a/cpuarch/src/vmsa.rs
+++ b/cpuarch/src/vmsa.rs
@@ -9,44 +9,51 @@
 use bitfield_struct::bitfield;
 use zerocopy::FromZeros;
 
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, FromZeros, PartialEq)]
+#[allow(dead_code)]
+pub struct GuestVMExit(u64);
+
+impl Default for GuestVMExit {
+    #[inline]
+    fn default() -> Self {
+        Self::INVALID
+    }
+}
+
 // AE Exitcodes
 // Table 15-35, AMD64 Architecture Programmer’s Manual, Vol. 2
-#[repr(u64)]
-#[derive(Clone, Copy, Default, Debug, FromZeros)]
-#[allow(dead_code)]
-pub enum GuestVMExit {
-    CR0_READ = 0,
-    MC = 0x52,
-    INTR = 0x60,
-    NMI = 0x61,
-    SMI = 0x62,
-    INIT = 0x63,
-    VINTR = 0x64,
-    PAUSE = 0x77,
-    HLT = 0x78,
-    SHUTDOWN = 0x7F,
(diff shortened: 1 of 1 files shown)


cpuarch/src/vmsa.rs
```diff
@@ -6,43 +6,64 @@
 
 use bitfield_struct::bitfield;
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[allow(dead_code, non_camel_case_types)]
```
Freax13, line 10:
```suggestion
#[allow(dead_code)]
```

cpuarch/src/vmsa.rs
```diff
@@ -6,43 +6,64 @@
...
+            ($exit:ident, $value:expr);
+        )+
+    ) => {
+        impl GuestVMExit {
+        $(
+            pub const $exit: Self = Self($value);
+        )+
+        }
+    };
+}
```
Freax13, line 25: I'm not sure if we need to use a macro for this. I get that the macro makes the code a bit less verbose, but it also makes it less obvious that there is nothing else going on.

I could understand the need for the macro if we also used it to generate a prettier `Debug` implementation.

Freax13: Yes, personally, I'd prefer that.

p4zuu: Let's switch then. That makes sense that we can avoid a macro if there is no need to generate specific `Debug` or even specific documentation

cpuarch/src/vmsa.rs
p4zuu: Done

cpuarch/src/vmsa.rs
p4zuu: Done

cpuarch/src/vmsa.rs
```diff
@@ -6,43 +6,50 @@
 
 use bitfield_struct::bitfield;
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[allow(dead_code)]
+pub struct GuestVMExit(u64);
```
Freax13, line 11:
```suggestion
#[repr(transparent)]
pub struct GuestVMExit(u64);
```

Freax13 (approved):

joergroedel:
Hey @p4zuu , can you please rebase this change to latest HEAD? Thanks!

p4zuu:
> Hey @p4zuu , can you please rebase this change to latest HEAD? Thanks!

Done :)

PR #491, opened 2024-10-22
kernel/idt: Ensure #HV handler RIP checks are executed when interrupts are enabled

roy-hopkins:
When testing restricted injection using an APIC timer firing into VMPL0 I experienced errors reported by the GHCB.rs in the code that detects whether the switch to VMPL2 was successful. This was down to a couple of separate issues. One of which is fixed by this PR. 

The current #HV handler checks EFLAGS.IF=1 at the start of the handler and immediately processes the #HV events if enabled without performing checks to see if RIP is within the VMPL switch or the iret window. The RIP windows are only checked if EFLAGS.IF=0. This means that if interrupts are enabled during either of these windows and a #HV occurs then the behaviour is undefined.

### DIFF
--- a/kernel/src/cpu/idt/entry.S
+++ b/kernel/src/cpu/idt/entry.S
@@ -84,11 +84,7 @@ asm_entry_hv:
 	pushq	$0
 	pushq	%rax
 	pushq 	%rbx
-        pushq   %rcx
-	// Check whether interrupts were enabled at the time of #HV.  If so,
-	// commit to processing all #HV events immediately.
-	testl 	$0x200, 0x30(%rsp)
-	jnz	continue_hv
+	pushq   %rcx
 	// Check whether the trap RIP is within the guest VMPL return window.
 	movq	0x20(%rsp), %rax // fetch RIP from the trap frame.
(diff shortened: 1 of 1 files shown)


msft-jlange:
> The current #HV handler checks EFLAGS.IF=1 at the start of the handler and immediately processes the #HV events if enabled without performing checks to see if RIP is within the VMPL switch or the iret window. The RIP windows are only checked if EFLAGS.IF=0. This means that if interrupts are enabled during either of these windows and a #HV occurs then the behaviour is undefined.

Roy, the current behavior is explicitly by design, since those code paths are always executed with interrupts disabled, and thus there should be no point in RIP checks when interrupts are enabled.  Can you explain what problem you're trying to fix?

roy-hopkins:
So the problem I am trying to fix is to allow interrupts occurring for VMPL0 to preempt a running VMPL2. I think I've misunderstood how this was supposed to work - with the 'aborted' switch to VMPL2 from the #HV handler being used to transfer control back to VMPL0. I'm still trying to understand how the new (now merged) PR #492 works in this situation but agree this PR is not the correct answer so I'll close it.

PR #492, opened 2024-10-22
requests: disable interrupts when reentering guest VMPL

msft-jlange:
When reentering the guest VMPL, no interrupts can be permitted once the SVSM begins reevaluating state that is presented to the guest OS upon reentry (such as guest APIC state), since handling an interrupt in this window may alter and corrupt the guest state being prepared.

Note that it is no longer necessary to explicitly check for #HV doorbell events in this flow because the use of #HV doorbell, which necessarily implies the use of Restricted Injection, causes the SVSM to run generally with interrupts enabled, and thus #HV doorbell events will naturally be processed any time interrupts are enabled, or when they are explicitly reenabled when the last `IrqGuard` is dropped.


kernel/src/requests.rs
```diff
@@ -173,6 +172,8 @@ pub fn request_loop() {
             flush_tlb_global_sync();
 
             switch_to_vmpl(GUEST_VMPL as u32);
+
+            drop(guard);
```
Freax13, line 176: This line isn't needed, but if your goal is to be explicit about this, I'm also fine with leaving this as is.

Freax13 (approved):

kernel/src/requests.rs
joergroedel: I think it does not hurt to have it explicitly there.

msft-jlange: It's needed to avoid a dead code warning.  Without this, `guard` is never referenced and generates a warning unless it is renamed `_guard`, which implies that the guard is never truly needed.  Including an explicit `drop()` ensures that the guard is known to be essential and that its lifetime scope - and thus the scope of disabled interrupts - is unambiguous.

PR #494, opened 2024-10-23
platform: move AP #HV doorbell initialization to platform

msft-jlange:
Initialization of the #HV doorbell page is a platform-specific operation.  Initialization on the BSP was already in the platform abstraction; initialization on APs should also be in the platform abstraction.

### DIFF
--- a/kernel/src/platform/snp.rs
+++ b/kernel/src/platform/snp.rs
@@ -30,7 +30,9 @@ use crate::utils::MemoryRegion;
 #[cfg(debug_assertions)]
 use crate::mm::virt_to_phys;
 
-use core::sync::atomic::{AtomicU32, Ordering};
+use core::sync::atomic::{AtomicBool, AtomicU32, Ordering};
+
+static SVSM_ENV_INITIALIZED: AtomicBool = AtomicBool::new(false);
 
 static GHCB_IO_DRIVER: GHCBIOPort = GHCBIOPort::new();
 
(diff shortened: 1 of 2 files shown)


kernel/src/platform/snp.rs
```diff
@@ -108,8 +108,15 @@ impl SvsmPlatform for SnpPlatform {
         cpu.setup_ghcb()
     }
 
-    fn setup_percpu_current(&self, cpu: &PerCpu) -> Result<(), SvsmError> {
+    fn setup_percpu_current(&self, cpu: &PerCpu, is_bsp: bool) -> Result<(), SvsmError> {
```
Freax13, line 111: Instead of adding a parameter here, we could also add a method `is_bsp(&self) -> bool` to `PerCpu`.

Freax13 (approved):

kernel/src/platform/snp.rs
msft-jlange: > Instead of adding a parameter here, we could also add a method `is_bsp(&self) -> bool` to `PerCpu`.

We could, but since `PerCpu` does not have that knowledge today, it that would still require the CPU construction code to pass in a parameter, and that would involve code passing `is_bsp` in roughly the same place that it does today.

msft-jlange: > We could, but since `PerCpu` does not have that knowledge today, it that would still require the CPU construction code to pass in a parameter, and that would involve code passing `is_bsp` in roughly the same place that it does today.

To eliminate the whole "is BSP" question, I've changed the code to track whether we've been through SVSM environment initialization.  If so, #HV doorbell initialization is performed during per-CPU setup.  Otherwise, it's completed during SVSM environment initialization.

PR #497, opened 2024-10-24
scripts/launch_guest.sh: fix restoring of terminal settings

stefano-garzarella:
The script runs with `set -e`, so if QEMU fails, we never restore Ctrl-C remapping to Ctrl-].

For example, this happens all the time when we run `scripts/test-in-svsm.sh` because QEMU exits with a value other than 0 (perhaps due to exit via isa-debug-exit).

To solve it, set a handler to re-install them when the script exits, also save the original values to restore exactly the initial configuration.


joergroedel (approved):
LGTM.

PR #503, opened 2024-10-28
various improvements to libmstpm

Freax13:
Among other things, this PR fixes a bug where the build script for libmstpm wouldn't rerun if the submodules in `libmstpm/deps/` were updated. It also changes the build script to only save output files in the directory specified on the `OUT_DIR` environment variable as recommended by [the cargo docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script).


osteffenrh (approved):

cclaudio (approved):
Looks good to me

PR #504, opened 2024-10-28
platform: implement Deref{Mut} for SvsmPlatformCell

Freax13:
This has the advantage that Rust will be able to auto-deref the SvsmPlatformCell into dyn SvsmPlatform, so we don't have to explicitly call as_dyn_ref.


msft-jlange:
Nice!

PR #505, opened 2024-10-29
Introduce UserPtr wrapper around GuestPtr

vijaydhanraj:
As suggested by @Freax13, introduce a `UserPtr` wrapper around `GuestPtr` to ensure 
- No invalid memory access is done.
- Validate the address passed from the userspace to make sure it isn't pointing to kernel memory.

This PR is next in series of PRs to add initial userspace support in SVSM. This is based on PR https://github.com/coconut-svsm/svsm/pull/502

### DIFF
--- a/kernel/src/error.rs
+++ b/kernel/src/error.rs
@@ -75,6 +75,8 @@ pub enum SvsmError {
     InvalidAddress,
     /// Error reported when convert a usize to Bytes
     InvalidBytes,
+    /// Error reported when converting to UTF-8
+    InvalidUtf8,
     /// Errors related to firmware parsing
     Firmware,
     /// Errors related to console operation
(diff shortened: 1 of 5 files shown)


Freax13 (approved):

p4zuu:
Even though we check we can only `read()/write()` to user memory, could we consider `UserPtr::read()` and `UserPtr::write()` can create UB? If yes, maybe we should set those methods unsafe

Freax13:
> Even though we check we can only `read()/write()` to user memory, could we consider `UserPtr::read()` and `UserPtr::write()` can create UB? If yes, maybe we should set those methods to be unsafe

This was considered, but we're not aware of any way to cause UB, so marking them as safe should be fine.

PR #509, opened 2024-11-06
Userspace support: Introduce exit and close syscalls

vijaydhanraj:
This is 4th PR in the userspace series. This PR implements exit and close syscalls.

### DIFF
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -294,7 +294,7 @@ extern "C" fn ex_handler_system_call(
     };
 
     ctxt.regs.rax = match input {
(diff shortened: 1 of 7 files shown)


joergroedel (commented):
This does implement the CLOSE but not the OPEN syscall, any reason for that?

vijaydhanraj:
Idea was to create a simple `init` process that would be able to launch a `DM`. For this basic model, we didn't need `OPEN` syscall that is the reason it isn't part of this PR. But we do plan to implement `OPEN` in the next set of userspace PRs.

Freax13 (approved):

vijaydhanraj:
Rebased to latest main after PR https://github.com/coconut-svsm/svsm/pull/508 is merged.

Issue #483, opened 2024-10-13
Fix ELF Parser Alignment Handling, Remove Custom Linker Script, and Improve Handling of .bss Section in Binary Parsing


vijaydhanraj:
As suggested by @Freax13 in [PR 467](https://github.com/coconut-svsm/svsm/pull/467), the following fixes and clean-ups need to be done.

- [ ] In the current ELF parser implementation, `verify_phdr` function returns an error when p_vaddr is not aligned. But since spec doesn't guarantee that p_vaddr must be aligned, our ELF parser needs to be fixed. Current WA is to use a custom linker script to align p_vaddr.

- [ ] Remove the custom linker script for the `init` and `dm` once the ELF parser issue is fixed.

- [ ]  When parsing the binary, if `p_memsz` > `p_filesz`, the range `[p_filesz..p_memsz)` needs to be zeroed as per spec. This is usually true for `.bss` section. But we cannot anonymously map just `.bss` as the segment may contain other sections like `.data`. To resolve this issue, extend `VMFileMapping::new` such that if `p_filesz` is not a multiple of page size, we should make a private copy of the page using `.try_cop_page()` and zero out the bytes after `p_filesz`.

stefano-garzarella:
@vijaydhanraj IIUC you are working on them, so I assigned this issue to you. Please let me know if it is not the case.

vijaydhanraj:
For issue 1, fix is part of https://github.com/coconut-svsm/svsm/pull/539. Not pursuing issue 3, please see discussion in here, https://github.com/coconut-svsm/svsm/pull/510#discussion_r1865131508.

PR #510, opened 2024-11-08
Fix ELF parser issues

vijaydhanraj:
This PR fixes the issues mentioned in https://github.com/coconut-svsm/svsm/issues/483. It takes care of issues 1 and 3 which should automatically resolve issue 2 as well.

### DIFF
--- a/kernel/src/mm/vm/mapping/file_mapping.rs
+++ b/kernel/src/mm/vm/mapping/file_mapping.rs
@@ -18,7 +18,7 @@ use crate::mm::vm::VMR;
 use crate::mm::PageRef;
 use crate::mm::{pagetable::PTEntryFlags, PAGE_SIZE};
 use crate::types::PAGE_SHIFT;
-use crate::utils::align_up;
+use crate::utils::{align_up, is_aligned, zero_mem_region};
 
 bitflags! {
     #[derive(Debug, PartialEq, Copy, Clone)]
@@ -90,7 +90,13 @@ impl VMFileMapping {
                 .mapping(offset + page_index * PAGE_SIZE)
                 .ok_or(SvsmError::Mem)?;
             if flags.contains(VMFileMappingFlags::Private) {
-                pages.push(page_ref.try_copy_page()?);
+                let private_page_ref = page_ref.try_copy_page()?;
+                if page_index == count - 1 && !is_aligned(size as u64, PAGE_SIZE as u64) {
+                    let start = private_page_ref.virt_addr() + (size & (PAGE_SIZE - 1));
+                    let end = private_page_ref.virt_addr() + PAGE_SIZE;
+                    zero_mem_region(start, end);
+                }
+                pages.push(private_page_ref);
             } else {
                 pages.push(page_ref);
             }
--- a/kernel/src/task/exec.rs
+++ b/kernel/src/task/exec.rs
@@ -11,6 +11,7 @@ use crate::mm::vm::VMFileMappingFlags;
 use crate::mm::USER_MEM_END;
 use crate::task::{create_user_task, current_task, schedule};
 use crate::types::PAGE_SIZE;
+use crate::utils::align_up;
 use elf::{Elf64File, Elf64PhdrFlags};
 
 fn convert_elf_phdr_flags(flags: Elf64PhdrFlags) -> VMFileMappingFlags {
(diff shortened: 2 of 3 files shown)


kernel/src/mm/vm/mapping/file_mapping.rs
```diff
@@ -90,7 +90,16 @@ impl VMFileMapping {
                 .mapping(offset + page_index * PAGE_SIZE)
                 .ok_or(SvsmError::Mem)?;
             if flags.contains(VMFileMappingFlags::Private) {
-                pages.push(page_ref.try_copy_page()?);
+                let private_page_ref = page_ref.try_copy_page()?;
+                if page_index == count - 1
+                    && flags.contains(VMFileMappingFlags::Write)
```
Freax13, line 95: Why do we need to check if the page is writable?

kernel/src/task/exec.rs
```diff
@@ -53,14 +54,32 @@ pub fn exec_user(binary: &str) -> Result<(), SvsmError> {
...
+            // Handle unaligned VirtAddr and Offset
+            let (start_aligned, offset, size) = if !virt_start.is_aligned(PAGE_SIZE) {
+                (
+                    virt_start.page_align(),
+                    file_offset - virt_start.page_offset(),
+                    file_size + virt_start.page_offset(),
+                )
+            } else {
+                (virt_start, file_offset, file_size)
+            };
```
Freax13, line 74: We don't need this if-block here, do we? The calculations also work out if we always use `virt_start.page_align()`, `file_offset - virt_start.page_offset()`, and `file_size + virt_start.page_offset()`, don't they?

kernel/src/mm/vm/mapping/file_mapping.rs
vijaydhanraj: I think we can skip this redundant check as when `try_copy_page` is called, a page is allocated and zeroed out, meaning it already writable. Also, `flags` may not even be relevant to this newly allocated page.

kernel/src/task/exec.rs
vijaydhanraj: True, wanted to avoid page_* methods if the start address is already aligned but will remove the if-block.

Freax13 (approved):

kernel/src/mm/vm/mapping/file_mapping.rs
```diff
@@ -90,7 +90,13 @@ impl VMFileMapping {
...
                 .ok_or(SvsmError::Mem)?;
             if flags.contains(VMFileMappingFlags::Private) {
-                pages.push(page_ref.try_copy_page()?);
+                let private_page_ref = page_ref.try_copy_page()?;
+                if page_index == count - 1 && !is_aligned(size as u64, PAGE_SIZE as u64) {
+                    let start = private_page_ref.virt_addr() + (size & (PAGE_SIZE - 1));
+                    let end = private_page_ref.virt_addr() + PAGE_SIZE;
+                    zero_mem_region(start, end);
+                }
+                pages.push(private_page_ref);
```
joergroedel, line 99: I wonder why this is necessary. File pages are zeroed out on allocation and try_copy_page() always copies full pages. The truncate operation on files also zeroes out all content after EOF when the file. So anything in the page after EOF should already be zero.

kernel/src/task/exec.rs
```diff
@@ -53,14 +54,26 @@ pub fn exec_user(binary: &str) -> Result<(), SvsmError> {
...
+            let offset = file_offset - virt_start.page_offset();
+            let size = file_size + virt_start.page_offset();
+            task.mmap_user(start_aligned, Some(&fh), offset, size, flags)?;
+
+            let size_aligned = align_up(size, PAGE_SIZE);
+            if size_aligned < len {
+                let start_anon = start_aligned.const_add(size_aligned);
+                let remaining_len = len - size_aligned;
+                task.mmap_user(start_anon, None, 0, remaining_len, flags)?;
+            }
```
joergroedel, line 76: Okay, I missed that and have a similar fix in #525 (As I have seen you just noticed as well :) ).

joergroedel (commented):
Okay, looks good execpt for the open question about zeroing memory in the file mapping path.

kernel/src/mm/vm/mapping/file_mapping.rs
joergroedel: Okay, so you are trying to support segment boundaries which are not aligned to page-size. Please help me understand how this interacts with segment permissions, e.g. RO vs. RW sections?

In any case, the file_mapping code is the wrong place to zero out partial pages, this needs to happen in the ELF loader.

vijaydhanraj: > Please help me understand how this interacts with segment permissions, e.g. RO vs. RW sections?

The new mapping created (comprising of the newly allocated and updated with the contents of file) is simply given the same permissions as the segment permissions.

> In any case, the file_mapping code is the wrong place to zero out partial pages, this needs to happen in the ELF loader.
 
Yes, ELF loader would be more appropriate but when we do `mmap_user` we don't seem to use `file_contents` returned as part of the `image_load_segment_iter` call. In `mmap_user` we end up reading again from the file handle and updating the mapping. So even if we update and zero-out partial pages as part of `file_contents` returned by `image_load_segment_iter` it may not be helpful. Felt `VMFileMapping::new()` might be more appropriate to zero out the region that are not aligned to page-size. Please let me know if you had something else in mind.

joergroedel: I picked up two of the three patches from this PR and included them into #525 . This change I left out because I think a proper fix requires some changes in how user-space tasks are set up.

To correctly fix this the ELF file needs to be loaded and mapped in the context of the new task. This makes sure that any memory which is about to be zeroed out is already mapped in the same `CR3` and can be directly accessed (from a CPU level, still need safe Rust abstractions).

This is a bigger change and we do not need to block ourselves on it, so meanwhile I suggest we keep a linker script for all user-space binaries which page-aligns the section boundaries.

vijaydhanraj: Agree, this makes sense. Should we create an issue to track this?

vijaydhanraj: Should I close this PR given that the commits are pulled into https://github.com/coconut-svsm/svsm/pull/539?

vijaydhanraj: As discussed in our community meeting closing this PR, as the following commits from this PR are pulled into https://github.com/coconut-svsm/svsm/pull/539
[elf: Fix elf program header verification logic](https://github.com/coconut-svsm/svsm/pull/510/commits/cd3cab9ea4529564252a5fa0e8b1a0a1a706a3f2)
[kernel/task: Remove p_vaddr alignment requirement from exec_user()](https://github.com/coconut-svsm/svsm/pull/510/commits/ee710d0f853eb03710bb86dad97d139a2fc4ac82)

Regarding [kernel/mm/vm/file-mapping: Zero-out memory not backed by file](https://github.com/coconut-svsm/svsm/pull/510/commits/c6261e4007a03dd93d24dc9a80b0ed50ea6b7391), as suggested by @joergroedel plan is to load and map elf file in the context of the new task which makes sure that any memory which is about to be zeroed out is already mapped and can be directly accessed. Until then keep a linker script for all user-space binaries which page-aligns the section boundaries.

PR #515, opened 2024-11-12
platform/tdp: replace tdx-tdcall crate with local implementation

msft-jlange:
The `tdx-tdcall` crate is built for different security and design assumptions than what is required for COCONUT-SVSM.  The TDCALL/TDVMCALL functionality required can be implemented locally, providing better integration with the COCONUT-SVSM environment.


msft-jlange:
@peterfang it would great to get your review on this

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,206 @@
...
+        vaddr = vaddr + size;
+    }
+    Ok(())
+}
+
+pub fn tdvmcall_halt() {
+    let pass_regs = (1 << 10) | (1 << 11) | (1 << 12);
+    unsafe {
+        asm!("tdcall",
+             in("eax") 0,
```
Freax13, line 159: Is there a reason we're using the 32-bit register names instead of the 64-bit ones?

Either way, AFAICT the rust compiler always [treats them as the 64-bit registers](https://doc.rust-lang.org/reference/inline-assembly.html#asm.register-names.supported-register-aliases), so this shouldn't be a problem.

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,206 @@
...
+    // I/O operation was performed by an untrusted source, the error
+    // information is not meaningfully different than a maliciously unreliable
+    // operation.
+
+    output
+}
+
+pub fn tdvmcall_io_write<T>(port: u16, data: T)
+where
+    T: Sized,
```
Freax13, line 195:
```suggestion
```
`T` must be `Sized` by default unless there's a `?Sized` bound. `From<T>` also implies that `T` is `Sized`.

This also applies to `tdvmcall_io_read`.

kernel/src/platform/tdp.rs
```diff
@@ -187,26 +166,26 @@ impl GHCIIOPort {
 
 impl IOPort for GHCIIOPort {
     fn outb(&self, port: u16, value: u8) {
-        tdvmcall_io_write_8(port, value);
+        tdvmcall_io_write::<u8>(port, value);
```
Freax13, line 169:
```suggestion
        tdvmcall_io_write(port, value);
```

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,206 @@
...
+             in("r14d") port as u32,
+             in("r15d") data,
+             lateout("r11d") output,
+             options(att_syntax));
+    }
+
+    // Ignore errors here.  The caller cannot handle them, and since the
+    // I/O operation was performed by an untrusted source, the error
+    // information is not meaningfully different than a maliciously unreliable
+    // operation.
```
Freax13, line 188: Information about errors could still be useful for debugging. We could use the `debug_assert!` family of macros. These are only active in development mode and don't do anything in release mode.

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,206 @@
...
+
+pub fn tdvmcall_halt() {
+    let pass_regs = (1 << 10) | (1 << 11) | (1 << 12);
+    unsafe {
+        asm!("tdcall",
+             in("eax") 0,
+             in("ecx") pass_regs,
+             in("r10d") 0,
+             in("r11d") TDVMCALL_HLT,
+             in("r12d") 0,
```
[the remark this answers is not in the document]
msft-jlange: Is there a reference I can read to understand the differences between `in`, `out`, and `lateout`, or the `=>` syntax you mention?  These assembler directives are still a bit of a mystery to me.

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,206 @@
...
+pub fn td_accept_physical_memory(region: MemoryRegion<PhysAddr>) -> Result<(), SvsmError> {
+    let mut addr = region.start();
+    let end = region.end();
+
+    while addr < end {
+        if addr.is_aligned(PAGE_SIZE_2M) && addr + PAGE_SIZE_2M <= end {
+            let ret = tdx_result(tdg_mem_page_accept(u64::from(addr) | 1));
+            match ret {
+                Err(TdxError::PageAlreadyAccepted) => {
+                    // The caller is expected not to accept a page twice unlee
```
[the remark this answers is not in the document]
msft-jlange: Thanks for catching my typos.

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,206 @@
...
+        asm!("tdcall",
+         in("rax") TDCALL_TDG_MEM_PAGE_ACCEPT,
+         in("rcx") gpa_info,
+         lateout("rax") ret,
+         options(att_syntax));
+    }
+    ret
+}
+
+pub fn td_accept_physical_memory(region: MemoryRegion<PhysAddr>) -> Result<(), SvsmError> {
```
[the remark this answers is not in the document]
msft-jlange: To test your reasoning, do you believe the existing `pvalidate_range` function should also be marked unsafe?  If not, can you help me understand what makes it different from `td_accept_physical_memory`?

I've heard different points of view on whether non-public functions require a declaration of unsafe.  Sounds like you believe all functions that have unsafe behavior should be declared `unsafe` regardless of scope.  Correct?  Just to enhance my understanding, what attribute of these functions renders them unsafe in your mind?

kernel/src/tdx/tdcall.rs
msft-jlange: There is a reason.  The x86-64 ISA guarantees that every 32-bit load is observed either as a zero-extended or a sign-extended version of the 32-bit value, depending on the opcode, but for 32-bit positive numbers, the result is always a zero-extension.  And the assembly for 32-bit operations is almost always smaller.  For example, `xor rax, rax` is 3 bytes while `xor eax, eax` is only two, but produces the same result.  Similarly, `mov eax, 0` is 5 bytes while `mov rax, 0` is 9.  So when writing assembly, it's always best practice to choose code instruction encodings that achieve the correct result in the minimum number of opcode bytes.  It might be the case that the rust compiler figures this out automatically, but there's no reason to depend on compiler smarts when explicit naming will guarantee the correct result.

kernel/src/tdx/tdcall.rs
msft-jlange: Thanks - I hadn't realized that `Sized` was always assumed in cases like these.  You'd think clippy would tell me that it's unnecessary (is there a lint we should be turning on?)

kernel/src/tdx/tdcall.rs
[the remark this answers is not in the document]
msft-jlange: > Is there a reference I can read to understand the differences between `in`, `out`, and `lateout`, or the `=>` syntax you mention? These assembler directives are still a bit of a mystery to me.

I found it: https://doc.rust-lang.org/reference/inline-assembly.html

kernel/src/tdx/tdcall.rs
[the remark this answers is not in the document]
msft-jlange: Oh, I see how this problem is unique to TDX - the low bits of the physical address are interpreted as part of the GPA info, so there must be a guarantee that they are zero coming in.

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,206 @@
...
+use crate::utils::MemoryRegion;
+
+use core::arch::asm;
+
+const TDCALL_TDG_MEM_PAGE_ACCEPT: u32 = 6;
+
+const TDVMCALL_HLT: u32 = 12;
+const TDVMCALL_IO: u32 = 30;
+
+fn tdg_mem_page_accept(gpa_info: u64) -> u64 {
```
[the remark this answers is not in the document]
msft-jlange: The existing `PageFrame` structure should do nicely for this.

kernel/src/tdx/tdcall.rs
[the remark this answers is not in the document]
Freax13: > To test your reasoning, do you believe the existing `pvalidate_range` function should also be marked unsafe? If not, can you help me understand what makes it different from `td_accept_physical_memory`?

Yes, I believe `pvalidate_range` should also be marked as unsafe.

> I've heard different points of view on whether non-public functions require a declaration of unsafe. Sounds like you believe all functions that have unsafe behavior should be declared `unsafe` regardless of scope. Correct?

Yes.

> Just to enhance my understanding, what attribute of these functions renders them unsafe in your mind?

These functions allow safe code to zero out arbitrary memory. It's essentially a specialized version of [write_bytes](https://doc.rust-lang.org/std/primitive.pointer.html#method.write_bytes) that always writes zeros. This could violate Rust's aliasing rules (e.g. write to memory pointed to by a shared reference) or violate the bit pattern validity rules of a type (e.g. write all zeros to a value of any of the `NonZeroU8/16/32/64` types). `pvalidate_range` is even worse as incorrect usage allows the host to change between two versions of the page at will.

kernel/src/tdx/tdcall.rs
Freax13: The Rust compiler doesn't differentiate between `eax`, `rax`, or even `ax`. It will always emit the same assembly, so we can't benefit from the optimization you've described above.

Another problem I just noticed is this [section](https://doc.rust-lang.org/reference/inline-assembly.html#asm.register-operands.smaller-value) in the Rust reference stating that if a small value is used to fill a register, the upper bits are undefined, they are not necessarily zero. There's even a warning against this: https://godbolt.org/z/KTK6b9aGv. My concern (however unlikely) is that the upper bits may contain secrets that we inadvertently leak to the host. IMO, when filling registers that are exposed to the host, we should always specify all 64 bits, so that nothing is undefined.

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,234 @@
...
+
+use core::arch::asm;
+
+const TDG_VP_TDVMCALL: u32 = 0;
+const TDCALL_TDG_MEM_PAGE_ACCEPT: u32 = 6;
+
+const TDVMCALL_HLT: u32 = 12;
+const TDVMCALL_IO: u32 = 30;
+
+fn gpa_info(frame: PageFrame) -> u64 {
```
Freax13, line 23: Out of curiosity, you've mentioned "gpa info" a couple of times, but I can't find this anywhere in the ABI spec. Where did you find this?

ABI spec 1.0:
image

ABI spec 1.5:
image

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,234 @@
...
+    // Callers are expected to guarantee alignment before calling this
+    // function.  If the caller verifies the alignment, then it can be done
+    // once without having to do it in a loop.
+    debug_assert!(u64::from(frame.address()) & (frame.size() - 1) as u64 == 0);
```
Freax13, line 27:
```suggestion
    debug_assert_eq!(u64::from(frame.address()) & (frame.size() - 1) as u64, 0);
```
If the value is not `0`, it'd be useful to know what it was.

What's the reasoning behind using `debug_assert!` rather than just `assert!`. Is see that we're already checking the alignment in `td_accept_physical_memory` and `td_accept_virtual_memory`, but we don't in `td_accept_virtual_4k` and `td_accept_virtual_2m`.

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,234 @@
...
+    ret
+}
+
+pub fn td_accept_physical_memory(region: MemoryRegion<PhysAddr>) -> Result<(), SvsmError> {
+    let mut addr = region.start();
+    if !addr.is_aligned(PAGE_SIZE) {
+        return Err(SvsmError::InvalidAddress);
+    }
+
+    let end = region.end();
```
Freax13, line 53: Can we also check the alignment of `end`?

This also applies to `td_accept_virtual_memory`.

kernel/src/tdx/tdcall.rs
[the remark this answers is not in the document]
msft-jlange: > Yes, I believe `pvalidate_range` should also be marked as unsafe.

By that logic, the page validation methods of `SvsmPlatform` should also be marked unsafe, since the platform-specific implementations have no ability to determine how to make the underlying calls safely.  That being the case, the request for `unsafe` handling goes far beyond the scope of this PR, which simply intends to replace the `tdx-tdcall` crate with local implementations.  I will mark the new local functions as unsafe, and add a TODO comment in the `TdpPlatform` versions, and the broader safety issues should be solved with a cross-platform PR.

kernel/src/tdx/tdcall.rs
msft-jlange: > Another problem I just noticed is this [section](https://doc.rust-lang.org/reference/inline-assembly.html#asm.register-operands.smaller-value) in the Rust reference stating that if a small value is used to fill a register, the upper bits are undefined, they are not necessarily zero. There's even a warning against this: https://godbolt.org/z/KTK6b9aGv. My concern (however unlikely) is that the upper bits may contain secrets that we inadvertently leak to the host. IMO, when filling registers that are exposed to the host, we should always specify all 64 bits, so that nothing is undefined.

That's an unfortunate gap in the Rust assembler implementation, because it appears to contradict the definitions of the x86-64 ISA, which is very counterintuitive.  In any case, I've verified that when using 64-bit registers, the compiler will emit 32-bit register instructions, so using 64-bit registers in the source will still result in the assembly I'm looking for.  I'll make the change.

kernel/src/tdx/tdcall.rs
msft-jlange: I must have hallucinated the term.  The same concept appears not only in TDG.MEM.PAGE.ACCEPT, but also other guest-side management operations like TDG.MEM.PAGE.ATTR.WR and host-side GPA management operations like TDH.MEM.PAGE.ADD or TDH.MEM.PAGE.DEMOTE.  It seems the documentation consistently uses the term "EPT mapping information" so I can change the name in our code.

peterfang:
> @peterfang it would great to get your review on this

Sure, I'll take a look

kernel/src/tdx/tdcall.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    while addr < end {
+        if addr.is_aligned(PAGE_SIZE_2M) && addr + PAGE_SIZE_2M <= end {
+            let ret = tdx_result(tdg_mem_page_accept(PageFrame::Size2M(addr)));
+            match ret {
+                Err(TdxError::PageAlreadyAccepted) => {
```
ereshetova, line 78: What is the usecase to allow such behavior, i.e. why we don't treat PageAlreadyAccepted as a bug condition?

msft-jlange: `PageAlreadyAccepted` does not necessarily indicate an error.  Suppose a TD has a private page at GPA X, and calls GHCI.MapGPA to convert it to a shared page.  The host may choose to insert a different HPA at GPA X+Shared without ever calling TDH.MEM.PAGE.REMOVE on GPA X.  This is acceptable to the TD.  Then, the TD may call GHCI.MapGPA to convert back to a pivate page.  The host may choose simply to remove GPA X+Shared, still without ever calling TDH.MEM.PAGE.REMOVE on GPA X.  There's nothing inappropriate about this flow.  The TD is expecting to accept a page of zeroes, so if it sees that the page was already accepted in this case, it should just zero the page and move on.

Similarly, not observing `PageAlreadyAccepted` does not mean that acceptance is safe.  A TD may call TDG.MEM.PAGE.ACCEPT on GPA X, and then may call TDG.MEM.PAGE.ACCEPT on GPA X a second time (perhaps intentionally, perhaps inadvertently).  If the host calls TDH.MEM.PAGE.REMOVE/TDH.MEM.PAGE.AUG in between the two calls, the second TDG.MEM.PAGE.ACCEPT call will succeed and zero the page, even if the TD expected that the page had already been accepted.

Since `PageAlreadyAccepted` is not a reliable way to detect an error condition, it's better just to make page acceptance be consistent, and to say that every call will result in a zeroed page regardless of what was there before.

Freax13 (approved):

kernel/src/tdx/tdcall.rs
ereshetova: >PageAlreadyAccepted does not necessarily indicate an error. Suppose a TD has a private page at GPA X, and calls >GHCI.MapGPA to convert it to a shared page. The host may choose to insert a different HPA at GPA X+Shared without >ever calling TDH.MEM.PAGE.REMOVE on GPA X. This is acceptable to the TD. Then, the TD may call GHCI.MapGPA to >convert back to a pivate page. The host may choose simply to remove GPA X+Shared, still without ever calling >TDH.MEM.PAGE.REMOVE on GPA X. There's nothing inappropriate about this flow. The TD is expecting to accept a page >of zeroes, so if it sees that the page was already accepted in this case, it should just zero the page and move on.

Interesting, indeed nothing should be wrong with this flow, but we so far haven't supported this for KVM (there hasnt been any requests). Given that guest indeed tracks by itself when it expect safe acceptance to happen, it should be fine. 

What is the usecase behind the above flow? Perf optimization in case you expect a lot of runtime conversions or?

msft-jlange: > What is the usecase behind the above flow? Perf optimization in case you expect a lot of runtime conversions or?

I'm not aware of a use case for this flow, but that's not relevant.  The architecture permits this flow, and thus it must be handled.  TD images built now will be expected to function without modification even as host architectures evolve over the next several years, on all hypervisors (not just KVM) - this is the essence of virtualization.  Anything a hypervisor can legitimately implement according to the architecture is something that the guest OS must be prepared to handle.

peterfang:
@msft-jlange I'm sharing my feedback a few hours too late...
The commit looks good to me and should be correct. My feedback is mostly around TDX error handling. Right now we only have a small handful of TDG calls so this isn't a problem, but I believe putting some logic in there would at least make it more aligned with the spec. I'll create a PR so we can continue this discussion.

PR #517, opened 2024-11-14
stage1: Rustify stage1 and update IGVM VTOM on TDP platforms

peterfang:
Move stage1/*.S to a Rust crate. This allows for better interfacing with other crates and reduces hard-coding and chances of breakage.

Update IgvmParamBlock.vtom in addition to Stage2LaunchInfo.vtom during IGVM boot. It ensures Stage2 & the SVSM kernel can use VTOM as expected.


peterfang:
This fails CI as stage1/stage1.o and stage1/reset.o no longer exist.

Makefile
```diff
@@ -141,42 +139,26 @@ ifneq ($(FS_FILE), none)
 endif
 	touch ${FS_BIN}
 
-stage1/stage1.o: stage1/stage1.S bin/stage2.bin bin/svsm-fs.bin bin/svsm-kernel.elf bin
+bin/svsm.bin: bin/stage2.bin bin/svsm-fs.bin bin/svsm-kernel.elf bin/meta.bin
 	ln -sf svsm-kernel.elf bin/kernel.elf
-	cc -c -DLOAD_STAGE2 -o $@ $<
+	cargo rustc --target i686-unknown-linux-gnu --manifest-path stage1/Cargo.toml ${CARGO_ARGS} --features load-stage2 --bin stage1 -- -C panic=abort
```
joergroedel, line 144: Do you see a change of making this work with the `x86-64-unknown-none` target? I'd like to avoid pull in another rust target into our build process.

joergroedel (changes requested):
Hi Peter,

I like this change, it will simplify the build process and allows more sharing of constants and code between the different stages.

Can you please adapt the CI files to make it pass with these changes and have a look at using the existing `x86-64-unknown-none` target? I tried that already but ran into linker errors due to 32-bit relocations in 64-bit code. But maybe there is a way around that.

kernel/src/platform/tdp.rs
```diff
@@ -44,6 +44,9 @@ impl SvsmPlatform for TdpPlatform {
     }
 
     fn env_setup(&mut self, debug_serial_port: u16, vtom: usize) -> Result<(), SvsmError> {
+        if vtom == 0 {
```
msft-jlange, line 47: May be simpler just to `assert_ne!(vtom, 0)` so that we can panic and get it over with (with a clear error message) instead of making some other component deal with a generic error.

Makefile
peterfang: I was wrestling with the "relocation truncated to fit" errors when using `x86-64-unknown-none`. I'll spend more time on this to see if there's a way to keep the target platform as is.

kernel/src/platform/tdp.rs
peterfang: That makes sense. Something's probably seriously wrong if we see VTOM == 0. I'll make the change.

Makefile
peterfang: I have found a way to remove this error. Will be updating this PR soon.

peterfang:
> Hi Peter,
> 
> I like this change, it will simplify the build process and allows more sharing of constants and code between the different stages.
> 
> Can you please adapt the CI files to make it pass with these changes and have a look at using the existing `x86-64-unknown-none` target? I tried that already but ran into linker errors due to 32-bit relocations in 64-bit code. But maybe there is a way around that.

Hi Joerg, I sidestepped the ld issue on the CI system using `--emit=obj -C linker=/usr/bin/true`. It should work as expected now.

joergroedel (approved):
Changes look good to me now, thanks.

PR #519, opened 2024-11-15
workspace: enable `unsafe_op_in_unsafe_fn` lint

msft-jlange:
It is good to mark functions as `unsafe` when the safety of their behavior is not guaranteed.  It is not goot to permit these functions to perform additional unsafe operations without warning.  All uses of unsafe code should require explicit `unsafe` annotation regardless of whether they appear in safe or unsafe functions.

### DIFF
--- a/kernel/src/cpu/sse.rs
+++ b/kernel/src/cpu/sse.rs
@@ -79,14 +79,16 @@ pub fn sse_init() {
 /// no other part of the code is accessing this memory at the same time.
 pub unsafe fn sse_save_context(addr: u64) {
     let save_bits = XCR0_X87_ENABLE | XCR0_SSE_ENABLE | XCR0_YMM_ENABLE;
-    asm!(
-        r#"
-        xsaveopt (%rsi)
-        "#,
-        in("rsi") addr,
-        in("rax") save_bits,
-        in("rdx") 0,
-        options(att_syntax));
+    unsafe {
+        asm!(
+            r#"
+            xsaveopt (%rsi)
+            "#,
+            in("rsi") addr,
+            in("rax") save_bits,
+            in("rdx") 0,
+            options(att_syntax));
+    }
 }
 
 /// # Safety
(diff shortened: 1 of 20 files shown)


kernel/src/cpu/sse.rs
```diff
@@ -79,14 +79,16 @@ pub fn sse_init() {
 /// no other part of the code is accessing this memory at the same time.
 pub unsafe fn sse_save_context(addr: u64) {
     let save_bits = XCR0_X87_ENABLE | XCR0_SSE_ENABLE | XCR0_YMM_ENABLE;
-    asm!(
+    unsafe {
+        asm!(
```
Freax13, line 83: rustfmt seems to choke on the multi-line asm! block. Could you please fix the indentation levels of these?

Freax13: rustfmt can't format multi-line asm! blocks it just leaves them as-is.

For example, previously we had this:
```rust
asm!(
    r#"
    xsaveopt (%rsi)
    "#,
    in("rsi") addr,
    in("rax") save_bits,
    in("rdx") 0,
    options(att_syntax));
```
Now with the new unsafe block, rustfmt shifts the `asm!` line one level to the right, but it just ignores all the other lines resulting in this:
```rust
unsafe {
    asm!(
    r#"
    xsaveopt (%rsi)
    "#,
    in("rsi") addr,
    in("rax") save_bits,
    in("rdx") 0,
    options(att_syntax));
}
```
Note how the `asm!` line and all the other lines are now on the same indentation level.


Please format them in such a way that everything in the parenthesis is indented one level more to the right than the start of the block (as was the case before this PR).

msft-jlange: Will do.  We should generate a document that describes this standard so that until rustfmt can do it correctly, people like me aren't required to guess at the expected format.

msft-jlange: I think I got them all.

PR #524, opened 2024-11-21
kernel/tdx: Fix TDX error codes

peterfang:
- Fix the definition of TdxError::PageAlreadyAccepted
- Handle TDX_OPERAND_BUSY in TDG.MEM.PAGE.ACCEPT
- Do sanity checks on status codes

### DIFF
--- a/kernel/src/tdx/error.rs
+++ b/kernel/src/tdx/error.rs
@@ -7,11 +7,17 @@
 use crate::error::SvsmError;
 
 #[derive(Clone, Copy, Debug, PartialEq)]
-pub enum TdxError {
-    Unknown(u64),
-    Unimplemented,
+pub enum TdxSuccess {
+    Success,
     PageAlreadyAccepted,
+    Unknown(u64),
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum TdxError {
     PageSizeMismatch,
+    Unimplemented,
+    Unknown(u64),
 }
 
 impl From<TdxError> for SvsmError {
@@ -20,14 +26,18 @@ impl From<TdxError> for SvsmError {
     }
 }
 
-pub fn tdx_result(err: u64) -> Result<u64, TdxError> {
+pub fn tdx_result(err: u64) -> Result<TdxSuccess, TdxError> {
     let code = err >> 32;
     if code < 0x8000_0000 {
-        return Ok(code);
-    }
-    match code {
-        0xC000_0B0A => Err(TdxError::PageAlreadyAccepted),
-        0xC000_0B0B => Err(TdxError::PageSizeMismatch),
-        _ => Err(TdxError::Unknown(err)),
+        match code {
+            0 => Ok(TdxSuccess::Success),
+            0x0000_0B0A => Ok(TdxSuccess::PageAlreadyAccepted),
+            _ => Ok(TdxSuccess::Unknown(err)),
+        }
+    } else {
+        match code {
+            0xC000_0B0B => Err(TdxError::PageSizeMismatch),
+            _ => Err(TdxError::Unknown(err)),
+        }
     }
 }
--- a/kernel/src/tdx/error.rs
+++ b/kernel/src/tdx/error.rs
@@ -41,3 +41,9 @@ pub fn tdx_result(err: u64) -> Result<TdxSuccess, TdxError> {
         }
     }
 }
+
+pub fn tdx_recoverable_error(err: u64) -> bool {
+    // Bit 63: ERROR
+    // Bit 62: NON_RECOVERABLE
+    (err >> 62) == 2
+}
--- a/kernel/src/tdx/error.rs
+++ b/kernel/src/tdx/error.rs
@@ -17,6 +17,13 @@ pub enum TdxSuccess {
 pub enum TdxError {
     PageSizeMismatch,
     Unimplemented,
+    Vmcall(TdVmcallError),
+    Unknown(u64),
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum TdVmcallError {
+    OperandInvalid,
     Unknown(u64),
 }
 
@@ -46,4 +53,12 @@ pub fn tdx_recoverable_error(err: u64) -> bool {
     // Bit 63: ERROR
     // Bit 62: NON_RECOVERABLE
     (err >> 62) == 2
+}
+
+pub fn tdvmcall_result(err: u64) -> Result<(), TdxError> {
+    match err {
+        0 => Ok(()),
+        0x8000_0000_0000_0000 => Err(TdxError::Vmcall(TdVmcallError::OperandInvalid)),
+        _ => Err(TdxError::Vmcall(TdVmcallError::Unknown(err))),
+    }
 }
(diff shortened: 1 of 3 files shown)


kernel/src/tdx/error.rs
```diff
@@ -10,24 +10,39 @@ use crate::error::SvsmError;
...
         SvsmError::Tdx(err)
     }
 }
 
-pub fn tdx_result(err: u64) -> Result<u64, TdxError> {
+pub fn tdx_result(err: u64) -> Result<(), TdxError> {
     let code = err >> 32;
-    if code < 0x8000_0000 {
-        return Ok(code);
+    if code == 0 {
```
msft-jlange, line 31: The TDX architecture defines results codes less than 0x8000_0000_0000_0000 as non-fatal conditions.  We can't treat a non-zero success condition as an error.  This is why this function is defined to return a `u64` in the `Ok` variant of the result - to provide a specific value of the success return in case it is a non-zero success value.

kernel/src/tdx/tdcall.rs
```diff
@@ -50,14 +50,28 @@ unsafe fn tdg_mem_page_accept(frame: PageFrame) -> u64 {
...
 }
 
+/// # Safety
+/// This function has the potential to zero arbitrary memory, so the caller is
+/// required to ensure that the supplied physical address range is appropriate
+/// for acceptance.
+unsafe fn accept_one_page(frame: PageFrame) -> Result<(), TdxError> {
+    for _ in 0..10000 {
+        match tdx_result(tdg_mem_page_accept(frame)) {
+            Err(TdxError::OperandBusy) => continue,
```
msft-jlange, line 68: My understanding is that TDX_OPERAND_BUSY only applies to TDH operations and not TDG operations.  When a TDG operation encounters a busy state, the TDX module will simply not advance RIP past the `tdcall` instruction, so the TD will naturally just reattempt the same call again.

kernel/src/tdx/tdcall.rs
```diff
@@ -50,14 +50,28 @@ unsafe fn tdg_mem_page_accept(frame: PageFrame) -> u64 {
...
     }
     ret
 }
 
+/// # Safety
+/// This function has the potential to zero arbitrary memory, so the caller is
+/// required to ensure that the supplied physical address range is appropriate
+/// for acceptance.
+unsafe fn accept_one_page(frame: PageFrame) -> Result<(), TdxError> {
+    for _ in 0..10000 {
```
msft-jlange, line 66: What makes 10000 - or any other arbitrary constant - a reasonable retry count?  Any retry count needs to be justified with on an architectural basis.

msft-jlange (changes requested):

kernel/src/tdx/error.rs
peterfang: Yes, there are recoverable and unrecoverable error codes. I was going through those and was unable to come up with a conclusion as to whether SVSM would view them as OK() cases, and hence this change.
Are you okay with `PageAlreadyAccepted` being an Ok() case then? That would also mean conditions like `tdx_result(ret).is_ok()` would include this case.

kernel/src/tdx/tdcall.rs
peterfang: `TDX_OPERAND_BUSY` is a possible outcome for TDG calls as well. In this case, `TDG.MEM.PAGE.ACCEPT` could fail to acquire an internal lock (i.e. SEPT lock) due to a host or a TDX module operation.

kernel/src/tdx/tdcall.rs
peterfang: It was an empirical number that I saw in reference code. Do you have any suggestions on how we can improve this?
There are also implementations in the Linux kernel (for TDH calls) where this is retried indefinitely until the call returns a difference status code. I wasn't not sure if we want to adopt that strategy in SVSM.

kernel/src/tdx/tdcall.rs
```diff
@@ -50,14 +50,28 @@ unsafe fn tdg_mem_page_accept(frame: PageFrame) -> u64 {
...
          options(att_syntax));
     }
     ret
 }
 
+/// # Safety
+/// This function has the potential to zero arbitrary memory, so the caller is
+/// required to ensure that the supplied physical address range is appropriate
+/// for acceptance.
+unsafe fn accept_one_page(frame: PageFrame) -> Result<(), TdxError> {
```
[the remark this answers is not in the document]
peterfang: I can confirm that `TDX_OPERAND_BUSY` is a case we should handle.

kernel/src/tdx/error.rs
msft-jlange: Somehow I had missed that TDX_PAGE_ALREADY_ACCEPTED was in the range of success codes.  I think a better approach here would be to define a `TdxSuccess` class to enumerate specific known success values, and to change this function to return `Result<TdxSuccess, TdxError>`.  Callers of `tdg_mem_page_accept` can match on `Ok(PageAlreadyAccepted)` instead of `Err(PageAlreadyAccepted)` to get the behavior we need.  But in general, `is_ok()` should always match all success codes.

kernel/src/tdx/tdcall.rs
msft-jlange: Looping forever is reasonable.  It is likely to produce a better outcome than random errors.

kernel/src/tdx/error.rs
```diff
@@ -10,24 +10,39 @@ use crate::error::SvsmError;
 pub enum TdxError {
     Unknown(u64),
     Unimplemented,
+    OperandBusy,
     PageAlreadyAccepted,
     PageSizeMismatch,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum TdVmcallError {
```
msft-jlange, line 19: I don't see any conversion path from `TdVmcallError` to `SvsmError`.  We want to ensure that all defined error types can be represented in `SvsmError`.  I suspect that the best way to do this is not to define `TdVmcallError` as a separate class, but rather as an element of `TdxError`.  Then your ` tdvmcall_result` function could return `TdxError::TdVmcallError(result)` and stay within the error class that is already established.

kernel/src/tdx/tdcall.rs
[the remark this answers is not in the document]
peterfang: Sure, I can simply retry on all recoverable error codes when doing a TDCALL.

kernel/src/tdx/error.rs
peterfang: Good idea. I'll make this change.

msft-jlange (approved):

kernel/src/tdx/tdcall.rs
```diff
@@ -216,35 +222,41 @@ pub unsafe fn td_accept_virtual_memory(region: MemoryRegion<VirtAddr>) -> Result
...
              in("r11") TDVMCALL_HLT,
              in("r12") 0,
+             lateout("rax") ret,
              lateout("r10") _,
              lateout("r11") _,
              lateout("r12") _,
              options(att_syntax));
     }
+    debug_assert!(tdx_result(ret).is_ok());
+    // r10 is guaranteed to be TDG.VP.VMCALL_SUCCESS
```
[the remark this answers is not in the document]
peterfang: Sure, `debug_assert`  make sense here.

PR #525, opened 2024-11-21
User -Mode Update

joergroedel:
This PR has an enhancement and a bug fix:

* Enhancement: Handle ELF LOAD segments where the virtual size is larger than the file-backed size.
* Fix: Make sure that half-initialized user-tasks do not end up on the TASKLIST or runqueue.


kernel/src/task/exec.rs
```diff
@@ -51,18 +52,34 @@ pub fn exec_user(binary: &str) -> Result<(), SvsmError> {
...
+        let file_len = file_end - file_offset;
         let flags = convert_elf_phdr_flags(seg.flags);
 
         if !virt_start.is_aligned(PAGE_SIZE) {
             return Err(SvsmError::Mem);
         }
 
         if file_offset > 0 {
-            task.mmap_user(virt_start, Some(&fh), file_offset, len, flags)?;
+            // Map file part
```
vijaydhanraj, line 75: ELF spec states that `file_len` (`p_filesz`)may not be larger than ` virt_len`(`p_memsz`), so it may be better to check for this error condition.

kernel/src/task/exec.rs
```diff
@@ -51,18 +52,34 @@ pub fn exec_user(binary: &str) -> Result<(), SvsmError> {
...
         let virt_start = VirtAddr::from(seg.vaddr_range.vaddr_begin);
         let virt_end = VirtAddr::from(seg.vaddr_range.vaddr_end).align_up(PAGE_SIZE);
+        let virt_len = virt_end - virt_start;
         let file_offset = seg.file_range.offset_begin;
-        let len = virt_end - virt_start;
+        let file_end = seg.file_range.offset_end;
+        let file_len = file_end - file_offset;
         let flags = convert_elf_phdr_flags(seg.flags);
 
         if !virt_start.is_aligned(PAGE_SIZE) {
```
vijaydhanraj, line 70: While we are at it, it is also better to fix the `virt_start` alignment requirement. ELF spec doesn't seem to mandate that `virt_start` (or in other words `p_vaddr`) should be page aligned.

kernel/src/task/exec.rs
vijaydhanraj: I already have a PR https://github.com/coconut-svsm/svsm/pull/510 opened that addresses ELF load segments which are not fully mapped from file (https://github.com/coconut-svsm/svsm/pull/510/commits/ee710d0f853eb03710bb86dad97d139a2fc4ac82) and also the above-mentioned fixes. Can you please take a look and see if it addresses your concern?

joergroedel: Yes, I just noticed that this is already fixed by one of your PRs, I left a review there.

kernel/src/task/exec.rs
joergroedel: Yes, your fixes handle that and are now included into this PR.

PR #526, opened 2024-11-21
kernel: remove unnecessary `unsafe`

msft-jlange:
Rust defines safety specifically with respect to memory guarantees enforced by the language (ownership, typing, bound, etc.) and specifically excludes overall program correctness (including data races and deadlock).  Functions that are safe with respect to the language definition of "safe" should not be declared `unsafe`.

### DIFF
--- a/kernel/src/cpu/irq_state.rs
+++ b/kernel/src/cpu/irq_state.rs
@@ -15,11 +15,9 @@ const EFLAGS_IF: u64 = 1 << 9;
 
 /// Unconditionally disable IRQs
 ///
-/// # Safety
-///
 /// Callers need to take care of re-enabling IRQs.
 #[inline(always)]
-pub unsafe fn raw_irqs_disable() {
+pub fn raw_irqs_disable() {
     unsafe {
         asm!("cli", options(att_syntax, preserves_flags, nomem));
     }
@@ -27,13 +25,11 @@ pub unsafe fn raw_irqs_disable() {
 
 /// Unconditionally enable IRQs
 ///
-/// # Safety
-///
 /// Callers need to make sure it is safe to enable IRQs. e.g. that no data
 /// structures or locks which are accessed in IRQ handlers are used after IRQs
 /// have been enabled.
 #[inline(always)]
-pub unsafe fn raw_irqs_enable() {
+pub fn raw_irqs_enable() {
     unsafe {
         asm!("sti", options(att_syntax, preserves_flags, nomem));
     }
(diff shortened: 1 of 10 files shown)


kernel/src/cpu/irq_state.rs
```diff
@@ -15,25 +15,19 @@ const EFLAGS_IF: u64 = 1 << 9;
 
 /// Unconditionally disable IRQs
 ///
-/// # Safety
-///
 /// Callers need to take care of re-enabling IRQs.
-#[inline(always)]
```
stefano-garzarella, line 21: Is removing `#[inline(always)]` intentional?

Not sure if we really need it or not, just asking ;-)

kernel/src/cpu/irq_state.rs
```diff
@@ -15,25 +15,19 @@ const EFLAGS_IF: u64 = 1 << 9;
...
 }
 
 /// Unconditionally enable IRQs
 ///
-/// # Safety
-///
 /// Callers need to make sure it is safe to enable IRQs. e.g. that no data
 /// structures or locks which are accessed in IRQ handlers are used after IRQs
 /// have been enabled.
-#[inline(always)]
```
stefano-garzarella, line 35: Ditto.

kernel/src/stage2.rs
```diff
@@ -455,9 +455,7 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) {
     // Shut down the GHCB
     unsafe {
         shutdown_percpu();
-    }
```
stefano-garzarella, line 458: I am not sure if this is a good practice, there is also a lint, to eventually enable, to discourage this:
https://rust-lang.github.io/rust-clippy/master/index.html#multiple_unsafe_ops_per_block

That said, I don't have a strong opinion since we already do this in several places, but maybe in the future we could avoid it and enable that lint.

msft-jlange: The reason `shutdown_percpu()` is unsafe is because the per-CPU can no longer be safely used.  So I put the call to `shutdown_percpu()` in the same `unsafe` block as the asm code that guarantees that we will no longer attempt to use the per-CPU to emphasize the coupling between these two concepts.  The second `unsafe` is what makes use of the first `unsafe` correct.

I'm not certain we will ever enable this lint as I'm not sure we will conclude that it improves the safety awareness within the code.  We can always change it if and when we make such a decision.

stefano-garzarella (approved):

PR #531, opened 2024-11-22
Add API to manage global mappings

joergroedel:
This PR implements a simple API to create global mappings of physical addresses and to manage their life-times.

The implementation is temporary and includes some code duplication with the existing page-table guard implementation. This is intentional as global mappings will be re-implemented using the VM code once the global part of the COCONUT address space is managed by a `VMR` struct. Therefore no time was spent making the existing guards more generic.

### DIFF
--- a/kernel/src/mm/address_space.rs
+++ b/kernel/src/mm/address_space.rs
@@ -141,14 +141,16 @@ const fn virt_from_idx(idx: usize) -> VirtAddr {
 pub const PGTABLE_LVL3_IDX_SHARED: usize = 511;
 
 /// Base Address of shared memory region
-pub const SVSM_SHARED_BASE: VirtAddr = virt_from_idx(PGTABLE_LVL3_IDX_SHARED);
+pub const SVSM_GLOBAL_BASE: VirtAddr = virt_from_idx(PGTABLE_LVL3_IDX_SHARED);
 
-/// Mapping range for shared stacks
-pub const SVSM_SHARED_STACK_BASE: VirtAddr = SVSM_SHARED_BASE.const_add(256 * SIZE_1G);
-pub const SVSM_SHARED_STACK_END: VirtAddr = SVSM_SHARED_STACK_BASE.const_add(SIZE_1G);
+/// Shared mappings region start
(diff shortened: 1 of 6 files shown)


joergroedel:
Just noticed that `shared_memory.rs` is a bad name and will lead to confusion wrt. private/shared mappings. I will rename it to `global_memory.rs`.

osteffenrh:
Awesome! Thank you!

kernel/src/mm/address_space.rs
```diff
@@ -143,10 +143,6 @@ pub const PGTABLE_LVL3_IDX_SHARED: usize = 511;
 /// Base Address of shared memory region
 pub const SVSM_SHARED_BASE: VirtAddr = virt_from_idx(PGTABLE_LVL3_IDX_SHARED);
```
stefano-garzarella, line 144: Should we remove also this in the first commit, which is used only by `SVSM_SHARED_STACK_BASE`?

As we add them back in the next commit, do we really need a patch to remove them?

kernel/src/mm/address_space.rs
```diff
@@ -143,6 +143,12 @@ pub const PGTABLE_LVL3_IDX_SHARED: usize = 511;
...
 
+/// Shared mappings region start
+pub const SVSM_SHARED_MAPPING_BASE: VirtAddr = SVSM_SHARED_BASE.const_add(256 * SIZE_1G);
+
+/// Shared mappings region end
+pub const SVSM_SHARED_MAPPING_END: VirtAddr = SVSM_SHARED_MAPPING_BASE.const_add(SIZE_1G);
+
```
stefano-garzarella, line 151: Should we rename in `SVSM_GLOBAL_*`?

joergroedel: Yes, renamed them the SVSM_GLOBAL_*, thanks.

kernel/src/mm/address_space.rs
joergroedel: I merged the two commits into one.

stefano-garzarella (approved):

PR #534, opened 2024-11-25
cpu: assign a CPU index to each CPU

msft-jlange:
It is convenient to be able to look up each CPU object by an index value (contiguous, starting from zero) and not just by APIC ID.  This will become critical when it becomes necessary to build a compact representation of a set of CPUs that are targets for an IPI message.


joergroedel (approved):
Changes look good to me.

Issue #228, opened 2024-01-25
Document `unsafe` blocks


00xc:
We should document all `unsafe` blocks in the codebase, explaining the rationale for why using the keyword is safe (i.e. `// SAFETY:` comments). This will allow us to enable the `undocumented_unsafe_blocks` clippy lint (e.g. `cargo clippy -- -D clippy::undocumented_unsafe_blocks`), and emit an error on new undocumented unsafe blocks.

stefano-garzarella:
I started to look into that, but we have a looooot of `unsafe` blocks without any comment.
It will take time, but in the meantime we should enforce new PRs to carry `SAFETY` comments. Not sure if we should add some scripts in the CI for now, till we can enable `undocumented_unsafe_blocks` clippy lint.

joergroedel:
I think we have to rely on reviews for making sure all new `unsafe` sections get a `SAFETY` comment.

stefano-garzarella:
> I think we have to rely on reviews for making sure all new `unsafe` sections get a `SAFETY` comment.

Yes, but we are human, and it can be missed, so I added a new pipeline for that, see https://github.com/coconut-svsm/svsm/pull/535

For example, we have recently added several with the work for user space (I will send another PR tomorrow to fix some of them, but I need help with some that I don't know how to document).

vijaydhanraj:
Hi @stefano-garzarella, I am able to repro missing `unsafe` block comment with recently introduced userspace PRs using below changes. Please let me know if you want me to cook up a patch for fixing it. 
```
diff --git a/Makefile b/Makefile
index 0916f1e1..26874350 100644
--- a/Makefile
+++ b/Makefile
@@ -175,10 +175,10 @@ bin/svsm-test.bin: bin/stage1-test
        objcopy -O binary $< $@
 
 clippy:
-       cargo clippy --workspace --all-features --exclude packit --exclude svsm-fuzz --exclude igvmbuilder --exclude igvmmeasure -- -D warnings
-       cargo clippy --workspace --all-features --exclude packit --exclude svsm-fuzz --exclude svsm --target=x86_64-unknown-linux-gnu -- -D warnings
-       RUSTFLAGS="--cfg fuzzing" cargo clippy --package svsm-fuzz --all-features --target=x86_64-unknown-linux-gnu -- -D warnings
-       cargo clippy --workspace --all-features --exclude packit --tests --target=x86_64-unknown-linux-gnu -- -D warnings
[...]

PR #535, opened 2024-11-25
CI: check new undocumented unsafe blocks

stefano-garzarella:
This is to prevent undocumented unsafe blocks from growing before we
add comments for all of them and manage to enable
`clippy::undocumented_unsafe_blocks` lint.
    
Progress documented at https://github.com/coconut-svsm/svsm/issues/228.
When we fix that issue, we may remove this pipeline.

~NOTE: the second patch it's just to test the new pipeline, I'll remove it if we want to merge this PR.~
I removed that patch, you can check the behavior with a new undocumented unsafe block in https://github.com/stefano-garzarella/svsm/pull/4

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -112,3 +112,75 @@ jobs:
         with:
           command: fmt
           args: --all -- --check --config "format_code_in_doc_comments=true"
+
+  # Check for new undocumented unsafe blocks. This is to prevent them from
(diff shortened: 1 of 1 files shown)


stefano-garzarella:
I removed that second patch, you can check the behavior with a new undocumented unsafe block in https://github.com/stefano-garzarella/svsm/pull/4

PR #536, opened 2024-11-26
cpu/idt: Handle interrupt nesting correctly

msft-jlange:
Exception handlers (other than IST-based exceptions) should run with interrupts enabled if interrupts were enabled at the time the exception was raised.  This is necessary to ensure that the `IrqState` nesting management is not confused by nesting that might occur within an exception handler.  In addition, most exceptions can safely tolerate interrupts, and need not impact system stability by suppressing interrupt delivery (this is particularly true for #VC exceptions, which can take a long time to execute due to emulation in the host).

IST-based exceptions must keep interrupts disabled as long as they execute on the IST stack (this currently applies only to #DF, which is fatal anyway).

Interrupt handlers must run with interrupts disabled, but need to update the `IrqState` in case they call into common code which attempts to disable interrupts.

The #HV handler must also run with interrupts disabled, and similarly needs to manage the `IrqState`.

### DIFF
--- a/kernel/src/cpu/idt/entry.S
+++ b/kernel/src/cpu/idt/entry.S
@@ -37,7 +37,7 @@ HV_DOORBELL_ADDR:
 	pushq	%rax
 .endm
 
-.macro default_entry_no_ist name: req handler:req error_code:req vector:req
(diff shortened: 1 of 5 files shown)


joergroedel (approved):
Looks good to me.

PR #537, opened 2024-11-26
syscall: add safety comments

### LINKED ISSUE #228: Document `unsafe` blocks

stefano-garzarella:
This is part of the effort to resolve #228 and enable `undocumented_unsafe_blocks` clippy lint.

Some patches have FIXME comments because I need help figuring out whether to mark the whole function unsafe.

The last commit is just a cleanup of a comment I noticed while adding the SAFETY comments.

I will open more PRs later with more SAFETY comments, I don't think it makes sense to open a giant PR with more than 200 comments, it would be hard to review.


kernel/src/cpu/control_regs.rs
```diff
@@ -106,6 +108,9 @@ pub fn read_cr0() -> CR0Flags {
 pub fn write_cr0(cr0: CR0Flags) {
     let reg = cr0.bits();
 
+    // SAFETY: The inline assembly set the processors CR0 register with flags
+    // defined by `struct CR0Flags`.
+    // FIXME: is this enough?
```
stefano-garzarella, line 113: Any hints on this?

kernel/src/cpu/control_regs.rs
```diff
@@ -124,6 +132,10 @@ pub fn read_cr2() -> usize {
 }
 
 pub fn write_cr2(cr2: usize) {
+    // SAFETY: The inline assembly set the processors CR2 register.
+    // FIXME: should this function be marked unsafe?
```
stefano-garzarella, line 136: Any hints on this?

kernel/src/cpu/idt/common.rs
```diff
@@ -74,9 +74,12 @@ impl X86ExceptionContext {
         self.frame.rip = new_rip;
 
         if is_cet_ss_supported() {
-            // Update the instruction pointer on the shadow stack.
             let return_on_stack = (self.ssp + 8) as *const usize;
             let return_on_stack_val = new_rip;
+            // SAFETY: Inline assembly to update the instruction pointer on
+            // the shadow stack.
+            // FIXME: should we mark the entire set_rip function as unsafe?
```
stefano-garzarella, line 81: Any hints on this?

kernel/src/cpu/apic.rs
```diff
@@ -244,6 +244,9 @@ impl LocalApic {
         // (core_remap_ca(), core_create_vcpu() or prepare_fw_launch()) so
         // they're safe to use.
         if let Ok(caa) = unsafe { calling_area.read() } {
+            // SAFETY: guest vmsa and ca are always validated before beeing updated
+            // (core_remap_ca(), core_create_vcpu() or prepare_fw_launch()) so
+            // they're safe to use.
             let _ = unsafe { calling_area.write(caa.update_no_eoi_required(0)) };
```
joergroedel, line 250: The `calling_area` is actually a case where the unsafe behavior can be hidden in the write functions. The calling area itself is volatile and not covered by safety guarantees at runtime or by the compiler. I think the read/write methods should be exposed as safe.

kernel/src/cpu/apic.rs
```diff
@@ -370,6 +373,9 @@ impl LocalApic {
                         // (core_remap_ca(), core_create_vcpu() or prepare_fw_launch())
                         // so they're safe to use.
                         if let Ok(caa) = unsafe { calling_area.read() } {
+                            // SAFETY: guest vmsa and ca are always validated before beeing upated
+                            // (core_remap_ca(), core_create_vcpu() or prepare_fw_launch())
+                            // so they're safe to use.
                             if unsafe { calling_area.write(caa.update_no_eoi_required(1)).is_ok() }
```
joergroedel, line 379: Same applies here.

kernel/src/cpu/control_regs.rs
joergroedel: The caller needs to make sure to not change any execution-state relevant bits like PE or PG.

kernel/src/cpu/control_regs.rs
joergroedel: Changing CR2 can impact the execution flow of the #PF handler. Since there is no user for this function, the safest is to just remove it. If there is a reason to write CR2 the code needs to make sure to restore the old value afterwards, which essentially needs a guard implementation.

kernel/src/cpu/control_regs.rs
```diff
@@ -142,6 +157,10 @@ pub fn read_cr3() -> PhysAddr {
 }
 
 pub fn write_cr3(cr3: PhysAddr) {
+    // SAFETY: The inline assembly set the processors CR3 register.
+    // FIXME: should this function be marked unsafe?
+    //        could change the state of the system, can that compromise
+    //        the memory safety?
```
joergroedel, line 163: Yes, changing the page-table is one of the core behavior which can impact memory safety. The SAFETY comment should state that writing CR3 needs to be accompanied by other actions to make sure a memory safe execution state is warranted (like changing the stack and register state).

kernel/src/cpu/control_regs.rs
```diff
@@ -191,6 +212,9 @@ pub fn read_cr4() -> CR4Flags {
 pub fn write_cr4(cr4: CR4Flags) {
     let reg = cr4.bits();
 
+    // SAFETY: The inline assembly set the processors CR4 register with flags
+    // defined by `struct CR4Flags`.
+    // FIXME: is this enough?
```
joergroedel, line 217: Same comment as for `CR0` applies here as well. Setting CR4 needs to make sure to not alter any execution state affecting memory safety (e.g. PSE or PAE).

kernel/src/cpu/cpuid.rs
```diff
@@ -55,6 +55,9 @@ impl CpuidResult {
         let mut result_ebx: u32;
         let mut result_ecx: u32;
         let mut result_edx: u32;
+        // SAFETY: Inline assembly to execute the CPUID instruction.
+        // Input registers (EAX, ECX) and output registers (EAX, EBX, ECX, EDX)
+        // are safely managed.
```
joergroedel, line 60: I would add that the CPUID instruction does not alter or affect any memory state.

kernel/src/cpu/idt/common.rs
joergroedel: Yeah, this is tricky. For `set_rip` the requirement is to not bypass compiler guarantees by changing the executions flow. For advancing the RIP in inline-assembly parts this is given, for instructions generated by the compiler the caller needs to make sure to update the rest of the execution state as actual hardware would have done it (e.g. for MMIO emulation/CPUID/MSR...).

kernel/src/cpu/idt/common.rs
```diff
@@ -378,6 +383,7 @@ pub fn triple_fault() {
         address: VirtAddr::from(0u64),
     };
 
+    // SAFETY: Inline assembly to load an invalid IDT to trigger a triple fault.
```
joergroedel, line 386: Better comment: This ends execution, this function will not return so memory safety is not an issue.

kernel/src/cpu/idt/svsm.rs
```diff
@@ -228,15 +229,21 @@ extern "C" fn ex_handler_control_protection(ctxt: &mut X86ExceptionContext, _vec
...
             let prev_rssp_ptr: GuestPtr<u64> = GuestPtr::new(VirtAddr::from(ctxt.ssp));
+            // SAFETY: `ssp` should be a valid guest address filled by the CPU
+            // in the X86ExceptionContext
             let prev_rssp = unsafe { prev_rssp_ptr.read() }
                 .expect("Failed to read address of previous shadow stack pointer");
             // The offset to the return pointer is different for RET and IRET.
             let offset = if code == NEAR_RET { 0 } else { 8 };
             let ret_ptr: GuestPtr<u64> = GuestPtr::new(VirtAddr::from(prev_rssp + offset));
+            // SAFETY: `ssp` should be a valid guest address filled by the CPU
+            // in the X86ExceptionContext
```
joergroedel, line 246: `s/should be/is/`

joergroedel (commented):
I left some comments on the safety of certain code parts.

kernel/src/cpu/control_regs.rs
stefano-garzarella: mmm, so should `write_cr0` be unsafe if we have those reqs on the caller?

kernel/src/cpu/control_regs.rs
stefano-garzarella: yeah, I agree, I'll remove the entire function!

kernel/src/cpu/control_regs.rs
stefano-garzarella: > The SAFETY comment should state that writing CR3 needs to be accompanied by other actions to make sure a memory safe execution state is warranted (like changing the stack and register state).

but that must be demanded to the caller, no?
So, should we mark the entire function unsafe?

kernel/src/cpu/control_regs.rs
joergroedel: Yes, this function needs to be unsafe.

kernel/src/cpu/control_regs.rs
joergroedel: Yes, that are caller-requirements and the function needs to be unsafe.

kernel/src/cpu/idt/svsm.rs
stefano-garzarella: done, thanks!

kernel/src/cpu/apic.rs
stefano-garzarella: So, IIUC you are suggesting to change `GuestPtr::write/read` in safe functions, right?
Should we add some more checks there?

joergroedel: No, `GuestPtr::read/write` need to be unsafe, because they can be used to write to SVSM memory as well, there are no checks to prevent that. For the `calling_area` this is different, as it is checked to not be in SVSM memory when set up, which makes use of `GuestPtr` safe.

kernel/src/cpu/control_regs.rs
stefano-garzarella: The caller is `PageTable::load()`, not sure if mark also it is unsafe or add a SAFETY comment. 
I added a FIXME in the code, please take a look.

stefano-garzarella:
v2:
- addressed most of @joergroedel comments
- TODO:
  - [ ] FIXME: should we mark PageTable::load() function unsafe?
  - [ ] calling_area comments

kernel/src/cpu/apic.rs
stefano-garzarella: yeah, that was also my concern, but now I'm confused, what are you suggesting here?

Creating a new struct for the calling area, wrapping the `GuestPtr`?

stefano-garzarella:
v3:
- moved some unsafe comments related to kernel/cpu to a new PR https://github.com/coconut-svsm/svsm/pull/550
- - TODO:
  - [ ] FIXME: should we mark PageTable::load() function unsafe? @joergroedel 
  - [ ] calling_area comments @joergroedel

kernel/src/mm/pagetable.rs
```diff
@@ -497,7 +497,12 @@ pub struct PageTable {
 impl PageTable {
     /// Load the current page table into the CR3 register.
     pub fn load(&self) {
```
p4zuu, line 499: I tend to say that we should set `PageTable::load()` unsafe. Loading an invalid page table can break memory safety. We should ensure this call is safe by forcing the caller to check that the new page table is correctly allocated, with the correct lifetime, etc.

kernel/src/cpu/apic.rs
joergroedel: No, a new wrapping struct is not required. The calling area is safe because it is always initialized with a pointer guaranteed to point to guest memory. So SVSM memory safety can not be harmed by accessing it.

kernel/src/mm/pagetable.rs
joergroedel: Yes, by the previous reasoning `PageTable::load()` needs also be marked as unsafe. The call-sites need safety comments explaining why it is safe to load the new page-table.

stefano-garzarella: Agree, I'll probably do in another PR because that function is wrapped by other, so not that easy. WDYT?

kernel/src/cpu/apic.rs
joergroedel: Ah, I was confused, sorry. I though the `calling_area.read/write` where specific to the calling area, but they are actually from `GuestPtr`. So the best here is to add safety comments stating that `caa` has been validated to point to guest memory.

In fact, the **proper** solution is to abstract the unsafe behavior away into accessors to SvsmCaa and pass an instance of this structure around.

joergroedel: > In fact, the **proper** solution is to abstract the unsafe behavior away into accessors to SvsmCaa and pass an instance of this structure around.

But that change is of course no merge requirement for this PR.

stefano-garzarella: > Ah, I was confused, sorry. I though the `calling_area.read/write` where specific to the calling area, but they are actually from `GuestPtr`. So the best here is to add safety comments stating that `caa` has been validated to point to guest memory.

Okay, now it's clear. I already added a SAFETY comment using the one some lines before, but yeah I can simplify that.

stefano-garzarella: BTW, instead of adding a new comment, which is similar to the comment a few lines above, which already says guest ca is always validated, I'm adding a single unsafe block because we're talking about the same things and it seems redundant.

kernel/src/cpu/apic.rs
stefano-garzarella: I'll do the same, a single unsafe block re-using the SAFETY comment we already have.

kernel/src/cpu/control_regs.rs
stefano-garzarella: This requires so many changes to dependencies, I move it to another PR.

kernel/src/mm/pagetable.rs
stefano-garzarella: I can do in this PR, it is not that hard I guess.

stefano-garzarella:
v3:
- added a new patch to mark `PageTable::load()` unsafe and also add SAFETY comments where it's used @p4zuu 
- fixed calling area safety comments @joergroedel

stefano-garzarella:
v4:
- fixed a `cargo fmt` warning
- added an unsafe block in gdbstub

kernel/src/debug/gdbstub.rs
```diff
@@ -191,7 +191,9 @@ pub mod svsm_gdbstub {
                 let cr3 = read_cr3();
                 let task = tl.get_task(id);
                 if let Some(task) = task {
-                    task.page_table.lock().load();
+                    unsafe {
```
p4zuu, line 194: Is this intended that clippy didn't catch the missing safety comment here?

stefano-garzarella: yeah, I mean clippy catched that, but our CI just counts them before and after applying the PR. Since I'm documented more than the ones I introduced, CI didn't fail. When we will enable clippy lint, it will be caught.

stefano-garzarella: BTW I can fix that, but any suggestion of SAFETY comment to put there?
@p4zuu @joergroedel

joergroedel: I think @roy-hopkins can help here. It can also be fixed in a later PR, no reason to block this one on the gdbstub comments.

stefano-garzarella: @joergroedel yep, I agree on doing in another PR where we will address all the SAFETY comments for the gdbstub.

I'll rebase this PR, we ca merge it.

stefano-garzarella:
v5:
- rebased on main
- removed patch `kernel/apic: fix safety comments` since that file was already fixed by another PR just merged.

PR #538, opened 2024-11-27
platform/snp: Add abstractions for FW operations

peterfang:
Add new abstractions to SvsmPlatform:

- prepare_fw() performs preparations for launching guest boot firmware
- launch_fw() launches guest boot firmware

Move FW-related functions in svsm.rs to snp_fw.rs as they're SEV-specific. Make sure snp_fw doesn't export anything unnecessary to the outside world.


kernel/src/cpu/cpuid.rs
```diff
@@ -4,19 +4,44 @@
...
+///
+/// The caller should verify that `dst` points to mapped memory whose size is
+/// at least 4K. We assert above at compile time that SnpCpuidTable fits within
+/// a page, so the write is safe.
+///
+/// The caller should verify not to corrupt arbitrary memory, as this function
+/// doesn't make any checks in that regard.
+pub unsafe fn copy_cpuid_table_to(dst: VirtAddr) {
+    let start = dst.as_mut_ptr::<u8>();
+    unsafe {
```
stefano-garzarella, line 36: Please add a `SAFETY` comment here. I know it's redundant because the function itself is unsafe, but if we enable  `undocumented_unsafe_blocks` clippy lint we will have a warning, see https://github.com/coconut-svsm/svsm/pull/537

kernel/src/platform/snp.rs
```diff
@@ -116,6 +121,30 @@ impl SvsmPlatform for SnpPlatform {
...
+        config: &SvsmConfig<'_>,
+        kernel_region: MemoryRegion<PhysAddr>,
+    ) -> Result<(), SvsmError> {
+        if let Some(fw_meta) = &config.get_fw_metadata() {
+            print_fw_meta(fw_meta);
+            write_guest_memory_map(config).expect("Failed to write guest memory map");
+            validate_fw_memory(config, fw_meta, &kernel_region).expect("Failed to validate memory");
+            copy_tables_to_fw(fw_meta, &kernel_region).expect("Failed to copy firmware tables");
+            validate_fw(config, &kernel_region).expect("Failed to validate flash memory");
+            prepare_fw_launch(fw_meta).expect("Failed to setup guest VMSA/CAA");
```
stefano-garzarella, line 135: I see that this is pre-existing, but since now this function return an `SvsmError`, can we avoid the `expect()` here and propagate the error to the caller?

peterfang: That's a good idea. I'll make that change.

stefano-garzarella (approved):

PR #542, opened 2024-11-28
virtio-blk storage via mmio using virtio-drivers [v2]

osteffenrh:
This is version 2 of https://github.com/coconut-svsm/svsm/pull/343.

Add support for virtio-blk storage devices using the virtio-mmio transport.

The [virtio-drivers](https://github.com/rcore-os/virtio-drivers/) crate from the rcore-os project is used. It required some modifications: https://github.com/osteffenrh/virtio-drivers/pull/2
Crate License: MIT.

This PR requires a patched version of Qemu (https://github.com/coconut-svsm/qemu/pull/13), adding virtio-mmio slots to the Q35 machine model.

This build on https://github.com/coconut-svsm/svsm/pull/531

How to test:
- Use Qemu from https://github.com/coconut-svsm/qemu/pull/13
- run the in-svsm tests `make test-in-svsm`

ToDo:
- Write block size / buffer size is limited to max 4k, needs improvement of bounce buffer handling
- Proper error codes for the block layer and driver impl.
- Supply the MMIO base address via IGVM file

Fixes since V1:
- SMP works, mmio range is mapped for all cpus
- rebased to latest versions of Coconut and virtio-drivers

Things for future PRs:
- encryption layer
- file system

CC: @stefano-garzarella ;-)

### DIFF
--- a/kernel/src/block/api.rs
+++ b/kernel/src/block/api.rs
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2024 Red Hat, Inc.
+//
+// Author: Stefano Garzarella <sgarzare@redhat.com>
+// Author: Oliver Steffen <osteffen@redhat.com>
+
+#[derive(Debug)]
+pub enum BlockDeviceError {
+    Failed, // ToDo: insert proper errors
+}
+
+pub trait BlockDriver {
+    fn read_blocks(&self, block_id: usize, buf: &mut [u8]) -> Result<(), BlockDeviceError>;
+    fn write_blocks(&self, block_id: usize, buf: &[u8]) -> Result<(), BlockDeviceError>;
+    fn block_size_log2(&self) -> u8;
+    fn size(&self) -> usize;
+}
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -326,6 +326,12 @@ pub extern "C" fn svsm_main() {
         panic!("Failed to launch FW: {e:#?}");
     }
 
+    {
+        use svsm::block::virtio_blk;
+        static MMIO_BASE: u64 = 0xfef03000;
+        let _blk = virtio_blk::VirtIOBlkDriver::new(PhysAddr::from(MMIO_BASE));
+    }
+
     start_kernel_task(request_processing_main, String::from("request-processing"))
         .expect("Failed to launch request processing task");
 
--- a/kernel/src/virtio/hal.rs
+++ b/kernel/src/virtio/hal.rs
@@ -1,0 +1,227 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2024 Red Hat, Inc.
+//
+// Author: Oliver Steffen <osteffen@redhat.com>
+
+extern crate alloc;
+use crate::locking::SpinLock;
+use alloc::vec::Vec;
+use core::{
+    cell::OnceCell,
+    ptr::{addr_of, NonNull},
+};
+use zerocopy::{FromBytes, Immutable, IntoBytes};
+
+use crate::{
+    address::{PhysAddr, VirtAddr},
+    cpu::{self, percpu::this_cpu},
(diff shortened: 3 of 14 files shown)


kernel/src/block/api.rs
```diff
@@ -0,0 +1,18 @@
...
+// Author: Oliver Steffen <osteffen@redhat.com>
+
+#[derive(Debug)]
+pub enum BlockDeviceError {
+    Failed, // ToDo: insert proper errors
+}
+
+pub trait BlockDriver {
+    fn read_blocks(&self, block_id: usize, buf: &mut [u8]) -> Result<(), BlockDeviceError>;
+    fn write_blocks(&self, block_id: usize, buf: &[u8]) -> Result<(), BlockDeviceError>;
```
joergroedel, line 15: This interface makes the safe-unsafe memory boundary a bit blurry and looks unsafe in itself (due to the slice parameters). I have seen this interface somewhat resembles the interface of the virtio-blk driver crate. Maybe it is the best strategy to import the drivers into our code-base and improve their interfaces.

Some things I'd like to see:
* Asynchronous interface with a separate request tracking data structure and polling methods.
* Better abstractions for the source/target memory regions of block access.

kernel/src/svsm.rs
```diff
@@ -319,6 +319,12 @@ pub extern "C" fn svsm_main() {
         panic!("Failed to launch FW: {e:#?}");
     }
 
+    {
+        use svsm::block::virtio_blk;
+        static MMIO_BASE: u64 = 0xfef03000;
+        let _blk = virtio_blk::VirtIOBlkDriver::new(PhysAddr::from(MMIO_BASE));
+    }
```
joergroedel, line 326: We should start thinking about a proper detection mechanism for SVSM-assigned devices.

kernel/src/virtio/hal.rs
```diff
@@ -0,0 +1,227 @@
...
+
+use crate::{
+    address::{PhysAddr, VirtAddr},
+    cpu::{self, percpu::this_cpu},
+    mm::{page_visibility::*, *},
+};
+
+struct PageStore {
+    pages: Vec<(PhysAddr, SharedBox<[u8; PAGE_SIZE]>)>,
+}
```
joergroedel, line 24: Nothing wrong with this code, but I think this uncovers a fundamental problem with `SharedBox` which we need to solve separately. Reading any any memory in a `SharedBox` is UB, so it should have a `read()/write()` interface instead of direct data access.

kernel/src/svsm.rs
stefano-garzarella: @joergroedel yeah, should be some runtime discover mechanism (e.g. fw-cfg) or a static configuration through IGVM? Or both?

kraxel: The qemu patch adds four virtio-mmio transport slots at `0xfef00000 ... 0xfef03fff`.  So the very minimum would be to loop over these four slots, check whenever a virtio device is present, if so what kind of virtio device it is.  Only after that attach a driver.

Separate question is whenever we'll just hardcode those four slots in both qemu and svsm, or whenever it makes sense to communicate the range from qemu to svsm via fw_cfg.  I honestly don't see much benefit in making this configurable.  There are not that many options to place the virtio-mmio transport slots anyway ...

osteffenrh: How about build time option that puts it into the igvm file as a start.
And use the range @kraxel picked as the default on both sides (Coconut and Qemu).

kernel/src/block/api.rs
osteffenrh: **About async:**
The virtio-drivers crate has non-blocking read/write functions, see https://github.com/osteffenrh/virtio-drivers/blob/63c5450f08bd557f609e0ffad7a5bd6281d6562e/src/device/blk.rs#L248
But I did not try that yet.

**About the second point:**
The buffer supplied to these functions is copied over to a new one, allocated via `SharedBox` when writing (and the reverse for reading). We are not sharing existing allocations with the host.

osteffenrh:
Closing this. See v3: #635

PR #543, opened 2024-12-02
cpu: implement interior mutability in `X86Tss`

msft-jlange:
The only fields of the TSS that ever require modification are the IST fields and RSP0.  Since RSP0 has to be modified on every task switch, it is undesirable to store TSS in a `Cell` because that requires copying the entire TSS in and out of the cell in order to modify just the single field.  Instead, the TSS can be made immutable, and can implement interior mutability for its own fields.  However, `Cell` cannot be used directly in this case because the X86 architecture defines the TSS such that the stack pointers are unaligned, and therefore assembly has to be used to modify the fields.

### DIFF
--- a/kernel/src/cpu/tss.rs
+++ b/kernel/src/cpu/tss.rs
@@ -6,7 +6,9 @@
 
 use super::gdt::GDTEntry;
 use crate::address::VirtAddr;
+use core::arch::asm;
 use core::num::NonZeroU8;
+use core::ptr::addr_of;
 
 // IST offsets
 pub const IST_DF: NonZeroU8 = unsafe { NonZeroU8::new_unchecked(1) };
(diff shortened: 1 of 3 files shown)


kernel/src/cpu/tss.rs
```diff
@@ -38,10 +40,32 @@ impl X86Tss {
         }
     }
 
-    pub fn set_ist_stack(&mut self, index: NonZeroU8, addr: VirtAddr) {
+    pub fn set_ist_stack(&self, index: NonZeroU8, addr: VirtAddr) {
```
p4zuu, line 43: Would it make sense to set `set_ist_stack()` and `set_rsp0()` `unsafe`? Incorrectly called, I guess they can both create memory issues, thus UB.
It's not an issue with these changes of course, it's also relevant with `main` branch code

msft-jlange: I don't see how these are unsafe.  The behavior of this code is identical to the behavior that we could expect if the stack pointer fields were wrapped in `Cell`, and defining the fields that way would certainly not require these functions to be marked `unsafe`.  The only reason I don't use `Cell` here is that the compiler refuses to place `Cell` on an unaligned boundary because that is not safe across all architectures, but the TSS code only applies to x86, where unaligned accesses are always safe.

msft-jlange: Ah, I understand your point and I agree with it.

p4zuu (approved):

PR #539, opened 2024-11-27
Add user-mode binary to build system

joergroedel:
This PR imports the helper utility used for user-mode bring-up into the COCONUT-SVSM repository. The build script and the recipes are enhanced to compile the binary and package it into the IGVM files via a generated file-system image.

This PR also creates the directory structure for user-mode modules and introduces the joint library.

PR #547, opened 2024-12-04
Userspace support: Refactor init to launch a simple device model

vijaydhanraj:
This PR adds an initial version of `init` to launch a dummy device model.  The initial `dm` version just calls `exit` syscall without doing anything else.

Note: This PR is built on top of https://github.com/coconut-svsm/svsm/pull/520 and https://github.com/coconut-svsm/svsm/pull/539 PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,8 @@ members = [
     "user/lib",
     # Init user-space module
     "user/init",
+    # device model module
+    "user/dm",
     # Release version identifier
     "release",
     # Virtio Drivers (MMIO)
@@ -48,6 +50,7 @@ syscall = { path = "syscall" }
 packit = { path = "packit" }
 userlib = { path = "user/lib" }
 userinit = { path = "user/init" }
+userdm = { path = "user/dm" }
 release = { path = "release" }
 virtio-drivers = { path = "virtio-drivers" }
 
--- a/user/dm/Cargo.toml
+++ b/user/dm/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "userdm"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "userdm"
+path = "src/main.rs"
+test = false
+doctest = false
+
+[dependencies]
+userlib.workspace = true
+
+[lints]
+workspace = true
--- a/user/dm/src/main.rs
+++ b/user/dm/src/main.rs
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2024 Intel Corporation.
+//
+// Author: Vijay Dhanraj <vijay.dhanraj@intel.com>
+
+#![no_std]
+#![no_main]
+
+use userlib::*;
+
+declare_main!(main);
+
+fn main() -> u32 {
+    0
+}
--- a/user/init/src/main.rs
+++ b/user/init/src/main.rs
@@ -1,41 +1,49 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2024 SUSE LLC
+//
+// Author: Joerg Roedel <jroedel@suse.de>
+
 #![no_std]
 #![no_main]
 
+use core::ffi::CStr;
 use userlib::*;
 
-use core::ptr::{addr_of, addr_of_mut};
-
-static mut SOME_BSS_DATA: [u64; 128] = [0; 128];
-static mut SOME_DATA: [u64; 128] = [0x01; 128];
-static SOME_RO_DATA: [u64; 128] = [0xee; 128];
-
-fn check(arr: &[u64; 128], val: u64) {
-    for v in arr.iter() {
-        if *v != val {
-            panic!("Unexpected array value");
-        }
-    }
-}
-
-fn write(arr: &mut [u64; 128], val: u64) {
-    for v in arr.iter_mut() {
-        *v = val;
-    }
-}
-
 declare_main!(main);
-
 fn main() -> u32 {
     println!("COCONUT-SVSM init process starting");
 
-    // SAFETY: Single-threaded process, so no data races. Safe to access global
-    // mutable data.
-    unsafe {
-        write(&mut *addr_of_mut!(SOME_DATA), 0xcc);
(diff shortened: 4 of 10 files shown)


Cargo.toml
```diff
@@ -38,6 +40,7 @@ syscall = { path = "syscall" }
 packit = { path = "packit" }
 userlib = { path = "user/lib" }
 userinit = { path = "user/init" }
+dm = { path = "user/dm" }
```
joergroedel, line 43: Please prefix all user-module crates with `user`, so in this case `userdm`. This simplifies things a lot later on.

Makefile
```diff
@@ -92,7 +92,7 @@ bin/coconut-test-vanadium.igvm: $(IGVMBUILDER) $(IGVMMEASURE) bin/stage1-trampol
 	$(IGVMMEASURE) --check-kvm --native-zero $@ measure
 
 test:
-	cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --workspace --exclude=user* --target=x86_64-unknown-linux-gnu
+	cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --workspace --exclude=user* --exclude=dm --target=x86_64-unknown-linux-gnu
```
joergroedel, line 95: E.g. here, with the `user` prefix the `--exclude user*` part also catches the crate.

user/dm/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "dm"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+userlib.workspace = true
+syscall = { path = "../../syscall" }
```
joergroedel, line 8: There is already an implicit dependency to the syscall crate via `userlib`. In general `userlib` should re-export everything from the `syscall` crate which a module might need directly.

user/init/src/main.rs
```diff
@@ -1,39 +1,75 @@
...
+            .lock()
+            .init(core::ptr::addr_of!(HEAP) as usize, HEAP_SIZE);
     }
-}
 
-fn write(arr: &mut [u64; 128], val: u64) {
-    for v in arr.iter_mut() {
-        *v = val;
+    let bin = CString::new("/").unwrap();
+    let Ok(obj) = opendir(&bin) else {
```
joergroedel, line 35: Can this be just `opendir(b"/\0")`?

joergroedel (commented):
As a general note the init process should have a configuration file which lists the modules to launch. We should also start thinking about a feature detection mechanism for the SVSM so that `init` can make decisions based on the environment it is running in. But that is outside the scope of this PR.

Cargo.toml
vijaydhanraj: Yes, this makes sense. I have added `user` prefix.

Makefile
vijaydhanraj: Yes, this is much better.

user/dm/Cargo.toml
vijaydhanraj: Thanks for pointing out and totally agree. I had missed to take advantage of the `userlib`. I have also refactored to make use of `declare_main` macro.

user/dm/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[no_mangle]
+pub extern "C" fn _start() -> ! {
+    dm_exit();
+}
+
+#[panic_handler]
+fn panic(_info: &PanicInfo<'_>) -> ! {
+    dm_exit();
+}
```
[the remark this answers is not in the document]
vijaydhanraj: Yes, taking advantage of `userlib` simplifies the code a lot.

user/init/src/main.rs
vijaydhanraj: `opendir` expects a `&CStr` type but using a raw byte slice may cause a compilation error. Also, current implementation is extensible if in future we want to change path from `/`.

vijaydhanraj:
> As a general note the init process should have a configuration file which lists the modules to launch. We should also start thinking about a feature detection mechanism for the SVSM so that `init` can make decisions based on the environment it is running in. But that is outside the scope of this PR.

Yes, this makes sense. I can look into this.

vijaydhanraj:
Hi @joergroedel can you please take a look?

user/init/src/main.rs
joergroedel: Yes, I understand, but creating a `CStr` usually requires allocator support. I am wondering whether there is a simpler way.

user/init/src/main.rs
```diff
@@ -1,41 +1,71 @@
...
+fn main() -> u32 {
+    println!("COCONUT-SVSM init process starting");
+
+    // SAFETY: The `begin` and `size` values refer to a valid, properly aligned, and
+    // accessible memory region.
+    unsafe {
+        HEAP_ALLOCATOR
+            .lock()
+            .init(core::ptr::addr_of!(HEAP) as usize, HEAP_SIZE);
     }
```
joergroedel, line 33: Heap allocator initialization needs to happen in the `userlib` crate. When `main` starts the heap should be initialized already.

user/lib/module.lds
```diff
@@ -48,12 +48,19 @@ SECTIONS
...
 		. = ALIGN(16);
 		_ebss = .;
 	} :bss
+
+	/* Fallback dummy section if `.bss` is empty */
+	. = ALIGN(4096);
+	.bss_dummy (NOLOAD) : {
+		LONG(0);
+	} :bss
+
```
joergroedel, line 63: Can you elaborate on why this is needed?

user/init/src/main.rs
vijaydhanraj: Understood. However, if we aim to avoid using the heap allocator, this comment may not be applicable. Could you please clarify your preference?

peterfang:
Hi @joergroedel, I've updated this PR with some minor changes. `userinit` is now a bit cleaner. I'm not able to resolve the previous conversations so I'd have to rely on you for that. Please take a look. Thanks!

msft-jlange:
Converting to a draft PR pending discussion about whether this design is consistent with the desired device management architecture.

PR #549, opened 2024-12-04
kernel: add additional safety comments

msft-jlange:
This PR adds additional missing safety comments to describe `unsafe` blocks.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -202,9 +202,9 @@ impl LocalApic {
         if self.lazy_eoi_pending {
             if let Some(virt_addr) = caa_addr {
                 let calling_area = GuestPtr::<SvsmCaa>::new(virt_addr);
-                // SAFETY: guest vmsa and ca are always validated before beeing updated
-                // (core_remap_ca(), core_create_vcpu() or prepare_fw_launch())
-                // so they're safe to use.
+                // SAFETY: guest vmsa and ca are always validated before being
+                // updated (core_remap_ca(), core_create_vcpu() or
+                // prepare_fw_launch()) so they're safe to use.
                 if let Ok(caa) = unsafe { calling_area.read() } {
                     if caa.no_eoi_required == 0 {
                         assert!(self.isr_stack_index != 0);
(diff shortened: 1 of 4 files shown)


kernel/src/cpu/apic.rs
```diff
@@ -202,9 +202,9 @@ impl LocalApic {
         if self.lazy_eoi_pending {
             if let Some(virt_addr) = caa_addr {
                 let calling_area = GuestPtr::<SvsmCaa>::new(virt_addr);
-                // SAFETY: guest vmsa and ca are always validated before beeing updated
-                // (core_remap_ca(), core_create_vcpu() or prepare_fw_launch())
-                // so they're safe to use.
+                // SAFETY: guest vmsa and ca are always validated before being
```
stefano-garzarella, line 205: Okay, I also touched that comment in https://github.com/coconut-svsm/svsm/pull/537 , but I agree that a single unsafe block is better.
BTW @joergroedel had a few comments, can you take a look too at https://github.com/coconut-svsm/svsm/pull/537#discussion_r1860456553 ?

kernel/src/platform/tdp.rs
```diff
@@ -110,6 +110,7 @@ impl SvsmPlatform for TdpPlatform {
             return Err(SvsmError::InvalidAddress);
         }
         match op {
+            // SAFETY: safety work on the address is yet to be completed.
```
stefano-garzarella, line 113: If it's still work in progress and safety is not guaranteed, would it be better to avoid the safety comment to remind us to check this better?

kernel/src/cpu/apic.rs
msft-jlange: I haven't thought much about this, though I am inclined to agree with Joerg's position that calling area accesses guarantee safety on their own.  I was imply fixing a typo in the comment that was already here.

kernel/src/platform/tdp.rs
msft-jlange: We need to enable the `undocumented_unsafe_blocks` lint as soon as possible to avoid future regression, so I think it's better to insert the safety comment now even though there is a `TODO` right next to it.  I'm not worried about the underlying safety issue being forgotten about.

stefano-garzarella (approved):

PR #554, opened 2024-12-05
kernel/pagetable: remove duplicate page table load

p4zuu:
Init page table is loaded twice, in the end of `init_page_table()` and just after in `svsm_start()`. We can remove the first one.


joergroedel (approved):
LGTM, thanks.

PR #555, opened 2024-12-05
cpu: use temporary GDT structures to load TSS

msft-jlange:
Configuring TSS on the current CPU requires a valid GDT entry, but the GDT entry only needs to valid long enough to execute the LTR instruction.  Modifying the global GDT for this purpose is undesirable, even if safety is guaranteed by wrapping the global GDT with a lock.  It is more straightforward to keep the global GDT constant, and to create a temporary writable GDT on each processor to use long enough to execute LTR and configure the TSS.

### DIFF
--- a/kernel/src/cpu/gdt.rs
+++ b/kernel/src/cpu/gdt.rs
@@ -35,11 +35,11 @@ impl GDTEntry {
     }
 
     pub const fn code_64_kernel() -> Self {
-        Self(0x00af9a000000ffffu64)
+        Self(0x00af9b000000ffffu64)
     }
 
     pub const fn data_64_kernel() -> Self {
-        Self(0x00cf92000000ffffu64)
+        Self(0x00cf93000000ffffu64)
     }
 
     pub const fn code_64_user() -> Self {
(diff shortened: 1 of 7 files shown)


msft-jlange:
> Hmm, this fails for me when switching from stage2 to the SVSM kernel:

Interesting, this worked for me when I wrote it but it also fails for me when I rebase to current.  I'll see what I can figure out.

msft-jlange:
> Interesting, this worked for me when I wrote it but it also fails for me when I rebase to current. I'll see what I can figure out.

There were a couple of problems here.  The old code caused a load of all segment registers (including CS) when the global GDT was constructed, and the new code skipped that step.  That shouldn't be a problem, except for a couple of issues that compound.  On SNP, we take a #VC fairly early due to executing `cpuid` (to calculate the XSAVE size during creation of the idle task), and the IDT dispatch causes a load of CS - and this becomes the first load of CS.  But by the time we execute the IDT dispatch, the global GDT has become read-only.  The compounding issue is that the GDT constructor is not consistent about setting accessed bits in GDT entries, and so the first load of the code selector requires writing to the GDT to set the accessed bit.  If the GDT is read-only, then the accessed bit cannot be set, so IDT dispatch raises #PF - which has the same problem, as does #DF, so we go to shutdown.

One important fix is to make sure the GDT constructor sets the accessed bit for all selectors, and the other is to make sure we bring back the missing load of selectors that occurs when the global GDT is established.

PR #559, opened 2024-12-06
cpu: Handle interrupt nesting correctly

msft-jlange:
This restores the logic to enable interrupts while handling exceptions, and also ensures that user mode runs with interrupts disabled when required.


joergroedel:
There was a small race-condition on my side, I already merged a fix for the user-if problem prior to seeing this PR. Sorry for that, but I cherry-picked the commit fixing the irq-state tracking into the `user-mode` PR and merged it via that path. I am closing this PR now.

PR #562, opened 2024-12-10
string: Deprecate FixedString

peterfang:
alloc::string::String is able to provide all of the functionalities of FixedString, with the added benefit of being part of the Rust core allocation library and a smaller memory footprint [1]. It also gives us native UTF-8 support which is important when importing file system images. Replace all instances of FixedString with String and be mindful of UTF-8 character boundaries when string truncation occurs.

[1] https://doc.rust-lang.org/alloc/string/struct.String.html#utf-8


kernel/src/acpi/tables.rs
```diff
@@ -182,8 +183,8 @@ impl ACPITable {
     ///
     /// This method returns the 4-character signature of the ACPI table, such as "APIC."
     #[expect(dead_code)]
-    fn signature(&self) -> FixedString<4> {
-        FixedString::from(self.header.sig)
+    fn signature(&self) -> String {
+        String::from_utf8(Vec::from(&self.header.sig)).unwrap()
```
p4zuu, line 187: Since this is dead code, I think we can afford changing the signature:
```suggestion
    fn signature(&self) -> Result<String, FromUtf8Error> {
        String::from_utf8(Vec::from(&self.header.sig))
```

kernel/src/syscall/class1.rs
```diff
@@ -36,8 +38,17 @@ pub fn sys_readdir(obj_id: u32, dirents: usize, size: usize) -> Result<u64, SysC
...
-        new_entry.file_name[..fname.len()].copy_from_slice(&fname);
+        // DirEnt.file_name must be nul-terminated
+        let bound = (0..min(name.len() + 1, new_entry.file_name.len()))
+            .rposition(|i| name.is_char_boundary(i))
+            .unwrap();
+        new_entry.file_name[..bound].copy_from_slice(&name.as_bytes()[..bound]);
+        if bound != name.len() {
+            log::info!(
+                "sys_readdir: truncation detected: {}",
+                str::from_utf8(&new_entry.file_name[..bound]).unwrap()
```
p4zuu, line 49: Since `&new_entry.file_name` looks user-provided, I guess there would be a way for a user to trigger the `unwrap()` by providing garbage string, I would prefer returning the error here instead of triggering a panic through `unwrap()`.

kernel/src/acpi/tables.rs
```diff
@@ -123,10 +124,10 @@ impl ACPITableHeader {
     /// Print a human-readable summary of the ACPI table header's fields
     #[expect(dead_code)]
     fn print_summary(&self) {
-        let sig = FixedString::from(self.sig);
-        let oem_id = FixedString::from(self.oem_id);
-        let oem_table_id = FixedString::from(self.oem_table_id);
-        let compiler_id = FixedString::from(self.compiler_id);
+        let sig = str::from_utf8(&self.sig).unwrap();
```
p4zuu, line 127: Same here about `unwrap()`s, I would prefer returning the error (we would need to change the function signature) instead of a possible panic

kernel/src/syscall/class1.rs
```diff
@@ -36,8 +38,17 @@ pub fn sys_readdir(obj_id: u32, dirents: usize, size: usize) -> Result<u64, SysC
         };
 
         let mut new_entry = DirEnt::default();
-        let fname = FileNameArray::from(name);
-        new_entry.file_name[..fname.len()].copy_from_slice(&fname);
+        // DirEnt.file_name must be nul-terminated
+        let bound = (0..min(name.len() + 1, new_entry.file_name.len()))
+            .rposition(|i| name.is_char_boundary(i))
+            .unwrap();
```
p4zuu, line 44: Same remark as `&new_entry.file_name` about the `unwrap()`

p4zuu (commented):
lgtm, but I would prefer avoiding calling `unwrap()` as much as possible

kernel/src/acpi/tables.rs
peterfang: Good idea. I'll make that change.

kernel/src/syscall/class1.rs
peterfang: Makes sense. I can return an error instead to avoid a user DOS attack.

peterfang: Upon a closer look, I now think `unwrap()` is probably an acceptable choice here. The content of `&new_entry.file_name` comes from SVSM's ramfs and the UTF-8 check was already done when SVSM unpacked the FS image (https://github.com/coconut-svsm/packit/blob/e2508f686440f6a703fb6c5c0c2fd338e55f1d38/src/file.rs#L76). The value of `bound` is protected by `is_char_boundary()` (I will return an error if the `rposition()` call fails). I can't think of a scenario in a functional SVSM environment where this string conversion would ever fail. Can you please take another look at this?

p4zuu: Thanks for pointing, that also makes sense to me. I think we can reasonably leave this `unwrap()` :)

peterfang: Thanks! I'll put a comment here to explain to use of `unwrap()`

p4zuu (approved):

PR #569, opened 2024-12-12
platform: support EOI on the native platform

msft-jlange:
Support EOI by issuing a write to the APIC EOI register.


joergroedel (approved):
LGTM.

PR #571, opened 2024-12-13
platform: Add abstractions for request processing

peterfang:
Add new abstractions to SvsmPlatform:

- request_task() is the kernel task for request processing
- request_loop() is the main vCPU request loop

Also, return Ok() for setup_percpu* on TDP platforms so they don't error out.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -12,7 +12,7 @@ use crate::error::SvsmError;
 use crate::mm::GuestPtr;
 use crate::platform::guest_cpu::GuestCpuState;
 use crate::platform::SVSM_PLATFORM;
-use crate::requests::SvsmCaa;
+use crate::sev::caa::SvsmCaa;
 use crate::sev::hv_doorbell::HVExtIntStatus;
 use crate::types::GUEST_VMPL;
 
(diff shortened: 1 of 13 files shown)


joergroedel:
Moving request processing to the platform implementation is not the right approach. The request-processing logic will at some point move to user-space, where the init process will decide, based on a feature detection mechanism, which services to run.

The right way to start here is to implement a basic version of that feature detection logic and then only start the request processing thread if running on an AMD platform. The outcome is the same as with moving stuff to the platform logic, but overall things are more generic with feature detection.

This is important because at some point there will also be AMD systems running a device model and not the standard request processing. A feature-based model can be more easily extended to support that.

peterfang:
> Moving request processing to the platform implementation is not the right approach. The request-processing logic will at some point move to user-space, where the init process will decide, based on a feature detection mechanism, which services to run.
> 
> The right way to start here is to implement a basic version of that feature detection logic and then only start the request processing thread if running on an AMD platform. The outcome is the same as with moving stuff to the platform logic, but overall things are more generic with feature detection.
> 
> This is important because at some point there will also be AMD systems running a device model and not the standard request processing. A feature-based model can be more easily extended to support that.

Understood. This totally makes sense. Should we start this work from the `CAPABILITIES` syscall?

peterfang:
Closing

PR #572, opened 2024-12-13
svsm_paging: Ensure page alignment during invalidation

peterfang:
Boot memory regions aren't always page-aligned. For example, kernel_fs_end cannot always be page-aligned because the size of the FS area has to match that of its original image exactly. Make sure the memory region always has a page-aligned size in
invalidate_boot_memory_region(). The starting address is always checked during page invalidation so just assume the caller will do the right thing.

### DIFF
--- a/kernel/src/svsm_paging.rs
+++ b/kernel/src/svsm_paging.rs
@@ -13,7 +13,7 @@ use crate::mm::pagetable::{PTEntryFlags, PageTable};
 use crate::mm::PageBox;
 use crate::platform::{PageStateChangeOp, PageValidateOp, SvsmPlatform};
 use crate::types::PageSize;
-use crate::utils::MemoryRegion;
+use crate::utils::{page_align_up, MemoryRegion};
 use bootlib::kernel_launch::KernelLaunchInfo;
 
 struct IgvmParamInfo<'a> {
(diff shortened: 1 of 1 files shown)


kernel/src/svsm_paging.rs
```diff
@@ -100,17 +100,23 @@ fn invalidate_boot_memory_region(
...
+        platform.validate_physical_page_range(valid_region, PageValidateOp::Invalidate)?;
 
         if config.page_state_change_required() {
-            platform.page_state_change(region, PageSize::Regular, PageStateChangeOp::Shared)?;
+            platform.page_state_change(
+                valid_region,
+                PageSize::Regular,
+                PageStateChangeOp::Shared,
+            )?;
         }
```
00xc, line 120: I think these already do the right thing underneath, right? They go through the region in jumps of at least `PAGE_SIZE`, so even if the end is not page-aligned, the full page will be invalidated / updated.

peterfang: That's true for the SEV-SNP implementation but not for the TDP implementation:
https://github.com/coconut-svsm/svsm/blob/b50c35dbca5ef587603ff3f0b08162e556e4c5b8/kernel/src/platform/tdp.rs#L109-L110

There were discussions about this alignment requirement when the implementation for TDP was put in but I'm also open to resolving this elsewhere in the call flow:
https://github.com/coconut-svsm/svsm/pull/515#discussion_r1841376433

kernel/src/svsm_paging.rs
```diff
@@ -100,17 +100,23 @@ fn invalidate_boot_memory_region(
     config: &SvsmConfig<'_>,
     region: MemoryRegion<PhysAddr>,
 ) -> Result<(), SvsmError> {
+    // Caller must ensure the memory region's starting address is page-aligned
+    let valid_region = MemoryRegion::new(region.start(), page_align_up(region.len()));
```
joergroedel, line 104: The naming is a bit unfortunate, can you rename it to `aligned_region`?

joergroedel (commented):
I see why this change is needed. Question(s): Can it cause problems elsewhere when we only enforce the alignment at this one call-site? Or would it be better to align the addresses in the `page_state_change()` platform call-backs?

peterfang:
> I see why this change is needed. Question(s): Can it cause problems elsewhere when we only enforce the alignment at this one call-site? Or would it be better to align the addresses in the `page_state_change()` platform call-backs?

Hi Joerg,

I checked all of the call sites for `validate_physical_page_range()` and `validate_virtual_page_range()` and this function is the only place right now where memory regions could be unaligned. `page_state_change()` is called after `validate_physical_page_range()` so for an unaligned memory region, the panic would be hit first.

kernel/src/svsm_paging.rs
peterfang: Sure, I'll make this change.

joergroedel (approved):
Thanks, looks good the me now.

joergroedel:
Will merge this after #577 and #578

kernel/src/svsm_paging.rs
joergroedel: On TDX the lower bits of the address have special meanings, so alignment is needed on this platform.

PR #573, opened 2024-12-13
igvmbuilder: set VTL2 protection flag where required

msft-jlange:
The VSM flavor of the IGVM file must specify the "VTL2 protectable" flag on the kernel memory region in order to execute correctly when running in VTL 2.


joergroedel (approved):
LGTM.

PR #574, opened 2024-12-16
vtpm: add safety comments and fix small issues

stefano-garzarella:
This is part of the effort for #228 to enable `undocumented_unsafe_blocks` clippy lint.

### DIFF
--- a/kernel/src/vtpm/tcgtpm/wrapper.rs
+++ b/kernel/src/vtpm/tcgtpm/wrapper.rs
@@ -16,7 +16,6 @@ use crate::{
 };
 
 use core::{
-    alloc::Layout,
     ffi::{c_char, c_int, c_ulong, c_void},
     ptr,
     slice::from_raw_parts,
@@ -28,15 +27,18 @@ use alloc::alloc::{alloc, alloc_zeroed, dealloc, realloc as _realloc};
 
 #[no_mangle]
 pub extern "C" fn malloc(size: c_ulong) -> *mut c_void {
-    let layout: Layout = layout_from_size(size as usize);
-    unsafe { alloc(layout).cast() }
+    if let Ok(layout) = layout_from_size(size as usize) {
+        return unsafe { alloc(layout).cast() };
+    }
+    ptr::null_mut()
 }
 
 #[no_mangle]
 pub extern "C" fn calloc(items: c_ulong, size: c_ulong) -> *mut c_void {
     if let Some(new_size) = items.checked_mul(size) {
-        let layout = layout_from_size(new_size as usize);
-        return unsafe { alloc_zeroed(layout).cast() };
+        if let Ok(layout) = layout_from_size(new_size as usize) {
+            return unsafe { alloc_zeroed(layout).cast() };
+        }
     }
     ptr::null_mut()
 }
--- a/kernel/src/vtpm/tcgtpm/wrapper.rs
+++ b/kernel/src/vtpm/tcgtpm/wrapper.rs
@@ -32,27 +32,31 @@ pub extern "C" fn malloc(size: c_ulong) -> *mut c_void {
         return ptr::null_mut();
     }
 
-    if let Ok(layout) = layout_from_size(size as usize) {
-        // SAFETY: layout is guaranteed to be non-zero size. Memory may not be
-        // initiatlized, but that's what the caller expects.
-        return unsafe { alloc(layout).cast() };
-    }
-    ptr::null_mut()
+    let Ok(layout) = layout_from_size(size as usize) else {
+        return ptr::null_mut();
+    };
+
+    // SAFETY: layout is guaranteed to be non-zero size. Memory may not be
+    // initiatlized, but that's what the caller expects.
+    unsafe { alloc(layout).cast() }
 }
 
 #[no_mangle]
 pub extern "C" fn calloc(items: c_ulong, size: c_ulong) -> *mut c_void {
-    if let Some(new_size) = items.checked_mul(size) {
-        if new_size == 0 {
-            return ptr::null_mut();
-        }
+    let Some(new_size) = items.checked_mul(size) else {
+        return ptr::null_mut();
+    };
 
-        if let Ok(layout) = layout_from_size(new_size as usize) {
-            // SAFETY: layout is guaranteed to be non-zero size.
-            return unsafe { alloc_zeroed(layout).cast() };
-        }
+    if new_size == 0 {
+        return ptr::null_mut();
     }
-    ptr::null_mut()
+
+    let Ok(layout) = layout_from_size(new_size as usize) else {
+        return ptr::null_mut();
+    };
+
+    // SAFETY: layout is guaranteed to be non-zero size.
+    unsafe { alloc_zeroed(layout).cast() }
 }
 
 #[no_mangle]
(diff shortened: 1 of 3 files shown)


stefano-garzarella:
CC @cclaudio

stefano-garzarella:
v2:
- rebased on main fixing a conflict with #576

stefano-garzarella:
CI is failing https://github.com/coconut-svsm/svsm/actions/runs/12428229021/job/34699433885?pr=574#step:7:1353 with:
```
make[1]: bindgen: No such file or directory
```

@joergroedel I'm not sure if it's related to this PR. 
Any change in the CI container?

stefano-garzarella:
It looks like for some reason now the CI container doesn't have `bindgen` installed, fixed in #583

stefano-garzarella:
v3:
- rebased on main after merging #583 that fixes the CI issue

joergroedel:
> v3:
> 
>     * rebased on main after merging [CI: install bindgen-cli #583](https://github.com/coconut-svsm/svsm/pull/583) that fixes the CI issue

Thanks, waiting a bit more for @cclaudio to approve.

cclaudio (approved):
LGTM.

cclaudio:
> > v3:
> > ```
> > * rebased on main after merging [CI: install bindgen-cli #583](https://github.com/coconut-svsm/svsm/pull/583) that fixes the CI issue
> > ```
> 
> Thanks, waiting a bit more for @cclaudio to approve.

For some reason my inbox message filters did not pick this one, sorry about that.

kernel/src/protocols/vtpm.rs
```diff
@@ -239,6 +239,7 @@ fn vtpm_command_request(params: &RequestParams) -> Result<(), SvsmReqError> {
         return Err(SvsmReqError::unsupported_call());
     }
 
+    // SAFETY: vaddr is just mapped, and its size is PAGE_SIZE
     let buffer = unsafe { from_raw_parts_mut(vaddr.as_mut_ptr::<u8>(), PAGE_SIZE) };
```
00xc, line 243: Is this truly safe though? Isn't this a shared buffer? Creating a mutable reference to it is unsound if the contents can be modified by the guest.

kernel/src/vtpm/tcgtpm/wrapper.rs
```diff
@@ -45,7 +59,29 @@ pub extern "C" fn calloc(items: c_ulong, size: c_ulong) -> *mut c_void {
...
+            return ptr::null_mut();
+        }
+
+        // SAFETY: layout_from_ptr() call ensures that `ptr` was allocated with
+        // this allocator and we are using the same `layout` used to allocate
+        // `ptr`. We also checked that `new_size` aligned does not overflow and
+        // it is not 0.
         return unsafe { _realloc(ptr, layout, new_size).cast() };
     }
     ptr::null_mut()
```
00xc, line 87: We can probably de-indent this, e.g.

```rust
let Some(layout) = layout_from_ptr(ptr) else {
    return ptr::null_mut();
};

if new_size == 0 {
    // SAFETY: layout_from_ptr() call ensures that `ptr` was allocated
    // with this allocator and we are using the same `layout` used to
    // allocate `ptr`.
    unsafe { dealloc(ptr, layout) };
    return ptr::null_mut();
}

// etc
```

kernel/src/protocols/vtpm.rs
stefano-garzarella: The guest should not touch it while issuing the SVSM call, which IIUC is also sync. @cclaudio can you confirm this?

kernel/src/vtpm/tcgtpm/wrapper.rs
stefano-garzarella: yeah, maybe also in `calloc`, I'd take a look

stefano-garzarella: Followed this suggestion in the entire file (see last patch)

stefano-garzarella:
v4:
- rebased on main
- add one patch (last one) reducing indentation [@00xc ]

kernel/src/protocols/vtpm.rs
00xc: The point is that the guest *could* touch it. Not sure what you mean by sync, but just having a mutable reference to memory that is being updated is undefined behavior. This is why `SharedBox` uses inline assembly, which can be racy but does not produce UB, IIRC.

stefano-garzarella: @00xc got it.
IIUC `SharedBox` is used for the host interface, and it uses `cpu::mem::{copy_bytes, write_bytes}`, should we use the same or extend SharedBox to support this use case (where the buffer is external)?

cclaudio: > Is this truly safe though? Isn't this a shared buffer? Creating a mutable reference to it is unsound if the contents can be modified by the guest.

Right that does not seem to be ideal. The vtpm protocol gets the physical address of a shared page and we are mapping it in the SVSM memory and passing a mutable buffer along to the TPM. @00xc eventually we will need to update that shared page with the response we get from the TPM, but until then we could do all the work in a private page. What do you think?

e.g. map the shared page as read only and copy its content to a private page and provide the private page (buffer) to the TPM. I think that should work since the TPM is running on the same address space as the SVSM kernel.

hmm ... we may need  to check if other protocols have a similar safety issue.

00xc: > > Is this truly safe though? Isn't this a shared buffer? Creating a mutable reference to it is unsound if the contents can be modified by the guest.
> 
> Right that does not seem to be ideal. The vtpm protocol gets the physical address of a shared page and we are mapping it in the SVSM memory and passing a mutable buffer along to the TPM. @00xc eventually we will need to update that shared page with the response we get from the TPM, but until then we could do all the work in a private page. What do you think?

Having a private copy would solve the issue, yes, but I'm not sure how big the performance penalty would be. The guest page should be read and written through a `GuestPtr` (maybe by adding a `read_ref()` method).

> UPDATE: I was talking to @tlendacky and he mentioned that the vTPM protocol provides the physical address of a page that is shared only between guest and the SVSM, so it is encrypted. IIUC, he suggested that we could just mark the page as read-only for non-zero VMPLs, with that it would be safe to pass it as a mutable buffer to the TPM. Once we are done with the page, we recover its permission flags. Although these page operations seem to be supported by the SVSM, I am not sure what is the API.

I'd say thats `rmp_revoke_guest_access()`.

stefano-garzarella: @00xc @cclaudio I'll drop commit d350ae4d96f11566fe2857eceaecd196d286d05b where we only add this `SAFETY` message, and we can fix it in another PR.

BTW, do you mean something like this?

```diff
diff --git a/kernel/src/protocols/vtpm.rs b/kernel/src/protocols/vtpm.rs
index 05dd8043..a798bfae 100644
--- a/kernel/src/protocols/vtpm.rs
+++ b/kernel/src/protocols/vtpm.rs
@@ -14,9 +14,11 @@ use alloc::vec::Vec;
 
 use crate::{
     address::{Address, PhysAddr},
+    cpu::flush_tlb_global_sync,
     mm::{valid_phys_address, GuestPtr, PerCPUPageMappingGuard},
     protocols::{errors::SvsmReqError, RequestParams},
-    types::PAGE_SIZE,
+    sev::utils::{rmp_grant_guest_access, rmp_revoke_guest_access},
+    types::{PageSize, PAGE_SIZE},
     vtpm::{vtpm_get_locked, TcgTpmSimulatorInterface, VtpmProtocolInterface},
 };
 
@@ -239,7 +241,14 @@ fn vtpm_command_request(params: &RequestParams) -> Result<(), SvsmReqError> {
         return Err(SvsmReqError::unsupported_call());
     }
 
-    // SAFETY: vaddr is just mapped, and its size is PAGE_SIZE
+    // Make sure the guest can't make modifications to the buffer
+    rmp_revoke_guest_access(vaddr, PageSize::Regular)?;
+
+    // TLB flush needed to propagate new permissions
+    flush_tlb_global_sync();
+
+    // SAFETY: vaddr is just mapped, we are sure the guest can't modify it,
+    // and its size is PAGE_SIZE
     let buffer = unsafe { from_raw_parts_mut(vaddr.as_mut_ptr::<u8>(), PAGE_SIZE) };
 
     let response_size = match cmd {
@@ -256,6 +265,12 @@ fn vtpm_command_request(params: &RequestParams) -> Result<(), SvsmReqError> {
         GuestPtr::<u32>::new(vaddr).write(response_size)?;
     }
 
+    // Make sure the guest can access again
+    rmp_grant_guest_access(vaddr, PageSize::Regular)?;
+
+    // TLB flush needed to propagate new permissions
+    flush_tlb_global_sync();
+
     Ok(())
 }
 
```

00xc: I'd say so, yes (although someone can correct me on the need for a TLB flush on both operations). We would also need reverting the RMP change on the error path, and that in turn might require holding `PVALIDATE_LOCK` (although I'm foggy on the attack vector that lock prevented, @joergroedel remembers perhaps). Anyhow, this can be handled in the review for the separate PR.

stefano-garzarella: @00xc thanks! yeah, we can discuss in the separate PR, I'm resolving this conversation.

@cclaudio do you plan to send a PR, otherwise I can do that.

cclaudio: @stefano-garzarella If you can do that great, I appreciate. I wish I had more arms ...

stefano-garzarella: @cclaudio sure, I'll post a PR :-)

PR #575, opened 2024-12-17
testing: support configurations other than SNP/QEMU

msft-jlange:
The `test-in-svsm` build configuration should support environments other than SNP running on QEMU.


joergroedel (commented):
The reason clippy fails is that with this PR `crate::testing` is references for `make test` builds, while it is only included for `make test_in_svsm` builds.

joergroedel (commented):
The changes work as proposed. But longer term I think the way the platform checks are done in the test cases will cause some problems. We should think about a better way to make the test-cases hypervisor specific, maybe by introducing per-hypervisor feature names.

PR #576, opened 2024-12-17
libtcgtpm: update _plat__NVEnable and _plat__NVDisable signatures

stefano-garzarella:
`_plat__NVEnable()` and `_plat__NVDisable()` APIs were updated in the new TCG TPM reference implementation to include an additional `paramSize` parameter (size_t) in their signatures.

This patch synchronizes the header file and Rust FFI calls with the new function prototypes.

Previously, the outdated header and calls passed fewer arguments than required by the new TCG TPM library. However, this mismatch did not cause issues because the new parameter (paramSize) is not used by the new implementation.

Additionally, the first parameter of `_plat__NVDisable()` has changed from `int delete` to `void* platParameter` in the new signature. However, the actual value passed (1) remains the same.

In the future we should find a better way to keep the headers synchronized and avoid this problem.

Fixes: 9894bf4d ("libtcgtpm: swtich from MS TPM to TCG TPM")

### DIFF
--- a/kernel/src/vtpm/tcgtpm/mod.rs
+++ b/kernel/src/vtpm/tcgtpm/mod.rs
@@ -162,11 +162,11 @@ impl VtpmInterface for TcgTpm {
         // 5. Power it on indicating it requires startup. By default, OVMF will start
         //    and selftest it.
 
-        unsafe { _plat__NVEnable(VirtAddr::null().as_mut_ptr::<c_void>()) };
+        unsafe { _plat__NVEnable(VirtAddr::null().as_mut_ptr::<c_void>(), 0) };
 
         let mut rc = self.manufacture(1)?;
         if rc != 0 {
-            unsafe { _plat__NVDisable(1) };
+            unsafe { _plat__NVDisable(1 as *mut c_void, 0) };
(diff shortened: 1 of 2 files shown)


stefano-garzarella:
CC @cclaudio

osteffenrh (approved):

cclaudio (approved):
LGTM. 
It seems that upstream has re-structured the platform header files. Later we may want to follow what they are doing by generatiing bindings for those headers.

https://github.com/TrustedComputingGroup/TPM/blob/main/TPMCmd/Simulator/src/simulatorPrivate.h , which includes the new platform headers.

stefano-garzarella:
> LGTM. 

Thanks for the review!

>It seems that upstream has re-structured the platform header files. Later we may want to follow what they are doing by generatiing bindings for those headers.
> 
> https://github.com/TrustedComputingGroup/TPM/blob/main/TPMCmd/Simulator/src/simulatorPrivate.h , which includes the new platform headers.

Yeah, I agree on that. I'll look into that, removing our intermediate header, and using `bindgen` directly on the headers provided by them!

PR #579, opened 2024-12-18
cpu: implement IPI support

msft-jlange:
This change includes logic to support sending inter-processor interrupts (IPIs) to other processors.  An IPI can be sent as a multicast message, or as a unicast message to a specific processor as a mutable message which receives a response.  All IPI delivery is synchronous, meaning that the sender will block until the IPI is delivered and the receiver has processed it.

### DIFF
--- a/kernel/src/cpu/cpuset.rs
+++ b/kernel/src/cpu/cpuset.rs
@@ -1,0 +1,161 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) Microsoft Corporation
+//
+// Author: Jon Lange (jlange@microsoft.com)
+
+use core::sync::atomic::{AtomicU64, Ordering};
+
+pub const MAX_CPUS: usize = 1024;
+
+/// Represents a set of CPUs, based on CPU index.  A maximum of `MAX_CPUS` can
+/// be represented.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct CpuSet {
+    bitmask: [u64; (MAX_CPUS + 63) / 64],
+}
+
+impl CpuSet {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Adds a CPU to the set.
+    ///
+    /// * `cpu_index`: the index of the CPU to add to the set.
+    pub fn add(&mut self, cpu_index: usize) {
+        self.bitmask[cpu_index >> 6] |= 1u64 << (cpu_index & 0x3F);
+    }
+
+    /// Removes a CPU from the set.
+    ///
+    /// * `cpu_index`: the index of the CPU to remove from the set.
+    pub fn remove(&mut self, cpu_index: usize) {
+        self.bitmask[cpu_index >> 6] &= !(1u64 << (cpu_index & 0x3F));
+    }
+
+    /// Produces an iterator to iterate over the set.
+    pub fn iter(&self) -> CpuSetIterator<'_> {
+        CpuSetIterator::new(self)
+    }
+}
+
+#[derive(Debug)]
+pub struct CpuSetIterator<'a> {
+    cpu_set: &'a CpuSet,
+    current_mask: u64,
+    mask_index: usize,
+}
+
+impl<'a> CpuSetIterator<'a> {
+    fn new(cpu_set: &'a CpuSet) -> Self {
+        Self {
+            cpu_set,
+            current_mask: cpu_set.bitmask[0],
+            mask_index: 0,
+        }
+    }
+}
+
+impl Iterator for CpuSetIterator<'_> {
+    type Item = usize;
+    fn next(&mut self) -> Option<usize> {
+        while self.current_mask == 0 {
+            self.mask_index += 1;
+            if self.mask_index == self.cpu_set.bitmask.len() {
+                return None;
+            }
+
+            self.current_mask = self.cpu_set.bitmask[self.mask_index];
+        }
+
+        let index = self.current_mask.trailing_zeros();
+        self.current_mask &= !(1u64 << index);
+        Some((self.mask_index << 6) | index as usize)
+    }
+}
+
+/// Represents a set of CPUs, based on CPU index, which supports atomic
+/// addition and removal.  A maximum of 1024 CPUs can be represented.
+#[derive(Debug, Default)]
+pub struct AtomicCpuSet {
+    bitmask: [AtomicU64; (MAX_CPUS + 63) / 64],
+}
+
+impl AtomicCpuSet {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Adds a CPU to the set.
+    ///
+    /// * `cpu_index`: the index of the CPU to add to the set.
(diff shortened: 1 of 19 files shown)


kernel/src/cpu/cpuset.rs
```diff
@@ -0,0 +1,159 @@
...
+// Author: Jon Lange (jlange@microsoft.com)
+
+use core::sync::atomic::{AtomicU64, Ordering};
+
+/// Represents a set of CPUs, based on CPU index.  A maximum of 1024 CPUs can
+/// be represented.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct CpuSet {
+    bitmask: [u64; 16],
+}
```
joergroedel, line 14: I think it is better to derive the `16` from a global `MAX_CPUS` constant, so that it can easily be changed.

kernel/src/cpu/ipi.rs
```diff
@@ -0,0 +1,566 @@
...
+///   selected by CPU index.
+/// * `AllButSelf` indicates all CPUs other than the sending processor.
+/// * `All` indicates all CPUs.
+#[derive(Clone, Copy, Debug)]
+pub enum IpiTarget {
+    Single(usize),
+    Multiple(CpuSet),
+    AllButSelf,
+    All,
+}
```
joergroedel, line 63: That is a rather big data structure (128 bytes or 2 cache lines), would it be more efficient to disallow `Copy` and `Clone` and force this to be passed around by reference?

kernel/src/cpu/ipi.rs
```diff
@@ -0,0 +1,566 @@
...
+    /// perform a deep copy of the contents of the source buffer into the
+    /// shared destination buffer.
+    ///
+    /// Arguments:
+    ///
+    /// *`src`: A pointer to the input message.
+    /// *`buffer`: A byte slice in shared memory which will be the target of the copy.
+    fn copy_to_shared(&self, buffer: &mut [u8])
+    where
+        Self: Sized,
```
joergroedel, line 122: Request `Copy` trait here? Also true for the other `copy_*` methods.

kernel/src/cpu/ipi.rs
```diff
@@ -0,0 +1,566 @@
...
+        }
+    }
+}
+
+// This function is the IPI workhorse.  As input, it takes an IpiHelper which
+// is the interface to the correct IPI message trait implementation.  This
+// is consumed as a dynamic dispatch trait to avoid explosion due to multiple
+// generic message implementations.
+pub fn send_ipi(
+    mut target_set: IpiTarget,
```
joergroedel, line 285: This can cause rather expensive copies for each function call, is it better to just borrow `IpiTarget`?

kernel/src/utils/scoped.rs
```diff
@@ -0,0 +1,107 @@
...
+
+use core::ops::{Deref, DerefMut};
+
+/// `ScopedRef` and `ScopedMut` are designed to solve the problem of managing
+/// lifetimes of references created from pointers.  Normally, when a reference
+/// is created from a pointer (such as with `ptr::as_ref()`), it is associated
+/// with the static lifetime, and as a result, the compiler is unable to
+/// determine whether the reference will live long enough for its intended
+/// use.  While functions like `ptr::as_ref()` can associate the reference with
+/// a lifetime, the compiler cannot usefully use this information to enfoce
```
joergroedel, line 15: Typo `enfoce` -> `enforce`

joergroedel (commented):
Thanks for working on this, Jon. I left a view review comments.

kernel/src/cpu/ipi.rs
msft-jlange: As we discussed, we will change `IpiTarget` so that it holds a reference to `CpuSet` if needed.  This ensures that the `IpiTarget` structure will be small and can safely be passed by value (which is needed in case the IPI logic needs to modify the IPI target to prevent sending an interrupt to the current CPU).  If a `CpuSet` reference is needed, the IPI logic will create a copy of the CPU set, but only if that copy is actually needed.

kernel/src/cpu/ipi.rs
msft-jlange: As we discussed, requiring `Copy` is unhelpful for two reasons.  First, requiring `Copy` doesn't solve the deep-copy problem that will exist if the IPI message contains references to other objects, because references inherently support `Copy` but in our design, the reference will not be valid on the target CPU due to address space differences.  This means that `Copy` doesn't eliminate the need for the copy method here.  Also, requiring `Copy` prevents the use of non-`Copy` types in IPI messages (such as `Vec`) which could safely be serialized into the shared buffer for use on the target CPU.  We do not want to impose such a constraint that prevents the use of such message types.

We agreed that we would add logic to the unit tests to ensure that the IPI logic does the correct work to project IPI messages into the shared address space, and that we would test this by implementing `Drop` in the unit tests to verify that it is called exactly the correct number of times.

kernel/src/cpu/ipi.rs
msft-jlange: As stated in another comment, this is largely resolved by changing `IpiTarget` to pass the `CpuSet` by reference, thus making copies cheap.

msft-jlange:
I've added the `Sync` trait bound to multicast messages since those messages can be referenced simultaneously by multiple threads.  `Sync` is not needed for unicast messages because they do not have multi-thread visibility.

kernel/src/cpu/ipi.rs
```diff
@@ -0,0 +1,623 @@
...
+        // interrupts.
+        if SVSM_PLATFORM.use_interrupts() {
+            let mut drop_count: usize = 0;
+            let message = TestIpi::new(4, &mut drop_count);
+            send_multicast_ipi(IpiTarget::All, &message);
+            drop(message);
+            // Verify that `drop()` was called exactly once on thie IPI
+            // message.
+            assert_eq!(drop_count, 1);
+        }
```
joergroedel, line 604: Turns out that the `SnpPlatform` assumes IRQs supported when it finds restricted injection enabled. Unfortunately this does not hold true with our latest development environment on the KVM side (which enables restricted injection, but has not yet full IRQ support). So these test-cases run on KVM as well and panic the test-in-svsm run.

joergroedel (commented):
Overall this looks good to me. There are two things I mentioned in separate comments which need attention.

kernel/src/cpu/ipi.rs
msft-jlange: That's very unfortunate, as it implies that there is no standard way for a VM to determine whether the host supports per-VMPL interrupts.  @tlendacky, I'm not sure whether this is something that should be added to the GHCB protocol or whether we should just hack around it for now because we know KVM will be updated to support per-VMPL interrupts in the near future.

For the time being, I'm just going to add a hack to our code to inhibit use of interrupts when running SNP under KVM, but we need a better long-term answer.

joergroedel: The workaround is functional, thanks for fixing this.

PR #581, opened 2024-12-18
platform/native: use SIPI for API startup

msft-jlange:
Native platforms that cannot make use of the Hyper-V hypercalls require the use of INIT/SIPI to perform multi-processor startup.

### DIFF
--- a/igvmbuilder/src/sipi.rs
+++ b/igvmbuilder/src/sipi.rs
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2024 Microsoft Corporation
+//
+// Author: Jon Lange <jlange@microsoft.com>
+
+use bootlib::kernel_launch::SIPI_STUB_GPA;
+use igvm::IgvmDirectiveHeader;
+use igvm_defs::{IgvmPageDataFlags, IgvmPageDataType, PAGE_SIZE_4K};
+
+pub fn add_sipi_stub(compatibility_mask: u32, directives: &mut Vec<IgvmDirectiveHeader>) {
+    // The SIPI stub is the code that is required on native platforms to
+    // transition the processor out of real mode and into 64-bit mode when APs
+    // are started.  It includes 16-bit code, 32-bit code, and 64-bit code.
+    // For simplicity, to avoid having to invoke multiple build elements to
+    // produce a number of separate, small code modules that are stitched
+    // together, this routine (somewhat awkwardly) simply just captures the
+    // required code bytes as a constant array, since this code is small and
+    // will almost never change.  The assembly code and corresponding
+    // disassembly are listed here for reference.
+    //
+    // F000: 0F 20 C0                       mov eax, cr0
+    // F003: 80 C8 01                       or al, 1
+    // F006: 0F 22 C0                       mov cr0, eax
+    // F009: 2E 66 0F 01 16 1A 00           lgdt cs:[001A]
+    // F010: EA 40 F0 08 00                 jmp 0008:F040
+    // F015: CC                             int 3
+    // F016: CC                             int 3
+    // F017: CC                             int 3
+    // F018: CC                             int 3
+    // F019: CC                             int 3
+    // F01A: 1F 00 20 F0 00 00
+    //
+    // GDT:
+    // F020: 00 00 00 00 00 00 00 00 // null selector
+    // F028: FF FF 00 00 00 9B CF 00 // 32-bit code
+    // F030: FF FF 00 00 00 9B AF 00 // 64-bit code
+    // F038: FF FF 00 00 00 93 CF 00 // data
+    //
+    // F040: 66 B8 18 00                    mov ax, 18h
+    // F044: 8E D8                          mov ds, ax
+    // F046: 8E D0                          mov ss, ax
+    // F048: 8E C0                          mov es, ax
+    // F04A: 8B 05 F8 FF 00 00              mov eax, [FFF8] // page table
+    // F050: 0F 22 D8                       mov cr3, eax
+    // F053: B9 80 00 00 C0                 mov ecx, C0000080h
+    // F058: 0F 32                          rdmsr
+    // F05A: 0F BA E8 08                    bts eax, 8 // EFER_LME
+    // F05E: 0F 30                          wrmsr
+    // F060: 0F 20 E0                       mov eax, cr4
+    // F063: 83 C8 20                       or eax, 20h // CR4_PAE
+    // F066: 0F 22 E0                       mov cr4, eax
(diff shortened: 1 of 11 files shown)


igvmbuilder/src/sipi.rs
```diff
@@ -0,0 +1,83 @@
...
+        0x00, 0x00, 0x00, 0xEA, 0x40, 0xF0, 0x00, 0x00, 0x08, 0x00, 0xCC, 0x1F, 0x00, 0x20, 0xF0,
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xFF, 0xFF, 0x00, 0x00, 0x00,
+        0x9B, 0xCF, 0x00, 0xFF, 0xFF, 0x00, 0x00, 0x00, 0x9B, 0xAF, 0x00, 0xFF, 0xFF, 0x00, 0x00,
+        0x00, 0x93, 0xCF, 0x00, 0x66, 0xB8, 0x18, 0x00, 0x8E, 0xD8, 0x8E, 0xD0, 0x8E, 0xC0, 0x8B,
+        0x05, 0xF8, 0xFF, 0x00, 0x00, 0x0F, 0x22, 0xD8, 0xB9, 0x80, 0x00, 0x00, 0xC0, 0x0F, 0x32,
+        0x0F, 0xBA, 0xE8, 0x08, 0x0F, 0x30, 0x0F, 0x20, 0xC0, 0x0F, 0xBA, 0xE8, 0x1F, 0x0F, 0x22,
+        0xC0, 0xEA, 0x71, 0xF0, 0x00, 0x00, 0x10, 0x00, 0x48, 0x8B, 0x25, 0x78, 0x0F, 0x00, 0x00,
+        0xBF, 0x00, 0x00, 0x01, 0x00, 0x2B, 0x3D, 0x79, 0x0F, 0x00, 0x00, 0xFF, 0x15, 0x5F, 0x0F,
+        0x00, 0x00, 0xFF, 0x25, 0x51, 0x0F, 0x00, 0x00,
+    ];
```
joergroedel, line 68: When moving to a state closer to a non-draft PR this array should be generated (ideally via `build.rs`). I suggest having an `trampoline.s` file and build a binary out of that which can then be included into one of SVSMs `.s` files. This will make this code much more maintainable.

kernel/src/platform/native.rs
```diff
@@ -172,6 +181,35 @@ impl SvsmPlatform for NativePlatform {
...
+        // INIT-SIPI to start the processor.  No second SIPI is required when
+        // running virtualized.
+        let icr = ApicIcr::new().with_destination(cpu.shared().apic_id());
+        let init_icr = icr.with_message_type(IcrMessageType::Init);
+        write_msr(APIC_MSR_ICR, init_icr.into());
+        let sipi_vector = SIPI_STUB_GPA >> 12;
+        let sipi_icr = icr
+            .with_message_type(IcrMessageType::Sipi)
+            .with_vector(sipi_vector.try_into().unwrap());
+        write_msr(APIC_MSR_ICR, sipi_icr.into());
```
joergroedel, line 211: The low-level APIC accesses should happen in the APIC code.

joergroedel (commented):
Apart from the comments I left this looks good to me, though I did not test it.

igvmbuilder/src/sipi.rs
msft-jlange: As we discussed, this can be deferred to a separate PR.

kernel/src/platform/native.rs
msft-jlange: There is no separate APIC code; the IRQ request routine is part of the platform object, so using platform-specific logic here is not unreasonable.  That said, this code is actually broken, so it would be better to call the `post_irq` method on the current platform object to ensure that there the correct APIC access is performed.

kernel/src/platform/native.rs
```diff
@@ -180,6 +189,35 @@ impl SvsmPlatform for NativePlatform {
...
         }
 
-        todo!();
+        // Translate this context into an AP start context and place it it in
+        // the AP startup transition page.
+        let ap_context = create_ap_start_context(context, self.transition_cr3);
+
+        let context_pa = PhysAddr::new(SIPI_STUB_GPA as usize);
+        let context_mapping =
+            PerCPUPageMappingGuard::create(context_pa, context_pa + PAGE_SIZE, 0)?;
```
peterfang, line 198: This can be replaced with `PerCPUPageMappingGuard::create_4k()`

peterfang:
I'll need more experimentation with this after it gets merged, but there is definitely an opportunity for the trampoline code in `sipi.rs` to be generic enough to replace most of stage1's `reset.S`.

kernel/src/cpu/smp.rs
```diff
@@ -46,6 +50,69 @@ pub fn start_secondary_cpus(platform: &dyn SvsmPlatform, cpus: &[ACPICPUInfo]) {
...
+
+        /* Begin execution from the starting RIP, which is at the top of the
+         * stack. */
+        ret
+        "#,
+    options(att_syntax)
+);
+
+pub fn create_ap_start_context(
+    initial_context: &hyperv::HvInitialVpContext,
```
peterfang, line 100: Is it possible to avoid using a hypervisor-specific structure here and instead use a hypervisor-agnostic one?

msft-jlange: The rest of the code base has already adopted this structure as the generic one.  See for example `SvsmPlatform::start_cpu()`.  We could switch the code base to a different structure, but I am not aware of any suitable, common alternative that specifies the minimum 64-bit startup state, and creating one would essentially involve duplicating this structure exactly - so what advantage would we gain by making a switch?

peterfang: It's for the sake of aesthetics. I'm reusing parts of this patch for TDX bring-up, so seeing HyperV structures in these common paths seems counter-intuitive, especially when these paths are also used by KVM.
I do agree that this could very well create code duplication but it's sometimes the cost of supporting multiple hosts/platforms. A `From<>` implementation will probably make this a bit less tedious.

Anyway it may not be that important at this stage of development but it's something I want to be mindful of. I don't object to punting it for now.

msft-jlange: I'm happy to make a change here but it's not clear what change will be best.  Using `From` requires having some other structure to convert to, and the challenge I see is that there is no obvious candidate for such a structure.  If you want to suggest a set of changes (or a specific approach to take), I would be happy to work with you to get it into the code.

Meanwhile, I'm going to resolve this thread based on your statement that we can defer this work outside of this PR.

peterfang: Sounds good. We can revisit this later since I can't think of a fast & easy way to do this either :-)

PR #582, opened 2024-12-19
platform: abstract tlb flush

msft-jlange:
Different platforms implement TLB flush in different ways, so TLB flush operations should be abstracted in the platform object.


joergroedel (approved):
Changes LGTM.

PR #583, opened 2024-12-20
CI: install bindgen-cli

stefano-garzarella:
`bindgen` is used in our build process to build libtcgtpm. CI recently started to fail because it was not installed.

Add a step to make sure `bindgen` is installed correctly in our workflows.


stefano-garzarella:
@joergroedel this should fix also the failure on `main` branch.
Something must have changed in the CI containers, where `bidngen` was previously installed a priori.

osteffenrh (approved):

PR #584, opened 2024-12-20
kernel/cpu: add safety comments

p4zuu:
Safety comments for:
- `stac` and `clac` instructions
- some TSS-related pieces of GDT code
- as described in #553 `write_msr()` should be unsafe since it can break memory safety. This part is still missing two comments [here](https://github.com/p4zuu/svsm/blob/f11a86f1ecbc33550ccaf443fe74dea15dceb5ac/kernel/src/sev/msr_protocol.rs#L148) and [here](https://github.com/p4zuu/svsm/blob/f11a86f1ecbc33550ccaf443fe74dea15dceb5ac/kernel/src/sev/msr_protocol.rs#L175) in the GHCB MSR protocol where I have doubt if the MSR request is subject to breaking memory safety or not, specifically `SNP_REG_GHCB_GPA_REQ` and `SNP_STATE_CHANGE_REQ` GHCB requests. My guess is a garbage request parameter can break the platform, but does it still count as memory safety?

### DIFF
--- a/kernel/src/cpu/gdt.rs
+++ b/kernel/src/cpu/gdt.rs
@@ -80,7 +80,18 @@ impl GDT {
 
         let tss_entries = &self.entries[idx..idx + 1].as_mut_ptr();
 
+        // SAFETY:
+        // For add():
+        //   - idx and idx + size_of::<GDTEntry>() don't overflow isize.
+        //   - the borrow checker guarantees that self is still allocated
+        //   - self.entries[6:8] fits in self's allocation.
+        // For write_volatile():
+        //   - the borrow checker guarantees that self.entries is allocated
+        //   - alignment is checked inside
         unsafe {
+            assert_eq!(align_of_val(&tss_entries.add(0)), align_of::<GDTEntry>());
+            assert_eq!(align_of_val(&tss_entries.add(1)), align_of::<GDTEntry>());
+
             tss_entries.add(0).write_volatile(desc0);
             tss_entries.add(1).write_volatile(desc1);
         }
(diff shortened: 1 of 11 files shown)


msft-jlange:
> I have doubt if the MSR request is subject to breaking memory safety or not, specifically `SNP_REG_GHCB_GPA_REQ` and `SNP_STATE_CHANGE_REQ` GHCB requests. My guess is a garbage request parameter can break the platform, but does it still count as memory safety?

Both of these GHCB operations can cause physical memory to be remapped (one may cause the GHCB data to appear in a different place, and one may cause memory to be remapped with a different encryption attribute).  Because both of these operations can cause mapped memory to change contents, they are technically unsafe.

kernel/src/cpu/gdt.rs
```diff
@@ -93,11 +104,12 @@ impl GDT {
...
 
-        unsafe {
-            self.set_tss_entry(desc0, desc1);
-            asm!("ltr %ax", in("ax") SVSM_TSS, options(att_syntax));
-            self.clear_tss_entry()
-        }
+        self.set_tss_entry(desc0, desc1);
+        // SAFETY: loading task register must me done in assembly.
+        // It's safe to do so as long as a global GDT is in use and still
+        // allocated, which is always our case.
```
msft-jlange, line 110: Actually, it is not necessarily the case that a global GDT is in use here.  However, the lifetime of the GDT doesn't matter, because once the task register is loaded, only the TSS needs to remain live.  For that reason, either this whole function should be `unsafe` (because the compiler cannot prove that the lifetime of the `tss` parameter cannot be proven) or the `tss` parameter should be declared as `&'static`.  A quick experiment suggests that the latter approach is workable as long as we modify `PerCpu::load()` and `PerCpu::load_isst()` to take `&'static self`.

kernel/src/cpu/vc.rs
```diff
@@ -577,7 +577,9 @@ mod tests {
     fn test_wrmsr_tsc_aux() {
         if is_qemu_test_env() && is_test_platform_type(SvsmPlatformType::Snp) {
             let test_val = 0x1234;
-            verify_ghcb_gets_altered(|| write_msr(MSR_TSC_AUX, test_val));
+            verify_ghcb_gets_altered(||
+                // SAFETY: writing to TSC MSR doesn't break memory safety.
```
msft-jlange, line 581: Should say `TSC_AUX MSR`.

msft-jlange (changes requested):

p4zuu:
> Both of these GHCB operations can cause physical memory to be remapped (one may cause the GHCB data to appear in a different place, and one may cause memory to be remapped with a different encryption attribute).  Because both of these operations can cause mapped memory to change contents, they are technically unsafe.

Thanks for your input on this Jon, and sorry for the massively delayed update. I entered a busy relocation mode and will start working on this again in a couple of weeks hopefully.

p4zuu:
Thanks for the initial review @msft-jlange. I addressed the comments and this is open again for review

kernel/src/cpu/gdt.rs
p4zuu: That sounds good to me. It builds fine with a `&'static self` for `PerCpu::load_gdt_tss()` and `PerCpu::load()`.

msft-jlange (approved):
Thanks for returning to this PR!  These changes look good.

PR #585, opened 2024-12-20
requests: handle interrupts when halted in the request loop

msft-jlange:
If the request loop determines that there is no active VMSA, the processor enters a halt state.  This halt state can terminate if an interrupt is received.  If that happens, the request loop will observe that no change to the VMSA mapping has been requested.  In this case, the request loop must return to the halt state rather than attempting to process any requests.

### DIFF
--- a/kernel/src/requests.rs
+++ b/kernel/src/requests.rs
@@ -71,7 +71,12 @@ pub fn update_mappings() -> Result<(), SvsmError> {
     let mut ret = Ok(());
 
     if !locked.needs_update() {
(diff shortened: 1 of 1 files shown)


msft-jlange:
@joergroedel, this is needed to make the IPI unit tests pass correctly.  Without this change, the tests will pass, but all APs will panic after the IPI test has completed because they cannot handle the wake from halt properly.  It is posted as a separate PR because it can merge separately.

joergroedel (approved):
LGTM.

PR #586, opened 2024-12-21
igvmbuilder: enable long mode paging in the initial VSM context

msft-jlange:
The VSM environment requires CR0.PG to be set when entering VTL 2 for the first time.  In this environment, the IGVM file must be built with initial page tables that map the low 4 GB, which permits stage 2 to start with paging enabled.  This also requires that stage 2 must start in long mode, because the paging mode cannot be changed once the VM starts executing (if CR0.PG must always be set, then it is not possible to change from 3-level PAE tables to 4-level tables).  Stage 2 will still start executing in 32-bit mode as usual, and will rebuild the page tables in its own memory area, as it always has, but that code can now execute with paging enabled as required by VSM.


joergroedel (approved):
LGTM.

PR #587, opened 2024-12-24
platform/native: require the use of X2APIC

msft-jlange:
For simplicity and consistency, require the native platform to use X2APIC so that APIC management can be identical across all native environments.

### DIFF
--- a/kernel/src/platform/native.rs
+++ b/kernel/src/platform/native.rs
@@ -7,7 +7,7 @@
 use crate::address::{PhysAddr, VirtAddr};
 use crate::console::init_svsm_console;
 use crate::cpu::cpuid::CpuidResult;
-use crate::cpu::msr::write_msr;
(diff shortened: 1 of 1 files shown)


kernel/src/platform/native.rs
```diff
@@ -41,6 +44,11 @@ impl NativePlatform {
 
 impl Default for NativePlatform {
     fn default() -> Self {
+        // Execution is not possible unless X2APIC is supported.
+        let features = CpuidResult::get(1, 0);
+        if (features.ecx & 0x200000) == 0 {
+            panic!("X2APIC is not supported");
+        }
```
joergroedel, line 51: The SVSM really needs a proper CPU feature checking framework. No need to implement it in this PR, though.

joergroedel (approved):
This looks good for now.

PR #588, opened 2025-01-06
task: assign each task a unique portion of the kernel address space

msft-jlange:
Each task uses a portion of the kernel address space to store task-specific memory, which is only visible in the context of the task that owns it.  However, references to task-specific memory can be passed between tasks, and Rust memory safety relies on the principle that any reference (which is captured by its virtual address) must refer to the same underlying data in every context in which that reference is used.  This safety principle is defeated if any address is valid in more than one task at a time.  Assigning each task a unique portion of the address space ensures that any per-task address that crosses a task boundary will cause a fault instead of referring to the wrong data.

### DIFF
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -20,8 +20,7 @@ use crate::cpu::irq_state::EFLAGS_IF;
 use crate::cpu::percpu::{current_task, PerCpu};
 use crate::cpu::shadow_stack::is_cet_ss_supported;
 use crate::cpu::sse::{get_xsave_area_size, sse_restore_context};
-use crate::cpu::X86ExceptionContext;
-use crate::cpu::{irqs_enable, X86GeneralRegs};
+use crate::cpu::{irqs_enable, X86ExceptionContext, X86GeneralRegs};
 use crate::error::SvsmError;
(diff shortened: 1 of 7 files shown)


kernel/src/task/tasks.rs
```diff
@@ -137,6 +171,11 @@ pub struct Task {
     /// Page table that is loaded when the task is scheduled
     pub page_table: SpinLock<PageBox<PageTable>>,
 
+    /// Virtual address region that has been allocated for this task.
+    /// This is not referenced but must be stored so that it is dropped when
+    /// the Task is dropped.
+    _ktask_region: TaskVirtualRegionGuard,
+
```
joergroedel, line 178: Suggestion for a follow-on change: Reference this member by reading the start and end addresses from it where needed and avoid the redundant struct members for that.

joergroedel (approved):

kernel/src/task/tasks.rs
msft-jlange: Once the `Task` structure exists, there are no more references to the start/end addresses of the range except from within the kernel `VMR` itself (which cannot take a dependency on a task-specific guard).  So I don't believe there are any places where I can take advantage of your suggestion.

PR #590, opened 2025-01-08
stage2: Use IGVM memory map to resize SVSM kernel

AdamCDunlap:
If the IGVM memory map contains a PARAVISOR_RESERVED entry, then place the SVSM kernel at that region instead of the region specified in the IgvmParamBlock. If no such entry exists in the memory map, then continue using the IgvmParamBlock.

This allows the hypervisor to dynamically resize the SVSM based on the machine shape without changing the IgvmParamBlock and thus the launch measurement. This is necessary because the SVSM uses more memory when there are more vCPUs (and, in the future, guest memory).

Note this relies on an unpublished IGVM spec change to add the PARAVISOR_RESERVED MemoryMapEntryType.

### DIFF
--- a/kernel/src/igvm_params.rs
+++ b/kernel/src/igvm_params.rs
@@ -7,7 +7,7 @@
 extern crate alloc;
 
 use crate::acpi::tables::ACPICPUInfo;
-use crate::address::{PhysAddr, VirtAddr};
+use crate::address::{Address, PhysAddr, VirtAddr};
 use crate::cpu::efer::EFERFlags;
 use crate::error::SvsmError;
 use crate::mm::{GuestPtr, PerCPUPageMappingGuard, PAGE_SIZE};
@@ -76,8 +76,34 @@ impl IgvmParams<'_> {
 
     pub fn find_kernel_region(&self) -> Result<MemoryRegion<PhysAddr>, SvsmError> {
         let kernel_base = PhysAddr::from(self.igvm_param_block.kernel_base);
-        let kernel_size: usize = self.igvm_param_block.kernel_size.try_into().unwrap();
-        Ok(MemoryRegion::<PhysAddr>::new(kernel_base, kernel_size))
+        let mut kernel_size = self.igvm_param_block.kernel_min_size;
+
+        // Check the untrusted hypervisor-provided memory map to see if the size of the kernel
+        // should be adjusted. The base location and mimimum and maximum size specified by the
+        // measured igvm_param_block are still respected to ensure a malicious memory map cannot
+        // cause the SVSM kernel to overlap anything important or be so small it causes weird
+        // failures. But if the hypervisor gives a memory map entry of type HIDDEN that starts at
+        // kernel_start, use the size of that entry as a guide. This allows the hypervisor to
+        // adjust the size of the SVSM kernel to what it expects will be needed based on the
+        // machine shape.
+        if let Some(memory_map_region) = self.igvm_memory_map.memory_map.iter().find(|region| {
+            region.entry_type == MemoryMapEntryType::HIDDEN
+                && region.starting_gpa_page_number.try_into() == Ok(kernel_base.pfn())
+        }) {
+            let region_size_bytes = memory_map_region
+                .number_of_pages
+                .try_into()
+                .unwrap_or(u32::MAX)
+                .saturating_mul(PAGE_SIZE as u32);
+            kernel_size = region_size_bytes.clamp(
+                self.igvm_param_block.kernel_min_size,
+                self.igvm_param_block.kernel_max_size,
+            );
+        }
+        Ok(MemoryRegion::<PhysAddr>::new(
+            kernel_base,
+            kernel_size.try_into().unwrap(),
+        ))
     }
 
     pub fn reserved_kernel_area_size(&self) -> usize {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -514,9 +514,9 @@ dependencies = [
 
 [[package]]
 name = "igvm"
-version = "0.3.3"
+version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7984b10433b50e06a06bd50c69bca4888a5d7de8975f64ea4c2a7687eb99b09d"
+checksum = "7b4ae8479aa3163c8a0fa716aa6ef08a6553e1097f8a89544f46fee695b5a162"
 dependencies = [
  "bitfield-struct 0.7.0",
  "crc32fast",
@@ -531,9 +531,9 @@ dependencies = [
 
 [[package]]
 name = "igvm_defs"
-version = "0.3.3"
+version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b64ec5588c475372ae830475d3ee9a7bd255407dcb9f03faf6d493556eb6105a"
+checksum = "e4f70c18b574e5c7fa6222c1f0ebd8bfe2e14b762573b799faf8697c044b0e2a"
 dependencies = [
  "bitfield-struct 0.7.0",
  "open-enum",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,8 +48,8 @@ bitflags = "2.4"
 clap = { version = "4.4.14", default-features = false }
 gdbstub = { version = "0.6.6", default-features = false }
 gdbstub_arch = { version = "0.2.4" }
-igvm = { version = "0.3.2", default-features = false }
-igvm_defs = { version = "0.3.2", default-features = false }
+igvm = { version = "0.3.4", default-features = false }
+igvm_defs = { version = "0.3.4", default-features = false }
 intrusive-collections = "0.9.6"
 libfuzzer-sys = "0.4"
 log = "0.4.17"
--- a/bootlib/src/igvm_params.rs
+++ b/bootlib/src/igvm_params.rs
@@ -136,11 +136,18 @@ pub struct IgvmParamBlock {
     /// memory region (e.g. for VMSA contents).
     pub kernel_reserved_size: u32,
 
-    /// The number of bytes in the kernel memory region.
-    pub kernel_size: u32,
-
     /// The guest physical address of the base of the kernel memory region.
     pub kernel_base: u64,
+
+    /// The minimum size to allocate for the kernel in bytes. If the hypervisor supplies a memory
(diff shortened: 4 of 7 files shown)


kernel/src/igvm_params.rs
```diff
@@ -75,6 +75,19 @@ impl IgvmParams<'_> {
     }
 
     pub fn find_kernel_region(&self) -> Result<MemoryRegion<PhysAddr>, SvsmError> {
+        // First check the memory map if there's a paravisor region to use.
```
msft-jlange, line 78: I don't think this can safely be done first.  The contents of the IGVM memory map are untrusted (they are provided by the host) which means that examining this list first will give the host the opportunity to override any region specifically chosen in the IGVM file, even if the guest is expecting to use the specifically described region.  I think there are only two safe ways to handle this.  Either (1) the IGVM-selected region should always be used if present, and the region proposed by the memory map should only be used if no IGVM-selected region is present, or (2) the IGVM parameter block in the IGVM file should specify whether the memory map should be consulted to preempt the region specified in the IGVM file.

kernel/src/igvm_params.rs
```diff
@@ -75,6 +75,19 @@ impl IgvmParams<'_> {
...
+        for entry in self.igvm_memory_map.memory_map.iter() {
+            if entry.entry_type == MemoryMapEntryType::PARAVISOR_RESERVED {
+                let starting_page: usize = entry.starting_gpa_page_number.try_into().unwrap();
+                let number_of_pages: usize = entry.number_of_pages.try_into().unwrap();
+                return Ok(MemoryRegion::new(
```
msft-jlange, line 83: This logic will always latch onto the first memory region that is described in the memory map as reserved for the paravisor.  Is that the right choice?  Is it better to find the biggest region?  Regardless of what choice is made, there should be a comment to acknowledge the fact that since the memory map is untrusted, there may be multiple paravisor regions present, and the comment should indicate the reasoning for whatever region is selected (even if the reasoning is cursory).

msft-jlange (changes requested):

kernel/src/igvm_params.rs
AdamCDunlap: If any supplied region is unsafe then I'd prefer to fix that. Having the param block override the hypervisor preference doesn't seem necessarily bad, but it does seem to indicate a lack of trust in the safety of supplying it.

Is there a case where the SVSM really needs to be at a specific location and is willing to fail if that location isn't available? I thought it wouldn't really care and the region being in the param block seems to support that.

msft-jlange: > If any supplied region is unsafe then I'd prefer to fix that. Having the param block override the hypervisor preference doesn't seem necessarily bad, but it does seem to indicate a lack of trust in the safety of supplying it.

This is the essence of confidential computing: the hypervisor is not trusted to supply safe configuration information; the guest VM has to take responsibility for ensuring that the hypervisor-provided configuration is reasonable and safe.  That doesn't necessarily mean that the hypervisor-provided configuration should not be the first choice, but it does mean that the SVSM must apply some sort of safety checks to ensure that the configuration is reasonable.  Another way to think about this is that the SVSM must be designed to be resilient in the face of fuzzing all of its inputs.

> I thought it wouldn't really care and the region being in the param block seems to support that.

The existing parameter block is measured, so it is not susceptible to modification by the hypervisor.  It's called `IgvmParamBlock` because it's a set of parameters embedded in the IGVM file (calculated by the file builder), not because it contains parameters inserted by the host through IGVM conventions (I agree that this naming may be confusing).  Because the existing parameters are measured, we cannot infer that the existing code doesn't care what their values are.

> Is there a case where the SVSM really needs to be at a specific location and is willing to fail if that location isn't available? 

The Hyper-V virtual firmware has some strong opinions about where the SVSM image must *not* be placed, and while that isn't exactly the same thing as relying on a specific location, there is some information there that needs to be conveyed about avoiding problematic collisions.  And the Hyper-V virtual firmware is designed to rely on a paravisor - meaning that it has no logic (and will never have logic) to ensure that its memory configuration is suitable, because it relies on the paravisor to provide that safety net.  I do think we can safely say that as long as the SVSM image doesn't collide with a no-trespassing region, then the firmware won't care where it is eventually placed.

Outside of Hyper-V, I think it's safe to say that every environment has at least some opinions about where the SVSM memory must not be placed.  It must not be placed to overlap any firmware image that is embedded in the IGVM file, because that would cause an unexpected overwrite of guest memory.  It must not be placed to overlap the stage2 area, because that would cause stage2 to overwrite itself during the relocation process.  It must not be placed to overlap the initial kernel or filesystem image, because that would cause corruption during the relocation process.  Some of this information can be deduced by stage2 today without needing additional information in the IGVM parameter block (though additional code would be required to calculate it) and some of it cannot be (for example, there is no way for the SVSM to know where the IGVM-embedded firmware is placed into memory unless the IGVM parameter block describes it).

So perhaps what we really need is to extend the `IgvmParamBlock` and/or `IgvmParamBlockFwInfo` to embed some small, fixed number of exclusion ranges (calculated by the IGVM file builder) that `find_kernel_region` can use for safety checks.  We can debate later whether exclusion should be enforced by ignoring regions that overlap the exclusion zone(s), or whether candidate regions should be trimmed to avoid intersection with an exclusion zone, or whether any observed overlap should result in immediate panic; the important thing to start with is a mechanism for defining safe/unsafe regions that can be enforced by the SVSM.

AdamCDunlap: I thought that we would be covered because stage2 already checks that the regions in the memory map do not overlap, but I no longer believe that is sufficient since there are other important regions that may not be in the memory map. Thanks for providing examples of when that's true.

Instead, I've redesigned the idea completely so that the IgvmParamBlock's kernel_start is always used, and the kernel_size is still used but it changes to act as a maximum kernel size. This means that the measured IgvmParamBlock will still ensure the kernel region does not overlap anything important but still allows the hypervisor to shrink the kernel size on smaller machine shapes which is the real goal.

I'm open to renaming kernel_size to max_kernel_size or adding a min_kernel_size as well.

msft-jlange: > I've redesigned the idea completely so that the IgvmParamBlock's kernel_start is always used, and the kernel_size is still used but it changes to act as a maximum kernel size. This means that the measured IgvmParamBlock will still ensure the kernel region does not overlap anything important but still allows the hypervisor to shrink the kernel size on smaller machine shapes which is the real goal.

I really like the new direction.

I believe it will be important to specify a minimum size for the kernel region.  The SVSM will definitely have an opinion about the minimum amount of memory that will be required for it to do anything useful, and that minimum will be based on the set of services that are composed into the image - meaning it must be part of the IGVM packaging, and not something that can be inferred entirely by kernel code.  Note that there will be some scaling calculations that the kernel will do on its own (like the amount of memory that will be required for page validation tracking bitmaps, which will depend on the memory map that is presented), but that isn't enough to calculate the correct minimum size.

To that point, I'm not convinced that stage2 is the right place to pin down the kernel memory region size, because the calculation of required bitmaps etc. will happen in the kernel itself (we need kernel code to be running in order to figure out what configuration it wants).  However, the current code structure requires the kernel to be a fixed size which is calculated in stage2, so I think your approach is good until such time in the future as we can overhaul the kernel's heap management.  We may need to add some interim logic in the kernel to ensure that the size selected by stage2 is going to work, but that's easy to add later on top of your code.

It's harder for me to understand why there is a practical maximum size.  The kernel doesn't have an upper bound on the amount of memory that it can manage as private memory, so if the host wants to suggest that the SVSM should consume huge amounts of memory for private use, then this shouldn't be any worse than poor resource management policy, which is outside the scope of security enforcement.  So I would argue against defining an upper bound unless you can help me see what I'm missing.

> I'm open to renaming kernel_size to max_kernel_size or adding a min_kernel_size as well.

For the sake of clarity, we should make sure the size field has a name that describes its meaning.  So `kernel_size` is insufficient, and we will need `min_kernel_size` and/or `max_kernel_size`.

AdamCDunlap: > The SVSM will definitely have an opinion about the minimum amount of memory that will be required for it to do anything useful

My thought was that if it didn't have enough space, it would panic when it tries to use the space. Note that even if it says "I must have at least X bytes" and the hypervisor doesn't give it that much space, there's not much it can do anyway except possibly panic earlier. Or the hypervisor could be lying and say it has X space, but it actually gave it less. A minimum size could be useful if (a) it could do something interesting (maybe a meaningful error message?) if given less than it or (b) it would misbehave (not just crash) if it was told it was given less than that. Or if the hypervisor could read that minimum size and use it to inform how big of a region it would allocate (this isn't practical for us, though).

> It's harder for me to understand why there is a practical maximum size.

If SVSM wants to put the kernel at address 0x1000 and there's Something Important at 0x2000, then we'd want the maximum size to ensure the kernel doesn't overwrite the important thing. If SVSM is expected to exist at a higher address than anything else important then the maximum size could just be a few terabytes.

> So kernel_size is insufficient, and we will need min_kernel_size and/or max_kernel_size.

Will do in the next version.

msft-jlange: > My thought was that if it didn't have enough space, it would panic when it tries to use the space.

This assumes that all of the code paths that are subject to allocation failure will correctly flow the failure back to a point where they can panic or otherwise fail gracefully.  I have a painful amount of experience with systems that don't do this correctly, i.e. they will fall over in mysterious ways when memory allocations fail.  The memory safety guarantees of Rust make this much less likely in COCONUT-SVSM, but I still get very nervous about proving that every allocation failure path is robust and can never result in insecure behavior.  This becomes much more worrisome once we introduce fallible allocators, which are most definitely on the roadmap.

> Note that even if it says "I must have at least X bytes" and the hypervisor doesn't give it that much space, there's not much it can do anyway except possibly panic earlier.

Panicking earlier means all of the panics are concentrated at one point that can be proven to be robust, so I claim that panicking earlier is a much better option.

> > It's harder for me to understand why there is a practical maximum size.
> If SVSM wants to put the kernel at address 0x1000 and there's Something Important at 0x2000, then we'd want the maximum size to ensure the kernel doesn't overwrite the important thing. If SVSM is expected to exist at a higher address than anything else important then the maximum size could just be a few terabytes.

This is an excellent point, so I agree that we need a maximum size.

AdamCDunlap: Updated PR to specify a minimum and maximum size in the IgvmBlockParam

msft-jlange (approved):
Thanks for getting this done!

AdamCDunlap:
Thanks for your help!

PR #593, opened 2025-01-20
task/schedule: Free init stack during first context switch

peterfang:
`init_stack` is now completely unused by the BSP and stops being used by any CPU after switching to the idle task in `schedule_init()`. Free up these memory pages during context switching. Since a Rust function is called for this purpose, make sure callee-saved registers are used for `switch_context`'s input arguments.


joergroedel (approved):
Changes look good to me, thanks.

Issue #591, opened 2025-01-09
Multiple soundness issues with `ImmutAfterInitCell`


msft-jlange:
`ImmutAfterInitCell` both implements `Deref` and has a method called `reinit`, which permits reinitialization of the contents of the cell.  In combination, this means it is possible for `reinit` to be called in one context while another context holds a shared (immutable) reference to the cell.  This violates borrowing safety rules.  At a minimum, `reinit` must be declared as `unsafe` because it is not safe, but it would be better to prohibit reinitialization altogether.

The one-time initialization checks for `ImmutAfterInitCell` are limited to debug builds, which means that non-debug builds are unsound because there are no initialization checks.  If there is a code path which can be invoked in non-debug builds which can attempt to reference an uninitialized cell or to reinitialize an initialized cell, then no checking is performed, leading to unsound behavior.

`Sync` guarantees around initialization depend on a point in time where multiple CPUs are started, under the assumption that cross-thread safety is maintained as long as only one CPU can execute.  However, the ability to invoke exception handlers (like the #VC handler) means that the exception handler can attempt to interact an `ImmutAfterInitCell` which is also being modified by the code that was interrupted, and that means that `Sync` rules need to apply even before multiple processors have started.

We need to fix `ImmutAfterInitCell` so that
- it is truly immutable after initialization (no reinitialize method)
- initialization checks are always performed
- it is `Sync` in all cases

This is similar to the requirements of `std::OnceLock`, except that is not available in `nostd` so we need a custom implementation.

msft-jlange:
Fixed with #595.

PR #595, opened 2025-01-22
utils: overhaul `ImmutAfterInitCell`

msft-jlange:
This change updates the design of `ImmutAfterInitCell` to resolve the issues described in #591, making it behave much more like a `Sync` version of `OnceCell`.


kernel/src/console.rs
```diff
@@ -34,19 +35,19 @@ impl Console {
...
     writer: &DEFAULT_SERIAL_PORT,
 });
-static CONSOLE_INITIALIZED: ImmutAfterInitCell<bool> = ImmutAfterInitCell::new(false);
+static CONSOLE_INITIALIZED: AtomicBool = AtomicBool::new(false);
 static CONSOLE_SERIAL: ImmutAfterInitCell<SerialPort<'_>> = ImmutAfterInitCell::uninit();
 
 fn init_console(writer: &'static dyn Terminal) -> ImmutAfterInitResult<()> {
     WRITER.lock().writer = writer;
-    CONSOLE_INITIALIZED.reinit(&true)?;
+    CONSOLE_INITIALIZED.store(true, Ordering::Relaxed);
```
joergroedel, line 43: `Ordering::Relaxed` is okay because the write to `WRITER` is locked separately, right? Please add a comment here as a reminder.

kernel/src/utils/immut_after_init.rs
```diff
@@ -52,141 +42,136 @@ pub enum ImmutAfterInitError {
...
-    fn set_init(&self) {
-        #[cfg(debug_assertions)]
-        self.init.store(true, Ordering::Relaxed);
-    }
-
     fn check_init(&self) -> ImmutAfterInitResult<()> {
-        #[cfg(debug_assertions)]
-        if !self.init.load(Ordering::Relaxed) {
-            return Err(ImmutAfterInitError::Uninitialized);
+        if self.init.load(Ordering::Relaxed) == IMMUT_INITIALIZED {
```
joergroedel, line 72: I think this requires `Ordering::Acquire` to make sure all changes before the `Ordering::Release` store in `complete_init` are visible.

kernel/src/utils/immut_after_init.rs
```diff
@@ -52,141 +42,136 @@ pub enum ImmutAfterInitError {
...
-        if self.init.load(Ordering::Relaxed) {
-            return Err(ImmutAfterInitError::AlreadyInit);
-        }
+    fn try_init(&self) -> ImmutAfterInitResult<()> {
+        self.init
+            .compare_exchange(
+                IMMUT_UNINIT,
+                IMMUT_INIT_IN_PROGRESS,
+                Ordering::Acquire,
+                Ordering::Relaxed,
```
joergroedel, line 85: I am not sure about this `Ordering::Relaxed` either, as I have seen the compiler eliminating an `Ordering::Acquire` load in favor of a previous `Ordering::Release` load.

joergroedel (commented):
Overall the changes look good to me. There are a few open questions about using atomics and ordering I'd like to get sorted before merging.

kernel/src/console.rs
msft-jlange: It's really because `CONSOLE_INITIALIZED` is always initialized while the system is still running single-proc, so no out-of-order observation is possible.  I'll add a comment to this effect.

kernel/src/utils/immut_after_init.rs
msft-jlange: You are correct.

kernel/src/utils/immut_after_init.rs
msft-jlange: `compare_exchange()` returns different values depending on whether the compare/exchange succeeds.  If the compare/exchange succeeds, it returns `Ok()` with the previous value (which was definitively known because it was part of the exchange), and if it fails, it will return `Err()` with the new value.  The second `Ordering` here describes the ordering that will be used to obtain the `Err()` value.  Since this code does not use the `Err()` value, the ordering is irrelevant, and thus no fence is required.

joergroedel (approved):
LGTM.

PR #598, opened 2025-01-23
Add support for TD SMP

peterfang:
This patch series reuses the logic introduced in https://github.com/coconut-svsm/svsm/pull/581. A software-based mailbox is used for AP signaling. TD APs will enter `ap_request_loop()` after this series is applied.

### DIFF
--- a/kernel/src/boot_stage2.rs
+++ b/kernel/src/boot_stage2.rs
@@ -5,14 +5,25 @@
 // Author: Joerg Roedel <jroedel@suse.de>
 
 use core::arch::global_asm;
+use core::mem::{offset_of, size_of};
 
+use bootlib::kernel_launch::{ApStartContext, STAGE2_HEAP_START};
 use svsm::{
     cpu::{
         efer::EFERFlags,
         msr::{EFER, SEV_STATUS},
     },
     mm::PGTABLE_LVL3_IDX_PTE_SELFMAP,
+    platform::tdp::TdMailbox,
+    types::PAGE_SIZE,
 };
+
+const MAILBOX_ADDR: u32 = bootlib::kernel_launch::SIPI_STUB_GPA;
+const AP_CTXT_ADDR: u32 = MAILBOX_ADDR + PAGE_SIZE as u32 - size_of::<ApStartContext>() as u32;
+
+// The mailbox page must be within the first 640 KB of memory so it can be
+// validated by the BSP in stage2. It also can't be in stage2's heap area.
+const _: () = assert!(MAILBOX_ADDR + (PAGE_SIZE as u32) <= STAGE2_HEAP_START);
 
 global_asm!(
     r#"
@@ -26,6 +37,7 @@ global_asm!(
 
         /* Save pointer to startup structure in EBP */
         movl %esp, %ebp
+
         /*
          * Load a GDT. Despite the naming, it contains valid
          * entries for both, "legacy" 32bit and long mode each.
@@ -98,19 +110,16 @@ global_asm!(
         orl %edx, %eax
         movl %eax, 0xF6C(%edi)
 
-        /* Signal APs */
-        movl $setup_flag, %edi
-        movl $1, (%edi)
         jmp 2f
 
-.Lskip_paging_setup:
-        movl $setup_flag, %edi
-.Lap_wait:
-        movl (%edi), %eax
-        test %eax, %eax
-        jz .Lap_wait
+    .Lskip_paging_setup:
+        movl $ap_flag, %edi
+    .Lap_wait_for_env:
+        cmpl $0, (%edi)
+        pause
+        je .Lap_wait_for_env
 
-2:
+    2:
         /* Enable 64bit PTEs, CR4.PAE. */
         movl %cr4, %eax
         bts $5, %eax
@@ -122,10 +131,9 @@ global_asm!(
         movl %eax, %ebx
         orl $({LME} | {NXE}), %eax
         cmp %eax, %ebx
-        jz 2f
+        jz 3f
         wrmsr
-        2:
-
+    3:
         /* Load the static page table root. */
         movl $pgtable, %eax
         movl %eax, %cr3
@@ -241,9 +249,23 @@ global_asm!(
         test %esi, %esi
         jz .Lbsp_main
 
-    .Lcheck_command:
-        /* TODO */
-        jmp .Lcheck_command
+        movl ${MAILBOX_VPIDX_ADDR}, %eax
+
+        /*
+         * The following wait-for-signal code must be written in asm because
+         * APs run stacklessly here.
+         */
+    .Lap_wait_for_signal:
+        /* vCPU index */
+        cmpl %esi, (%eax)
+        pause
+        jne .Lap_wait_for_signal
+
+        /* See smp.rs for prerequisites. */
+        movl ${AP_CTXT_ADDR}, %edi
+        movl ${AP_STARTUP_RIP_ADDR}, %eax
+        movq (%rax), %rax
+        jmp *%rax
 
     .Lbsp_main:
         /* Clear out .bss and transfer control to the main stage2 code. */
@@ -268,7 +290,8 @@ global_asm!(
         .data
 
         .align 4
-    setup_flag:
+        .globl ap_flag
+    ap_flag:
         .long 0
 
     idt32:
@@ -304,15 +327,19 @@ global_asm!(
         .word gdt64_end - gdt64 - 1
         .quad gdt64
 
-        .align 4096
+        .align {PAGE_SIZE}
         .globl pgtable
     pgtable:
-        .fill 7 * 4096, 1, 0
+        .fill 7*{PAGE_SIZE}, 1, 0
     pgtable_end:"#,
+    PAGE_SIZE = const PAGE_SIZE,
     PGTABLE_LVL3_IDX_PTE_SELFMAP = const PGTABLE_LVL3_IDX_PTE_SELFMAP,
     EFER = const EFER,
     LME = const EFERFlags::LME.bits(),
     NXE = const EFERFlags::NXE.bits(),
     SEV_STATUS = const SEV_STATUS,
+    MAILBOX_VPIDX_ADDR = const MAILBOX_ADDR + offset_of!(TdMailbox, vcpu_index) as u32,
+    AP_CTXT_ADDR = const AP_CTXT_ADDR,
+    AP_STARTUP_RIP_ADDR = const AP_CTXT_ADDR + offset_of!(ApStartContext, initial_rip) as u32,
     options(att_syntax)
 );
--- a/kernel/src/platform/tdp.rs
+++ b/kernel/src/platform/tdp.rs
@@ -9,10 +9,12 @@ use crate::console::init_svsm_console;
 use crate::cpu::cpuid::CpuidResult;
 use crate::cpu::msr::read_msr;
 use crate::cpu::percpu::PerCpu;
+use crate::cpu::smp::create_ap_start_context;
 use crate::cpu::x86::apic::{apic_register_bit, APIC_MSR_ISR};
 use crate::error::SvsmError;
 use crate::hyperv;
 use crate::io::IOPort;
+use crate::mm::PerCPUPageMappingGuard;
 use crate::platform::{PageEncryptionMasks, PageStateChangeOp, PageValidateOp, SvsmPlatform};
 use crate::tdx::tdcall::{
     td_accept_physical_memory, td_accept_virtual_memory, tdvmcall_halt, tdvmcall_io_read,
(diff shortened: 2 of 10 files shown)


kernel/src/platform/tdp.rs
```diff
@@ -20,13 +22,39 @@ use crate::tdx::TdxError;
...
+// Both structures must fit in a page
+const _: () = assert!(mem::size_of::<TdMailbox>() + mem::size_of::<ApStartContext>() <= PAGE_SIZE);
+
+// SAFETY: caller must ensure `mailbox` points to a valid memory address.
+unsafe fn wakeup_ap(mailbox: *mut TdMailbox, apic_id: u32) {
+    // SAFETY: caller must ensure the address is valid and not aliased.
+    unsafe {
+        (*mailbox).apic_id = 0;
+        (*mailbox).command = 0;
+        _mm_sfence();
```
msft-jlange, line 51: The store fences here are not required because Intel platforms guarantee that writes from one processor are always visible to other processors in the order they appear in the program.  A store fence would only be required if a subsequent read in this routine could depend on another processor observing these stores completing, but there are no subsequent reads here.

kernel/src/boot_stage2.rs
```diff
@@ -241,9 +250,32 @@ global_asm!(
...
+         */
     .Lcheck_command:
-        /* TODO */
-        jmp .Lcheck_command
+        /* Command */
+        movl (%ecx), %eax
+        test %eax, %eax
+        jz .Lcheck_command
+
+        lfence
```
msft-jlange, line 266: These load fences are unnecessary for the same reason that the store fences are unnecessary (see my other comments).

kernel/src/boot_stage2.rs
```diff
@@ -241,9 +250,32 @@ global_asm!(
...
+
+        /* APIC ID */
+        movl (%edx), %eax
+        cmp %esi, %eax
```
msft-jlange, line 270: This logic appears to assume that the VP index in RSI will always have the same value as the APIC ID.  That is not the case on NUMA systems that do not have a power-of-two CPU count in each node.  The NUMA APIC architecture specifies that the APIC ID is divided into some bits that represent the processor node and some other bits that are unique within the node.  If, for example, there are 12 vCPUs on a two-node system, then APIC IDs will range from 0-5 and from 8-13 (there will be no 6 or 7).

I think the only way this will work is if the BSP passes the VP index and not the APIC ID.  This is available as `cpu.shared().cpu_index()`.

kernel/src/boot_stage2.rs
```diff
@@ -241,9 +250,32 @@ global_asm!(
...
+        movl ${MAILBOX_APICID_ADDR}, %edx
+
+        /*
+         * The following wait-for-signal code must be written in asm because
+         * APs run stacklessly here.
+         */
     .Lcheck_command:
-        /* TODO */
-        jmp .Lcheck_command
+        /* Command */
```
msft-jlange, line 261: Why is it necessary to have both a command and a VP index?  There is only one command, meaning "start", so isn't it sufficient for each vCPU to spin until it sees its APIC ID in the mailbox without also needing to compare a command identifier?

kernel/src/boot_stage2.rs
```diff
@@ -98,19 +110,16 @@ global_asm!(
...
-        /* Signal APs */
-        movl $setup_flag, %edi
-        movl $1, (%edi)
         jmp 2f
 
-.Lskip_paging_setup:
-        movl $setup_flag, %edi
-.Lap_wait:
+    .Lskip_paging_setup:
+        movl $ap_flag, %edi
```
msft-jlange, line 116: Why is this flag necessary?  The BSP will never write a non-zero VP index into the mailbox until it is ready to start APs, so shouldn't the VP index check be enough to guarantee that all APs spin until they are ready to start?

kernel/src/platform/tdp.rs
peterfang: Thanks for pointing this out. I was overly cautious with the semantics and not taking Intel's architecture into consideration. Since this page is WB it will respect Intel's memory ordering properties. I do see that this could still be reordered by the Rust compiler so I'll change them to compiler barriers.

kernel/src/boot_stage2.rs
peterfang: I'll remove the lfences here. No compiler barriers are needed since this is already asm code.

kernel/src/boot_stage2.rs
peterfang: Thanks for the reminder. `cpu.shared().cpu_index()` makes sense because it maps to the vCPU index assigned by the TDX module.
QEMU always assigns APIC IDs according to CPU topology so my code wouldn't work for fwcfg boot (which isn't supported by TDP right now). However I don't see such logic in `IgvmParams::load_cpu_info()`. Is that subject to change at some point?

kernel/src/boot_stage2.rs
peterfang: This is a much-simplified version of the TD mailbox implemented in OVMF and the original version has a command field for operations other than wake-up. It is indeed not needed here. I'll remove it.

kernel/src/boot_stage2.rs
peterfang: Correct, APs should just spin on the VP index field and this is not needed. I'll remove it.

kernel/src/platform/tdp.rs
msft-jlange: Once you remove the command field and retain only the VP index, then there's only one write in this function and thus there are no ordering concerns that require decoration for the compiler.  If you really want to force the value to be observed immediately, then you should define the mailbox to define the VP index as `AtomicU32` and you can write it with `Ordering::Release` semantics.  It's not clear to me whether this strictly necessary to remain within the scope of defined behavior.

kernel/src/boot_stage2.rs
msft-jlange: The right way to associate VP index and ACPI ID is through the MADT.  At some point, `IgvmParams::load_cpu_info()` needs to be updated to do that (which should not be excessively difficult because the MADT is available as an IGVM parameter).

kernel/src/platform/tdp.rs
peterfang: Ah, you're absolutely correct. There's no need to reset the VP index to 0 first. Thanks for making this much simpler :-)

kernel/src/boot_stage2.rs
msft-jlange: Now that the mailbox has been reduced simply to the VP index, why not just have `ap_flag` become the VP index instead?  That would eliminate the need to have a separate mailbox page as well as any page table construction associated with that page.

peterfang: That's a great idea. However that means I'd have to pass the address of this VP index field all the way to `start_secondary_cpus()` which is something I'd like to avoid if possible. Using the mailbox page is convenient since `SIPI_STUB_GPA` is already being used for native AP bring-up and `ApStartContext` would end up being at the same location. Do you prefer this implementation?

msft-jlange: I see your point, and it's one I didn't have to contend with when doing the SIPI work because that only applies to non-confidential VMs (where there is no concept of accept/validate).  With that in mind, I think your use of `ap_flag` is fine, and is simpler than trying to pass its address all the way down to the AP start loop.

peterfang: Sounds good!

kernel/src/boot_stage2.rs
```diff
@@ -98,19 +110,16 @@ global_asm!(
...
+    .Lap_wait_for_env:
         movl (%edi), %eax
         test %eax, %eax
-        jz .Lap_wait
+        jz .Lap_wait_for_env
```
msft-jlange, line 120: Can you add a `PAUSE` in the spin loop?  There are scenarios in which this helps with spin loop management.

kernel/src/stage2.rs
```diff
@@ -43,6 +46,7 @@ use svsm::utils::{is_aligned, MemoryRegion};
 use release::COCONUT_VERSION;
 
 extern "C" {
+    static mut ap_flag: u32;
```
msft-jlange, line 49: You might consider making this `static ap_flag: AtomicU32`.  If you change the later store to read `ap_flag.store(1, Ordering::Release)` then you can avoid the use of `unsafe` (and you will also not need to worry about compiler ordering).

kernel/src/boot_stage2.rs
peterfang: Sure, that's a great suggestion.

kernel/src/stage2.rs
peterfang: Unfortunately this still requires `unsafe` due to it being an extern static:

    error[E0133]: use of extern static is unsafe and requires unsafe block
       --> kernel/src/stage2.rs:131:5
        |
    131 |     ap_flag.store(1, Ordering::Release);
        |     ^^^^^^^ use of extern static
        |
        = note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior

msft-jlange: I see what you mean.  My preference would still be to define this as an atomic because the cross-thread data dependency is much more clear than a simple `mut` variable (and the ordering semantics of the write make the compiler's code more predictable) but I don't think it's a high priority.

peterfang: Making sure the compiler won't be reordering this store sounds like a good benefit to me. We can continue using the non-atomic load on the AP side due to x86's memory ordering properties. I'll make this change.

kernel/src/boot_stage2.rs
```diff
@@ -98,19 +110,18 @@ global_asm!(
...
-.Lap_wait:
+    .Lskip_paging_setup:
+        movl $ap_flag, %edi
+    .Lap_wait_for_env:
         movl (%edi), %eax
         test %eax, %eax
-        jz .Lap_wait
+        jnz 2f
+        pause
+        jmp .Lap_wait_for_env
```
msft-jlange, line 122:
```suggestion
        pause
        jz .Lap_wait_for_env
```

There's nothing wrong with executing `PAUSE` if the loop isn't taken; `PAUSE` hint exiting occurs when the same `PAUSE` is run too many times in a row.  If you follow my suggestion, you can drop an extra branch and make the code tighter.

kernel/src/boot_stage2.rs
```diff
@@ -241,9 +251,26 @@ global_asm!(
...
+         * The following wait-for-signal code must be written in asm because
+         * APs run stacklessly here.
+         */
+    .Lap_wait_for_signal:
+        /* vCPU index */
+        movl (%ecx), %eax
+        cmp %esi, %eax
+        je 4f
+        pause
+        jmp .Lap_wait_for_signal
```
msft-jlange, line 266:
```suggestion
        pause
        jne .Lap_wait_for_signal
```

Same as earlier suggestion.

kernel/src/boot_stage2.rs
```diff
@@ -98,19 +110,18 @@ global_asm!(
...
         jmp 2f
 
-.Lskip_paging_setup:
-        movl $setup_flag, %edi
-.Lap_wait:
+    .Lskip_paging_setup:
+        movl $ap_flag, %edi
+    .Lap_wait_for_env:
         movl (%edi), %eax
         test %eax, %eax
```
msft-jlange, line 119:
```suggestion
        cmpl (%edi), $0
```

It's better to perform the comparison in a single instruction instead of load followed by a compare.  This code predates your change, but you may as well clean it up while you're in this code.

kernel/src/boot_stage2.rs
peterfang: Makes sense. I'll do that.

kernel/src/boot_stage2.rs
peterfang: Sure I'll clean that up.

kernel/src/platform/tdp.rs
```diff
@@ -22,13 +24,33 @@ use crate::tdx::TdxError;
...
+use core::{mem, ptr};
 
 #[cfg(test)]
 use bootlib::platform::SvsmPlatformType;
 
 static GHCI_IO_DRIVER: GHCIIOPort = GHCIIOPort::new();
 static VTOM: ImmutAfterInitCell<usize> = ImmutAfterInitCell::uninit();
 
+#[derive(Debug)]
+#[repr(C, packed)]
```
msft-jlange, line 37: I find that `packed` creates compilation issues because the compiler will assume that all fields are unaligned (which isn't true), and therefore `packed` should only be used if it's really necessary.  A structure that has only a single field will never encounter packing issues, so it clearly isn't necessary.  But it's also not a problem here, so I don't think you need to change it - but you should think about it for the future.

msft-jlange (approved):

kernel/src/platform/tdp.rs
peterfang: Thanks for the reminder. I used `packed` for the first version of the mailbox implementation because it's used in assembly code and has multiple fields in the struct. That said, we can now use `offset_of!` in complex cases to make sure the offset is always correct so perhaps it's not needed as much even when there's dependency on assembly code.

PR #600, opened 2025-01-23
cpu: restrict IPI-based TLB flush to multi-proc environments

msft-jlange:
To accommodate platforms that cannot make use of SVSM interrupts, TLB flushes should avoid attempts to send flush IPIs unless a second CPU is actually started.


joergroedel (approved):
LGTM.

PR #601, opened 2025-01-23
platform: implement `SvsmPlatform::is_external_interrupt` for TDP.

msft-jlange:
The TDP platform must use APIC virtualization to check whether a given interrupt event originates from a hardware interrupt or a software interrupt.  This change implements the logic to read the required state from the APIC.


kernel/src/cpu/x86/apic.rs
```diff
@@ -0,0 +1,16 @@
...
+// Author: Jon Lange <jlange@microsoft.com>
+
+pub const APIC_MSR_EOI: u32 = 0x80B;
+pub const APIC_MSR_ISR: u32 = 0x810;
+pub const APIC_MSR_ICR: u32 = 0x830;
+
+// Returns the MSR offset and bitmask to identify a specific vector in an
+// APIC register (IRR, ISR, or TMR).
+pub fn apic_register_bit(vector: usize) -> (u32, u32) {
+    let index: u8 = (vector & 0xFF) as u8;
```
peterfang, line 14: nit: I believe the following would suffice since Rust always truncates in this case
```suggestion
    let index = vector as u8;
```

msft-jlange: You are correct.

peterfang:
Looks good to me, thanks!

PR #603, opened 2025-01-28
kernel/vtpm: fix potential UB when accessing guest buffer

stefano-garzarella:
Creating a mutable reference to the guest buffer, could have been an UB because the guest could potentially modify it.

So, let's copy the guest buffer to an internal one and then copy it back to give the result to the guest. To avoid overflowing the stack, allocate the buffer with Vec.

### DIFF
--- a/kernel/src/protocols/vtpm.rs
+++ b/kernel/src/protocols/vtpm.rs
@@ -107,27 +107,17 @@ impl TpmSendCommandRequest {
             return Err(SvsmReqError::invalid_parameter());
         }
 
-        let mut length = self.inbuf_size as usize;
+        let length = self.inbuf_size as usize;
 
         let tpm_cmd = self
             .inbuf
             .get(..length)
             .ok_or_else(SvsmReqError::invalid_parameter)?;
-        let mut buffer: Vec<u8> = Vec::with_capacity(SEND_COMMAND_RESP_OUTBUF_SIZE);
-        buffer.extend_from_slice(tpm_cmd);
-
-        // The buffer slice must be large enough to hold the TPM command response
-        buffer.resize(SEND_COMMAND_RESP_OUTBUF_SIZE, 0);
 
         let vtpm = vtpm_get_locked();
-        vtpm.send_tpm_command(buffer.as_mut_slice(), &mut length, self.locality)?;
+        let response = vtpm.send_tpm_command(tpm_cmd, self.locality)?;
 
-        if length > buffer.len() {
-            return Err(SvsmReqError::invalid_request());
-        }
-        buffer.truncate(length);
-
-        Ok(buffer)
+        Ok(response)
     }
 }
 
--- a/kernel/src/protocols/vtpm.rs
+++ b/kernel/src/protocols/vtpm.rs
@@ -8,9 +8,8 @@
 
 extern crate alloc;
 
-use core::mem::size_of;
-
 use alloc::vec::Vec;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
 use crate::{
     address::{Address, PhysAddr},
@@ -71,8 +70,8 @@ const SVSM_VTPM_QUERY: u32 = 0;
 const SVSM_VTPM_COMMAND: u32 = 1;
 
 /// TPM_SEND_COMMAND request structure (SVSM spec, table 16)
-#[derive(Clone, Copy, Debug)]
 #[repr(C, packed)]
+#[derive(FromBytes, IntoBytes, Immutable, KnownLayout, Clone, Copy, Debug)]
 struct TpmSendCommandRequest {
     /// MSSIM platform command ID
     command: u32,
@@ -87,15 +86,8 @@ struct TpmSendCommandRequest {
 impl TpmSendCommandRequest {
     // Take as slice and return a reference for Self
     pub fn try_from_as_ref(buffer: &[u8]) -> Result<&Self, SvsmReqError> {
-        let buffer = buffer
-            .get(..size_of::<Self>())
-            .ok_or_else(SvsmReqError::invalid_parameter)?;
-
-        // SAFETY: TpmSendCommandRequest has no invalid representations, as it
-        // is comprised entirely of integer types. It is repr(packed), so its
-        // required alignment is simply 1. We have checked the size, so this
-        // is entirely safe.
-        let request = unsafe { &*buffer.as_ptr().cast::<Self>() };
+        let request =
+            Self::ref_from_bytes(buffer).map_err(|_| SvsmReqError::invalid_parameter())?;
 
         if !request.validate() {
             return Err(SvsmReqError::invalid_parameter());
@@ -130,8 +122,8 @@ impl TpmSendCommandRequest {
 const SEND_COMMAND_RESP_OUTBUF_SIZE: usize = PAGE_SIZE - 4;
 
 /// TPM_SEND_COMMAND response structure (SVSM spec, table 17)
-#[derive(Clone, Copy, Debug)]
 #[repr(C, packed)]
+#[derive(FromBytes, IntoBytes, Immutable, KnownLayout, Clone, Copy, Debug)]
 struct TpmSendCommandResponse {
     /// Size of the output buffer
     outbuf_size: u32,
@@ -142,17 +134,7 @@ struct TpmSendCommandResponse {
 impl TpmSendCommandResponse {
     // Take as slice and return a &mut Self
     pub fn try_from_as_mut_ref(buffer: &mut [u8]) -> Result<&mut Self, SvsmReqError> {
-        let buffer = buffer
-            .get_mut(..size_of::<Self>())
-            .ok_or_else(SvsmReqError::invalid_parameter)?;
-
-        // SAFETY: TpmSendCommandResponse has no invalid representations, as it
-        // is comprised entirely of integer types. It is repr(packed), so its
-        // required alignment is simply 1. We have checked the size, so this
-        // is entirely safe.
-        let response = unsafe { &mut *buffer.as_mut_ptr().cast::<Self>() };
-
-        Ok(response)
+        Self::mut_from_bytes(buffer).map_err(|_| SvsmReqError::invalid_parameter())
     }
 
     /// Write the response to the outbuf
--- a/kernel/src/protocols/vtpm.rs
+++ b/kernel/src/protocols/vtpm.rs
@@ -8,13 +8,16 @@
 
 extern crate alloc;
 
-use core::{mem::size_of, slice::from_raw_parts_mut};
+use core::mem::size_of;
 
 use alloc::vec::Vec;
 
 use crate::{
     address::{Address, PhysAddr},
-    mm::{valid_phys_address, GuestPtr, PerCPUPageMappingGuard},
+    mm::{
+        guestmem::{copy_slice_to_guest, read_bytes_from_guest},
+        valid_phys_address, GuestPtr, PerCPUPageMappingGuard,
+    },
     protocols::{errors::SvsmReqError, RequestParams},
     types::PAGE_SIZE,
     vtpm::{vtpm_get_locked, TcgTpmSimulatorInterface, VtpmProtocolInterface},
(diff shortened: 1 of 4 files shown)


stefano-garzarella:
This is a followup of #574 with a reference to the discussion we had here: https://github.com/coconut-svsm/svsm/pull/574#discussion_r1900871887 (thanks @00xc and @cclaudio)

I'm reporting the last comment from Carlos, to restart the discussion:

> [00xc](https://github.com/00xc) [Jan 16, 2025](https://github.com/coconut-svsm/svsm/pull/574#discussion_r1919034235) •
> 
> I'd say so, yes (although someone can correct me on the need for a TLB flush on both operations). We would also need to revert the RMP change on the error path, and that in turn might require holding PVALIDATE_LOCK (although I'm foggy on the attack vector that lock prevented, @joergroedel remembers perhaps). Anyhow, this can be handled in the review for the separate PR.

I handled the error path, moving the error handling after the section, not sure what to do with PVALIDATE_LOCK.
@joergroedel any suggestion here?

msft-jlange:
I don't think this is the sort of pattern we want to develop for interaction with guest memory.  This change will revoke all guest access to the page that is being used by the TPM, and depending on how the guest is built (where it may be reading data, polling for memory changes, etc.), the guest may encounter an unexpected and fatal memory exception as a result of the permission change.

This pattern of modifying guest memory from the SVSM will be increasingly common as the SVSM begins to implement paravisor functionality.  A much better pattern will be for the SVSM to be written defensively, using atomics or other `Sync`-safe accesses that assume the possibility for concurrent access - this is preferable to just assuming mutable access to the data and hoping for the best.

stefano-garzarella:
> I don't think this is the sort of pattern we want to develop for interaction with guest memory. This change will revoke all guest access to the page that is being used by the TPM, and depending on how the guest is built (where it may be reading data, polling for memory changes, etc.), the guest may encounter an unexpected and fatal memory exception as a result of the permission change.

@msft-jlange yeah, I see your point.

> 
> This pattern of modifying guest memory from the SVSM will be increasingly common as the SVSM begins to implement paravisor functionality. A much better pattern will be for the SVSM to be written defensively, using atomics or other `Sync`-safe accesses that assume the possibility for concurrent access - this is preferable to just assuming mutable access to the data and hoping for the best. If that's difficult, then consider copying the guest memory into a local SVSM buffer that can be accessed mutably for the operation, where the contents can be copied back to guest memory after the operation completes.

Maybe we should use something like https://github.com/rust-vmm/vm-memory/blob/main/src/volatile_memory.rs and when we need to share a buffer with C code, we will need to allocate a tmp one.

stefano-garzarella:
> If that's difficult, then consider copying the guest memory into a local SVSM buffer that can be accessed mutably for the operation, where the contents can be copied back to guest memory after the operation completes.

@msft-jlange I tried that in this version, but now I have a problem with `GuestPtr::read()`.
I just pushed the code, so you can take a look, but I have a double fault the first time OVMF sends a vTPM command:
```
[SVSM] vTPM buffer - start 2139054080 offset 2752 end 2139062272
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 7b5bcddf
[SVSM] ERROR: Info: panicked at kernel/src/cpu/idt/svsm.rs:149:9:
Double-Fault at RIP 0xffffff8000208658 RSP: 0xfffffe0280003eb0 CR2: 0xfffffe0280003ea8
[SVSM] ---BACKTRACE---:
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 7b5bcddf
[SVSM] ERROR: Info: panicked at kernel/src/address.rs:443:21:
attempt to subtract with overflow
[SVSM] ---BACKTRACE---:
[SVSM]   [0xffffff8000234ba0] #
[SVSM]   [0xffffff800015e13e] #
[SVSM]   [0xffffff80001c54e6] #
[SVSM]   [0xffffff80001c5b2d] #
[SVSM]   [0xffffff80001c5e69] #
[SVSM]   [0xffffff80001c2494] #
[SVSM]   [0xffffff80001c36ab] #
[SVSM]   [0xffffff80001c6111] #
[SVSM]   [0xffffff8000129bd6] #
[SVSM]   [0xffffff80002317ab] #
```

Note: in the first line, I printed the paddr and the offset, so the buffer is crossing 2 pages. I thought it was the problem, but removing the vTPM driver in OVMF, and booting Linux, where the driver allocates the buffer page aligned, I have the same behavior, of course when Linux tries to access the vTPM :
```
[SVSM] vTPM buffer - start 4325560320 offset 0 end 4325564416
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 7b5bcddf
[SVSM] ERROR: Info: panicked at kernel/src/cpu/idt/svsm.rs:149:9:
Double-Fault at RIP 0xffffff8000208659 RSP: 0xfffffe0280003340 CR2: 0xfffffe0280003338
[SVSM] ---BACKTRACE---:
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 7b5bcddf
[SVSM] ERROR: Info: panicked at kernel/src/address.rs:443:21:
attempt to subtract with overflow
[SVSM] ---BACKTRACE---:
[SVSM]   [0xffffff8000234bc0] #
[SVSM]   [0xffffff800015e13e] #
[SVSM]   [0xffffff80001c54e6] #
[SVSM]   [0xffffff80001c5b2d] #
[SVSM]   [0xffffff80001c5e69] #
[SVSM]   [0xffffff80001c2494] #
[SVSM]   [0xffffff80001c36ab] #
[SVSM]   [0xffffff80001c6111] #
[SVSM]   [0xffffff8000129bd6] #
[SVSM]   [0xffffff80002317cb] #
```

So I tried to reduce the size of the buffer, for example if I do something like this I can't see the double fault if the buffer is page aligned or not (from PAGE_SIZE - 900 it double faults) :
```
            let buf_ptr = GuestPtr::<[u8; PAGE_SIZE - 1000]>::new(vaddr);
```

Any suggestions on where I'm doing it wrong?

stefano-garzarella:
> Any suggestions on where I'm doing it wrong?

I can answer my-self.... I did a stack overflow! Increasing the stack size, it works well.

So, should I avoid that, right?
Any suggestions on alternatives?

stefano-garzarella:
> > Any suggestions on where I'm doing it wrong?
> 
> I can answer my-self.... I did a stack overflow! Increasing the stack size, it works well.
> 
> So, should I avoid that, right? Any suggestions on alternatives?

Maybe we can add `GuestPtr::read_ref()` and I can do something like this:
```
let mut buffer: Vec<u8> = vec![0; PAGE_SIZE];
let buf_slice: &mut [u8; PAGE_SIZE] = buffer.as_mut_slice().try_into().expect("Failed to initialize vTPM buffer");

let buf_ptr = GuestPtr::<[u8; PAGE_SIZE]>::new(vaddr);
unsafe { buf_ptr.read_ref(buf_slice)? };

let response_size = tpm_send_command_request(buf_slice)?;

unsafe { buf_ptr.write_ref(buf_slice)? };
```

stefano-garzarella:
v2:
- added `GuestPtr::read_ref`
- copied vTPM buffer from the guest to an internal one and copied it back to return results

@msft-jlange @cclaudio @00xc please take a look.

kernel/src/protocols/vtpm.rs
```diff
@@ -239,10 +239,32 @@ fn vtpm_command_request(params: &RequestParams) -> Result<(), SvsmReqError> {
         return Err(SvsmReqError::unsupported_call());
     }
 
-    let buffer = unsafe { from_raw_parts_mut(vaddr.as_mut_ptr::<u8>(), PAGE_SIZE) };
-
     let response_size = match cmd {
-        TpmPlatformCommand::SendCommand => tpm_send_command_request(buffer)?,
+        TpmPlatformCommand::SendCommand => {
+            // Let's use Vec to avoid consuming too much stack.
+            let mut buffer: Vec<u8> = vec![0; PAGE_SIZE];
```
msft-jlange, line 245: It would be nice to avoid having to allocate an entire page for every request.  Is there some header in the payload that indicates how big the request is?  If so, you could read that header into a stack local (which should be small), allocate your buffer based on how big the header indicates the data to be, then copy the stack local into the buffer and use `read_ref` to read the remainder of the command block into the buffer.

msft-jlange, line 245: You should consider using `let mut uninit_buffer: Box<MaybeUninit<[u8]>> = Box::new_uninit_slice(PAGE_SIZE)` and then you can avoid having to zero-initialize the contents that you are about to write over (and you can also avoid the intermediate `Vec` representation).  This would require `GuestPtr::read_ref()` to become `GuestPtr::read_ptr()` (because creating a reference to uninitialized data is UB but creating a pointer to it is not), and after the copy is done, you could use `let mut buffer: &[u8] = unsafe { uninit_buffer.assume_init() }` to make your buffer safely usable once the copy initialization is complete.

msft-jlange:
> @msft-jlange @cclaudio @00xc please take a look.

Much better now!  I left a couple of comments that could tighten up the use of the intermediate buffer, but from a safety perspective, what you have now is perfectly acceptable.

kernel/src/protocols/vtpm.rs
stefano-garzarella: Looking in the SVSM specification, it actually does not say anywhere that the buffer must be PAGE_SIZE, so I think you are right. Although the Linux driver for now always passes a large PAGE_SIZE buffer since it's used also for the response, so in practice it should be the same, but I see your point, I'll try it!

stefano-garzarella: Cool, thanks for this, I like it, I'll try!

That said, I just discovered that during the path from this code to the TCG TPM C code, we allocate/copy the buffer several times:
- in `TpmSendCommandRequest::send()` https://github.com/coconut-svsm/svsm/blob/035bb52967ad938bee3820283b4b6ae4a8ce6a33/kernel/src/protocols/vtpm.rs#L116
- in `TcgTpm::send_tpm_command()` https://github.com/coconut-svsm/svsm/blob/035bb52967ad938bee3820283b4b6ae4a8ce6a33/kernel/src/vtpm/tcgtpm/mod.rs#L97

I think the second case maybe we cannot avoid, since the C code seems to expect 2 different buffers for input and output, but maybe, since we have to copy them from/to the guest anyway, we could pass 2 buffers from the beginning and avoid these allocations/copies.

I'll try to see if we can refactor a bit.

stefano-garzarella:
@deeglaze since I rebased this on main, I touched a bit also the code added with #662 , so if you can take a look/test, it will be great.

kernel/src/vtpm/tcgtpm/tss.rs
```diff
@@ -132,10 +137,10 @@ pub fn create_ek<T: TcgTpmSimulatorInterface>(
...
     let mut cmd = create_mtauth_ek_cmd(tpmt_public);
 
-    checked_send(vtpm, &mut cmd, /*set_len=*/ false)?;
+    let mut response = checked_send(vtpm, &mut cmd, /*set_len=*/ false)?;
 
     // Get size (UINT16) of TPMT_PUBLIC at offset 18.
     // Note this is output from the TPM, so its value is trusted.
-    let size_of_tpmt_public = u16::from_be_bytes([cmd[18], cmd[19]]);
-    Ok(cmd[20..(20 + size_of_tpmt_public) as usize].to_vec())
+    let size_of_tpmt_public: usize = u16::from_be_bytes([response[18], response[19]]).into();
```
deeglaze, line 144:
```suggestion
    let size_of_tpmt_public = u16::from_be_bytes([response[18], response[19]]) as usize;
```
Just for consistency of how we do this elsewhere.

kernel/src/protocols/vtpm.rs
```diff
@@ -239,10 +232,13 @@ fn vtpm_command_request(params: &RequestParams) -> Result<(), SvsmReqError> {
         return Err(SvsmReqError::unsupported_call());
     }
 
-    let buffer = unsafe { from_raw_parts_mut(vaddr.as_mut_ptr::<u8>(), PAGE_SIZE) };
-
     let response_size = match cmd {
```
deeglaze, line 235: Do you not want to also use guestmem::read_from_guest for more simply reading the TpmPlatformCommand value without needing the direct use of PerCPUPageMappingGuard?

stefano-garzarella: Yep, I'd like to simplify this a bit, plus other changes, but I'm not sure to do that in this PR or in another.

00xc: I'd say since we're changing this code, let's try to do the correct thing and use the right interface

stefano-garzarella: I did it in this PR, thanks for encouraging me :-)

deeglaze (approved):

PR #605, opened 2025-01-30
github/workflows: install bindgen-cli in publish-docs.yml

stefano-garzarella:
`bindgen` is used in our build process to build libtcgtpm. This workflow uses `make docsite` which will build also libtcgtpm, but it is currently failing with:

    make[1]: bindgen: No such file or directory

Let's add also a comment in other .yml to remember to install it in future pipelines.

In the future, we may integrate `bindgen` in the build.rs instead of invoking it from the Makefile.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -27,6 +27,13 @@ jobs:
             profile: minimal
             override: true
 
+      # bindgen-cli is required to generate bindings for TPM 2.0 Reference Implementation
+      - name: Install bindgen-cli
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: bindgen-cli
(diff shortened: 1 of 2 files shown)


stefano-garzarella:
This PR can be discarded if we merge: https://github.com/coconut-svsm/svsm/pull/606

cclaudio:
> This PR can be discarded if we merge: https://github.com/coconut-svsm/svsm/pull/606

#606 seems a better approach, unless it is not ready and we need a temporary solution

stefano-garzarella:
> > This PR can be discarded if we merge: #606
> 
> #606 seems a better approach, unless it is not ready and we need a temporary solution

Thanks for checking! I'm closing it as we agreed in the SVSM call today.

PR #606, opened 2025-01-30
libtcgtpm: use bindgen in build.rs instead of bindgen-cli in Makefile

stefano-garzarella:
We had several issues in our CI because we forgot to install `bindgen-cli` used in the libtcgtpm/Makefile to generate the TPM bindings.

Instead of using `bindgen-cli` (and then having as a requirement to have it installed), let's use `bindgen` in build.rs, this way the dependency is handled directly by cargo.

Let's also clean up the references to `libtcgtpm/src/bindings.rs` since bindings were already generated in the OUT_DIR.

Remove the reference to install `bindgen-cli` in the CI workflows, Documentation, and container image.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,12 +28,6 @@ jobs:
             override: true
             components: rustfmt, rust-src, clippy
 
-      - name: Install bindgen-cli
-        uses: actions-rs/cargo@v1
-        with:
-          command: install
-          args: bindgen-cli
-
       - name: Install TPM 2.0 Reference Implementation build dependencies
         run: sudo apt install -y autoconf autoconf-archive pkg-config build-essential automake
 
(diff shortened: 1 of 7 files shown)


stefano-garzarella:
`Rust / Check unsafe blocks` is failing because the `base` branch still requires `bindgen-cli`.
To make it pass, I can undo the change where I removed its installation for that workflow, but not sure if it makes sense or not, since after merging this PR, installing `bindgen-cli` is useless. @joergroedel WDYT?

cclaudio (approved):
LGTM

cclaudio:
I was not able to reproduce the `Rust / Check unsafe blocks` issue. I fetched the PR and run

```bash
cargo clippy --workspace --all-features --exclude packit --exclude stage1 --exclude svsm-fuzz --exclude igvmbuilder --exclude igvmmeasure --quiet -- -W clippy::undocumented_unsafe_blocks
```
Although that prints many unsafe warnings, the command exited with success (exit 0).

stefano-garzarella:
> I was not able to reproduce the `Rust / Check unsafe blocks` issue. I fetched the PR and run

Yeah, it's expected. We introduced a new CI workflow to avoid new unsafe blocks not documented. 
That workflow checks out the `base` branch of the PR (e.g. `main`) and run that command to counts the number of unsafe blocks not documented, in order to compare them with the PR applied.

In this case I removed the `bindgen-cli` installation in the workflow, so that step fails.
I can put it back in that workflow, and maybe we can remove later, but when this PR will be merged, that should not be a problem anymore.

stefano-garzarella:
v2:
- re-added the step to install `bindgen-cli` in the CI worflow that checks unsafe rust blocks, since there we checkout the `base` branch that still needs it. I'll open another PR to complete the cleanup, so we should have green CI with this PR [@joergroedel]

Issue #608, opened 2025-02-07
IgvmParamBlock.is_qemu field is ambiguous


jxu023:
In the fashion of [platform: suppress SVSM use of interrupts on KVM/QEMU](https://github.com/coconut-svsm/svsm/commit/a16ac6818cf5037c8b36396c9edb7cf1bb3bb1ec#diff-f7396369cdc48a2a9aa52e9ed0151d747472f3ebc8443665980091148c0ff837) we'd also like to disable SVSM interrupts on vanadium.

However, the is_qemu field used as a hypervisor check is also used to signify the presence of qemu devices like FwCfg and TestDev. These may also available on Vanadium.

Thus, to disambiguate is_qemu and provide a more general hypervisor check, let's add a hypervisor field to IgvmParamBlock, and perhaps rename is_qemu to something else. Then we can configure the use of TestDev with SVSM on Vanadium and run SVSM with suppressed interrupts on Vanadium.

PR #609, opened 2025-02-07
Two enums

jxu023:
This PR intends to suppress interrupts on Vanadium, and also enable the use of TestDev on Vanadium. To do this, it also disambiguates the IgvmParamBlock.is_qemu field as described in https://github.com/coconut-svsm/svsm/issues/608

### DIFF
--- a/kernel/src/config.rs
+++ b/kernel/src/config.rs
@@ -186,4 +186,11 @@ impl SvsmConfig<'_> {
             SvsmConfig::IgvmConfig(igvm_params) => igvm_params.is_qemu(),
         }
     }
+
+    pub fn hypervisor(&self) -> bootlib::igvm_params::Hypervisor {
+        match self {
+            SvsmConfig::FirmwareConfig(_) => bootlib::igvm_params::Hypervisor::Qemu,
+            SvsmConfig::IgvmConfig(igvm_params) => igvm_params.hypervisor(),
+        }
+    }
 }
--- a/kernel/src/config.rs
+++ b/kernel/src/config.rs
@@ -180,10 +180,10 @@ impl SvsmConfig<'_> {
         }
     }
 
-    pub fn is_qemu(&self) -> bool {
+    pub fn has_qemu_fw_services(&self) -> bool {
         match self {
             SvsmConfig::FirmwareConfig(_) => true,
-            SvsmConfig::IgvmConfig(igvm_params) => igvm_params.is_qemu(),
+            SvsmConfig::IgvmConfig(igvm_params) => igvm_params.has_qemu_fw_services(),
         }
     }
 
--- a/kernel/src/igvm_params.rs
+++ b/kernel/src/igvm_params.rs
@@ -417,8 +417,8 @@ impl IgvmParams<'_> {
         self.igvm_param_block.use_alternate_injection != 0
     }
 
-    pub fn is_qemu(&self) -> bool {
-        self.igvm_param_block.is_qemu != 0
+    pub fn has_qemu_fw_services(&self) -> bool {
+        self.igvm_param_block.has_qemu_fw_services != 0
     }
 
     pub fn hypervisor(&self) -> bootlib::igvm_params::Hypervisor {
--- a/kernel/src/igvm_params.rs
+++ b/kernel/src/igvm_params.rs
@@ -420,4 +420,8 @@ impl IgvmParams<'_> {
     pub fn is_qemu(&self) -> bool {
         self.igvm_param_block.is_qemu != 0
     }
+
+    pub fn hypervisor(&self) -> bootlib::igvm_params::Hypervisor {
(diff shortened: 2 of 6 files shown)


AdamCDunlap:
has_qemu_fw_services is still too generic for us. The "qemu fw sevices" come from both the pc-testdev device (installed in `scripts/launch-guest.sh`) and some code in the `scripts/test-in-svsm.sh` file. Vanadium supports a device similar to pc-testdev, but not the IORequests from the script. I think it'd make sense to split up the "qemu fw services" into 2 different flags.

I'm also a bit confused about having both the hypervisor enum and a boolean (qemu fw services) that solely depends on the enum. IMO it'd be better to have a bunch of booleans in the IgvmParamBlock that specify behavior rather than the hypervisor enum. So in that case we'd have "has_qemu_testdev", "has_test_iorequests", and "suppress_svsm_interrupts". Igvmbuilder would specify which of these to set based on the hypervisor.

jxu023:
For suppress_svsm_interrupts, it's currently conditioned on both SNP and the hypervisor. Since IGVM is arch-agnostic for snp/tdx, it seems the determination of this may be necessary to do at runtime instead of igvm build time.

> I'm also a bit confused about having both the hypervisor enum and a boolean (qemu fw services) that solely depends on the enum.

Perhaps it's better to remove the qemu_fw_services/is_qemu field entirely from IgvmParamsBlock then and just make do with the hypervisor field? Then we can make those checks based on hypervisor.

AdamCDunlap:
> For suppress_svsm_interrupts, it's currently conditioned on both SNP and the hypervisor. Since IGVM is arch-agnostic for snp/tdx, it seems the determination of this may be necessary to do at runtime instead of igvm build time.

Name it `supress_svsm_interrupts_on_snp`?

> Then we can make those checks based on hypervisor.

That would be better than having both, but IMO individual bools are more "morally correct" than a hypervisor enum. The idea should be that the IGVM could boot on any hypervisor even if practically there are different requirements from each hypervisor.

jxu023:
> The idea should be that the IGVM could boot on any hypervisor even if practically there are different requirements from each hypervisor.

Ah I see, that makes sense; I was initially confused since we're building different IGVM files for each hypervisor; but this makes sense if we're moving towards that direction.

I'll write up a separate PR and then close this one.

kernel/src/config.rs
```diff
@@ -180,10 +180,17 @@ impl SvsmConfig<'_> {
...
-    pub fn is_qemu(&self) -> bool {
+    pub fn has_qemu_fw_services(&self) -> bool {
         match self {
             SvsmConfig::FirmwareConfig(_) => true,
-            SvsmConfig::IgvmConfig(igvm_params) => igvm_params.is_qemu(),
+            SvsmConfig::IgvmConfig(igvm_params) => igvm_params.has_qemu_fw_services(),
+        }
+    }
+
+    pub fn hypervisor(&self) -> bootlib::igvm_params::Hypervisor {
```
msft-jlange, line 190: Since the only definition of `Hypervisor` that is usable by the kernel environment is the one in `bootlib::igvm_params`, can we just `use bootlib::igvm_params::Hypervisor` and avoid the need for explicit namespace decoration here and in other uses?

kernel/src/igvm_params.rs
```diff
@@ -417,7 +417,11 @@ impl IgvmParams<'_> {
         self.igvm_param_block.use_alternate_injection != 0
     }
 
-    pub fn is_qemu(&self) -> bool {
-        self.igvm_param_block.is_qemu != 0
+    pub fn has_qemu_fw_services(&self) -> bool {
+        self.igvm_param_block.has_qemu_fw_services != 0
+    }
+
+    pub fn hypervisor(&self) -> bootlib::igvm_params::Hypervisor {
```
msft-jlange, line 424: Above comment about `use bootlib::igvm_params::Hypervisor` applies here as well.

kernel/src/stage2.rs
```diff
@@ -428,10 +429,12 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
     )
     .expect("Failed to map and validate heap");
 
-    // Determine whether use of interrupts n the SVSM should be suppressed.
+    // Determine whether use of interrupts in the SVSM should be suppressed.
     // This is required when running SNP under KVM/QEMU.
     let suppress_svsm_interrupts = match platform_type {
-        SvsmPlatformType::Snp => config.is_qemu(),
+        SvsmPlatformType::Snp => {
+            config.hypervisor() == Hypervisor::Vanadium || config.hypervisor() == Hypervisor::Qemu
```
msft-jlange, line 436: Can we define a method like `SvsmConfig::can_use_svsm_interrupts` so that all of the per-hypervisor logic can exist in the config object instead of having decisions leak out to other places like this one?

PR #613, opened 2025-02-12
requests: refactor SVSM request loop

msft-jlange:
Move the SVSM request loop out of the idle task and into a separate task, and invoke it only on SNP platforms.  This is a step towards moving SVSM request processing out of the kernel, and enabling it to be excluded in configurations that only support paravisor processing.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -42,8 +42,8 @@ use crate::mm::{
     SVSM_PERCPU_VMSA_BASE, SVSM_SHADOW_STACKS_INIT_TASK, SVSM_SHADOW_STACK_ISST_DF_BASE,
     SVSM_STACK_IST_DF_BASE,
 };
-use crate::platform::{SvsmPlatform, SVSM_PLATFORM};
-use crate::requests::{request_loop, request_processing_main};
+use crate::platform::{halt, SvsmPlatform, SVSM_PLATFORM};
+use crate::requests::{request_loop_main, request_processing_main};
 use crate::sev::ghcb::{GhcbPage, GHCB};
 use crate::sev::hv_doorbell::{allocate_hv_doorbell_page, HVDoorbell};
 use crate::sev::msr_protocol::{hypervisor_ghcb_features, GHCBHvFeatures};
@@ -56,6 +56,7 @@ use crate::types::{
     PAGE_SHIFT, PAGE_SHIFT_2M, PAGE_SIZE, PAGE_SIZE_2M, SVSM_TR_ATTRIBUTES, SVSM_TSS,
 };
 use crate::utils::MemoryRegion;
+use alloc::format;
 use alloc::string::String;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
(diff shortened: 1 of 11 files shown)


kernel/src/cpu/percpu.rs
```diff
@@ -1369,3 +1372,30 @@ pub fn process_requests() {
...
 }
+
+#[no_mangle]
+pub extern "C" fn cpu_idle_loop() {
+    // Start request processing on this CPU if required.
+    if SVSM_PLATFORM.start_svsm_request_loop() {
+        start_kernel_task(request_processing_main, String::from("request-processing"))
+            .expect("Failed to launch request processing task");
+        start_kernel_task(request_loop_main, String::from("request-loop"))
+            .expect("Failed to launch request loop task");
```
joergroedel, line 1383: Can we add the APIC id of the CPU to the task-names, maybe using `format!()`?

kernel/src/requests.rs
```diff
@@ -181,7 +185,7 @@ pub fn request_loop() {
             drop(guard);
         } else {
             log::debug!("No VMSA or CAA! Halting");
-            halt();
+            go_idle();
```
joergroedel, line 188: So this is a workaround until COCONUT gets a decent event system and IPIs on all platforms, right?

joergroedel (commented):
In general this LGTM. I stumbled a bit over the `go_idle()` implementation, but after some investigation it looks okay since it takes the scheduler/runqueue (which disables IRQs) lock.
Left two comments with questions for clarification.

kernel/src/cpu/percpu.rs
msft-jlange: Yes, though in general I prefer to transition us away from APIC ID and towards CPU index for internal CPU identification.  This is because CPU index is guaranteed to be dense (which is important for structures like `CpuSet` which holds a bitmap) and APIC ID is not guaranteed to be dense due to NUMA considerations.  This distinction already causes problems for TDX which requires the use of CPU index at certain places in the architecture (like MP startup), and I believe that using CPU index everywhere other than selecting APIC destinations will provide much better consistency across the code base.  This change isn't going to happen overnight, but I'd prefer to stop adding APIC ID in new places like this one and use CPU index instead.  We can discuss that later if necessary.

kernel/src/requests.rs
msft-jlange: I'm not certain about that.  My past experience with paravisor design suggests that every time the paravisor gets control, it may need to evaluate guest state to determine what work needs to be performed.  There is not a strong contract between host and VMPL 0 that guarantees that the host will send a specific interrupt notification when waking VMPL 0 from idle, so we may find that it is always necessary for the VMM loop to get control after every wake from idle so it can make decisions about how to proceed.  If we find someday that every possible wake-from-idle scenario can be covered by an interrupt handler signaling the correct event, then we can replace this with an event wait, but I'm not ready to assume yet that all such cases can be covered by explicit event notifications.

kernel/src/cpu/percpu.rs
msft-jlange: It turns out that `format!()` requires `std` so it cannot be used directly here.  My preference for now is not to add the CPU index, since the current code doesn't include the CPU index, and to save the formatting changes for a future PR.

msft-jlange: I found `format!()` in `alloc` so it should be usable.

msft-jlange: After everything I wrote above about CPU index vs. APIC ID, I'm going to stick with APIC ID for this PR because that's how the existing request tasks work.  I'll defer the conversion to CPU index to a future PR.

PR #619, opened 2025-02-14
syscall: Implement CAPABILITIES

peterfang:
The `CAPABILITIES` syscall takes an index as its input and returns the corresponding capability in a u64 as per the syscall ABI.
Bits [2:0] of GlobalFeatureFlags are defined as the platform type.


joergroedel (approved):
LGTM, thanks for implementing this.

PR #622, opened 2025-02-19
Allow early stack unwinding

peterfang:
Allow the stack unwinder to work right after preparing the console in the SVSM kernel. Use an early #PF handler and the early exception table to determine if the PerCpu page is ready. If there is no PerCpu available, perform best-effort unwinding based on the default stack regions.

### DIFF
--- a/kernel/src/cpu/extable.rs
+++ b/kernel/src/cpu/extable.rs
@@ -4,14 +4,17 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
-extern "C" {
-    pub static exception_table_start: u8;
-    pub static exception_table_end: u8;
-}
-
 use super::idt::common::X86ExceptionContext;
 use crate::address::{Address, VirtAddr};
+use crate::utils::MemoryRegion;
 use core::mem;
+
+extern "C" {
+    static exception_table_start: u8;
+    static exception_table_end: u8;
+    static early_exception_table_start: u8;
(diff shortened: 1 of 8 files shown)


kernel/src/cpu/extable.rs
```diff
@@ -67,9 +67,20 @@ pub fn dump_exception_table() {
...
+        MemoryRegion::from_addresses(
+            VirtAddr::from(&raw const early_exception_table_start),
+            VirtAddr::from(&raw const early_exception_table_end),
+        )
+    } else {
+        MemoryRegion::from_addresses(
+            VirtAddr::from(&raw const exception_table_start),
+            VirtAddr::from(&raw const exception_table_end),
+        )
+    };
```
joergroedel, line 81: Instead of the `early` flag, please split up the function into

* `handle_exception_table()`
* `handle_exception_table_early()` and
* `handle_exception_table_common()` for the common part of both code paths.

This will make the call-sites more readable.

joergroedel (changes requested):
Hi @peterfang ,

Thanks for these changes, I am fine with the probe-approach to detecting presence of `PerCpu`. I left two comments to address.

Thanks!

peterfang:
> Hi @peterfang ,
> 
> Thanks for these changes, I am fine with the probe-approach to detecting presence of `PerCpu`. I left two comments to address.
> 
> Thanks!

Thanks for reviewing. All of the comments have been addressed.

PR #625, opened 2025-02-22
ghcb: Use APIC ID List NAE event to get APIC IDs

### LINKED ISSUE #623: SVSM assumes APIC IDs are sequential and fails to boot if there are gaps
Currently, SVSM assumes APIC IDs go from 0 to cpu count: https://github.com/coconut-svsm/svsm/blob/035bb52967ad938bee3820283b4b6ae4a8ce6a33/kernel/src/igvm_params.rs#L253C1-L259C11 . However, some hypervisors do not assign APIC IDs this way and instead leave gaps for reasons beyond my understanding (something about last level caches and threads per core). When SVSM tries to boot an AP with an APIC ID that was not assigned by the hypervisor, SVSM panics.

So we need to modify SVSM to instead get the list of APIC IDs from the hypervisor. I believe there's 3 reasonable ways to do this:
1. ACPI tables. It appears this is what happens in the fw_cfg version [load_acpi_cpu_info](https://github.com/coconut-svsm/svsm/blob/035bb52967ad938bee3820283b4b6ae4a8ce6a33/kernel/src/acpi/tables.rs#L466). This probably wouldn't work with IGVM since the ACPI tables aren't supplied to SVSM
2. A new IGVM parameter. I don't see this parameter currently existing in the IGVM documentation -- it just has the number of VPs, not which APIC IDs they use
3. The SVM_VMGEXIT_GET_APIC_IDS hypercall

I'm inclined to go with the hypercall. My only reservation is that we had some internal concern about the KVM implementation of the hypercall, but since EDK2 uses it as well, it's probably OK to add it here as well.

I can draft a PR but would appreciate any feedback

AdamCDunlap:
The GHCB protocol includes an NAE to get the list of APIC IDs for the guest, so use it on SEV-SNP. Previously, the list of APIC IDs was either gotten via ACPI table or using IGVM to get the number of vCPUs and assuming the APIC IDs were sequential. This is still done if the platform is not SNP since other platforms do not provide a way to get this list.

Fixes #623


kernel/src/svsm.rs
```diff
@@ -312,16 +313,17 @@ pub extern "C" fn svsm_main() {
...
-    let cpus = config.load_cpu_info().expect("Failed to load ACPI tables");
-    let mut nr_cpus = 0;
-
-    for cpu in cpus.iter() {
-        if cpu.enabled {
-            nr_cpus += 1;
-        }
-    }
+    // Get the list of APIC IDs from the platform if possible. If the platform does not
+    // support the operation,get them from the config instead.
```
deeglaze, line 317:
```suggestion
    // support the operation, get them from the config instead.
```

msft-jlange:
This change doesn't work on TDX because it uses an SNP-specific GHCB protocol.  It is possible to solve this in a cross-platform way using the MADT, and I just posted #626 to do exactly this.  If this doesn't work correctly on QEMU, then a better approach would be to use `FwCfg` on QEMU builds to obtain the MADT and extract APIC IDs that way.  I'm prepared to update my PR if required.  I would strongly prefer to avoid any SNP-specific logic since a cross-platform technique is available.

PR #626, opened 2025-02-24
igvm: use MADT for determining APIC IDs

msft-jlange:
The IGVM format specifies a mechanism for the host to inject an MADT into the guest address space.  This MADT should be used to determine the set of APIC IDs associated with the set of CPUs that will be started by the guest.

### DIFF
--- a/igvmbuilder/src/gpa_map.rs
+++ b/igvmbuilder/src/gpa_map.rs
@@ -65,6 +65,7 @@ pub struct GpaMap {
     pub igvm_param_block: GpaRange,
     pub general_params: GpaRange,
     pub memory_map: GpaRange,
+    pub madt: GpaRange,
     pub guest_context: GpaRange,
     // The kernel region represents the maximum allowable size. The hypervisor may request that it
     // be smaller to save memory on smaller machine shapes. However, the entire region should not
@@ -139,7 +140,12 @@ impl GpaMap {
 
         let igvm_param_block = GpaRange::new_page(kernel_fs.get_end())?;
         let general_params = GpaRange::new_page(igvm_param_block.get_end())?;
-        let memory_map = GpaRange::new_page(general_params.get_end())?;
+        let madt_size = match options.hypervisor {
+            Hypervisor::HyperV | Hypervisor::Vanadium => PAGE_SIZE_4K,
+            Hypervisor::Qemu => 0,
+        };
+        let madt = GpaRange::new(general_params.get_end(), madt_size)?;
+        let memory_map = GpaRange::new_page(madt.get_end())?;
         let guest_context = if let Some(firmware) = firmware {
             if firmware.get_guest_context().is_some() {
                 // Locate the guest context after the memory map parameter page
@@ -171,6 +177,7 @@ impl GpaMap {
             igvm_param_block,
             general_params,
             memory_map,
+            madt,
             guest_context,
             kernel,
             vmsa,
--- a/igvmbuilder/src/igvm_builder.rs
+++ b/igvmbuilder/src/igvm_builder.rs
@@ -44,7 +44,8 @@ pub const ANY_NATIVE_COMPATIBILITY_MASK: u32 = NATIVE_COMPATIBILITY_MASK | VSM_C
 // Parameter area indices
 const IGVM_GENERAL_PARAMS_PA: u32 = 0;
 const IGVM_MEMORY_MAP_PA: u32 = 1;
-const IGVM_PARAMETER_COUNT: u32 = 2;
+const IGVM_MADT_PA: u32 = 2;
+const IGVM_PARAMETER_COUNT: u32 = 3;
 
 const _: () = assert!(size_of::<IgvmParamBlock>() as u64 <= PAGE_SIZE_4K);
 const _: () = assert!(size_of::<IgvmGuestContext>() as u64 <= PAGE_SIZE_4K);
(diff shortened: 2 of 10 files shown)


igvmbuilder/src/gpa_map.rs
```diff
@@ -139,7 +140,8 @@ impl GpaMap {
 
         let igvm_param_block = GpaRange::new_page(kernel_fs.get_end())?;
         let general_params = GpaRange::new_page(igvm_param_block.get_end())?;
-        let memory_map = GpaRange::new_page(general_params.get_end())?;
+        let madt = GpaRange::new_page(general_params.get_end())?;
```
deeglaze, line 143: Is this table something we can measure into a service manifest and/or rtmr? With the oem and table ids undigested for lookup purposes?

msft-jlange: I suppose so, but why?  The selection of APIC IDs is an attribute of the environment and shouldn't affect the trustworthiness of the VM (once the VM is capable of sanitizing the table, which is work that remains to be done).  We don't measure any other parts of firmware config, such as the memory map, so what benefit is gained by including a measurement of the APIC ID table?

Even if we do want to measure some aspects of the environment, I believe that is entirely outside the scope of this PR, which aims only to collect the correct association between vCPU and APIC ID according to the way the host wants to emulate cross-processor interrupt delivery.

deeglaze: Alright, well this is already available through the qemu fwcfg interface, so I’d prefer to avoid needing to tightly couple different VMM backends to populate this region. I won’t block this, but I think there are multiple strategies to get this information, and Adam’s is the most straightforward for Sev-snp.

msft-jlange: I strongly disagree that SNP vs TDX vs native is the correct axis to pivot on.  The fact that firmware config exposes this is great for QEMU but on Hyper-V, the only way to get the MADT is through the unmeasured IGVM parameter (which I think will show up in QEMU at some point anyway).  I will be updating this PR to fall back to firmware config when the IGVM MADT is not available, which will give us the comprehensive support we need with a minimum of backend fuss.

In the long run, we really should standardize on a single, cross-architecture, cross-host way to pass parameters from the host into the confidential guest.  I don't think QEMU firmware config is a candidate for that because it is too coupled to how QEMU works and can never really be workable on Hyper-V.  IGVM was intended to be exactly that standard, and the IGVM work done so far in QEMU suggests that it is likely to be a suitable choice.  Since we don't have convergence yet, the COCONUT-SVSM code can continue to support all of the formats it knows how to parse today, though it would be nice to standardize in the future.

igvmbuilder/src/igvm_builder.rs
```diff
@@ -345,12 +349,17 @@ impl IgvmBuilder {
...
         self.directives.push(IgvmDirectiveHeader::ParameterArea {
             number_of_bytes: PAGE_SIZE_4K,
             parameter_area_index: IGVM_MEMORY_MAP_PA,
             initial_data: vec![],
         });
+        self.directives.push(IgvmDirectiveHeader::ParameterArea {
+            number_of_bytes: PAGE_SIZE_4K,
+            parameter_area_index: IGVM_MADT_PA,
+            initial_data: vec![],
+        });
```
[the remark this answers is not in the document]
msft-jlange: > QEMU stumbles over these sections in the IGVM file and exits with an error:

That was a mistake in the PR.  I added all of the logic for the SVSM to obtain the MADT from firmware config instead of the IGVM file, but forgot to make the change in the IGVM builder to exclude the MADT directive in QEMU builds.  It should work correctly now.

msft-jlange:
> We need to find a way to get this resolved with QEMU, allow building IGVM files without the MADT directives for QEMU or adding support for these directives to QEMU.

I just fixed the PR to build the QEMU IGVM without the MADT directive.  That was my intention but forgot that logic in the last version of the PR.

AdamCDunlap:
I checked the latest version with Vanadium and I got a successful boot with 96 vCPUs and non-contiguous APIC IDs. Thanks!

igvmbuilder/src/igvm_builder.rs
```diff
@@ -378,6 +394,13 @@ impl IgvmBuilder {
                 parameter_area_index: IGVM_MEMORY_MAP_PA,
             },
         ));
+        self.directives.push(IgvmDirectiveHeader::ParameterInsert(
```
AdamCDunlap, line 397: Should this be under an `if self.gpa_map.madt.get_size() != 0`? If I understand correctly, this will cause the loader to insert the all-zeros parameter area (since the MADT was not actually added to it) to the madt region, but the madt region is 0 sized. It would probably work since the madt region overlaps the "general params" region and the next directive would overwrite it, but it still seems error prone.

It might make sense to have 1 if statement and then bundle together to the parameter area creation, madt insertion, and parameter area insertion so all the code for it is together.

AdamCDunlap (approved):

igvmbuilder/src/igvm_builder.rs
AdamCDunlap:
```
        let madt = GpaRange::new(general_params.get_end(), madt_size)?;
        let memory_map = GpaRange::new_page(madt.get_end())?;
...
        let madt_offset = param_page_offset + PAGE_SIZE_4K as u32;
        let memory_map_offset = madt_offset + self.gpa_map.madt.get_size() as u32;
```

Isn't this saying that (if madt size is 0), the madt parameter area and the memory map parameter areas will be inserted at the same GPA? (Above I said it overlapped with the general params, but it looks like it's actually the memory map.) I agree it's not a problem if an all-0s parameter area is inserted, but this appears they're being inserted at the same location and thus overwriting each other.

msft-jlange: Good point.  I'll change it so that all of the MADT parameter management is contingent on wanting to have the MADT, but it won't be behind a single `if`.

PR #627, opened 2025-02-25
github/workflows: print clippy output in case of errors

stefano-garzarella:
In the `unsafe-check` job, if `cargo clippy` fails, we don't have any log since we are redirecting the stderr to a file to count the number of undocumented unsafe blocks.

We don't expect it to fail, since other steps in `check` job should check this case, but it has happened in some cases.

Print the file content in case of `cargo clippy` returns with error, so we can debug better CI failures.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -149,7 +149,7 @@ jobs:
 
       - name: Clippy with undocumented_unsafe_blocks for PR branch
         run: |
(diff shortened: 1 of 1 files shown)


osteffenrh:
Could also do `cargo clippy ... > file || ( cat file; exit 1)` to keep the CI pipeline behavior the same and stop if the command fails and not have it execute the remaining commands. But it is also fine as it is, I guess.

PR #629, opened 2025-02-26
Add Virtualization Exception handler

peterfang:
Intel TD can receive #VE in certain cases and it should be handled by SVSM. Introduce a virtualization exception handler to handle them. Only the CPUID exit reason is handled right now.


joergroedel (approved):
LGTM, thanks @peterfang. I will also request a review from @msft-jlange and merge upon his approval.

kernel/src/tdx/tdcall.rs
```diff
@@ -16,11 +17,24 @@ use bitfield_struct::bitfield;
...
+const TDG_VP_VEINFO_GET: u32 = 3;
 const TDG_MEM_PAGE_ACCEPT: u32 = 6;
 
+const TDVMCALL_CPUID: u32 = 10;
 const TDVMCALL_HLT: u32 = 12;
 const TDVMCALL_IO: u32 = 30;
 
+/// Virtualization exception information
+#[derive(Clone, Copy, Debug)]
+pub struct TdVeInfo {
```
msft-jlange, line 29: This needs to be `#[repr(C)]` because it must match the ABI structure definition.

msft-jlange: nvm

kernel/src/cpu/ve.rs
```diff
@@ -0,0 +1,38 @@
...
+use crate::tdx::tdcall::{tdcall_get_ve_info, tdvmcall_cpuid};
+use crate::tdx::TdxError;
+
+const VMX_EXIT_REASON_CPUID: u32 = 10;
+
+pub fn handle_virtualization_exception(ctx: &mut X86ExceptionContext) -> Result<(), SvsmError> {
+    let veinfo = tdcall_get_ve_info().expect("Failed to get #VE info");
+
+    match veinfo.exit_reason {
+        VMX_EXIT_REASON_CPUID => handle_cpuid(ctx),
```
msft-jlange, line 18: I'd like to consider adding IO emulation in the near future because that should be low-hanging fruit.  There is much that can be copied from the instruction emulator and/or the #VC handler.

kernel/src/cpu/ve.rs
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
msft-jlange, line 1: Can we put this file under `kernel\src\tdx` instead?  I'd like to move towards a model where architecture-specific code is in architecture-specific directories.  There's a lot of SNP code that doesn't follow this pattern today but I'd like to avoid making it any worse than it already is.

kernel/src/cpu/ve.rs
peterfang: I can add IO emulation after this PR. While we're on this subject, how do you feel about possibly replacing our current instruction decoder with a public crate such as [zydis](https://crates.io/crates/zydis)? I just checked and it supports `no_std` + decoder-only compilation.

The reason for this is because x86 instruction decoding is notoriously difficult to get right and bugs in this area tend to be quite challenging to triage. The decoder we previously upstreamed was an in-house development and not fully stress-tested or fuzzed. We could also continue to improve our coconut fuzzer but I'm open to exploring whether we could just delegate that to a well-known decoder. This of course would take time and would not be done right away.

kernel/src/cpu/ve.rs
peterfang: I agree. I'll do that.

kernel/src/cpu/ve.rs
msft-jlange: > While we're on this subject, how do you feel about possibly replacing our current instruction decoder with a public crate such as [zydis](https://crates.io/crates/zydis)? I just checked and it supports `no_std` + decoder-only compilation.

The instruction emulator represents a significant portion of the attack surface because bugs in the instruction emulator permit tricking the emulator into executing the wrong code, which is tantamount to a remote code execution vulnerability.  For this reason, I think it's critical for the instruction emulator to be as small as possible, tailored only to the bare minimum of instruction forms that we think are ever appropriate to emulate.  For provenance guarantees, it's even better if that is code in our tree so that we are not required to audit every change that is ingested every time we update an imported crate.  So I am generally against replacing our emulator and would prefer to see us narrowly tailor it to our specific needs.

I also believe we should be designing COCONUT-SVSM to rely on virtualization exceptions as little as possible - it is a custom-built system with a robust platform abstraction, so falling back to architectural primitives should seldom be necessary.  In fact, I'm now questioning whether IO handling is ever going to be necessary, so I would advocate avoiding adding it all until we can identify some code that requires use of IO emulation through virtualization exceptions.

peterfang: Makes sense. I'm pretty sure that CPUID VEs will be needed at some point but we can add the rest as needed.

msft-jlange: > Makes sense. I'm pretty sure that CPUID VEs will be needed at some point but we can add the rest as needed.

I'm inclined to agree but the behavior of intercept management on SNP is much messier so it might be the case that we have to end up removing all CPUID uses, even in user mode.  For now at least, having CPUID VE is a good thing.

ereshetova:
I am obviously late to this review, and I dont have any objections to this PR, but want to bring attention to one point for the future. In general blindly forwarding the cpuid (and MSR) requests to the vmm via tdvmcall can be a security risk. So at some point of time when we will reach the point that coconut-svsm will be handling cpuid requests from L2 VMs there should be some filtering in place to prevent blind forwarding. Otherwise we have to rely on L2 OS to not making cpuid requests that VMM can manipulate, but i think this is impractical generic assumption (albeit Linux guest does exactly this, it only issues tdvmcalls for kvm sw range of CPUID leaves, but not for MSRs). Also, we have a new feature that will be part of future TDX module, which is called "#VE reduction" (see VE reduction feature changes in the latest spec: https://cdrdv2.intel.com/v1/dl/getContent/733575 ), which does reduce greatly the number of CPUID leaves and MSRs that would cause #VE to be injected into the guest at the first place.

peterfang:
> I am obviously late to this review, and I dont have any objections to this PR, but want to bring attention to one point for the future. In general blindly forwarding the cpuid (and MSR) requests to the vmm via tdvmcall can be a security risk. So at some point of time when we will reach the point that coconut-svsm will be handling cpuid requests from L2 VMs there should be some filtering in place to prevent blind forwarding. Otherwise we have to rely on L2 OS to not making cpuid requests that VMM can manipulate, but i think this is impractical generic assumption (albeit Linux guest does exactly this, it only issues tdvmcalls for kvm sw range of CPUID leaves, but not for MSRs). Also, we have a new feature that will be part of future TDX module, which is called "#VE reduction" (see VE reduction feature changes in the latest spec: https://cdrdv2.intel.com/v1/dl/getContent/733575 ), which does reduce greatly the number of CPUID leaves and MSRs that would cause #VE to be injected into the guest at the first place.

Noted. Thanks @ereshetova 
In principal, L2 CPUIDs/MSRs will be handled with greater care in L1 VMM.

PR #630, opened 2025-02-26
cpu: compile for shadow stack support by default

msft-jlange:
Shadow stack support should be compiled into the code for all builds by default so that shadow stacks are enabled on all platforms that support CET.


p4zuu:
SVSM kernel still boots fine with these changes on my CET non-coco machine, but I can't test on other platforms.
However, do you know if shadow-stack is widely supported on modern non-coco CPUs? If we plan to fully support non-coco platforms and that shadow-stack CPUs aren't very common, maybe we should annoucne in the doc that the default build requires a CPU supporting shadow-stacks. But overall, enabling shadow-stacks by default seems a good idea to me

joergroedel (approved):
LGTM.

osteffenrh:
Somehow SVSM does not boot anymore for me with this.
It just hangs when starting the kernel.
Do I need to change anything else?

```
$ scripts/launch_guest.sh --qemu ../build/bin/qemu-system-x86_64 --igvm bin/coconut-qemu.igvm
=============================
Launching SVSM guest
=============================
QEMU:         ../build/bin/qemu-system-x86_64
QEMU Version: 9.1.50
IGVM:         bin/coconut-qemu.igvm
IMAGE:
=============================
Press Ctrl-] to interrupt
=============================
[Stage2] COCONUT Secure Virtual Machine Service Module Version f6dd6702
[Stage2] SVSM memory region: [0x0000008000000000-0x0000008001000000]
[Stage2] Order-02: total pages:     0 free pages:     0
[Stage2] Order-03: total pages:     0 free pages:     0
[Stage2] Order-04: total pages:     2 free pages:     2
[Stage2] Order-05: total pages:     3 free pages:     3
[Stage2] Total memory: 572KiB free memory: 516KiB
[Stage2]   kernel_virtual_base      = 0xffffff8000000000
```

joergroedel:
I merged the fix in our Linux tree, everyone needs to update their host kernel in order to use shadow stacks.

stefano-garzarella:
> I merged the fix in our Linux tree, everyone needs to update their host kernel in order to use shadow stacks.

@joergroedel should we send an email to the ML ?

joergroedel:
> @joergroedel should we send an email to the ML ?

I am about to write one :)

osteffenrh:
Thanks for fixing, @joergroedel. Works for me :-)

p4zuu:
@joergroedel why is the COCONUT kernel panicking when trying to access the shadow stack MSRs? MSRs read/write are supposed to be handled if the #VC was raised from kernel space

tlendacky:
> @joergroedel why is the COCONUT kernel panicking when trying to access the shadow stack MSRs? MSRs read/write are supposed to be handled if the #VC was raised from kernel space

The kernel shadow stack MSRs are swap type A, which means that they are restored from the VMSA on VMRUN. Given that, you can never really set the MSR in the guest if access to it is being intercepted. Since they can't be set, trying then to use a shadow stack instruction will #UD, which is likely the reason for the crash.

tlendacky:
> I merged the fix in our Linux tree, everyone needs to update their host kernel in order to use shadow stacks.

Shouldn't the shadow stack support only be enabled if CPUID indicates it is present/available?

PR #631, opened 2025-02-27
Update Alternate Injection GHCB HV feature bit and NAE events SW_EXITCODE

MelodyHuibo:
Since the GHCB HV feature bit 7 and 8 are taken, update the Alternate Injection bit to bit 9.

Same as the NAE events related to Alternate Injection, update their SW_EXITCODE.


msft-jlange:
> Since the GHCB HV feature bit 7 and 8 are taken, update the Alternate Injection bit to bit 9.
> Same as the NAE events related to Alternate Injection, update their SW_EXITCODE.

Is there a newer version of the GHCB spec that specifies these bit positions and NAE codes?

MelodyHuibo:
Yes, I just updated the GHCB with Alternate Injection, it will be sent out
later for review.

Thanks,
Melody

PR #633, opened 2025-02-27
Typo fix: contiguous

deeglaze:


stefano-garzarella:
@deeglaze can you add S-o-b to make CI happy?

nit: maybe also a prefix, something like `kernel/mm: ...`

PR #634, opened 2025-02-28
qemu/native: place svsm kernel below 512G

kraxel:
Makes igvm work in native platform mode on somewhat older intel hardware with phys-bits = 39 (which can't address memory above 512G).

### DIFF
--- a/igvmbuilder/src/gpa_map.rs
+++ b/igvmbuilder/src/gpa_map.rs
@@ -125,8 +125,8 @@ impl GpaMap {
         // Choose the kernel base and maximum size.
         let kernel = match options.hypervisor {
             Hypervisor::Qemu => {
-                // Place the kernel area at 512 GB with a maximum size of 16 MB.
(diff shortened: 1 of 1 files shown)


joergroedel:
Which Intel hardware is affected by this limitation? I am also trying to understand the use-case of running the SVSM on such older systems.

kraxel:
> Which Intel hardware is affected by this limitation? I am also trying to understand the use-case of running the SVSM on such older systems.

My workstation, skylake I think.  Use case is development + testing.  At least some areas can be worked on without requiring a machine with SNP support.  But I guess I can live with using TCG, svsm is small enough that it boots fast even on emulation.

joergroedel:
> My workstation, skylake I think. Use case is development + testing. At least some areas can be worked on without requiring a machine with SNP support. But I guess I can live with using TCG, svsm is small enough that it boots fast even on emulation.

Okay, I am a bit cautious with this change as I am not sure if it could break things for others. Closing as per your comment.

PR #635, opened 2025-03-01
virtio-blk storage via mmio using virtio-drivers [v3]

osteffenrh:
This PR adds support for virtio-blk storage devices using the virtio-mmio transport.

A minimized version of the [rcore-os/virtio-drivers crate]() (MIT License) is added 
to the codebase as a sub-crate. Most of the unused code was removed (PCI support, unused device types,
example programs,...), resulting in about 3800 lines of code, see https://github.com/osteffenrh/virtio-drivers/pull/3

This PR requires a patched version of Qemu (coconut-svsm/qemu#13), which features virtio-mmio slots in the Q35 machine model (thanks @kraxel).

Qemu communicates the MMIO base address via the fw_cfg interface, making use of OVMF's
[HardwareInfoLib](https://edk2.groups.io/g/devel/message/86051). Work is already ongoing to [add this to upstream Qemu](https://lists.nongnu.org/archive/html/qemu-devel/2025-02/msg03923.html).
Device information are stored in the `etc/hardware-info` fw_cfg file in a simple binary format.
Building on top of that, the PR introduces a new device type to store MMIO addresses for use by the SVSM.

How to test:
 - Use Qemu from (coconut-svsm/qemu#13)
 - run the in-svsm tests make test-in-svsm

Fixes since V2 (#542):
- pull virtio code into the repo + minimize it
- device discovery via fw_cfg

Things for future PRs:
- encryption layer
- file system / partitioning


osteffenrh:
- Added read + write tests.
- Requests that are larger than PAGE_SIZE are automatically split up to keep bounce buffer size small and avoid needing multiple pages (could be improved in the future)

p4zuu:
Would MMIO #VC handling help here? I can start working on it if that's easier for this PR. We could also implement it after this PR is merged if this one is faster

osteffenrh:
> > Yes, that would allow us to get rid of some code and would be the better solution. I pretty much all modifications I had to do to the virtio-drivers create code could go then.
> 
> Alright, I will start working on it then, unless you prefer to do so or already started working on it. That might take a couple of weeks since I'm working on SVSM on my free time now

Awesome, thanks!

kernel/src/block/error.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2025 Red Hat, Inc.
+//
+// Author: Oliver Steffen <osteffen@redhat.com>
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum BlockDeviceError {
+    Failed, // ToDo: insert proper errors
```
stefano-garzarella, line 9: and also a documentation, I can help with that

kernel/src/block/virtio_blk_demo.rs
```diff
@@ -0,0 +1,33 @@
+use crate::{
```
stefano-garzarella, line 1: I'd add this demo in a separate commit, so we can easily remove that commit from the PR

kernel/Cargo.toml
```diff
@@ -37,6 +37,7 @@ libtcgtpm = { workspace = true, optional = true }
 zerocopy = { workspace = true, features = ["alloc", "derive"] }
 release.workspace = true
 virtio-drivers.workspace = true
+sha2 = { workspace = true, features = ["force-soft"] }
```
stefano-garzarella, line 40: Is this used only for tests?

If yes, please move to `dev-dependencies`

kernel/src/testing.rs
```diff
@@ -40,6 +40,8 @@ pub enum IORequest {
     NOP = 0x00,
     /// get SEV-SNP pre-calculated launch measurement (48 bytes) from the host
     GetLaunchMeasurement = 0x01,
+    /// Virtio-blk tests: Get md5 hash of the svsm state disk image
+    GetStateImageMd5 = 0x10,
```
stefano-garzarella, line 44: `0x02` should be fine

kernel/src/testing.rs
```diff
@@ -40,6 +40,8 @@ pub enum IORequest {
     NOP = 0x00,
     /// get SEV-SNP pre-calculated launch measurement (48 bytes) from the host
     GetLaunchMeasurement = 0x01,
+    /// Virtio-blk tests: Get md5 hash of the svsm state disk image
```
stefano-garzarella, line 43: `s/md5/sha256` right?
Can you check other places?

stefano-garzarella (commented):
I've done a quick first review and it already seems to be in good shape. There are several todo's in the commit messages and code, what's the plan for these? Do we fix them before merging or after?
I can help with some!

virtio-drivers/src/device/blk.rs
```diff
@@ -0,0 +1,884 @@
+//! Driver for VirtIO block devices.
```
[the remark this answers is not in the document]
osteffenrh: @joergroedel, how should we handle this? Code is imported with little modifications from the upstream repo. None of them have a SPDX line. Also do we need the author fields?

kernel/src/block/virtio_blk_demo.rs
osteffenrh: I removed the code. It should not get merged anyway. We have the test code that exercises the virtio-blk driver.

osteffenrh: Posting it here in case...
```patch
From 138d7a5854a543e1c6c083da8634a3b38e914505 Mon Sep 17 00:00:00 2001
From: Oliver Steffen <osteffen@redhat.com>
Date: Mon, 24 Mar 2025 10:59:00 +0100
Subject: [PATCH] demo code

demo
---
 kernel/src/block/mod.rs             |  1 +
 kernel/src/block/virtio_blk_demo.rs | 33 +++++++++++++++++++++++++++++
 kernel/src/svsm.rs                  |  6 ++++++
 3 files changed, 40 insertions(+)
 create mode 100644 kernel/src/block/virtio_blk_demo.rs

diff --git a/kernel/src/block/mod.rs b/kernel/src/block/mod.rs
index 213d4f8f..d5620e39 100644
--- a/kernel/src/block/mod.rs
+++ b/kernel/src/block/mod.rs
@@ -7,5 +7,6 @@
 pub mod api;
 pub mod error;
 pub mod virtio_blk;
+pub mod virtio_blk_demo;
 
 pub use error::BlockDeviceError;
diff --git a/kernel/src/block/virtio_blk_demo.rs b/kernel/src/block/virtio_blk_demo.rs
new file mode 100644
index 00000000..d522f1c0
--- /dev/null
+++ b/kernel/src/block/virtio_blk_demo.rs
@@ -0,0 +1,33 @@
+use crate::{
+    address::PhysAddr,
+    block::{api::BlockDriver, virtio_blk::VirtIOBlkDriver},
+    fw_cfg::FwCfg,
+};
+extern crate alloc;
+use alloc::vec::Vec;
+
+pub fn run_demo(fw_cfg: &FwCfg<'_>) {
+    log::info!("Virtio blk demo");
+
+    let addresses = fw_cfg.get_virtio_mmio_addresses().unwrap_or_default();
+
+    log::info!(
+        " Got these virtio mmio addresses from fw_cfg: {:x?}",
+        addresses
+    );
+
+    let mut devices: Vec<VirtIOBlkDriver> = addresses
+        .iter()
+        .inspect(|a| log::info!("  checking for device at {a:016x}"))
+        .filter_map(|a| VirtIOBlkDriver::new(PhysAddr::from(*a)).ok())
+        .inspect(|_d| log::info!("   found a device!"))
+        .collect();
+
+    log::info!(" found {} virtio_blk devides", devices.len());
+
+    if let Some(dev) = devices.pop() {
+        log::info!("First device's size is {} bytes", dev.size());
+    }
+
+    panic!("--- END OF DEMO ---");
+}
diff --git a/kernel/src/svsm.rs b/kernel/src/svsm.rs
index 3f491c8d..b6622f44 100755
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -333,6 +333,12 @@ pub extern "C" fn svsm_main() {
         panic!("Failed to prepare guest FW: {e:#?}");
     }
 
+    #[cfg(not(test))]
+    {
+        use svsm::fw_cfg::FwCfg;
+        svsm::block::virtio_blk_demo::run_demo(&FwCfg::new(SVSM_PLATFORM.get_io_port()));
+    }
+
     vtpm_init().expect("vTPM failed to initialize");
 
-- 
2.48.1

```

kernel/src/virtio/error.rs
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2025 Red Hat, Inc.
+//
+// Author: Oliver Steffen <osteffen@redhat.com>
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum VirtioError {
+    InvalidDeviceType,
```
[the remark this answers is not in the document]
osteffenrh: With some refactoring we can get rid of `InvalidDeviceType`.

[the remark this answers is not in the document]
osteffenrh: > With some refactoring we can get rid of `InvalidDeviceType`.

Not sure if that would make it much nicer. Leaving it in for now, and adding documentation to it.

kernel/src/block/error.rs
osteffenrh: Added documentation.

virtio-drivers/src/device/blk.rs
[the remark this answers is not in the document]
joergroedel: Author field is not mandatory, but please add an SPDX line which matches the license from the source repo.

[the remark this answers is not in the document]
osteffenrh: Done.

osteffenrh:
MMIO read + write are now part of the SvsmPlatform trait and implemented for SNP.

kernel/src/platform/mod.rs
```diff
@@ -190,6 +190,14 @@ pub trait SvsmPlatform {
     fn start_svsm_request_loop(&self) -> bool {
         false
     }
+
+    fn mmio_write(&self, _paddr: PhysAddr, _data: &[u8]) -> Result<(), SvsmError> {
```
p4zuu, line 194: Like you did in virtio/hal/mmio_read/write, I would prefer setting `SvsmPlatform::mmio_write()` and `SvsmPlatform::mmio_read()` unsafe, these 2 could either break memory safety or leak data so callers should ensure safety before use

osteffenrh: Ok, I'll mark the SvsmPlatform mmio functions `unsafe`, also in the SnpPlatform impl, of course.
There the GHCB ones are called. Should those be marked unsafe too?

osteffenrh: Marked everything unsafe.

virtio-drivers/src/device/blk.rs
```diff
@@ -16,7 +16,8 @@ const QUEUE_SIZE: u16 = 16;
 const SUPPORTED_FEATURES: BlkFeature = BlkFeature::RO
     .union(BlkFeature::FLUSH)
     .union(BlkFeature::RING_INDIRECT_DESC)
-    .union(BlkFeature::RING_EVENT_IDX);
+    .union(BlkFeature::RING_EVENT_IDX)
+    .union(BlkFeature::VERSION_1);
```
stefano-garzarella, line 20: We should explain the "why" in the commit.
Something like, “the implementation respects VIRTIO 1.0, so better enable VIRTIO_F_VERSION_1 to be compliant and not force device to use legacy implementation”

kernel/src/error.rs
```diff
@@ -102,6 +103,8 @@ pub enum SvsmError {
     Apic(ApicError),
     /// Errors related to Hyper-V.
     HyperV(u16),
+    /// Errors related to Virtio devices.
```
stefano-garzarella, line 107: s/devices/drivers ?

kernel/src/virtio/devices.rs
```diff
@@ -0,0 +1,65 @@
...
+impl VirtIOBlkDevice {
+    pub fn new(mmio_base: PhysAddr) -> Result<Box<Self>, SvsmError> {
+        virtio_init();
+
+        let mem = map_global_range_4k_shared(mmio_base, PAGE_SIZE, PTEntryFlags::data())?;
+
+        // Not expected to fail, because mem exists.
+        let header = NonNull::new(mem.addr().as_mut_ptr()).unwrap();
+
+        // SAFETY: `header` is the MMIO config area; we have to trust the content is valid.
```
stefano-garzarella, line 44: is it coming from the host, right? (maybe we should mention it)

Should we do some kind of checks/validations? (@joergroedel)

kernel/src/virtio/hal.rs
```diff
@@ -0,0 +1,231 @@
...
+        // TODO: allow more than one page
+        assert!(buffer.len() <= PAGE_SIZE);
+
+        let shared_page = SharedBox::<[u8; PAGE_SIZE]>::try_new_zeroed().unwrap();
+
+        if direction == virtio_drivers::BufferDirection::DriverToDevice {
+            let src = buffer.as_ptr().cast::<u8>();
+            let dst = shared_page.addr().as_mut_ptr::<u8>();
+
+            // SAFETY: Assume `buffer` is valid because it is supplied by the virtio-drivers crate.
```
stefano-garzarella, line 136: Is `buffer` safety demanded to the caller? should we update this message?

kernel/src/virtio/hal.rs
```diff
@@ -0,0 +1,231 @@
...
+        if let Some(shared_page) = SHARED_MEM.lock().get_mut().unwrap().pop(paddr.into()) {
+            let vaddr = phys_to_virt(paddr.into());
+            let va_from_shared = shared_page.addr();
+            assert!(vaddr == va_from_shared);
+
+            if direction == virtio_drivers::BufferDirection::DeviceToDriver {
+                let dst = buffer.as_ptr().cast::<u8>();
+                let src = vaddr.as_mut_ptr::<u8>();
+
+                // SAFETY: Assume `buffer` is valid and can hold at leader `buffer.len()`
```
stefano-garzarella, line 176: s/leader/least ?

I think also here we should mention that `buffer` safety is demanded to the caller.

BTW I don't get it, I'm confused about `can hold at least buffer.len()`. Is it related to `src`?

kernel/src/virtio/hal.rs
```diff
@@ -0,0 +1,231 @@
...
+    ///
+    /// `src` must be properly alinged and reside at a readable memory address.
+    unsafe fn mmio_read<T: FromBytes + Immutable>(src: &T) -> T {
+        let paddr = this_cpu()
+            .get_pgtable()
+            .phys_addr(VirtAddr::from(addr_of!(*src)))
+            .unwrap();
+
+        let mut b = alloc::vec![0; size_of::<T>()];
+        // SAFETY: We are trusting the caller (the virtio driver) to ensure validity of `paddr` and alignment of data.
```
stefano-garzarella, line 205: Are we validating paddr here in the previous line or am I wrong?

kernel/src/virtio/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod devices;
```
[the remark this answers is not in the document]
stefano-garzarella: Still missing, maybe you missed this comment.

scripts/launch_guest.sh
```diff
@@ -36,6 +38,14 @@ while [[ $# -gt 0 ]]; do
       shift
       shift
       ;;
+    --state)
+      STATE_ENABLE="x-svsm-virtio-mmio=on"
```
stefano-garzarella, line 42: Please document this in the commit description or here.
It's "unstable" right?

Cargo.toml
```diff
@@ -60,7 +60,7 @@ intrusive-collections = "0.9.6"
 libfuzzer-sys = "0.4"
 log = "0.4.17"
 p384 = { version = "0.13.0" }
-sha2 = "0.10.8"
+sha2 = { version = "0.10.8", default-features = false }
```
stefano-garzarella, line 63: Should we enable default features where it was used before (i.e. `igvmmeasure/Cargo.toml` ?

kernel/src/platform/mod.rs
```diff
@@ -190,6 +190,26 @@ pub trait SvsmPlatform {
...
     }
+
+    /// Perfrom a write to a memory-mapped IO area
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that `pa` points to a properly aligned memory location and the
+    /// memory accessed is part of a valid MMIO range.
+    unsafe fn mmio_write(&self, _paddr: PhysAddr, _data: &[u8]) -> Result<(), SvsmError> {
+        todo!()
```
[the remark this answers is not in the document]
osteffenrh: Returning `SvsmError::NotImplemented` is also fine with me.

Is it better to stub the mmio functions in all platfrom impls that way, or provide a default implementation of them in the base?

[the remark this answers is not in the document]
osteffenrh: Also, shall we add `NotSupported` to `SvsmError` or would `NotSupported` do as well?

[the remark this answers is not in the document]
stefano-garzarella: > Is it better to stub the mmio functions in all platfrom impls that way, or provide a default implementation of them in the base?

I would add a default implementation, only if it actually does something, so I would always tend to defer to implementations if it's not possible to implement something by default as in this case. Then each implementation decides what's best, whether to return an error or panic if they don't expect this to be called. In the latter case, though, I think it's better to use `panic!()` or `unimplemented!()`.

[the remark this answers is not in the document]
stefano-garzarella: > Also, shall we add `NotSupported` to `SvsmError` or would `NotSupported` do as well?

yep, I guess we can add it

Cargo.toml
```diff
@@ -26,6 +26,8 @@ members = [
     "user/init",
```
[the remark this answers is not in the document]
osteffenrh: Including the history in the coconut repo.

kernel/src/virtio/devices.rs
osteffenrh: Not sure what we can check there. This is based on the MMIO address handed over from Qemu.

osteffenrh:
The change history to the virtio drivers to minimize the code is now also included in this PR. This way we can see what was done and maybe even redo the steps in case we want to sync with the upstream driver code later.

osteffenrh:
Added a cargo feature called `virtio-drivers` to enable the virtio driver and the block device driver (disabled by default).
The in-svsm tests (`make test-in-svsm`) enables it and runs basic read-write checks with a set of different request sizes.

kernel/src/virtio/devices.rs
joergroedel: Yes, I agree that the value is hard to validate. A wrong value is equal to a denial of service from the host.

joergroedel (approved):
Changes look good to me. Thanks for all you efforts @osteffenrh and everyone else involved.

virtio-drivers/LICENSE
stefano-garzarella, line 1: Why removing their license file?

virtio-drivers/src/hal.rs
stefano-garzarella, line 1: There is a `FIXME` in the [commit](https://github.com/coconut-svsm/svsm/pull/635/commits/dfed6badec6a9d58b39c5d4721c918b27c1e1126) description. I think Copy trait is fine.

stefano-garzarella (approved):
I left minor comments, but overall it LGTM! Great work!

osteffenrh:
Addressed the comments and rebased onto main.

stefano-garzarella (approved):

PR #636, opened 2025-03-03
Cleanup X2APIC code and fix IRQ delivery on the Native Platform with QEMU

joergroedel:
Cleanup the X2APIC code by moving accesses for `kernel/src/cpu/x86/apic.rs` and calling the accessors from the native and TDP platform implementations.

Also fix IRQ delivery with QEMU/KVM by setting the ASE bit in APIC.SPIV on X2APIC initialization.

### DIFF
--- a/kernel/src/cpu/x86/apic.rs
+++ b/kernel/src/cpu/x86/apic.rs
@@ -52,3 +52,13 @@ pub fn x2apic_eoi() {
     // SAFETY: writing to EOI MSR doesn't break memory safety.
     unsafe { write_msr(MSR_X2APIC_EOI, 0) };
 }
+
+/// Write a command to the Interrupt Command Register.
+///
+/// # Arguments
+///
+/// - `icr` - The 64-bit value describing the interrupt command.
+pub fn x2apic_icr_write(icr: u64) {
+    // SAFETY: writing to ICR MSR doesn't break memory safety.
+    unsafe { write_msr(MSR_X2APIC_ICR, icr) };
+}
--- a/kernel/src/cpu/x86/apic.rs
+++ b/kernel/src/cpu/x86/apic.rs
@@ -46,3 +46,9 @@ pub fn x2apic_enable() {
         }
     }
 }
+
+/// Send an End-of-Interrupt notification to the X2APIC.
+pub fn x2apic_eoi() {
+    // SAFETY: writing to EOI MSR doesn't break memory safety.
+    unsafe { write_msr(MSR_X2APIC_EOI, 0) };
+}
--- a/kernel/src/cpu/x86/apic.rs
+++ b/kernel/src/cpu/x86/apic.rs
@@ -6,12 +6,18 @@
 // Author: Jon Lange <jlange@microsoft.com>
 // Author: Joerg Roedel <jroedel@suse.de>
 
+use crate::cpu::msr::{read_msr, write_msr};
+
 /// End-of-Interrupt register MSR offset
 pub const MSR_X2APIC_EOI: u32 = 0x80B;
 /// Interrupt-Service-Register base MSR offset
 pub const MSR_X2APIC_ISR: u32 = 0x810;
 /// Interrupt-Control-Register register MSR offset
 pub const MSR_X2APIC_ICR: u32 = 0x830;
+
+const MSR_APIC_BASE: u32 = 0x1B;
+const APIC_ENABLE_MASK: u64 = 0x800;
+const APIC_X2_ENABLE_MASK: u64 = 0x400;
 
 /// Get the MSR offset relative to a bitmap base MSR and the mask for the MSR
 /// value to check for a specific vector bit being set in IRR, ISR, or TMR.
@@ -23,4 +29,20 @@ pub const MSR_X2APIC_ICR: u32 = 0x830;
 pub fn apic_register_bit(vector: usize) -> (u32, u32) {
     let index: u8 = vector as u8;
     ((index >> 5) as u32, 1 << (index & 0x1F))
+}
+
+/// Enables the X2APIC by setting the AE and EXTD bits in the APIC base address
+/// register.
+pub fn x2apic_enable() {
+    // Enable X2APIC mode.
+    let apic_base = read_msr(MSR_APIC_BASE);
+    let apic_base_x2_enabled = apic_base | APIC_ENABLE_MASK | APIC_X2_ENABLE_MASK;
+    if apic_base != apic_base_x2_enabled {
+        // SAFETY: enabling X2APIC mode allows accessing APIC's control
+        // registers through MSR accesses, so enabling it doesn't break
+        // memory safety itself.
+        unsafe {
+            write_msr(MSR_APIC_BASE, apic_base_x2_enabled);
+        }
+    }
 }
--- a/kernel/src/cpu/x86/apic.rs
+++ b/kernel/src/cpu/x86/apic.rs
@@ -13,8 +13,13 @@ pub const MSR_X2APIC_ISR: u32 = 0x810;
 /// Interrupt-Control-Register register MSR offset
 pub const MSR_X2APIC_ICR: u32 = 0x830;
 
-// Returns the MSR offset and bitmask to identify a specific vector in an
-// APIC register (IRR, ISR, or TMR).
+/// Get the MSR offset relative to a bitmap base MSR and the mask for the MSR
+/// value to check for a specific vector bit being set in IRR, ISR, or TMR.
+///
+/// # Returns
+///
+/// A `(u32, u32)` tuple with the MSR offset as the first and the vector
+/// bitmask as the second value.
 pub fn apic_register_bit(vector: usize) -> (u32, u32) {
     let index: u8 = vector as u8;
     ((index >> 5) as u32, 1 << (index & 0x1F))
--- a/kernel/src/cpu/x86/apic.rs
+++ b/kernel/src/cpu/x86/apic.rs
@@ -1,12 +1,17 @@
 // SPDX-License-Identifier: MIT
 //
 // Copyright (c) Microsoft Corporation
+// Copyright (c) SUSE LLC
 //
 // Author: Jon Lange <jlange@microsoft.com>
+// Author: Joerg Roedel <jroedel@suse.de>
 
-pub const APIC_MSR_EOI: u32 = 0x80B;
-pub const APIC_MSR_ISR: u32 = 0x810;
-pub const APIC_MSR_ICR: u32 = 0x830;
+/// End-of-Interrupt register MSR offset
+pub const MSR_X2APIC_EOI: u32 = 0x80B;
+/// Interrupt-Service-Register base MSR offset
+pub const MSR_X2APIC_ISR: u32 = 0x810;
+/// Interrupt-Control-Register register MSR offset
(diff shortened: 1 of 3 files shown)


msft-jlange:
Nice find on the spurious vector enable bit!

There are two things you've done here that I generally do not agree with.  First, the existing `cpu/apic.rs` file contains code exclusively related to APIC emulation, and you have now extended that to include consumption of APIC functionality for the SVSM itself.  For the sake of good partitioning in code organization, I think we will really want to have all APIC emulation code by itself, so that it is clear from reading that the code module is designed exclusively for APIC emulation.  I don't have any issue with you moving APIC consumer code into a common location, but I do not believe it should coexist with the APIC emulator.

Second, your change uses native MSR access and relies on virtualization exception handling to provide platform abstractions.  COCONUT-SVSM is designed as a fully enlightened system and is aware of confidential platforms, and I believe we will want to move in a direction where we minimize the logic in the virtualization exception handlers only to those instructions where we have no reasonable alternative to take advantage of platform abstraction.  The virtualization exception handlers do not have any context about the calling code and cannot make robust security decisions about whether any particular emulated action is safe, and therefore we should avoid requiring them as much as possible.  All of the APIC consumer code in COCONUT-SVSM has access to the platform abstractions and should use them for host access.  This was my rationale for putting the existing APIC code into the platform abstraction (where the abstraction boundary is high level - implying APIC behavior - and not low level like MSR access - and this enhances robustness).

I would like to see your change adhere to the principles described here, or at least to have a discussion about whether these principles are appropriate.

joergroedel:
> Nice find on the spurious vector enable bit!
> 
> There are two things you've done here that I generally do not agree with. First, the existing `cpu/apic.rs` file contains code exclusively related to APIC emulation, and you have now extended that to include consumption of APIC functionality for the SVSM itself. For the sake of good partitioning in code organization, I think we will really want to have all APIC emulation code by itself, so that it is clear from reading that the code module is designed exclusively for APIC emulation. I don't have any issue with you moving APIC consumer code into a common location, but I do not believe it should coexist with the APIC emulator.

I agree with that and moved the APIC consumer code to `kernel/src/cpu/x86/apic.rs`. This file already contained some definitions for the existing APIC consumers. The `kernel/src/cpu/apic.rs` file instead rightly contains the APIC emulation code, so both use-cases are implemented in different files.

> Second, your change uses native MSR access and relies on virtualization exception handling to provide platform abstractions. COCONUT-SVSM is designed as a fully enlightened system and is aware of confidential platforms, and I believe we will want to move in a direction where we minimize the logic in the virtualization exception handlers only to those instructions where we have no reasonable alternative to take advantage of platform abstraction. The virtualization exception handlers do not have any context about the calling code and cannot make robust security decisions about whether any particular emulated action is safe, and therefore we should avoid requiring them as much as possible. All of the APIC consumer code in COCONUT-SVSM has access to the platform abstractions and should use them for host access. This was my rationale for putting the existing APIC code into the platform abstraction (where the abstraction boundary is high level - implying APIC behavior - and not low level like MSR access - and this enhances robustness).

Is my understanding correct that AMD SEV-SNP without Secure AVIC is the only platform where the X2APIC MSRs are intercepted (and thus need special handling)?

> I would like to see your change adhere to the principles described here, or at least to have a discussion about whether these principles are appropriate.

I agree in general that avoiding #VC/#VE exceptions is a reasonable goal, but it is also often a tradeoff for platform-dependent performance optimization. In the case of X2APIC most platforms (all where X2APIC MSRs are not intercepted) will get better performance by using `rdmsr/wrmsr` because there are less indirect calls and compiler inlining. While introducing an abstraction to do direct GHCB calls makes it perform better on platforms where the MSR are intercepted.

We can discuss what the right trade-off in this particular case is and address it in a follow-up PR.

msft-jlange:
> I agree with that and moved the APIC consumer code to `kernel/src/cpu/x86/apic.rs`. This file already contained some definitions for the existing APIC consumers. The `kernel/src/cpu/apic.rs` file instead rightly contains the APIC emulation code, so both use-cases are implemented in different files.

Apologies, it looks like I misread the PR and thought you had put all of this into the same file.  Your proposed organization is good.

> Is my understanding correct that AMD SEV-SNP without Secure AVIC is the only platform where the X2APIC MSRs are intercepted (and thus need special handling)?

No, when running native builds under Hyper-V, we would like to take advantage of the "lazy EOI" optimization and perform EOI by writing to the VP Assist Page instead of the EOI MSR (I haven't implemented that yet but I would like to do so).  This is equivalent to how `NoEoiRequired` works with SEV-SNP Restricted Injection.

Note that on SEV-SNP, the #VC handler would require special logic to determine whether the target MSR is APIC EOI, and if it is, then it would have to detect whether Restricted Injection is active, and if so, to redirect the APIC write to an update to `NoEoiRequired`.  Placing this logic into the #VC handler violates the principle that APIC logic should live in the APIC layer.  In my mind, the best path forward would be for the platform abstraction to continue to implement EOI, though the other APIC functions (such as initialization) could still move into a common location because they do not require platform-specific logic.

> We can discuss what the right trade-off in this particular case is and address it in a follow-up PR.

Before we submit this PR, we should recognize that our TDX support does not currently have a #VE handler implementation, and the one proposed in the active PR only covers CPUID and not MSR accesses.  Consequently, this PR would break TDX.  I wonder whether we should just submit the native initialization fix first and then submit the reorganization after we have a solid plan for virtualization exceptions on all platforms.

kernel/src/cpu/x86/apic.rs
```diff
@@ -1,16 +1,99 @@
...
+        write_msr(MSR_APIC_BASE, apic_base_x2_enabled);
+    }
+    // Set SW-enable in SPIV to enable IRQ delivery
+    x2apic_sw_enable();
+}
+
+/// Send an End-of-Interrupt notification to the X2APIC.
+pub fn x2apic_eoi() {
+    write_msr(MSR_X2APIC_EOI, 0);
+}
```
msft-jlange, line 56: If we keep EOI in the platform abstraction, then this could be written like this and avoid #VC in the SNP case.

```suggestion
pub fn x2apic_eoi(wrmsr: FnOnce<(u32, u64)>) {
    wrmsr(MSR_X2APIC_EOI, 0);
}
```

Passing a closure to perform the WRMSR permits each platform to implement this optimally without requiring any virtualization exception.

joergroedel:
> No, when running native builds under Hyper-V, we would like to take advantage of the "lazy EOI" optimization and perform EOI by writing to the VP Assist Page instead of the EOI MSR (I haven't implemented that yet but I would like to do so). This is equivalent to how `NoEoiRequired` works with SEV-SNP Restricted Injection.

Okay, given these requirements I think the best way forward is to introduce an `X86Apic` trait and also implementations to overwrite certain operations with platform optimised variants. That is something I thought about while working on this PR already, but I omitted it because of time required to make all the changes (and because it is not a blocker yet). I left it for a future PR for now.

> Note that on SEV-SNP, the #VC handler would require special logic to determine whether the target MSR is APIC EOI, and if it is, then it would have to detect whether Restricted Injection is active, and if so, to redirect the APIC write to an update to `NoEoiRequired`. Placing this logic into the #VC handler violates the principle that APIC logic should live in the APIC layer. In my mind, the best path forward would be for the platform abstraction to continue to implement EOI, though the other APIC functions (such as initialization) could still move into a common location because they do not require platform-specific logic.

Yes, agreed. Putting special logic into VE/VC handlers is not a good or maintainable solution.

> Before we submit this PR, we should recognize that our TDX support does not currently have a #VE handler implementation, and the one proposed in the active PR only covers CPUID and not MSR accesses. Consequently, this PR would break TDX. I wonder whether we should just submit the native initialization fix first and then submit the reorganization after we have a solid plan for virtualization exceptions on all platforms.

TDX has a hardware-virtualised APIC, does accessing the X2APIC MSRs still cause VE? I don't think so, because one of the patches in this PR moves a direct X2APIC MSR access out of the TDP code into `x86/apic.rs`. Functionally this PR should not change any behaviour, it just adds some indirection.

msft-jlange:
> Okay, given these requirements I think the best way forward is to introduce an `X86Apic` trait and also implementations to overwrite certain operations with platform optimised variants. That is something I thought about while working on this PR already, but I omitted it because of time required to make all the changes (and because it is not a blocker yet). I left it for a future PR for now.

I don't think we need to go that far.  As a first step, I'm advocating leaving the APIC code common as you have done, and only making EOI platform-specific, as this should be much simpler (and more efficient) than creating a separate trait that has to be implemented separately by every platform.  And we could even keep the EOI code in the common APIC code if that common EOI logic knew that it was supposed to call the platform to do the actual EOI logic.  In fact, you could imagine something like this, where the platform abstraction exposes a function like this:
```
fn platform_eoi(common_handler: FnOnce(FnOnce(u32, u64)));
```

then the common EOI code would do this:
```
pub fn eoi() {
    SVSM_PLATFORM.platform_eoi(|wrmsr_fn| { wrmsr_fn(APIC_MSR_EOI, 0) });
}
```

and the implementation of platform_eoi might look like this (on native, for example):
```
fn platform_eoi(common_handler: FnOnce(FnOnce(u32, u64))) {
    common_handler(|msr, value| { wrmsr(msr, value) });
}
```

This needs some work, but it would allow us to keep APIC knowledge in the APIC layer while keeping platform knowledge in the platform layer, using closures to avoid the explosion of additional traits that make things more complex.

> TDX has a hardware-virtualised APIC, does accessing the X2APIC MSRs still cause VE? I don't think so, because one of the patches in this PR moves a direct X2APIC MSR access out of the TDP code into `x86/apic.rs`. Functionally this PR should not change any behaviour, it just adds some indirection.

You are correct.  I misspoke earlier (and edited my comment) because access to the EOI MSR on TDX will not result in a #VE.

joergroedel:
> and the implementation of platform_eoi might look like this (on native, for example):
> 
> ```
> fn platform_eoi(common_handler: FnOnce(FnOnce(u32, u64))) {
>     common_handler(|msr, value| { wrmsr(msr, value) });
> }
> ```

This is still a very special-purpose solution, longer term this needs a more complete abstraction that could also cover a fully paravirtualized APIC. For SNP without AVIC we need to overwrite all MSR reads/writes with GHCB calls, for example. For now I can live with this solution, though.

kernel/src/platform/tdp.rs
```diff
@@ -200,10 +199,7 @@ impl SvsmPlatform for TdpPlatform {
     fn eoi(&self) {}
```
msft-jlange, line 199: Should we implement this while we're at it?
```suggestion
    fn eoi(&self) {
        x2apic_eoi();
    }
```

joergroedel:
> This is still a very special-purpose solution, longer term this needs a more complete abstraction that could also cover a fully paravirtualized APIC. For SNP without AVIC we need to overwrite all MSR reads/writes with GHCB calls, for example. For now I can live with this solution, though.

Okay, I take that back and implemented a more generic APIC infrastructure in PR #639 .

kernel/src/platform/tdp.rs
joergroedel: Yes, that makes sense. I added it.

kernel/src/cpu/x86/apic.rs
joergroedel: As per our discussion, lets solve this in PR #639

PR #637, opened 2025-03-04
Enable UMIP

peterfang:
UMIP (User-Mode Instruction Prevention) prevents user-mode applications from executing SGDT/SIDT/SLDT/SMSW/STR. A `#GP(0)` is raised if any of these instrctions is executed in CPL > 0 and CR4.UMIP = 1. It is generally a good idea to enable such prevention before user space comes up because these instructions can leak critical information regarding memory layout and configuration.


joergroedel (approved):
LGTM. Also nice cleanup of the CPU feature checking functions, thanks for that.

PR #639, opened 2025-03-07
Introduce APIC Driver Infrastructure

joergroedel:
This PR implements a generic driver infrastructure for X86 APICs. It allows to provide drivers for APICs with a hardware interface (XAPIC, X2APIC), paravirtualized APICs, or a mixture of both.

To achieve this, the APIC handling currently implemented in platform code is moved out to the new infrastructure and made accessible via the PerCpu code. The PR is based on the changes in PR #636 .

Two APIC drivers are implemented with this PR:

1. A driver using the X2APIC hardware interface.
2. A driver for AMD SEV-SNP which uses the X2APIC interface via direct GHCB calls and special EOI  handling in case RestrictedInjection is enabled.

One important change to mention is that this PR also sets an APIC driver for the Intel TDP platform, which results in the enable bits in MSR_APIC_BASE being set.


kernel/src/cpu/x86/x2apic.rs
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    /// Creates new instance of [`X2Apic`] which does not need to be enabled
+    /// via MSR_APIC_BASE.
+    pub fn new_skip_msr_enable() -> Self {
+        Self {
+            skip_msr_enable: true,
+        }
+    }
```
peterfang, line 54: Since I'm reworking https://github.com/coconut-svsm/svsm/pull/657, the native x2apic driver won't be needing this function anymore since it's only used for native platforms and they always start in xAPIC mode. On the other hand, TDP platforms will always assume that LAPIC starts in x2APIC mode (this is true for both Hyper-V and KVM hosts).

peterfang (commented):
Comment posted

joergroedel:
@msft-jlange Can you please test these changes on the various Hyper-V configurations?

kernel/src/cpu/x86/x2apic.rs
peterfang: `MSR_APIC_BASE` accesses will always result in #VEs. In my latest version of https://github.com/coconut-svsm/svsm/pull/657, the TDP APIC driver always uses GHCI for `MSR_APIC_BASE` accesses.

peterfang: > Okay, the flag is needed to enable X2APIC mode on native, while this can be skipped on TDP because that already starts in X2APIC mode.
> 
> The question I have is, does TDX/TDP handle writes to the MSR_APIC_BASE or will that cause problems?

Wouldn't `skip_msr_enable` always be false on native platforms?

peterfang:
Just as a side note, I find it a bit confusing that there's an `LApic` in `kernel/src/cpu/x86/apic.rs` and also an `LocalApic` in `kernel/src/cpu/apic.rs` and both are being used by `PerCpu`. But this may not be in the scope of this PR.

msft-jlange:
> Just as a side note, I find it a bit confusing that there's an `LApic` in `kernel/src/cpu/x86/apic.rs` and also an `LocalApic` in `kernel/src/cpu/apic.rs` and both are being used by `PerCpu`. But this may not be in the scope of this PR.

We can rename one of them to `GuestLocalApic` in a future PR.  I think that would resolve the confusion.

msft-jlange:
I worry a little bit about the use of generics in this design - it promotes a lot of code duplication, which will just get worse once TDX implements `GHCIApicAccessor`.  I wonder whether a better design would be for `X86Apic` to be an object, not a trait, and to have it contain a member which is a `&dyn ApicAccess`.  This would eliminate the distinction between `X86Apic`, `X86ApicDriver`, and `RawX86Apic`, and would prevent multiple instances of `RawX86Apic<A>`.

kernel/src/cpu/x86/x2apic.rs
peterfang: I think you already alluded to the generics design in this PR in your comment. I implemented the `X86Apic` and `ApicAccess` traits for TDP APIC instead of extending the existing `X2Apic` driver since some of TDP's APIC registers require GHCI if we want to avoid #VEs.

joergroedel:
@msft-jlange Can you please test these changes on the various Hyper-V configurations?

> I worry a little bit about the use of generics in this design - it promotes a lot of code duplication, which will just get worse once TDX implements `GHCIApicAccessor`. I wonder whether a better design would be for `X86Apic` to be an object, not a trait, and to have it contain a member which is a `&dyn ApicAccess`. This would eliminate the distinction between `X86Apic`, `X86ApicDriver`, and `RawX86Apic`, and would prevent multiple instances of `RawX86Apic<A>`.

Okay, I changed the implementation to have a single `X86Apic` object per CPU and all special casing now happens in the implementations of the `ApicAccessor` trait. @msft-jlange and @peterfang, can you please have a look and test this on your side? I hope to merge this soon.

kernel/src/cpu/x86/x2apic.rs
joergroedel: This will need some updating with the latest changes, but it should only get simpler to implement a TDP specific APIC driver with them. The `skip_msr_enable` flag is gone as well.

msft-jlange:
Unfortunately, this doesn't work under SNP with shadow stacks enabled.  In SNP, the call to `PerCpu::setup_on_cpu()` ends up in `SnpPlatform::setup_percpu_current()` which calls `apic_enable()`.  Setting aside the question of why APIC enablement is in platform-specific logic instead of common logic, `apic_enable()` relies on #VC handling to complete the MSR accesses.  When shadow stacks are enabled, the #VC handler executes WRSS to adjust the shadow stack prior to the IRET from the #VC handler.  Unfortunately, at this point in CPU setup, we have not yet called `enable_shadow_stacks()` on the BSP, so the IA32_S_CET MSR has not yet been configured at all - certainly not with the WR_SHSTK_EN bit, so the attempt to execute WRSS results in #UD.  That is a fatal exception.

I believe the correct approach is to extend the `ApicAccess` trait to enable reads and writes to the APIC_BASE MSR in addition to the APIC MSRs.  You could even pass in the MSR number as a parameter so the MSR identifier can remain within the APIC driver instead of the accessor.  But I think the native MSR access will have to be avoided.

kernel/src/cpu/x86/snp_apic.rs
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: MIT
```
msft-jlange, line 1: Can we move this file to the `kernel/src/sev` directory?  In the long run, I think we will prefer all SNP-specific code to be in a single directory (same with TDX and native) so that when compiling a binary for a single platform, we can easily use `cfg` targets to splice out entire directories - much easier than doing it file by file.

joergroedel: Done.

msft-jlange:
> Thanks for testing this, I followed your suggestion and added a method to the `ApicAccess` trait to update the APIC_BASE MSR.

The Hyper-V tests now pass.  I'll review the rest of the code later today and then sign off.

kernel/src/cpu/percpu.rs
```diff
@@ -356,6 +357,9 @@ pub struct PerCpu {
     /// address space.
     shared_global: OnceCell<&'static PerCpuShared>,
 
+    /// APIC access object
+    apic: X86Apic,
```
msft-jlange, line 361: Why is this per-CPU?  The only contents of this object are a pointer to a dynamic APIC accessor trait, which does not appear to vary from CPU to CPU.  Would it be better if this were a static?

joergroedel: It is not strictly necessary now, but since the APIC is fundamentally a per-CPU resource it makes sense to store it alongside other per-CPU data. This also allows to move other other platform-specific and APIC-related state into that structure (like the HV doorbell page state).

joergroedel:
Rebased to latest main branch.

kernel/src/cpu/percpu.rs
msft-jlange: I'm not convinced, but I'm not going to object.  I foresee a future where we need to define an architecture-specific extension to the `PerCpu` so we can have a flavor for SNP, a flavor for TDX, etc.  I think when that happens, we'll have something like an `enum PerCpuArch` that will include HV doorbell information in the SNP flavor, and this will leave the APIC as nothing more than a shell for the system-wide dynamic trait object.  But we can clean that up in the future if that's where we end up.

joergroedel: Agreed, we can clean up things in the future if needed.

PR #641, opened 2025-03-11
Revert "cpu: compile for shadow stack support by default"

joergroedel:
The commit makes shadow-stack support be compiled in for all builds. Unfortunately there is an issue in the shadow-stack support code which makes the SVSM kernel fail to boot in an SEV-SNP environment on KVM. Revert this patch until the issue is resolved.

This reverts commit 2a73e041e6617a3691207cf7d319d088d0af78ae.

Cc @msft-jlange


joergroedel:
The issue is probably insufficient host support for ShadowStack virtualization (maybe the MSRs are intercepted and cause #VC). I am investigating.

joergroedel:
And indeed, disabling the shadow stack MSR intercepts on the host makes the SVSM boot again.

joergroedel:
Problem is fixed with a host kernel update. This an be closed.

PR #644, opened 2025-03-17
igvmbuilder: prevalidate kernel hashes area

osteffenrh:
Make the kernel hashes memory area of OVMF as prevalidated. This makes
it accessible by the firmware and thus the kernel hashes feature usable.

The AmdSev variant of OVMF for example relies on this area to be
readable.


osteffenrh:
Cc: @kraxel

kraxel:
Makes sense.  Gets things going with an unused/empty kernel hashes page.

Related question: How will these hashes be added to the image at load? IGVM has the concept of 'parameters' for things which can't be determined at compile time (memory map for example). Should IGVM declare the kernel hashes being parameters?

kraxel (approved):

PR #645, opened 2025-03-18
Update verification to use latest verus syntax && generalize and move align proofs to fix verification for address alignment

ziqiaozhou:
* Verus unified the attribute-based and existing macro styles.
* Fix broken verification due to changes in address and util.
* Move align proofs into util module.
* Use vstd ops (new change in verus supports operators)
* Add trusted SpecPartialOrd and SpecPartialCmp for VirtAddr and PhysAddr (done by verus macro)
* Remove assumption for VirtAddr::sub(VirtAddr).
* Update verus version to latest version.
* Update verusfmt version to v0.5.3 to use latest syntax.


kernel/src/address.rs
```diff
@@ -430,17 +524,18 @@ impl<T> From<*mut T> for VirtAddr {
...
-    #[requires(self.sub_requires(other))]
-    #[ensures(|ret: InnerAddr| self.sub_ensures(other, ret))]
-    fn sub(self, other: VirtAddr) -> Self::Output {
-        sign_extend(self.0 - other.0)
+    fn sub(self, other: VirtAddr) -> InnerAddr {
+        proof! {
+            use_type_invariant(self);
+            use_type_invariant(other);
+        }
+        (self.0 - other.0) & ((1usize << (SIGN_BIT + 1)) - 1)
```
ziqiaozhou, line 538: I modified this function implementation since I noticed that the use of sub(vaddr1, vaddr2) is used to get the available memory size between two addresses in the existing code in existing code. The original implementation only makes sense if they are both high/low addresses. The new implementation makes more sense no matter whether the two addresses are both high/low addresses. This helps to drop some assumptions in the code (e.g., MemoryRegion::get_virt_offset inside mm/alloc.rs).

kernel/src/address.rs
```diff
@@ -135,38 +164,50 @@ pub trait Address:
...
+            from_spec::<_, usize>(*self) + size - 1 <= usize::MAX,
+        returns
+            crosses_page_spec(*self, size),
     )]
-    #[ensures(|ret: bool | ret == crosses_page_spec(*self, size))]
     fn crosses_page(&self, size: usize) -> bool {
         let start = self.bits();
         let x1 = start / PAGE_SIZE;
-        let x2 = (start + size - 1) / PAGE_SIZE;
+        let x2 = (start + (size - 1)) / PAGE_SIZE;
```
ziqiaozhou, line 177: Modified the crosses_page implementation to avoid potential integer overflow to ensure panic-free code with verification.

kernel/src/address.rs
joergroedel: Thanks for the fix!

PR #648, opened 2025-03-19
kernel/cpu/vc: support MMIO operations

p4zuu:
Since we already decode MMIO instructions, we can emulate them in the runtime #VC handler. MMIO NAE is currently only handled if it's raised from kernelspace.

**I don't have access to a SEV-SNP platform, I would need someone to test it, test-in-svsm included. Thanks!**

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -30,6 +30,7 @@ pub const SVM_EXIT_CPUID: usize = 0x72;
 pub const SVM_EXIT_IOIO: usize = 0x7b;
 pub const SVM_EXIT_MSR: usize = 0x7c;
 pub const SVM_EXIT_RDTSCP: usize = 0x87;
+pub const SVM_EXIT_NPF: usize = 0x400;
 pub const X86_TRAP_DB: usize = 0x01;
 pub const X86_TRAP: usize = SVM_EXIT_EXCP_BASE + X86_TRAP_DB;
 
@@ -154,6 +155,7 @@ pub fn handle_vc_exception(ctx: &mut X86ExceptionContext, vector: usize) -> Resu
         (SVM_EXIT_MSR, Some(ins)) => handle_msr(ctx, ghcb, ins),
         (SVM_EXIT_RDTSC, Some(DecodedInsn::Rdtsc)) => ghcb.rdtsc_regs(&mut ctx.regs),
         (SVM_EXIT_RDTSCP, Some(DecodedInsn::Rdtsc)) => ghcb.rdtscp_regs(&mut ctx.regs),
+        (SVM_EXIT_NPF, Some(DecodedInsn::Mov)) => handle_mmio(ctx, &insn_ctx),
         _ => Err(VcError::new(ctx, VcErrorType::Unsupported).into()),
     }?;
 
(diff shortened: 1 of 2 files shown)


msft-jlange:
What are the devices that we expect to support using MMIO emulation?  Why do we think that relying on the MMIO #VC handler - which has no context at all about the security assumptions of the underlying request - provides a safe path for handling MMIO emulation?  Another alternative approach would be for MMIO #VC exceptions to be fatal, and to require consumers of MMIO to be sufficiently enlightened to make the host calls as required - why would automatic #VC handling be superior to requiring pervasive enlightenment?

msft-jlange:
> I don't see what you mean, sorry. I agree that emulating automatic MMIO #VC opens a insecure door, that's why I restricted it to CPL 0, but how would you restrict MMIO handling to "sufficiently enlightened" consumers?

What I mean is that instead of having CPL 0 access MMIO addresses directly, we could add `mmio_read` and `mmio_write` methods to the `SvsmPlatform` trait which would perform the GHCB/GHCI operations as required.  We could also decorate these functions as `unsafe` so that all callers take explicit responsibility for evaluating whether any particular access is safe and secure. 
 This ensures that the responsibility for determining safety lies solely in the hands of whichever driver is performing the access and not in some common code that is unaware of the meaning of each access.

I agree that restricting this to CPL 0 is far superior to giving user mode the unrestricted ability to interact with MMIO space, possibly unexpectedly, but I still worry about the potential for erroneous accesses from CPL 0.

You could argue that memory safety inherent to Rust would prevent any erroneous access anywhere, because safe code could never obtain a pointer to MMIO space without knowing what it was getting.  But this applies only to safe code, and we have plenty of unsafe code (out of necessity, to support core OS functionality) and any bugs in that code that result in stray MMIO accesses would be made far, far worse if the #VC handler would blindly perform the MMIO emulation.  As a principle for making all of our code as robust as possible, I still believe that platform methods to perform explicit MMIO emulation will always be more faithful to the intention and will do a better job shielding us from unexpected bugs.

p4zuu:
> What I mean is that instead of having CPL 0 access MMIO addresses directly, we could add `mmio_read` and `mmio_write` methods to the `SvsmPlatform` trait which would perform the GHCB/GHCI operations as required. We could also decorate these functions as `unsafe` so that all callers take explicit responsibility for evaluating whether any particular access is safe and secure. This ensures that the responsibility for determining safety lies solely in the hands of whichever driver is performing the access and not in some common code that is unaware of the meaning of each access.
> 
> I agree that restricting this to CPL 0 is far superior to giving user mode the unrestricted ability to interact with MMIO space, possibly unexpectedly, but I still worry about the potential for erroneous accesses from CPL 0.

Thanks for the further explanation. I didn't really think of the explicit platform `mmio_read/write()` solution but I fully agree, this is much safer than opening the emulation door. It seems that it's already what @osteffenrh implemented in #635 (missing adding `mmio_read/write` methods in `SvsmPlatform` trait, and making them unsafe which I also agree with).
@osteffenrh would you agree adding this in your PR and would this solution work for you? I mean only using the platform MMIO methods and leaving the automatic MMIO #VC unsupported, which is already the case btw and is already panicking if raised.

osteffenrh:
> @osteffenrh would you agree adding this in your PR and would this solution work for you? I mean only using the platform MMIO methods and leaving the automatic MMIO #VC unsupported, which is already the case btw and is already panicking if raised.

Right, the PR already does explicit mmio accesses. I handle that in the virtio-Hal implementation. I would be open to make use of mmio functions provided by the SvsmPlatform instead. Does anybody want to implement those or should I do that as part of the virtio PR?

p4zuu:
> Right, the PR already does explicit mmio accesses. I handle that in the virtio-Hal implementation. I would be open to make use of mmio functions provided by the SvsmPlatform instead. Does anybody want to implement those or should I do that as part of the virtio PR?

It seems that your PR only misses the `SvsmPlatform` new methods so I think it's easier to add those in your PR if you can but I'm happy to do it if you prefer. I don't think there is other MMIO needs for now so I guess we can leave it in your PR

p4zuu:
Since @osteffenrh is adding `mmio_read/write()` methods to the platform trait to perform explicit MMIO read/write in #635, this PR is useless now. Moreover, MMIO #VC is already unsupported and raising one should result with a panic which is the intended behavior for now. Closing

PR #649, opened 2025-03-19
task: Remove alignment requirement for TaskVirtualRegionGuard indices

AdamCDunlap:
Each TaskVirtualRegionGuard gets a unique index that indicates where its vaddr region is. Previously only even indices were allocated since the alloc function was given an alignment of 1, but this is unncessary as every index is valid.

This doubles the number of possible tasks from 512 to 1024. By default, 3 tasks are created per vCPU, so this doubles the number of vCPUs supported from 170 to 341 (in theory). Actually tested with 224 vCPUs.


msft-jlange:
Thanks for catching and fixing my mistake!

msft-jlange (approved):

Issue #650, opened 2025-03-20
Core query protocol incorrect about core protocol itself


deeglaze:
the CORE_PROTOCOL value `1` in kernel/src/protocols/core.rs does not match the spec's 0. There is also no specified APIC protocol :(

joergroedel:
Fix has been merged. Closing.

PR #651, opened 2025-03-21
Fix core_protocol value in query implementation

deeglaze:
The individual protocol constants in the top level protocol module are the correct values.

Fixes Issue #650


deeglaze:
I'm not sure if the protocol numbers should be defined within each protocol implementation file and them exported in the mod.rs, or just use the constant from the mod.rs.

deeglaze:
@joergroedel thoughts?

joergroedel (approved):
LGTM. Thanks for the fix!

PR #652, opened 2025-03-21
cpu/ipi: do not target offline CPUs

msft-jlange:
When sending an IPI to all CPUs or all-but-self, it may be the case that the global CPU array is tracking CPUs that are not yet online.  These CPUs should be ignored when sending IPIs because they are not able to respond to IPI requests.

### DIFF
--- a/kernel/src/cpu/ipi.rs
+++ b/kernel/src/cpu/ipi.rs
@@ -364,16 +364,22 @@ pub fn send_ipi(
             }
         }
         _ => {
+            let mut target_count: usize = 0;
             for cpu in PERCPU_AREAS.iter() {
-                ipi_board.pending.fetch_add(1, Ordering::Relaxed);
-                cpu.as_cpu_ref().ipi_from(sender_cpu_index);
+                // Ignore the current CPU and CPUs that are not online.
+                let cpu_shared = cpu.as_cpu_ref();
+                if cpu_shared.is_online() && cpu_shared.apic_id() != this_cpu().get_apic_id() {
+                    target_count += 1;
+                    cpu_shared.ipi_from(sender_cpu_index);
+                }
             }
-            send_interrupt = true;
 
-            // Remove the current CPU from the target set and completion
-            // calculation, since no interrupt is required to ensure that
-            // IPI handlng can be performed locally.
-            ipi_board.pending.fetch_sub(1, Ordering::Relaxed);
+            // Record the count of targets that will need to respond before
+            // this IPI can complete.
+            ipi_board.pending.store(target_count, Ordering::Relaxed);
+
+            // Send an interrupt only if there are targets to receive it.
+            send_interrupt = target_count != 0;
 
             // Only include the current CPU if requested.
             if let IpiTarget::All = target_set {


kernel/src/cpu/ipi.rs
```diff
@@ -364,16 +364,22 @@ pub fn send_ipi(
...
+            let mut target_count: usize = 0;
             for cpu in PERCPU_AREAS.iter() {
-                ipi_board.pending.fetch_add(1, Ordering::Relaxed);
-                cpu.as_cpu_ref().ipi_from(sender_cpu_index);
+                // Ignore the current CPU and CPUs that are not online.
+                let cpu_shared = cpu.as_cpu_ref();
+                if cpu_shared.is_online() && cpu_shared.apic_id() != this_cpu().get_apic_id() {
+                    target_count += 1;
+                    cpu_shared.ipi_from(sender_cpu_index);
+                }
```
peterfang, line 374: Would something like this be a bit cleaner?
```suggestion
            for cpu in PERCPU_AREAS
                .iter()
                .map(|c| c.as_cpu_ref())
                .filter(|c| c.is_online() && c.apic_id() != this_cpu().get_apic_id())
            {
                target_count += 1;
                cpu.ipi_from(sender_cpu_index);
            }
```

kernel/src/cpu/ipi.rs
```diff
@@ -364,16 +364,22 @@ pub fn send_ipi(
             }
         }
         _ => {
+            let mut target_count: usize = 0;
             for cpu in PERCPU_AREAS.iter() {
-                ipi_board.pending.fetch_add(1, Ordering::Relaxed);
-                cpu.as_cpu_ref().ipi_from(sender_cpu_index);
+                // Ignore the current CPU and CPUs that are not online.
+                let cpu_shared = cpu.as_cpu_ref();
+                if cpu_shared.is_online() && cpu_shared.apic_id() != this_cpu().get_apic_id() {
```
peterfang, line 371: Since `sender_cpu_index` is already used as input, does it make sense to use `cpu_shared.cpu_index() != sender_cpu_index` instead?

peterfang (commented):
Added some minor comments

kernel/src/cpu/ipi.rs
msft-jlange: I have a change coming that will overhaul all uses of local APIC ID and replace most of them with references to the CPU index (which will cover this code as well).  However, I can't submit that until some of my open PRs merge first.  I'd prefer to make the change comprehensively instead of tackling this one separately.

kernel/src/cpu/ipi.rs
msft-jlange: That's a great question.  I don't know how `map` and `filter` affect stack consumption and iterator performance.  It's easy for me to imagine that both `map` and `filter` cause the compiler to generate separate iterator objects, which implies that the loop iteration may occur multiple times (once for each iterator) and that stack consumption grows to accommodate the new iterator objects.  Why take the risk of such inefficiencies for such a small change in readability?

peterfang: I don't have issues with this change the way it is. When I was working on a different project I was asked numerous times by the Rust reviewers to avoid C-like coding styles like this (which is natural to me as a C programmer) and was constantly encouraged to use iterators instead. I'm not sure if the Rust compiler does anything to address your stack consumption concerns. Since I'm not the enforcer of how Rusty this project should be I'll just leave it at that.

msft-jlange: In my mind, the two things that matter more than anything in a code base are how maintainable the code is, and how well it performs.  I don't personally see a connection between deep embrace of a language's peculiar syntax and maintainability - as long as someone reading the code can clearly understand what it does and can modify it if necessary, then we've met the maintainability goal.  I would be all for embracing Rust-specific iteration constructs if it made the code more maintainable but I don't think it matter much in this case, just as you are saying.  Rust for the sake of Rust is not very compelling by itself; Rust for the sake of safety and maintainability makes much more sense.
 
If I had more time, I would look at the optimized code generation for both examples to see which one is more efficient and would go with that, but I think that will have to be left as an exercise for a future date.  My AI buddy tells me that code comparisons within the `foreach` loop are likely to be more efficient, but who knows if that's actually the truth.

PR #653, opened 2025-03-21
Add safe memory constructs for large requests 

deeglaze:
With arbitrary length memory access in service requests, we need to be able to read/write memory across multiple pages safely.

This is groundwork for the attestation service to write manifests and read "selectors" of arbitrary length.

### DIFF
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -20,7 +20,7 @@ use core::arch::asm;
 use core::ffi::c_char;
 use core::mem::{size_of, MaybeUninit};
 use syscall::PATH_MAX;
-use zerocopy::FromBytes;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
 #[inline]
 pub fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
@@ -523,6 +523,80 @@ pub fn copy_slice_to_guest(src: &[u8], dst: PhysAddr) -> Result<(), SvsmError> {
         let destination = guard.virt_addr().bits() + offset;
         copy_bytes(source, destination, size)
     }
+}
+
+/// Reads a vector of bytes from a physical address region outside of SVSM use.
+///
+/// # Arguments
+///
+/// * `src`: The physical address designating the start of continguous physical
+///   memory to read from.
+/// * `size`: The length of the physical address region to read into a vector.
+///
+/// # Returns
+///
+/// This function returns a `Result` that indicates the success or failure of the operation.
+/// On success, returns a vector of length `size`.
+/// If the physical address region cannot be mapped, it returns `Err(SvsmError::Mem)`.
+/// If the physical address region cannot be read, it returns `Err(SvsmError::Fault)`.
+/// If the physical address region is not allocated to the guest, it returns
+///   `Err(SvsmError::InvalidAddress)`.
+pub fn read_bytes_from_guest(src: PhysAddr, size: usize) -> Result<Vec<u8>, SvsmError> {
+    let mut result = Vec::with_capacity(size);
+    // SAFETY: The vector's capacity, `size` has been populated by copy_from_guest.
+    // The translation of bytes to T are checked.
+    unsafe {
+        copy_from_guest(src, result.as_mut_ptr(), size)?;
+        result.set_len(size);
+    }
+    Ok(result)
+}
+
+/// Reads an instance of T from a physical address region outside of SVSM use.
+///
+/// # Arguments
+///
+/// * `src`: The physical address designating the start of continguous physical
+///   memory to read from.
+///
+/// # Returns
+///
+/// This function returns a `Result` that indicates the success or failure of the operation.
+/// On success, returns an instance of T.
+/// If the physical address region cannot be mapped, it returns `Err(SvsmError::Mem)`.
+/// If the physical address region cannot be read, it returns `Err(SvsmError::Fault)`.
+/// If the physical address region is not allocated to the guest, it returns
+///   `Err(SvsmError::InvalidAddress)`.
+pub fn read_from_guest<T: KnownLayout + Sized>(src: PhysAddr) -> Result<T, SvsmError> {
+    let mut t: MaybeUninit<T> = MaybeUninit::uninit();
+    // SAFETY: copy_from_guest does not read `t`, so it's safe to take a mutable pointer.
+    // The `t` layout is known, so populating through the casted *mut u8 is safe.
+    // The size of T is allocated on the SVSM stack, so it is fully owned.
+    // copy_from_guest populates the full contents of t so it is same to assume t is initialized.
+    unsafe {
+        copy_from_guest(src, t.as_mut_ptr().cast::<u8>(), size_of::<T>())?;
+        Ok(t.assume_init())
+    }
+}
+
+/// Writes a value to a physical address region outside of SVSM use.
+///
+/// # Arguments
+///
+/// * `v`: The value to write to guest memory
+/// * `dst`: The physical address designating the start of continguous physical
+///   memory to write to.
+///
+/// # Returns
+///
+/// This function returns a `Result` that indicates the success or failure of the operation.
+/// If the physical address region cannot be mapped, it returns `Err(SvsmError::Mem)`.
+/// If the physical address region cannot be read, it returns `Err(SvsmError::Fault)`.
+/// If the physical address region is not allocated to the guest, it returns
+///   `Err(SvsmError::InvalidAddress)`.
+#[inline]
+pub fn write_to_guest<T: IntoBytes + Immutable>(v: &T, dst: PhysAddr) -> Result<(), SvsmError> {
+    copy_slice_to_guest(v.as_bytes(), dst)
 }
 
 #[cfg(test)]
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -6,11 +6,14 @@
 
 extern crate alloc;
 
-use crate::address::{Address, VirtAddr};
+use crate::address::{Address, PhysAddr, VirtAddr};
 use crate::cpu::x86::smap::{clac, stac};
 use crate::error::SvsmError;
 use crate::insn_decode::{InsnError, InsnMachineMem};
-use crate::mm::{USER_MEM_END, USER_MEM_START};
+use crate::mm::{
+    memory::valid_phys_region, ptguards::PerCPUPageMappingGuard, USER_MEM_END, USER_MEM_START,
+};
+use crate::utils::MemoryRegion;
 use alloc::string::String;
 use alloc::vec::Vec;
 use core::arch::asm;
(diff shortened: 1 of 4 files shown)


kernel/src/mm/ptguards.rs
```diff
@@ -198,6 +202,39 @@ impl<T: Copy> MemMappingGuard<T> {
             })
     }
 
+    /// Reads a vector of data from a virtual address region specified by an offset and count
+    ///
+    /// # Safety
+    ///
+    /// The caller must verify not to read from arbitrary memory regions. The region to read must
+    /// be checked to guarantee the memory is mapped by the guard and is valid for reading.
+    ///
```
maurer, line 211: You also need to require that the bit pattern present is valid for `T`, and that the original offset used to construct the MemMappingGuard matches alignment restrictions for `T` (not the argument to this function, your base value)

kernel/src/mm/ptguards.rs
```diff
@@ -198,6 +202,39 @@ impl<T: Copy> MemMappingGuard<T> {
...
+            .checked_mul(offset)
+            .ok_or(SvsmError::InvalidAddress)?;
+        let size = core::mem::size_of::<T>()
+            .checked_mul(count)
+            .ok_or(SvsmError::InvalidAddress)?;
+        self.virt_addr_region(start, size)
+            .map_or(Err(SvsmError::Mem), |region|
+                    // SAFETY: Assured by caller.
+                    unsafe {
+                        Ok(from_raw_parts(region.start().as_ptr::<T>(), count).to_vec())
```
maurer, line 234: Since this is a VMM, this is likely roughly equivalent to `copy_from_user` in a traditional kernel. What mechanisms are being deployed to ensure the supervised VM isn't editing the underlying memory during this procedure? If it can still edit the memory, then the fact that you (temporarily) have a `&[T]` pointing at the mutating memory triggers UB.

Separately, if it can still edit the memory, and `T` is only valid for some bit patterns (e.g. enums, bools, references), then the caller would never be able to assert that the memory would still have legal bit patterns in it.

Your options here are:
1. Have the caller assert that the memory being read has been locked for reading somehow, and the VM can't mutate it.
2. Require that `T` be a type which supports *all* bit patterns (`zerocopy` has some good abstractions for this) and build up the vector via `ptr::read` or similar rather than ever producing a reference into the VM space.

kernel/src/mm/ptguards.rs
```diff
@@ -221,13 +258,49 @@ impl<T: Copy> MemMappingGuard<T> {
...
+        let size = core::mem::size_of::<T>();
+        let start = size.checked_mul(offset).ok_or(SvsmError::InvalidAddress)?;
+        let len = size
+            .checked_mul(data.len())
+            .ok_or(SvsmError::InvalidAddress)?;
+        self.virt_addr_region(start, len)
+            .map_or(Err(SvsmError::Mem), |region| {
+                // SAFETY: Assured by caller.
+                unsafe {
+                    from_raw_parts_mut(region.start().as_mut_ptr::<T>(), len).copy_from_slice(data);
```
maurer, line 298: You produce a `&mut [T]`, which means that you are guaranteeing you have unique access, and that the memory region has valid bit patterns for T, triggering UB if that is not the case.

I'd suggest using `std::ptr::copy` or `std::ptr::copy_non_overlapping` instead to avoid materializing a reference.

kernel/src/mm/ptguards.rs
```diff
@@ -243,16 +316,149 @@ impl<T: Copy> MemMappingGuard<T> {
...
+    /// * `start` - The starting address for a `PerCPUPageMappingGuard` to associate with the
+    ///             `MemMappingGuard`.
+    /// * `count` - The number of contiguous T objects to map.
+    ///
+    /// # Returns
+    ///
+    /// Self is returned.
+    pub fn new(start: PhysAddr, count: usize) -> Result<Self, SvsmError> {
+        let size = count.checked_mul(size_of::<T>()).ok_or(SvsmError::Mem)?;
+        let region = MemoryRegion::checked_new(start, size).ok_or(SvsmError::Mem)?;
```
maurer, line 367: Does this check that `start` is maximally aligned? If not, you still need to check that `start` is at least `align_of::<T>` aligned.

kernel/src/mm/ptguards.rs
```diff
@@ -243,16 +316,149 @@ impl<T: Copy> MemMappingGuard<T> {
...
+/// Represents a guard for a specific memory range mapping, which will
+/// unmap the specific memory range after being dropped.
+/// Does all necessary validity checks that the memory region is valid in order for
+/// reads and writes to be safe.
+#[derive(Debug)]
+pub struct SafeMemMappingGuard<T> {
+    guard: MemMappingGuard<T>,
+}
+
+impl<T: Copy> SafeMemMappingGuard<T> {
```
maurer, line 328: Either this needs a guard that `T` takes all bit patterns (`zerocopy::FromBytes` would do the trick), or your `new` function will need to be `unsafe` to have the caller assert that they know the contents match bit pattern restrictions.

kernel/src/mm/ptguards.rs
deeglaze: Just like in linux, there's nothing to stop mid-copy mutation from the userspace. We know the kernel doesn't do this, but there's no guarantee. I'm not really sure what can be done about that.

maurer: The answer then is option 2 - `T: zerocopy::FromBytes + Copy` and do your reads through pointers rather than through slices.

kernel/src/mm/ptguards.rs
deeglaze: I've changed the type-generic function read_from_guest to allocate bytes first and use read_from_bytes.

kernel/src/mm/ptguards.rs
deeglaze: Changed, thanks.

kernel/src/mm/ptguards.rs
deeglaze: Changed, thanks.

kernel/src/mm/ptguards.rs
deeglaze: Removed some of the type generics to simplify into a copy_{to, from}_guest pair of primitives.

kernel/src/mm/guestmem.rs
```diff
@@ -428,6 +431,136 @@ pub fn copy_to_user(src: &[u8], dst: VirtAddr) -> Result<(), SvsmError> {
...
+/// This function returns a `Result` that indicates the success or failure of the operation.
+/// On success, returns a vector of length `size`.
+/// If the physical address region cannot be mapped, it returns `Err(SvsmError::Mem)`.
+/// If the physical address region cannot be read, it returns `Err(SvsmError::Fault)`.
+pub fn read_bytes_from_guest(src: PhysAddr, size: usize) -> Result<Vec<u8>, SvsmError> {
+    let mut result = Vec::with_capacity(size);
+    // SAFETY: The vector's capacity, `size` has been populated by copy_from_guest.
+    // The translation of bytes to T are checked.
+    unsafe {
+        copy_from_guest(src, result.as_mut_slice())?;
```
maurer, line 519: Unfortunately, `result.as_mut_slice()` materialized a `&mut [u8]` where the value of all the elements is `uninit`, which isn't legal. Either use an explicitly zeroed vector, or use `Vec<MaybeUninit<u8>>` and `.assume_init` when you're done, or use pointers.

I think everything else looks fine.

deeglaze: Fixed. PTAL.

kernel/src/mm/guestmem.rs
```diff
@@ -428,6 +431,73 @@ pub fn copy_to_user(src: &[u8], dst: VirtAddr) -> Result<(), SvsmError> {
...
+///   memory to read from.
+/// * `dst`: The pointer to the beginning of the guest memory range to populate.
+/// * `size`: The number of bytes to write from src to dst.
+///
+/// # Returns
+///
+/// This function returns a `Result` that indicates the success or failure of the operation.
+/// If the physical address region cannot be mapped, it returns `Err(SvsmError::Mem)`.
+/// If the physical address region cannot be read, it returns `Err(SvsmError::Fault)`.
+pub fn copy_from_guest(src: PhysAddr, dst: *mut u8, size: usize) -> Result<(), SvsmError> {
```
stefano-garzarella, line 456: Let's move the discussion started at https://github.com/coconut-svsm/svsm/pull/662#discussion_r2035876831 here.

I think this should be marked `unsafe` since the caller must ensure that `dst` is sufficiently big (`size`).
Otherwise, we can use a slice (like in `copy_to_guest`) to have a similar interface. If the caller has no slice and is concerned about performance, it can always call `std::slice::from_raw_parts_mut(result.as_mut_ptr(), size)`

kernel/src/mm/guestmem.rs
```diff
@@ -498,6 +498,70 @@ pub fn copy_to_guest(src: &[u8], dst: PhysAddr) -> Result<(), SvsmError> {
...
+/// # Returns
+///
+/// This function returns a `Result` that indicates the success or failure of the operation.
+/// On success, returns an instance of T.
+/// If the physical address region cannot be mapped, it returns `Err(SvsmError::Mem)`.
+/// If the physical address region cannot be read, it returns `Err(SvsmError::Fault)`.
+pub fn read_from_guest<T: FromBytes + Sized>(src: PhysAddr) -> Result<T, SvsmError> {
+    // Should be stack allocatable, but using size_of in array bounds is not possible
+    // in current Rust. The zero, copy, read sequence could be better optimized too.
+    let mut result_bytes = vec![0u8; size_of::<T>()];
```
stefano-garzarella, line 542: Why do we initialize it to 0 here, while in `read_bytes_from_guest` we don't?

kernel/src/mm/guestmem.rs
```diff
@@ -498,6 +498,70 @@ pub fn copy_to_guest(src: &[u8], dst: PhysAddr) -> Result<(), SvsmError> {
...
+    copy_from_guest(src, result_bytes.as_mut_ptr(), size_of::<T>())?;
+    T::read_from_bytes(&result_bytes).map_err(|_| SvsmError::InvalidBytes)
+}
+
+/// Writes a value to a physical address region outside of SVSM use.
+///
+/// # Arguments
+///
+/// * `v`: The value to write to guest memory
+/// * `src`: The physical address designating the start of continguous physical
```
stefano-garzarella, line 552: this should be `dst`

kernel/src/mm/guestmem.rs
```diff
@@ -498,6 +498,70 @@ pub fn copy_to_guest(src: &[u8], dst: PhysAddr) -> Result<(), SvsmError> {
...
+///
+/// * `src`: The physical address designating the start of continguous physical
+///   memory to read from.
+///
+/// # Returns
+///
+/// This function returns a `Result` that indicates the success or failure of the operation.
+/// On success, returns an instance of T.
+/// If the physical address region cannot be mapped, it returns `Err(SvsmError::Mem)`.
+/// If the physical address region cannot be read, it returns `Err(SvsmError::Fault)`.
```
stefano-garzarella, line 538: Should we mention also `SvsmError::InvalidBytes` ?

kernel/src/mm/guestmem.rs
deeglaze: I used slices before and @maurer noted the undefined behavior that induces. The use of a pointer is deliberate.

kernel/src/mm/guestmem.rs
deeglaze: Good point. I should use read_bytes_from_guest directly and then T::read_from_bytes to avoid the extra linear scan. I couldn't find a suitable means to have a single pass T creation due to the alignment requirements and strange additional trait bound for IntoBytes.

kernel/src/mm/guestmem.rs
deeglaze: Removed the need.

kernel/src/mm/guestmem.rs
stefano-garzarella: Okay, but in that case this should be `unsafe` be because it must be the caller's responsibility to make sure that `size` is correct.

kernel/src/mm/guestmem.rs
deeglaze: Fixed.

kernel/src/mm/guestmem.rs
deeglaze: I've renamed the safe functions to copy_slice_{to,from}_guest and changed copy_from_guest to an unsafe function. Thanks!

PR #654, opened 2025-03-22
tdp: do not load EFER during AP startup

msft-jlange:
The TDX module does not permit EFER to be reloaded when running in an L1, so skip the reload of EFER during the AP startup flow when running under TDX.

### DIFF
--- a/kernel/src/cpu/smp.rs
+++ b/kernel/src/cpu/smp.rs
@@ -6,6 +6,7 @@
 
 use crate::acpi::tables::ACPICPUInfo;
 use crate::address::{Address, VirtAddr};
+use crate::cpu::efer::EFERFlags;
 use crate::cpu::idt::idt;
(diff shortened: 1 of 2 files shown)


kernel/src/cpu/smp.rs
```diff
@@ -92,8 +93,14 @@ global_asm!(
          * environment and context structure from the address space. */
         movq    %r8, %cr0
         movq    %r10, %cr4
+
+        /* Check to see whether EFER.LME is specified.  If not, then EFER
+         * should not be reloaded. */
+        testl   ${LME}, %ecx
```
peterfang, line 99: `%eax`?

peterfang (commented):
Comment posted

kernel/src/cpu/smp.rs
msft-jlange: Oops - good catch.

peterfang:
LGTM now. Thanks for the fix. We're not actively changing bits in EFER when jumping from stage2 to SVSM so I neglected this.

PR #655, opened 2025-03-25
Make IDT and GDT read-only after init

p4zuu:
Implements a way to set pages read-only at runtime to allow making IDT and global GDT read-only after they're initialized.

**I don't have access to a CoCo platform, I would need someone to test it, test-in-svsm included. Thanks!**


osteffenrh:
This does not boot for me:
```
$ ./scripts/launch_guest.sh --qemu ../build/bin/qemu-system-x86_64 --igvm ./bin/coconut-qemu.igvm --image ../f41.qcow2
=============================
Launching SVSM guest
=============================
QEMU:         ../build/bin/qemu-system-x86_64
QEMU Version: 9.1.50
IGVM:         ./bin/coconut-qemu.igvm
IMAGE:        ../f41.qcow2
=============================
Press Ctrl-] to interrupt
=============================
[Stage2] COCONUT Secure Virtual Machine Service Module Version 4c6b3af5
[Stage2] SVSM memory region: [0x0000008000000000-0x0000008001000000]
[Stage2] Order-02: total pages:     0 free pages:     0
[Stage2] Order-03: total pages:     0 free pages:     0
[Stage2] Order-04: total pages:     2 free pages:     2
[Stage2] Order-05: total pages:     3 free pages:     3
[Stage2] Total memory: 572KiB free memory: 516KiB
[Stage2]   kernel_virtual_base      = 0xffffff8000000000
[SVSM] COCONUT Secure Virtual Machine Service Module Version 4c6b3af5
[SVSM] Order-00: total pages:    47 free pages:     0
[SVSM] Order-01: total pages:     0 free pages:     0
[SVSM] Order-02: total pages:     1 free pages:     1
[SVSM] Order-03: total pages:     1 free pages:     1
[SVSM] Order-04: total pages:     1 free pages:     1
[SVSM] Order-05: total pages:   100 free pages:   100
[SVSM] Total memory: 13100KiB free memory: 12912KiB
[SVSM] Boot stack @ [0xffffff8000302000-0xffffff800030a000]
[SVSM] Guest Memory Regions:
[SVSM]   [0x0000000000000000-0x0000000080000000]
[SVSM]   [0x0000000100000000-0x0000000280000000]
[SVSM] Ignoring MADT entry with type 1
[SVSM] Ignoring MADT entry with type 2
[SVSM] Ignoring MADT entry with type 2
[SVSM] Ignoring MADT entry with type 2
[SVSM] Ignoring MADT entry with type 2
[SVSM] Ignoring MADT entry with type 2
[SVSM] Ignoring MADT entry with type 4
[SVSM] 4 CPU(s) present
[SVSM] Launching AP with APIC-ID 1
[SVSM] AP with APIC-ID 1 is online
[SVSM] Launching AP with APIC-ID 2
[SVSM] Launching request-processing task on CPU 1
[SVSM] Launching request loop task on CPU 1
[SVSM] AP with APIC-ID 2 is online
[SVSM] Launching AP with APIC-ID 3
[SVSM] Launching request-processing task on CPU 2
[SVSM] AP with APIC-ID 3 is online
[SVSM] Launching request loop task on CPU 2
[SVSM] Brought 3 AP(s) online
[SVSM] Launching request-processing task on CPU 3
[SVS% 
```
Is there something I can check to give some more info?

msft-jlange:
I gave this a try, and the test-in-svsm build runs correctly for me under Hyper-V on SNP, so I don't yet know where the problem might be.

p4zuu:
> I gave this a try, and the test-in-svsm build runs correctly for me under Hyper-V on SNP, so I don't yet know where the problem might be.

Thanks for testing @msft-jlange. Did you also face a SVSM kernel boot hang like @osteffenrh when trying to boot a guest?

osteffenrh:
It is not hanging, Qemu simply exits without any message. Exit code is 0.

kernel/src/mm/pagetable.rs
```diff
@@ -1186,6 +1186,21 @@ impl PageTable {
             entry.set(make_private_address(paddr), flags);
         }
     }
+
+    /// Makes the memory region pages read-only.
+    ///
+    /// # Safety
+    ///
+    /// The caller should verify that `region` can be made read-only, i.e. that
+    /// no write can happen or that a #PF raised by any write temptative is
```
peterfang, line 1195: write temptative -> tentative write?

kernel/src/mm/pagetable.rs
```diff
@@ -1186,6 +1186,21 @@ impl PageTable {
...
+    /// expected.
+    pub unsafe fn make_region_ro(&mut self, region: MemoryRegion<VirtAddr>) {
+        for page in region.iter_pages(PageSize::Regular) {
+            if let Mapping::Level0(entry) = self.walk_addr(page) {
+                entry.set(entry.address(), PTEntryFlags::data_ro());
+            }
```
peterfang, line 1201: What about huge pages? I believe it's possible to encounter 2M pages here.

kernel/src/mm/ro_after_init.rs
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
peterfang, line 1: Can you add `Copyright` and `Author:` to the header here?

kernel/src/mm/pagetable.rs
```diff
@@ -1186,6 +1186,21 @@ impl PageTable {
...
+    ///
+    /// # Safety
+    ///
+    /// The caller should verify that `region` can be made read-only, i.e. that
+    /// no write can happen or that a #PF raised by any write temptative is
+    /// expected.
+    pub unsafe fn make_region_ro(&mut self, region: MemoryRegion<VirtAddr>) {
+        for page in region.iter_pages(PageSize::Regular) {
+            if let Mapping::Level0(entry) = self.walk_addr(page) {
+                entry.set(entry.address(), PTEntryFlags::data_ro());
```
peterfang, line 1200: `PTEntryFlags` has other constructors such as `task_data_ro()`, albeit it's not being used anywhere. Is it better to just remove `PTEntryFlags::WRITABLE` & `PTEntryFlags::DIRTY` here?

kernel/src/mm/ro_after_init.rs
```diff
@@ -0,0 +1,85 @@
...
+    };
+}
+
+extern "C" {
+    pub static ro_after_init_start: u8;
+    pub static ro_after_init_end: u8;
+}
+
+#[derive(Debug)]
+pub struct ROAfterInit(MemoryRegion<VirtAddr>);
```
peterfang, line 22: This looks like a specialized struct that deals with specific memory ranges that are shared by all CPUs (since it does a global TLB shootdown in the end). Does it need to be public right now? Are there other use cases that come to mind?

peterfang (commented):
It boots on my TDX platform after rebasing to main. Will probably need someone to help retest this on SNP before merging to make sure nothing breaks.

Aside from the comments above, one observation is that the #PF handler will probably silently ignore these faults and cause the faulting CPU to be stuck without giving any useful information. This however can be improved later.

kernel/src/cpu/gdt.rs
```diff
@@ -54,6 +55,10 @@ impl GDTEntry {
 
 const GDT_SIZE: u16 = 8;
 
+// SAFETY: after this section's pages are made read-only, a write to this section
+// is a memory saftey violation triggering a #PF, which is the intended
```
msft-jlange, line 59: Typo
```suggestion
// is a memory safety violation triggering a #PF, which is the intended
```

kernel/src/cpu/idt/common.rs
```diff
@@ -394,6 +395,10 @@ impl ReadLockGuard<'static, IDT> {
     }
 }
 
+// SAFETY: after this section's pages are made read-only, a write to this section
+// is a memory saftey violation triggering a #PF, which is the intended
```
msft-jlange, line 399: Typo
```suggestion
// is a memory safety violation triggering a #PF, which is the intended
```

kernel/src/mm/pagetable.rs
msft-jlange: This routine really only works if it is called on a region that is backed by 4 KB pages - in fact, it only works if it is called on a region that is aligned (start and size) to the boundaries of the page table entries mapping the region.  Since it is declared `unsafe`, it is reasonable to assume that the caller has verified whatever is required, though this really should be explicitly stated in the comments.  It wouldn't be a bad idea to assert that the entries are 4 KB entries, but as an unsafe function, it's not strictly required (so maybe a debug assert would be best here).

p4zuu: What @msft-jlange wrote was actually my assumptions too, but that's indeed not clear in the code. 
What about the following to make it clearer:
- rename the method to `make_region_4k()`
- explicit in the comment that this is only meant for 4k pages
- add in the unsafe requirements that the caller must check that the region start and size must be 4k aligned
- add debug assert checking `entry.huge()` is false

kernel/src/mm/ro_after_init.rs
p4zuu: I added `// Copyright (c) Coconut-SVSM authors`  as seen somewhere else in the code. I'm writing this on my free time now

p4zuu: Done :)

kernel/src/mm/ro_after_init.rs
p4zuu: Hum, not really atm. We still need one public method/function that gets called during kernel setup.
Since `make_ro_after_init()` is currently the only use and it's used only for a given kernel section, I guess we can make the struct private and leave the `make_ro_after_init()` public. Would it be better?
Just noticing, it's not clear that `make_ro_after_init()` operates on one specific section, I'll add doc and maybe rename the function

peterfang: > Hum, not really atm. We still need one public method/function that gets called during kernel setup. Since `make_ro_after_init()` is currently the only use and it's used only for a given kernel section, I guess we can make the struct private and leave the `make_ro_after_init()` public. Would it be better? 

That sounds great to me. Thanks!

kernel/src/mm/pagetable.rs
p4zuu: I'm a bit puzzled about leaving the eventuality of having `.ro_after_init` section mapped in a 2M page. On the other hand, it's probably overkill to hack into `init_page_table()` to ensure the section is 4K mapped. We can probably take the risk if this is very unlikely, but I'm open to further verifications

p4zuu:
> Having the section ending on page boundary was intended, thanks for catching this! I thought I verified that but apparently it was not done properly. I'll apply your suggestion

@osteffenrh maybe we should keep ` = ALIGN(4096);` before `.bss` as well to ensure bss is properly aligned. By placing the align directive inside the `.ro_after_init` section, I think we ensure ro_after_init size is paged-aligned, but if we add another unaligned section between ro_after_init and bss, that might screw bss page-alignment

kernel/src/mm/ro_after_init.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        // SAFETY: writing to ro_after_init_start is supposed to fail preventing memory safety break.
+        let res_w1 = unsafe {
+            GuestPtr::<u8>::new(VirtAddr::from(&raw const ro_after_init_start)).write(0x41)
+        };
+        assert!(res_w1.is_err());
+
+        // SAFETY: writing to ro_after_init_end is supposed to fail preventing memory safety break.
+        let res_w2 = unsafe {
+            GuestPtr::<u8>::new(VirtAddr::from(&raw const ro_after_init_end)).write(0x41)
```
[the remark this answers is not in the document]
msft-jlange: > Just noticing that allowing to write to a `GuestPtr` created from a `&raw const` looks like a soundness issue in GuestPtr

The `GuestPtr` isn't being created from the reference.  It's being created from `VirtAddr`.  The `VirtAddr` is being created from a reference.  There are no soundness issues with either link in this chain: `VirtAddr` just represents a virtual address and implies nothing about mutability, and `GuestPtr` assumes that the virtual address in `VirtAddr` is useful (this assumption cannot be verified, which is why the `GuestPtr` methods are `unsafe`).  Because this is implemented as an unsafe operation, it is up to the caller (here) to ensure the necessary soundness properties, because the compiler cannot do anything on its own.

kernel/src/mm/pagetable.rs
```diff
@@ -1186,6 +1191,27 @@ impl PageTable {
...
+    /// expected.
+    /// The caller must also ensure that the region start and size are 4k
+    /// aligned.
+    pub unsafe fn make_region_ro_4k(&mut self, region: MemoryRegion<VirtAddr>) {
+        for page in region.iter_pages(PageSize::Regular) {
+            if let Mapping::Level0(entry) = self.walk_addr(page) {
+                debug_assert!(entry.global());
+                debug_assert!(!entry.huge());
+
+                entry.set(entry.address(), PTEntryFlags::data_ro());
```
joergroedel: This clears the `C` bit under SEV-SNP, which explains the crashed people have seen (including me).

Please change this method so that it preserves the shared/private mapping attributes when making a page read-only. Please also handle the case where the region is mapped with 2M before.

kernel/src/cpu/idt/common.rs
```diff
@@ -394,6 +395,10 @@ impl ReadLockGuard<'static, IDT> {
     }
 }
 
+// SAFETY: after this section's pages are made read-only, a write to this section
+// is a memory safety violation triggering a #PF, which is the intended
+// behavior.
+#[unsafe(link_section = ro_after_init_section!())]
 static IDT: RWLock<IDT> = RWLock::new(IDT::new());
```
joergroedel: This also makes the `RWLock` read-only, preventing any read-access to the IDT. It needs to be split up so that only the IDT itself, and not the lock, are on the read-only page.

msft-jlange: > This also makes the `RWLock` read-only, preventing any read-access to the IDT. It needs to be split up so that only the IDT itself, and not the lock, are on the read-only page.

This is an excellent, point, and is problematic.  The design of `RWLock` requires the lock state and the data protected by the lock to live in the same allocation, which means that either both must be read-only or neither can be read-only.  We really don't want an implementation where attempting to acquire the lock for shared access causes a system crash.

I believe that fixing this will require a redesign of how the IDT is managed.  I can take a crack at that in a separate PR, and it would be natural to make the IDT read-only as part of that PR.  Once that's done, this PR would be great for making the GDT read-only.

joergroedel: If we want to keep the `RWLock` the solution could be putting the lock around a `Box` which is constructed from a raw pointer to the read-only IDT in memory.

msft-jlange: > If we want to keep the `RWLock` the solution could be putting the lock around a `Box` which is constructed from a raw pointer to the read-only IDT in memory.

I do not believe that is possible.  `Box` can only represent data in the heap (as allocated by the global allocator) because `Box::drop()` will always call the global allocator's `free` function.  We definitely don't want portions of the global heap to be marked read-only.

I have another plan in mind which I believe will work well, which involves splitting up IDT initialization to avoid the need for any lock at all.  Let me see if I can make that work before we look for unconventional alternatives.

p4zuu:
> Hi Thomas, thanks for your work on this. I tracked the boot crash on SNP down and left two comments for things that need to be changed.

Thanks a lot Joerg! Let me work on that

kernel/src/mm/pagetable.rs
p4zuu: For 2M pages, we discussed [here](https://github.com/coconut-svsm/svsm/pull/655#discussion_r2099826657) that since this function is unsafe, we should be ok as long as it's proper advertised that this function is meant for 4k pages. Would you prefer properly handling the 2M page case instead of a `debug_assert!()`?

joergroedel: The thing is that the `debug_assert!(!entry.huge())` will never trigger because it happens in the `Mapping::Level0` path already. This `if` statement needs an else with an assert. I think the code also needs to check whether the PTE is marked present.

p4zuu: @joergroedel just added changes. I now check the region start address alignment to iterate over the pages with regular or huge page size, and then handle `Mapping::Level1` mapping the same way we handle `Mapping::Level0` but setting the `HUGE` PTE flag in addition. Hope it's ok

kernel/src/cpu/idt/common.rs
p4zuu: Changed this after rebase on #726. It now sets the `GLOBAL_IDT` in the ro_after_init region. It means that the `ImmutAfterInitCell` in which the static IDT is contained will be set ro as well, but since we don't expect another `.init()` on this cell, we should be fine. Can you please confirm @msft-jlange?

kernel/src/mm/pagetable.rs
p4zuu: Truly sorry @peterfang I completely forgot to reply to this.
As said [in this comment](https://github.com/coconut-svsm/svsm/pull/655#discussion_r2172100885), it now checks the region start address to handle the huge page case. Can you please review this again?

kernel/src/cpu/idt/common.rs
msft-jlange: > It means that the `ImmutAfterInitCell` in which the static IDT is contained will be set ro as well, but since we don't expect another `.init()` on this cell, we should be fine. Can you please confirm @msft-jlange?

It should be fine, but there's really no point in protecting `GLOBAL_IDT`.  That variable only holds a pointer to the actual IDT, so corrupting it won't affect IDT dispatch at all.  It would be better to avoid protecting something that has no security function.

kernel/src/mm/pagetable.rs
```diff
@@ -119,6 +119,11 @@ fn make_private_address(paddr: PhysAddr) -> PhysAddr {
     PhysAddr::from(paddr.bits() & !shared_pte_mask() | private_pte_mask())
 }
 
+// Returns true if the address is shared.
+fn is_shared(paddr: PhysAddr) -> bool {
+    (paddr.bits() & shared_pte_mask()) != 0
```
peterfang: This is incomplete since `private_pte_mask()` also needs to be considered. Maybe something like this to be more failproof:

```suggestion
    paddr == make_shared_address(strip_confidentiality_bits(paddr))
```

kernel/src/mm/pagetable.rs
```diff
@@ -1186,6 +1196,53 @@ impl PageTable {
...
+        region: MemoryRegion<VirtAddr>,
+    ) -> Result<(), SvsmError> {
+        let page_size = if region.start().is_aligned(PageSize::Huge as usize) {
+            PageSize::Huge
+        } else {
+            debug_assert!(region.start().is_aligned(PageSize::Regular as usize));
+            PageSize::Regular
+        };
+
+        for page in region.iter_pages(page_size) {
```
peterfang: Since this function now supports both 4K and 2M pages, `page_size` is determined by each individual page. So using `iter_pages()` doesn't cover all the cases anymore.

kernel/src/mm/pagetable.rs
```diff
@@ -1186,6 +1196,53 @@ impl PageTable {
...
+            match self.walk_addr(page) {
+                Mapping::Level0(entry) | Mapping::Level1(entry) => {
+                    if !entry.present() || !entry.global() {
+                        return Err(SvsmError::Mem);
+                    }
+
+                    let mut flags = if is_shared(entry.address()) {
+                        PTEntryFlags::data_ro().bits() as usize | shared_pte_mask()
```
peterfang: Please use `make_shared_address()` instead to cover all platforms.

kernel/src/mm/pagetable.rs
```diff
@@ -1186,6 +1196,53 @@ impl PageTable {
...
+                    } else {
+                        PTEntryFlags::data_ro().bits() as usize | private_pte_mask()
```
peterfang: Please use `make_private_address()` instead to cover all platforms.

kernel/src/mm/pagetable.rs
peterfang: Thanks for making the changes. I just left my comments.

kernel/src/cpu/idt/common.rs
p4zuu: Yes you're right. I did not pay attention enough and missed that the cell only holds a pointer to the IDT. I'll change this

kernel/src/mm/pagetable.rs
p4zuu: Thanks for catching this. Your suggestion looks to me so I applied it. @joergroedel can you please double-check this is correct?
Also, I noticed that I called this `is_shared()` helper with `entry.address()` as argument. However, the `address()` method also strips the C-bit (and other PTE flags). This was wrong I think, it should be better to call this helper with the raw PTE value which contains the C-bit. It means that the `is_shared()` helper also compares the PTE flags. I suppose it's ok, but I can create another helper function that strips all flags except the C-bit

kernel/src/mm/pagetable.rs
p4zuu: Thanks for catching this, it's changed now

kernel/src/mm/pagetable.rs
p4zuu: Thanks for catching this, it's changed now

kernel/src/mm/pagetable.rs
p4zuu: Sorry I don't get what you mean, do you mean that a multiple-page region can be backed by both 4K and 2M pages, so `iter_pages()` needs to verify the page size at each iteration?

kernel/src/mm/pagetable.rs
peterfang: I'm okay either way since the PTE flags aren't used to indicate a shared/private address

peterfang: Glad you caught it! Comparing the PTE flags shouldn't be a problem since they're not used to indicate a shared/private address.

kernel/src/mm/pagetable.rs
peterfang: Yes that's what I meant. The memory region passed into `make_region_ro()` could contain both 4K and 2M pages (e.g. a 2M-aligned starting address with size 2M+4K) so iterating over a fixed page size can be problematic.

p4zuu:
> How about just creating a dedicated data section for it? We can build infrastructure for RO-after-init heap space later if there's a need for it.

Do you mean a new section for the global page heap? I guess we could, but I'm more concerned about having a read-only portion of global memory (like described [here](https://github.com/coconut-svsm/svsm/pull/655#discussion_r2155088919))

peterfang:
> Do you mean a new section for the global page heap? I guess we could, but I'm more concerned about having a read-only portion of global memory (like described [here](https://github.com/coconut-svsm/svsm/pull/655#discussion_r2155088919))

Actually I was thinking of keeping it separate from the global heap. Instead of putting the IDT page in a generic section like `.ro_after_init`, maybe we can put it in its own IDT section and make that section read-only as well. The `ImmutAfterInitCell` struct holding the IDT page would still remain in `.ro_after_init`. Like what was said previously, we probably don't want to make a part of the heap area read-only.

msft-jlange:
> Actually I was thinking of keeping it separate from the global heap. Instead of allocating the IDT page from a generic area like the heap, maybe we can put it in its own IDT section and make that section read-only. The IDT section could possibly be a subsection of `.ro_after_init`. The `ImmutAfterInitCell` struct holding the IDT page would also remain in `.ro_after_init`. Like what was said previously, we probably don't want to make a part of the heap area read-only.

No, it's better to avoid keeping this out of the span of the ELF image.  Allocating it from the page heap is the right thing to do.  There is no concern about changing the permissions of page-sized allocations because they are not heap objects and have no header/footer data.  The page heap allocates individual pages as full pages, and each of those is entitled to apply its own unique page permissions.

kernel/src/mm/pagetable.rs
p4zuu: That makes sense. I switched back to handling 4k pages only as before, that's easier

p4zuu:
> > Actually I was thinking of keeping it separate from the global heap. Instead of allocating the IDT page from a generic area like the heap, maybe we can put it in its own IDT section and make that section read-only. The IDT section could possibly be a subsection of `.ro_after_init`. The `ImmutAfterInitCell` struct holding the IDT page would also remain in `.ro_after_init`. Like what was said previously, we probably don't want to make a part of the heap area read-only.
> 
> No, it's better to avoid keeping this out of the span of the ELF image. Allocating it from the page heap is the right thing to do. There is no concern about changing the permissions of page-sized allocations because they are not heap objects and have no header/footer data. The page heap allocates individual pages as full pages, and each of those is entitled to apply its own unique page permissions.

Thanks for your input folks. I moved to make the IDT entries `PageBox` read-only

kernel/src/mm/ro_after_init.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub static ro_after_init_start: u8;
+    pub static ro_after_init_end: u8;
+}
+
+#[derive(Debug)]
+pub struct ROAfterInit(MemoryRegion<VirtAddr>);
+
+impl ROAfterInit {
+    pub fn new(start: VirtAddr, end: VirtAddr) -> Self {
+        Self(MemoryRegion::new(start, end - start))
```
peterfang: Nit: Using `MemoryRegion::from_addresses()` will give you the added benefit of checking whether end >= start.

p4zuu: Good point, I just changed it. Thanks!

peterfang (approved):

peterfang:
LGTM, @msft-jlange @joergroedel can you take another look? Thanks.

kernel/src/mm/ro_after_init.rs
```diff
@@ -0,0 +1,94 @@
...
+    };
+}
+
+extern "C" {
+    pub static ro_after_init_start: u8;
+    pub static ro_after_init_end: u8;
+}
+
+#[derive(Debug)]
+pub struct ROAfterInit(MemoryRegion<VirtAddr>);
```
msft-jlange: I don't understand the point of the `ROAfterInit` structure at all.  It doesn't appear to encapsulate any unique data, nor any functionality specific to that data.  The `make_ro()` function could simply be declared as
```
pub unsafe fn make_ro(region: MemoryRegion<VirtAddr) -> Result<(), SvsmError>
```
and it would do exactly the same thing without requiring the definition of a new structure.  What makes the use of a specialized structure a better approach?

p4zuu: You're right. This is an early-dev leftover, but I agree it's clearer to simply get rid of the `ROAfterInit` struct. Thanks for noticing
Changed now

msft-jlange (approved):

Issue #646, opened 2025-03-18
Linux boot broken after 2d6a7fc7


osteffenrh:
This commit
```
commit 2d6a7fc75b39ebf1458b06e6cafff7b06fc490af (HEAD)
Author: Jon Lange <jlange@microsoft.com>
Date:   Sun Feb 23 11:51:06 2025 -0800

    igvm: use MADT for determining APIC IDs

    The IGVM format specifies a mechanism for the host to inject an MADT
    into the guest address space.  This MADT should be used to determine the
    set of APIC IDs associated with the set of CPUs that will be started by
    the guest.

    Not all VMMs can support injection of an MADT through IGVM parameters.
    When building for these systems, the IGVM file does not request an MADT,
    and the MADT is obtained instead through QEMU firmware config.

    Signed-off-by: Jon Lange <jlange@microsoft.com>
```
Causes the Linux kernel (6.11, from our local repo) to encounter an error during boot:

serial output:
```
[...]
[    1.947534] VFS: Disk quotas dquot_6.6.0
[    1.948352] VFS: Dquot-cache hash table entries: 512 (order 0, 4096 bytes)
[    1.949987] ACPI Error: AE_BAD_PARAMETER, Thread 33652736 could not acquire Mutex [ACPI_MTX_
Tables] (0x2) (20240827/utmutex-252)
[    1.952156] ACPI Error: Mutex [ACPI_MTX_Tables] (0x2) is not acquired, cannot release (20240
827/utmutex-289)
[    1.954051] pnp: PnP ACPI: disabled
[    2.007201] PM-Timer failed consistency check  (0xff00) - aborting.
[...]
[    2.684374] atomic64_test: passed for x86-64 platform with CX8 and with SSE
[    2.686027] shpchp: Standard Hot Plug PCI Controller Driver version: 0.4
[    2.687491] ------------[ cut here ]------------
[    2.688383] refcount_t: addition on 0; use-after-free.
[    2.689433] WARNING: CPU: 1 PID: 1 at lib/refcount.c:25 refcount_warn_saturate+0xe5/0x110
[    2.690992] Modules linked in:
[    2.691571] CPU: 1 UID: 0 PID: 1 Comm: swapper/0 Not tainted 6.13.6-200.fc41.x86_64 #1
[    2.693037] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS edk2-stable202502-39-gb
483f751 02/02/2022
[...]

PR #658, opened 2025-03-27
config, fw_cfg: Read cpu count from fw_cfg item on Qemu

osteffenrh:
Due to a problem with ACPI tables and Linux when running under Qemu [1],
determine the CPU count via a hidden fw_cfg item instead and don't read
any ACPI tables.

[1] https://github.com/coconut-svsm/svsm/issues/646

### DIFF
--- a/kernel/src/config.rs
+++ b/kernel/src/config.rs
@@ -130,7 +130,17 @@ impl<'a> SvsmConfig<'a> {
             }
         }
 
-        load_fw_cpu_info(self.fw_cfg.as_ref().unwrap())
+        // Fall back to fw_cfg, if present, and read a hidden file to get CPU count
+        if let Some(c) = self.fw_cfg.as_ref() {
+            Ok((0..c.get_cpu_count())
+                .map(|i| ACPICPUInfo {
+                    apic_id: i,
+                    enabled: true,
+                })
+                .collect())
+        } else {
+            load_fw_cpu_info(self.fw_cfg.as_ref().unwrap())
+        }
     }
 
     pub fn should_launch_fw(&self) -> bool {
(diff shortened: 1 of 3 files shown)


kernel/src/config.rs
```diff
@@ -130,7 +130,17 @@ impl<'a> SvsmConfig<'a> {
             }
         }
 
-        load_fw_cpu_info(self.fw_cfg.as_ref().unwrap())
+        // Fall back to fw_cfg, if present, and read a hidden file to get CPU count
+        if let Some(c) = self.fw_cfg.as_ref() {
+            Ok((0..c.get_cpu_count())
+                .map(|i| ACPICPUInfo {
+                    apic_id: i,
+                    enabled: true,
```
msft-jlange, line 138: This doesn't work when APIC ID doesn't match CPU index.  This is why the original change was made in the first place.

msft-jlange:
We need to stop layering new hacks into the code and just make the correct code work properly.  It sounds like there is a PR available to QEMU to generate the MADT correctly so that it can be used both by COCONUT-SVSM and OVMF.  That direction should be preferred over some other QEMU hack that makes a separate APIC ID list available - both are QEMU fixes, so why have two?  And really, the right thing to do is to teach QEMU to inject the IGVM MADT parameter, so if we're going to require a QEMU fix, that's really the one we should be aiming for, because that is the correct long-term solution.

osteffenrh:
I should have mentioned that I am only proposing this as an alternative to https://github.com/coconut-svsm/qemu/pull/19

osteffenrh:
No longer needed. The workaround was merged in our Qemu.

PR #661, opened 2025-03-28
svsm: correctly support TDP on Hyper-V

msft-jlange:
Correct behavior of TDP on Hyper-V requires the SVSM to make Hyper-V hypercalls.  This change includes the infrastructure required to detect a Hyper-V interface when running on TDP (or SNP), the infrastructure required to make hypercalls on TDP (or SNP), and the logic required for the SVSM to operate correctly when running TDP on top of Hyper-V.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -26,7 +26,7 @@ use crate::cpu::vmsa::{svsm_code_segment, svsm_data_segment, svsm_gdt_segment, s
 use crate::cpu::{IrqState, LocalApic};
 use crate::error::{ApicError, SvsmError};
 use crate::hyperv;
-use crate::hyperv::HypercallPagesGuard;
+use crate::hyperv::{HypercallPagesGuard, IS_HYPERV};
 use crate::locking::{LockGuard, RWLock, RWLockIrqSafe, SpinLock};
 use crate::mm::alloc::{allocate_pages, free_page};
 use crate::mm::page_visibility::{make_page_private, make_page_shared};
@@ -667,6 +667,10 @@ impl PerCpu {
         self.current_stack.set(stack);
     }
 
+    pub fn get_cpu_index(&self) -> usize {
+        self.shared().cpu_index()
+    }
+
     pub fn get_apic_id(&self) -> u32 {
         self.shared().apic_id()
     }
@@ -884,6 +888,12 @@ impl PerCpu {
 
         // Complete platform-specific initialization.
         platform.setup_percpu(self)?;
+
+        // Allocate hypercall pages if running on Hyper-V, unless this is the
+        // BSP (where they will be allocated later).
+        if self.shared.cpu_index() != 0 && *IS_HYPERV {
+            self.allocate_hypercall_pages()?;
+        }
 
         Ok(())
     }
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -28,7 +28,8 @@ use crate::error::{ApicError, SvsmError};
 use crate::hyperv;
 use crate::hyperv::HypercallPagesGuard;
 use crate::locking::{LockGuard, RWLock, RWLockIrqSafe, SpinLock};
-use crate::mm::alloc::allocate_pages;
+use crate::mm::alloc::{allocate_pages, free_page};
+use crate::mm::page_visibility::{make_page_private, make_page_shared};
 use crate::mm::pagetable::{PTEntryFlags, PageTable};
 use crate::mm::virtualrange::VirtualRange;
 use crate::mm::vm::{
@@ -584,6 +585,28 @@ impl PerCpu {
     /// Allocates hypercall input/output pages for this CPU.
     pub fn allocate_hypercall_pages(&self) -> Result<(), SvsmError> {
         let vaddr = allocate_pages(2)?;
+
+        // Attempt to make each page shared.  This must be done carefully so
+        // that state is correctly restored in the event of failure.
+        // SAFETY: the virtual addresses are correct because the pages were
+        // just allocated.  If conversion is successful, these pages will never
+        // be freed, and if conversion if unsuccessful, state will correctly be
+        // restored here before the pages are freed.
+        unsafe {
+            let mut r = make_page_shared(vaddr);
+            if r.is_ok() {
+                let r2 = make_page_shared(vaddr + PAGE_SIZE);
+                if r2.is_err() {
+                    make_page_private(vaddr).expect("Failed to restore private page");
+                    r = r2;
+                }
+            }
+            if r.is_err() {
+                free_page(vaddr);
+                return r;
+            }
+        }
+
         let pages = (
             VirtPhysPair::new(vaddr),
             VirtPhysPair::new(vaddr + PAGE_SIZE),
--- a/kernel/src/hyperv/hv.rs
+++ b/kernel/src/hyperv/hv.rs
@@ -18,12 +18,119 @@ use crate::mm::pagetable::PTEntryFlags;
 use crate::mm::{virt_to_phys, SVSM_HYPERCALL_CODE_PAGE};
 use crate::types::PAGE_SIZE;
 use crate::utils::immut_after_init::ImmutAfterInitCell;
+use crate::utils::unsafe_copy_bytes;
 
 use core::arch::asm;
 use core::cell::RefMut;
-use core::slice;
+use core::marker::PhantomData;
+use core::mem;
+use core::mem::MaybeUninit;
+use core::ptr;
 
 use bitfield_struct::bitfield;
+
+#[derive(Debug)]
+struct HypercallInput<H, T: ?Sized> {
+    header: *mut H,
+    rep_count: usize,
+    _phantom: PhantomData<T>,
+}
+
+impl<H, T: ?Sized> HypercallInput<H, T> {
+    // SAFETY: the caller must guarantee the safety of the header pointer so
+    // that this function can assume the safety guarantees expected for
+    // subsequent access.
+    unsafe fn new(header: *mut H) -> Self {
+        Self {
+            header,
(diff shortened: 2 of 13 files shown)


kernel/src/cpu/percpu.rs
```diff
@@ -584,6 +585,29 @@ impl PerCpu {
...
+            if r.is_ok() {
+                let r2 = make_page_shared(vaddr + PAGE_SIZE);
+                if r2.is_err() {
+                    make_page_private(vaddr).expect("Failed to restore private page");
+                    r = r2;
+                }
+            }
+            if r.is_err() {
+                free_page(vaddr);
+                free_page(vaddr + PAGE_SIZE);
```
peterfang, line 606: I don't think this is needed since `vaddr` was allocated with order == 2

msft-jlange: Agreed.  The name "free page" is misleading (it is singular and implies that only one page will be freed) but I see that the allocation order is considered.

peterfang: I agree. Maybe we can think of a better name for this function down the road (`free_pages`?).

kernel/src/platform/mod.rs
```diff
@@ -120,9 +121,31 @@ pub trait SvsmPlatform {
...
+    fn setup_hyperv_hypercalls(&self) -> Result<(), SvsmError> {
+        Ok(())
+    }
+
+    /// Perform a hypercall.
+    /// # Safety
+    /// Hypercalls may have side-effects that affect the integrity of the
+    /// system, and the caller must take responsibility for ensuring that the
+    /// hypercall operation is safe.
+    unsafe fn hypercall(
```
peterfang, line 134: Does this also include hypercalls for KVM hosts in the future? Technically speaking, only one hypercall ABI would be used at runtime so this abstraction could be common across different hosts (with generic arguments), but just want to make sure if you have this in mind. Otherwise, can we prefix this with `hyperv_`?

kernel/src/platform/tdp.rs
```diff
@@ -125,11 +146,27 @@ impl SvsmPlatform for TdpPlatform {
...
         _size: PageSize,
-        _op: PageStateChangeOp,
+        op: PageStateChangeOp,
     ) -> Result<(), SvsmError> {
-        Err(TdxError::Unimplemented.into())
+        // The cast to u32 below is awkward, but the is_aligned() function
+        // requires its type to be convertible to u32 - which usize is not -
+        // and for an alignment check, only the low 32 bits are needed anyway.
+        if !region.start().is_aligned(PAGE_SIZE)
+            || !is_aligned(region.len() as u32, PAGE_SIZE as u32)
```
peterfang, line 157: `is_aligned()` now supports `usize` after [`ec6254ab1a8fecc416dc5c4ef48b8aaa9388c836`](https://github.com/coconut-svsm/svsm/commit/ec6254ab1a8fecc416dc5c4ef48b8aaa9388c836) is merged so this conversion is no longer needed. There are 2 other such conversions in this file so you can update them as well.

kernel/src/hyperv/hv.rs
```diff
@@ -211,12 +321,41 @@ pub fn hyperv_setup_hypercalls() -> Result<(), SvsmError> {
...
+    // First, determine if this is a Hyper-V system.
+    let is_hyperv = is_hyperv_hypervisor();
+    IS_HYPERV
+        .init(is_hyperv)
+        .expect("Hyper-V support already initialized");
+
+    if is_hyperv {
+        // If this is the BSP, then configure hypercall pages.
+        let cpu = this_cpu();
+        if cpu.get_cpu_index() == 0 {
```
peterfang, line 342: This function is only called by the BSP so is this check necessary? If it got called erroneously by an AP, it would panic at `IS_HYPERV.init()` anyway.

kernel/src/hyperv/hv.rs
```diff
@@ -211,12 +321,41 @@ pub fn hyperv_setup_hypercalls() -> Result<(), SvsmError> {
     Ok(())
 }
 
+pub fn finish_hyperv_setup() -> Result<(), SvsmError> {
```
peterfang, line 324: Is this meant for APs? I don't see this getting called anywhere. If that's the case, 'CURRENT_VTL' can only be initialized by the BSP.

kernel/src/utils/unsafe_copy.rs
```diff
@@ -0,0 +1,24 @@
...
+
+/// # Safety
+///
+/// The caller is required to ensure that the source/destination pointers and
+/// size will not result in overwriting memory unexpectedly.
+pub unsafe fn unsafe_copy_bytes(src: usize, dst: usize, size: usize) {
+    // SAFETY: the caller is responsible for ensuring that the addresses are
+    // correct and safe.
+    unsafe {
+        asm!("cld",
```
peterfang, line 17: Since there's a `cld` here, do we need to update `copy_bytes()` and `write_bytes()` accordingly?

kernel/src/mm/page_visibility.rs
```diff
@@ -34,8 +34,9 @@ use zerocopy::{FromBytes, FromZeros};
 ///
 /// Converting the memory at `vaddr` must be safe within Rust's memory model.
 /// Notably any objects at `vaddr` must tolerate unsynchronized writes of any
-/// bit pattern.
-unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
+/// bit pattern.  In addition, the caller must take responsibility for
+/// returning a page to the private state if it is ever freed.
+pub unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
```
peterfang, line 39: For TDX platforms, the guest is required to flush all cache lines prior to private->shared page conversions in order to avoid #MCE. This can be done through `wbinvd` or `clflushopt`. Since this needs to be done before the `TDVMCALL_MAP_GPA` call, we can also flush TLBs before the call.

kernel/src/mm/page_visibility.rs
```diff
@@ -69,7 +70,12 @@ unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
 /// # Arguments
 ///
 /// * `vaddr` - The virtual address of the page to be made private.
-unsafe fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
+///
+/// # Safety
+///
+/// Converting the memory at `vaddr` must be safe within Rust's memory model.
+/// No outstanding references to the page may exist.
+pub unsafe fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
```
peterfang, line 78: For TDX platforms, the guest does not need to flush TLBs or cache lines before shared->private conversions as it's the host VMM's responsibility. Therefore, `flush_tlb_global_sync()` here can be skipped for TDX, but I'm not sure if you prefer to keep it the way it is for the sake of simplicity.

kernel/src/mm/page_visibility.rs
msft-jlange: The guest can never be required to take action to prevent #MCE - otherwise, a malicious guest could crash the host system by neglecting to perform the cache flush.  The host must take responsibility for flushing the cache after the conversion is complete.

kernel/src/utils/unsafe_copy.rs
msft-jlange: Good question.  This was modeled after `unsafe_copy_bytes()`, which includes `CLD`.  Since this PR is extending existing behavior, I think that pursuing the correct pattern across the entire code base is outside the scope of this PR.

kernel/src/hyperv/hv.rs
msft-jlange: This was left over from a previous draft of the code and it should have been deleted.

kernel/src/hyperv/hv.rs
msft-jlange: This is another leftover from code reorganization.

kernel/src/mm/page_visibility.rs
peterfang: I guess the Linux TDX guest does it because it doesn't know whether the host would reliably do the flushing. If the Hyper-V host always flushes the converted private page using `TDH.PHYMEM.PAGE.WBINVD`, then it's okay that we skip this. The KVM host does it so it's safe on this side.

msft-jlange: Yes, Hyper-V always does the flush because it cannot make any assumptions about what the guest does.  Glad to know that KVM has the same opinion.

peterfang:
LGTM now, thanks!

kernel/src/svsm.rs
```diff
@@ -294,6 +295,8 @@ pub extern "C" fn svsm_main() {
         .env_setup_svsm()
         .expect("SVSM platform environment setup failed");
 
+    hyperv_setup().expect("failed to complete Hyper-V setup");
+
```
joergroedel, line 299: This adds an unconditional call to `is_hyperv_hypervisor()` for all platforms, which panics on KVM with this message:

```
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 4080c5951290
[SVSM] ERROR: Info: panicked at kernel/src/hyperv/hv.rs:275:45:
called `Option::unwrap()` on a `None` value
[SVSM] ---BACKTRACE---:
[SVSM]   [ffffff8000238035]
[SVSM]   [ffffff8000237df9]
[SVSM]   [ffffff80001b19cf]
[SVSM]   [ffffff80001b1ffe]
[SVSM]   [ffffff8000128b60]
[SVSM]   [ffffff800018cc4a]
[SVSM]   [ffffff800018cc50]
[SVSM] ---END---
```

joergroedel (commented):
Need to fix `is_hyperv_hypervisor()` up before it can be merged.

joergroedel:
Closing as changes have been merged through PR #672

PR #669, opened 2025-04-11
platform: make `SVSM_PLATFORM` a dynamic reference

msft-jlange:
All callers of `SVSM_PLATFORM` expect to call through a dynamic `SvsmPlatform` trait reference.  For efficiency, the dynamic trait reference should be stored in the global.  This is better than storing a cell containing an enum which must be matched every time the static platform is referenced.


joergroedel (approved):
LGTM, testing passed as well.

PR #671, opened 2025-04-12
Update to 1.84.1 for strict_provenance use in copy_bytes

deeglaze:
This improves address provenance hygiene to keep exposed addresses as close to their use as can be managed.

Suggested-by: Adam Dunlap <acdunlap@google.com>

### DIFF
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -44,6 +44,8 @@ pub mod svsm_gdbstub {
     const INT3_INSTR: u8 = 0xcc;
     const MAX_BREAKPOINTS: usize = 32;
 
+    // The static mutable reference to the stack is protected by the GDB_STATE lock.
+    #[allow(static_mut_refs)]
     pub fn gdbstub_start(platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
         unsafe {
             let mut target = GdbStubTarget::new();
@@ -55,7 +57,7 @@ pub mod svsm_gdbstub {
                 .run_state_machine(&mut target)
                 .expect("Failed to start GDB state machine");
             *GDB_STATE.lock() = Some(SvsmGdbStub { gdb, target });
-            GDB_STACK_TOP = GDB_STACK.as_mut_ptr().offset(GDB_STACK.len() as isize - 1) as u64;
+            GDB_STACK_TOP = GDB_STACK.as_mut_ptr().offset(GDB_STACK.len() as isize - 1);
         }
         GDB_INITIALISED.store(true, Ordering::Relaxed);
         Ok(())
@@ -120,6 +122,9 @@ pub mod svsm_gdbstub {
                 }
             }
 
+            // SAFETY: swaps the stack with GDB_STACK_TOP, which is held uniquely mutably
+            // under the GDB_STACK_LOCK, and then sets the stack to the result of handle_stop,
+            // which is the expected behavior.
             unsafe {
                 asm!(
                     r#"
@@ -132,7 +137,7 @@ pub mod svsm_gdbstub {
                     in("rsi") exception_type as u64,
                     in("rdi") &mut task_ctx,
                     in("rdx") &mut gdb_state,
-                    in("rax") GDB_STACK_TOP,
+                    in("rax") GDB_STACK_TOP.expose_provenance(),
                     options(att_syntax));
             }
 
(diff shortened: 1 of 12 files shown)


stefano-garzarella:
I'd split this in 2 patches, and I think we can merge the update to 1.84 right away.
About `exposed_provenance`, it looks like it's an ustable api, so not sure we should use it, at least until it becomes stable.

deeglaze:
It's stable in 1.84, hence the upgrade. I'll split this though.

stefano-garzarella:
> It's stable in 1.84, hence the upgrade. I'll split this though.

aaa, the CI is failing because we should update the toolchain also in `.github/workflows/rust.yml`, not sure if there is a way to support `rust-toolchain.toml` in the github workflow.

deeglaze:
I've updated the workflow.

stefano-garzarella:
> I've updated the workflow.

I saw it just after I wrote the comment :-)
Although now it looks like the CI has not started (at least in my view), but without error messages or anything.

kernel/src/debug/gdbstub.rs
```diff
@@ -120,6 +124,9 @@ pub mod svsm_gdbstub {
                 }
             }
 
+            // SAFETY: swaps the stack with GDB_STACK_TOP, which is held uniquely mutably
```
deeglaze, line 127: I think the final `movq    %rax, %rsp` might actually be incorrect and should be `movq %rsp, %rax` given context clues. /cc @roy-hopkins

stefano-garzarella:
IMO it is worth mentioning in the commit that `action-rs` is deprecated, so we need to find something else.

[Here](https://github.com/input-output-hk/ce-netsim/issues/3) I saw that they suggest this: https://github.com/dtolnay/rust-toolchain

But IMO it's also fine to do it by hand as you did, but again, I would split into 2 commits, one where we replace `action-rs` with some other thing, but keeping 1.82 and another where we update to 1.84 with the necessary changes.

deeglaze:
Thanks. Restructured commits.

user/init/Cargo.toml
```diff
@@ -3,6 +3,11 @@ name = "userinit"
 version = "0.1.0"
 edition = "2021"
 
+[[bin]]
+name = "init"
+path = "src/main.rs"
+test = false
+
```
stefano-garzarella, line 10: Why this is needed?

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.82.0"
+channel = "1.84.0"
```
stefano-garzarella, line 2: 1.84.1 is available: https://releases.rs/docs/1.84.1/, should we use that one?

user/init/Cargo.toml
deeglaze: Just marking the panic handler as not for test didn't do the trick for the duplicate definition of the panic handler for the test build, so I needed to disable tests for init.

kernel/src/debug/gdbstub.rs
```diff
@@ -55,7 +57,9 @@ pub mod svsm_gdbstub {
                 .run_state_machine(&mut target)
                 .expect("Failed to start GDB state machine");
             *GDB_STATE.lock() = Some(SvsmGdbStub { gdb, target });
-            GDB_STACK_TOP = GDB_STACK.as_mut_ptr().offset(GDB_STACK.len() as isize - 1) as u64;
+            GDB_STACK_TOP = GDB_STACK
+                .as_mut_ptr()
+                .offset(core::mem::size_of_val(&GDB_STACK) as isize - 1);
```
stefano-garzarella, line 62: I may have missed something, but I don't understand this change.

I tried a palyground to see the result, and indeed it's different:

```
GDB_STACK_TOP old 0x5d58fba19978
GDB_STACK_TOP new 0x5d58fb9a9978
GDB_STACK_TOP old - new 0x70000
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cf4f60067423f23322e3d0e4bd02c316

Is this intentional?

stefano-garzarella:
@deeglaze I found a way to avoid specifying the toolchain in the CI, see #674.
If you prefer include those commits here (about the second commit, I guess you can squash it with your own if you prefer), otherwise I'll make the PR ready after we merge this one.

kernel/src/debug/gdbstub.rs
deeglaze: It's not. I was trying to avoid using `len` for clippy, but the size is not the len due to u64 accounting for 8 bytes. I'll fix that.

rust-toolchain.toml
deeglaze: Done.

deeglaze:
> @deeglaze I found a way to avoid specifying the toolchain in the CI, see #674. If you prefer include those commits here (about the second commit, I guess you can squash it with your own if you prefer), otherwise I'll make the PR ready after we merge this one.

I've incorporated your changes here.

stefano-garzarella (approved):

kernel/src/debug/gdbstub.rs
deeglaze: Resolving since the gdb stub is mostly irrelevant to this change.

PR #672, opened 2025-04-16
Fixed up Hyperv TDP Support

joergroedel:


joergroedel:
This is the fixed-up version of PR #661

PR #674, opened 2025-04-17
Use rust-toolchain.toml version in the CI and bump it to 1.84.1

stefano-garzarella:
`rustup` will use the toolchain defined in rust-toolchain.toml if
no version is specified when installing a toolchain.

Without the specified version and using minimal profile, `rustup`
does not install the components specified with `-c`, so add another
step to install them. Also remove unnecessary components for the
`unsafe-check` job.

There is a patch release for the 1.84 toolchain, let's use it.

Based on #671

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,8 +19,11 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Install specified rust toolchain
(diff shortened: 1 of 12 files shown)


osteffenrh (approved):

stefano-garzarella:
Changes incorporated in #671

PR #675, opened 2025-04-17
scripts: Fix QEMU launcher nocc mode

osteffenrh:
There are some fixes needed to make this actually work on regular hardware, *sigh*. It is complicated... ;-)

The cbit utility does not work on non-cc hardware.
Skip the detection of the C-bit position in nocc mode.

When running in nocc mode on non-cc hardware, we can't use the kvm
backend and need to use tcg instead. To avoid another command line
option for the accelerator, let's simply always use tcg in nocc
mode.

When running with the tcg, the EPYC-v4 vcpu model does not work.
Let's use `max` with the `smep` instructions enabled instead.


scripts/launch_guest.sh
```diff
@@ -75,12 +75,25 @@ if (( QEMU_MAJOR < 9)) && [ "$CGS" = "nocc" ]; then
   exit 1
 fi
 
+# Default settings
+CPU=EPYC-v4
+ACCEL=kvm
+IGVM_OBJ=""
```
stefano-garzarella, line 81: I would move these defaults on top, near COM1_SERIAL, etc.

scripts/launch_guest.sh
```diff
@@ -75,12 +75,25 @@ if (( QEMU_MAJOR < 9)) && [ "$CGS" = "nocc" ]; then
...
+
+if [ "$CGS" = "sev" ]; then
+  C_BIT_POS=$("$SCRIPT_DIR/../utils/cbit")
```
stefano-garzarella, line 84: I'd leave this default on top, and change it if CSG is not sev when you set also CPU, ACCEL, etc.
I mean, it's strange to treat this differently from others.

scripts/launch_guest.sh
```diff
@@ -63,6 +62,7 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
+
```
stefano-garzarella, line 65: leftover?

scripts/launch_guest.sh
osteffenrh: ok.
Just need to make sure not to call the cbit util in nocc mode.

stefano-garzarella (approved):

PR #676, opened 2025-04-17
cpu: resolve safety issues with `PERCPU_AREAS`

msft-jlange:
The `PERCPU_AREAS` structure is not multi-thread safe because it contains a `Vec` that is modified as additional CPUs are started.  This would be safe when executing in a single-processor environment, but as each processor is started, additional CPU state is pushed onto `PERCPU_AREAS`, causing it to continue to change even after other processors have started running.  In order for `PERCPU_AREAS` to be sound, it must be fully populated before any additional processor can begin running.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -57,6 +57,7 @@ use crate::types::{
     PAGE_SHIFT, PAGE_SHIFT_2M, PAGE_SIZE, PAGE_SIZE_2M, SVSM_TR_ATTRIBUTES, SVSM_TSS,
 };
 use crate::utils::MemoryRegion;
+use alloc::boxed::Box;
 use alloc::format;
 use alloc::string::String;
 use alloc::sync::Arc;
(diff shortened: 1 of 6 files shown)


kernel/src/cpu/percpu.rs
```diff
@@ -98,7 +80,7 @@ pub static PERCPU_AREAS: PerCpuAreas = PerCpuAreas::new();
 // should only occur after all writes are done.
 #[derive(Debug)]
 pub struct PerCpuAreas {
-    areas: UnsafeCell<Vec<PerCpuInfo>>,
+    areas: UnsafeCell<Vec<&'static PerCpuShared>>,
```
joergroedel, line 83: The comment above this structure should be updated, describing in more detail why it is still safe (especially explaining that initialization for **all** CPUs happens from the BSP).

This can be done in a follow-up PR.

joergroedel (approved):
Changes LGTM, testing also passed.

kernel/src/cpu/percpu.rs
msft-jlange: I tried to make that clear with my statement in the comment that says, "writes to the structure only occur at initialization, from CPU 0, and reads should only occur after all writes are done."  Since "CPU 0" is the same as "BSP", I think it already says what you want it to say.  What part of the comment needs further clarification?

PR #677, opened 2025-04-17
Use strict_provenance for copy_bytes

deeglaze:
The request to use pointers more explicitly instead of `as u64` came from @AdamCDunlap. This is a type safety enhancement.

### DIFF
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -19,6 +19,7 @@ use alloc::vec::Vec;
 use core::arch::asm;
 use core::ffi::c_char;
 use core::mem::{size_of, MaybeUninit};
+use core::ptr::{with_exposed_provenance, with_exposed_provenance_mut};
 use syscall::PATH_MAX;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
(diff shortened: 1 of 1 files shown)


kernel/src/mm/guestmem.rs
```diff
@@ -462,15 +461,15 @@ fn checked_guest_region(start: PhysAddr, size: usize) -> Result<MemoryRegion<Phy
 pub unsafe fn copy_from_guest(src: PhysAddr, dst: *mut u8, size: usize) -> Result<(), SvsmError> {
     let region = checked_guest_region(src, size)?;
     let start = region.start().page_align();
-    let offset = region.start().page_offset();
+    // Offset will always be 0..4K, so this is infallible.
+    let offset: isize = region.start().page_offset().try_into().unwrap();
```
stefano-garzarella, line 465: I'm fine with this, or we can just do `as isize` like we do in line 382.

stefano-garzarella (approved):

kernel/src/mm/guestmem.rs
deeglaze: Done.

kernel/src/mm/guestmem.rs
```diff
@@ -513,15 +512,16 @@ pub fn copy_slice_to_guest(src: &[u8], dst: PhysAddr) -> Result<(), SvsmError> {
     let size = src.len();
     let region = checked_guest_region(dst, src.len())?;
     let start = region.start().page_align();
-    let offset = region.start().page_offset();
+    // Offset will always be 0..4K so this is infallible.
+    let offset: isize = region.start().page_offset().try_into().unwrap();
```
stefano-garzarella, line 516: At this point, we can do `as isize` also here. (I guess you just forgot to push the change :-) )

deeglaze: Done.

stefano-garzarella (approved):

joergroedel (approved):
Changes lgtm, thanks Dionna.

PR #679, opened 2025-04-23
Add verification for page frame allocator with safe permission-based memory access

ziqiaozhou:
This is a large PR since it starts to introduce the ghost tracked permission to reason unsafe memory access using unforgeable tracked memory permissions.

## Description of the PR

This PR add verification codes for allocator. 

### A new proof-related data type (verify_proof/src/frac_xxx.rs)

I introduced an unforgeable tracked memory permissions FracTypedPerm<T> based on verus's default memory permission [PointsTo<T>](https://github.com/verus-lang/verus/blob/main/source/vstd/raw_ptr.rs) (As for how the memory permission token is used, you can also refer to [VeriSMo](https://www.usenix.org/system/files/osdi24-zhou.pdf) to know more usage and properties it can provide, if you are really interested in). This permission type allows us to define the ownership of data when dealing with raw memory where Rust's existing ownership does not work well, especially for memory allocator. Only when the code has &mut FracTypedPerm<T>  and its shared == total_share, the code can modify the memory protected by the permission; When the code has &FracTypedPerm<T> where share > 0, the code can read the memory. This is a critical property for some memory area inside memory allocator.  All ghost functions provided by the type are fully verified to satisfying the properties we defined using state machine reasoning.

Other proof helper functions in verify_proof/src are not tighly related to the semantic of SVSM code and are just math proofs we used in svsm's proofs.

### Ghost data types defined for allocator verification.

`MemoryRegionPerms`: stores all memory permissions managed by MemoryRegion
`MRFreePerms`: stores all free memory permissions managed by MemoryRegion
`PageInfoDb`: Stores memory permissions pointing to PageInfo in reserved memory used by MemoryRegion. Those permissoins could be shared as readonly via FracTypedPerm when a page is allocated.
`MemRegionMapping`: A memory permission pointing to the global state of the memory region to ensure allocated memory and the global mem region shares the same mapping and address provenance.

### Verification of allocator (alloc_perm.verus.rs, alloc_free.verus.rs, alloc.rs)

With the above ghost permissions, in memory allocator, we need to prove the correct use of raw memory pointer and some invariants we defined for the memory region permissions stored in the memory allocator. verus checks the correct use of those memory permissions since we defined the trusted pointer TCB in tcb_ptr.rs to require a proper memory permission to read/write a raw ptr.

Another way to think about the verification here is that you can treat ghost permissions as a real memory block stored in the allocator, and then the allocator will return a memory block when allocating an instance from itself and will take a memory block when freeing an instance. When merging or splitting pages, the allocator will merge two blocks into one or split one into two. The use of those "memory block" must follow the rust ownership. With the ghost permission token, we can do this without introducing real memory overhead.

At a high level, it guarantees that

1. The MemoryRegion never returns a page for more than once.
2. The returned page must be a page managed by the region.
3. The page info stored is correct. If the page is allocated out of the allocator, the allocator should not be able to update the page info for that page until the page is freed. And thus, checking the page info is valid to know the actual type of the page.
4. A memory page to be freed must be an allocated memory. For any unsafe/unsafe code that tries to dealloc a memory page via API, that code must has ownership of the page.
5. The page counting is correct.
6. The code cannot change the definition of memory region (start, end, mapping) when some pages are allocated out of the allocator.

Among them, 1,2,4,6 are implicitly implied through the permission mechanism at any time point. 3 is partially implied by the permission and verified as an invariant in MRFreePerms::ens_perm_valid. 
Property 5 is verified for all pub or API functions by ensuring MemoryRegion::wf_perms which cannot be an invariant since MemoryRegion has some intermediate states (via some private functions) that can violate that property.

In addition, the verification helps to avoid some potential panic (e.g., Rust may panic if underflow/overflow happens) and provides proofs for some unreachable state. For all verified code, we can guarantee it will never cause overflow and underflow problems.  (If unverified code violates the preconditions we defined, it may still cause the problem, but we will incrementally ensure that will not happen).

### Encode and decode properties for some data types in allocator (alloc_type.verus.rs).

The allocator code also has many decode/encode function pairs, and I have proved the correct encode/decode relationship in SpecDecoderProof::proof_encode_decode (If a variable x can be encoded, the encoded value must be able to be decoded back to x) See that in alloc_type.verus.rs. 

## New information about the verification

In this PR, you will find three new syntax:
1. Defines extra **ghost** arguments passed to a function. The arguments does not exist in final compiled code.
```rust
#[verus_spec(with Tracked(x): Tracked<XType>)]
fn xyz() {}
```

2. Pass ghost arguments to a function call.
```rust
proof_with!(Tracked(x))
let _ = xyz();
```

3. Prove while declaring ghost/tracked variable that could be used in other proof blocks inside the same function
```rust
proof_decl!{
   let ghost x = 1usize;
   let tracked y = Map::tracked_empty();
}
```

4. Prove the loop with invariant
```rust
#[verus_spec(invariant true)] 
loop {}
```

5. Allow panic!() by enabling "vpanic" feature in builtin_macros and vstd.

In verus, the default setting disallows any panic to ensure the verified code is panic-free. But `svsm` has used panic in many places, and so we need to enable "vpanic" feature to allow panic to be used and also need to check the parameters passed to panic macro. 

## Changes

As before, verification-only files are
* under `verify_proof/src/` and `verify_externals/src/` 
* `kernel/src/*.verus.rs` are only used by verification.

Added executable files:
* [kernel/src/utils/safe_ptr.rs](https://github.com/coconut-svsm/svsm/compare/main...MSRSSP:svsm:verus-alloc-main#diff-854db1c5e0ea3471d8ebeffa0017dcf07e364bab029b32ef062cc43abfc87254) adds new raw pointer read/write function with a safe tracked permission.
Modified executable files:
* [kernel/src/address.rs](https://github.com/coconut-svsm/svsm/compare/main...MSRSSP:svsm:verus-alloc-main#diff-4345e2aa126d200ea1f34e2ef0bcffce704adc159e175aa570b28591752dbd11)
* [kernel/src/mm/address_space.rs](https://github.com/coconut-svsm/svsm/compare/main...MSRSSP:svsm:verus-alloc-main#diff-b4e8f094354e7e8101561be3695b473c69bf35e2c27f8780ed63e739831e64d3)
* [kernel/src/mm/alloc.rs](https://github.com/coconut-svsm/svsm/compare/main...MSRSSP:svsm:verus-alloc-main#diff-7eace8c7386d76dc9aa61d4f5d6e28af507602e921daad687f4064bf109d45c1)

NOTE: Inline comments will be left on executable code changes for further context.

## Verification Results
verification result could be found here at https://github.com/MSRSSP/svsm/actions/runs/14611944129/job/40991530406
Each commit is verifiable without issue (see per-commit CI result here at https://github.com/MSRSSP/svsm/pull/8)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ zerocopy = { version = "0.8.2", features = ["derive"] }
 # Verus repos
 builtin = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
 builtin_macros = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
+state_machines_macros = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
 vstd = { git = "https://github.com/verus-lang/verus", rev ="6c66898", features = ["alloc"], default-features = false }
 
 # Verification libs
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,9 +85,9 @@ zerocopy = { version = "0.8.2", features = ["derive"] }
 
 # Verus repos
 builtin = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
-builtin_macros = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
+builtin_macros = { git = "https://github.com/verus-lang/verus", rev ="6c66898", features = ["vpanic"], default-features = false }
 state_machines_macros = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
-vstd = { git = "https://github.com/verus-lang/verus", rev ="6c66898", features = ["alloc"], default-features = false }
+vstd = { git = "https://github.com/verus-lang/verus", rev ="6c66898", features = ["alloc", "allow_panic"], default-features = false }
 
 # Verification libs
 verify_proof = { path = "verify_proof", default-features = false  }
--- a/kernel/src/address.rs
+++ b/kernel/src/address.rs
@@ -358,12 +358,22 @@ impl VirtAddr {
 
     #[inline]
     #[verus_verify(external_body)]
+    #[verus_spec(ret =>
+        with Tracked(provenance): Tracked<vstd::raw_ptr::IsExposed>
+        ensures
+            ret == ptr_from_data::<T>(PtrData { addr: self@, provenance: provenance@, metadata: () }),
+    )]
     pub fn as_ptr<T>(&self) -> *const T {
         self.0 as *const T
     }
 
     #[inline]
     #[verus_verify(external_body)]
+    #[verus_spec(ret =>
+        with Tracked(provenance): Tracked<vstd::raw_ptr::IsExposed>
+        ensures
+            ret == ptr_mut_from_data::<T>(PtrData { addr: self@, provenance: provenance@, metadata: () }),
+    )]
     pub fn as_mut_ptr<T>(&self) -> *mut T {
         self.0 as *mut T
     }
--- a/kernel/src/address.rs
+++ b/kernel/src/address.rs
@@ -364,7 +364,7 @@ impl VirtAddr {
             ret == ptr_from_data::<T>(PtrData { addr: self@, provenance: provenance@, metadata: () }),
     )]
     pub fn as_ptr<T>(&self) -> *const T {
-        self.0 as *const T
+        core::ptr::with_exposed_provenance(self.0)
     }
 
     #[inline]
@@ -375,7 +375,7 @@ impl VirtAddr {
             ret == ptr_mut_from_data::<T>(PtrData { addr: self@, provenance: provenance@, metadata: () }),
     )]
     pub fn as_mut_ptr<T>(&self) -> *mut T {
-        self.0 as *mut T
+        core::ptr::with_exposed_provenance_mut(self.0)
     }
 
     #[inline]
--- a/kernel/src/address.verus.rs
+++ b/kernel/src/address.verus.rs
@@ -8,7 +8,11 @@ use crate::utils::util::{
     align_down_integer_ens, align_up_integer_ens, proof_align_down, proof_align_up,
 };
 use verify_external::convert::{exists_into, forall_into, FromSpec};
+use verify_external::hw_spec::SpecVAddrImpl;
+use vstd::set_lib::set_int_range;
+use vstd::std_specs::cmp::PartialOrdSpec;
 use vstd::std_specs::ops::AddSpec;
+
 verus! {
 
 pub broadcast group sign_extend_proof {
@@ -183,6 +187,7 @@ impl VirtAddr {
     pub open spec fn spec_add_ensures(self, offset: InnerAddr, ret: VirtAddr) -> bool {
         &&& self.offset() + offset == ret.offset()
         &&& ret === VirtAddr::from_spec((self@ + offset) as InnerAddr)
+        &&& ret === VirtAddr::from_spec((self.offset() + offset) as InnerAddr)
     }
 }
 
@@ -262,6 +267,75 @@ impl FromSpec<VirtAddr> for u64 {
     }
 }
 
+/// @Property: address can be identified by an integer.
+impl SpecVAddrImpl for VirtAddr {
+    #[verifier(inline)]
+    open spec fn spec_int_addr(&self) -> Option<int> {
+        Some(self@ as int)
+    }
+
+    #[verifier(opaque)]
+    open spec fn region_to_dom(&self, size: nat) -> Set<int> {
+        if self.is_canonical() {
+            Set::new(
+                |v: int|
+                    exists|addr: VirtAddr|
+                        addr@ == v && v <= usize::MAX && addr.is_canonical() && self.offset()
+                            <= addr.offset() < self.offset() + size,
+            )
+        } else {
+            Set::empty()
+        }
+    }
+
+    #[verus_verify(spinoff_prover)]
+    proof fn lemma_unique(v1: &Self, v2: &Self) {
+    }
+
+    #[verus_verify(spinoff_prover)]
+    proof fn lemma_vaddr_region_len(&self, size: nat)
+        ensures
+            self.is_canonical() ==> self.region_to_dom(size).len() > 0,
+    {
+        reveal(<VirtAddr as SpecVAddrImpl>::region_to_dom);
+        if self.is_canonical() {
+            assert(self.region_to_dom(size).contains(self@ as int));
+        }
+        self.lemma_valid_small_size(1, size);
+        vstd::set_lib::lemma_int_range(0, usize::MAX + 1);
+        vstd::set_lib::lemma_len_subset(self.region_to_dom(1), set_int_range(0, usize::MAX + 1));
+        vstd::set_lib::lemma_len_subset(self.region_to_dom(size), set_int_range(0, usize::MAX + 1));
+        vstd::set_lib::lemma_len_subset(self.region_to_dom(1), self.region_to_dom(size));
+    }
+
+    #[verus_verify(spinoff_prover)]
+    proof fn lemma_valid_small_size(&self, size1: nat, size2: nat) {
+        reveal(<VirtAddr as SpecVAddrImpl>::region_to_dom);
+    }
+}
+
+impl VirtAddr {
+    #[verus_verify(spinoff_prover)]
+    pub proof fn lemma_region_to_dom_merge(self, size1: nat, vaddr2: VirtAddr, size2: nat)
+        requires
+            self.is_canonical() && vaddr2.is_canonical(),
+            vaddr2.offset() == self.offset() + size1,
+        ensures
+            self.region_to_dom(size1) + vaddr2.region_to_dom(size2) == self.region_to_dom(
+                size1 + size2,
+            ),
+            self.region_to_dom(size1 + size2).difference(self.region_to_dom(size1))
+                == vaddr2.region_to_dom(size2),
+    {
+        reveal(<VirtAddr as SpecVAddrImpl>::region_to_dom);
+        assert(self.region_to_dom(size1) + vaddr2.region_to_dom(size2) =~= self.region_to_dom(
+            size1 + size2,
+        ));
+        assert(self.region_to_dom(size1 + size2).difference(self.region_to_dom(size1))
+            =~= vaddr2.region_to_dom(size2))
+    }
+}
+
 // Define a view (@) for PhysAddr
 impl View for PhysAddr {
     type V = InnerAddr;
@@ -325,6 +399,26 @@ impl vstd::std_specs::ops::AddSpecImpl<InnerAddr> for PhysAddr {
 
     open spec fn add_spec(self, offset: InnerAddr) -> PhysAddr {
         PhysAddr::from_spec((self@ + offset) as InnerAddr)
+    }
+}
+
+impl vstd::std_specs::cmp::PartialOrdSpecImpl<VirtAddr> for VirtAddr {
+    open spec fn obeys_partial_cmp_spec() -> bool {
+        true
+    }
+
+    open spec fn partial_cmp_spec(&self, other: &VirtAddr) -> Option<core::cmp::Ordering> {
+        PartialOrdSpec::partial_cmp_spec(&self@, &other@)
+    }
+}
+
+impl vstd::std_specs::cmp::PartialOrdSpecImpl<PhysAddr> for PhysAddr {
+    open spec fn obeys_partial_cmp_spec() -> bool {
+        true
+    }
+
+    open spec fn partial_cmp_spec(&self, other: &PhysAddr) -> Option<core::cmp::Ordering> {
+        PartialOrdSpec::partial_cmp_spec(&self@, &other@)
     }
 }
 
--- a/kernel/src/address.verus.rs
+++ b/kernel/src/address.verus.rs
@@ -9,6 +9,7 @@ use crate::utils::util::{
 };
 use verify_external::convert::{exists_into, forall_into, FromSpec};
 use verify_external::hw_spec::SpecVAddrImpl;
+use vstd::raw_ptr::{ptr_from_data, ptr_mut_from_data, PtrData};
 use vstd::set_lib::set_int_range;
 use vstd::std_specs::cmp::PartialOrdSpec;
 use vstd::std_specs::ops::AddSpec;
@@ -402,6 +403,7 @@ impl vstd::std_specs::ops::AddSpecImpl<InnerAddr> for PhysAddr {
     }
 }
 
+/// Assumptions for PartialOrd since it is derived.
 impl vstd::std_specs::cmp::PartialOrdSpecImpl<VirtAddr> for VirtAddr {
     open spec fn obeys_partial_cmp_spec() -> bool {
         true
@@ -412,6 +414,7 @@ impl vstd::std_specs::cmp::PartialOrdSpecImpl<VirtAddr> for VirtAddr {
     }
 }
 
+/// Assumptions for PartialOrd since it is derived.
 impl vstd::std_specs::cmp::PartialOrdSpecImpl<PhysAddr> for PhysAddr {
     open spec fn obeys_partial_cmp_spec() -> bool {
         true
@@ -419,6 +422,42 @@ impl vstd::std_specs::cmp::PartialOrdSpecImpl<PhysAddr> for PhysAddr {
 
     open spec fn partial_cmp_spec(&self, other: &PhysAddr) -> Option<core::cmp::Ordering> {
         PartialOrdSpec::partial_cmp_spec(&self@, &other@)
+    }
+}
+
+/// Assumptions for PartialEq since it is derived.
+pub assume_specification[ <VirtAddr as PartialEq<VirtAddr>>::eq ](
+    x: &VirtAddr,
+    y: &VirtAddr,
+) -> bool
+;
+
+/// Assumptions for PartialEq since it is derived.
+impl vstd::std_specs::cmp::PartialEqSpecImpl<VirtAddr> for VirtAddr {
+    open spec fn obeys_eq_spec() -> bool {
+        true
+    }
+
+    open spec fn eq_spec(&self, other: &VirtAddr) -> bool {
+        self@ == other@
+    }
+}
+
+/// Assumptions for PartialEq since it is derived.
+pub assume_specification[ <PhysAddr as PartialEq<PhysAddr>>::eq ](
+    x: &PhysAddr,
+    y: &PhysAddr,
+) -> bool
+;
+
+/// Assumptions for PartialEq since it is derived.
+impl vstd::std_specs::cmp::PartialEqSpecImpl<PhysAddr> for PhysAddr {
+    open spec fn obeys_eq_spec() -> bool {
+        true
+    }
+
+    open spec fn eq_spec(&self, other: &PhysAddr) -> bool {
+        self@ == other@
     }
 }
 
--- a/kernel/src/address_inner.verus.rs
+++ b/kernel/src/address_inner.verus.rs
@@ -18,7 +18,16 @@ pub spec const VADDR_MAX_BITS: nat = 48;
 pub spec const VADDR_LOWER_MASK: InnerAddr = 0x7fff_ffff_ffff as InnerAddr;
 
 #[verifier(inline)]
-pub spec const VADDR_UPPER_MASK: InnerAddr = !VADDR_LOWER_MASK;
+pub spec const VADDR_UPPER_MASK: InnerAddr = 0xffff_8000_0000_0000u64 as InnerAddr;
+
+// Prove the relationship among max_bits, upper and lower masks.
+proof fn lemma_bit_mask_relationship()
+    ensures
+        VADDR_UPPER_MASK == !VADDR_LOWER_MASK,
+        VADDR_LOWER_MASK == (1usize << (VADDR_MAX_BITS - 1) as usize) - 1,
+{
+    assert(VADDR_UPPER_MASK == !VADDR_LOWER_MASK) by (bit_vector);
+}
 
 #[verifier(inline)]
 pub spec const VADDR_RANGE_SIZE: InnerAddr = 0x1_0000_0000_0000u64 as InnerAddr;
--- a/kernel/src/mm/address_space.rs
+++ b/kernel/src/mm/address_space.rs
@@ -8,6 +8,11 @@ use crate::address::{PhysAddr, VirtAddr};
 use crate::mm::pagetable::{PageFrame, PageTable};
 use crate::utils::immut_after_init::ImmutAfterInitCell;
 
+use verus_stub::*;
+#[cfg(verus_keep_ghost)]
+include!("address_space.verus.rs");
+
+#[verus_verify]
 #[derive(Debug, Copy, Clone)]
 #[cfg_attr(not(any(test, target_os = "none")), expect(dead_code))]
 pub struct FixedAddressMappingRange {
@@ -17,6 +22,12 @@ pub struct FixedAddressMappingRange {
 }
 
 impl FixedAddressMappingRange {
+    #[verus_spec(ret =>
+        requires
+            Self::req_new(virt_start, virt_end, phys_start),
+        ensures
+            ret@ == LinearMap::spec_new(virt_start, virt_end, phys_start)
+    )]
     pub fn new(virt_start: VirtAddr, virt_end: VirtAddr, phys_start: PhysAddr) -> Self {
         Self {
             virt_start,
@@ -26,12 +37,17 @@ impl FixedAddressMappingRange {
     }
 
     #[cfg(target_os = "none")]
+    #[verus_spec(ret =>
+        ensures
+            ret == self@.spec_phys_to_virt(paddr@ as int)
+    )]
     fn phys_to_virt(&self, paddr: PhysAddr) -> Option<VirtAddr> {
+        proof! {self.use_type_invariant();}
         if paddr < self.phys_start {
             None
         } else {
             let size: usize = self.virt_end - self.virt_start;
-            if paddr >= self.phys_start + size {
+            if paddr - self.phys_start >= size {
                 None
             } else {
                 let offset: usize = paddr - self.phys_start;
--- a/kernel/src/mm/alloc.rs
+++ b/kernel/src/mm/alloc.rs
@@ -19,6 +19,9 @@ use core::{cmp, ptr, slice};
 #[cfg(any(test, fuzzing))]
 use crate::locking::LockGuard;
 
+#[cfg(verus_keep_ghost)]
+include!("alloc.verus.rs");
+
 /// Represents possible errors that can occur during memory allocation.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum AllocError {
--- a/kernel/src/mm/alloc.rs
+++ b/kernel/src/mm/alloc.rs
@@ -104,7 +104,6 @@ impl PageStorageType {
     const TYPE_SHIFT: u64 = 4;
     const TYPE_MASK: u64 = (1u64 << Self::TYPE_SHIFT) - 1;
     const NEXT_SHIFT: u64 = 12;
-    const NEXT_MASK: u64 = !((1u64 << Self::NEXT_SHIFT) - 1);
     const ORDER_MASK: u64 = (1u64 << (Self::NEXT_SHIFT - Self::TYPE_SHIFT)) - 1;
     // Slab item sizes are encoded in a u16
     const SLAB_MASK: u64 = 0xffff;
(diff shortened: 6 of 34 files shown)


kernel/src/mm/alloc.rs
```diff
@@ -802,39 +1462,73 @@ impl MemoryRegion {
...
             }
             PageInfo::Slab(_si) => {
+                proof_with!(Tracked(perm));
                 self.free_page_order(pfn, 0);
             }
-            PageInfo::Compound(ci) => {
-                let mask = (1usize << ci.order) - 1;
-                let start_pfn = pfn & !mask;
-                self.free_page_order(start_pfn, ci.order);
-            }
```
ziqiaozhou, line 836: This can easily cause unexpected free and cause it hard to prove safety. Removed since it is not need. In current code, VMSA page is represented with two pages to guarantee it will free all pages together and does not need hack anymore.

kernel/src/mm/alloc.rs
```diff
@@ -802,39 +1462,73 @@ impl MemoryRegion {
...
+    #[verus_spec(
+        with Tracked(perm): Tracked<PgUnitPerm<DeallocUnit>>
+        requires
+            old(self).req_try_to_merge_page(pfn, order, perm),
+        ensures
+            old(self).ens_free_page_order(self, pfn, order),
+    )]
+    #[verus_verify(spinoff_prover)]
     fn free_page_order(&mut self, pfn: usize, order: usize) {
-        match self.try_to_merge_page(pfn, order) {
```
ziqiaozhou, line 806: Fix clippy warning due to use of proof_with!() before try_to_merge_page

kernel/src/mm/alloc.rs
```diff
@@ -603,24 +1045,41 @@ impl MemoryRegion {
...
-        self.refill_page_list(0)?;
-
-        let pfn = self.get_next_page(0)?;
         let pg = PageInfo::Slab(SlabPageInfo {
             item_size: u64::from(item_size),
         });
-        self.write_page_info(pfn, pg);
-        Ok(self.start_virt + (pfn * PAGE_SIZE))
+        proof_with!(Tracked(perm));
+        self.allocate_pages_info(0, pg)
```
ziqiaozhou, line 1063: Reuse defined functions to avoid extra complexity.

kernel/src/utils/safe_ptr.rs
```diff
@@ -0,0 +1,124 @@
...
+    ///   while this shared permission reference is in use.
+    #[inline(always)]
+    #[verus_spec(v =>
+        with Tracked(perm): Tracked<&'a FracTypedPerm<T>>
+    )]
+    unsafe fn v_borrow<'a>(self) -> &'a T {
+        proof_decl! {
+            let tracked ptr_perm = perm.borrow();
+        }
+        vstd::raw_ptr::ptr_ref(self, verus_exec_expr!(Tracked(&ptr_perm)))
```
ziqiaozhou, line 78: FracTypedPerm is built above PointsTo. vstd::raw_ptr::ptr_ref is a safe pointer wrapper from verus vstd lib (See https://github.com/verus-lang/verus/blob/main/source/vstd/raw_ptr.rs). It is just a wrapper to unsafe { &*ptr } that requires the correct use of tracked memory permission PointsTo<T>, to convert unsafe raw memory access to a verified safe behavior. Right now, our verification TCB is from verus-vstd lib, we can add `tcb` crate into the project in the future to list all trusted primitives with proper tracked variable as guardian (for unsafe functions(e.g., ptr_ref, pvalidate, rmpdajust, etc.).

kernel/src/utils/safe_ptr.rs
```diff
@@ -0,0 +1,124 @@
...
+        with Tracked(perm): Tracked<&mut FracTypedPerm<T>>
+        ensures
+            perm@ == old(perm)@.update_value(MemContents::Init(v)),
+    )]
+    unsafe fn v_write(self, v: T) {
+        proof_decl! {
+            let tracked mut ptr_perm = perm.extract();
+            ptr_perm.leak_contents();
+        }
+        vstd::raw_ptr::ptr_mut_write(self, verus_exec_expr!(Tracked(&mut ptr_perm)), v);
```
ziqiaozhou, line 119: Similar to ptr_ref, ptr_mut_write is a wrapper of core::ptr::write(ptr, v) with tracked memory permission to guard the usage.

ziqiaozhou:
@joergroedel , I updated the proof code to 
1. Reduce the proof block complexity with proof macros.
2. Support newer rust version. However, latest verus only supports 1.85.1 and older and so we need to manually modify the version in kernel/Cargo.toml to allow 1.85.1 before running `cargo verify`. I have tested it locally and in a forked repo with [CI](https://github.com/MSRSSP/svsm/actions/runs/15602192598/job/43944088563?pr=8). 
3. Avoid introducing extra dependencies from Verus in non-verification mode. My old change introduced some trusted executable code for pointer with provenance and tracked permission. Since the main branch has upgraded to 1.86, we can use rust's standard lib to track provenance, and I added two trusted functions in kernel/utils/tcb_ptr.rs with tracked memory permission checks to guarantee the safe read/write access to raw pointers.

Please let me know if you need any improvement or clarification to the change. Each commit is compliable and verifiable and so I can also dive this large PR to smaller ones if you prefer smaller PRs.

ziqiaozhou:
This PR has been rebased on the update verus https://github.com/coconut-svsm/svsm/pull/761.
All commits can pass the default CI and the verification CI in https://github.com/MSRSSP/svsm/pull/12.

stefano-garzarella:
> This PR has been rebased on the update verus #761. All commits can pass the default CI and the verification CI in [MSRSSP#12](https://github.com/MSRSSP/svsm/pull/12).

Okay, since there is a dependency on another, I'll move this to Draft, please mark it ready when we merge #761

ziqiaozhou:
> > This PR has been rebased on the update verus #761. All commits can pass the default CI and the verification CI in [MSRSSP#12](https://github.com/MSRSSP/svsm/pull/12).
> 
> Okay, since there is a dependency on another, I'll move this to Draft, please mark it ready when we merge #761

I had a conflict appointment and so cannot join the SVSM Development Call on July 23. But the PR is rebased and so should be ready for review. I did not push all commit one by one but they all passed the CI before rebasing and should all work well for standard build and verification.

verify_proof/Cargo.toml
```diff
@@ -8,14 +8,16 @@ edition = "2021"
 [dependencies]
 builtin = { workspace = true, optional = true }
 builtin_macros = { workspace = true }
-vstd = { workspace = true, optional = true }
+vstd = { workspace = true }
```
stefano-garzarella: why this is not optional anymore?

verify_proof/src/bits.rs
```diff
@@ -3,6 +3,5 @@
 // Copyright (c) Microsoft Corporation
 //
 // Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
-
```
stefano-garzarella: please avoid unrelated changes like this

verify_proof/Cargo.toml
```diff
@@ -8,14 +8,16 @@ edition = "2021"
...
+state_machines_macros = { workspace = true }
 paste = "1.0"
 seq-macro = "0.3"
 
 [lints]
-workspace = true
```
stefano-garzarella: why removing this?

kernel/src/types.verus.rs
```diff
@@ -3,6 +3,7 @@
 // Copyright (c) Microsoft Corporation
 //
 // Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
+#[cfg(verus_keep_ghost)]
```
stefano-garzarella: why this is needed?

verify_proof/Cargo.toml
ziqiaozhou: I can set it back to optional. I used to set vstd as required since this lib does not contains any executable code and should only be used for verification.

verify_proof/Cargo.toml
ziqiaozhou: Re-enabled the lint again. I think I disabled it when developing due to some lint issue but now is resolved. Thanks for noticing that.

kernel/src/types.verus.rs
ziqiaozhou: This is needed since the verus_stub does not do proper erasing for the content inside the verus macro. Currently the only places that we need to keep content in verus is for constant declaration.  If using verus builtin macro, the verus will only keep executable code and erase all ghost codes when cfg(not(verus_keep_ghost)).

ziqiaozhou:
> I tried to review, but commit message are very poor. Many of us know very little about verus, so please try to explain better in the commit messages what they do, at least to get an idea of what we are reviewing. Having a bulleted list (as I already pointed out in another PR), is not the standard practice we follow.
> 
> Also please check TODO, I saw at least 1 commit message with a TODO inside, not sure if it's on purpose.

I updated the comments and the PR description. It is a little bit long, but I hope it helps you to understand the PR. 

In addition, I also added some comments in code to help developers to understand some spec and struct I defined.

kernel/src/address.verus.rs
```diff
@@ -402,6 +403,10 @@ impl vstd::std_specs::ops::AddSpecImpl<InnerAddr> for PhysAddr {
     }
 }
 
+} // verus!
+verus! {
```
stefano-garzarella: why we need 2 `verus!` blocks here?

verify_proof/src/bits.verus.rs
```diff
@@ -388,4 +418,322 @@ pub broadcast proof fn lemma_bit_usize_shr_is_div(v: usize, n: usize)
...
+        requires
+            bits2 <= m < u64::BITS,
+            bits1 == m - bits2,
+    ;
+    lemma_bit_u64_bitmask_and_min_effect(x >> bits1, bits2, bits2);
+    lemma_bit_u64_bitmask_and_min_effect(x, bits1, m);
+}
+
 } // verus!
+macro_rules! bit_xor_neighbor {
```
stefano-garzarella: Why this is out of the `verus!` macro?

verify_proof/src/frac_perm.rs
```diff
@@ -0,0 +1,404 @@
...
+        let tracked mut perm = FracPerm::empty(self.total());
+        tracked_swap(self, &mut perm);
+        let tracked (ret, mut new) = perm.extract();
+        tracked_swap(self, &mut new);
+        ret
+    }
+}
+
+} // verus!
+verus! {
```
stefano-garzarella: Why closing and re-opening it?

kernel/src/mm/address_space.rs
```diff
@@ -26,12 +37,17 @@ impl FixedAddressMappingRange {
...
+            ret == self@.spec_phys_to_virt(paddr@ as int)
+    )]
     fn phys_to_virt(&self, paddr: PhysAddr) -> Option<VirtAddr> {
+        proof! {self.use_type_invariant();}
         if paddr < self.phys_start {
             None
         } else {
             let size: usize = self.virt_end - self.virt_start;
-            if paddr >= self.phys_start + size {
+            if paddr - self.phys_start >= size {
```
stefano-garzarella, line 50: IIUC this is a real fix on the code, so please put it at least in a separate commit, but even better in a different PR.

verify_proof/src/bits.verus.rs
stefano-garzarella: I guess because it's a macro. BTW can we move this on top of this file with others macro?

Cargo.toml
```diff
@@ -72,9 +72,9 @@ zerocopy = { version = "0.8.2", features = ["derive"] }
 
 # Verus repos
 builtin = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
-builtin_macros = { git = "https://github.com/verus-lang/verus", rev ="6c66898", default-features = false }
+builtin_macros = { git = "https://github.com/verus-lang/verus", rev ="6c66898", features = ["vpanic"], default-features = false }
```
stefano-garzarella, line 88: why we need this new feature?

I think these are things that we should describe in the commit message...

kernel/src/address.rs
```diff
@@ -358,14 +358,24 @@ impl VirtAddr {
...
     #[inline]
     #[verus_verify(external_body)]
+    #[verus_spec(ret =>
+        with Tracked(provenance): Tracked<vstd::raw_ptr::IsExposed>
+        ensures
+            ret == ptr_from_data::<T>(PtrData { addr: self@, provenance: provenance@, metadata: () }),
+    )]
     pub fn as_ptr<T>(&self) -> *const T {
-        self.0 as *const T
+        core::ptr::with_exposed_provenance(self.0)
```
stefano-garzarella, line 367: IMO, changes to the code should go in a different commits/PR with a good description in the commit message of the issue and maybe also on how it was found.

kernel/src/address.rs
```diff
@@ -358,14 +358,24 @@ impl VirtAddr {
...
     #[inline]
     #[verus_verify(external_body)]
+    #[verus_spec(ret =>
+        with Tracked(provenance): Tracked<vstd::raw_ptr::IsExposed>
+        ensures
+            ret == ptr_mut_from_data::<T>(PtrData { addr: self@, provenance: provenance@, metadata: () }),
+    )]
     pub fn as_mut_ptr<T>(&self) -> *mut T {
-        self.0 as *mut T
+        core::ptr::with_exposed_provenance_mut(self.0)
```
stefano-garzarella, line 378: Ditto

stefano-garzarella (commented):
I did a quick review, let me summarize here what we should improve IMHO:
- avoid mixing `verus` changes and code changes in the same patch. If there is an issue, please fix it in a separate commit or even in another PR.
- break lines in commit description (e.g. https://github.com/coconut-svsm/svsm/pull/679/commits/424d39698c1a28f7692d7558bc82a4e12562fcff)
- explain changes in the commit description. I appreciated the PR description, but in the future, we will go through commits in our git history to debug issues or understand the verification code. So finding a commit like https://github.com/coconut-svsm/svsm/pull/679/commits/daeb5a9b1bdf86c3dd8f296eef49b1f2d540d07c that do ` 7 files changed, 412 insertions(+), 2 deletions(-)` fixing also a potential overflow in the code, with just 1 line description, is not great IMHO. How can we understand that in 1 or 2 years?
- please check the last commit description https://github.com/coconut-svsm/svsm/pull/679/commits/f9c79d84cad38b1bca6763369bbaa7e6067a0e7f
  - `Update verus version` where?
  - `Remove unncessary NEXT_MASK.` I looked in the diff, I can't see any `NEXT_MASK`
  - It looks like a summary of the entire series, so I'm confused... again, please put more effort in commit descriptions, describing the reasons for the change, and briefly what it changes.

verify_proof/src/bits.verus.rs
ziqiaozhou: Yes. I move it above.

verify_proof/src/frac_perm.rs
ziqiaozhou: Thanks for catching it.

Cargo.toml
ziqiaozhou: Documented it in comment at b48b27b9687457a49ab5245e5028c896939d5807. Verus expects verified code is panic free by default and so I need to enable panic feature to allow svsm to use panic. This is the first place we have panic inside verified code.

kernel/src/address.rs
ziqiaozhou: Now the reason is documented in 3e2ea4239b596fc5dc1688d4b947d17d78dd1b62

The change is needed since verus disallow such casting without using a trusted API.

ziqiaozhou:
> I did a quick review, let me summarize here what we should improve IMHO:
> 
> * avoid mixing `verus` changes and code changes in the same patch. If there is an issue, please fix it in a separate commit or even in another PR.
> * break lines in commit description (e.g. [424d396](https://github.com/coconut-svsm/svsm/commit/424d39698c1a28f7692d7558bc82a4e12562fcff))
> * explain changes in the commit description. I appreciated the PR description, but in the future, we will go through commits in our git history to debug issues or understand the verification code. So finding a commit like [daeb5a9](https://github.com/coconut-svsm/svsm/commit/daeb5a9b1bdf86c3dd8f296eef49b1f2d540d07c) that do ` 7 files changed, 412 insertions(+), 2 deletions(-)` fixing also a potential overflow in the code, with just 1 line description, is not great IMHO. How can we understand that in 1 or 2 years?
> * please check the last commit description [f9c79d8](https://github.com/coconut-svsm/svsm/commit/f9c79d84cad38b1bca6763369bbaa7e6067a0e7f)
>   
>   * `Update verus version` where?
>   * `Remove unncessary NEXT_MASK.` I looked in the diff, I can't see any `NEXT_MASK`
>   * It looks like a summary of the entire series, so I'm confused... again, please put more effort in commit descriptions, describing the reasons for the change, and briefly what it changes.

Thanks for the comments. I have splitted my changes in executable code from verus proofs. Now all comments start with $path: $description pattern and so it is clear `xxx.verus` path (e.g., kernel/mm/alloc.verus) indicate that change only add verus-related proofs/specs and does not really change exec code.  `verify_proof`/`verify_external` are all verus code without executable.

kernel/src/address_inner.verus.rs
```diff
@@ -18,7 +18,7 @@ pub spec const VADDR_MAX_BITS: nat = 48;
 pub spec const VADDR_LOWER_MASK: InnerAddr = 0x7fff_ffff_ffff as InnerAddr;
 
 #[verifier(inline)]
-pub spec const VADDR_UPPER_MASK: InnerAddr = !VADDR_LOWER_MASK;
```
stefano-garzarella, line 21: Is this a fix?

kernel/src/mm/alloc.rs
```diff
@@ -591,11 +589,6 @@ impl MemoryRegion {
         }
     }
 
-    /// Calculates the end virtual address of the memory region.
-    fn end_virt(&self) -> VirtAddr {
```
stefano-garzarella, line 464: IIUC we can remove this, because it's not used, right?

kernel/src/mm/alloc.rs
```diff
@@ -557,11 +589,28 @@ impl MemoryRegion {
...
+    )]
     unsafe fn page_info_mut_ptr(&mut self, pfn: usize) -> *mut PageStorageType {
+        proof! { self.lemma_page_info_ptr(pfn);}
         // SAFETY: The start of the managed virtual address space always
         // contains the meta-data about the pages. So this is safe, given the
         // caller supplied a valid pfn.
-        unsafe { self.start_virt.as_mut_ptr::<PageStorageType>().add(pfn) }
+        unsafe {
+            {
+                proof_with!(Tracked(self.perms.borrow().info_ptr_exposed));
```
stefano-garzarella: Why we need to put this in the `unsafe` block?

kernel/src/mm/alloc.rs
```diff
@@ -571,71 +620,180 @@ impl MemoryRegion {
...
+        }
         self.check_pfn(pfn);
 
         let info: PageStorageType = pi.to_mem();
         // SAFETY: we have checked that the pfn is valid via check_pfn() above.
-        unsafe { self.page_info_mut_ptr(pfn).write(info) };
+        // proved to be safe enough if passing valid tracked perm.
+        unsafe {
+            proof_with!(Tracked(perm));
+            crate::utils::tcb_ptr::ptr_write(self.page_info_mut_ptr(pfn), info);
```
stefano-garzarella: Since we are touching code, what about moving this in a separate commit?

kernel/src/mm/alloc.rs
```diff
@@ -799,39 +1111,68 @@ impl MemoryRegion {
...
         // Write new compound head
-        let pg = PageInfo::Allocated(AllocatedInfo { order: order + 1 });
+        let pg = PageInfo::Allocated(AllocatedInfo { order: new_order });
+
+        proof_with!(Tracked(&mut head_info));
         self.write_page_info(pfn, pg);
 
         // Write compound pages
-        let pg = PageInfo::Compound(CompoundInfo { order: order + 1 });
-        for i in 1..nr_pages {
```
stefano-garzarella, line 695: IMO this change should go in a separate patch

kernel/src/mm/alloc.rs
```diff
@@ -929,34 +1360,67 @@ impl MemoryRegion {
...
+    #[verus_verify(spinoff_prover)]
     fn free_page_order(&mut self, pfn: usize, order: usize) {
-        match self.try_to_merge_page(pfn, order) {
+        proof_decl! {
+            let tracked mut perm = perm;
+        }
+
+        proof_with!(Tracked(&mut perm));
+        let merged = self.try_to_merge_page(pfn, order);
+        match merged {
```
stefano-garzarella, line 1371: Why this change? I mean, why we need `merged` variable now?

kernel/src/utils/tcb_ptr.rs
```diff
@@ -0,0 +1,70 @@
...
+//
+// Copyright (c) Microsoft Corporation
+//
+// Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
+// The safe pointer lib as trusted verification TCB.
+// It provides a safe interface to access a pointer with a tracked memory permission.
+// This is a basic support for read and write access to private memory.
+// TODO:
+// - Move the spec into verify_external after verus supports extra arguments in external functions.
+// - Extend FracTypedPointer to support memory operations for confidential VMs.
```
stefano-garzarella, line 11: Still valid, right? Maybe we should mention in the commit description.

(If we move the code change in a new commit, IMHO we can also move this new file in that commit)

kernel/src/address_inner.verus.rs
ziqiaozhou: Not a fix but a simplification to proofs. In verus, VADDR_UPPER_MASK = !VADDR_LOWER_MASK is encodeded as !VADDR_LOWER_MASK instead of the final evaluated constant 0xffff_8000_0000_0000 and so I need to use bit-vec solving when using VADDR_UPPER_MASK. I choose to do that since the original use of VADDR_UPPER_MASK are mostly inside context where we will use bit-vec solver or we will introduce a lemma to solve it. But later, when more and more places start to use address, I noticed that most external code only cares its actual value.

kernel/src/mm/alloc.rs
ziqiaozhou: Yes. The only place end_virt was used is in write_page_info (c6aea54e2290bb5df33bc5eed20fcfd31bec6e1c) and that place has been changed to avoid overflow.

kernel/src/mm/alloc.rs
ziqiaozhou: This is to add a ghost variable to self.start_virt.as_mut_ptr::<PageStorageType>(). Currently, proof_with!(...) must be used together with as_mut_ptr so that verification tool will know the tracked ghost var is applied to that function call, and so I put it inside the same block with self.start_virt.as_mut_ptr::<PageStorageType>(). The original code has used `unsafe` and so I keep it there, but it does not need unsafe block in verified code since the safety is guarded by the exposed provenance tracked.

kernel/src/mm/alloc.rs
ziqiaozhou: This is a limitation of proof_with!() used inside verus_spec attribute macro. proof_with must be used close enough to try_to_merge_page where try_to_merge_page is used in a let-statement or is just a function call and could not be used in a complexed pattern like proof_with!(Tracked(&mut perm)) match elf.try_to_merge_page(pfn, order){ ... }. Such limitation might be resolved in a later version of verus but now this is how proof_with is supported in executable code.

kernel/src/mm/alloc.rs
```diff
@@ -431,11 +431,14 @@ impl MemoryRegion {
     /// The caller must provide a valid pfn, otherwise the returned pointer is
     /// undefined, as the compiler is allowed to optimize assuming there will
     /// be no arithmetic overflows.
-    unsafe fn page_info_mut_ptr(&mut self, pfn: usize) -> *mut PageStorageType {
```
stefano-garzarella, line 431: @00xc @joergroedel since you touched this code, can you check if this is okay?

kernel/src/mm/alloc.rs
```diff
@@ -431,11 +431,14 @@ impl MemoryRegion {
...
+    #[expect(clippy::needless_pass_by_ref_mut)]
+    fn page_info_mut_ptr(&mut self, pfn: usize) -> *mut PageStorageType {
         // SAFETY: The start of the managed virtual address space always
```
stefano-garzarella: This should be removed if we don't have the `unsafe` block anymore

kernel/src/mm/alloc.rs
```diff
@@ -445,11 +448,13 @@ impl MemoryRegion {
     /// The caller must provide a valid pfn, otherwise the returned pointer is
     /// undefined, as the compiler is allowed to optimize assuming there will
     /// be no arithmetic overflows.
-    unsafe fn page_info_ptr(&self, pfn: usize) -> *const PageStorageType {
+    fn page_info_ptr(&self, pfn: usize) -> *const PageStorageType {
         // SAFETY: The start of the managed virtual address space always
```
stefano-garzarella: Ditto

kernel/src/mm/alloc.rs
stefano-garzarella: Please mention that in the commit description

kernel/src/mm/alloc.rs
```diff
@@ -600,15 +600,17 @@ impl MemoryRegion {
...
+        }
     }
 
     /// Reads page information for a given page frame number.
     fn read_page_info(&self, pfn: usize) -> PageInfo {
         self.check_pfn(pfn);
 
         // SAFETY: we have checked that the pfn is valid via check_pfn() above.
-        let info = unsafe { self.page_info_ptr(pfn).read() };
+        let info = unsafe { crate::utils::tcb_ptr::ptr_read(self.page_info_ptr(pfn)) };
```
stefano-garzarella: As we usually do, can you add `use crate::utils::tcb_ptr::{ptr_read, ptr_write}` on top?

kernel/src/mm/alloc.rs
```diff
@@ -502,18 +503,28 @@ pub struct MemInfo {
 
 /// Memory region with its physical/virtual addresses, page count, as well
 /// as other details.
-#[derive(Debug, Default)]
```
stefano-garzarella, line 374: Why `Default` is removed?

kernel/src/mm/alloc.rs
stefano-garzarella: Thanks!

kernel/src/mm/alloc.rs
stefano-garzarella: I see, thanks!

kernel/src/address_inner.verus.rs
stefano-garzarella: thanks, I'd mention it in the commit description, but since it's verus code I'll not be too picky, so up to you!

kernel/src/mm/alloc.rs
ziqiaozhou: Based on the original comment, the function was marked unsafe since the returned pointer might be
undefined due to arithmetic overflows, and the compiler's optimization may assume there is no overflow and tried to use undefined ptr causing UB. That should be due to the original use of ptr::add(), Now I changed it to ptr::wrapping_add and thus should not cause undefined pointer. If start_virt or pfn is not valid, the result can still lead to invalid raw pointer which is true for VirtAddr::as_ptr(). Whether to keep both page_info_mut_ptr calculatin and ptr_read/ptr_write unsafe, or only do unsafe block for ptr read/write is trivial and does not matter too much in verus's world. It depends on the SVSM's preference. To me, I feel it is more natural to remove unsafe here since there is no unsafe functions called.

kernel/src/mm/alloc.rs
ziqiaozhou: MemoryRegion's default method is never used.

In a later verification change, we need to define a Tracked ghost.  Due to
security consideration, verus does not allow Default for an arbitrary Tracked<A> to avoid accidental duplicated tracked tokens. Only a limited Tracked types
(like Tracked<Option<A>>) may be supported to use Default but not in current verus version.  For now, it seems
that there is no need to make it complicated since MemoryRegion's
default is not used and should always be replacable by MemoryRegion::new()

kernel/src/mm/alloc.rs
ziqiaozhou: Updated the commit comment.

kernel/src/address_inner.verus.rs
ziqiaozhou: Updated the comment. Also put a tiny proof there to indicate that it is still equivalent to !VADDR_LOWER_MASK.

kernel/src/mm/alloc.rs
ziqiaozhou: Explained in a separate commit: 41aff028dec775c16415f2f7ce1f6305bc6973a3

kernel/src/mm/alloc.rs
stefano-garzarella: @ziqiaozhou I see, LGTM!

kernel/src/mm/alloc.rs
stefano-garzarella: Great, thanks for adding it in the commit description!

stefano-garzarella:
@ziqiaozhou thanks! LGTM, except for the unneeded safety comment in https://github.com/coconut-svsm/svsm/pull/679#discussion_r2241958411

PR #684, opened 2025-04-29
kernel/sev/ghcb: Always supply RCX on GHCB CPUID NAE event

tlendacky:
RCX must always be supplied when sending a GHCB CPUID NAE event to the hypervisor. When RCX is not supplied, the GHCB call will (or should) fail. The KVM hypervisor will log the following:

  kvm_amd: kvm [15381]: vcpu0, guest rIP: 0x0 vmgexit: exit code 0x72 input is not valid

Update the GHCB cpuid() function to supply RCX with a value of zero.

### DIFF
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -308,9 +308,10 @@ impl GHCB {
         Ok(())
     }
 
-    pub fn cpuid(&self, eax: u32) -> Result<CpuidResult, SvsmError> {
+    pub fn cpuid(&self, eax: u32, ecx: u32) -> Result<CpuidResult, SvsmError> {
         self.clear();
         self.set_rax_valid(eax as u64);
+        self.set_rcx_valid(ecx as u64);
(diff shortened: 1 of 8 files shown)


kernel/src/sev/ghcb.rs
```diff
@@ -311,6 +311,7 @@ impl GHCB {
     pub fn cpuid(&self, eax: u32) -> Result<CpuidResult, SvsmError> {
         self.clear();
         self.set_rax_valid(eax as u64);
+        self.set_rcx_valid(0);
```
msft-jlange, line 314: Because the CPUID protocol defines RCX, the function `cpuid()` really should be modified to take RCX as an input (`pub fn cpuid(&self, eax: u32, ecx: u32)`), and all of the callers should be modified to pass zero where they don't require a specific value.

tlendacky: Makes sense, I'll update the PR to have cpuid() take an ecx value.

msft-jlange (approved):

joergroedel (approved):
LGTM.

PR #686, opened 2025-05-04
build: do not hardcode host target triple & port build tool to Rust

00xc:
* Currently, the default target triple for the workspace is set to `x86_64-unknown-none`. This means that all workspace members that are meant to be built for the host (e.g. igvmbuilder, igvmmeasure) need their target triple explicitly hardcoded to `x86_64-unknown-linux-gnu` in various different places. This is a problem because that might break builds in non-Linux, non-x86 environments. Instead, do not set a default target triple, and let cargo figure out what it should use for the host. For bare metal components (stage 1, stage 2, the SVSM kernel) we can set the target triple explicitly because they are not meant to work outside bare metal x86 environments (at least in the near future).
* Ignore `*.igvm` build artifacts from git version control.
* Port the `build.py` script to Rust. This allows for a more strongly-defined format for the build recipes (using `serde`) and removes a dependency on Python.
* Fix a recent change to `userinit` that caused the resulting binary to not be found by the build tool.


Documentation/docs/installation/INSTALL.md
```diff
@@ -213,13 +213,13 @@ $ git submodule update --init
 That checks out the SVSM which can be built by
 
 ```
-$ FW_FILE=/path/to/firmware/OVMF.fd ./build configs/qemu-target.json
+$ FW_FILE=/path/to/firmware/OVMF.fd cargo xbuild configs/qemu-target.json
```
osteffenrh, line 216: I get an error when running this:

```
$ FW_FILE=../OVMF-svsm-gf2ba2b2a99.fd cargo xbuild configs/qemu-target.json
[...]
 Compiling igvmbuilder v0.1.0 (/home/osteffen/svsm/igvmbuilder)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.19s
error: unexpected argument '--stage1' found

  tip: a similar argument exists: '--stage2'

Usage: igvmbuilder <--tdx-stage1 <TDX_STAGE1>|--stage2 <STAGE2>|--kernel <KERNEL>|--filesystem <FILESYSTEM>|--firmware <FIRMWARE>|--output <OUTPUT>|--comport <COMPORT>|HYPERVISOR|--verbose|--sort|--policy <POLICY>|--snp|--tdp|--native|--vsm|--debug|--sev-features <SEV_FEATURES>|--alt-injection>

For more information, try '--help'.
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
````

Comment (00xc) [reply Documentation/docs/installation/INSTALL.md:216]: Should be fixed now :) thanks for testing it

**Comment by joergroedel** [comment]
Wow, impressive. Thanks for sending this, @00xc . I will have a more in-depth look soon.

**Comment by 00xc** [comment]
> I am hitting another problem.It builds the `userinit`, but then does not find the file:
> 
> ```
> $ FW_FILE=../OVMF-svsm-gdecfdf4f30.fd cargo xbuild configs/qemu-target.json  
> [...]
>    Compiling regex v1.11.1
>    Compiling libtcgtpm v0.1.0 (/home/osteffen/coconut/svsm/libtcgtpm)
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 22.57s
> Building userinit...
>    Compiling userinit v0.1.0 (/home/osteffen/coconut/svsm/user/init)
>    Compiling userlib v0.1.0 (/home/osteffen/coconut/svsm/user/lib)
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
> objcopy: 'target/x86_64-unknown-none/debug/userinit': No such file
> Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
> ```

I'm also hitting this with the old Python script, could you test it?

**Comment by 00xc** [comment]
> I added it to get rid of a duplicate panic handler build error when upgrading the Rust version. If you can avoid that on 1.84.1 without my addition, that's great.

Hm, I'm not finding that issue. Do you remember which build command caused it?

**Comment by 00xc** [comment]
FTR this would also fix the build (we'd have to apply the to other configs as well):

```patch
diff --git a/configs/qemu-target.json b/configs/qemu-target.json
index 4bad575a..0076989b 100644
--- a/configs/qemu-target.json
+++ b/configs/qemu-target.json
@@ -33,7 +33,8 @@
     },
     "fs": {
         "modules": {
-            "userinit": {
+            "init": {
+                "binary": true,
                 "path": "/init"
             }
         }
```

deeglaze:
> > I added it to get rid of a duplicate panic handler build error when upgrading the Rust version. If you can avoid that on 1.84.1 without my addition, that's great.
> 
> Hm, I'm not finding that issue. Do you remember which build command caused it?

I don't recall. I was just running all the commands that are in the github actions.

00xc:
> > > I added it to get rid of a duplicate panic handler build error when upgrading the Rust version. If you can avoid that on 1.84.1 without my addition, that's great.
> > 
> > 
> > Hm, I'm not finding that issue. Do you remember which build command caused it?
> 
> I don't recall. I was just running all the commands that are in the github actions.

The following Github actions command was not properly excluding the user crates (and clippy was emitting a warning about that):

```diff
-        run: cargo clippy --all-features --workspace --tests --exclude packit --exclude user* -- -D warnings
+        run: cargo clippy --all-features --workspace --tests --exclude packit --exclude 'user*' -- -D warnings
```

Your change effectively made clippy ignore the user crate when it came to tests anyway through `test = false` for the binary. I'm keeping your change, updating the binary name so that it matches. I've also fixed tests for `userlib`, so I'm removing the `--exclude` altogether.

osteffenrh:
This now works for me, with and without the `--release` option.

stefano-garzarella (approved):
great work, LGTM!

00xc:
Should be all good now. I was not placing the `.igvm` file in `bin/` so the test scripts were failing to find the file.

osteffenrh (approved):

msft-jlange:
If I run `make test-in-hyperv` then the `bin` directory contains `coconut-test-hyperv.igvm` as expected.  If I then run `cargo xbuild configs/hyperv-target.json` then the test IGVM file disappears.  I do not believe that running any build should remove unrelated build artifacts unless some sort of "clean" operation is explicitly requested.  Can we avoid deleting files when running xbuild?

joergroedel (approved):
Run all my tests successfully, so this looks good to me.

I approve this PR, but would like to see a redirect added from the previous `./build` tool invocation to `cargo xbuild`, so that this change does not break too many developer workflows.

00xc:
> Run all my tests successfully, so this looks good to me.
> 
> I approve this PR, but would like to see a redirect added from the previous `./build` tool invocation to `cargo xbuild`, so that this change does not break too many developer workflows.

Added a script that simply runs cargo xbuild

00xc:
> If I run `make test-in-hyperv` then the `bin` directory contains `coconut-test-hyperv.igvm` as expected. If I then run `cargo xbuild configs/hyperv-target.json` then the test IGVM file disappears. I do not believe that running any build should remove unrelated build artifacts unless some sort of "clean" operation is explicitly requested. Can we avoid deleting files when running xbuild?

This should be fixed now

msft-jlange:
> Added a script that simply runs cargo xbuild

The script should probably execute `cargo xbuild $*`.  Without the additional arguments, I just get this error output:
```
$ ./build configs/hyperv-target.json
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/xbuild`
error: the following required arguments were not provided:
  <RECIPES>...

Usage: xbuild <RECIPES>...
```

(note that for some reason GitHub will not let me attach this comment to the specific line of the script, but I'm sure you know what I'm talking about)

00xc:
> > Added a script that simply runs cargo xbuild
> 
> The script should probably execute `cargo xbuild $*`. Without the additional arguments, I just get this error output:
> 
> ```
> $ ./build configs/hyperv-target.json
>     Finished `release` profile [optimized] target(s) in 0.07s
>      Running `target/release/xbuild`
> error: the following required arguments were not provided:
>   <RECIPES>...
> 
> Usage: xbuild <RECIPES>...
> ```
> 
> (note that for some reason GitHub will not let me attach this comment to the specific line of the script, but I'm sure you know what I'm talking about)

Ugh yes, I definitely forgot to pass argv 😄 should be good now

msft-jlange (approved):

PR #687, opened 2025-05-06
vmm: introduce VMM API

msft-jlange:
This change introduces the first VMM API, which permits a processing loop to enter a guest and obtain exit information when the guest exits.  VMPL transitions unrelated to events described by guest exit messages are handled transparently so callers will only obtain relevant exit message information.


kernel/src/vmm/execloop.rs
```diff
@@ -0,0 +1,146 @@
...
+            vmsa.enable();
+
+            irq_guard
+        };
+
+        flush_tlb_global_sync();
+
+        switch_to_vmpl(GUEST_VMPL as u32);
+
+        drop(guard);
```
peterfang, line 124: Is this clearer? `let guard =` can even be omitted and just let `irq_guard` drop automatically as it goes out of scope.

```suggestion
        let mut vmsa_ref = cpu.guest_vmsa_ref();
        let caa_addr = vmsa_ref.caa_addr();
        let vmsa = vmsa_ref.vmsa();

        // Modify guest registers before disabling interrupts.
        for reg in regs {
            set_guest_register(vmsa, reg);
        }
        regs = &[];

        let guard = {
            // Interrupts must be disabled before updating APIC state.
            let irq_guard = IrqGuard::new();

            // Update APIC interrupt emulation state if required.
            cpu.update_apic_emulation(vmsa, caa_addr);

            // Make VMSA runnable again by setting EFER.SVME
            vmsa.enable();

            flush_tlb_global_sync();
            
            switch_to_vmpl(GUEST_VMPL as u32);
            
            irq_guard
        };
        drop(guard);
```

stefano-garzarella: @peterfang with your version, maybe we can avoid the explicit drop just not returning the guard, I mean something liker this:
```
        {
            // Interrupts must be disabled before updating APIC state.
            let irq_guard = IrqGuard::new();
            // Update APIC interrupt emulation state if required.
            cpu.update_apic_emulation(vmsa, caa_addr);
            // Make VMSA runnable again by setting EFER.SVME
            vmsa.enable();
            flush_tlb_global_sync();
            
            switch_to_vmpl(GUEST_VMPL as u32);
        }
```

msft-jlange: Peter's suggestion won't work because `vmsa_ref` must be dropped before calling `switch_to_vmpl()`.  I could clarify that in a comment, but that's the reason that `vmsa_ref` is captured within a separate block.  The challenge is that the IRQ guard must be established within that block but must outlive it, which is why it is passed out and dropped later.

I can look into explicitly dropping `vmsa_ref` without needing any lexical scope block, but if it's not trivial (which it may not be due to the number of other values that borrow from `vmsa_ref`), then keeping the scope block may actually be the simplest.

PR #692, opened 2025-05-09
task: eliminate extra per-task shadow stack

msft-jlange:
There is no reason to create an extra "exception" shadow stack for each task.  When an exception is delivered in kernel mode, exception delivery pushes its state onto the current shadow stack without switching stacks (unless the exception is an IST vector, but those are never per-task). It is necessary to configure a shadow stack to be used during transitions from user mode to kernel mode (this is what `PL0_SSP` is for) but entry to the kernel should always be on the task's primary kernel stack.  Consequently, for a user task, the task's primary stack can be used as the transition stack as long as it has a correct busy token at its base.


joergroedel (approved):
LGTM, thanks for making this simpler.

PR #693, opened 2025-05-12
libtcgtpm: Switch to OpenSSL 3

osteffenrh:
Policies in certain environments might prohibit the use of OpenSSL 1.x, which is currently used for the vTPM.

This PR switches over to OpenSSL 3.4.1.

Many thanks to @cclaudio for digging up his old branch for me which enables OpenSSL 3.1.

Unfortunately, this change results in a larger SVSM binary, not sure if this is a problem.

Adding `-ffunction-sections` and `-fdata-sections` to the OpenSSL build config, as well as disabling the `dh` feature, reduces the size again somewhat.

Here is a comparison of the sizes of the build for QEMU:

|               |  Size (bytes) |        | 
| :------------ |          ---: |    --- |
| openssl 1.1.1 |       5669136 |   +0%  |
| openssl 3.1.4 |       6614448 | +16.7% |
| openssl 3.4.1 |       6849712 | +20.8% |
| openssl 3.4.1 (gcc opts) |       6247024 | +10.6% |
| openssl 3.4.1 (gcc opts + no-dh) | 6147952 | + 8.4% |

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -123,6 +123,12 @@ jobs:
           name: clippy-warnings-pr
           path: clippy_warnings_pr.txt
 
+      # Ensure submodules are rebuilt correctly
+      - name: Clean workspace
+        run: |
+          git clean -xdf
+          git submodule foreach git clean -xdf
+
       - name: Checkout base branch
         uses: actions/checkout@v3
         with:
--- a/libtcgtpm/deps/libcrt/src/stub.c
+++ b/libtcgtpm/deps/libcrt/src/stub.c
@@ -142,6 +142,14 @@ int usleep(unsigned usec)
     return -1;
 }
 
+int sleep(unsigned usec)
+{
+    NOT_IMPLEMENTED;
+    return -1;
+}
+
+int timezone = 0;
+
 // dirent.h
 
 DIR *opendir(const char *name)
@@ -173,6 +181,11 @@ int stat(const char *__restrict path, struct stat *restrict buf)
 // time.h
 
 char *ctime(const time_t *t) {
-	NOT_IMPLEMENTED;
-	return NULL;
+    NOT_IMPLEMENTED;
+    return NULL;
+}
+
+time_t mktime(struct tm *tm) {
+    NOT_IMPLEMENTED;
+    return 0;
 }
--- a/libtcgtpm/deps/openssl_svsm.conf
+++ b/libtcgtpm/deps/openssl_svsm.conf
@@ -9,7 +9,7 @@ my %targets = (
         inherit_from    => [ "BASE_unix" ],
         perlasm_scheme  => "elf",
         CC              => "gcc",
-        CFLAGS          => add(combine(picker(default => "-Wall",
+        CFLAGS          => add(combine(picker(default => "-Wall -ffunction-sections -fdata-sections",
                                           debug => "-g -O0",
                                           # When compiled with any optmization (e.g. -O1),
                                           # the gcc generates code with instructions that are
--- a/libtcgtpm/deps/openssl_svsm.conf
+++ b/libtcgtpm/deps/openssl_svsm.conf
@@ -20,7 +20,7 @@ my %targets = (
                                           release => "-O0"),
                                    "-fPIE -m64 -nostdinc -nostdlib -static -fno-stack-protector -mno-red-zone")),
         bn_ops          => "SIXTY_FOUR_BIT_LONG",
-        lib_cppflags    => add("-DL_ENDIAN -DNO_SYSLOG -DOPENSSL_SMALL_FOOTPRINT -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE"),
+        lib_cppflags    => add("-DL_ENDIAN -DNO_SYSLOG -DOPENSSL_SMALL_FOOTPRINT -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DOPENSSL_USE_USLEEP"),
         sys_id          => "SVSM"
     },
 );
--- a/libtcgtpm/Makefile
+++ b/libtcgtpm/Makefile
@@ -46,8 +46,12 @@ $(OPENSSL_MAKEFILE):
 	(cd $(OPENSSL_DIR) && \
 		./Configure \
 			--config=$(DEPS_DIR)/openssl_svsm.conf \
+			--api=1.1.1 \
 			SVSM \
+			disable-legacy \
 			no-afalgeng \
+			no-aria \
+			no-asm \
 			no-async \
 			no-autoerrinit \
 			no-autoload-config \
@@ -56,12 +60,18 @@ $(OPENSSL_MAKEFILE):
 			no-capieng \
 			no-cast \
 			no-chacha \
+			no-cmac \
+			no-cmp \
 			no-cms \
 			no-ct \
 			no-deprecated \
 			no-des \
 			no-dgram \
 			no-dsa \
+			no-dso \
+			no-dtls \
+			no-dtls1_2-method \
+			no-dtls1-method \
 			no-dynamic-engine \
 			no-ec2m \
 			no-engine \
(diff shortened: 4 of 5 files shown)


osteffenrh:
I reordered the OpenSSL config options alphabetically, but did not remove any option. It might look more complicated than it actually is.

osteffenrh:
Also I an not an OpenSSL expert. Maybe the size can be optimized more by disabling more features?

osteffenrh:
Made it work up to OpenSSL 3.4.1.

Goal is 3.5, but there we need to wait for an update of the TCG TPM, since it limits the allowed OpenSSL versions.

Also, OpenSSL 3.5 needs some fixes for building it for SVSM, see https://github.com/openssl/openssl/pull/27284.
These are in the master, but not released yet. I guess we'll have to wait a bit.

osteffenrh:
Also added `-fdata-sections` for another 45kB reduction. Thanks @nicstange !

libtcgtpm/deps/libcrt/src/stub.c
```diff
@@ -173,6 +181,11 @@ int stat(const char *__restrict path, struct stat *restrict buf)
 // time.h
 
 char *ctime(const time_t *t) {
-	NOT_IMPLEMENTED;
-	return NULL;
+	  NOT_IMPLEMENTED;
+	  return NULL;
```
stefano-garzarella, line 185: Since we are touching these lines, can you remove the tabs?

libtcgtpm/deps/tpm-20-ref
stefano-garzarella, line 1: Whit this bump I think we can also remove an hack I did for unused variable, since that release should include the fix.

I mean this:
```diff
diff --git a/libtcgtpm/Makefile b/libtcgtpm/Makefile
index 26e00344..c7b0b569 100644
--- a/libtcgtpm/Makefile
+++ b/libtcgtpm/Makefile
@@ -109,11 +109,6 @@ TCGTPM_CFLAGS += -static -nostdinc -fno-stack-protector -fPIE -mno-sse -mno-red-
 TCGTPM_CFLAGS += -DFILE_BACKED_NV=NO
 TCGTPM_CFLAGS += -I$(LIBCRT_DIR)/include
 TCGTPM_CFLAGS += -I$(OPENSSL_DIR)/include
-# This is needed to workaround an issue with an unused variable when
-# FILE_BACKED_NV is set to NO. The real fix is posted here:
-# https://github.com/TrustedComputingGroup/TPM/pull/7
-# When that PR is merged, we can remove this line.
-TCGTPM_CFLAGS += -Wno-error=unused-variable
 
 TCGTPM_EXTRA_CFLAGS = -I $(DEPS_DIR)/TpmConfiguration
```

.github/workflows/rust.yml
stefano-garzarella, line 1: I would move https://github.com/coconut-svsm/svsm/pull/693/commits/d10dd2e71ac5d3b3bec629464df1b6245df16f0c at the beginning of the PR

stefano-garzarella (commented):
Cool, thanks to work on that!

Just a question, the main reason for increasing the version step-by-step is to simplify any revert in the future, right?

osteffenrh:
> Cool, thanks to work on that!
> 
> Just a question, the main reason for increasing the version step-by-step is to simplify any revert in the future, right?

Yeah, this way we can easily switch back to a different OpenSSL version if needed.

But I am open to squashing all that, if we don't want to keep individual commits.

stefano-garzarella:
> > Cool, thanks to work on that!
> > Just a question, the main reason for increasing the version step-by-step is to simplify any revert in the future, right?
> 
> Yeah, this way we can easily switch back to a different OpenSSL version if needed.
> 
> But I am open to squashing all that, if we don't want to keep individual commits.

Nope, I think your approach is great, but next time explain it in the PR description.

I'll test this a bit before approve, but LGTM!

nicstange:
It might perhaps be worthwhile to add `no-dh` to the openssl `Configure` invocations -- AFAIK it shouldn't be needed by a TPM and seems to gain another ~100kB saving.

(Switching the `Configure` target from `--debug` to `--release` OTOH is probably not really worth it: that results in a ~20kB difference, but might have the potential to break things, given the comment in `openssl_svsm.conf` about FPU usage when optimizations are enabled).

Other than that, after briefly skimming through the binary, I couldn't find any other obvious candidate openssl feature to disable.

osteffenrh:
Added another commit on top that disables `dh`.
A quick test: booting a guest and reading the PCRs worked.

nicstange (approved):
LGTM, thanks!

stefano-garzarella (approved):
Looks good also to me! Great work!

@cclaudio @deeglaze @IT302 any objection?

.github/workflows/rust.yml
```diff
@@ -122,6 +122,12 @@ jobs:
           name: clippy-warnings-pr
           path: clippy_warnings_pr.txt
 
+      # Ensure submodules are rebuild correctly
```
deeglaze, line 125:
```suggestion
      # Ensure submodules are rebuilt correctly
```

deeglaze:
> I'm double-checking with TPM WG chair to see if that's resolved.

Chris says OpenSSL 3 "works" but the bad behavior is still there, and every dependency update needs to be carefully considered. The reference implementation has a process for testing every openssl dependency is tested appropriately due to the troublesome behavior, so the commits that change the dependency beyond what is in the reference implementation should be first determined to be safe by the TCG experts first.

nicstange:
> Chris says OpenSSL 3 "works" but the bad behavior is still there, and every dependency update needs to be carefully considered. The reference implementation has a process for ensuring every openssl dependency is tested appropriately due to the troublesome behavior, so the commits that change the dependency beyond what is in the reference implementation should be first determined to be safe by the TCG experts first.

Judging from tpm-20-ref commit 5cef8a0 ("cryptolibs/Ossl/BnToOsslMath: Support openssl 3.x (>= 3.1)"), they seem to have measures in place to enforce usage with verified openssl versions only.

stefano-garzarella:
@deeglaze so IIUC the fact that the PR https://github.com/TrustedComputingGroup/TPM/pull/8 was merged is not enough, right?
What we should wait? A formal release/tag?

libtcgtpm/deps/openssl_svsm.conf
```diff
@@ -19,7 +19,7 @@ my %targets = (
                                           # vtpm_init()->libtpm:manufacture()->libcrypto:BN_CTX_new()
                                           release => "-O0"),
                                    "-fPIE -m64 -nostdinc -nostdlib -static -fno-stack-protector -mno-red-zone")),
-        bn_ops          => "SIXTY_FOUR_BIT_LONG",
```
osteffenrh, line 22: Thinking about this a bit more, I am not confident that this is the right thing to do.

AFAIU `SIXTY_FOUR_BIT_LONG` should be correct, and it seems that both `SIXTY_FOUR_BIT_LONG` and `SIXTY_FOUR_BIT` are defined during build:

```
gcc  -Icrypto -Iinclude -Iproviders/implementations/include -Iproviders/common/include   -Wall -g -O0 -fPIE -m64 -nostdinc -nostdlib -static -fno-stack-protector -mno-red-zone -DL_ENDIAN -DNO_SYSLOG -DOPENSSL_SMALL_FOOTPRINT -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DOPENSSLDIR="\"/usr/local/ssl\"" -DENGINESDIR="\"/usr/local/lib/engines-3\"" -DMODULESDIR="\"/usr/local/lib/ossl-modules\"" -DOPENSSL_BUILDING_OPENSSL -I/home/osteffen/work/svsm/coconut/svsm/libtcgtpm/deps/libcrt/include  -c -o providers/common/libcommon-lib-provider_ctx.o providers/common/provider_ctx.c
In file included from include/openssl/asn1.h:36,
                 from include/internal/cryptlib.h:27,
                 from providers/common/der/der_rsa_key.c:17:
include/openssl/bn.h:45:11: warning: "BN_ULONG" redefined
   45 | #  define BN_ULONG        unsigned long long
      |           ^~~~~~~~
include/openssl/bn.h:37:11: note: this is the location of the previous definition
   37 | #  define BN_ULONG        unsigned long
      |           ^~~~~~~~
```

osteffenrh: There was a duplicate definition of `SIXTY_FOUR_BIT` in `libcrt/include/libcrt.h`, which is not needed.

I replaced the commit that changed the `bn_ops` with a removal of that define, leaving `bn_ops` as `SIXTY_FOUR_BIT_LONG`.

Thanks @nicstange for investigating!

deeglaze (approved):

PR #694, opened 2025-05-13
add uefi variable store protocol

kraxel:
Initial bits of a UEFI variable store protocol implementation.

https://gitlab.com/kraxel/virt-firmware-rs/-/tree/main/varstore is a uefi
variable store implementation in rust.  Its far from being complete, but
should be good enough to get a linux guest boot up to the login prompt
(without secure boot).

The design idea is to re-use serialization protocol edk2 uses to talk to
SMM mode code.  The only change on the edk2 side is that the buffers are
passed to SVSM instead of SMM.  The same approach is used by the qemu
uefi variable store (new in qemu 10.0), which also receives request
buffers from the firmware.

So SVSM needs a simple protocol to send request buffers.  This is the
purpose of this PR.  Protocol documentation is in the `uefivars.md` file
created by the first patch.

TODO: Once merged use coconfs storage support to make uefi variables
persistent (will be a separate PR).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,7 @@ serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
+virtfw-varstore = { version = "0.6.1" }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,7 @@ serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
+virtfw-libefi = { version = "0.6" }
 virtfw-varstore = { version = "0.6.1" }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,7 @@ serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
+virtfw-varstore = { version = "0.6.1" }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
 embedded-io = { version = "0.6.1" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,7 @@ serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
+virtfw-libefi = { version = "0.6" }
 virtfw-varstore = { version = "0.6.1" }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
--- a/Documentation/docs/protocols/uefivars.md
+++ b/Documentation/docs/protocols/uefivars.md
@@ -1,0 +1,69 @@
+
+## SVSM UEFI MM PROTOCOL
+
+Process MM communication requests from edk2 firmware.
+
+Usually this request serialization format is used by edk2 for
+communication between normal mode and management mode (MM for short).
+
+This SVSM protocol is a thin wrapper to allow edk2 firmware send those
+requests to the SVSM instead.  This allows the SVSM to provide services
+which are usually running in SMM mode on bare metal or
+non-confidential VMs, specifically an UEFI variable store.
+
+### protocol number
+
+```
+pub const SVSM_UEFI_MM_PROTOCOL: u32 = 4;
+```
+
+### protocol request
+
+```
+const SVSM_UEFI_MM_REQUEST: u32 = 1;
+```
+Parameters:
+ * RCX: buffer address (guest physical).  Must be page aligned.
+ * RDX: buffer size
+
+Request SVSM to process the communication buffer as a UEFI request.
+The response format is request-specific.
+
+### request format
+
+The request serialization format is specified by edk2, the short
+overview below is purely informational.
+
+The buffer starts with this header:
+
+```
+// EFI_MM_COMMUNICATE_HEADER
+#[repr(C, packed)]
+pub struct MmCoreHeader {
+    pub guid: [u8; 16],
+    pub size: u64,
+}
+```
+
+The `guid` specifies the receiver of the message (typically an efi
+protocol guid or an efi event guid), and therefore also the format of
+the request data following this header.  The `size` field specified
+the size of the request data (excluding the header).
+
+### uefi spec references
+
+ * UEFI Platform Initialization Specification, section IV-5.7 "MM
+   Communication Protocol" describes the EFI_MM_COMMUNICATE_HEADER
+   struct.
+
+### edk2 code references
+
+ * MdePkg/Include/Protocol/MmCommunication.h
+ * MdeModulePkg/Include/Guid/SmmVariableCommon.h
+ * MdeModulePkg/Include/Guid/VarCheckPolicyMmi.h
+
+### linux references
+
+The linux kernel has an efi variable driver implementation using this
+request format in `drivers/firmware/efi/stmm`.  It sends those
+requests to the edk2 MM code running in arm secure partition.
--- a/configs/qemu-target.json
+++ b/configs/qemu-target.json
@@ -14,7 +14,7 @@
     },
     "kernel": {
         "svsm": {
-            "features": "vtpm",
+            "features": "vtpm,uefivars",
             "binary": false
         },
         "stage2": {
--- a/configs/qemu-target.json
+++ b/configs/qemu-target.json
@@ -14,7 +14,7 @@
     },
     "kernel": {
         "svsm": {
-            "features": "vtpm,uefivars",
+            "features": "vtpm,uefivars,secureboot",
             "binary": false
         },
         "stage2": {
--- a/configs/qemu-target.json
+++ b/configs/qemu-target.json
@@ -14,7 +14,7 @@
     },
     "kernel": {
         "svsm": {
-            "features": "vtpm",
+            "features": "vtpm,uefivars",
             "binary": false
         },
         "bldr": {
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -86,6 +86,7 @@ noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
 block = []
 uefivars = ["dep:virtfw-varstore"]
+secureboot = []
 
 [dev-dependencies]
 sha2 = { workspace = true, features = ["force-soft"] }
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -60,6 +60,7 @@ release.workspace = true
 sha2 = { workspace = true, features = ["force-soft"] }
 uuid.workspace = true
 virtio-drivers = { workspace = true, optional = true }
+virtfw-varstore = { workspace = true, optional = true }
 
 verus_builtin = { workspace = true, optional = true }
 verus_builtin_macros = { workspace = true, optional = true }
@@ -84,6 +85,7 @@ verus = ["verus_all", "verify_proof/noverify", "verify_external/noverify"]
 noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
 block = []
+uefivars = ["dep:virtfw-varstore"]
 
 [dev-dependencies]
 sha2 = { workspace = true, features = ["force-soft"] }
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -60,6 +60,7 @@ release.workspace = true
 sha2 = { workspace = true, features = ["force-soft"] }
 uuid.workspace = true
 virtio-drivers = { workspace = true, optional = true }
+virtfw-libefi = { workspace = true, optional = true }
 virtfw-varstore = { workspace = true, optional = true }
 
 verus_builtin = { workspace = true, optional = true }
@@ -85,7 +86,7 @@ verus = ["verus_all", "verify_proof/noverify", "verify_external/noverify"]
 noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
 block = []
-uefivars = ["dep:virtfw-varstore"]
+uefivars = ["dep:virtfw-varstore", "dep:virtfw-libefi"]
 secureboot = []
 
 [dev-dependencies]
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -55,6 +55,7 @@ release.workspace = true
 sha2 = { workspace = true, features = ["force-soft"] }
 uuid.workspace = true
 virtio-drivers = { workspace = true, optional = true }
+virtfw-libefi = { workspace = true, optional = true }
 virtfw-varstore = { workspace = true, optional = true }
 
 verus_builtin = { workspace = true, optional = true }
@@ -81,7 +82,7 @@ noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
 block = []
 vsock = []
-uefivars = ["dep:virtfw-varstore"]
+uefivars = ["dep:virtfw-varstore", "dep:virtfw-libefi"]
 secureboot = ["uefivars"]
 
 [dev-dependencies]
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -82,6 +82,7 @@ virtio-drivers = ["dep:virtio-drivers"]
 block = []
 vsock = []
 uefivars = ["dep:virtfw-varstore"]
+secureboot = ["uefivars"]
 
 [dev-dependencies]
 sha2 = { workspace = true, features = ["force-soft"] }
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -55,6 +55,7 @@ release.workspace = true
 sha2 = { workspace = true, features = ["force-soft"] }
 uuid.workspace = true
 virtio-drivers = { workspace = true, optional = true }
+virtfw-varstore = { workspace = true, optional = true }
 
 verus_builtin = { workspace = true, optional = true }
 verus_builtin_macros = { workspace = true, optional = true }
@@ -80,6 +81,7 @@ noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
 block = []
 vsock = []
+uefivars = ["dep:virtfw-varstore"]
 
 [dev-dependencies]
 sha2 = { workspace = true, features = ["force-soft"] }
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -17,6 +17,8 @@ use crate::greq::{
     services::get_regular_report,
 };
 use crate::mm::guestmem::{copy_slice_to_guest, read_bytes_from_guest, read_from_guest};
+#[cfg(all(feature = "uefivars", not(test)))]
+use crate::protocols::uefivars::uefi_mm_get_manifest;
 use crate::protocols::{RequestParams, errors::SvsmReqError};
 use crate::utils::MemoryRegion;
 #[cfg(all(feature = "vtpm", not(test)))]
@@ -39,6 +41,8 @@ const SVSM_ATTEST_SINGLE_SERVICE: u32 = 1;
 
 #[cfg(all(feature = "vtpm", not(test)))]
 const SVSM_ATTEST_VTPM_GUID: Uuid = uuid!("c476f1eb-0123-45a5-9641-b4e7dde5bfe3");
+#[cfg(all(feature = "uefivars", not(test)))]
+const SVSM_ATTEST_UEFI_MM_GUID: Uuid = uuid!("a4453a59-9e1b-4787-a033-1986d6adbe55");
 
 // According to
 // https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/drivers/virt/coco/sev-guest/sev-guest.c#L370
@@ -542,6 +546,10 @@ fn attest_multiple_services(params: &mut RequestParams) -> Result<(), SvsmReqErr
 
     #[cfg(all(feature = "vtpm", not(test)))]
     services.push(SVSM_ATTEST_VTPM_GUID, vtpm_get_manifest()?);
+
+    #[cfg(all(feature = "uefivars", not(test)))]
+    services.push(SVSM_ATTEST_UEFI_MM_GUID, uefi_mm_get_manifest()?);
+
     let manifest = services.to_vec()?;
     let mut nonce_and_manifest = attest_op.get_nonce()?;
     nonce_and_manifest.extend_from_slice(manifest.as_slice());
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -17,6 +17,8 @@ use crate::greq::{
     services::get_regular_report,
 };
 use crate::mm::guestmem::{copy_slice_to_guest, read_bytes_from_guest, read_from_guest};
+#[cfg(all(feature = "uefivars", not(test)))]
+use crate::protocols::uefivars::uefi_mm_get_manifest;
 use crate::protocols::{RequestParams, errors::SvsmReqError};
 use crate::utils::MemoryRegion;
 #[cfg(all(feature = "vtpm", not(test)))]
@@ -38,6 +40,8 @@ const SVSM_ATTEST_SINGLE_SERVICE: u32 = 1;
 
 #[cfg(all(feature = "vtpm", not(test)))]
 const SVSM_ATTEST_VTPM_GUID: Uuid = uuid!("c476f1eb-0123-45a5-9641-b4e7dde5bfe3");
+#[cfg(all(feature = "uefivars", not(test)))]
+const SVSM_ATTEST_UEFI_MM_GUID: Uuid = uuid!("a4453a59-9e1b-4787-a033-1986d6adbe55");
 
 // According to
 // https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/drivers/virt/coco/sev-guest/sev-guest.c#L370
@@ -545,6 +549,10 @@ fn attest_multiple_services(params: &mut RequestParams) -> Result<(), SvsmReqErr
 
     #[cfg(all(feature = "vtpm", not(test)))]
     services.push(SVSM_ATTEST_VTPM_GUID, vtpm_get_manifest()?);
+
+    #[cfg(all(feature = "uefivars", not(test)))]
+    services.push(SVSM_ATTEST_UEFI_MM_GUID, uefi_mm_get_manifest()?);
+
     let manifest = services.to_vec()?;
     let mut nonce_and_manifest = attest_op.get_nonce()?;
     nonce_and_manifest.extend_from_slice(manifest.as_slice());
@@ -569,12 +577,13 @@ fn attest_single_service_handler(params: &mut RequestParams) -> Result<(), SvsmR
 
     // Extract the GUID from the Attest Single Service Operation structure.
     // The GUID is used to determine the specific service to be attested.
-    // Currently, only the VTPM service with the GUID 0xebf176c4_2301a545_9641b4e7_dde5bfe3
-    // is supported, see 8.3.1 of the spec "Secure VM Service Module for SEV-SNP Guests
-    // 58019 Rev. 1.00" for more details.
     match attest_op.get_guid() {
         #[cfg(all(feature = "vtpm", not(test)))]
         SVSM_ATTEST_VTPM_GUID => attest_single_vtpm(params, &attest_op),
+        #[cfg(all(feature = "uefivars", not(test)))]
+        SVSM_ATTEST_UEFI_MM_GUID => {
+            attest_single_service(uefi_mm_get_manifest()?.as_slice(), params, &attest_op)
+        }
         _ => Err(SvsmReqError::unsupported_protocol()),
     }
 }
--- a/kernel/src/protocols/uefivars.rs
+++ b/kernel/src/protocols/uefivars.rs
@@ -98,6 +98,15 @@ pub fn uefi_mm_protocol_init() -> Result<(), SvsmReqError> {
     let mut store = STORE.lock();
     store.reset();
 
+    #[cfg(all(feature = "secureboot", not(test)))]
+    {
+        // hard coded configuration for now.
+        store.enroll_pk_mgmt();
+        store.enroll_kek_microsoft();
+        store.enroll_db_microsoft_uefi();
+        store.enroll_dbx_native();
+    }
+
     // In case a TPM is present, shim's fallback.efi will setup efi
     // boot variables then reboot.  This is not going to work until we
     // have persistence support for the UEFI variable store.  So turn
--- a/kernel/src/protocols/uefivars.rs
+++ b/kernel/src/protocols/uefivars.rs
@@ -17,8 +17,13 @@
 //! is implemented by the virtfw_varstore crate.  The later implements
 //! the MM protocols needed to provide an UEFI variable store.
 
+extern crate alloc;
+use alloc::vec::Vec;
+use bitfield_struct::bitfield;
 use core::mem::size_of;
+use zerocopy::{Immutable, IntoBytes};
 
+use virtfw_libefi::efivar::ids::PK;
 use virtfw_varstore::mm::core::{MmCoreHeader, core_request_dispatch};
 use virtfw_varstore::store::EfiVarStore;
 
@@ -120,4 +125,42 @@ pub fn uefi_mm_protocol_init() -> Result<(), SvsmReqError> {
     store.quirk_disable_shim_reboot(true);
 
     Ok(())
+}
+
+#[bitfield(u32)]
+#[derive(IntoBytes, Immutable)]
+pub struct UefiMmManifestFlags {
+    // non-volatile uefi variables are written to persistent storage
+    pub persistent_nv_vars: bool,
+    // secure boot is enabled
+    pub secureboot_enabled: bool,
+    // secure boot databases ('db' + 'dbx') can be updated by the
+    // guest (assuming proper pkcs7 signature of course).
+    // This should only be set in case secure boot is enabled.
+    pub secureboot_db_update: bool,
+
+    #[bits(29)]
+    _reserved: u32,
+}
+
+#[allow(dead_code)]
+#[derive(IntoBytes, Immutable)]
+struct UefiMmManifest {
+    pub version: u32,
+    pub flags: UefiMmManifestFlags,
+}
+
+pub fn uefi_mm_get_manifest() -> Result<Vec<u8>, SvsmReqError> {
+    let store = STORE.lock();
+
+    let pk = store.get(&PK.into());
+    let sb = pk.is_ok();
+
+    let flags = UefiMmManifestFlags::new()
+        .with_persistent_nv_vars(false)
+        .with_secureboot_enabled(sb)
+        .with_secureboot_db_update(false);
+    let manifest = UefiMmManifest { version: 0, flags };
+
+    Ok(manifest.as_bytes().to_vec())
 }
--- a/kernel/src/protocols/uefivars.rs
+++ b/kernel/src/protocols/uefivars.rs
@@ -104,6 +104,15 @@ pub fn uefi_mm_protocol_init() -> Result<(), SvsmReqError> {
     let mut store = STORE.lock();
     store.reset();
 
+    #[cfg(all(feature = "secureboot", not(test)))]
+    {
+        // hard coded configuration for now.
+        store.enroll_pk_mgmt();
+        store.enroll_kek_microsoft();
+        store.enroll_db_microsoft_uefi();
+        store.enroll_dbx_native();
+    }
+
     // In case a TPM is present, shim's fallback.efi will setup efi
     // boot variables then reboot.  This is not going to work until we
     // have persistence support for the UEFI variable store.  So turn
--- a/kernel/src/protocols/uefivars.rs
+++ b/kernel/src/protocols/uefivars.rs
@@ -31,6 +31,9 @@ use crate::protocols::errors::SvsmReqError;
 
 const SVSM_UEFI_MM_REQUEST: u32 = 1;
 
+pub const UEFI_MM_PROTOCOL_VERSION_MIN: u32 = 1;
+pub const UEFI_MM_PROTOCOL_VERSION_MAX: u32 = 1;
+
 // current edk2 uses 64k
 pub const UEFI_MM_BUFFER_LIMIT: usize = 256 * 1024;
 
--- a/kernel/src/protocols/uefivars.rs
+++ b/kernel/src/protocols/uefivars.rs
@@ -1,0 +1,111 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (C) 2025 Red Hat, Inc.
+//
+// Author: Gerd Hoffmann <kraxel@redhat.com>
+
+//! UEFI MM protocol implementation
+//!
+//! Process MM communication requests from edk2 firmware.
+//!
+//! Usually this request serialization format is used by edk2 for
+//! communication between normal mode and management mode (MM for
+//! short).
+//!
+//! This SVSM protocol is a thin wrapper to allow edk2 firmware send
+//! those requests to the SVSM instead.  The actual request processing
+//! is implemented by the virtfw_varstore crate.  The later implements
+//! the MM protocols needed to provide an UEFI variable store.
+
+use core::mem::size_of;
+
+use virtfw_varstore::mm::core::{MmCoreHeader, core_request_dispatch};
+use virtfw_varstore::store::EfiVarStore;
+
+use crate::address::{Address, PhysAddr};
+use crate::locking::SpinLock;
+use crate::mm::guestmem::{copy_slice_to_guest, read_bytes_from_guest, read_from_guest};
+use crate::mm::valid_phys_address;
+use crate::protocols::RequestParams;
+use crate::protocols::errors::SvsmReqError;
+
+const SVSM_UEFI_MM_REQUEST: u32 = 1;
+
+// current edk2 uses 64k
+pub const UEFI_MM_BUFFER_LIMIT: usize = 256 * 1024;
+
+static STORE: SpinLock<EfiVarStore> = SpinLock::new(EfiVarStore::new());
+
+fn check_buffer(addr: u64, size: usize) -> Result<(), SvsmReqError> {
+    let paddr = PhysAddr::from(addr);
+    if paddr.is_null() {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+    if !valid_phys_address(paddr) {
+        return Err(SvsmReqError::invalid_address());
+    }
+    if paddr.page_offset() != 0 {
+        return Err(SvsmReqError::invalid_address());
+    }
+    if size < size_of::<MmCoreHeader>() {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+    if size > UEFI_MM_BUFFER_LIMIT {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+    Ok(())
+}
+
+// process uefi mm request in passed buffer
+fn uefi_mm_request(params: &RequestParams) -> Result<(), SvsmReqError> {
+    let addr = params.rcx;
+    let size = params.rdx as usize;
+
+    let mut store = STORE.lock();
+
+    // check buffer parameters
+    check_buffer(addr, size)?;
+    log::debug!("uefi mm buffer: 0x{addr:x} +0x{size:x}");
+
+    let paddr = PhysAddr::from(addr);
+    let mmcore = read_from_guest::<MmCoreHeader>(paddr)?;
+    let boffset = size_of::<MmCoreHeader>();
+    let bsize = mmcore.size as usize;
+    if bsize > size - boffset {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+
+    let req = read_bytes_from_guest(paddr + boffset, bsize)?;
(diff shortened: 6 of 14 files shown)


kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,37 @@
...
+const SVSM_UEFI_MM_REQUEST: u32 = 23;
+
+// static mut STORE: EfiVarStore = EfiVarStore::new();
+
+fn uefi_mm_request(_params: &RequestParams) -> Result<(), SvsmReqError> {
+    // TODO: use static varstore (needs virtfw_varstore fix)
+    let mut store = EfiVarStore::new();
+
+    // TODO: pass communication buffer addr+size from edk2.
+    // TODO: map communication buffer
```
deeglaze, line 19: Given the virtio-blk support in both OVMF and (soon) SVSM, is the plan to use virtqueues for logic reuse?

kraxel: No.  The plan is to extend https://github.com/tianocore/edk2/tree/master/OvmfPkg/VirtMmCommunicationDxe so it can talk to both the qemu implementation and the svsm implementation.

kraxel:
> Number two is to setup the buffer once with a separate request (this is closer to the qemu device implementation):
> 
>     * SVSM_UEFI_MM_SETUP(paddr,size)
>     * SVSM_UEFI_MM_RESET
>     * SVSM_UEFI_MM_REQUEST

Picked this approach in the latest push.  Actual implementation is not there yet for the most part, but the intended API should be clear from the function declarations and comments.

kraxel:
The edk2 bits: https://github.com/tianocore/edk2/pull/11085

uefivars.md
```diff
@@ -0,0 +1,94 @@
...
+the request data following this header.  The `size` field specified
+the size of the request data (excluding the header).
+
+### edk2 references
+
+ * MdePkg/Include/Protocol/MmCommunication.h
+ * MdeModulePkg/Include/Guid/SmmVariableCommon.h
+ * MdeModulePkg/Include/Guid/VarCheckPolicyMmi.h
+
+### linux refernces
```
deeglaze, line 90:
```suggestion
### linux references
```

uefivars.md
```diff
@@ -0,0 +1,94 @@
+
+## SVSM UEFI MM PROTOCOL
+
+Process MM communication requests from edk2 firmware.
+
+Usually this request serialization format is used by edk2 for
+communication between normal mode and management mode (MM for short).
+
+This SVSM protocol is a thin wrapper to allow edk2 firmware send those
+requests to the SVSM instead.  This allows the SVSM provide services
```
deeglaze, line 10:
```suggestion
requests to the SVSM instead.  This allows the SVSM to provide services
```

uefivars.md
```diff
@@ -0,0 +1,94 @@
...
+non-confidential VMs, specifically an UEFI variable store.
+
+### protocol number
+
+```
+pub const SVSM_UEFI_MM_PROTOCOL: u32 = 4;
+```
+preliminary, not yet officially assigned
+
+### query command
```
deeglaze, line 21: I think this should be in a section titled "calls" since each protocol has call numbers it specifies support for at different protocol versions.

uefivars.md
```diff
@@ -0,0 +1,94 @@
...
+
+```
+const SVSM_UEFI_MM_QUERY: u32 = 1;
+```
+
+Returns success in case the protocol is supported by the svsm.
```
deeglaze, line 27: I don't think we need to worry about adding a feature query call until there is a protocol version that needs the new feature and is optional in a dimension distinct from protocol version.

uefivars.md
```diff
@@ -0,0 +1,94 @@
...
+
+Return parameters:
+
+ * RCX: protocol feature flags.  There are no features defined yet.
+   All bits are reserved and must be zero.
+
+### setup command
+
+```
+const SVSM_UEFI_MM_SETUP: u32 = 2;
```
deeglaze, line 37: Setup is an overloaded term. Can we say this is REGISTER_BUFFER?

uefivars.md
kraxel: It's also meant as a way to probe whenever the protocol is present.

uefivars.md
kraxel: Yes.

uefivars.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+Setup buffer for messages.
+
+Parameters:
+ * RCX: buffer address (guest physical).
+ * RDX: buffer size
```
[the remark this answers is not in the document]
kraxel: Current edk2 side implementation uses page alignment (and 64k max size).

I'm fine with making page alignment a hard requirement.  Will also simplify the mapping code a bit.

Should we restrict the buffer size too?  Or allow the svsm restrict the size and throw some specific error in case the buffer is too big?

uefivars.md
```diff
@@ -0,0 +1,94 @@
...
+
+### reset command
+
+```
+const SVSM_UEFI_MM_RESET: u32 = 3;
+```
+
+Reset protocol (communication buffer).
```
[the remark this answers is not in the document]
kraxel: Well, this is meant to wipe any protocol state, right now this is more a UNREGISTER_BUFFER operation and probably wouldn't do much more in the future.  Guess we should rename this too.

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    // map buffer
+    let start = paddr.page_align();
+    let offset = paddr.page_offset();
+    let end = (paddr + buffer.size).page_align_up();
+    let guard = PerCPUPageMappingGuard::create(start, end, 0)?;
+    let vaddr = guard.virt_addr() + offset;
+
+    // SAFETY: vaddr comes from a new mapped region.
+    let buffer = unsafe { from_raw_parts_mut(vaddr.as_mut_ptr::<u8>(), buffer.size) };
```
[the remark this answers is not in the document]
kraxel: I'll have a look.

uefivars.md
deeglaze: The core protocol's query call is meant for that purpose.

kernel/src/protocols/uefivars.rs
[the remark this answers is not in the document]
deeglaze: There's additional drawback here from reading directly from the guest during processing when you might have validation to do ahead of time. It's much safer to copy in, copy out. This code is still missing additional safety checks for the whole region (you only check the initial address) and the requisite preconditions for creating a slice without UB. The copy in requirement is specifically to get a safety condition about the non-mutability of the slice from another source:

> The memory referenced by the returned slice must not be mutated for the duration of lifetime 'a, except inside an UnsafeCell.

So please use the library that was specifically written for handling guest memory reads and writes.

uefivars.md
kraxel: ok, makes sense.

/me notes that the TPM protocol is not hooked up in core_query_protocol (ping @stefano-garzarella). I've looked at how the TPM protocol handles things fir guidance ...

kernel/src/protocols/uefivars.rs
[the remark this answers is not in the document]
kraxel: The variable store operates in copy in + copy out mode, it's just the code is not in the svsm repo, see https://gitlab.com/kraxel/virt-firmware-rs/-/blob/main/varstore/src/mm/core.rs

If you think this is not good enough I can switch to the use svsm guestmem functions instead, I only need to rearrange the varstore code first.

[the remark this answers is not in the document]
deeglaze: I don't think SVSM should make assumptions about how the guest will use this ABI since it's a security boundary. The PerCPUMapping and valid_phys_address checks are hard to reason about for safety, so any guest memory interactions should be through guestmem. If you have any ideas for abstractions that make clean use of a single PerCPUMapping for a sequence of memory transactions, then I would suggest you build them there and use them here. In the meantime, we can live with an additional page table interaction given the already costly boundary that is getting crossed to switch VMPLs and further use a paravirtual device to interact with a block device.

uefivars.md
stefano-garzarella: @kraxel thanks for pinging me! I'll fix (or feel free to open a PR if you want)

kernel/src/protocols/uefivars.rs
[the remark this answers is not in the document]
stefano-garzarella: > Hmm, drawback is we'll go map the buffer twice per request then, first to read the request, then to write the response.

yeah, I was thinking the same when using that API in #603 , maybe we can try to create some abstraction. I'd suggest opening a PR with just that if you have something.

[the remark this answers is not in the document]
kraxel: > yeah, I was thinking the same when using that API in #603 , maybe we can try to create some abstraction. I'd suggest opening a PR with just that if you have something.

It's actually three times.  First read fixed-size header, then read + write actual request (both variable size).  See most recent push which switches to guestmem functions (incremental patches for now, will clean up the branch once it passed testing).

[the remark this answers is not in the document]
tlendacky: If the buffer will need to be mapped on each call anyway, does it makes sense to just have the single SVSM_UEFI_MM_REQUEST call which supplies the buffer address and size? It would be simpler and eliminate the need to register / unregister the buffer.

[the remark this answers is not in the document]
kraxel: Well, the current protocol design gives the SVSM some flexibility how it wants handle the buffer mapping.  It can work with temporary (per cpu) mappings on each request, or it can work with a permanent (global) mapping.  Passing the buffer on each request takes the second option off the table.

If we declare we want stick to the temporary mappings forever then yes, we can go for a single REQUEST call.

Documentation/docs/protocols/uefivars.md
```diff
@@ -0,0 +1,83 @@
...
+
+```
+const SVSM_UEFI_MM_REGISTER_BUFFER: u32 = 1;
+```
+
+Setup communication buffer for messages.
+
+Parameters:
+ * RCX: buffer address (guest physical).  Must be page aligned.
+ * RDX: buffer size
```
tlendacky, line 27: How do you want the protocol to behave if a buffer is already registered when this call is made? Implicitly unregister the current buffer and register the new buffer or return an error? Also, is there any alignment requirements for the buffer?

uefivars.md
```diff
@@ -0,0 +1,83 @@
...
+ * RCX: buffer address (guest physical).  Must be page aligned.
+ * RDX: buffer size
+
+#### unregister buffer command
+
+```
+const SVSM_UEFI_MM_UNREGISTER_BUFFER: u32 = 2;
+```
+
+Clear communication buffer configuration.
```
tlendacky: How do you want this to behave if no buffer is registered, just return success or return an error?

Documentation/docs/protocols/uefivars.md
```diff
@@ -0,0 +1,83 @@
...
+const SVSM_UEFI_MM_REQUEST_DOORBELL: u32 = 3;
+```
+
+Request doorbell for SVSM to process the communication buffer as a UEFI request.
+The response format is requester-specific.
+
+### request format
+
+The request serialization format is specified by edk2, the short
+overview below is purely informational.
```
tlendacky, line 35: Do you need something to indicate that there is a new request in the buffer? What if a malicious EFI component registers a new buffer to prevent the request from actually being processed or is there something in the request protocol to verify execution of the request?

Documentation/docs/protocols/uefivars.md
kraxel: Current implementation implicitly unregisters.  Buffer should be page aligned (see quoted patch).

uefivars.md
kraxel: Current implementation returns success.

Documentation/docs/protocols/uefivars.md
kraxel: That kind of DoS is indeed possible with the buffer register/unregister.  Passing the buffer address on each request would prevent that, and it would also allow multipe users.  The linux kernel could have a efi variable driver which calls the svsm protocol directly, and that driver would not break variable access via uefi runtime services.

Documentation/docs/protocols/uefivars.md
tlendacky: Doh! right there in front of me.

Documentation/docs/protocols/uefivars.md
tlendacky: Given that, do you want to stick to the register/unregister/request format or just have a single call with the buffer to be processed? I think the latter would be safer/better.

kraxel: Yes, switching to a single call looks increasingly attractive.  I'll check next week.

tlendacky: Btw, what manifest information do you want to supply for this service for the attestation report?

kraxel: > Btw, what manifest information do you want to supply for this service for the attestation report?

Hmm, good question.  Are there any guidelines?

I see the vtpm includes the (public) ek, which represents the identity of the VM.  Which makes sense to me.  There isn't something simliar in the uefi variable store.  The secure boot configuration (specifically 'db' + 'dbx') could be relevant for attestation.  But right now it is compiled in and covered by launch measurement, so I think it would be duplication.

Related: https://github.com/microsoft/igvm/issues/91

I'm looking into ways to (a) make db/dbx configurable, and (b) launch a linux kernel directly.  Right now my idea is to have a igvm directive added to the spec with pointers to memory pages containing db / dbx / kernel, so you can easily update existing igvm files (see linked issue for details).

The db/dbx pages can be covered by launch measurement.  But they don't have to.  Maybe it is better to put them into the attestation manifest instead?

deeglaze: Since UEFI variables are going to be managed by SVSM, and it's up to the SVSM to provide protections of the persisted content, I think attesting to the persistence properties is prudent. Anything configurable and security-relevant needs to be measured. We're going to need to enable different behavior with flag flips as they become available and roll out according to production principles.

Now for what specifically should be attested, I think that we could use TCG non-volatile storage attributes TPMA_NV and TPMA_NV_EXP can be used as inspiration https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-2-Version-184_pub.pdf 
I wouldn't recommend making the properties configurable at the per-variable level, so many bits are unnecessary.

tlendacky: > The db/dbx pages can be covered by launch measurement. But they don't have to. Maybe it is better to put them into the attestation manifest instead?

If the db/dbx contents will ever be updateable outside of the OVMF binary then it would make sense to make them part of the manifest. Will they be? Or would it make sense that if they ever change, then the OVMF binary has to be rebuilt?

deeglaze: They are updateable by the UEFI specification with use of the appropriate keys. What you want to state is that the variable store does not permit SetVariable to non-volatile variables, then that should be attested. It might be plain from the behavior of the SVSM binary as measured while not yet configurable, so we could potentially wait to codify how these properties can be attested in a later manifest version. I'd much rather have some concretely reported, like nv_immutable for disallowing mutation of non-volatile UEFI variables.

kraxel: Sketched a manifest, see the uefi_mm_get_manifest patch just pushed.

Writing non-volatile variables works fine, it is just that the changes are not actually persistent because we do not yet have encrypted storage support.

Writing non-volatile *authenticated* variables (including 'db' and 'dbx') is not supported yet due to lack of pkcs7 signature verification.  Also doesn't make much sense to support until we have persistence support for the efi variables.  Throws EFI_UNSUPPORTED errors right now.  Long-term we probably want a config switch to allow/deny that, and of course attest the state of the config switch.

tlendacky: I was thinking the KEK database should also be included since that can be updated, too. From the commit, I noticed that you have a flags entry, I can add that and define the current flags that you have listed. Does something like this work for you?

```
Byte Offset   Size (Bytes)   Meaning
0x000         4              Total size of manifest data
0x004         4              Flags
0x008         4              Offset of signature database (db)
0x00C         4              Size of signature database (db)
0x010         4              Offset of revoked signatures database (dbx) 
0x014         4              Size of revoked signatures database (dbx) 
0x018         4              Offset of Key Enrollment Key database (KEK) 
0x01C         4              Size of Key Enrollment Key database (KEK) 
0x020         Variable       Db / dbx / KEK 
```
If data for a database isn't present, then the offset and size for that database must be set to zero.

tlendacky: Also, this is the GUID I have assigned for the UEFI Management Mode protocol: a4453a59-9e1b-4787-a033-1986d6adbe55

kraxel: > ```
> Byte Offset   Size (Bytes)   Meaning
> 0x000         4              Total size of manifest data
> 0x004         4              Flags
> 0x008         4              Offset of signature database (db)
> 0x00C         4              Size of signature database (db)
> 0x010         4              Offset of revoked signatures database (dbx) 
> 0x014         4              Size of revoked signatures database (dbx) 
> 0x018         4              Offset of Key Enrollment Key database (KEK) 
> 0x01C         4              Size of Key Enrollment Key database (KEK) 
> 0x020         Variable       Db / dbx / KEK 
> ```

Any specific reason for including the offsets and the total size?  These can be calculated from the size fields, and I'd prefer to not have redundant data in there.

Not sure whenever it makes sense to include the KEK.  For allowing / denying EFI binaries 'db' and 'dbx' are relevant.  If we want include KEK (because KEK signatures can update db+dbx) then we should include PK too (because PK signatures can update KEK).

tlendacky: > Any specific reason for including the offsets and the total size? These can be calculated from the size fields, and I'd prefer to not have redundant data in there.

What is the format for db /dbx? Is there an identifier that would allow to distinguish which db is which? We would need that documented somewhere in order to be sure this data can be parsed properly.



> Not sure whenever it makes sense to include the KEK. For allowing / denying EFI binaries 'db' and 'dbx' are relevant. If we want include KEK (because KEK signatures can update db+dbx) then we should include PK too (because PK signatures can update KEK).

Since the KEK does allow for updating the db and dbx, it and the PK should probably be included, too. @joergroedel, @deeglaze and anyone else, any thoughts?

deeglaze: I don't think that including the variables in the manifest makes sense at all, given that they're measured into PCR7 when their contents are actually security-relevant.

tlendacky: > I don't think that including the variables in the manifest makes sense at all, given that they're measured into PCR7 when their contents are actually security-relevant.

So just the flags field then?

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,197 @@
...
+    Ok(())
+}
+
+#[derive(IntoBytes, Immutable)]
+#[bitfield(u32)]
+pub struct UefiMmManifestFlags {
+    // non-volatile uefi variables are written to persistent storage
+    pub persistent_nv_vars: bool,
+    // secure boot is enabled
+    pub secureboot_enabled: bool,
```
deeglaze, line 130: This is measured into PCR7. I wouldn't duplicate it here.

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,197 @@
...
+    // off that behaviour for now (via EFI variable).
+    store.quirk_disable_shim_reboot(true);
+
+    Ok(())
+}
+
+#[derive(IntoBytes, Immutable)]
+#[bitfield(u32)]
+pub struct UefiMmManifestFlags {
+    // non-volatile uefi variables are written to persistent storage
```
deeglaze, line 127: I think this is enough only if the persistent storage for UEFI variables is owned by the UEFI variable service.
If the storage volume is homogeneous across different services' usage, you're punting the persistent storage properties to an unspecified device that will also need an attestation manifest to claim things like "no pending writes. encrypted with attested key. rollback protected by counter service root key X."

I think different platforms will have different capabilities. For instance, secure boot failures confuse the crap out of a lot of customers, and it's really useful for Google customer service agents to have access to the secure boot variables to help diagnose problems. That and to perform larger queries to find instances that are running with older known-vulnerable databases to either report to the customer or intervene transparently.

With a heterogeneous storage device, the variables can be stored encrypted and rollback-safe with the attested key in one volume, and with a host-visible key in another. Boot can verify the values are the same to report tampering, only trust the attested values, but still make the values visible to customer service when they get the Google key released to decrypt the state as part of normal business procedure.

For this reason I think the backing storage device's service GUID and some volume identifier within the service should be what's recorded. This can all be for v2 of the manifest when we really figure out the persistent state aspects, but I don't think the format here is long for this world.

Documentation/docs/protocols/uefivars.md
deeglaze: Responded elsewhere about the flags. The storage properties of the variables are the most important. If PKCS7 authenticode verification is configurable, then that should be attested too. It shouldn't be to be stardards compliant but you never know.

kernel/src/protocols/uefivars.rs
kraxel: > This can all be for v2 of the manifest when we really figure out the persistent state aspects, but I don't think the format here is long for this world.

Sounds like adding a version field is a good idea.

Documentation/docs/protocols/uefivars.md
kraxel: > What is the format for db /dbx?

EFI signature database.

> Is there an identifier that would allow to distinguish which db is which?

No.

> We would need that documented somewhere in order to be sure this data can be parsed properly.

Yes, the order is important, and the size too.
(assuming we are adding these to the manifest, when relying on PCR7 instead this is moot ...).

tlendacky: Right...  so based on the comment below, what about the V1 manifest just having a version field and a flags field:

```
Byte Offset   Size (Bytes)   Meaning
0x000         4              Version
0x004         4              Flags
```
We would need to define what the version field would mean and how it gets set/updated or whether it just corresponds to the manifest version.

We would need to define the flags provided in the flags field with all reserved bits being set as zero.

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,105 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (C) 2025 Red Hat, Inc.
+//
+// Author: Gerd Hoffmann <kraxel@redhat.com,>
```
luigix25: extra comma?

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,105 @@
...
+    let boffset = size_of::<MmCoreHeader>();
+    let bsize = mmcore.size as usize;
+    if boffset + bsize > size {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+
+    let req = read_bytes_from_guest(paddr + boffset, bsize)?;
+    let rsp = core_request_dispatch(&mut store, &mmcore.guid, &req);
+    if rsp.len() > bsize {
+        // should not happen (add SvsmReqError::internal_error() ?)
```
luigix25: mmh, leftover todo?

kernel/Cargo.toml
```diff
@@ -60,6 +60,7 @@ release.workspace = true
 sha2 = { workspace = true, features = ["force-soft"] }
 uuid.workspace = true
 virtio-drivers = { workspace = true, optional = true }
+virtfw-varstore = { version = "0.6.1", optional = true }
```
luigix25: up until now all dependencies have been added to the "top level" Cargo.toml. Should we do the same here?

kernel/src/protocols/uefivars.rs
kraxel: Yes, will drop.

kernel/src/protocols/uefivars.rs
kraxel: More an RfC.  How to handle errors which should in theory never ever happen?  Is it ok to panic then?  If so, then I could siimply `assert!(rsp.len() <= bsize)` ...

kernel/Cargo.toml
kraxel: Guess that makes sense for consistency, even if I don't expect that being ever used outside the `kernel/` crate.

kernel/src/protocols/uefivars.rs
luigix25: mmh good point. How critical would be a wrong-sized buffer?

I don't honestly know, maybe asserting is too much. I'm more for an error, like you did.

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    // check buffer parameters
+    check_buffer(addr, size)?;
+    log::debug!("uefi mm buffer: 0x{addr:x} +0x{size:x}");
+
+    let paddr = PhysAddr::from(addr);
+    let mmcore = read_from_guest::<MmCoreHeader>(paddr)?;
+    let boffset = size_of::<MmCoreHeader>();
+    let bsize = mmcore.size as usize;
+    if boffset + bsize > size {
```
luigix25: IIUC `bsize` is read directly from the guest memory and it's unchecked.
This means that potentially this sum can wrap around and this check can be skipped.

usize is 64 bit, so IMO you have to do it on purpose..

kernel/src/protocols/uefivars.rs
kraxel: It would be a bug in the `virtfw-varstore` crate and not something the guest should be able to trigger.  If the return buffer is too small the `virtfw-varstore` can return a response to the guest saying so (not visible here, see https://gitlab.com/kraxel/virt-firmware-rs/-/blob/main/varstore/src/mm/variable.rs, grep for BUFFER_TOO_SMALL).

kernel/src/protocols/uefivars.rs
kraxel: Updated in incremental patch [to be squashed, left separate for now for easier review]

luigix25: lgtm, thanks!

Documentation/docs/protocols/uefivars.md
```diff
@@ -0,0 +1,70 @@
+
```
luigix25, line 1: should this file be placed in in Documentation/docs/developer?

uefivars.md
```diff
@@ -0,0 +1,70 @@
...
+requests to the SVSM instead.  This allows the SVSM to provide services
+which are usually running in SMM mode on bare metal or
+non-confidential VMs, specifically an UEFI variable store.
+
+### protocol number
+
+```
+pub const SVSM_UEFI_MM_PROTOCOL: u32 = 4;
+```
+preliminary, not yet officially assigned
```
luigix25: is this still true?

Given https://github.com/tianocore/edk2/pull/11085, probably yes

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,111 @@
...
+//!
+//! Process MM communication requests from edk2 firmware.
+//!
+//! Usually this request serialization format is used by edk2 for
+//! communication between normal mode and management mode (MM for
+//! short).
+//!
+//! This SVSM protocol is a thin wrapper to allow edk2 firmware send
+//! those requests to the SVSM instead.  The actual request processing
+//! is implemented by the virtfw_varstore crate.  The later implements
```
luigix25: /s/later/latter

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,111 @@
...
+use crate::locking::SpinLock;
+use crate::mm::guestmem::{copy_slice_to_guest, read_bytes_from_guest, read_from_guest};
+use crate::mm::valid_phys_address;
+use crate::protocols::RequestParams;
+use crate::protocols::errors::SvsmReqError;
+
+const SVSM_UEFI_MM_REQUEST: u32 = 1;
+
+// current edk2 uses 64k
+pub const UEFI_MM_BUFFER_LIMIT: usize = 256 * 1024;
```
luigix25: does this need to be public?

kernel/Cargo.toml
```diff
@@ -86,6 +86,7 @@ noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
 block = []
 uefivars = ["dep:virtfw-varstore"]
+secureboot = []
```
luigix25: does secureboot need to depend on uefivars?

kernel/src/protocols/uefivars.rs
```diff
@@ -121,3 +126,41 @@ pub fn uefi_mm_protocol_init() -> Result<(), SvsmReqError> {
...
+    #[bits(29)]
+    _reserved: u32,
+}
+
+#[allow(dead_code)]
+#[derive(IntoBytes, Immutable)]
+struct UefiMmManifest {
+    pub version: u32,
+    pub flags: UefiMmManifestFlags,
+}
```
luigix25: what about dropping dead_code and renaming fields to` _version` and `_flags`?

kernel/src/protocols/uefivars.rs
```diff
@@ -121,3 +126,41 @@ pub fn uefi_mm_protocol_init() -> Result<(), SvsmReqError> {
...
+    let store = STORE.lock();
+
+    let pk = store.get(&PK.into());
+    let sb = pk.is_ok();
+
+    let flags = UefiMmManifestFlags::new()
+        .with_persistent_nv_vars(false)
+        .with_secureboot_enabled(sb)
+        .with_secureboot_db_update(false);
+    let manifest = UefiMmManifest { version: 0, flags };
```
luigix25, line 157: why version 0?

Cargo.toml
```diff
@@ -93,6 +93,7 @@ serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
+virtfw-libefi = { version = "0.6" }
```
luigix25, line 97: latest version is 0.6.1, worth updating?

Cargo.toml
```diff
@@ -93,6 +93,7 @@ serde = { version = "1.0.215", default-features = false }
...
 virtfw-varstore = { version = "0.6.1" }
```
luigix25, line 98: ditto with 0.6.2

luigix25 (changes requested):

kernel/src/protocols/uefivars.rs
stefano-garzarella: yep, an assert in this case should be fine.

Documentation/docs/protocols/uefivars.md
```diff
@@ -0,0 +1,70 @@
...
+ * MdePkg/Include/Protocol/MmCommunication.h
+ * MdeModulePkg/Include/Guid/SmmVariableCommon.h
+ * MdeModulePkg/Include/Guid/VarCheckPolicyMmi.h
+
+### linux references
+
+The linux kernel has an efi variable driver implementation using this
+request format in `drivers/firmware/efi/stmm`.  It sends those
+requests to the edk2 MM code running in arm secure partition.
```
stefano-garzarella, line 69: What about specifying from which versions they are available upstream in EDK2 and Linux?

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+    Ok(())
+}
+
+// process uefi mm request in passed buffer
+fn uefi_mm_request(params: &RequestParams) -> Result<(), SvsmReqError> {
+    let addr = params.rcx;
+    let size = params.rdx as usize;
+
+    let mut store = STORE.lock();
```
stefano-garzarella: Do we really need to lock this spinlock from here?

What about this:
```diff
diff --git a/kernel/src/protocols/uefivars.rs b/kernel/src/protocols/uefivars.rs
index 00b54e51..b535c9e6 100644
--- a/kernel/src/protocols/uefivars.rs
+++ b/kernel/src/protocols/uefivars.rs
@@ -69,8 +69,6 @@ fn uefi_mm_request(params: &RequestParams) -> Result<(), SvsmReqError> {
     let addr = params.rcx;
     let size = params.rdx as usize;
 
-    let mut store = STORE.lock();
-
     // check buffer parameters
     check_buffer(addr, size)?;
     log::debug!("uefi mm buffer: 0x{addr:x} +0x{size:x}");
@@ -84,7 +82,7 @@ fn uefi_mm_request(params: &RequestParams) -> Result<(), SvsmReqError> {
     }
 
     let req = read_bytes_from_guest(paddr + boffset, bsize)?;
-    let rsp = core_request_dispatch(&mut store, &mmcore.guid, &req);
+    let rsp = core_request_dispatch(&mut STORE.lock(), &mmcore.guid, &req);
     if rsp.len() > bsize {
         // should not happen (add SvsmReqError::internal_error() ?)
         log::debug!("uefi mm: response buffer too big");
```

kernel/src/protocols/uefivars.rs
```diff
@@ -121,3 +126,41 @@ pub fn uefi_mm_protocol_init() -> Result<(), SvsmReqError> {
...
+    // This should only be set in case secure boot is enabled.
+    pub secureboot_db_update: bool,
+
+    #[bits(29)]
+    _reserved: u32,
+}
+
+#[allow(dead_code)]
+#[derive(IntoBytes, Immutable)]
+struct UefiMmManifest {
```
stefano-garzarella, line 142: IIUC this will be parsed by edk2, so should we use `#[repr(C, packed)]` ?

kernel/src/protocols/attest.rs
```diff
@@ -542,6 +546,10 @@ fn attest_multiple_services(params: &mut RequestParams) -> Result<(), SvsmReqErr
 
     #[cfg(all(feature = "vtpm", not(test)))]
     services.push(SVSM_ATTEST_VTPM_GUID, vtpm_get_manifest()?);
+
+    #[cfg(all(feature = "uefivars", not(test)))]
+    services.push(SVSM_ATTEST_UEFI_MM_GUID, uefi_mm_get_manifest()?);
```
stefano-garzarella, line 554: Should we add something also in `attest_single_service_handler()`?

configs/qemu-target.json
stefano-garzarella, line 1: > [testing] uefivars: turn on secure boot
> Flip switch.
> 
> RfC: What should be the default for the secureboot feature?

Do we need to wait for the persistent storage or this can still be useful right now?

If it's already useful maybe we can discuss with the TSC what they prefer for now.

kernel/Cargo.toml
stefano-garzarella: +1 on this.

Also I'm not sure if it makes sense to add a prefix like `uefivars-` ? Maybe no.

Documentation/docs/protocols/uefivars.md
kraxel: Does it make sense to have a new `Documentation/docs/protocol` directory?

uefivars.md
kraxel: No, it's in the spec and also merged already (commit e0e0e428d04ba247b331b9b4d2ccdac20c95993f).

I'll drop the line.

kernel/src/protocols/uefivars.rs
kraxel: Well, the fields are used, so I wouldn't prefix them with `_`.  They get written, then the whole struct is converted to a byte vector.  They are never read as fields though, which why the rust compiler considers them unused.

kernel/src/protocols/uefivars.rs
kraxel: Don't remember.  Guess I should better double-check the whole section against the latest svsm protocol spec by @tlendacky.

Documentation/docs/protocols/uefivars.md
kraxel: I don't think this matters here.  Note that we have /two/ protocols here, the one is the serialization protocol for the requests (used in various places for years), and the other is the svsm protocol to pass those buffers between edk2 + svsm (which is very simple because it can piggyback on the existing serialization protocol).

The latter is not yet supported by edk2 even though I have a branch.  For the latter it'll make sense to document which edk2 version you'll need once it is merged.

kernel/src/protocols/attest.rs
kraxel: Seems only the vtpm service is supported there right now.  If this makes sense for the uefi vars too (not sure it does) at minimum we'll need a guid assigned in the spec.

configs/qemu-target.json
kraxel: It really depends on the use case.  If you want boot standard distro kernels simply enrolling the standard microsoft certificates will work just fine.  If you want something more strict you'll have to configure svsm at compile time somehow because you can't (yet) boot the guest with a pre-configured variable store on coconfs.

kraxel: Given that `cargo xbuild` has a cmd line switch to turn on features but not to turn off features it is probably most convinient to have it off by default ...

stefano-garzarella: @kraxel thanks for the clarification. So that said, I agree that maybe it's better to turn it off by default and perhaps turn it on for tests (if we have tests running on that part)

Documentation/docs/protocols/uefivars.md
luigix25: i think so!

kernel/src/protocols/attest.rs
stefano-garzarella: In the latest spec (https://docs.amd.com/v/u/en-US/58019.pdf - Revision: 1.01 - Issue Date: March 2026) I see:

> 10.2 Service Attestation Data
> 
> 10.2.1 Service Attestation GUID
> The UEFI Management Mode service attestation GUID is a4453a59-9e1b-4787-a033-1986d6adbe55.
> 
> 10.2.2 SVSM_ATTEST_SERVICES Manifest Data
> The manifest data used in an SVSM_ATTEST_SERVICES call is the same format as the manifest version 0 format of the SVSM_ATTEST_SINGLE_SERVICE call, see section 10.2.3.1.
> 
> 10.2.3 SVSM_ATTEST_SINGLE_SERVICE Manifest Data
> 
> 10.2.3.1 Manifest Version 0
> The manifest data used in an SVSM_ATTEST_SINGLE_SERVICE call for manifest version 0 is
> specified according to Table 45.

kernel/src/protocols/uefivars.rs
luigix25: yeah I see. I don't have a strong opinion though.

kernel/src/protocols/uefivars.rs
kraxel: ok, switching to assert then.

Documentation/docs/protocols/uefivars.md
kraxel: ok.

kernel/src/protocols/uefivars.rs
kraxel: fixed.

kernel/Cargo.toml
kraxel: fixed

Cargo.toml
kraxel: /me checks git log.

No.

Cargo.toml
kraxel: Also no.  No changes for the varstore core code.

kernel/src/protocols/uefivars.rs
kraxel: fixed.

kernel/src/protocols/uefivars.rs
kraxel: specs: https://docs.amd.com/v/u/en-US/58019.pdf

yes, version 0 is correct, and the other bits too.

luigix25: maybe worth a comment?

kernel/src/protocols/uefivars.rs
kraxel: It's not edk2, but yes, the struct layout is ABI.  I think for a struct with two `u32` it doesn't make an actual difference in struct layout, but it is good nevertheless to clarify our intention.

kernel/src/protocols/attest.rs
kraxel:
```
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -583,6 +583,10 @@ fn attest_single_service_handler(params: &mut RequestParams) -> Result<(), SvsmR
     match attest_op.get_guid() {
         SVSM_ATTEST_VTPM_GUID => attest_single_vtpm(params, &attest_op),
+        SVSM_ATTEST_UEFI_MM_GUID => {
+            attest_single_service(uefi_mm_get_manifest()?.as_slice(), params, &attest_op)
+        }
+
         _ => Err(SvsmReqError::unsupported_protocol()),
     }
 }
```
Compiles.  Not fully sure this actually makes sense.

stefano-garzarella: > Compiles.

I guess the final code will need a guard on `uefivars` feature.

> Not fully sure this actually makes sense.

@kraxel What are your concerns?

kernel/src/protocols/uefivars.rs
kraxel: no.

kernel/src/protocols/attest.rs
kraxel: If there is an actual use case for that.

Why does the single-service attestation exist in the first place?

kraxel: Maybe I'm over-thinking this, included the change now.

stefano-garzarella: IIUC was added to better support keylime use case where just the vTPM needed to be attested (see https://lore.kernel.org/linux-coco/6303283f-cf1c-8be6-9359-69d556a89554@amd.com/).

@tlendacky @jejb do you think makes sense to support SVSM_ATTEST_UEFI_MM_GUID also in the SVSM_ATTEST_SINGLE_SERVICE call or it is useless?

That said, I think we should also add the APIC Emulation at least in the multiple services version, but this is another issue.

configs/qemu-target.json
kraxel: dropped from config and turned on for test builds now.

configs/qemu-target.json
```diff
@@ -14,7 +14,7 @@
     },
     "kernel": {
         "svsm": {
-            "features": "vtpm",
+            "features": "vtpm,uefivars",
```
stefano-garzarella, line 17: leftover (from https://github.com/coconut-svsm/svsm/pull/694#discussion_r3380638274) or intentional?

Documentation/docs/protocols/uefivars.md
```diff
@@ -0,0 +1,69 @@
+
+## SVSM UEFI MM PROTOCOL
```
stefano-garzarella, line 2: Should we mention that the `uefivars` feature must be enabled to support this? (somewhere in this file)

kernel/Cargo.toml
```diff
@@ -82,6 +82,7 @@ virtio-drivers = ["dep:virtio-drivers"]
 block = []
 vsock = []
 uefivars = ["dep:virtfw-varstore"]
+secureboot = ["uefivars"]
```
stefano-garzarella, line 86: Should we document this feature too in the new file we are adding the documentation to exaplain what this is supposed to do?

kernel/src/protocols/uefivars.rs
```diff
@@ -0,0 +1,161 @@
...
+    // secure boot databases ('db' + 'dbx') can be updated by the
+    // guest (assuming proper pkcs7 signature of course).
+    // This should only be set in case secure boot is enabled.
+    pub secureboot_db_update: bool,
+
+    #[bits(29)]
+    _reserved: u32,
+}
+
+#[allow(dead_code)]
```
stefano-garzarella:
```suggestion
```

I just tried this change also with `make clippy CARGO_HACK=1` and I didn't see any error/warning. Do we really need it?

stefano-garzarella (commented):
Some minor comments, mainly to improve a bit the documentation around the new added features.

The code LGTM!

configs/qemu-target.json
kraxel: This turns on the variable store support by default but does not turn on secure boot.  Which I think makes sense.

kernel/src/protocols/uefivars.rs
kraxel: Hmm, interesting.  With `#[repr(C,packed)]` added it seems to not be needed any more.

configs/qemu-target.json
stefano-garzarella: @kraxel oh I see, makes sense to me too, maybe we can mention this in the commit.

kernel/src/protocols/uefivars.rs
stefano-garzarella: @kraxel oh, cool, maybe with that it knows that it's an FFI and so fields can be unused in the rust code. Good to know!

Documentation/docs/protocols/uefivars.md
kraxel: I'd tend to think the build instructions in `INSTALLATION.md` instructions would be the better place, especially as the edk2 side needs some care too.

Preparing the edk2 PR right now: https://github.com/tianocore/edk2/pull/12662

/me wants land this in the next (2026-08) edk2 release.

stefano-garzarella: @kraxel yep, `INSTALLATION.md` seems a better place.

It would be nice to have some instructions also on how to test it for us, to check it's working during releases (I don't know, like some tools we can use in Linux to probe secureboot status, etc.)

kraxel: Looking at the edk2 build build docs I guess it makes sense to rework them in a separate PR, they are a bit outdated.  svsm fork is not needed any more, all upstream in tianocore, also the grub/shim mess should be sorted meanwhile and 2026 linux distros should boot just fine without any hacks.  Will look into that.

stefano-garzarella: @kraxel yep a follow up is fine by me. Thanks for that in advance!

kernel/Cargo.toml
stefano-garzarella: So this can go with the follow up as well.

kernel/src/protocols/attest.rs
stefano-garzarella: I guess we can remove this comment now.

stefano-garzarella (dismissed):
LGTM, a minor minor comment left that we can fix only if you need to respin.

kernel/src/protocols/attest.rs
kraxel: done

luigix25 (approved):

stefano-garzarella (approved):

joergroedel (approved):
Really cool! Thanks @kraxel and @stefano-garzarella for bringing this over the finish line.

PR #695, opened 2025-05-13
block: suppress virtio tests on hyper-v

msft-jlange:
Hyper-V doesn't support virtio-blk, so the virtio-blk tests should not be run on that platform.


stefano-garzarella (approved):

osteffenrh (approved):

osteffenrh:
Sorry about that.

PR #696, opened 2025-05-13
CI: treat nightly failures as warnings

msft-jlange:
Failures of the nightly toolchain provide useful information about keeping the build healthy, but those failures should not block PRs from merging.


joergroedel (approved):
LGTM.

PR #698, opened 2025-05-13
Verification: add wrapper for verification

luigix25:
This adds two new crates: `verification` and `verification_derive`.

First one provides a dummy implementation for proc-macros like `#[verus_verify]`, `#[verus_spec]` or `proof!`.
The second one exports either the dummy implementation or the real one, according to the feature(s) set during compilation.

This can be useful, for example, for packaging svsm: there will be no need to package all the dependencies required for the verification part, since it is not necessary.

This is my first PR here, and I'm very open to any comments or suggestions!


stefano-garzarella:
@ziqiaozhou please, can you take a look?

IIUC the main goal is to reduce the dependencies when `verus`/`verus_all` features are not enabled to simplify packaging of svsm in distro.

kernel/Cargo.toml
```diff
@@ -56,7 +54,7 @@ enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 vtpm = ["dep:libtcgtpm"]
 nosmep = []
 nosmap = []
-verus_all = ["builtin", "verify_proof/verus", "verify_external/verus"]
+verus_all = ["verify_proof/verus", "verify_external/verus"]
 verus = ["verus_all", "verify_proof/noverify", "verify_external/noverify"]
```
stefano-garzarella, line 58: Maybe we should move the set of features of dependencies in the new crate and here just enabled the new features you added in the wrapper crate.

stefano-garzarella: Nope, `kernel` is still using those, so they should stay here.

verification_derive/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+#![no_std]
+
+#[cfg(any(feature = "verus", feature = "verus_all"))]
+pub use builtin_macros::*;
+
+#[cfg(any(feature = "verus", feature = "verus_all"))]
+pub use vstd::prelude::*;
+
+#[cfg(not(any(feature = "verus", feature = "verus_all")))]
+pub use verification::*;
```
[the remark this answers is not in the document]
ziqiaozhou: verus_stubs sounds good to me.

kernel/Cargo.toml
ziqiaozhou:
```
builtin = { workspace = true, optional = true}
```

builtin crate should be optionally added since verus macro generated spec/proof code will use builtin.

```suggestion
verus_all = ["builtin", "verify_proof/verus", "verify_external/verus", "verus/enable"]
verus = ["verus_all", "verify_proof/noverify", "verify_external/noverify"]
```

verification_derive/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[dependencies]
+
+builtin_macros = { workspace = true, optional = true }
+vstd = { workspace = true, optional = true }
+verification = { workspace = true }
+
+[features]
+default = []
+verus_all = ["dep:builtin_macros", "dep:vstd"]
+verus = ["verus_all"]
```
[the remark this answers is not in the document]
ziqiaozhou: But I do not think this crate need two features, maybe just a feature called "enable" to enable verus? When enabled, we use real verus macro. Maybe this verifcation_derive can be called verus, another called verus_stub

kernel/src/types.rs
```diff
@@ -7,7 +7,8 @@
 use crate::error::SvsmError;
 use crate::sev::vmsa::VMPL_MAX;
 
-use builtin_macros::*;
+use verification_derive::*;
+#[cfg(any(feature = "verus", feature = "verus_all"))]
```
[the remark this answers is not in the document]
luigix25: will do, thanks!

luigix25:
Thanks everybody for the feedback!

luigix25:
I should have addressed all the comments!

Cargo.lock
stefano-garzarella, line 1: We are updating a lot of dependencies which are not related, can you check this?

kernel/Cargo.toml
```diff
@@ -56,7 +56,7 @@ enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 vtpm = ["dep:libtcgtpm"]
 nosmep = []
 nosmap = []
-verus_all = ["builtin", "verify_proof/verus", "verify_external/verus"]
+verus_all = ["builtin", "vstd", "verify_proof/verus", "verify_external/verus", "verus_stub/enable"]
```
stefano-garzarella, line 59: if we call the new crate `verus_stub`, IMO it is better to call the feature `disable`, so maybe is a bit more clear that we are disabling the stubs. If we call `verus` (which I slightly prefer), I think `enable` is fine.

stefano-garzarella (commented):
Just minor comments, but overall LGTM, thanks!

I'd like an approval from @ziqiaozhou before merge this.

Cargo.lock
luigix25: my IDE made a mess, sorry for that. should be ok now :)

kernel/Cargo.toml
luigix25: verus is also the name of a feature in the `kernel` crate. This is why I opted for verus_stub.
TBH I don't even know if it's allowed to have a crate named like a feature. 

I renamed the feature to `disable`.

ziqiaozhou (approved):
It looks good to me. I may need to make some changes in my unmerged PR to pick this change by defining some stubs. But overall it is a good idea to drop the dependencies from verus. Thanks for making this change!

stefano-garzarella (approved):

PR #701, opened 2025-05-16
uefivars: reserve uefi mm protocol number

kraxel:
This protocol allows svsm receive mm communication requests from edk2,
which in turn allows svsm implement an UEFI variable service.

Signed-off-by: Gerd Hoffmann <kraxel@redhat.com>

### DIFF
--- a/kernel/src/protocols/mod.rs
+++ b/kernel/src/protocols/mod.rs
@@ -21,6 +21,7 @@ pub const SVSM_CORE_PROTOCOL: u32 = 0;
 pub const SVSM_ATTEST_PROTOCOL: u32 = 1;
 pub const SVSM_VTPM_PROTOCOL: u32 = 2;
 pub const SVSM_APIC_PROTOCOL: u32 = 3;
+pub const SVSM_UEFI_MM_PROTOCOL: u32 = 4;
 
 #[derive(Debug, Default, Clone, Copy)]
 pub struct RequestParams {


stefano-garzarella:
I don't know if it's better to discuss this on linux-coco/coconut-svsm mailing lists.
@tlendacky what do you suggest?

deeglaze:
I think a full specification of the protocol to correspond to the protocol number allocation ought to be sent to the list similar to how Jon did with the APIC protocol. Otherwise we have the problem of "the code is the spec" and we end up with bug-for-bug compliance problems.

tlendacky:
> Doing svsm protocol calls from uefi runtime code seems to be a non-trivial challenge though because linux relocates the CAA page (see also https://mail.8bytes.org/pipermail/svsm-devel/2025-May/000572.html).

See the SVSM spec, section 4, Discovery (most specifically 4.2 Post-Boot), where you'll be able to determine if an SVSM is active and obtain the current CAA.

stefano-garzarella (approved):
In the `[SVSM draft specification (v1.01 draft #3)](https://lore.kernel.org/coconut-svsm/39cc8435-5643-4a16-8eb5-5e12f15566a1@amd.com/) sent by @tlendacky we have: 

> The UEFI Management Mode protocol has protocol ID value 4.

So I guess we can now merge this.

joergroedel (approved):

PR #703, opened 2025-05-19
kernel: remove manual UUID implementation

00xc:
Since the uuid crate is already being used in the SVSM kernel and other packages in the workspace, rely on it instead of the in-kernel implementation. While we are at it, centralize GUIDs into the bootlib crate, so that the kernel and igvmbuilder can access the same definitions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,6 +189,7 @@ dependencies = [
 name = "bootlib"
 version = "0.1.0"
 dependencies = [
+ "uuid",
  "zerocopy 0.8.10",
 ]
 
--- a/bootlib/Cargo.toml
+++ b/bootlib/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+uuid.workspace = true
 zerocopy.workspace = true
 
 [lints]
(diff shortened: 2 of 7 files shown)


stefano-garzarella:
> Since the uuid crate is already being used in the SVSM kernel and other packages in the workspace, rely on it instead of the in-kernel implementation. While we are at it, centralize GUIDs into the bootlib crate, so that the kernel and igvmbuilder can access the same definitions.

Thanks, I like this, but can we split in 2 patches (one to remove our implementation and one to centralize GUIDs)?

osteffenrh (approved):

osteffenrh:
Boots my Linux guest on SNP :heavy_check_mark:

stefano-garzarella:
We discussed a bit the idea of replacing our `uuid` implementation here: https://github.com/coconut-svsm/svsm/pull/541#discussion_r1890526628

I'm going to approve this, but I'd like @joergroedel to take a second look since I don't know why we introduced our implementation.

00xc:
> I'm going to approve this, but I'd like @joergroedel to take a second look since I don't know why we introduced our implementation.

As far as I remember, the original implementation came in before we pulled the uuid crate, which is more convenient, so it was just a matter of replacing the old implementation.

stefano-garzarella:
> > I'm going to approve this, but I'd like @joergroedel to take a second look since I don't know why we introduced our implementation.
> 
> As far as I remember, the original implementation came in before we pulled the uuid crate, which is more convenient, so it was just a matter of replacing the old leftover implementation.

Ah, I see! Okay since tests are fine (I run locally `test-in-svsm`), @osteffenrh approved, and CI is happy, I think we can merge this, so less work for @joergroedel when he will be back. If we find something wrong, we can revert/fix later.

PR #704, opened 2025-05-22
Makefile: allow building with empty features

00xc:
Currently there is no easy way to build the SVSM from the command line with only the default features. The Makefile will build the SVSM with the `vtpm` feature unless `FEATURES` is overriden. However, overriding `FEATURES` to an empty value (to avoid building the `vtpm` feature) will result in passing the `--features` flag to cargo without any actual features. Thus, pass only this flag when there are features explicitly enabled. Also, do the equivalent for `FEATURES_TEST`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,13 @@
 FEATURES ?= vtpm
+ifneq ($(FEATURES),)
 SVSM_ARGS += --features ${FEATURES}
+endif
 
 FEATURES_TEST ?= virtio-drivers
-SVSM_ARGS_TEST += --no-default-features --features ${FEATURES_TEST}
+SVSM_ARGS_TEST += --no-default-features
+ifneq ($(FEATURES_TEST),)
+SVSM_ARGS_TEST += --features ${FEATURES_TEST}
+endif
 
 ifdef RELEASE
 TARGET_PATH=release


Makefile
```diff
@@ -1,8 +1,13 @@
 FEATURES ?= vtpm
+ifneq ($(FEATURES),)
 SVSM_ARGS += --features ${FEATURES}
+endif
```
stefano-garzarella, line 4: Since we are touching it, I was thinking on something similar, but supporting also `--no-default-feature`. I mean when FEATURES is empty, we can build without defaults.

I mean something like we have for tests:
```suggestion
FEATURES ?= default,vtpm
SVSM_ARGS += --no-default-features
ifneq ($(FEATURES),)
SVSM_ARGS += --features ${FEATURES}
endif
```

WDYT?

stefano-garzarella (commented):
At least for FEATURES, I used `FEATURES=default make` for that, but I thought about this change too. Thanks for this!

Makefile
00xc: I think as it is the user can already do this by setting `SVSM_ARGS=--no-default-features` from the command line, which is more intuitive, as default features should be opt-out.

stefano-garzarella: yep, right, let's not overcomplicate it.

stefano-garzarella (approved):

PR #707, opened 2025-05-23
mm/page_visibility: `SharedBox` improvements

00xc:
* Fix `SharedBox` trait bounds so that methods are truly safe
* Attempt to restore page visibility on failure during `SharedBox::try_new_zeroed()`.
* Remove the last user of the low level `make_page_shared()` and `make_page_private()` functions, and then make them private to the `page_visibility` module.

### DIFF
--- a/kernel/src/mm/page_visibility.rs
+++ b/kernel/src/mm/page_visibility.rs
@@ -36,7 +36,7 @@ use zerocopy::{FromBytes, FromZeros, Immutable, IntoBytes};
 /// Notably any objects at `vaddr` must tolerate unsynchronized writes of any
 /// bit pattern.  In addition, the caller must take responsibility for
 /// returning a page to the private state if it is ever freed.
-pub unsafe fn make_page_shared(vaddr: VirtAddr) -> Result<(), SvsmError> {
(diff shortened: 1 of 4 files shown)


kernel/src/mm/page_visibility.rs
```diff
@@ -114,14 +114,27 @@ impl<T> SharedBox<T> {
...
         for offset in (0..core::mem::size_of::<T>()).step_by(PAGE_SIZE) {
-            unsafe {
-                make_page_shared(vaddr + offset)?;
+            let r1 = unsafe { make_page_shared(vaddr + offset) };
+            if let Err(e1) = r1 {
+                for off in (0..offset).step_by(PAGE_SIZE) {
+                    // SAFETY: we previously marked this page as shared in this same function.
+                    let r2 = unsafe { make_page_private(vaddr + off) };
+                    if let Err(e2) = r2 {
+                        log::error!("Failed to restore page visibility (e2:?), leaking memory!");
```
msft-jlange: This should be grounds for a panic, as the error cannot be handled gracefully in any meaningful way.

00xc: Okay, I changed the log and return for a panic

PR #708, opened 2025-05-24
utils: remove duplicated `copy_bytes()` implementation

00xc:
`unsafe_copy_bytes()` does the exact same thing as `copy_bytes()`, so remove it. [Rust guarantees that the direction flag on entry to an assembly block will be cleared](https://doc.rust-lang.org/reference/inline-assembly.html#r-asm.rules.x86-df), so the `cld` instruction is not necessary.


msft-jlange:
Can we keep the name `unsafe_copy_bytes`?  It is significant here because the Hyper-V code is copying data to/from unsynchronized memory which means standard memory access soundness guarantees (like pointer access soundness) do not apply.  I think it is important that callers of this function know that they are getting an implementation that expressly addresses the fact that the underlying memory behavior is unsound (as if it were volatile).  The name `copy_bytes` suggests that the implementation could be anything that causes bytes to be copied, such as `pointer::copy_to()` which would definitely be unsound for the type of memory being copied here.

msft-jlange (approved):

PR #709, opened 2025-05-26
Documentation: update the mailing list address

stefano-garzarella:
We are shutting down the old mailing list, update all references to the mailing list in our documentation.


peterfang (approved):
Thanks for the housekeeping!

PR #710, opened 2025-05-28
Launcher improvements

osteffenrh:
Here are some little convenience improvements for the QEMU launcher script.

1) Pass-through of arbitrary arguments to the QEMU call. This is convenient for
   debugging, for example.
2) Don't set the boot device explicitly. This allows booting from other devices
   for testing/experiments.
3) Automatically build the cbit util, if it does not exist.
   (Very handy, because I *always* forget to build it!)


00xc (approved):

joergroedel (approved):
LGTM.

PR #711, opened 2025-05-29
kernel/attest: validate AttestServicesOp and AttestSingleServiceOp

stefano-garzarella:
Commit 0ae2ab71 ("Add attest_multiple_services") used `read_from_guest` to create an instance of `AttestServicesOp` or `AttestSingleServiceOp`, but we didn't validate them like we do in `try_from_as_ref()`.

So let's provide for both a new `try_from_guest()` method where we can validate them. Probably in the future we should provide a trait (e.g. `FromGuest`) that each structure can implement in order to standardize our code.

Fixes: 0ae2ab71 ("Add attest_multiple_services")

### DIFF
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -56,6 +56,17 @@ pub struct AttestServicesOp {
 }
 
 impl AttestServicesOp {
+    /// Take a PhysAddr and return Self
+    pub fn try_from_guest(paddr: PhysAddr) -> Result<Self, SvsmReqError> {
+        let ops = read_from_guest::<Self>(paddr).map_err(|_| SvsmReqError::invalid_parameter())?;
+
+        if !ops.is_reserved_clear() {
+            return Err(SvsmReqError::invalid_parameter());
+        }
+
+        Ok(ops)
+    }
+
     /// Take a slice and return a reference for Self
     pub fn try_from_as_ref(buffer: &[u8]) -> Result<&Self, SvsmReqError> {
         let ops: &Self =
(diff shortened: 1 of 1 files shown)


kernel/src/protocols/attest.rs
```diff
@@ -56,6 +56,17 @@ pub struct AttestServicesOp {
 }
 
 impl AttestServicesOp {
+    /// Take a PhysAddr and return Self
+    pub fn try_from_guest(paddr: PhysAddr) -> Result<Self, SvsmReqError> {
+        let ops = read_from_guest::<Self>(paddr).map_err(|_| SvsmReqError::invalid_parameter())?;
+
+        if !ops.is_reserved_clear() {
+            return Err(SvsmReqError::invalid_parameter());
+        }
```
deeglaze, line 65: Do we not also want to check the validity requirements of the nonce and manifest?
We can then say that there's the PreAttestServicesOp type that is structurally all the fields to read, and try_from_guest returns AttestServicesOp which is ensured to have valid fields. That allows all the get_ methods to not need to do any wellformedness checks.

deeglaze: I think even moreso here and there. both TpmSendCommandRequest and AttestServicesOp ought to have their fields in TpmSendCommandRequestState/AttestServicesOpState structs that get embedded in TpmSendCommandRequest/AttestServicesOp only if their values pass validation. This ensures that operations are only associated with the struct type that maintains the wellformedness invariant.

stefano-garzarella: If we make fields private and provide only `try_from*` method to build those structures, we should have the same, without embending them, no?
Or maybe I'm missing something.

stefano-garzarella: > If we make fields private and provide only `try_from*` method to build those structures, we should have the same, without embending them, no? Or maybe I'm missing something.

@deeglaze WDYT?

stefano-garzarella:
This issue was fixed by https://github.com/coconut-svsm/svsm/pull/1073 so I'm closing this and eventually open a new one to introduce `try_from_guest()` which can avoid this issue at all since checked at creation time.

PR #685, opened 2025-04-30
libtcgtpm/build.rs: Use include files of libcrt for bindgen

joergroedel:
Explicitly supply the include directory of libcrt to the bindgen call in build.rs to make sure it always finds `stddef.h`.

PR #713, opened 2025-05-29
Re-enable `vtpm` feature on tests

stefano-garzarella:
I'd like to add tests for the vtpm emulation, so I'm trying to re-enable the feature disabled by #685

Commit c2e95ca9 ("libtcgtpm/build.rs: Use include files of libcrt for bindgen") added `-isystemdeps/libcrt/include/` clang arg to the bindgen call. This though produces an empty `bindings.rs` in our CI (ubuntu-latest), thus leading `make test` to fail by not finding symbols as `TPM_Manufacture`, etc.

This happens because in libcrt.h we have this on top:
```
    #if defined(__pie__)
    #pragma GCC visibility push (hidden)
    #endif
```

and probably on Ubuntu `-fno-pie` is not enabled by default.

So let's invoke clang always with `-fno-pie` to find all the symbols to generate the bindings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ ifneq ($(FEATURES),)
 SVSM_ARGS += --features ${FEATURES}
 endif
 
-FEATURES_TEST ?= virtio-drivers
+FEATURES_TEST ?= vtpm,virtio-drivers
 SVSM_ARGS_TEST += --no-default-features
 ifneq ($(FEATURES_TEST),)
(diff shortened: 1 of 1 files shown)


deeglaze:
> not giving precedence to the system libraries

That seems counter to what the flag actually does though. Are we sure that the relative path is correct? Add a #include <nonstandard.h> that you create in this libcrt include directory and see if it's found.

deeglaze (approved):

PR #715, opened 2025-06-03
Make exception context 16b-aligned

peterfang:
It is generally a good practice to have exception frames aligned on a 16-byte boundary. Currently, X86ExceptionContext is off by 8 bytes which makes it difficult to ensure all following call stacks are 16-byte-aligned.

Remove the padding between ssp and the remaining fields to form a naturally 16-byte-aligned exception context. Make the corresponding changes in both Stage2 and SVSM to follow this new layout. This actually fixes Stage2's exception handling which was previously broken because it didn't push enough data into the stack to form a correct exception context.

This is the precursor to adding stacktrace support to Stage2. This PR also includes some clean-ups.


joergroedel (commented):
Testing looks good to me.

@msft-jlange Can you please test this on Hyper-V as well?

kernel/src/cpu/idt/svsm_entry.S
```diff
@@ -203,7 +203,15 @@ restart_hv:
...
 	// instruction.
-	testl	${IF}, 0x10(%rcx)
+	//
+	// At this point, RCX holds the stack pointer at the time of the
+	// IRET that was aborted.  In other words, it is pointing to
+	// &X86ExceptionContext.frame.  Rewind RCX back to pointing at the
+	// beginning of X86ExceptionContext to facilitate asm-const usages.
+	// This is essentially the reverse of the
+	// "addq ${EXCEP_FRAME_OFF}, %rsp" operation right before IRET.
+	subq	${EXCEP_FRAME_OFF}, %rcx
```
msft-jlange: This makes me nervous.  Although the code is correct, the modification of rcx in this way implies that the data above rcx is valid.  However, after this subtraction, rcx points below the stack pointer, and thus points to data that is not stable.  The offset you use below (`EXCEP_FLAGS_OFF`) is above the current stack pointer, but adjusting rcx below rsp in this way invites problems in the future - not least because the use of EXCEP_* implies that `rcx` points at a complete and intact exception context, which it most certainly does not.

Can we keep this like the old code, where we just execute `testl ${IF}, {EXCEP_FLAGS_OFF - EXCEPT_FRAME_OFF}, %rcx`?

kernel/src/cpu/idt/svsm_entry.S
```diff
@@ -216,48 +224,49 @@ restart_hv:
...
-	// in the top three QWORDs of the current stack.
+	//
+	// The QWORDs that have been pushed into the current stack are the
+	// dummy error code and the RAX, RBX, and RCX values.  Copy them back
+	// to the previous exception context and start reconstructing (or
+	// "taking over") the exception context from which the IRET intends to
+	// return.
 	movq	0*8(%rsp), %rax
-	movq	%rax, -4*8(%rcx)
+	movq	%rax, {EXCEP_RCX_OFF}(%rcx)
```
msft-jlange: This will have to changed based on the previous comment about needing to remove the adjustment of rcx.  While you could apply the same math here (`EXCEPT_RCX_OFF - EXCEPT_FRAME_OFF`), I think the comment is quite self-explanatory: it doesn't matter what meaning applies to the three QWORDs being moved, only that there are three that need to be moved intact so the rest of the exception frame can be rebuilt.  For this reason, I don't think symbolic constants offer much value here; the old code with its constant manipulation is just fine.

kernel/src/cpu/idt/svsm_entry.S
```diff
@@ -216,48 +224,49 @@ restart_hv:
...
 	movq	1*8(%rsp), %rax
-	movq	%rax, -3*8(%rcx)
+	movq	%rax, {EXCEP_RBX_OFF}(%rcx)
 	movq	2*8(%rsp), %rax
-	movq	%rax, -2*8(%rcx)
-	leaq	-4*8(%rcx), %rsp
+	movq	%rax, {EXCEP_RAX_OFF}(%rcx)
+	// Load the previous stack pointer and continue reconstruction at RDX.
+	leaq	{EXCEP_RCX_OFF}(%rcx), %rsp
```
msft-jlange: By the same reasoning as above, I don't think symbolic constants are needed here.  It is sufficient to make rsp point at the last of the three QWORDs that were copied, which can be determined from the constant value.  Thus, I would return to the original code here.

msft-jlange (changes requested):
Please make the requested changes to the #HV handler, or explain why you disagree with my comments.  The goal here should be to avoid any changes to the generated code, only to change the symbolic constants used to generate it.

kernel/src/cpu/idt/svsm_entry.S
peterfang: Your concern is very valid. I can make that change.

peterfang:
> Please make the requested changes to the #HV handler, or explain why you disagree with my comments. The goal here should be to avoid any changes to the generated code, only to change the symbolic constants used to generate it.

I have addressed the concerns. Thanks for reviewing.

PR #717, opened 2025-06-04
cpu/mem: take pointers in `unsafe_copy_bytes()`

00xc:
Make the function signature clearer by taking generic pointers. This also simplifies pointer arithmetic at the callers, as the size parameter now indicates the number of elements of size `T` to be copied.


kernel/src/cpu/mem.rs
```diff
@@ -29,7 +30,8 @@ pub unsafe fn unsafe_copy_bytes(src: usize, dst: usize, size: usize) {
 /// This function has all the safety requirements of `core::ptr::write_bytes` except
 /// that data races are explicitly permitted.
 #[inline(always)]
-pub unsafe fn write_bytes(dst: usize, size: usize, value: u8) {
+pub unsafe fn write_bytes<T>(dst: *mut T, count: usize, value: u8) {
```
msft-jlange, line 33: It seems strange to me that the target pointer is of type `T` but the value is of type `u8`.  Shouldn't these be consistent?  If `T` is a multi-byte type, then blasting it with repeated copies of a single byte doesn't have useful meaning (unless the value is zero, in which case this could just become `zero_bytes()`).

00xc: Well if the value was `T` then there is no point to this function vs `unsafe_copy_bytes`. So yes, this is perhaps too generic for its purpose, but there are users like `PageRef::fill()` that would need to be changed if we updated this to `zero_bytes()`, which would feel more intrusive. I'd say let's just keep it, even if for the time being we're only filling the bytes with zero.

00xc: The point of this PR is to avoid having `size_of::<T>()` on every caller to reduce the chance of someone passing in the wrong size.

msft-jlange: > The point of this PR is to avoid having `size_of::<T>()` on every caller to reduce the chance of someone passing in the wrong size.

That's great, but that appears to impact the change from `size` to `count` and not the use of the fill value.  If `T` is a multi-byte type, then how can a single byte fill value ever be meaningful, unless filling with zeroes?  Zeroing a structure makes lots of sense, but filling with arbitrary fill value - which is unlikely to have any consistent meaning in a multi-byte structure - doesn't seem to be conducive to sound behavior.  If all of the callers of this function just use it to zero data, then can we just omit `value` altogether and turn this into `zero_bytes()`?

> Well if the value was `T` then there is no point to this function vs `unsafe_copy_bytes`.

I think that's only true for a count of 1.  If the count is greater than one, then this function can be used to stamp out an entire array from a single template element, which `unsafe_copy_bytes()` cannot do.  That said, I don't think there's much call for an unsafe function that can populate an entire array from a template.  Unsafe zeroing seems far more useful.

00xc: > That's great, but that appears to impact the change from size to count and not the use of the fill value. If T is a multi-byte type, then how can a single byte fill value ever be meaningful, unless filling with zeroes?

You can still use the function as before by passing a `*mut u8`, in which case you might want to fill with a different value.

msft-jlange: > You can still use the function as before by passing a `*mut u8`, in which case you might want to fill with a different value.

Fair enough.  I still think it's odd but you've convinced me that this is valuable and that there's no more practical way to do it.

msft-jlange (approved):

PR #721, opened 2025-06-05
Add a small TPM software stack for commands to populate NV

deeglaze:
The EK template population logic populates the "low range" NV indices for the EK credential profile 2.0 templates and a couple more in the Google Component OEM range if the google-vtpm-ek feature is enabled.

The TPM initialization sequence adds TPM_CC_STARTUP with TPM_SU_CLEAR to allow for the NV population commands to complete. SVSM still expects for the next stage of boot to perform the selftests.

The ek_templates::populate step runs three commands for NV index:

1. NV_UndefineSpace to ensure the index is clear. Failure is ignored.
2. NV_DefineSpace to assign attributes, empty authorization, and size.
3. NV_Write to populate initial contents of the NV index.

A Rand trait is added to the crypto crate to make implementation substitution simpler.

The templates here are specifically for Google Component OEM range NV indices, and they follow the exact template used in Compute Engine.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -29,6 +29,7 @@ bitfield-struct.workspace = true
 bitflags.workspace = true
 gdbstub = { workspace = true, optional = true }
 gdbstub_arch = { workspace = true, optional = true }
+getrandom = { workspace = true, optional = true }
 igvm_defs = { workspace = true, features = ["unstable"] }
 intrusive-collections.workspace = true
 log = { workspace = true, features = ["max_level_info", "release_max_level_info"] }
@@ -54,6 +55,7 @@ test.workspace = true
 default = []
 enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 vtpm = ["dep:libtcgtpm"]
+google-vtpm-ek = ["dep:libtcgtpm", "dep:getrandom", "vtpm"]
 nosmep = []
 nosmap = []
 verus_all = ["builtin", "vstd", "verify_proof/verus", "verify_external/verus", "verus_stub/disable"]
--- a/kernel/src/vtpm/tcgtpm/ek_templates.rs
+++ b/kernel/src/vtpm/tcgtpm/ek_templates.rs
@@ -5,23 +5,52 @@
 // Author: Dionna Glaze <dionnaglaze@google.com>
 //
 
+extern crate alloc;
+
+use crate::crypto::rand::Rand;
+#[cfg(feature = "google-vtpm-ek")]
+use crate::crypto::rand::Reader as RandReader;
+use alloc::vec::Vec;
+
+use crate::protocols::errors::SvsmReqError;
+use crate::vtpm::{
+    tcgtpm::{tss, SvsmVTpmError},
+    TcgTpmSimulatorInterface,
+};
+
 // All constants are defined in TCG Trusted Platform Module Library Part 2 - Structures
 // with the names stated in comments.
 //
 // Table 9. definition of TPM_ALG_ID constants:
+// TPM_ALG_CFB     (0x0043)
+// TPM_ALG_ECC     (0x0023)
+// TPM_ALG_ECDSA   (0x0018)
 // TPM_ALG_RSA     (0x0001)
-// TPM_ALG_AES     (0x0006)
+// TPM_ALG_RSASSA  (0x0014)
 // TPM_ALG_SHA256  (0x000B)
 // TPM_ALG_NULL    (0x0010)
-// TPM_ALG_CFB     (0x0043)
 //
 // Table 33. TPMA_OBJECT bits
 // fixedTPM(1)
 // fixedParent(4)
 // sensitiveDataOrigin(5)
+// userWithAuth(6)
 // adminWithPolicy(7)
 // restricted(16)
 // decrypt(17)
+// sign/encrypt(18)
+//
+// Table 225. TPMA_NV bits
+// TPMA_NV_PPWRITE(0)
+// TPMA_NV_WRITEDEFINE(13)
+// TPMA_NV_PPREAD(16)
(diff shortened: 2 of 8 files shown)


kernel/Cargo.toml
```diff
@@ -54,6 +55,7 @@ test.workspace = true
 default = []
 enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 vtpm = ["dep:libtcgtpm"]
+google-vtpm-ek = ["dep:libtcgtpm", "dep:getrandom", "vtpm"]
```
stefano-garzarella, line 58: Instead of static configuration, would be better to have some kind of runtime configuration (e.g. through IGVM or something else?)

kernel/src/vtpm/tcgtpm/ek_templates.rs
```diff
@@ -59,3 +88,204 @@ pub const DEFAULT_PUBLIC_AREA: [u8; 314] = [
...
+    0x00, 0x10, //  symmetric { algorithm TPM_ALG_NULL
+    // keybits skipped,
+    // mode skipped }
+    0x00, 0x18, 0x00, 0x0B, // scheme { ecc scheme TPM_ALG_ECDSA, asym scheme TPM_ALG_SHA256}
+    0x00, 0x03, // curveID { TPM_ECC_NIST_P256 }
+    0x00, 0x10, // kdf { scheme TPM_ALG_NULL, details skipped }
+          // } ECC parameters
+];
+
+#[cfg(feature = "google-vtpm-ek")]
```
stefano-garzarella, line 152: Can we put all of this function in a `google-vtpm` module?

kernel/src/vtpm/tcgtpm/ek_templates.rs
```diff
@@ -59,3 +88,204 @@ pub const DEFAULT_PUBLIC_AREA: [u8; 314] = [
...
+    tss::write_nv(vtpm, nvindex, public_area)?;
+    Ok(())
+}
+
+/// Initializes contents for NV indices that are "manufacturer populated" for EK templates.
+///
+/// Arguments:
+///
+/// * `vtpm`: An implementation of [`TcgTpmSimulatorInterface`] to send commands to.
+#[allow(unused)]
```
stefano-garzarella, line 231: Can we use the feature flag also here?

kernel/src/vtpm/tcgtpm/ek_templates.rs
```diff
@@ -59,3 +88,204 @@ pub const DEFAULT_PUBLIC_AREA: [u8; 314] = [
...
+            0x01c1_0003,
+            google_ecc_signing_template(rand)?.as_slice(),
+        )?;
+    }
+    #[cfg(not(feature = "google-vtpm-ek"))]
+    let _ = rand;
+    Ok(())
+}
+
+#[allow(dead_code)]
```
stefano-garzarella, line 259: Ditto.

kernel/src/vtpm/tcgtpm/ek_templates.rs
```diff
@@ -59,3 +88,204 @@ pub const DEFAULT_PUBLIC_AREA: [u8; 314] = [
...
+    {
+        let mut rand = RandReader {};
+        return populate(vtpm, &mut rand);
+    }
+    // Randomness is not used outside of google-vtpm-ek.
+    #[cfg(not(feature = "google-vtpm-ek"))]
+    {
+        let mut nop = Nonrandom {};
+        return populate(vtpm, &mut nop);
+    }
```
stefano-garzarella, line 280: Can we do a bit more rusty code here, e.g. using a trait and struct/enum to wrap the tpm around the type (e.g. google, default, etc.)?

kernel/Cargo.toml
IT302: I like the idea of introducing runtime configuration for the vTPM and possibly other components. This approach would enable OS distributions, for example, to provide a single binary that supports a much broader range of use cases, simplifying distribution and deployment. Ideally, a service call manifest should be implemented to attest to all runtime configurations.

deeglaze: I didn't want to introduce Google-only bloat to the build without an opt-out. Can we leave a way to dynamically enable/disable the feature when built in as a follow-up?

kernel/src/vtpm/tcgtpm/ek_templates.rs
deeglaze: The default templates are set in this function as well though.

kernel/src/vtpm/tcgtpm/ek_templates.rs
deeglaze: If we use a trait for the SpinLocked statically allocated VTPM, then we get into a world of hurt with the type checker. Can you say more about what it is you're looking for?

kernel/src/vtpm/tcgtpm/ek_templates.rs
stefano-garzarella: Why we need `#[allow(unused)]` ?

kernel/src/vtpm/tcgtpm/ek_templates.rs
stefano-garzarella: I'm looking for avoiding using `#[cfg(feature = "google-vtpm-ek")]` and `#[cfg(not(feature = "google-vtpm-ek"))]` all over the place. That will be hard to maintain and review.
IMO, if we really need a cfg, we should keep the usage in very few places.

kernel/Cargo.toml
stefano-garzarella: okay, but I'd like to make it clear that it's temporary solution, maybe we should use a prefix "temp_" or "unstable_" and add a TODO/FIXME here to explain next steps.

PR #722, opened 2025-06-07
hyperv: simplify HypercallPagesGuard

00xc:
The virtual-physical address pairs for the hypercall pages do not need to be stored twice, as the given `RefMut` can be split into two individual `RefMut`s, one for each pair.

### DIFF
--- a/kernel/src/hyperv/hv.rs
+++ b/kernel/src/hyperv/hv.rs
@@ -134,12 +134,8 @@ impl<T> HypercallOutput<T> {
 
 #[derive(Debug)]
 pub struct HypercallPagesGuard<'a> {
-    // The page reference is never actually read; it exists here to bind the
-    // lifetime of the `Ref` to the lifetime of the `HypercallPagesGuard`.
-    // The input/output pages are unwrapped during construction.
-    _page_ref: RefMut<'a, (VirtPhysPair, VirtPhysPair)>,
-    pub input: VirtPhysPair,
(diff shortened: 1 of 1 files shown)


kernel/src/hyperv/hv.rs
```diff
@@ -157,9 +153,8 @@ impl<'a> HypercallPagesGuard<'a> {
     /// responsible for ensuring that both virtual and physical addresses are
     /// correct and usable.
     pub unsafe fn new(page_ref: RefMut<'a, (VirtPhysPair, VirtPhysPair)>) -> Self {
-        let (input, output) = *page_ref;
+        let (input, output) = RefMut::map_split(page_ref, |r| (&mut r.0, &mut r.1));
```
msft-jlange, line 156: Can we change this to `(&mut r.input, &mut r.output)` to be more resilient to possible reordering of structure elements in the future?

00xc: That would require adding a new structure with those fields, which I don't know if is worth it just for one function.

msft-jlange: > That would require adding a new structure with those fields, which I don't know if is worth it just for one function.

I see - I missed that this was operating on the tuple, not the `HypercallPagesGuard` itself.  In that case, there is no structure ordering to worry about.

msft-jlange (approved):

PR #725, opened 2025-06-19
Support Kernel threads in COCONUT-SVSM

joergroedel:
Make the required refactoring of `struct Task` to support multiple threads for one task. Build the code on-top to create and launch kernel threads and use them for the request-loop.

This PR only contains the bare minimum to support kernel threads, supporting user-threads needs more additions like tracking the relationship between tasks and their threads and better exit-code handling.


kernel/src/cpu/shadow_stack.rs
```diff
@@ -78,3 +90,91 @@ bitflags! {
...
+    ///
+    /// This won't place any tokens on the shadow stack.
+    Init,
+    /// A shadow stack to be used during normal execution of a task.
+    ///
+    /// This will create a shadow stack with a shadow stack restore token.
+    Normal {
+        /// The address of the first instruction that will be executed by the task.
+        entry_return: usize,
+        /// The address of the fucntion that's executed when the task exits.
```
deeglaze:
```suggestion
        /// The address of the function that's executed when the task exits.
```

kernel/src/mm/vm/mapping/kernel_stack.rs
```diff
@@ -66,11 +69,12 @@ impl VMKernelStack {
     /// # Arguments
     ///
     /// * `size` - Size of the kernel stack, without guard pages
+    /// * `shadow` - Whether this mapping is for a shadow stack
     ///
     /// # Returns
     ///
     /// Initialized stack on success, Err(SvsmError::Mem) on error
-    pub fn new_size(size: usize) -> Result<Self, SvsmError> {
+    pub fn new_size(size: usize, shadow: bool) -> Result<Self, SvsmError> {
```
deeglaze: Style-wise, I think that booleans in public APIs are best avoided. I'd prefer for this twiddly bit to be in a private method and set to true in new_size_shadow() the way you have a new_shadow with size=PAGE_SIZE below.

kernel/src/mm/vm/mapping/kernel_stack.rs
```diff
@@ -92,11 +97,37 @@ impl VMKernelStack {
...
+    ///
+    /// The function panics if called on a non-shadow-stack object.
+    pub fn shadow_page(&self) -> PageRef {
+        assert!(self.shadow);
+        self.alloc.page(0)
     }
 
     /// Create a new [`VMKernelStack`] with the default size, packed into a
-    /// [`Mapping`]. This function / will already allocate the backing pages for
+    /// [`Mapping`]. This function will already allocate the backing pages for
```
deeglaze: "Will already" is weird. It will allocate, or it will expect the backing pages to already be allocated?

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+    _ktask_region: TaskVirtualRegionGuard,
+
+    /// Task virtual memory range for use at CPL 0
+    vm_kernel_range: VMR,
+
+    /// Task virtual memory range for use at CPL 3 - None for kernel tasks
+    vm_user_range: Option<VMR>,
+}
+
+// SAFETY: This is safe because all members of the struct are `Send + Sync`.
```
deeglaze: If this were true, wouldn't it be automatically derived for TaskMM?

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+unsafe impl Sync for TaskMM {}
+
+impl TaskMM {
+    pub fn create(user_vmr: Option<VMR>) -> Result<Self, SvsmError> {
```
deeglaze, line 76: Public methods should be documented.

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        Ok(TaskMM {
+            _ktask_region: ktask_region,
+            vm_kernel_range,
+            vm_user_range: user_vmr,
+        })
+    }
+
+    pub fn kernel_range(&self) -> &VMR {
```
deeglaze, line 103: Public methods should be documented.

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+        &self.vm_kernel_range
+    }
+
+    pub fn user_range(&self) -> Option<&VMR> {
```
deeglaze, line 112: Public methods should be documented.

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+        self.vm_user_range.as_ref()
+    }
+
+    pub fn has_user(&self) -> bool {
+        self.vm_user_range.is_some()
+    }
+}
+
+pub struct TaskKernelMapping {
```
deeglaze, line 127: Public structs should be documented.

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    pub fn kernel_range(&self) -> &VMR {
+        &self.vm_kernel_range
+    }
+
+    pub fn user_range(&self) -> Option<&VMR> {
+        self.vm_user_range.as_ref()
+    }
+
+    pub fn has_user(&self) -> bool {
```
deeglaze, line 121: Public methods should be documented.

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+pub struct TaskKernelMapping {
+    mm: Arc<TaskMM>,
+    va: VirtAddr,
+}
+
+impl TaskKernelMapping {
+    pub fn new(mm: Arc<TaskMM>, mapping: Arc<Mapping>) -> Result<Self, SvsmError> {
```
deeglaze, line 146: Public methods should be documented.

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,127 @@
...
+        let va = mm.kernel_range().insert(mapping)?;
+        Ok(Self { mm, va })
+    }
+
+    pub fn virt_addr(&self) -> VirtAddr {
```
deeglaze, line 156: Public methods should be documented.

kernel/src/task/tasks.rs
```diff
@@ -384,10 +358,28 @@ impl Task {
             name,
             vm_user_range: Some(vm_user_range),
             rootdir: root,
+            thread_of: None,
         };
         Self::create_common(cpu, create_args)
     }
 
+    pub fn create_thread(
```
deeglaze, line 380: Public methods should be documented.

kernel/src/task/tasks.rs
```diff
@@ -647,27 +639,27 @@ impl Task {
...
+        if self.mm.has_user() {
+            let vmr = self.mm.user_range().unwrap();
+            vmr.remove(addr)?;
+            Ok(())
+        } else {
+            Err(SvsmError::Mem)
         }
-
-        self.vm_user_range.as_ref().unwrap().remove(addr)?;
-        Ok(())
```
deeglaze: I think that return on error is better than the rightward drift added here.

```
if !self.mm.has_user() {
  return Err(SvsmError::Mem);
}

let vmr = self.mm.user_range().unwrap();
vmr.remove(addr)?;
Ok(())
```

kernel/src/task/schedule.rs
```diff
@@ -282,6 +282,41 @@ pub fn start_kernel_task(
     Ok(task)
 }
 
+/// Creates, initializes and starts a new kernel thread of the currently
+/// running kernel task. Note that the thread has already started to run before
+/// this function returns.
+///
+/// # Arguments
+///
+/// * `entry` -  The function to run as the new tasks main function
```
deeglaze: s/new tasks/new task's/

kernel/src/task/schedule.rs
```diff
@@ -282,6 +282,41 @@ pub fn start_kernel_task(
...
+/// * `start_parameter` - Parameter of type `usize` to pass to the new threads entry function.
```
deeglaze: s/new threads entry/new thread's main/

Please use "entry" in both places or "main" in both places.

kernel/src/task/schedule.rs
```diff
@@ -687,4 +687,4 @@ global_asm!(
 ///    to the context switch shadow stacks by executing `saveprevssp`.
 ///
 /// We just switched between two shadow stack tables in different page tables :)
-const CONTEXT_SWITCH_RESTORE_TOKEN: VirtAddr = SVSM_CONTEXT_SWITCH_SHADOW_STACK.const_add(0xff8);
+const CONTEXT_SWITCH_RESTORE_TOKEN: VirtAddr = SVSM_CONTEXT_SWITCH_SHADOW_STACK.const_add(0x1ff8);
```
deeglaze, line 734: Is this constant explained anywhere? I'm guessing the added 0x1000 is for that "free guard page", but no addition to the comment makes me worried that the details will get lost over time.

kernel/src/cpu/shadow_stack.rs
joergroedel: Fixed.

kernel/src/mm/vm/mapping/kernel_stack.rs
joergroedel: Makes sense, the `new_size()` is private now. It is never called outside of this file, so there is no need for a `new_size_shadow()` method.

kernel/src/mm/vm/mapping/kernel_stack.rs
joergroedel: Removed `already`.

kernel/src/task/task_mm.rs
joergroedel: Fixed.

kernel/src/task/task_mm.rs
joergroedel: Fixed.

kernel/src/task/task_mm.rs
joergroedel: Fixed.

kernel/src/task/task_mm.rs
joergroedel: Fixed.

kernel/src/task/task_mm.rs
joergroedel: Fixed.

kernel/src/task/task_mm.rs
joergroedel: Fixed.

kernel/src/task/task_mm.rs
joergroedel: Fixed.

kernel/src/task/tasks.rs
joergroedel: Fixed.

kernel/src/task/tasks.rs
joergroedel: Okay, that made me think again and I came up with a more idiomatic version:

```
    pub fn munmap_user(&self, addr: VirtAddr) -> Result<(), SvsmError> {
        let vmr = self.mm.user_range().ok_or(SvsmError::Mem)?;
        vmr.remove(addr).map(|_| ())
    }
```

joergroedel: And did the same change to the `mmap_user()` method.

kernel/src/task/schedule.rs
joergroedel: Fixed.

kernel/src/task/schedule.rs
joergroedel: Fixed.

kernel/src/task/schedule.rs
joergroedel: Added a comment, thanks.

kernel/src/task/task_mm.rs
joergroedel: You are right, this is not solved yet with the update. I will look into `Send/Sync` annotations separately.

joergroedel: Figured it out, the underlying problem was that `struct VMR` was not `Sync` by default, although it is `Sync` in practice. I added a marker and explained the reasoning, also removed a couple other redundant `Send` and `Sync` annotations. New version is pushed.

kernel/src/mm/vm/mapping/kernel_stack.rs
```diff
@@ -36,10 +36,10 @@ impl VMKernelStack {
     /// # Returns
     ///
     /// Virtual address to program into the hardware stack register
```
00xc: This documentation should be updated to reflect the change in the return value

kernel/src/mm/vm/range.rs
```diff
@@ -66,6 +66,24 @@ pub struct VMR {
...
+//      -> RBTree::Link is used in 'struct VMM' (!Sync)
+//        -> RBTree::Link and VMM are used in VMMAdapter, which makes it !Sync
+//           -> !Sync propagates up through RBTree -> RWLock
+//
+// The usages of the VMM and VMMAdapter in struct VMR are safe nonetheless,
+// because a struct VMM is only used internally in the VMR implementation and
+// never exposed externally. Creation and modification of struct VMM instances
+// and VMMAdapter only happen with the RWLock held for write, which ensures
+// exclusive and non-concurrent access.
+unsafe impl Sync for VMR {}
```
msft-jlange: This seems unnecessarily dangerous.  I understand your point that the `tree` field is `Sync` in practice, but the blanket implementation of `Sync` for `VMR` will mean that any future changes to `VMR` that introduce non-`Sync` fields will be silently covered up, potentially leading to soundness issues.  Can we instead do something like
```
type VmmAdapterTree = RWLock<RBTree<VMMAdapter>>;
unsafe impl Sync for VmmAdapterTree;
```
This would make VMR implicitly `Sync` and would eliminate the possibility of problems arising unexpectedly due to future changes.

msft-jlange: Actually, as I think about it, it would be even better to do something like this:
```
struct VmmAdapterTree {
    tree: RWLock<RBTree<VMMAdapter>>,
}
impl VmmAdapterTree {
    fn lock_read(&self); // locks for read as normal
    unsafe fn lock_write(&self); // declared unsafe because it violates Sync rules
}
unsafe impl Sync for VmmAdapterTree;
```
This would mean that every attempt to acquire a write lock on the tree would require an `unsafe` decoration and therefore justification at the call site for why the `Sync` rules are true.  Your explanation above for why write locks are safe because of locking conventions is compelling, but there is nothing in the code to enforce that to be true.  By requiring the use of `unsafe` at the point of every write lock, we can analyze the soundness in the context of every write lock instead of hoping your invariant remains true forever.

joergroedel: I dopped this change since `struct VMR` is now `Sync` by default.

joergroedel:
> It seems that after your change, `TaskPointer` refers to a thread and not a task. We should probably change the name to `ThreadPointer` and make similar name changes throughout the code base so that we can meaningfully differentiate the term "thread" (which means an execution context) and "task" (which means a set of threads and an associated address space). If, on the other hand, you want "task" to continue to mean thread, then we should make a different set of changes so that we no longer have a "per-task" address space but a "per-process" address space. Either way, the term "task" must have the same meaning everywhere in the code base, whether for a scheduler item, an address space, or whatever we decide it to mean.

Yes, that is a bit confusing right now. I plan to make the change in a future PR which introduces a better distinction between Tasks and Threads. The current `struct Task` will become `struct Thread` and `TaskPointer` will be renamed to `ThreadPointer`. A new `struct Task` will be introduced which to group threads into tasks.

kernel/src/mm/vm/mapping/kernel_stack.rs
joergroedel: Right, thanks. This is fixed now.

kernel/src/mm/address_space.rs
```diff
@@ -174,14 +174,16 @@ pub const SVSM_PERCPU_STACKS_BASE: VirtAddr = SVSM_PERCPU_BASE.const_add(SIZE_LE
 pub const SVSM_SHADOW_STACKS_INIT_TASK: VirtAddr = SVSM_PERCPU_STACKS_BASE;
 
 /// Stack address to use during context switches
-pub const SVSM_CONTEXT_SWITCH_STACK: VirtAddr = SVSM_SHADOW_STACKS_INIT_TASK.const_add(PAGE_SIZE);
+pub const SVSM_CONTEXT_SWITCH_STACK: VirtAddr =
+    SVSM_SHADOW_STACKS_INIT_TASK.const_add(STACK_TOTAL_SIZE);
```
msft-jlange, line 178: This doesn't look right.  I think we need to add the size of a shadow stack (because this comes after `SVSM_SHADOW_STACKS_INIT_TASK`), but that is much smaller than `STACK_TOTAL_SIZE`.  A normal stack is 9 pages (8 for the stack and 1 for the guard page) while a shadow stack is 2 pages (1 for the stack and 1 for the guard page).  It's not harmful to over-allocate VA space for shadow stacks, but in the interest of clarity, it's probably beneficial to define a `SHADOW_STACK_TOTAL_SIZE` constant and use it everywhere we need to describe the VA span of a shadow stack.

kernel/src/task/task_mm.rs
```diff
@@ -0,0 +1,168 @@
+// SPDX-License-Identifier: MIT
```
msft-jlange, line 1: Would it be more appropriate to place this file under `mm` instead of under `task` since it really appears to define memory management primitives within the task address space, and not task primitives related to memory management?

kernel/src/mm/address_space.rs
joergroedel: It is at least not wrong or incorrect. You are right that the mappings could be more compact, but leaving some extra space for the shadow stacks does not hurt either. It also does not increase memory consumption, as all percpu stack mappings still fit into a 2MB range, using only one PTE page.

kernel/src/task/task_mm.rs
joergroedel: I'd like to keep it under `task/`, as it is specific to the task/thread implementation and only uses the primitives from `mm/`, and does not extend the memory management functionality itself.

kernel/src/mm/address_space.rs
msft-jlange: > It is at least not wrong or incorrect. You are right that the mappings could be more compact, but leaving some extra space for the shadow stacks does not hurt either. It also does not increase memory consumption, as all percpu stack mappings still fit into a 2MB range, using only one PTE page.

I agree that the code as written has no errors, but I still find it confusing, because the code reserves more virtual address space than it consumes, and there is no explanation for why that is the case.  Anyone reading the code is likely to assume that the discrepancy represents a mistake and not a conscious choice.  I believe we should either make the change I suggested, or we should add explicit comments explaining why the discrepancy exists.  Whatever we choose can certainly be done in a future PR.

PR #726, opened 2025-06-23
cpu: eliminate static IDT allocation

msft-jlange:
Split the IDT into an early-boot IDT (only large enough to contain exception vectors), which is allocated on the stack, and a global IDT (with full contents), which is allocated from persistent memory.  This makes it much easier to make the IDT read-only after it has been populated.

### DIFF
--- a/kernel/src/cpu/idt/common.rs
+++ b/kernel/src/cpu/idt/common.rs
@@ -13,14 +13,12 @@ use crate::cpu::gdt::GLOBAL_GDT;
 use crate::cpu::registers::{X86GeneralRegs, X86InterruptFrame};
 use crate::cpu::shadow_stack::is_cet_ss_supported;
 use crate::insn_decode::{InsnError, InsnMachineCtx, InsnMachineMem, Register, SegRegister};
-use crate::locking::{RWLock, ReadLockGuard, WriteLockGuard};
-use crate::mm::GuestPtr;
+use crate::mm::{GuestPtr, PageBox};
 use crate::platform::SVSM_PLATFORM;
 use crate::types::{Bytes, SVSM_CS};
 use alloc::boxed::Box;
 use core::arch::asm;
 use core::mem;
-use core::ops::Deref;
 
 pub const DE_VECTOR: usize = 0;
 pub const DB_VECTOR: usize = 1;
@@ -305,24 +303,31 @@ impl IdtEntry {
     }
 }
 
-const IDT_ENTRIES: usize = 256;
+pub const EARLY_IDT_ENTRIES: usize = 32;
 
 #[repr(C, packed(2))]
 #[derive(Default, Clone, Copy, Debug)]
 struct IdtDesc {
-    size: u16,
+    limit: u16,
     address: VirtAddr,
 }
 
-#[derive(Copy, Clone, Debug)]
-pub struct IDT {
-    entries: [IdtEntry; IDT_ENTRIES],
+#[derive(Debug)]
+pub struct IDT<'a> {
+    entries: &'a mut [IdtEntry],
 }
 
-impl IDT {
-    pub const fn new() -> Self {
-        IDT {
-            entries: [IdtEntry::no_handler(); IDT_ENTRIES],
+impl<'a> IDT<'a> {
+    pub const fn new(entries: &'a mut [IdtEntry]) -> Self {
+        Self { entries }
+    }
+
+    /// # Safety
+    /// The caller is required to supply the virtual address of a freshly
+    /// allocated page.
+    pub fn new_from_page(page: PageBox<[IdtEntry]>) -> Self {
+        Self {
+            entries: PageBox::leak(page),
         }
     }
 
@@ -343,14 +348,21 @@ impl IDT {
         self
     }
 
+    pub fn base_limit(&self) -> (u64, u16) {
+        let base: *const IdtEntry = self.entries.as_ptr();
+        let limit = mem::size_of_val(self.entries) - 1;
+        (base as u64, limit.try_into().unwrap())
+    }
+
     /// Load an IDT.
     /// # Safety
     /// The caller must guarantee that the IDT lifetime must be static so that
     /// its entries are always available to the CPU.
     pub unsafe fn load(&self) {
-        let desc: IdtDesc = IdtDesc {
-            size: (IDT_ENTRIES * 16) as u16,
-            address: VirtAddr::from(self.entries.as_ptr()),
+        let (base, limit) = self.base_limit();
+        let desc = IdtDesc {
+            limit,
+            address: VirtAddr::from(base),
         };
 
         // SAFETY: Inline assembly to load an IDT. `'static` lifetime ensures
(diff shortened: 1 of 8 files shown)


kernel/src/cpu/idt/common.rs
```diff
@@ -314,16 +313,25 @@ struct IdtDesc {
...
+        Self { entries }
+    }
+
+    /// # Safety
+    /// The caller is required to supply the virtual address of a freshly
+    /// allocated page.
+    pub unsafe fn new_from_page(addr: VirtAddr) -> Self {
+        let idt_ptr = addr.as_mut_ptr::<IdtEntry>();
+        let count = PAGE_SIZE / mem::size_of::<IdtEntry>();
+        // SAFETY; The caller guarantees the suitability of the pointer.
```
stefano-garzarella: I think the CI is failing for this line (sorry, it looks like clippy is picky about it):
```suggestion
        // SAFETY: The caller guarantees the suitability of the pointer.
```

msft-jlange: Thank you for catching my typo!

kernel/src/cpu/idt/common.rs
```diff
@@ -314,15 +313,25 @@ struct IdtDesc {
...
-            entries: [IdtEntry::no_handler(); IDT_ENTRIES],
+impl<'a> IDT<'a> {
+    pub const fn new(entries: &'a mut [IdtEntry]) -> Self {
+        Self { entries }
+    }
+
+    /// # Safety
+    /// The caller is required to supply the virtual address of a freshly
+    /// allocated page.
+    pub unsafe fn new_from_page(addr: VirtAddr) -> Self {
```
00xc: We should probably take a `PageBox` here and manually leak it. That would make this function completely safe, as we are guaranteed to take ownership of the page, and the backing memory is valid by the type's own invariants. We could even allocate it inside this method. For the first option, this would be my suggestion:

```diff
diff --git a/kernel/src/cpu/idt/common.rs b/kernel/src/cpu/idt/common.rs
index 56bc71fb..36ca7ede 100644
--- a/kernel/src/cpu/idt/common.rs
+++ b/kernel/src/cpu/idt/common.rs
@@ -13,13 +13,13 @@ use crate::cpu::gdt::GLOBAL_GDT;
 use crate::cpu::registers::{X86GeneralRegs, X86InterruptFrame};
 use crate::cpu::shadow_stack::is_cet_ss_supported;
 use crate::insn_decode::{InsnError, InsnMachineCtx, InsnMachineMem, Register, SegRegister};
-use crate::mm::{GuestPtr, PAGE_SIZE};
+use crate::mm::{GuestPtr, PageBox};
 use crate::platform::SVSM_PLATFORM;
 use crate::types::{Bytes, SVSM_CS};
 use alloc::boxed::Box;
 use core::arch::asm;
 use core::mem;
-use core::slice;
+use zerocopy::FromZeros;
 
 pub const DE_VECTOR: usize = 0;
 pub const DB_VECTOR: usize = 1;
@@ -215,7 +215,7 @@ pub fn user_mode(ctxt: &X86ExceptionContext) -> bool {
 
 // The base addresses of the IDT should be aligned on an 8-byte boundary
 // to maximize performance of cache line fills.
-#[derive(Copy, Clone, Default, Debug)]
+#[derive(Copy, Clone, Default, Debug, FromZeros)]
 #[repr(C, packed(8))]
 pub struct IdtEntry {
     low: u64,
@@ -323,16 +323,9 @@ impl<'a> IDT<'a> {
         Self { entries }
     }
 
-    /// # Safety
-    /// The caller is required to supply the virtual address of a freshly
-    /// allocated page.
-    pub unsafe fn new_from_page(addr: VirtAddr) -> Self {
-        let idt_ptr = addr.as_mut_ptr::<IdtEntry>();
-        let count = PAGE_SIZE / mem::size_of::<IdtEntry>();
-        Self {
-            // SAFETY: The caller guarantees the suitability of the pointer.
-            entries: unsafe { slice::from_raw_parts_mut(idt_ptr, count) },
-        }
+    pub fn new_from_page(page: PageBox<[IdtEntry; EARLY_IDT_ENTRIES]>) -> Self {
+        let entries = PageBox::leak(page);
+        Self { entries }
     }
 
     pub fn init(&mut self, handler_array: *const u8, size: usize) -> &mut Self {
diff --git a/kernel/src/cpu/idt/svsm.rs b/kernel/src/cpu/idt/svsm.rs
index 10a303bc..6be55f66 100644
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -23,8 +23,7 @@ use crate::cpu::shadow_stack::IS_CET_SUPPORTED;
 use crate::cpu::X86ExceptionContext;
 use crate::debug::gdbstub::svsm_gdbstub::handle_debug_exception;
 use crate::error::SvsmError;
-use crate::mm::alloc::allocate_page;
-use crate::mm::GuestPtr;
+use crate::mm::{GuestPtr, PageBox};
 use crate::task::{is_task_fault, terminate};
 use crate::tdx::ve::handle_virtualization_exception;
 use crate::utils::immut_after_init::ImmutAfterInitCell;
@@ -126,13 +125,10 @@ pub unsafe fn early_idt_init(idt: &mut IDT<'_>) {
 
 pub fn idt_init() -> Result<(), SvsmError> {
     // Allocate a page of memory to use as the IDT.
-    let idt_page = allocate_page()?;
+    let idt_page = PageBox::try_new_zeroed()?;
+    let mut idt = IDT::new_from_page(idt_page);
     // SAFETY: the page used for the IDT was just allocated for exclusive use.
-    let mut idt = unsafe {
-        let mut idt = IDT::new_from_page(idt_page);
-        early_idt_init(&mut idt);
-        idt
-    };
+    unsafe { early_idt_init(&mut idt) };
 
     // Switch #PF handler to the default one
     idt.set_entry(PF_VECTOR, IdtEntry::entry(asm_entry_pf));
```

msft-jlange: Good idea!  I did this a little bit differently, and that included some more refactoring which eliminated another use of `unsafe` and also fixed a bug.

msft-jlange:
@joergroedel can you help me figure out why the QEMU boot test fails?  This latest version boots correctly on Hyper-V.

p4zuu:
Thanks for helping on this @msft-jlange!
About the QEMU failing test, I also have the same error on native platform. It's actually weird that the early page-fault handler is called. Since `idt_init()` is already called when the #PF is triggered, the runtime pf handler must be called. Seems that IDT is incorrectly updated when calling `idt_init()`. It's maybe not the root of the issue but this isn't expected

kernel/src/cpu/idt/svsm.rs
```diff
@@ -22,15 +22,30 @@ use crate::cpu::registers::RFlags;
...
+use crate::utils::immut_after_init::ImmutAfterInitCell;
 use core::arch::global_asm;
+use core::mem;
 use core::mem::offset_of;
+use core::num::NonZero;
 
 use crate::syscall::*;
 use syscall::*;
 
+pub static GLOBAL_IDT: ImmutAfterInitCell<IDT<'_>> = ImmutAfterInitCell::uninit();
```
p4zuu, line 38: Can we enforce a `IDT<'static>` here? I guess we only expect static IDT to be loaded in the global IDT

msft-jlange: The anonymous lifetime (`'_`) in a static global implies the `'static` lifetime because the underlying variable has the `'static` lifetime.  Making the lifetime explicit here wouldn't actually change anything.

00xc: @msft-jlange I'm not sure this is true. You could create an `IDT<'a>` (e.g. on the stack) and then do `GLOBAL_IDT.init(stack_idt)`.

msft-jlange: > @msft-jlange I'm not sure this is true. You could create an `IDT<'a>` (e.g. on the stack) and then do `GLOBAL_IDT.init(stack_idt)`.

Isn't the borrow checker obligated to prevent mistakes here?  If would expect your example to cause the compiler to generate a "may not live long enough" error.  Otherwise, the compiler would generate UB code by permitting an object to go out of scope while a static still refers to it.

00xc: Ah yes, you're right, please ignore my comment.

joergroedel (approved):
Thanks @msft-jlange , changes LGTM.

PR #727, opened 2025-06-23
mm/page_visibility, hyperv: multiple safety & ergonomic improvements

00xc:
There are multiple uses of types behind a `SharedBox`, which have the particularity that reside on hypervisor-shared memory, which means that:
1. The contents may be updated at any time by the hypervisor without synchronization.
2. The contents may contain arbitrary data, including invalid type representations.

The first use case is the `HvDoorbell` page. Since this type is `Sync` (cannot suffer from data races) and `FromBytes` (has no invalid representations), it should be accessible from behind a `SharedBox` transparently. Thus, allow for types following these characteristics to be accessed through a `SharedBox` via `Deref` or `AsRef`.

The second case is the Hyper-V hypercall pages. In this case, keep the semantic information of memory being shared by not leaking the backing pages, and simply allow unsafely copying bytes from/to the shared memory inside the `HypercallInput` and `HypercallOutput` type methods. The aforementioned methods now have the proper `IntoBytes` / `FromBytes` requirements. While we are at it, remove clippy warning suppressions by properly giving out mutable borrows of the backing pages, ensuring only one piece of code is attempting to mutate them. Finally, remove some unsafe blocks thanks to the type safety we introduced.

### DIFF
--- a/kernel/src/hyperv/hv.rs
+++ b/kernel/src/hyperv/hv.rs
@@ -4,7 +4,7 @@
 //
 // Author: Jon Lange (jlange@microsoft.com)
 
-use crate::address::{VirtAddr, VirtPhysPair};
+use crate::address::{PhysAddr, VirtAddr};
 use crate::cpu::mem::unsafe_copy_bytes;
 use crate::cpu::msr::write_msr;
 use crate::cpu::percpu::{this_cpu, PerCpu};
@@ -14,6 +14,7 @@ use crate::error::SvsmError::HyperV;
 use crate::hyperv;
 use crate::hyperv::{HvInitialVpContext, HyperVMsr};
 use crate::mm::alloc::allocate_pages;
+use crate::mm::page_visibility::SharedBox;
 use crate::mm::pagetable::PTEntryFlags;
 use crate::mm::{virt_to_phys, SVSM_HYPERCALL_CODE_PAGE};
 use crate::platform::SVSM_PLATFORM;
@@ -27,6 +28,29 @@ use core::mem;
 use core::mem::MaybeUninit;
 
 use bitfield_struct::bitfield;
+
+/// An raw, owned page in shared memory used for Hyper-V hypercalls.
+#[derive(Debug)]
+pub struct HypercallPage {
+    // Shared page.
+    page: SharedBox<[u8; PAGE_SIZE]>,
+    // Physical address of the shared page.
+    paddr: PhysAddr,
+}
+
+impl HypercallPage {
+    /// Attempts to allocate a new shared hypercall page.
+    pub fn try_new() -> Result<Self, SvsmError> {
+        let page = SharedBox::<[u8; PAGE_SIZE]>::try_new_zeroed()?;
+        let paddr = virt_to_phys(page.addr());
+        Ok(Self { page, paddr })
+    }
+
+    /// Gets the virtual address of the shared page.
+    fn vaddr(&self) -> VirtAddr {
+        self.page.addr()
+    }
+}
 
 #[derive(Debug)]
 struct HypercallInput<H, T: ?Sized> {
@@ -129,8 +153,8 @@ impl<T> HypercallOutput<T> {
 
 #[derive(Debug)]
 pub struct HypercallPagesGuard<'a> {
-    pub input: RefMut<'a, VirtPhysPair>,
-    pub output: RefMut<'a, VirtPhysPair>,
+    pub input: RefMut<'a, HypercallPage>,
+    pub output: RefMut<'a, HypercallPage>,
     _irq_guard: IrqGuard,
 }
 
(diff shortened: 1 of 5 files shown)


kernel/src/hyperv/hv.rs
```diff
@@ -27,66 +28,87 @@ use core::mem;
...
+    _phantom1: PhantomData<H>,
+    _phantom2: PhantomData<T>,
 }
 
-impl<H, T: ?Sized> HypercallInput<H, T> {
-    // SAFETY: the caller must guarantee the safety of the header pointer so
-    // that this function can assume the safety guarantees expected for
-    // subsequent access.
-    unsafe fn new(header: *mut H) -> Self {
+impl<'a, 'b, H, T: ?Sized> HypercallInput<'a, 'b, H, T> {
```
msft-jlange: The old code used the `?Sized` bound but I don't think this is actually necessary, because `()` is `Sized` and therefore all of the instantiations of `HypercallInput` will always use a `Sized` form of `T`.  Maybe better to remove it now if it isn't actually needed.

kernel/src/hyperv/hv.rs
```diff
@@ -27,66 +28,85 @@ use core::mem;
...
 
-impl<H, T> HypercallInput<H, T> {
-    // SAFETY: the caller must guarantee the safety of the header pointer so
-    // that this function can assume the safety guarantees expected for
-    // subsequent access.
-    unsafe fn new_rep(header: *mut H, page_size: usize) -> Self {
-        let rep_count = (page_size - mem::size_of::<H>()) / mem::size_of::<T>();
+    fn new_rep(page: &'a mut RefMut<'b, HypercallPage>) -> Self {
+        const { assert!(size_of::<H>() <= PAGE_SIZE) };
+        let rep_count = (PAGE_SIZE - mem::size_of::<H>()) / mem::size_of::<T>();
```
msft-jlange, line 100: We should probably add `const { assert_ne!(mem::size_of::<T>(), 0); }` so that any attempt to use a zero-sized `T` will result in a compilation error instead of a runtime divide-by-zero exception.

kernel/src/hyperv/hv.rs
```diff
@@ -27,66 +28,85 @@ use core::mem;
...
+    }
+
+    /// Gets the virtual address of the shared page.
+    fn vaddr(&self) -> VirtAddr {
+        self.page.addr()
+    }
+}
+
+#[derive(Debug)]
+struct HypercallInput<'a, 'b, H, T: ?Sized> {
```
msft-jlange:
```suggestion
struct HypercallInput<'a, 'b, H, T> {
```

It turns out there are no valid cases in which `T` will not be `Sized`, since even the null type `()` is `Sized` to zero.

msft-jlange: Can you add a comment explaining why two lifetimes are needed?  Your earlier explanation to me about one lifetime tracking the borrow of the `HypercallInput` from the `HypercallPagesGuard` and another one associated with the borrow that the `HypercallPagesGuard` inherently holds was really helpful and should be stated here.

kernel/src/hyperv/hv.rs
```diff
@@ -96,24 +116,37 @@ impl<H, T> HypercallInput<H, T> {
...
+    _phantom: PhantomData<T>,
 }
 
-impl<T> HypercallOutput<T> {
-    // SAFETY: the caller must guarantee the safety of the array pointer so
-    // that this function can assume the safety guarantees expected for
-    // subsequent access.
-    unsafe fn new(array: *const T, rep_count: usize) -> Self {
-        Self { array, rep_count }
+impl<'a, 'b, T> HypercallOutput<'a, 'b, T> {
```
msft-jlange, line 137: Can you add a comment here as well about two lifetimes?  Maybe just refer back to the comments describing `HypercallInput`.

kernel/src/hyperv/hv.rs
00xc: Sounds good. I used the regular `assert` because `assert_ne` does not work in const context for some reason.

kernel/src/hyperv/hv.rs
00xc: Sure, let me know if the explanation I added seems sufficient.

kernel/src/hyperv/hv.rs
msft-jlange: Looks like `assert_ne!` requires `PartialEq` and I'll bet that once trait bounds get involved, the compiler can't work out the `const`ness of the assertion.  In any case, I agree that `assert!` is perfectly adequate here.

kernel/src/hyperv/hv.rs
msft-jlange: I think it is sufficient.

PR #729, opened 2025-06-24
scripts/test-in-svsm: avoid leaving `head` processes sleeping

stefano-garzarella:
`head` seems not to handle pipe closure well, in fact especially when the test fails, we leave many `head` processes sleeping, even if the pipe is closed.

Use `read`, and optimize the loop by avoiding opening the pipe every cycle, but assign it to fd 3.

In this case `kill` may fail by finding the subprocess already out because the pipe can be already closed.


joergroedel (approved):
LGTM.

PR #730, opened 2025-06-24
utils: remove `ImmutAfterInitRef`

msft-jlange:
Because `ImmutAfterInitCell` can be initialized by value as well as by reference, it is suitable for holding a global static reference. Therefore, `ImmutAfterInitRef` no longer adds any unique value, and the code can be cleaned up by removing it.

### DIFF
--- a/kernel/src/cpu/cpuid.rs
+++ b/kernel/src/cpu/cpuid.rs
@@ -6,18 +6,18 @@
 
 use crate::address::VirtAddr;
 use crate::types::PAGE_SIZE;
-use crate::utils::immut_after_init::ImmutAfterInitRef;
+use crate::utils::immut_after_init::ImmutAfterInitCell;
 use cpuarch::snp_cpuid::SnpCpuidTable;
 
 use core::arch::asm;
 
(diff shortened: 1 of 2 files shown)


kernel/src/cpu/cpuid.rs
```diff
@@ -39,7 +40,7 @@ pub unsafe fn copy_cpuid_table_to(dst: VirtAddr) {
         start.write_bytes(0, PAGE_SIZE);
         start
             .cast::<SnpCpuidTable>()
-            .copy_from_nonoverlapping(&*CPUID_PAGE, 1);
+            .copy_from_nonoverlapping(ptr::from_ref(*CPUID_PAGE), 1);
```
00xc: Nit: you can simply do the following, as `ImmutAfterInitCell` implements `Deref`, and `&T` can be automatically coerced to `*const T` by the compiler:
```suggestion
            .copy_from_nonoverlapping(*CPUID_PAGE, 1);
```

00xc (approved):

kernel/src/cpu/cpuid.rs
msft-jlange: I thought I had tried that and hit an error, but apparently not.  I agree that this is better.

joergroedel (approved):
LGTM.

PR #731, opened 2025-06-25
stage2: add safety comments

msft-jlange:
Document additional `unsafe` blocks in stage2.

### DIFF
--- a/kernel/src/stage2.rs
+++ b/kernel/src/stage2.rs
@@ -117,6 +117,9 @@ fn setup_env(
         PhysAddr::from(u64::from(STAGE2_START)),
     );
 
+    // SAFETY: the CPUID page address specified in the launch info was mapped
+    // by the loader, which promises to supply a correctly formed CPUID page
+    // at that address.
     let cpuid_page = unsafe { &*(launch_info.cpuid_page as *const SnpCpuidTable) };
     register_cpuid_table(cpuid_page);
     paging_init(platform, true).expect("Failed to initialize early paging");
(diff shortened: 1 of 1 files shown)


kernel/src/stage2.rs
```diff
@@ -220,7 +221,9 @@ fn get_svsm_config(
 }
 
 /// Loads a single ELF segment and returns its virtual memory region.
-fn load_elf_segment(
+/// # Safety
+/// The caller is required to supply the correct virtual address.
```
00xc: We should specify that we mean the virtual addresses for the ELF segment. The caller should also verify that the physical address is valid and unused.

kernel/src/stage2.rs
```diff
@@ -117,6 +117,7 @@ fn setup_env(
         PhysAddr::from(u64::from(STAGE2_START)),
     );
 
+    // SAFETY: the CPUID page address was correctly configured by the caller.
```
00xc: We should explain how we know that it is correctly configured.

kernel/src/stage2.rs
```diff
@@ -266,6 +271,8 @@ fn load_kernel_elf(
     let elf_start = PhysAddr::from(launch_info.kernel_elf_start as u64);
     let elf_end = PhysAddr::from(launch_info.kernel_elf_end as u64);
     let elf_len = elf_end - elf_start;
+    // SAFETY: the caller is required to provide the correct virtual address
```
stefano-garzarella: Should this function (`load_kernel_elf`) be unsafe if it puts safety requirements on the caller?

kernel/src/stage2.rs
```diff
@@ -336,9 +349,12 @@ fn load_igvm_params(
 
     // Copy the contents over
     let src_addr = VirtAddr::from(launch_info.igvm_params as u64);
-    let igvm_src = unsafe { slice::from_raw_parts(src_addr.as_ptr::<u8>(), igvm_vregion.len()) };
-    let igvm_dst = unsafe {
-        slice::from_raw_parts_mut(igvm_vregion.start().as_mut_ptr::<u8>(), igvm_vregion.len())
+    // SAFETY: the caller is required to specify the correct virtual addresses.
```
stefano-garzarella: Ditto for `load_igvm_params()`, should we mark it unsafe?

msft-jlange:
I'm moving the `pvalidate`-related changes to a separate PR.

PR #732, opened 2025-06-25
sev: add safety comments

msft-jlange:
Document and clean up use of `unsafe` blocks in SEV utilties and calling functions.

### DIFF
--- a/kernel/src/platform/snp_fw.rs
+++ b/kernel/src/platform/snp_fw.rs
@@ -186,18 +186,18 @@ fn validate_fw_mem_region(
         let guard = PerCPUPageMappingGuard::create_4k(paddr)?;
         let vaddr = guard.virt_addr();
 
-        pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid)?;
-
-        // Make page accessible to guest VMPL
-        rmp_adjust(
-            vaddr,
-            RMPFlags::GUEST_VMPL | RMPFlags::RWX,
-            PageSize::Regular,
-        )?;
-
-        // SAFETY: we trust PerCPUPageMappingGuard::create_4k() to return a
-        // valid pointer to a correctly mapped region of size PAGE_SIZE.
+        // SAFETY: the virtual address mapping is known to point to the guest
+        // physical address range supplied by the caller.
         unsafe {
+            pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid)?;
+
+            // Make page accessible to guest VMPL
+            rmp_adjust(
+                vaddr,
+                RMPFlags::GUEST_VMPL | RMPFlags::RWX,
+                PageSize::Regular,
+            )?;
+
             zero_mem_region(vaddr, vaddr + PAGE_SIZE);
         }
     }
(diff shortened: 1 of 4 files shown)


kernel/src/platform/snp_fw.rs
```diff
@@ -186,18 +186,17 @@ fn validate_fw_mem_region(
         let guard = PerCPUPageMappingGuard::create_4k(paddr)?;
         let vaddr = guard.virt_addr();
 
-        pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid)?;
+        // SAFETY: the caller is required to specify a guest address range.
```
00xc: Then the function we're in should be unsafe as well to indicate this requirement to the caller.

kernel/src/sev/utils.rs
```diff
@@ -52,14 +52,23 @@ impl fmt::Display for SevSnpError {
     }
 }
 
+/// # Safety
+/// The caller is required to ensure that conversion of the virtual address
+/// range does not violate memory safety.
 fn pvalidate_range_4k(region: MemoryRegion<VirtAddr>, valid: PvalidateOp) -> Result<(), SvsmError> {
```
stefano-garzarella: I think you just forgot since you also put safety's comment, but I think this should be marked unsafe, right?

kernel/src/sev/utils.rs
```diff
@@ -52,14 +52,23 @@ impl fmt::Display for SevSnpError {
...
+        }
     }
 
     Ok(())
 }
 
+/// # Safety
+/// The caller is required to ensure that conversion of the virtual address
+/// range does not violate memory safety.
 pub fn pvalidate_range(
```
stefano-garzarella: Ditto

kernel/src/sev/utils.rs
```diff
@@ -76,15 +85,23 @@ pub fn pvalidate_range(
         {
             // Try to validate as a huge page.
             // If we fail, try to fall back to regular-sized pages.
```
stefano-garzarella: nit: I'd move this comment with `pvalidate` in the unsafe block

kernel/src/platform/snp_fw.rs
msft-jlange: In this case, the comment is wrong - the virtual address range was calculated above from the `PerCPUPageMappingGuard` and that's what makes the virtual address safe.  I'll update the comment.

kernel/src/sev/utils.rs
msft-jlange: I actually think it reads better with this comment on the outside of the unsafe block, because the comment applies to the action that requires the use of the unsafe block.

stefano-garzarella: I thought it was more about what it is inside the `match err`, but I don't have a strong opinion, both are fine.

msft-jlange:
I'm moving the `pvalidate`-related changes to a separate PR.

kernel/src/platform/snp_fw.rs
```diff
@@ -186,18 +186,18 @@ fn validate_fw_mem_region(
         let guard = PerCPUPageMappingGuard::create_4k(paddr)?;
         let vaddr = guard.virt_addr();
 
-        pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid)?;
+        // SAFETY: the virtual address mapping is known to point to the guest
+        // physical address range supplied by the caller.
+        unsafe {
+            pvalidate(vaddr, PageSize::Regular, PvalidateOp::Valid)?;
```
stefano-garzarella, line 192: This can stay out of the `unsafe` block for now, right?

msft-jlange: #748 requires this to be `unsafe`.  I don't want to require that these two PRs merge in a specific order, so they both have the same `unsafe` construction here.  That removes the merge conflict and will enable them to complete in either order.

stefano-garzarella (approved):

PR #741, opened 2025-06-26
kernel/task: make `current_task_terminated()` safe

stefano-garzarella:
In commit 051e2844 ("kernel/syscall: Implement two testing syscalls") we marked `current_task_terminated()` unsafe because it can panic, but this should not be related to memory safety.


joergroedel:
> @joergroedel I may have missed something, do you remember why we marked `current_task_terminated()` unsafe?

The function is unsafe since its inception, not sure why. The commit message does not give much clue. I can probably be marked safe as it does not harm any memory safety guarantees.

PR #743, opened 2025-06-26
fuzz: add safety comments

### LINKED ISSUE #734: fuzz: document unsafe blocks
Undocumented unsafe blocks per module (commit dccbcb2cc3845d5491348a5f1df02434f519dad7)

     18 /fuzz
     18 /fuzz/fuzz_targets
      1 /fuzz/fuzz_targets/acpi.rs
     11 /fuzz/fuzz_targets/alloc.rs
      1 /fuzz/fuzz_targets/fs.rs
      5 /fuzz/fuzz_targets/page_alloc.rs

00xc:
Document all uses of unsafe in the fuzzing harnesses.

Closes #734

### DIFF
--- a/fuzz/fuzz_targets/alloc.rs
+++ b/fuzz/fuzz_targets/alloc.rs
@@ -50,35 +50,53 @@ impl PoisonAllocator {
         }
     }
 
+    /// # Safety
+    ///
+    /// The caller must provide valid memory given by the allocator.
     unsafe fn unpoison_mem(&self, ptr: *mut u8, size: usize) {
+        // SAFETY: the caller must guarantee `ptr` is valid
         unsafe {
             ptr.write_bytes(Self::WRITE_BYTE, size);
         }
     }
 
+    /// # Safety
+    ///
+    /// The caller must provide valid memory given by the allocator.
     unsafe fn poison_mem(&self, ptr: *mut u8, size: usize) {
+        // SAFETY: the caller must guarantee `ptr` is valid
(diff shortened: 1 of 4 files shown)


fuzz/fuzz_targets/alloc.rs
```diff
@@ -50,43 +50,65 @@ impl PoisonAllocator {
...
 
-    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
-        unsafe {
-            let ptr = self.heap.alloc(layout);
-            if !ptr.is_null() {
-                self.unpoison_mem(ptr, layout.size());
-            }
-            ptr
+    fn alloc(&self, layout: Layout) -> *mut u8 {
+        // SAFETY: this cannot cause UB
```
stefano-garzarella: Why?

Should we check that `layout` have non-zero size?

stefano-garzarella (commented):
Thanks! Just a comment, the rest LGTM!

fuzz/fuzz_targets/alloc.rs
00xc: I think `SvsmAllocator` handles zero-sized allocations in a well-defined manner (it will allocate a block from the smallest slab), even if the `GlobalAlloc` trait does not guarantee it. I can add a comment in that aspect.

stefano-garzarella: Okay, I see! Thanks for the comment!

PR #744, opened 2025-06-26
kernel/debug: add missing safety comments

stefano-garzarella:
Add safety comments around `unsafe` blocks and mark some functions `unsafe` if the safety is demanded to the caller.
Add a FIXME in place where unsafe could be very dangerous and silence `clippy::undocumented_unsafe_blocks` lint in gdbstub.rs
Mark all 'gdbstub' public functions unsafe.
Add a compilation error if gdbstub is enabled in release (we already had a comment on discourage that)

### DIFF
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -195,8 +195,10 @@ pub mod svsm_gdbstub {
     }
 
     impl GdbTaskContext {
+        // FIXME: This is unsafe and might cause a crash, but okay until we find
+        // a better way since this is a debug module.
         #[must_use = "The task switch will have no effect if the context is dropped"]
-        fn switch_to_task(id: u32) -> Self {
+        unsafe fn switch_to_task(id: u32) -> Self {
             let cr3 = if is_current_task(id) {
                 0
             } else {
@@ -204,6 +206,10 @@ pub mod svsm_gdbstub {
                 let cr3 = read_cr3();
                 let task = tl.get_task(id);
                 if let Some(task) = task {
+                    // FIXME: cr3 being switched to might not map the current
+                    // CPUs per-cpu space. This might cause a crash when any
+                    // external event, like an #HV comes in when in the other
+                    // tasks context.
                     unsafe {
                         task.page_table.lock().load();
                     }
@@ -486,8 +492,9 @@ pub mod svsm_gdbstub {
                 if let Some(task) = task {
                     // The registers are stored in the top of the task stack as part of the
                     // saved context. We need to switch to the task pagetable to access them.
-                    let _task_context = GdbTaskContext::switch_to_task(tid.get() as u32);
+                    // FIXME: GdbTaskContext::switch_to_task() could be dangerous
                     unsafe {
+                        let _task_context = GdbTaskContext::switch_to_task(tid.get() as u32);
                         *regs = X86_64CoreRegs::from(&*(task.rsp as *const TaskContext));
                     };
                     regs.regs[7] = task.rsp;
@@ -538,7 +545,8 @@ pub mod svsm_gdbstub {
         ) -> gdbstub::target::TargetResult<(), Self> {
             // Switch to the task pagetable if necessary. The switch back will
             // happen automatically when the variable falls out of scope
-            let _task_context = GdbTaskContext::switch_to_task(tid.get() as u32);
+            // FIXME: GdbTaskContext::switch_to_task() could be dangerous
+            let _task_context = unsafe { GdbTaskContext::switch_to_task(tid.get() as u32) };
             let start_addr = VirtAddr::from(start_addr);
             for (off, dst) in data.iter_mut().enumerate() {
                 // SAFETY: Unsafe but ok since this is a debug access
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -11,6 +11,11 @@
 //
 #[cfg(feature = "enable-gdb")]
 pub mod svsm_gdbstub {
+    // Comment the following 2 lines if you really want to enable gdbstub in
+    // release build because memory safety is not guaranteed.
+    #[cfg(not(debug_assertions))]
+    compile_error!("This module must not be built in release mode!");
+
     use crate::address::{Address, VirtAddr};
     use crate::cpu::control_regs::read_cr3;
     use crate::cpu::idt::common::{X86ExceptionContext, BP_VECTOR, DB_VECTOR, VC_VECTOR};
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -4,6 +4,12 @@
 //
 // Author: Roy Hopkins <rhopkins@suse.de>
 
+// Some unsafe blocks can really compromise memory safety, so we are unable to
+// provide appropriate SAFETY comments. However, this is acceptable since
+// all public functions are a nop if `enable-gdb` feature is disabled.
+// `enable-gdb` can't be enabled on release builds.
+#![allow(clippy::undocumented_unsafe_blocks)]
+
 //
 // For release builds this module should not be compiled into the
 // binary. See the bottom of this file for placeholders that are
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -44,11 +44,17 @@ pub mod svsm_gdbstub {
     const INT3_INSTR: u8 = 0xcc;
     const MAX_BREAKPOINTS: usize = 32;
 
+    /// # Safety
+    ///
+    /// Starting the gdbstub is not multi-thread safe, so the caller must
+    /// ensure to call this function on a single processor.
     // The static mutable reference to the stack is protected by the GDB_STATE lock.
     #[allow(static_mut_refs)]
-    pub fn gdbstub_start(platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
+    pub unsafe fn gdbstub_start(platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
+        let mut target = GdbStubTarget::new();
+        // SAFETY: the caller guarantees to call this function on a single processor,
+        // so it's safe to use mutable static variables.
         unsafe {
-            let mut target = GdbStubTarget::new();
             #[expect(static_mut_refs)]
             let gdb = GdbStubBuilder::new(GdbStubConnection::new(platform))
                 .with_packet_buffer(&mut PACKET_BUFFER)
@@ -169,6 +175,8 @@ pub mod svsm_gdbstub {
             log::info!("***********************************");
             log::info!("* Waiting for connection from GDB *");
             log::info!("***********************************");
+            // SAFETY: Inline assembly to generate a software interrupt,
+            // which does not change any state related to memory safety.
             unsafe {
                 asm!("int3");
             }
@@ -211,6 +219,8 @@ pub mod svsm_gdbstub {
     impl Drop for GdbTaskContext {
         fn drop(&mut self) {
             if self.cr3 != 0 {
+                // SAFETY: This is safe because it just restores the previous
+                // page-table, given that the percpu mappings is taken care of.
                 unsafe {
                     asm!("mov %rax, %cr3",
                          in("rax") self.cr3,
@@ -752,7 +762,11 @@ pub mod svsm_gdbstub {
     use crate::cpu::X86ExceptionContext;
     use crate::platform::SvsmPlatform;
 
-    pub fn gdbstub_start(_platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
+    /// # Safety
+    ///
+    /// Starting the gdbstub is not multi-thread safe, so the caller must
+    /// ensure to call this function on a single processor.
+    pub unsafe fn gdbstub_start(_platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


kernel/src/debug/gdbstub.rs
```diff
@@ -196,6 +204,7 @@ pub mod svsm_gdbstub {
                 let cr3 = read_cr3();
                 let task = tl.get_task(id);
                 if let Some(task) = task {
+                    // SAFETY: TODO
```
stefano-garzarella: In `kernel/src/cpu/percpu.rs` we have:
```
        // SAFETY: along with the page table we are also uploading the right
        // TSS and ISST to ensure a memory safe execution state
        unsafe { self.get_pgtable().load() };
```

Is this safe here?
@roy-hopkins @joergroedel you touched this code, any clue?

kernel/src/debug/gdbstub.rs
```diff
@@ -211,6 +220,7 @@ pub mod svsm_gdbstub {
     impl Drop for GdbTaskContext {
         fn drop(&mut self) {
             if self.cr3 != 0 {
+                // SAFETY: TODO
```
stefano-garzarella: I guess it's safe since we read it in `switch_to_task`, but it's enough?
BTW should we use `write_cr3`? (maybe in another patch)

kernel/src/debug/gdbstub.rs
```diff
@@ -477,6 +487,7 @@ pub mod svsm_gdbstub {
                     // The registers are stored in the top of the task stack as part of the
                     // saved context. We need to switch to the task pagetable to access them.
                     let _task_context = GdbTaskContext::switch_to_task(tid.get() as u32);
+                    // SAFETY: TODO
```
stefano-garzarella: Should we do some checks on `task.rsp` or can we trust the value as it is?

@roy-hopkins @joergroedel any clue on this?

kernel/src/debug/stacktrace.rs
```diff
@@ -134,8 +136,10 @@ impl StackUnwinder {
             return UnwoundStackFrame::Invalid;
         }
 
+        // SAFETY: TODO
```
stefano-garzarella: Not sure if we should mark the entire function unsafe and demand to the caller

@00xc any clue on this?

kernel/src/debug/stacktrace.rs
```diff
@@ -134,8 +136,10 @@ impl StackUnwinder {
...
         let rbp = unsafe { rsp.as_ptr::<VirtAddr>().read_unaligned() };
         let rsp = rsp + mem::size_of::<VirtAddr>();
+        // SAFETY: TODO
```
stefano-garzarella: Not sure if we should mark the entire function unsafe and demand to the caller

kernel/src/debug/stacktrace.rs
```diff
@@ -152,6 +156,7 @@ impl StackUnwinder {
             return UnwoundStackFrame::Invalid;
         }
 
+        // SAFETY: TODO
```
stefano-garzarella: Not sure if we should mark the entire function unsafe and demand to the caller

stefano-garzarella:
I need some help on some safety comments :-)

kernel/src/debug/gdbstub.rs
joergroedel: @roy-hopkins might know better, but I think this is safe because all of GDBStub and the memory it uses are globally mapped and are thus accessible from any task context.

joergroedel: One issue here is, though, that the `cr3` being switched to might not map the current CPUs per-cpu space. This might cause a crash when any external event, like an #HV comes in when in the other tasks context.

kernel/src/debug/gdbstub.rs
joergroedel: This is safe because it just restores the previous page-table, given that the percpu mappings is taken care of.

kernel/src/debug/gdbstub.rs
joergroedel: Given that switching to the other tasks page-table is safe (which it seems currently not), this is safe because the code runs with the target-tasks page-table loaded and can access its stack.

joergroedel: In general there is likely a better way to implement accesses to a foreign task context. I need to think about this some more.

kernel/src/debug/stacktrace.rs
joergroedel: This function always works on the stacks of the current context, so de-referencing pointers from the stacks of the context is safe.

stefano-garzarella:
As discussed in TSC meeting, debug code is really unsafe in some places, so the idea is to silence `clippy` in these files and make public API `unsafe`, so that we will use it carefully.

kernel/src/debug/gdbstub.rs
stefano-garzarella: I'll remove `SAFETY` here and left a comment

stefano-garzarella:
v2:
- removed all `SAFETY: TODO`
- added some`SAFETY: `  comments provided by @joergroedel 
- added a `FIXME` in place where `unsafe` could be very dangerous
- marked all public functions unsafe [@msft-jlange]
- added a compilation error if gdbstub is enabled in release (we already had a comment on discourage that)
- silenced clippy::undocumented_unsafe_blocks lint in gdbstub.rs [@msft-jlange]

@joergroedel @msft-jlange PTAL!

stefano-garzarella:
v3:
- fixed wrong SAFETY comment in `kernel/src/cpu/idt/svsm.rs`

kernel/src/debug/gdbstub.rs
```diff
@@ -4,13 +4,24 @@
 //
 // Author: Roy Hopkins <rhopkins@suse.de>
 
+// Some unsafe blocks can really compromise memory safety, so we are unable to
+// provide appropriate SAFETY comments. However, this is acceptable since
+// all public functions are a nop if `enable-gdb` feature is disabled.
+// `enable-gdb` can't be enabled on release builds.
```
msft-jlange, line 10: Safety guarantees are still expected on non-release builds.  The more relevant point here is that debugger operations are inherently unsafe (by design), and that the undocumented unsafe code in this module is only exercised during debugger operations.

kernel/src/debug/gdbstub.rs
```diff
@@ -4,13 +4,24 @@
...
+
 //
 // For release builds this module should not be compiled into the
 // binary. See the bottom of this file for placeholders that are
 // used when the gdb stub is disabled.
 //
 #[cfg(feature = "enable-gdb")]
 pub mod svsm_gdbstub {
+    // Comment the following 2 lines if you really want to enable gdbstub in
+    // release build because memory safety is not guaranteed.
```
msft-jlange, line 21: Excluding debugger support from release builds is about security, not about memory safety.  Including the `compile_error` is a good idea just to create a higher barrier to prevent accidental inclusion, but we should still be accurate with the comments.

kernel/src/debug/gdbstub.rs
```diff
@@ -81,7 +102,12 @@ pub mod svsm_gdbstub {
         }
     }
 
-    pub fn handle_debug_exception(ctx: &mut X86ExceptionContext, exception: usize) {
+    /// # Safety
+    ///
+    /// Some gdbstub operations are very dangerous and could compromise memory
+    /// safety, but these are not avoidable for now, so the caller must be sure
+    /// to use gdbstub carefully (and only in debug mode).
+    pub unsafe fn handle_debug_exception(ctx: &mut X86ExceptionContext, exception: usize) {
```
msft-jlange, line 110: It's not clear to me that this function is really unsafe, because it does not interpret the caller's parameters in a way that could affect memory safety.  The function already appears to include sufficient logic to prevent execution outside of debugger support, and debugger support isn't expected to maintain safety anyway.

kernel/src/debug/gdbstub.rs
```diff
@@ -164,11 +190,18 @@ pub mod svsm_gdbstub {
         }
     }
 
-    pub fn debug_break() {
+    /// # Safety
+    ///
+    /// Some gdbstub operations are very dangerous and could compromise memory
+    /// safety, but these are not avoidable for now, so the caller must be sure
+    /// to use gdbstub carefully (and only in debug mode).
+    pub unsafe fn debug_break() {
```
msft-jlange, line 198: Executing INT3 does not affect memory safety, so this function is not inherently unsafe.

kernel/src/debug/gdbstub.rs
```diff
@@ -187,15 +220,21 @@ pub mod svsm_gdbstub {
     }
 
     impl GdbTaskContext {
+        // FIXME: This is unsafe and might cause a crash, but okay until we find
+        // a better way since this is a debug module.
         #[must_use = "The task switch will have no effect if the context is dropped"]
-        fn switch_to_task(id: u32) -> Self {
+        unsafe fn switch_to_task(id: u32) -> Self {
```
msft-jlange, line 226: Even if we suppress the requirement for undocumented unsafe blocks, we should still write `Safety` comments for unsafe functions.

kernel/src/debug/gdbstub.rs
```diff
@@ -752,11 +795,29 @@ pub mod svsm_gdbstub {
...
 
-    pub fn gdbstub_start(_platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
+    /// # Safety
+    ///
+    /// Starting the gdbstub is not multi-thread safe, so the caller must
+    /// ensure to call this function on a single processor.
+    ///
+    /// Some gdbstub operations are very dangerous and could compromise memory
+    /// safety, but these are not avoidable for now, so the caller must be sure
+    /// to use gdbstub carefully (and only in debug mode).
```
msft-jlange, line 805: It's better to point out that this function is declared `unsafe` only because the gdb-enabled version is `unsafe`, not because it contains any unsafe code.

kernel/src/debug/gdbstub.rs
```diff
@@ -44,11 +55,21 @@ pub mod svsm_gdbstub {
...
     // The static mutable reference to the stack is protected by the GDB_STATE lock.
     #[allow(static_mut_refs)]
-    pub fn gdbstub_start(platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
+    pub unsafe fn gdbstub_start(platform: &'static dyn SvsmPlatform) -> Result<(), u64> {
```
msft-jlange, line 68: I'm not sure this function really needs to be `unsafe`.  Enabling the debugger itself does not compromise memory safety; it is use of the debugger that does.  Code should only be marked `unsafe` if the code itself is unsafe, not just because it enables a scenario that could compromise safety later.

kernel/src/debug/gdbstub.rs
msft-jlange: And if you remove the `unsafe` declaration from the gdb-enabled version, then obviously this whole comment can go away.

kernel/src/debug/gdbstub.rs
stefano-garzarella: Okay, I'll update the comment. thanks.

kernel/src/debug/gdbstub.rs
stefano-garzarella: Perhaps I misunderstood when we discussed this in the TSC. But I thought we agreed to mark all public functions of the module as unsafe because, in some cases, we cannot guarantee safety internally.

If I get it wrong, I can revert this change.

kernel/src/debug/gdbstub.rs
stefano-garzarella: Same of https://github.com/coconut-svsm/svsm/pull/744#discussion_r2235121532

kernel/src/debug/gdbstub.rs
stefano-garzarella: Okay, but this is one of the function that it's really unsafe IIUC, what comment should we write?

kernel/src/debug/gdbstub.rs
stefano-garzarella: But if debugger operations are inherently unsafe, how can we provide safety guarantees if this module is enabled/used?

kernel/src/debug/gdbstub.rs
stefano-garzarella: Same of https://github.com/coconut-svsm/svsm/pull/744/files#r2235121532

kernel/src/debug/gdbstub.rs
stefano-garzarella: Good point (for both), let's discuss on https://github.com/coconut-svsm/svsm/pull/744/files#r2235121532 next step and I'd fix in a way or the other.

kernel/src/debug/gdbstub.rs
stefano-garzarella: I just noticed that I added `FIXME` to explain why this is really unsafe, so not sure what to do here, since the user of this function can't provide any guarantee IIUC.

stefano-garzarella: The FIXME comments was based on @joergroedel comment from https://github.com/coconut-svsm/svsm/pull/744#discussion_r2171361486

> One issue here is, though, that the cr3 being switched to might not map the current CPUs per-cpu space. This might cause a crash when any external event, like an #HV comes in when in the other tasks context.

kernel/src/debug/stacktrace.rs
```diff
@@ -55,6 +55,8 @@ fn is_stage2() -> bool {
 impl StackUnwinder {
     pub fn unwind_this_cpu() -> Self {
         let mut rbp: usize;
+        // SAFETY: Inline assembly to read RPB, which does not change any state
```
msft-jlange, line 58:
```suggestion
// SAFETY: Inline assembly to read RBP, which does not change any state
```

msft-jlange, line 58:
```suggestion
        // SAFETY: Inline assembly to read RBP, which does not change any state
```

msft-jlange:
@stefano-garzarella I think this is moot now that #770 has been merged.  Do you agree?

PR #745, opened 2025-06-27
Migration

ruzickajakub:
All validated guest pages can be transferred from the source SVSM to the target SVSM. Currently, no packaging (no confidentiality) is performed, but a hash of all transferred/received blocks is computed that can be compared to verify that the channel is working correctly. Also, the main migration function of the source and destination SVSM is busy waiting for a signal from QEMU to start outbound or incoming migration.

A single shared page called MigrationPage is used for data transfer and communication with QEMU. The migration page contains two registers and a buffer: a status register, a data register, and a data buffer. The status register is used to signal a change in status (e.g., migration starts, migration is complete). The data register is used to signal that a new page has been prepared in the data buffer by the provider or processed by the consumer. The roles of provider and consumer are switched between SVSM and QEMU on the source and destination machines.

The QEMU implementation of this communication is implemented in a patch[1]. A hack in edkII [2] is required to successfully start the guest. Further patches are required on my machine. Since they are also needed in case migration changes are not implemented, I have them in a separate development branch [3]. I provide it so that one can check it in case the machine does not start or crashes.

[1] https://github.com/coconut-svsm/qemu/pull/23
[2] https://github.com/ruzickajakub/edk2/commit/7209a4ac8a4bc4ee34a2a91e28763ad419298f3f
[3] https://github.com/ruzickajakub/coconut-svsm/tree/migration-dev

### DIFF
--- a/kernel/src/mm/alloc.rs
+++ b/kernel/src/mm/alloc.rs
@@ -44,8 +44,8 @@ impl From<AllocError> for SvsmError {
     }
 }
 
-/// Maximum order of page allocations (up to 128kb)
-pub const MAX_ORDER: usize = 6;
+/// Maximum order of page allocations (up to 512kb)
+pub const MAX_ORDER: usize = 8;
 
 /// Calculates the order of a given size for page allocation.
 ///
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -29,6 +29,7 @@ pub mod io;
 pub mod kernel_region;
 pub mod locking;
 pub mod mm;
+pub mod migration;
 pub mod platform;
 pub mod protocols;
 pub mod requests;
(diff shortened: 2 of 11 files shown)


msft-jlange:
Converting this to draft PR pending discussion on whether this is the desired approach.

kernel/src/mm/alloc.rs
```diff
@@ -44,8 +44,8 @@ impl From<AllocError> for SvsmError {
     }
 }
 
-/// Maximum order of page allocations (up to 128kb)
-pub const MAX_ORDER: usize = 6;
+/// Maximum order of page allocations (up to 512kb)
```
osteffenrh, line 47: This does not match the commit description where it says 256kB

stefano-garzarella:
This PR has been stalled for a while and the codebase has changed significantly, so we're closing it for now, but feel free to open a new one in the future. Since this is a large topic, it would be best to discuss the architecture first and agree on the next steps. Thanks!

PR #746, opened 2025-06-27
kernel: clean up `Send` and `Sync` bounds on locks

msft-jlange:
To improve the detection of synchronization issues, it's helpful to place bounds on `Send` and `Sync` for operations involving locks, instead of simply requiring them when the locks themselves require `Send` and `Sync`.  This makes detection of synchronization issues more obvious.

Changing the lock bounds in this way uncovered a handful of issues with existing data structures, which are being made `Sync`-safe as part of this change as well.


joergroedel (approved):
Changes look good to me, thanks for taking care of this.

PR #747, opened 2025-06-30
Document unsafe blocks in memory management code

joergroedel:
This PR documents all not-yet documented unsafe blocks in the memory management code.

### DIFF
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -23,11 +23,26 @@ use core::ptr::{with_exposed_provenance, with_exposed_provenance_mut};
 use syscall::PATH_MAX;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
+/// Read one byte from a virtual address.
+///
+/// # Arguments
+///
+/// - `v` - Virtual address to read.
+///
+/// # Returns
+///
+/// `Ok(u8)` with the value read on success, `Err(SvsmError)` on failure.
+///
+/// # Safety
+///
+/// Any safety requirements for accessing raw pointers apply here as well.
 #[inline]
-pub fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
+pub unsafe fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
     let mut rcx: u64;
     let mut val: u64;
 
+    // SAFETY: Assembly dereferences the pointer, which is safe when the
+    // function's safety requirements are fulfilled.
     unsafe {
         asm!("1: movb ({0}), %al",
              "   xorq %rcx, %rcx",
@@ -51,7 +66,7 @@ pub fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
     }
 }
 
-/// Writes 1 byte at a virtual address.
+/// Writes one byte at a virtual address.
 ///
 /// # Safety
 ///
@@ -66,6 +81,8 @@ pub fn read_u8(v: VirtAddr) -> Result<u8, SvsmError> {
 pub unsafe fn write_u8(v: VirtAddr, val: u8) -> Result<(), SvsmError> {
     let mut rcx: u64;
 
+    // SAFETY: Assembly writes to virtual address, safe when function's safety
+    // requirements are fulfilled.
     unsafe {
         asm!("1: movb %al, ({0})",
              "   xorq %rcx, %rcx",
@@ -88,12 +105,27 @@ pub unsafe fn write_u8(v: VirtAddr, val: u8) -> Result<(), SvsmError> {
     }
 }
 
+/// Read one word from a virtual address.
+///
+/// # Arguments
+///
+/// - `v` - Virtual address to read.
+///
+/// # Returns
+///
+/// `Ok(u16)` with the value read on success, `Err(SvsmError)` on failure.
+///
+/// # Safety
+///
+/// Any safety requirements for accessing raw pointers apply here as well.
 #[expect(dead_code)]
 #[inline]
 unsafe fn read_u16(v: VirtAddr) -> Result<u16, SvsmError> {
     let mut rcx: u64;
     let mut val: u64;
 
+    // SAFETY: Assembly dereferences the pointer, which is safe when the
+    // function's safety requirements are fulfilled.
     unsafe {
         asm!("1: movw ({0}), {1}",
              "   xorq %rcx, %rcx",
@@ -117,12 +149,27 @@ unsafe fn read_u16(v: VirtAddr) -> Result<u16, SvsmError> {
     }
 }
 
+/// Read one dword from a virtual address.
+///
+/// # Arguments
+///
+/// - `v` - Virtual address to read.
+///
+/// # Returns
+///
+/// `Ok(u32)` with the value read on success, `Err(SvsmError)` on failure.
+///
+/// # Safety
+///
+/// Any safety requirements for accessing raw pointers apply here as well.
 #[expect(dead_code)]
 #[inline]
 unsafe fn read_u32(v: VirtAddr) -> Result<u32, SvsmError> {
     let mut rcx: u64;
     let mut val: u64;
 
+    // SAFETY: Assembly dereferences the pointer, which is safe when the
+    // function's safety requirements are fulfilled.
     unsafe {
         asm!("1: movl ({0}), {1}",
              "   xorq %rcx, %rcx",
@@ -146,12 +193,27 @@ unsafe fn read_u32(v: VirtAddr) -> Result<u32, SvsmError> {
     }
 }
 
+/// Read one qword from a virtual address.
+///
+/// # Arguments
+///
+/// - `v` - Virtual address to read.
+///
+/// # Returns
+///
+/// `Ok(u32)` with the value read on success, `Err(SvsmError)` on failure.
+///
+/// # Safety
+///
+/// Any safety requirements for accessing raw pointers apply here as well.
 #[expect(dead_code)]
 #[inline]
 unsafe fn read_u64(v: VirtAddr) -> Result<u64, SvsmError> {
     let mut rcx: u64;
     let mut val: u64;
 
+    // SAFETY: Assembly dereferences the pointer, which is safe when the
+    // function's safety requirements are fulfilled.
     unsafe {
         asm!("1: movq ({0}), {1}",
              "   xorq %rcx, %rcx",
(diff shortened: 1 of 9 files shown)


kernel/src/mm/guestmem.rs
```diff
@@ -90,10 +94,12 @@ pub unsafe fn write_u8(v: VirtAddr, val: u8) -> Result<(), SvsmError> {
 
 #[expect(dead_code)]
 #[inline]
-unsafe fn read_u16(v: VirtAddr) -> Result<u16, SvsmError> {
+fn read_u16(v: VirtAddr) -> Result<u16, SvsmError> {
```
msft-jlange: This must be `unsafe` because it can read through any arbitrary pointer.  Pointer reads are UB under a number of circumstances (dangling pointers, misaligned pointers, read/write races, etc).

kernel/src/cpu/apic.rs
```diff
@@ -205,7 +205,7 @@ impl LocalApic {
                 // SAFETY: guest vmsa and ca are always validated before being
                 // updated (core_remap_ca(), core_create_vcpu() or
                 // prepare_fw_launch()) so they're safe to use.
```
msft-jlange: Remove the `SAFETY` comment if it is no longer needed.

kernel/src/mm/guestmem.rs
```diff
@@ -119,10 +125,12 @@ unsafe fn read_u16(v: VirtAddr) -> Result<u16, SvsmError> {
 
 #[expect(dead_code)]
 #[inline]
-unsafe fn read_u32(v: VirtAddr) -> Result<u32, SvsmError> {
+fn read_u32(v: VirtAddr) -> Result<u32, SvsmError> {
```
msft-jlange: This should be `unsafe` as described above.

kernel/src/mm/guestmem.rs
```diff
@@ -148,10 +156,12 @@ unsafe fn read_u32(v: VirtAddr) -> Result<u32, SvsmError> {
 
 #[expect(dead_code)]
 #[inline]
-unsafe fn read_u64(v: VirtAddr) -> Result<u64, SvsmError> {
+fn read_u64(v: VirtAddr) -> Result<u64, SvsmError> {
```
msft-jlange: This should be `unsafe` as described above.

kernel/src/mm/guestmem.rs
```diff
@@ -234,9 +257,11 @@ impl<T: Copy> GuestPtr<T> {
     ///
     /// Returns an error if the specified address is not mapped.
     #[inline]
-    pub unsafe fn read(&self) -> Result<T, SvsmError> {
+    pub fn read(&self) -> Result<T, SvsmError> {
```
msft-jlange: The safety of this function is questionable.  It uses an unsafe copy, so it is not subject to the same pointer UB behavior described elsewhere.  However, there is no guarantee that the `GuestPtr` actually points to guest memory, and it's not clear what will happen if a read is attempted on SVSM memory.

kernel/src/cpu/idt/svsm.rs
```diff
@@ -297,23 +297,19 @@ extern "C" fn ex_handler_control_protection(ctxt: &mut X86ExceptionContext, _vec
         code @ (NEAR_RET | FAR_RET_IRET) => {
             // Read the return address on the normal stack.
             let ret_ptr: GuestPtr<u64> = GuestPtr::new(VirtAddr::from(ctxt.frame.rsp));
```
msft-jlange: Why does this code use GuestPtr at all?  This is the local #CP handler so the stack frame is in SVSM memory.  Aren't these just normal memory reads?

kernel/src/mm/page_visibility.rs
```diff
@@ -233,18 +237,24 @@ impl<T: FromBytes + Sync> AsRef<T> for SharedBox<T> {
     }
 }
 
+// SAFETY: FIXME: This is not safe. Any shared memory can be externally
+// modified, which is undefined behavior in Rusts memory model. So a SharedBox
+// can never be Send, even if T is Send.
```
msft-jlange: I don't think this is a problem for `Send`, since `Send` simply means that the data object has the same meaning on one thread as on another (`Send` typically describes references).  What you write here is definitely true for `Sync` so it's good that you're removing that trait.

kernel/src/mm/vm/mapping/api.rs
```diff
@@ -31,7 +31,7 @@ pub struct VMPageFaultResolution {
     pub flags: PTEntryFlags,
 }
 
-pub trait VirtualMapping: core::fmt::Debug {
+pub trait VirtualMapping: core::fmt::Debug + Send + Sync {
```
msft-jlange, line 34: This may or may not be needed after #746.

msft-jlange: Actually, this is better than what I did in #746, so you should keep this and you should remove the `Send + Sync` that I added to all `where T: VirtualMapping` bounds.

kernel/src/cpu/apic.rs
joergroedel: With the read functions staying unsafe, this comment needs to stay as well.

kernel/src/mm/guestmem.rs
joergroedel: Okay, good point. I changed it to keep the read functions unsafe.

kernel/src/mm/guestmem.rs
joergroedel: Dito.

kernel/src/mm/guestmem.rs
joergroedel: Dito.

kernel/src/mm/guestmem.rs
joergroedel: This stays unsafe as well.

kernel/src/cpu/idt/svsm.rs
joergroedel: No idea, maybe @Freax13 has an idea? I think changing that is outside the scope of this PR.

kernel/src/mm/vm/mapping/api.rs
joergroedel: Done.

kernel/src/mm/page_visibility.rs
joergroedel: Thinking more about this, I agree. Making `SharedBox<T>` `Send` when `T` is `Send` should be safe.

kernel/src/cpu/idt/svsm.rs
Freax13: I think I used `GuestPtr` because it allows reads to fail gracefully and works without unsafe code. We do something similar in the [#VC handler](https://github.com/coconut-svsm/svsm/blob/3b1f13585a310ec77416d9bcf8926d05d69e0ee6/kernel/src/cpu/vc.rs#L310-L318), and IIRC that's where I took that from.

Arguably, this code shouldn't use `GuestPtr` because it's not accessing guest memory.

kernel/src/mm/page_visibility.rs
Freax13: If a type implements `Send`, that means it's safe for another thread to take ownership of that type. You can also think of this as sending a mutable reference. If a type implements `Sync`, that means it's safe to send a shared reference to another thread.

In practice, `SharedBox` can always be `Send` and `Sync` regardless of whether `T` implements `Send`/`Sync` because we don't allow direct access to the bytes, and all accesses happen through read and write methods. These methods have bounds on zerocopy traits that convert the raw bytes into instances. This (safely) skips any `Send`/`Sync` requirements.

Btw, the `impl<T: FromBytes + Sync> Deref for SharedBox<T>` impl above is unsound.
> The target is Sync, so there cannot be any data races.

That's wrong. `i32` implements `Sync` but there can clearly be data races for a `i32`.

kernel/src/mm/guestmem.rs
```diff
@@ -88,12 +105,27 @@ pub unsafe fn write_u8(v: VirtAddr, val: u8) -> Result<(), SvsmError> {
...
+///
+/// Any safety requirements for accessing raw pointers apply here as well.
 #[expect(dead_code)]
 #[inline]
 unsafe fn read_u16(v: VirtAddr) -> Result<u16, SvsmError> {
     let mut rcx: u64;
     let mut val: u64;
 
+    // SAFETY: Assembly dereferences the pointer, which is safe when the
+    // functions safety requirements are fulfilled.
```
msft-jlange:
```suggestion
    // function's safety requirements are fulfilled.
```

kernel/src/mm/guestmem.rs
```diff
@@ -117,12 +149,27 @@ unsafe fn read_u16(v: VirtAddr) -> Result<u16, SvsmError> {
...
+///
+/// Any safety requirements for accessing raw pointers apply here as well.
 #[expect(dead_code)]
 #[inline]
 unsafe fn read_u32(v: VirtAddr) -> Result<u32, SvsmError> {
     let mut rcx: u64;
     let mut val: u64;
 
+    // SAFETY: Assembly dereferences the pointer, which is safe when the
+    // functions safety requirements are fulfilled.
```
msft-jlange:
```suggestion
    // function's safety requirements are fulfilled.
```

kernel/src/mm/guestmem.rs
```diff
@@ -146,12 +193,27 @@ unsafe fn read_u32(v: VirtAddr) -> Result<u32, SvsmError> {
...
+///
+/// Any safety requirements for accessing raw pointers apply here as well.
 #[expect(dead_code)]
 #[inline]
 unsafe fn read_u64(v: VirtAddr) -> Result<u64, SvsmError> {
     let mut rcx: u64;
     let mut val: u64;
 
+    // SAFETY: Assembly dereferences the pointer, which is safe when the
+    // functions safety requirements are fulfilled.
```
msft-jlange:
```suggestion
    // function's safety requirements are fulfilled.
```

kernel/src/mm/guestmem.rs
```diff
@@ -173,10 +235,18 @@ unsafe fn read_u64(v: VirtAddr) -> Result<u64, SvsmError> {
...
+/// might happen during the operation.
+///
+/// # Safety
+///
+/// The caller must make sure that writing to `dst` does not harm memory safety.
 #[inline]
 unsafe fn copy_bytes(src: *const u8, dst: *mut u8, size: usize) -> Result<(), SvsmError> {
     let mut rcx: u64;
 
+    // SAFETY: Safe as long as the functions safety requirements are met. Any
```
msft-jlange:
```suggestion
    // SAFETY: Safe as long as the function's safety requirements are met. Any
```

kernel/src/cpu/idt/svsm.rs
joergroedel: Thanks for the comment. As I said, replacing `GuestPtr` here is outside the scope of this PR.

msft-jlange (approved):

PR #748, opened 2025-07-01
platform: mark page validation as `unsafe`

msft-jlange:
Page validation/invalidation operations affect memory safety and must be marked as `unsafe`.


kernel/src/platform/native.rs
```diff
@@ -146,6 +146,9 @@ impl SvsmPlatform for NativePlatform {
         Ok(())
     }
 
+    /// # Safety
+    /// The caller is required to ensure the safety of the validation operation
+    /// on this memory range.
```
peterfang: This looks like it's for `validate_virtual_page_range()`?

kernel/src/platform/snp.rs
```diff
@@ -49,6 +49,9 @@ static VTOM: ImmutAfterInitCell<usize> = ImmutAfterInitCell::uninit();
 
 static APIC_EMULATION_REG_COUNT: AtomicU32 = AtomicU32::new(0);
 
+/// # Safety
+/// The caller is required to ensure the safety of the validation operation on
+/// this memory range.
 fn pvalidate_page_range(range: MemoryRegion<PhysAddr>, op: PvalidateOp) -> Result<(), SvsmError> {
```
peterfang: Should this function be marked `unsafe`?

kernel/src/platform/native.rs
msft-jlange: Yes, thanks for catching that.

kernel/src/platform/snp.rs
msft-jlange: No, because it doesn't consume virtual addresses.  This comment was from an earlier draft of the PR and can be removed.

PR #749, opened 2025-07-01
Split is_qemu in IgvmConfig with more specific feature detection

### LINKED ISSUE #608: IgvmParamBlock.is_qemu field is ambiguous

AdamCDunlap:
Currently, is_qemu decides whether interrupts are supported on SEV-SNP, if the qemu "testdev" device exists for tests, and if the "IORequests" can be used for other tests. Remove the is_qemu boolean and replace it with individual booleans for each of these, so hypervisors that act similarly but not identically can specify what they support.

This is an evolution of #609, and resolves #608


bootlib/src/igvm_params.rs
```diff
@@ -119,9 +119,17 @@ pub struct IgvmParamBlock {
...
+    pub has_qemu_testdev: u8,
+
+    /// Indicates whether SVSM can use "IORequest"s to assist with testing.
+    pub has_test_iorequests: u8,
+
+    /// Indicates whether SVSM should use an IO port for logging.
+    pub has_io_port: u8,
 
     #[doc(hidden)]
     pub _reserved: [u8; 4],
```
AdamCDunlap: I'm not sure the purpose of this reserved field. I'm guessing it's to guarantee alignment of future fields (given that this struct is packed), but since there's no errors being thrown that seems unnecessary.

msft-jlange: > I'm not sure the purpose of this reserved field. I'm guessing it's to guarantee alignment of future fields (given that this struct is packed), but since there's no errors being thrown that seems unnecessary.

It is to make the alignment padding explicit.  Please keep the padding bytes correct.

bootlib/src/igvm_params.rs
```diff
@@ -119,9 +119,17 @@ pub struct IgvmParamBlock {
...
+    pub suppress_svsm_interrupts_on_snp: u8,
+
+    /// Indicates whether SVSM can assume that the qemu testdev device exists to assist testing.
+    pub has_qemu_testdev: u8,
+
+    /// Indicates whether SVSM can use "IORequest"s to assist with testing.
+    pub has_test_iorequests: u8,
+
+    /// Indicates whether SVSM should use an IO port for logging.
+    pub has_io_port: u8,
```
msft-jlange: Can we make this more descriptive, at least to indicate that this is a logging port?  Just saying that the VMM offers an I/O port doesn't say much about how it is expected to be used and how that relates to any possible VMM-specific functionality (which is something that `is_qemu` did try to cover, even if sloppily).

AdamCDunlap: I dug in a little more about how this is used and it looks like it's just for talking to the qemu FwCfg interface. It also looks like this interface is currently only ever used if the IGVM parameters don't contain the MADT. If qemu has support for MADT, then this may be dead code.

Renamed to has_fw_cfg_port for now.

bootlib/src/igvm_params.rs
AdamCDunlap: Fixed padding and added some compile-time assertions for the alignment of a couple of fields.

msft-jlange:
Does this work mean that #609 is no longer needed?

joergroedel (approved):
Changes LGTM.

PR #750, opened 2025-07-01
cpu/ipi: improve safety

msft-jlange:
This change improves safety of the IPI mechanism by encapsulating all of the IPI functionality in the IPI module without any reliance on functions in `PerCpu` or `PerCpuShared` other than a single unsafe function that cannot be abused elsewhere.

Now that all of the IPI logic is together, and synchronization guarantees can be ensured, the IPI state associated with a CPU can be marked `Sync`, since the rules can be upheld in one place.


joergroedel (approved):
Thanks, this is a great step forward. `IpiBoard` not visible anymore outside IPI code is a good improvement.

PR #753, opened 2025-07-02
config: require the presence of IGVM parameters

msft-jlange:
All boot configurations now require the use of IGVM, so the presence of IGVM parameters can be assumed.  There is no reason to include code that can provide configuration information in the absence of IGVM parameters.

### DIFF
--- a/kernel/src/fw_cfg.rs
+++ b/kernel/src/fw_cfg.rs
@@ -6,17 +6,12 @@
 
 extern crate alloc;
 
-use crate::address::{Address, PhysAddr};
 use crate::error::SvsmError;
-use crate::mm::pagetable::max_phys_addr;
-use crate::utils::MemoryRegion;
-use bootlib::kernel_launch::{STAGE2_MAXLEN, STAGE2_START};
 use zerocopy::FromBytes;
 
 use super::io::IOPort;
 use alloc::string::String;
 use alloc::vec::Vec;
-use core::iter;
 use core::mem::size_of;
 
 const FW_CFG_CTL: u16 = 0x510;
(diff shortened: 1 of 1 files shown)


msft-jlange:
@AdamCDunlap can you please take a look?

kernel/src/config.rs
```diff
@@ -61,20 +57,15 @@ fn check_ovmf_regions(
 #[derive(Debug)]
 pub struct SvsmConfig<'a> {
```
AdamCDunlap, line 58: I'm wondering if this class is even useful at this point. Most of the methods simply delegate the same call to igvm_params, so users of this class could mostly be changed to store an IgvmParams instead. The one function that uses fw_cfg, load_cpu_info, is only called in one place which has easy access to the SvsmPlatform already to easily construct the FwCfg object inline.

If you think keeping this class is worthwhile, maybe consider replacing some of the delegated calls with a call to `get_igvm_params`, e.g. instead of `config.page_state_change_required()`, have `config.get_igvm_params().page_state_change_required()`

AdamCDunlap (commented):
This change works with Google's hypervisor. Thanks!

kernel/src/config.rs
msft-jlange: I did consider that, and I think the scales tip slightly in favor of keeping the object.  There are a couple of methods today where the class contains real code (enumerating APIC IDs being a primary example) and it's nice to hide that somewhere.  You are correct that the rest of the methods could be eliminated, but there's not a lot of benefit in doing so - deciding to call `get_igvm_params()` from the outside vs. just using the reference on the inside doesn't change the compiled code nor does it really impact how readable the code is.  Given that it really does seem like a toss-up, I went in favor of minimizing the change.  We can of course revisit this, but I see no hurry in doing so.

AdamCDunlap: Okay, I think those are reasonable choices.

kernel/src/fw_cfg.rs
```diff
@@ -210,120 +125,4 @@ impl<'a> FwCfg<'a> {
...
-            if header.hw_type == TYPE_SVSM_VIRTIO_MMIO {
-                addresses.push(Self::read_from_it::<SimpleDevice>(&mut it)?.mmio_address);
-            } else {
-                for _ in 0..header.size as usize {
-                    it.next();
-                }
-            }
-        }
-        Ok(addresses)
-    }
```
joergroedel: This method is still needed for the `virtio-blk` tests.

joergroedel (commented):
Looks good, except the removed method which is needed for the virtio implementation.

kernel/src/fw_cfg.rs
msft-jlange: Should be fixed now.

PR #754, opened 2025-07-03
kernel/insn_decode: Add safety comments

peterfang:
Make sure unsafe blocks are documented.

### DIFF
--- a/kernel/src/insn_decode/insn.rs
+++ b/kernel/src/insn_decode/insn.rs
@@ -302,9 +302,13 @@ pub mod test_utils {
             }
 
             match size {
+                // SAFETY: caller must ensure `pa` points to a valid address
                 Bytes::One => Ok(unsafe { *(pa as *const u8) } as u64),
+                // SAFETY: caller must ensure `pa` points to a valid address
                 Bytes::Two => Ok(unsafe { *(pa as *const u16) } as u64),
+                // SAFETY: caller must ensure `pa` points to a valid address
                 Bytes::Four => Ok(unsafe { *(pa as *const u32) } as u64),
+                // SAFETY: caller must ensure `pa` points to a valid address
                 Bytes::Eight => Ok(unsafe { *(pa as *const u64) }),
                 _ => Err(InsnError::HandleMmioRead),
             }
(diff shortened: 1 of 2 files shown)


kernel/src/insn_decode/insn.rs
```diff
@@ -299,9 +299,13 @@ pub mod test_utils {
             }
 
             match size {
+                // SAFETY: caller must ensure `pa` points to a valid address
```
stefano-garzarella, line 305: should we mark the function `unsafe` if we demand it to the caller?

kernel/src/insn_decode/insn.rs
```diff
@@ -319,9 +323,13 @@ pub mod test_utils {
...
```
stefano-garzarella, line 332: Ditto

stefano-garzarella (commented):
Thanks for this! I left a couple of comments about marking functions `unsafe`

kernel/src/insn_decode/insn.rs
peterfang: Good question. Right now this only applies to testing. When this is used for instruction emulation, MMIO is usually emulated by software. However, MMIO emulation is implementation-specific so it can definitely end up with a real unsafe memory operation. I expect those memory operations to be marked `unsafe` but we currently don't have these cases. What are your thoughts on this?

kernel/src/insn_decode/insn.rs
peterfang: Same as the previous comments.

kernel/src/insn_decode/insn.rs
stefano-garzarella: Oh I see it now!

Not a strong opinion, can be fine also in this way, but maybe I'd mark it `unsafe` for now, since I suspect it will be unsafe when we will implement it, at least till we pass `pa: usize`.

peterfang: Make sense to me. I'll make this change. Thanks!

peterfang:
@stefano-garzarella I just updated this PR. Please take a look. Thanks!

stefano-garzarella (approved):

PR #755, opened 2025-07-04
Document unafe blocks in CPU code - The easy parts

joergroedel:
Document unsafe blocks in `kernel/src/cpu/`. Leave out IPI and VMSA code, as these need more work.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -65,16 +65,19 @@ use cpuarch::vmsa::VMSA;
 // PERCPU areas virtual addresses into shared memory
 pub static PERCPU_AREAS: PerCpuAreas = PerCpuAreas::new();
 
-// We use an UnsafeCell to allow for a static with interior
-// mutability. Normally, we would need to guarantee synchronization
-// on the backing datatype, but this is not needed because writes to
-// the structure only occur at initialization, from CPU 0, and reads
-// should only occur after all writes are done.
+// We use an UnsafeCell to allow for a static with interior mutability.
+// Normally, we would need to guarantee synchronization on the backing
+// datatype, but this is not needed because writes to the structure only occur
+// at initialization, from CPU 0, and reads of the Vec should only occur after
+// all writes are done.
 #[derive(Debug)]
 pub struct PerCpuAreas {
     areas: UnsafeCell<Vec<&'static PerCpuShared>>,
 }
 
+// SAFETY: Any operation that can affect synchronization safety is declared as
+// unsafe, and therefore callers guarantee that synchronization is always
+// maintained. Also see comment above struct declaration.
 unsafe impl Sync for PerCpuAreas {}
 
 impl PerCpuAreas {
@@ -86,7 +89,7 @@ impl PerCpuAreas {
 
     /// # Safety
     /// The areas vector obtained here is not multi-thread safe, so the caller
-    /// must guarantee that is not used in a context that expects multiwthread
+    /// must guarantee that is not used in a context that expects multi-thread
     /// safety.
     unsafe fn get_areas(&self) -> &Vec<&'static PerCpuShared> {
         // SAFETY: the caller guarantees that accessing the unsafe cell is
@@ -1169,6 +1172,9 @@ impl PerCpu {
 }
 
 pub fn this_cpu() -> &'static PerCpu {
+    // SAFETY: The PerCPU area is always mapped at the same virtual address, so
+    // dereferencing a pointer to that address is safe. The PerCPU area is also
+    // never freed, so using a static lifetime is safe as well.
     unsafe { &*SVSM_PERCPU_BASE.as_ptr::<PerCpu>() }
 }
 


kernel/src/cpu/percpu.rs
```diff
@@ -69,12 +69,15 @@ pub static PERCPU_AREAS: PerCpuAreas = PerCpuAreas::new();
 // mutability. Normally, we would need to guarantee synchronization
 // on the backing datatype, but this is not needed because writes to
 // the structure only occur at initialization, from CPU 0, and reads
-// should only occur after all writes are done.
+// should only occur after all writes are done. The exception is the IpiState
+// structure, which is changed at runtime. But IpiState is Sync, which makes
+// the runtime access safe.
```
msft-jlange: The attributes of `IpiState` have nothing to do with how this structure operates.  The only member of this structure is a `Vec` of references, and nothing is implied about whether the data described by those references are `Sync`.  The only factor that impacts whether this structure is `Sync` is how the `UnsafeCell` is used, and that's described by the existing comment.

kernel/src/cpu/percpu.rs
```diff
@@ -69,12 +69,15 @@ pub static PERCPU_AREAS: PerCpuAreas = PerCpuAreas::new();
...
 #[derive(Debug)]
 pub struct PerCpuAreas {
     areas: UnsafeCell<Vec<&'static PerCpuShared>>,
 }
 
+// SAFETY: See comment above struct declaration
```
msft-jlange: You might want to expand this and say that any operation that can affect synchronization safety is declared as `unsafe`, and therefore callers of the unsafe operations guarantee that synchronization is always maintained when using this structure.

kernel/src/cpu/percpu.rs
```diff
@@ -1169,6 +1172,8 @@ impl PerCpu {
 }
 
 pub fn this_cpu() -> &'static PerCpu {
+    // SAFETY: The PerCPU area is always mapped at the same virtual address, so
+    // dereferencing a pointer to that address is safe.
```
msft-jlange: We should also point out that the backing memory for the `PerCpu` is never freed, and therefore it is safe to assume the `'static` lifetime.

kernel/src/cpu/vc.rs
```diff
@@ -340,6 +340,7 @@ mod tests {
     fn test_has_memory_encryption_info_cpuid() {
         const CPUID_EXTENDED_FUNCTION_INFO: u32 = 0x8000_0000;
         const CPUID_MEMORY_ENCRYPTION_INFO: u32 = 0x8000_001F;
+        // SAFETY: CPUID_EXTENDED_FUNCTION_INFO is a valid CPUID leaf.
```
msft-jlange: CPUID is inherently a safe operation because it cannot impact memory safety.  The fact that it is declared as `unsafe` is an issue with the core library that I hope will be fixed someday.  For our purposes, it is sufficient to point out that CPUID can never affect safety.

kernel/src/cpu/vc.rs
```diff
@@ -388,13 +391,15 @@ mod tests {
     const TESTDEV_ECHO_LAST_PORT: u16 = 0xe0;
 
     fn inb(port: u16) -> u8 {
+        // SAFETY: The assembly code does not modify any memory.
```
msft-jlange: The behavior of IN/OUT is not entirely clear because some synthetic devices can use I/O-based registers to control how memory actions are performed.  I suspect it would be better to declare the IN/OUT functions as `unsafe`.

kernel/src/cpu/percpu.rs
joergroedel: Added, thanks.

kernel/src/cpu/percpu.rs
joergroedel: Right. I updated this hunk to just be more specific about the `Vec`.

kernel/src/cpu/vc.rs
joergroedel: Updated.

kernel/src/cpu/vc.rs
joergroedel: Done.

kernel/src/cpu/percpu.rs
joergroedel: When the callers need to ensure synchronization, doesn't that mean that the struct itself is not Sync?

msft-jlange: `Sync` means that a type is safe to share between threads.  I interpret "safe to share" to mean that all non-`unsafe` accesses to the type are safe to share between threads.  In this case, all of the non-`unsafe` methods observe the required multi-thread safety guarantees.  The only method that can affect multi-thread safety (`create_new()`) is `unsafe` and requires the caller to ensure that it only executes in a non-multithreaded environment.  Since the only method that can affect multi-thread safety is `unsafe`, I don't believe this impacts the `Sync` property.

kernel/src/cpu/vc.rs
```diff
@@ -350,6 +351,7 @@ mod tests {
         if is_test_platform_type(SvsmPlatformType::Snp) {
             const CPUID_VENDOR_INFO: u32 = 0;
 
+            // SAFETY: CPUID_VENDOR_INFO is a valid CPUID leaf.
```
msft-jlange: CPUID never affects safety.

msft-jlange (dismissed):

kernel/src/cpu/vc.rs
joergroedel: Thanks, fixed.

kernel/src/cpu/percpu.rs
joergroedel: Okay, fine for now. I updated the comment.

PR #757, opened 2025-07-07
kernel/task: Add safety comments

peterfang:
Make sure unsafe functions are always marked `unsafe` and unsafe blocks are documented. Also, use debug_assert!() to make sure raw pointers can be used safely.


joergroedel (approved):
The safety comments look correct to me, thanks @peterfang.

PR #759, opened 2025-07-09
libtcgtpm/libcrt: Halt execution if stub is called

osteffenrh:
If a subbed function in the C library is called, stop the execution
using abort().
Also elevate the warning message to an error.

These stubs should never be called. Simply printing a warning is not
enough, since it can easily be missed.

I use the `about()` function that we already present in the libcrt,
which calls `sev::msr_protocol::request_termination_msr()` in the Rust part.

### DIFF
--- a/libtcgtpm/deps/libcrt/src/stub.c
+++ b/libtcgtpm/deps/libcrt/src/stub.c
@@ -2,7 +2,12 @@
 
 #include <libcrt.h>
 
-#define NOT_IMPLEMENTED printf("WARNING: %s not implemented\n", __func__)
+#define NOT_IMPLEMENTED                                                        \
+  do {                                                                         \
+    printf("ERROR: libcrt: %s not implemented\n", __func__);                   \
+    abort();                                                                   \
+  } while (0)
+
 
 // errno.h
 


stefano-garzarella:
Yeah, LGTM! @cclaudio WDYT?

luigix25:
Code LGTM! I'm not sure how should I test it, I tried your diff, but probably i'm doing something wrong

osteffenrh:
> Code LGTM! I'm not sure how should I test it, I tried your diff, but probably i'm doing something wrong

Hmm, are you booting into a full guest OS? Maybe `cock_gettime()` is not getting called somehow? Maybe try a different C function that is actually implemented (be aware, something like `malloc()` might break `printf`)

luigix25:
>Hmm, are you booting into a full guest OS?

Yes, using `launch_guest.sh`! I have a qcow2 with Fedora

luigix25:
>rm libtcgtpm/deps/libcrt/libcrt.a and rebuild ?
Good catch, it triggers the error now! I wonder why `make` didn't catch it. I'll investigate

osteffenrh:
@luigix25 @stefano-garzarella I made the libcrt target phony. Now the libcrt rebuilds automatically.
I think we might have similar issues with the other dependencies...

luigix25:
> > @luigix25 @stefano-garzarella I made the libcrt target phony. Now the libcrt rebuilds automatically. I think we might have similar issues with the other dependencies...
> 
> yep, IMO we should fix the entire Makefile when calling `make` but of course not in this PR.

Yeah.. I just realized that if you do `make distclean` from the project root, it will invoke the openssl `configure` and may fail. It's kind of weird

cclaudio (approved):
Yeah, that's a good idea. LGTM

stefano-garzarella (approved):

PR #760, opened 2025-07-14
[RfC] add memmap protocol

kraxel:


kraxel:
We'll go use the igvm memory map parameter instead, as discussed in Aug 6th meeting.

Issue #700, opened 2025-05-15
Rust toolchain 1.84.0 incompatible with verus


luigix25:
Commit `96c75b8b1b16bfc97490792d4de6a1da46e2a234` bumped the rust toolchain to version 1.84.0, but verus, used for code verification, is still using 1.82.0.

This means that running `cargo verify` returns this error:

`error: rustc 1.82.0 is not supported by the following packages:
  svsm@0.1.0 requires rustc 1.84.0
`

Unfortunately upstream verus is still using 1.82.0

stefano-garzarella:
CC @ziqiaozhou

stefano-garzarella:
@ziqiaozhou I assigned this to you, please let me know if this is fine by you.

PR #761, opened 2025-07-15
Update verus to support rust 1.88 and under

ziqiaozhou:
* Update operator usage with the latest external trait features.
* Distinguish usize vs u64 proofs due to https://github.com/verus-lang/verus/pull/1745
* Repair deprecated verus grammars
* This should close https://github.com/coconut-svsm/svsm/issues/700


ziqiaozhou:
To run verus, users need to re-run the `script/vinstall.sh` to update all verus related tools and need to install 1.88 rust and then run  `cargo verify`. So the current verus version should be compatible with svsm project until svsm decides to update to >1.88

stefano-garzarella:
@ziqiaozhou please add `Closes #700` in the commit message when @luigix25 will confirm it, so the issue will be automatically closed.

kernel/src/address.rs
```diff
@@ -280,10 +280,6 @@ impl ops::Sub<PhysAddr> for PhysAddr {
     type Output = InnerAddr;
 
     #[inline]
-    #[verus_spec(ret =>
```
stefano-garzarella, line 283: Is this removing intentional?

ziqiaozhou: Yes. This is related to the latest change in verus https://github.com/verus-lang/verus/pull/1465. Verus team has added a new feature to better support trait spec in a more elegant way, based on our need. Instead of adding all spec inside implementation, now we can define a spec trait to add annotation for an impl trait. 

Verus has already defined some spec template for all operator traits. For example, instead of adding all pre/post conditions in the implementation, we can implement SubSpecImpl trait inside verus marco to define sub_req, sub_spec. But the verus's template only allows us to define returned value and so I still need to keep some specifications that are not equivalent to verus_spec(returns spec_xx())

stefano-garzarella: Okay, I see. Next time please describe it the commit message.

luigix25:
@stefano-garzarella yes, I had to manually install the rust toolchain too.
I wonder why `scripts/vinstall.sh` worked fine on your system without the override in place. Maybe we should modify that script?

stefano-garzarella:
> @stefano-garzarella yes, I had to manually install the rust toolchain too. I wonder why `scripts/vinstall.sh` worked fine on your system without the override in place. 

I guess because initially I run what you suggested, so it installed the right toolchain, but IMO we should document that.

> Maybe we should modify that script?

Or document in `Documentation/docs/developer/VERIFICATION.m` that prerequisite if verus can't be built with 1.86

ziqiaozhou:
> > @stefano-garzarella yes, I had to manually install the rust toolchain too. I wonder why `scripts/vinstall.sh` worked fine on your system without the override in place.
> 
> I guess because initially I run what you suggested, so it installed the right toolchain, but IMO we should document that.
> 
> > Maybe we should modify that script?
> 
> Or document in `Documentation/docs/developer/VERIFICATION.m` that prerequisite if verus can't be built with 1.86

I updated the script and added the reason:

> Verus requires a specific version of Rust (e.g., 1.88.0) because it depends
  on the compiler’s internal libraries for verification. Thanks to Rust’s strong
  backward compatibility guarantees, Verus can analyze code written against
  older versions. However, some newer Verus features depend on recent rustc
  versions, so we may occasionally upgrade the toolchain to support those
  features, which does not guarantee it is the exact version defined by
  `rust-toolchain.toml`

scripts/vinstall.sh
```diff
@@ -5,8 +5,17 @@
 #
 # Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
 # A script to install verus tools
+VERISMO_REV=7545c6c
 
-VERISMO_REV=a120c45
+# Verus is based on 1.88.0, so we need to install the toolchain first.
+if ! rustup show | grep '1.88'; then
```
stefano-garzarella: Next time `verus` will update the toolchain, we need to change 4 lines here. Can we at least use a variable where to store the required toolchain version?

scripts/vinstall.sh
```diff
@@ -5,8 +5,17 @@
...
+  rustup install 1.88
+fi
+if ! rustup show | grep '1.88.0-x86_64-unknown-linux-gnu'; then
+  rustup target add x86_64-unknown-none --toolchain 1.88
```
stefano-garzarella: This is not working here, I still have the same issue:
```
$ cargo verify                                          
RUSTC="verus-rustc" RUSTUP_TOOLCHAIN="1.88.0" VERUS="/home/stefano/.cargo/git/checkouts/verus-e4ebf515fa1de14c/6c66898/source/target-verus/release/verus" VERUS_TARGETS="svsm,verify_external,verify_proof,verus_stub,vstd" "cargo" "build" "--features" "verus"
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-none` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-none`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `svsm` (lib) due to 1 previous error
cargo build failed with status: exit status: 101
```

But if I use `1.88.0` I can run it:
```
$ rustup target add x86_64-unknown-none --toolchain 1.88.0
info: downloading component 'rust-std' for 'x86_64-unknown-none'
info: installing component 'rust-std' for 'x86_64-unknown-none'

$ cargo verify                                            
RUSTC="verus-rustc" RUSTUP_TOOLCHAIN="1.88.0" VERUS="/home/stefano/.cargo/git/checkouts/verus-e4ebf515fa1de14c/6c66898/source/target-verus/release/verus" VERUS_TARGETS="svsm,verify_external,verify_proof,verus_stub,vstd" "cargo" "build" "--features" "verus"
...
verification results:: 80 verified, 0 errors
verification results:: 0 verified, 0 errors
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.52s
```

stefano-garzarella:
Or maybe we can just set and export RUSTUP_TOOLCHAIN:


```
$ cat scripts/vinstall.sh
#!/bin/bash
# SPDX-License-Identifier: MIT OR Apache-2.0
#
# Copyright (c) Microsoft Corporation
#
# Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
VERISMO_REV=7545c6c
export RUSTUP_TOOLCHAIN="1.87.0"
RUSTUP_TARGET="x86_64-unknown-none"

# Verus is based on 1.88.0, so we need to install the toolchain first.
if ! rustup toolchain list | grep "$RUSTUP_TOOLCHAIN"; then
  rustup install
fi
if ! rustup target list --installed | grep "$RUSTUP_TARGET"; then
  rustup target add "$RUSTUP_TARGET"
fi

# Install verus tools
cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV cargo-v
builtin=`cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "builtin_macros") | .targets[].src_path'`
verus=`dirname $builtin`/../../../source/target-verus/release/verus
if [ -f ${verus} ]; then
    echo "verus (${verus}) is already built"
else
    # build the verus using the source code from builtin/vstd defined in
    # Cargo.toml so that the verus lib and tool are compatible.
   cargo v prepare-verus
fi

# verus-rustc as a wrapper to call verus with proper rustc flags.
cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV verus-rustc

# verus formatter
cargo install --git https://github.com/verus-lang/verusfmt --rev v0.5.7
```

ziqiaozhou:
Thanks for trying out and suggest the improved scripts. I updated based on your suggestion and further improved it by getting RUSTUP_TOOLCHAIN from verus toolchain file.

stefano-garzarella:
@ziqiaozhou okay, now it's much better, but please squash the last commit with the previous one. I don't think make sense to introduce code in a commit and remove it in the next commit of the same PR.

nit: I'd suggest following the stile of the other commit messages (e.g. avoid bulleted lists)

stefano-garzarella (approved):

Issue #737, opened 2025-06-26
kernel: document unsafe blocks


stefano-garzarella:
Undocumented unsafe blocks per module (commit dccbcb2cc3845d5491348a5f1df02434f519dad7)

    146 /kernel
    146 /kernel/src
     30 /kernel/src/cpu
      3 /kernel/src/cpu/ipi.rs
      2 /kernel/src/cpu/percpu.rs
      1 /kernel/src/cpu/shadow_stack.rs
      1 /kernel/src/cpu/sse.rs
     21 /kernel/src/cpu/vc.rs
      2 /kernel/src/cpu/vmsa.rs
      9 /kernel/src/debug
      5 /kernel/src/debug/gdbstub.rs
      4 /kernel/src/debug/stacktrace.rs
      1 /kernel/src/fs
      1 /kernel/src/fs/init.rs
     10 /kernel/src/insn_decode
     10 /kernel/src/insn_decode/insn.rs
      8 /kernel/src/locking
      4 /kernel/src/locking/rwlock.rs
      4 /kernel/src/locking/spinlock.rs
     62 /kernel/src/mm
     21 /kernel/src/mm/alloc.rs
     13 /kernel/src/mm/guestmem.rs
      9 /kernel/src/mm/pagebox.rs
      7 /kernel/src/mm/pagetable.rs
      5 /kernel/src/mm/page_visibility.rs
      4 /kernel/src/mm/ptguards.rs
      1 /kernel/src/mm/validate.rs
      2 /kernel/src/mm/vm
      2 /kernel/src/mm/vm/mapping
      2 /kernel/src/mm/vm/mapping/api.rs
      6 /kernel/src/sev
      6 /kernel/src/sev/utils.rs
      7 /kernel/src/stage2.rs
      1 /kernel/src/syscall
      1 /kernel/src/syscall/class0.rs
      6 /kernel/src/task
      3 /kernel/src/task/schedule.rs
      3 /kernel/src/task/tasks.rs
      6 /kernel/src/utils
      6 /kernel/src/utils/immut_after_init.rs

stefano-garzarella:
kernel/syscall warning should be related to #741

stefano-garzarella:
Current status of our main branch (commit 00b24f830a318a40b56b492b917466e28fde12e2):
```
Undocumented unsafe blocks per module

    105 kernel
    105 kernel/src
     30 kernel/src/cpu
      3 kernel/src/cpu/ipi.rs
[...]

PR #764, opened 2025-07-23
Remove unsafe custom checks

stefano-garzarella:
All our unsafe blocks are finally documented, so we can enable the clippy lint and remove custom checks.


stefano-garzarella:
Marked as draft since this depends on #737 and #744

PR #767, opened 2025-07-28
hyperv: fix hypercall pages addresses

msft-jlange:
Hyper-V expects hypercall page addresses to be passed with the correct shared GPA bits.  However, the virtual-to-physical translation layer always strips both confidentiality and shared bits during address translation.  A new translation method must be defined to preserve those bits.

### DIFF
--- a/kernel/src/mm/pagetable.rs
+++ b/kernel/src/mm/pagetable.rs
@@ -120,7 +120,11 @@ fn make_private_address(paddr: PhysAddr) -> PhysAddr {
 }
 
 fn strip_confidentiality_bits(paddr: PhysAddr) -> PhysAddr {
-    PhysAddr::from(paddr.bits() & !(shared_pte_mask() | private_pte_mask()))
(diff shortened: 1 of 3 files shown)


kernel/src/mm/pagetable.rs
```diff
@@ -460,14 +469,18 @@ pub enum PageFrame {
...
         match *self {
             Self::Size4K(pa) => pa,
             Self::Size2M(pa) => pa,
             Self::Size1G(pa) => pa,
         }
     }
 
+    pub fn address(&self) -> PhysAddr {
+        strip_confidentiality_bits(self.page_frame())
+    }
```
peterfang: Why does `PageFrame::address()` only strip the C bit?

msft-jlange: You're right; this is an error.  I reworked some of the logic but missed this spot, so I'll update.

joergroedel (approved):
LGTM.

PR #768, opened 2025-07-29
mm: introduce `PerCPUMapping` for type-specific mappings

msft-jlange:
The existing `PerCPUPageMappingGuard` exposes only virtual addresses for its mappings, which means (a) they are not type-safe and (b) the lifetimes of any references derived from those addresses are not coupled to the lifetimes of the underlying mappings.  The new `PerCPUMapping` object solves both of these problems by exposing a typed reference that is bound to the lifetime of the underlying mapping.  This is now used in the handful of places that expect a typed mapping; other users of PerCPUPageMappingGuard that only require a virtual address for memory management operations (such as page validation) continue to use the address-based mapping guards.


joergroedel (dismissed):
LGTM. This fixes the unsafe code in the VMSA implementation and passed my testing.

kernel/src/mm/ptguards.rs
```diff
@@ -137,142 +138,66 @@ impl Drop for PerCPUPageMappingGuard {
...
-        self.virt_addr_region(offset * size, size)
-            .map_or(Err(SvsmError::Mem), |region| {
-                // SAFETY: The region was mapped by PerCPUPageMappingGuard and
-                // is safe to access.
-                unsafe {
-                    *(region.start().as_mut_ptr::<T>()) = data;
-                }
-                Ok(())
-            })
+impl<T> AsRef<T> for PerCPUMapping<T> {
```
00xc: This should likely only be implemented for `FromBytes` and `Sync` types

kernel/src/mm/ptguards.rs
```diff
@@ -137,142 +138,66 @@ impl Drop for PerCPUPageMappingGuard {
...
-                    .virt_addr()
-                    .checked_add(self.start_off + offset)?,
-                len,
-            )
-            .filter(|v| self.guard.mapping.region().contains_region(v))
-        } else {
-            None
-        }
+impl<T> AsMut<T> for PerCPUMapping<T> {
+    fn as_mut(&mut self) -> &mut T {
```
00xc, line 186: This should likely only be implemented for types that implement `FromBytes`, `Sync` and `IntoBytes`.

kernel/src/mm/ptguards.rs
```diff
@@ -137,142 +138,66 @@ impl Drop for PerCPUPageMappingGuard {
...
+        let addr = self.mapping.virt_addr() + self.offset;
+        // SAFETY: the mapping is known to be unique and valid by the
```
00xc, line 188: It is not known to be unique though, two mappings can be created for the same physical address

msft-jlange: That's true, but I'm not yet ready to make the leap that all interpretations of a physical address are `unsafe`.  That seems like a better step for your broader PR which tackles memory safety.  It is the case here that as long as the mapping was unique, then the reference is also unique.  Do you think it would be better to clarify the comment to say that?

00xc: Still, this is an unsound API, since it allows the user to violate memory safety without using `unsafe`, and that is not okay. My suggestion would be either to make the `create` method unsafe, adding the required expectations (for example, the mapping being unique) or adding an unsafe `as_mut`-like method instead of implementing `AsMut`.

kernel/src/mm/ptguards.rs
msft-jlange: Not sure I agree.  `PerCpuMapping<T>` does not implement `Sync` so it doesn't matter whether `T` is `Sync`.  `Sync` is only required here if multiple mappings can exist, which is not known.

`FromBytes` and `IntoBytes` are no more necessary here than they would be for a pointer cast to a reference.  The `PerCPUMapping` implies a reference to arbitrary memory, not a reference to a byte stream, so it's more appropriate to think of `PerCPUMapping<T>` as wrapping a `*const T` than a `[u8]`.

kernel/src/mm/ptguards.rs
msft-jlange: See above comment.

kernel/src/mm/ptguards.rs
00xc: > Not sure I agree. PerCpuMapping<T> does not implement Sync so it doesn't matter whether T is Sync. Sync is only required here if multiple mappings can exist, which is not known.

That is not the point. The point is to enforce that once may acquire `&T` from `PerCpuMapping<T>` if and only if `T` has no interior mutability.

> FromBytes and IntoBytes are no more necessary here than they would be for a pointer cast to a reference. The PerCPUMapping implies a reference to arbitrary memory, not a reference to a byte stream, so it's more appropriate to think of PerCPUMapping<T> as wrapping a *const T than a [u8].

Also not the point. The idea here is to enforce that the underlying type has no invalid representations. Otherwise acquiring a `&T` could be undefined behavior (for example if `T` is a `bool` and the underlying memory contains a numeric value other than 1 or 0).

kernel/src/mm/ptguards.rs
msft-jlange: I can accept that `create()` should be `unsafe` because the previous version of the code required `unsafe` in order to consume the physical address as data, and we don't want this PR to regress safety.  However, I don't understand how implementing `AsMut` creates problems relative to an explicit `as_mut()` method (which can't be implemented on its own according to Clippy).  Is the concern that `AsMut` entitles the compiler to perform conversions without requiring explicit syntax?  If so, I believe that marking `create()` as `unsafe` would be sufficient because then obtaining a `PerCPUMapping` requires the caller to consider all aspects of safety, including implicit reference conversions.

00xc: No, the concern is that you cannot mark `AsMut::as_mut` as unsafe because the method signature is safe in the trait definition, thus you need a separate method (you must name it differently so that clippy does not complain).

`create` must be unsafe to indicate that the type has certain assumptions about the memory it is mapping, and the user must uphold those invariants for the type to be safe in the first place (just like `Box::from_raw` is unsafe but all of the other `Box` methods are safe, because the assumption is that the underlying memory is valid, not aliased, heap-allocated, etc).

kernel/src/mm/ptguards.rs
msft-jlange: > > Not sure I agree. PerCpuMapping does not implement Sync so it doesn't matter whether T is Sync. Sync is only required here if multiple mappings can exist, which is not known.
> That is not the point. The point is to enforce that once may acquire `&T` from `PerCpuMapping<T>` if and only if `T` has no interior mutability.

But interior mutability is clearly present in a mutable reference, and this object `PerCPUMapping` offers mutable references, and thus interior mutability must be acceptable on `T`.  From my understanding, `Sync` just guarantees that the type permits multiple, simultaneous references to the same data, and that's only required if multiple callers obtain mappings to the same object.  If the object is designed to permit only a single reference at a time (e.g. a single mutable reference at a time), then `Sync` is never required.  If we make `create()` `unsafe` and leave it to callers to interpret the synchronization concerns, then this seems appropriate.

> Also not the point. The idea here is to enforce that the underlying type has no invalid representations. Otherwise acquiring a `&T` could be undefined behavior (for example if `T` is a `bool` and the underlying memory contains a numeric value other than 1 or 0).

The following sequence should be legal:
```
let pa = some_physical_addr;
let foo = PerCPUMapping::<MaybeUninit::<T>>::create(pa);
// initialize the contents of foo.
drop(foo);
// ...
let bar = PerCPUMapping::<T>::create(pa);
// use the contents of bar
```

This sequence is perfectly valid even if `T` does not implement `FromBytes` or `IntoBytes` because it is fundamentally no different than using a heap pointer to achieve the same characteristics.  The only difference here is that the persistent address of the object is physical, where the virtual address mapping may be different every time it is used - but the data representation has identical characteristics to just filling a `MaybeUninit` and then using it later.

kernel/src/mm/ptguards.rs
msft-jlange: > No, the concern is that you cannot mark `AsMut::as_mut` as unsafe because the method signature is safe in the trait definition, thus you need a separate method (you must name it differently so that clippy does not complain).

But if `create()` is `unsafe`, then the caller can be required to ensure that reference acquisition is always appropriate, and thus `AsMut` can be implemented correctly because the caller of `create()` promises the soundness of `AsMut`.  Is this not correct?

00xc: > But if create() is unsafe, then the caller can be required to ensure that reference acquisition is always appropriate, and thus AsMut can be implemented correctly because the caller of create() promises the soundness of AsMut. Is this not correct?

Sure, but you can segregate the different sources of unsafety: aliasing is a different potential cause of UB than invalid representations. So you can require the caller to guarantee that memory is not aliased and valid, and enforce the rest of the type safety requirements through the traits I suggest in my other comments.

msft-jlange: > Sure, but you can segregate the different sources of unsafety: aliasing is a different potential cause of UB than invalid representations. So you can require the caller to guarantee that memory is not aliased and valid, and enforce the rest of the type safety requirements through the traits I suggest in my other comments.

I think that implies that certain uses of `create()` might be "safer" if `T` implements the other traits (`FromBytes`, `IntoBytes`, `Sync`).  But due to the potential for aliasing, `create()` must always be `unsafe` and thus I see no clear benefit from requiring trait bounds since (a) we want callers to able to avoid these trait bounds when they know it is safe to do so and (b) all callers of `create()` must be `unsafe` anyway.  I certainly don't want to rule out (a), and it's not at all clear how introducing trait bounds for some subset of methods makes any of the rest of the safety matrix any clearer.

kernel/src/mm/ptguards.rs
00xc: > But interior mutability is clearly present in a mutable reference, and this object PerCPUMapping offers mutable references

Not through `AsRef`, which is what my comment was about. You want different bounds on `AsRef` and `AsMut`.

>  If the object is designed to permit only a single reference at a time (e.g. a single mutable reference at a time)

But it is not, two CPUs may map the same physical address and get a `&T` thinking that doing so is safe, while one or more of them mutate the underlying memory through interior mutability (e.g. if the `T` has a `Cell` in it).

> This sequence is perfectly valid even if T does not implement FromBytes or IntoBytes because it is fundamentally no different than using a heap pointer to achieve the same characteristics.

It is not if you acquire a `&T` if `T` has invalid representations and the memory contents of `pa` are such an invalid representation. That is why you want to bound the `AsRef` implementation, so that you may create a `PerCPUMapping` to *any* `T`, but limit acquiring a `&T` to only those `T`s that have the right characteristics.

kernel/src/mm/ptguards.rs
00xc: > I think that implies that certain uses of create() might be "safer" if T implements the other traits (FromBytes, IntoBytes, Sync). But due to the potential for aliasing, create() must always be unsafe and thus I see no clear benefit from requiring trait bounds since (a) we want callers to able to avoid these trait bounds when they know it is safe to do so and (b) all callers of create() must be unsafe anyway. I certainly don't want to rule out (a), and it's not at all clear how introducing trait bounds for some subset of methods makes any of the rest of the safety matrix any clearer.

No, `create()` is always unsafe due to potential aliasing. That's one thing that can cause UB. Then there's the fact that we want to permit getting mutable or immutable references to the underlying `T`, and that is unsafe in its own right, even if there is no aliasing, due to potential invalid representations. We want to prevent the second one through the type system, and ask the user to take care to prevent the fist one.

kernel/src/mm/ptguards.rs
00xc: Just to be clear, I'm not suggesting adding the trait bounds to the type itself, just to the trait implementations, e.g.

```rust
impl<T: FromBytes + Sync> AsRef<T> for PerCPUMapping<T> {
    ...
}
```

This is pretty much the same as what I already did with `SharedBox` in #707 by the way

msft-jlange: > But it is not, two CPUs may map the same physical address and get a `&T` thinking that doing so is safe, while one or more of them mutate the underlying memory through interior mutability (e.g. if the `T` has a `Cell` in it).

I see what you mean.  So are you advocating bounding `AsRef` to `Sync` and defining a separate `as_ref_unsafe()` for non-`Sync` types in which the caller must take responsibility for knowing the reference is safe?

> > This sequence is perfectly valid even if T does not implement FromBytes or IntoBytes because it is fundamentally no different than using a heap pointer to achieve the same characteristics.
> It is not if you acquire a `&T` if `T` has invalid representations and the memory contents of `pa` are such an invalid representation. That is why you want to bound the `AsRef` implementation, so that you may create a `PerCPUMapping` to _any_ `T`, but limit acquiring a `&T` to only those `T`s that have the right characteristics.

Surely it is not unsafe to use `AsRef` in the case where `T` is a `MaybeUninit`, right?  Or is it the case that `MaybeUninit` implements `FromBytes`?

00xc: > I see what you mean. So are you advocating bounding AsRef to Sync and defining a separate as_ref_unsafe() for non-Sync types in which the caller must take responsibility for knowing the reference is safe?

I think we need to bound both AsRef and AsMut (the bounds are different). We can add on top the unsafe variants if you want for those types that do not meet the criteria, although I think there will be no users.

00xc: > Surely it is not unsafe to use AsRef in the case where T is a MaybeUninit, right? Or is it the case that MaybeUninit implements FromBytes?

That would just give you a `&MaybeUninit<T>`, which is fine

msft-jlange: > > Surely it is not unsafe to use AsRef in the case where T is a MaybeUninit, right? Or is it the case that MaybeUninit implements FromBytes?
> That would just give you a `&MaybeUninit<T>`, which is fine

I'm asking a slightly different question.  You point out that using `AsRef` over `MaybeUninit` is fine.  But if `MaybeUninit` doesn't implement `FromBytes`, and if `AsRef` is bounded to `FromBytes`, then how can this safe operation work correctly?

00xc: `FromBytes` is implemented for `MaybeUninit<T>` (even if `T` doesn't implement `FromBytes` itself):

https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html#impl-FromBytes-for-MaybeUninit%3CT%3E

So the case you describe would simply work.

msft-jlange: > `FromBytes` is implemented for `MaybeUninit<T>` (even if `T` doesn't implement `FromBytes` itself):
> So the case you describe would simply work.

OK, in that case I'll just add the trait bounds and if we find there are cases where the bound requirements are causing problems, we can discuss what to do about them.

msft-jlange: It turns out that `MaybeUninit<T>` does not implement `IntoBytes`, so `AsMut` cannot be derived, but a mutable reference is required for `MaybeUninit::write()`.  That means that the following code sequence is impossible as a result of the trait bounds:

```
        let mut context_mapping = unsafe{PerCPUMapping::<MaybeUninit::<ApStartContext>>:: create(PhysAddr::new(context_pa))?};
        context_mapping.as_mut().write(create_ap_start_context(&context, 0));
```

For now, I'm going to remove the `IntoBytes` bound unless you have a better suggestion for how this can be done without involving unsafe code.

00xc: I would then keep the `FromBytes` bound at least (the main difference is that `IntoBytes` does not allow padding, which can cause the user to leak data through a write). Another option is that keep the `IntoBytes` bound and you add a special method for this case, something like:

````rust
impl<T> PerCPUMapping<MaybeUninit<T>> {
    fn into_init(self, val: T) -> PerCPUMapping<T> { ... }
}
```

msft-jlange: > I would then keep the `FromBytes` bound at least (the main difference is that `IntoBytes` does not allow padding, which can cause the user to leak data through a write). Another option is that keep the `IntoBytes` bound and you add a special method for this case, something like:

I am keeping `FromBytes`.  I don't think `IntoBytes` is really all that important here because the point of this object is to map data that is private to the kernel (shared host data, user data, and guest-visible data all have different object types that should be used instead) and thus data leaks are not a concern.  I also don't think the approach of defining a separate implementation for each safe type that doesn't implement `IntoBytes` is unlikely to scale well and should only be used if there is a truly compelling benefit, which I don't see here.

Also, I don't believe the `Sync` bound is needed for `AsMut` because acquiring a mutable reference implies sole ownership, which should mean that cross-context visibility is unimportant, but I'm keeping `Sync` for now because it doesn't appear to be in the way.

kernel/src/mm/ptguards.rs
```diff
@@ -137,142 +139,68 @@ impl Drop for PerCPUPageMappingGuard {
...
-        let size = core::mem::size_of::<T>();
-        self.virt_addr_region(offset * size, size)
-            // SAFETY: The region was mapped by PerCPUPageMappingGuard and is
-            // safe to access.
-            .map_or(Err(SvsmError::Mem), |region| unsafe {
-                Ok(*(region.start().as_ptr::<T>()))
-            })
+    /// The caller must guarantee that the physical address is valid.
+    pub unsafe fn create(paddr: PhysAddr) -> Result<Self, SvsmError> {
+        let offset = u64::from(paddr) as usize & (PAGE_SIZE - 1);
```
peterfang: Nit:
```suggestion
        let offset = paddr.bits() & (PAGE_SIZE - 1);
```

PR #769, opened 2025-07-29
Documentation and RamFS Fixes

joergroedel:
This PR fixes two problems:

1. The documentation on how to run the Verus verifier is not linked in the documentation.
2. An out-of-bounds `Vec` access in RamFS code.

### DIFF
--- a/kernel/src/fs/ramfs.rs
+++ b/kernel/src/fs/ramfs.rs
@@ -95,12 +95,17 @@ impl RawRamFile {
         let page_offset = page_offset(file_offset);
         let page_index = file_offset / PAGE_SIZE;
 
+        // Make sure the page exists before trying to read from it.
+        let Some(page) = self.pages.get(page_index) else {
+            return Ok(0);
+        };
+
         // Minimum of space bytes-to-read on the page and remaining space in buffer
         let buffer_min = min(buffer.size() - buffer_offset, PAGE_SIZE - page_offset);
         // Make sure to not read beyond EOF
         let size = min(self.size.checked_sub(file_offset).unwrap(), buffer_min);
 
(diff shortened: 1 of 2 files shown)


kernel/src/fs/ramfs.rs
```diff
@@ -95,6 +95,13 @@ impl RawRamFile {
         let page_offset = page_offset(file_offset);
         let page_index = file_offset / PAGE_SIZE;
 
+        // When the file is empty (self.pages.len() == 0) the code below would
```
Freax13: It'd be a bit more idiomatic to use `get`:

```rust
let Some(page) = self.pages.get(page_index) else {
    return Ok(0);
};

// ...

page.copy_to_buffer(buffer, buffer_offset, page_offset, size)
```

Freax13 (approved):

kernel/src/fs/ramfs.rs
joergroedel: Good suggestion, updated the patch, thanks.

kernel/src/fs/ramfs.rs
```diff
@@ -95,12 +95,17 @@ impl RawRamFile {
         let page_offset = page_offset(file_offset);
         let page_index = file_offset / PAGE_SIZE;
 
+        // Make sure the page exists before trying to read from it.
+        let Some(page) = self.pages.get(page_index) else {
+            return Ok(0);
```
stefano-garzarella, line 100: Should we return an error on this case?
Or at least document this behavior.

joergroedel: No, this is just a read attempt beyond EOF, which usually returns `0-bytes-read`

stefano-garzarella: I see, I would document that, but for sure not in this PR.

PR #771, opened 2025-07-30
Fix `test_make_ro_after_init()`

joergroedel:
The test fails because it expects a write right after the read-only section to fail. Fix the test so that it only expects failures when writing to the actual read-only section.


stefano-garzarella:
@joergroedel mmm, when we introduced this failure?

stefano-garzarella (approved):
ah, I see, should be #655 we recently merged. LGTM, @joergroedel feel free to merge or require others approvals.

joergroedel:
> ah, I see, should be #655 we recently merged. LGTM, @joergroedel feel free to merge or require others approvals.

Merged, thanks for your review.

PR #772, opened 2025-07-30
Documentation: use `alloc` fuzzer in the example

stefano-garzarella:
`fw_meta` fuzzer was removed by commit 4ae85dfa ("firmware: remove dead code") since we don't use it anymore.

Replace `fw_meta` with `alloc` fuzzer in the example of our fuzzing documentation.


msft-jlange:
Thanks!  I was going to propose exactly the same change after reading the docs yesterday.

Issue #580, opened 2024-12-18
vTPM: generate TCG TPM bindings directly from TPM headers


stefano-garzarella:
from @cclaudio https://github.com/coconut-svsm/svsm/pull/576#pullrequestreview-2512286364:

> It seems that upstream has re-structured the platform header files. Later we may want to follow what they are doing by generatiing bindings for those headers.
> 
> https://github.com/TrustedComputingGroup/TPM/blob/main/TPMCmd/Simulator/src/simulatorPrivate.h , which includes the new platform headers.

This should prevent issues like https://github.com/coconut-svsm/svsm/pull/576

TODO:
- [ ] remove `libtcgtpm/deps/libtcgtpm.h`
- [ ] use bindgen on `libtcgtpm/deps/tpm-20-ref/TPMCmd/Simulator/src/simulatorPrivate.h`  or a subset of headers included there

PR #774, opened 2025-07-31
libtcgtpm: generate bindings from tpm-20-ref headers

stefano-garzarella:
Remove libtcgtpm/deps/libtcgtpm.h since it must be kept in sync with tpm-20-ref code to avoid issues like the one fixed by commit c480aa69 ("libtcgtpm: update _plat__NVEnable and _plat__NVDisable signatures") and generate the bindings directly from TPM headers (i.e. deps/tpm-20-ref/TPMCmd/Platform/include/Platform.h)

We need `-nostdinc` (like we do in the libtcgtpm/Makefile) to avoid clang including std libraries and some new include paths needed by the headers.

Closes: #580


joergroedel (approved):
LGTM.

PR #766, opened 2025-07-28
Introduce Virtio-Vsock support

luigix25:
This PR introduces a wrapper to the virtio-drivers crate like what has been done for virtio-blk. This provides blocking calls to `connect` `close` `send` and `recv`.

Vsock is more flexible than the serial port as it has the concept of port, and we can reuse vsock for anything else.

I had to make 2 changes to the virtio-drivers crate, the idea is to open a PR in that repo. Before proceeding with the PR, I'd like to take some feedback here, maybe more changes are needed.

To compile it you need to enable the feature `vsock` and `virtio-drivers`.

To test it you can use `make test-in-svsm`

Note: If a vsock device is not attached, svsm will not crash.

PR #775, opened 2025-08-05
kernel/platform: Add mmio support for native platform

luigix25:
Currently native platform does not support mmio operations. This can be useful when developing or testing drivers that rely on MMIO, like virtio-vsock or virtio-block.

Tested on top of my RFC #766 that introduces Vsock support, using `make test-in-svsm`

### DIFF
--- a/kernel/src/platform/mod.rs
+++ b/kernel/src/platform/mod.rs
@@ -217,19 +217,19 @@ pub trait SvsmPlatform: Sync {
     ///
     /// # Safety
     ///
-    /// Caller must ensure that `pa` points to a properly aligned memory location and the
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
-    unsafe fn mmio_write(&self, _paddr: PhysAddr, _data: &[u8]) -> Result<(), SvsmError>;
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError>;
 
     /// Perfrom a read from a memory-mapped IO area
     ///
     /// # Safety
     ///
-    /// Caller must ensure that `paddr` points to a properly aligned memory location and the
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
     unsafe fn mmio_read(
         &self,
-        paddr: PhysAddr,
+        vaddr: VirtAddr,
         data: &mut [MaybeUninit<u8>],
     ) -> Result<(), SvsmError>;
 }
--- a/kernel/src/platform/native.rs
+++ b/kernel/src/platform/native.rs
@@ -233,13 +233,13 @@ impl SvsmPlatform for NativePlatform {
         Ok(())
     }
 
-    unsafe fn mmio_write(&self, _paddr: PhysAddr, _data: &[u8]) -> Result<(), SvsmError> {
+    unsafe fn mmio_write(&self, _vaddr: VirtAddr, _data: &[u8]) -> Result<(), SvsmError> {
         unimplemented!()
     }
 
     unsafe fn mmio_read(
         &self,
-        _paddr: PhysAddr,
+        _vaddr: VirtAddr,
         _data: &mut [MaybeUninit<u8>],
     ) -> Result<(), SvsmError> {
         unimplemented!()
--- a/kernel/src/platform/native.rs
+++ b/kernel/src/platform/native.rs
@@ -233,15 +233,123 @@ impl SvsmPlatform for NativePlatform {
         Ok(())
     }
 
-    unsafe fn mmio_write(&self, _vaddr: VirtAddr, _data: &[u8]) -> Result<(), SvsmError> {
-        unimplemented!()
+    /// Perform a write to a memory-mapped IO area
+    ///
+    /// This function expects data to be 1, 2, 4 or 8 bytes long.
+    ///
+    /// It is not possible to loop and write one byte at a time because mmio devices (e.g., those emulated by QEMU)
+    /// expect certain registers to be written with a single operation. Using a generic on SvsmPlatform is
+    /// not possible because it uses the dyn trait.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
+    /// memory accessed is part of a valid MMIO range.
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError> {
+        match data.len() {
+            1 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_write_type::<u8>(vaddr, data);
+                }
+            }
+            2 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_write_type::<u16>(vaddr, data);
+                }
+            }
+            4 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_write_type::<u32>(vaddr, data);
+                }
+            }
+            8 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_write_type::<u64>(vaddr, data);
+                }
+            }
+            _ => return Err(SvsmError::InvalidBytes),
+        };
+
+        Ok(())
     }
 
+    /// Perform a read from a memory-mapped IO area
+    ///
+    /// This function expects reads to be 1, 2, 4 or 8 bytes long.
+    ///
+    /// It is not possible to loop and read one byte at a time because mmio devices (e.g., those emulated by QEMU)
+    /// expect certain registers to be read with a single operation. Using a generic on SvsmPlatform is
+    /// not possible because it uses the dyn trait.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
+    /// memory accessed is part of a valid MMIO range.
     unsafe fn mmio_read(
         &self,
-        _vaddr: VirtAddr,
-        _data: &mut [MaybeUninit<u8>],
+        vaddr: VirtAddr,
+        data: &mut [MaybeUninit<u8>],
     ) -> Result<(), SvsmError> {
-        unimplemented!()
+        match data.len() {
+            1 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_read_type::<u8>(vaddr, data);
+                }
+            }
+            2 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_read_type::<u16>(vaddr, data);
+                }
+            }
+            4 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_read_type::<u32>(vaddr, data);
+                }
+            }
+            8 => {
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+                unsafe {
+                    mmio_read_type::<u64>(vaddr, data);
+                }
+            }
+            _ => return Err(SvsmError::InvalidBytes),
+        };
+
+        Ok(())
     }
+}
+
+unsafe fn mmio_write_type<T: Copy>(vaddr: VirtAddr, data: &[u8]) {
+    let data_ptr = data.as_ptr().cast::<T>();
+    let ptr = vaddr.as_mut_ptr::<T>();
+
+    // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+    unsafe {
+        if data_ptr.is_aligned() {
+            ptr.write_volatile(data_ptr.read());
+        } else {
+            ptr.write_volatile(data_ptr.read_unaligned());
+        }
+    };
+}
+
+unsafe fn mmio_read_type<T>(vaddr: VirtAddr, data: &mut [MaybeUninit<u8>]) {
+    let data_ptr = data.as_mut_ptr().cast::<T>();
+    let ptr = vaddr.as_mut_ptr::<T>();
+
+    // SAFETY: We are trusting the caller to ensure validity of `vaddr`.
+    unsafe {
+        if data_ptr.is_aligned() {
+            data_ptr.write(ptr.read_volatile());
+        } else {
+            data_ptr.write_unaligned(ptr.read_volatile());
+        }
+    };
 }
--- a/kernel/src/platform/snp.rs
+++ b/kernel/src/platform/snp.rs
@@ -381,9 +381,11 @@ impl SvsmPlatform for SnpPlatform {
     ///
     /// # Safety
     ///
-    /// Caller must ensure that `paddr` points to a properly aligned memory location and the
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
-    unsafe fn mmio_write(&self, paddr: PhysAddr, data: &[u8]) -> Result<(), SvsmError> {
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError> {
+        let paddr = this_cpu().get_pgtable().phys_addr(vaddr)?;
+
         // SAFETY: We are trusting the caller to ensure validity of `paddr` and alignment of data.
         unsafe { crate::cpu::percpu::current_ghcb().mmio_write(paddr, data) }
     }
@@ -392,13 +394,14 @@ impl SvsmPlatform for SnpPlatform {
     ///
     /// # Safety
     ///
-    /// Caller must ensure that `paddr` points to a properly aligned memory location and the
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
     unsafe fn mmio_read(
         &self,
-        paddr: PhysAddr,
+        vaddr: VirtAddr,
         data: &mut [MaybeUninit<u8>],
     ) -> Result<(), SvsmError> {
+        let paddr = this_cpu().get_pgtable().phys_addr(vaddr)?;
         // SAFETY: We are trusting the caller to ensure validity of `paddr` and alignment of data.
         unsafe { crate::cpu::percpu::current_ghcb().mmio_read(paddr, data) }
     }
--- a/kernel/src/platform/tdp.rs
+++ b/kernel/src/platform/tdp.rs
@@ -280,13 +280,13 @@ impl SvsmPlatform for TdpPlatform {
         Ok(())
     }
 
-    unsafe fn mmio_write(&self, _paddr: PhysAddr, _data: &[u8]) -> Result<(), SvsmError> {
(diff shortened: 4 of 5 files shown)


kernel/src/platform/mod.rs
```diff
@@ -218,15 +218,15 @@ pub trait SvsmPlatform: Sync {
     ///
     /// Caller must ensure that `pa` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
```
stefano-garzarella: Please update the documentation here and in other places.

kernel/src/platform/snp.rs
```diff
@@ -393,8 +395,10 @@ impl SvsmPlatform for SnpPlatform {
     ///
     /// Caller must ensure that `paddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
-    unsafe fn mmio_read(&self, paddr: PhysAddr, data: &mut [u8]) -> Result<(), SvsmError> {
+    unsafe fn mmio_read(&self, vaddr: VirtAddr, data: &mut [u8]) -> Result<(), SvsmError> {
         // SAFETY: We are trusting the caller to ensure validity of `paddr` and alignment of data.
+        let paddr = this_cpu().get_pgtable().phys_addr(vaddr).unwrap();
```
stefano-garzarella: Is it okay to panic here?

kernel/src/virtio/hal.rs
```diff
@@ -197,18 +196,16 @@ unsafe impl virtio_drivers::Hal for SvsmHal {
...
     unsafe fn mmio_read<T: FromBytes + Immutable>(src: &T) -> T {
-        let paddr = this_cpu()
-            .get_pgtable()
-            .phys_addr(VirtAddr::from(addr_of!(*src)))
-            .unwrap();
-
-        let mut b = alloc::vec![0; size_of::<T>()];
         // SAFETY: We are trusting the caller (the virtio driver) to ensure `src` is a valid MMIO
         // address and that it is alinged properly.
+        let mut buffer = alloc::vec![0; size_of::<T>()];
```
stefano-garzarella: Why this change?
the `// SAFETY: ` comment should be on top of the unsafe block, also avoid this kind of unrelated change, use another patch for that (I mean replacing `b` with `buffer`).

kernel/src/platform/native.rs
```diff
@@ -237,11 +237,75 @@ impl SvsmPlatform for NativePlatform {
...
 
-    unsafe fn mmio_write(&self, _vaddr: VirtAddr, _data: &[u8]) -> Result<(), SvsmError> {
-        unimplemented!()
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError> {
+        let number = vaddr.as_usize();
+
+        match data.len() {
+            1 => {
+                let ptr = number as *mut u8;
+                let data = u8::from_le_bytes(data.try_into().unwrap());
```
stefano-garzarella: why panic here? (and in other places)

stefano-garzarella: and also why from little endian?

kernel/src/platform/native.rs
```diff
@@ -237,11 +237,75 @@ impl SvsmPlatform for NativePlatform {
         Ok(())
     }
 
-    unsafe fn mmio_write(&self, _vaddr: VirtAddr, _data: &[u8]) -> Result<(), SvsmError> {
-        unimplemented!()
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError> {
+        let number = vaddr.as_usize();
+
+        match data.len() {
```
stefano-garzarella, line 249: Should we do a loop if `data.len()` is > 8 ? Or it's not expected?
(this should be documented IMO)

kernel/src/platform/native.rs
```diff
@@ -237,11 +237,75 @@ impl SvsmPlatform for NativePlatform {
         Ok(())
     }
 
-    unsafe fn mmio_write(&self, _vaddr: VirtAddr, _data: &[u8]) -> Result<(), SvsmError> {
-        unimplemented!()
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError> {
+        let number = vaddr.as_usize();
```
stefano-garzarella: please use a better name, or just remove it and use `vaddr.as_usize()` in all the place.

kernel/src/platform/native.rs
```diff
@@ -237,11 +237,75 @@ impl SvsmPlatform for NativePlatform {
...
+            _ => return Err(SvsmError::InvalidBytes),
+        };
+
+        Ok(())
     }
 
-    unsafe fn mmio_read(&self, _vaddr: VirtAddr, _data: &mut [u8]) -> Result<(), SvsmError> {
-        unimplemented!()
+    unsafe fn mmio_read(&self, vaddr: VirtAddr, data: &mut [u8]) -> Result<(), SvsmError> {
+        let number = vaddr.as_usize();
```
stefano-garzarella: Ditto, `number` is really useless as name IMO. Try `vaddr_raw`, or again just use `vaddr.as_usize()` where you need it, it's much more readable.

kernel/src/platform/native.rs
stefano-garzarella, line 1: Can you describe better in the commit description and maybe also on top of `mmio_write|read` why you need to implement in that way?

stefano-garzarella (changes requested):

kernel/src/platform/native.rs
stefano-garzarella: > why panic here? (and in other places)

this should be fine, since we know the size. The `from_le_bytes` usage is still not clear to me, IMO we should use `from_ne_bytes`.

kernel/src/platform/snp.rs
luigix25: Do you want to propagate the error to the caller? Before this, the panic was in the SvsmHal.

kernel/src/virtio/hal.rs
luigix25: oops that was unintentional, I'll roll back to using `b` as a variable name.

kernel/src/platform/snp.rs
stefano-garzarella: This function returns `Result<(), SvsmError>` so I'm trying to understand why you want to panic here instead of just propagating it. I mean simply this:
```suggestion
        let paddr = this_cpu().get_pgtable().phys_addr(vaddr)?;
```

kernel/src/platform/native.rs
luigix25: The virtio spec (https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-182001r1) says that values are little endian. Using `from_ne_bytes` we are using the host's endianess right? Would this cause an issue on a big endian machine?

edit: never mind, we can have non-virtio mmio devices.

kernel/src/platform/native.rs
luigix25: I don't expect to have read or write operations bigger than 8 bytes. I'll write a comment explaining that

luigix25:
@stefano-garzarella thanks for the review!
I should have addressed all your comments.

kernel/src/platform/native.rs
```diff
@@ -237,11 +237,111 @@ impl SvsmPlatform for NativePlatform {
...
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
+    /// memory accessed is part of a valid MMIO range.
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError> {
+        match data.len() {
+            1 => {
+                let ptr: *mut u8 = vaddr.as_usize() as *mut u8;
+                let data = u8::from_ne_bytes(data.try_into().unwrap());
```
osteffenrh: Can't we somehow cast this instead of creating a copy?

kernel/src/platform/native.rs
```diff
@@ -237,11 +237,111 @@ impl SvsmPlatform for NativePlatform {
...
     }
 
-    unsafe fn mmio_write(&self, _paddr: PhysAddr, _data: &[u8]) -> Result<(), SvsmError> {
-        unimplemented!()
+    /// Perform a write to a memory-mapped IO area
+    ///
+    /// This function expects data to be 1, 2, 4 or 8 bytes long.
+    ///
+    /// It is not possible to loop and write one byte at a time because this would violate
+    /// the virtio spec. Using a generic on SvsmPlatform is not possible because it uses the
```
stefano-garzarella: This module is not related to virtio, so I'd not mention that. I think we should mention here that mmio devices (e.g., those emulated by QEMU) expect certain registers to be read with a single read operation rather than byte by byte.

kernel/src/virtio/hal.rs
```diff
@@ -197,17 +196,15 @@ unsafe impl virtio_drivers::Hal for SvsmHal {
...
         let mut b = alloc::vec![0; size_of::<T>()];
         // SAFETY: We are trusting the caller (the virtio driver) to ensure `src` is a valid MMIO
         // address and that it is alinged properly.
         unsafe {
-            SVSM_PLATFORM.mmio_read(paddr, &mut b).unwrap();
+            SVSM_PLATFORM
+                .mmio_read(VirtAddr::from(addr_of!(*src)), &mut b)
+                .unwrap();
         }
+
```
stefano-garzarella: nit: unrelated change

kernel/src/platform/tdp.rs
```diff
@@ -270,11 +270,11 @@ impl SvsmPlatform for TdpPlatform {
         Ok(())
     }
 
-    unsafe fn mmio_write(&self, _paddr: PhysAddr, _data: &[u8]) -> Result<(), SvsmError> {
+    unsafe fn mmio_write(&self, _paddr: VirtAddr, _data: &[u8]) -> Result<(), SvsmError> {
```
stefano-garzarella: Please update also the name of the variable

kernel/src/platform/mod.rs
```diff
@@ -216,17 +216,17 @@ pub trait SvsmPlatform: Sync {
...
 
     /// Perfrom a read from a memory-mapped IO area
     ///
     /// # Safety
     ///
-    /// Caller must ensure that `pa` points to a properly aligned memory location and the
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
-    unsafe fn mmio_read(&self, _paddr: PhysAddr, _data: &mut [u8]) -> Result<(), SvsmError>;
+    unsafe fn mmio_read(&self, _vaddr: VirtAddr, data: &mut [u8]) -> Result<(), SvsmError>;
```
stefano-garzarella: why removing `_` only to `data`?

(BTW I think we don't need `_` in the trait variables, so feel free to remove it all over the place)

kernel/src/platform/native.rs
stefano-garzarella: Maybe no, because IIUC `write_volatile` wants a value, not a pointer/reference. For that reason I suggested to use generics for `data`, but we can't since the trait is used as dynamic object :-(

luigix25: you mean just for this case?

osteffenrh: The other cases (u16, u32, u64) too.
Probably not super relevant, probably optimized away.

`ptr.write_volatile(*(data.as_ptr() as *const u16));` ?

kernel/src/platform/native.rs
```diff
@@ -237,11 +237,111 @@ impl SvsmPlatform for NativePlatform {
...
+    /// dyn trait.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
+    /// memory accessed is part of a valid MMIO range.
+    unsafe fn mmio_write(&self, vaddr: VirtAddr, data: &[u8]) -> Result<(), SvsmError> {
+        match data.len() {
+            1 => {
+                let ptr: *mut u8 = vaddr.as_usize() as *mut u8;
```
stefano-garzarella: VirtAddr implements `as_mut_ptr`, so I think here (and in other places, you can simplify in this way:

```suggestion
                let ptr = vaddr.as_mut_ptr::<u8>();
```

stefano-garzarella:
@luigix25 I just posted #782 that maybe can simplify a bit also the code in this PR

luigix25:
@stefano-garzarella yes! I'll wait until that PR is merged and then rebase

kernel/src/platform/mod.rs
```diff
@@ -217,19 +217,19 @@ pub trait SvsmPlatform: Sync {
...
     ///
     /// # Safety
     ///
-    /// Caller must ensure that `paddr` points to a properly aligned memory location and the
+    /// Caller must ensure that `vaddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
     unsafe fn mmio_read(
         &self,
-        paddr: PhysAddr,
+        paddr: VirtAddr,
```
stefano-garzarella: `vaddr` here, right?

kernel/src/platform/native.rs
```diff
@@ -233,15 +233,119 @@ impl SvsmPlatform for NativePlatform {
...
     }
 
-    unsafe fn mmio_write(&self, _vaddr: VirtAddr, _data: &[u8]) -> Result<(), SvsmError> {
-        unimplemented!()
+    /// Perform a write to a memory-mapped IO area
+    ///
+    /// This function expects data to be 1, 2, 4 or 8 bytes long.
+    ///
+    /// It is not possible to loop and write one byte at a time because mmio devices (e.g., those emulated by QEMU)
+    /// expect certain registers to be writeen with a single operation. Using a generic on SvsmPlatform is
```
stefano-garzarella: s/writeen/written

kernel/src/platform/native.rs
```diff
@@ -233,15 +233,119 @@ impl SvsmPlatform for NativePlatform {
...
+        match data.len() {
+            1 => {
+                let ptr = vaddr.as_mut_ptr::<u8>();
+
+                // SAFETY: We are trusting the caller to ensure validity of `vaddr` and alignment of data.
+                let value = unsafe { ptr.read_volatile().to_ne_bytes() };
+
+                for (d, s) in data.iter_mut().zip(value.iter()) {
+                    d.write(*s);
+                }
```
stefano-garzarella: What about doing this:
```rust
                unsafe {
                    data.as_mut_ptr().cast::<u8>().write(ptr.read_volatile());
                }
```

We can eventually create a function for that with a generic `T` (something similar also for the `mmio_write` where all the cases has the same pattern.

kernel/src/platform/mod.rs
luigix25: oops, yes :)

kernel/src/platform/native.rs
luigix25: much cleaner, thanks!

stefano-garzarella: > We can eventually create a function for that with a generic `T` (something similar also for the `mmio_write` where all the cases has the same pattern.

Why not doing this ^ ?

luigix25: > > We can eventually create a function for that with a generic `T` (something similar also for the `mmio_write` where all the cases has the same pattern.
> 
> Why not doing this ^ ?

done!

kernel/src/platform/native.rs
```diff
@@ -233,15 +233,113 @@ impl SvsmPlatform for NativePlatform {
...
+            }
+            _ => return Err(SvsmError::InvalidBytes),
+        };
+
+        Ok(())
     }
 
+    /// Perform a read from a memory-mapped IO area
+    ///
+    ///  This function expects reads to be 1, 2, 4 or 8 bytes long.
```
stefano-garzarella:
```suggestion
    /// This function expects reads to be 1, 2, 4 or 8 bytes long.
```

kernel/src/platform/native.rs
```diff
@@ -233,15 +233,113 @@ impl SvsmPlatform for NativePlatform {
...
+        Ok(())
+    }
+}
+
+unsafe fn mmio_write_type<T: Copy>(vaddr: VirtAddr, data: &[u8]) {
+    let ptr = vaddr.as_mut_ptr::<T>();
+
+    // SAFETY: We are trusting the caller to ensure validity of `vaddr` and alignment of data.
+    unsafe {
+        ptr.write_volatile(*(data.as_ptr() as *const T));
```
stefano-garzarella:
```suggestion
        ptr.write_volatile(*(data.as_ptr().cast::<T>));
```

I mean, both are fine, but please be consistent between the 2 functions you are adding.

kernel/src/platform/native.rs
```diff
@@ -233,15 +233,113 @@ impl SvsmPlatform for NativePlatform {
...
+    };
+}
+
+unsafe fn mmio_read_type<T>(vaddr: VirtAddr, data: &mut [MaybeUninit<u8>]) {
+    let ptr = vaddr.as_mut_ptr::<T>();
+
+    // SAFETY: We are trusting the caller to ensure validity of `vaddr` and alignment of data.
+    unsafe {
+        data.as_mut_ptr().cast::<T>().write(ptr.read_volatile());
```
stefano-garzarella: I think this is also unsafe because `data.as_mut_ptr().cast::<T>()` should be aligned (and valid).

So should we add a couple of checks for that here or in the caller?

I mean for example something like this for the alignment:
```
    let data_ptr = data.as_ptr().cast::<T>();

    if !data_ptr.is_aligned() {
        return Err(SvsmError::...);
    }
```

About the size, we can demand it to the caller or add another check here.

kernel/src/platform/native.rs
stefano-garzarella: Ditto here about `data` alignment.

kernel/src/platform/native.rs
stefano-garzarella: Or we should demand it to the `kernel/src/virtio/hal.rs`, but we need to document that and check it.

luigix25: Does it make sense to perform a mmio operation on an unaligned address? If not, then we can do the check in the platform's code so every operation is covered by the check. Otherwise we can enforce it just for virtio in `kernel/src/virtio/hal.rs`.

stefano-garzarella: The problem is not the mmio operation itself, is the `data` buffer provided by the caller that should be aligned if we use `T::write()` (see https://doc.rust-lang.org/core/primitive.pointer.html#method.write or https://doc.rust-lang.org/core/ptr/fn.write.html)

luigix25: uhm ok thanks. If it's ok for you I'd add this check in `native.rs`. imho in this way it's clear why this check is in place. No strong opinion though.

stefano-garzarella: @luigix25 ok, maybe you can check the alignment and use `ptr::write` or `ptr::write_unaligned()` depending on it

luigix25: What if we always use `write_unaligned`?

luigix25:
Added a check on `data_ptr` alignment and use read/write_aligned() or read()/write()

kernel/src/platform/native.rs
stefano-garzarella: I guess performances will be affected. In the end I think will be like we did before, copying byte by byte.

luigix25: ok thanks. I updated the PR using write_unaligned or write() according to the alignment.

osteffenrh (approved):
Tested it on top of #823 :heavy_check_mark:

stefano-garzarella (approved):
In the last TSC meeting we discussed this PR and decided to merge, but for long term we should improve our MMIO interface. Of course this is outside the scope of this PR so we can postpone it.

I'm approving and wait @joergroedel to take a final look before merging it.

PR #776, opened 2025-08-05
CI: remove custom steps and increase test matrix

stefano-garzarella:
This contains the last 3 patches originally contained in #764 

The goal is to use the targets of our Makefile as much as possible, so we have a single point to maintain.
Also add more steps to build the release image, test it, and use `--all-features` to test the build with all features enabled.


stefano-garzarella:
CI without this series: https://github.com/coconut-svsm/svsm/actions/runs/16754397261/job/47432898524
`5m 34s `

CI with this series: https://github.com/coconut-svsm/svsm/actions/runs/16755476512/job/47436866938?pr=776
`7m 57s `

```
Build debug image with all features     32s
Build release image                     48s

Run tests on release image            1m 8s
```

So around `2m 30s` more added by the last commit. Not sure if we want this or not, I can remove that commit.

PR #780, opened 2025-08-06
Makefile: add native flag for qemu builds

luigix25:
The IGVM file produced by `make` does not have the native flag set. This means that `launch_guest [..] --nocc` will not work.

Add `native` option to qemu and test-qemu


stefano-garzarella:
IIUC we already enable it in `configs/qemu-target.json`, so I'd mention that to make it clear that we are aligning the two.

In the commit description, I'd replace `launch_guest [..] --nocc` with `scripts/launch_guest.sh --nocc` and I'd also try to truncate lines around 75 cols like we usually do.

stefano-garzarella:
@joergroedel any objection on merging this?
IIRC we just forgot to enable it after QEMU started to support `native`

PR #781, opened 2025-08-07
igvmbuilder/ovmf: Rework using slices and zerocopy

peterfang:
Rework the OVMF parser in `igvmbuilder`. Take advantage of Rust slice's properties and the `zerocopy` crate and reduce code redundancy.

### DIFF
--- a/igvmbuilder/src/ovmf_firmware.rs
+++ b/igvmbuilder/src/ovmf_firmware.rs
@@ -14,6 +14,7 @@ use bootlib::igvm_params::{IgvmGuestContext, IgvmParamBlockFwInfo};
 use igvm::IgvmDirectiveHeader;
 use igvm_defs::{IgvmPageDataFlags, IgvmPageDataType, PAGE_SIZE_4K};
 use uuid::Uuid;
+use zerocopy::{FromBytes, KnownLayout};
 
 use crate::firmware::Firmware;
 use crate::igvm_builder::{NATIVE_COMPATIBILITY_MASK, SNP_COMPATIBILITY_MASK};
@@ -21,33 +22,36 @@ use crate::igvm_builder::{NATIVE_COMPATIBILITY_MASK, SNP_COMPATIBILITY_MASK};
 // Offset from the end of the file where the OVMF table footer GUID should be.
 const FOOTER_OFFSET: usize = 32;
 
+#[derive(FromBytes, KnownLayout)]
+#[repr(C)]
+struct GuidBlockFooter {
+    len: u16,
+    guid: [u8; 16],
+}
+
+#[derive(FromBytes, KnownLayout)]
+#[repr(C)]
+struct GuidBlockMetadata {
+    offset_from_end: u32,
+}
+
+#[derive(FromBytes, KnownLayout)]
+#[repr(C)]
+struct GuidBlockResetVector {
+    vector_address: u32,
+    compatibility_mask: u32,
+}
+
+#[derive(FromBytes, KnownLayout)]
+#[repr(C)]
 struct SevMetadataEntry {
     base: u32,
     len: u32,
     metadata_type: u32,
 }
 
-impl SevMetadataEntry {
-    pub fn size() -> usize {
-        size_of::<Self>()
-    }
-}
-
-impl TryFrom<&[u8]> for SevMetadataEntry {
-    type Error = Box<dyn Error>;
-
-    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
-        if value.len() < Self::size() {
-            return Err("Cannot parse OVMF metadata entry - invalid buffer size".into());
-        }
-        Ok(Self {
-            base: read_u32(&value[0..4])?,
-            len: read_u32(&value[4..8])?,
-            metadata_type: read_u32(&value[8..12])?,
-        })
-    }
-}
-
+#[derive(FromBytes, KnownLayout)]
+#[repr(C)]
 struct MetadataDesc {
     _sig: u32,
     _len: u32,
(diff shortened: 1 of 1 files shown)


osteffenrh (approved):

igvmbuilder/src/ovmf_firmware.rs
```diff
@@ -234,21 +227,21 @@ impl OvmfFirmware {
...
-        let len = in_file.metadata()?.len() as usize;
-        if len > 0xffffffff {
-            return Err("OVMF firmware is too large".into());
-        }
-        let mut data = Vec::with_capacity(len);
-        if in_file.read_to_end(&mut data)? != len {
+        // Must fit within the 32-bit address space.
+        let len =
+            u32::try_from(in_file.metadata()?.len()).map_err(|_| "OVMF firmware is too large")?;
+        let len = NonZeroUsize::new(len as usize).ok_or("OVMF firmware is empty")?;
```
stefano-garzarella: Why using `NonZeroUsize`?
I mean, we use `len` only in this function, so can we just check if it's 0 and return the error in that case, so we don't need to call `.get()`

igvmbuilder/src/ovmf_firmware.rs
```diff
@@ -77,146 +80,136 @@ impl TryFrom<&[u8]> for SevMetadata {
...
-        Err("Invalid buffer passed to read_u16".into())
-    } else {
-        Ok(data[0] as u16 + ((data[1] as u16) << 8))
-    }
+    let val = u16::from_le_bytes(
+        data.try_into()
+            .map_err(|_| "Invalid buffer passed to read_u16")?,
+    );
+    Ok(val)
 }
```
stefano-garzarella: Next commit is removing these functions, so maybe we can just avoid touching them.

igvmbuilder/src/ovmf_firmware.rs
```diff
@@ -77,146 +80,136 @@ impl TryFrom<&[u8]> for SevMetadata {
...
+        &data[..data.len() - table_size],
+    ))
 }
 
 fn parse_sev_metadata(
     data: &[u8],
-    table_data_offset: usize,
-    firmware: &mut IgvmParamBlockFwInfo,
+    fw_img: &[u8],
+    fw_info: &mut IgvmParamBlockFwInfo,
```
stefano-garzarella, line 82: nit: I'd move this change in a separate commit, since several lines are just renaming, this should make the patch easy to review, bisect, backport, etc.

igvmbuilder/src/ovmf_firmware.rs
```diff
@@ -234,21 +227,21 @@ impl OvmfFirmware {
...
         let mut fw_info = IgvmParamBlockFwInfo {
             // OVMF is located to end at 4GB by default.
-            start: (0xffffffff - len + 1) as u32,
-            size: len as u32,
+            start: ((1usize << 32) - len.get()) as u32,
+            size: len.get() as u32,
             ..Default::default()
         };
-        // OVMF is not compatible with TDP by default
+        // OVMF is not compatible with TDP by default.
```
stefano-garzarella, line 210: nit: unrelated change

igvmbuilder/src/ovmf_firmware.rs
```diff
@@ -122,31 +82,34 @@ fn parse_sev_metadata(
...
+    let (desc, mut buf) = MetadataDesc::read_from_prefix(&fw_img[offset..])
+        .map_err(|e| format!("Cannot parse OVMF metadata descriptor: {e}"))?;
 
-    for i in 0..metadata.num_desc as usize {
-        let desc_offset = offset + SevMetadata::size() + i * MetadataDesc::size();
-        let metadata_desc =
-            MetadataDesc::try_from(&fw_img[desc_offset..desc_offset + MetadataDesc::size()])?;
-        match metadata_desc.metadata_type {
+    for _ in 0..desc.num_desc as usize {
+        let (entry, remainder) = SevMetadataEntry::read_from_prefix(buf)
```
stefano-garzarella, line 95: Can we keep `metadata_desc` name to reduce the diff size?

igvmbuilder/src/ovmf_firmware.rs
stefano-garzarella: Or just call `.get()` here:
```suggestion
        let len = NonZeroUsize::new(len as usize).ok_or("OVMF firmware is empty")?.get();
```

but again I don't understand why not just checking if it's 0, I may miss something.

peterfang: I agree, there's not a lot of value add from using `NonZeroUsize`. I wanted to make sure it never refers to an empty file because that would overflow `IgvmParamBlockFwInfo.start`. I now agree it makes the code less readable. I'll change it back to `usize` and just check if it's 0.

igvmbuilder/src/ovmf_firmware.rs
peterfang: On rare occasions commits get reverted, so maybe we can keep this change? I did test it to make sure it works as expected.

igvmbuilder/src/ovmf_firmware.rs
peterfang: Yep, will move non-functional changes to a separate commit.

igvmbuilder/src/ovmf_firmware.rs
peterfang: Will move to a separate commit.

igvmbuilder/src/ovmf_firmware.rs
peterfang: I'd prefer to move the name changes to a separate commit. I'm trying to match the names to the OVMF code:
`SevMetadata` -> generic metadata descriptor
`MetadataDesc` -> an entry within the metadata region

igvmbuilder/src/ovmf_firmware.rs
stefano-garzarella: @peterfang I see, sure, let's keep it!

igvmbuilder/src/ovmf_firmware.rs
stefano-garzarella: @peterfang sure, this is also fine by me

PR #782, opened 2025-08-07
kernel/virtio: avoid zeroing and extra copying the destination buffer

stefano-garzarella:
While reviewing #775 I figured out that `SvsmHal::mmio_read()` could be improved a bit.

`SvsmHal::mmio_read()` currently allocates the destination buffer, zeros it, and then copies its value. Avoid all this by allocating `MaybeUninit::<T>` on the stack and passing a pointer to `SVSM_PLATFORM.mmio_read()`.

Cascade the changes to the functions to use a pointer, since the destination buffer may be uninitialized.

### DIFF
--- a/kernel/src/platform/mod.rs
+++ b/kernel/src/platform/mod.rs
@@ -20,6 +20,7 @@ use tdp::TdpPlatform;
 
 use core::arch::asm;
 use core::fmt::Debug;
+use core::mem::MaybeUninit;
 
 use crate::address::{PhysAddr, VirtAddr};
 use crate::config::SvsmConfig;
@@ -224,9 +225,13 @@ pub trait SvsmPlatform: Sync {
     ///
     /// # Safety
     ///
-    /// Caller must ensure that `pa` points to a properly aligned memory location and the
+    /// Caller must ensure that `paddr` points to a properly aligned memory location and the
     /// memory accessed is part of a valid MMIO range.
-    unsafe fn mmio_read(&self, _paddr: PhysAddr, _data: &mut [u8]) -> Result<(), SvsmError>;
+    unsafe fn mmio_read(
+        &self,
+        paddr: PhysAddr,
+        data: &mut [MaybeUninit<u8>],
+    ) -> Result<(), SvsmError>;
 }
 
 //FIXME - remove Copy trait
--- a/kernel/src/virtio/hal.rs
+++ b/kernel/src/virtio/hal.rs
@@ -9,6 +9,7 @@ use crate::{locking::SpinLock, platform::SVSM_PLATFORM};
 use alloc::vec::Vec;
 use core::{
     cell::OnceCell,
+    mem::{size_of, MaybeUninit},
     ptr::{addr_of, NonNull},
 };
 use zerocopy::{FromBytes, Immutable, IntoBytes};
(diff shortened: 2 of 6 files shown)


kernel/src/platform/mod.rs
```diff
@@ -224,9 +224,11 @@ pub trait SvsmPlatform: Sync {
...
     /// # Safety
     ///
-    /// Caller must ensure that `pa` points to a properly aligned memory location and the
-    /// memory accessed is part of a valid MMIO range.
-    unsafe fn mmio_read(&self, _paddr: PhysAddr, _data: &mut [u8]) -> Result<(), SvsmError>;
+    /// Caller must ensure that `paddr` points to a properly aligned memory location and the
+    /// memory accessed is part of a valid MMIO range. The `buf` pointer must be valid for
+    /// writes of `size` bytes.
+    unsafe fn mmio_read(&self, paddr: PhysAddr, buf: *mut u8, size: usize)
+        -> Result<(), SvsmError>;
```
joergroedel: Can we pass a reference to the `MaybeUnit<[u8]>`instead? That would make the API use safer than passing raw pointers. It puts more responsibilities on the caller, but if it uses `MaybeUnit` anyway that is not a problem.

stefano-garzarella: We can't use `MaybeUnit<[u8]>`, because `Sized` trait is required:
```
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> kernel/src/platform/mod.rs:230:59
    |
230 |     unsafe fn mmio_read(&self, paddr: PhysAddr, buf: &mut MaybeUninit<[u8]>)
    |                                                           ^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `[u8]`
note: required by an implicit `Sized` bound in `core::mem::MaybeUninit`
```

But we can use `[MaybeUninit<u8>]`, let me try that.

stefano-garzarella:
@joergroedel done, the results it's much better, thanks for the idea!

I don't like much the `MaybeUninit::<T>` -> `[MaybeUninit<u8>]` cast I did. With `MaybeUninit::as_bytes_mut()` is much cleaner, but it's still unstable.

kernel/src/virtio/hal.rs
```diff
@@ -202,13 +203,21 @@ unsafe impl virtio_drivers::Hal for SvsmHal {
             .phys_addr(VirtAddr::from(addr_of!(*src)))
             .unwrap();
 
-        let mut b = alloc::vec![0; size_of::<T>()];
+        let mut b = MaybeUninit::<T>::uninit();
         // SAFETY: We are trusting the caller (the virtio driver) to ensure `src` is a valid MMIO
-        // address and that it is alinged properly.
+        // address and that it is alinged properly. If SVSM_PLATFORM.mmio_read() doesn't fail
```
luigix25: nit: `alinged`

kernel/src/virtio/hal.rs
```diff
@@ -202,13 +203,21 @@ unsafe impl virtio_drivers::Hal for SvsmHal {
...
         unsafe {
-            SVSM_PLATFORM.mmio_read(paddr, &mut b).unwrap();
+            // MaybeUninit::as_bytes_mut() can avoid this, but it's still
+            // unstable. When it will be stabilized, we can simply use
+            // `b.as_bytes_mut()` instead of creating `b_slice`.
+            let b_slice = core::slice::from_raw_parts_mut(
+                b.as_mut_ptr().cast::<MaybeUninit<u8>>(),
+                size_of::<T>(),
+            );
+            SVSM_PLATFORM.mmio_read(paddr, b_slice).unwrap();
```
luigix25, line 218: WDYT about using `expect`?

stefano-garzarella: It's pre-existing, we should fix in another PR/commit

kernel/src/virtio/hal.rs
stefano-garzarella: this is also pre-existing, but since I'm touching that comment, I can fix here

stefano-garzarella: mmm, we have that in several places. I'll open another PR for that:

```
$ git grep alinged

kernel/src/virtio/hal.rs:    /// `src` must be properly alinged and reside at a readable memory address.
kernel/src/virtio/hal.rs:        // address and that it is alinged properly. If SVSM_PLATFORM.mmio_read() doesn't fail
kernel/src/virtio/hal.rs:    /// `dst` must be properly alinged and reside at a writable memory address.
virtio-drivers/src/hal.rs:    /// `src` must be properly alinged and reside at a readable memory address.
virtio-drivers/src/hal.rs:    /// `dst` must be properly alinged and reside at a writable memory address.
virtio-drivers/src/hal.rs:        // SAFETY: dst is assumed to be properly alinged and within writeble memory
```

luigix25: maybe a spellchecker in the CI could be useful. At least for the commit messages

stefano-garzarella: @luigix25 yep, good point, if you have any idea, please submit a PR

stefano-garzarella: PR opened here: https://github.com/coconut-svsm/svsm/pull/804

joergroedel (approved):

PR #786, opened 2025-08-15
cpu/apic: Fix handle_icr_write

MelodyHuibo:
SVSM has been checking bits of trigger mode and assert in ICR message for IPI, however the Hardware ignores those bits and this is how Linux has been doing it since forever.

Fix handle_icr_write() to ignore those bits just like real hardware does.


peterfang:
Please fix the formatting issue and re-submit. Thanks!

peterfang (approved):

Issue #779, opened 2025-08-06
Configuring SVSM features via build recipes is broken


joergroedel:
Currently the build fails when additional features are specified in a build recipe. Tried this with changing the `"vtpm"` string to `"vtpm,attest"`.

stefano-garzarella:
@joergroedel I just tried `"features": "vtpm,enable-gdb",` and it worked, so maybe it's more related to the `attest` feature itself, or how we build `aproxy` (just guessing)

joergroedel:
> [@joergroedel](https://github.com/joergroedel) I just tried `"features": "vtpm,enable-gdb",` and it worked, so maybe it's more related to the `attest` feature itself, or how we build `aproxy` (just guessing)

Interesting, I will do more investigation. Specifying `attest` spits out a big pile of errors. But it has to wait until after the release.

00xc:
This is not due to xbuild. It has something to do with how cargo decides to resolve dependencies. For some reason, when building the SVSM kernel as a binary (`cargo build --bin svsm`, or `binary: true` in the recipe), the dependency feature resolution is different than when building as a package (`cargo build -p svsm` or `binary: false`). Somehow this causes cargo to attempt to build serde with the `std` feature enabled.

This fixes it for me:

```diff
diff --git a/configs/qemu-target.json b/configs/qemu-target.json
index 4bad575a..6878396f 100644
--- a/configs/qemu-target.json
+++ b/configs/qemu-target.json
@@ -14,8 +14,8 @@
     },
     "kernel": {
         "svsm": {
-            "features": "vtpm",
-            "binary": true
+            "features": "vtpm,attest",
+            "binary": false
         },
         "stage2": {
             "manifest": "kernel/Cargo.toml",
```

NOCC boot seems to work. Could someone check on an SEV/TDX host?

v-thakkar:
I checked this solution and it works on SEV-SNP for me.

PR #787, opened 2025-08-18
configs: build SVSM as package

00xc:
Do not build the SVSM kernel as a binary (`--bin`) but as a package instead (`--package`) when invoking cargo. This needs to be done because building as a binary uses different dependency resolution logic in cargo, which in the end may end up building dependencies for a std target, causing breakage in builds.

Fixes: #779


joergroedel (approved):
Thanks a lot, this fixes the build problem.

Issue #683, opened 2025-04-29
GIT_VERSION reporting require clean rebuild


osteffenrh:
The git revision reported by SVSM can be incorrect, if SVSM was not built from a clean checkout.

for example
```
$ make
$ git checkout $OLD_REV
$ make
```
will report a wrong git revision in SVSM's output.

To get the correct revision reported, one needs to remove the build files,
for example by `git clean -xdf` or alike.

joergroedel:
Thanks for the report, I will investigate.

PR #792, opened 2025-08-22
release, xbuild: Move version string generation to xbuild

joergroedel:
Generating the version string in release/build.rs is error prone because the build.rs is not re-run on every build, only when the release crate itself has changes. Changes outside of the release crate will not be captured in the version string.

Fix this problem by moving the generation of the version string file into the xbuild utility, which runs on every build and is a much better place to generate accurate version strings.

Fixes: #683

**Update:**

This got a bit bigger now, but the `Makefile` targets work again. In addition there are more cleanups around the Github workflows and the Makefile, which now use `cargo xbuild` where possible. The `cargo xbuild` utility was extended to allow specifying package features without editing the build recipes.


joergroedel:
This PR breaks `make` builds on clean repositories. I will fix that and update this PR.

xbuild/src/main.rs
stefano-garzarella, line 1: nit: typo in the https://github.com/coconut-svsm/svsm/pull/792/commits/b5d943542912dd3d25b1a67b934c9b90cfc02faa description `s/feathre/feature`

xbuild/src/main.rs
```diff
@@ -174,12 +174,18 @@ impl ComponentConfig {
...
                     if let Some((a, b)) = f.split_once(':') {
                         if pkg == a {
+                            // Package specific feature
                             Some(String::from(b.trim()))
                         } else {
                             None
                         }
                     } else {
-                        None
+                        // Features without package prefix are only used for 'svsm'
```
stefano-garzarella: Should we also update the doc comment on top of the `--feature/-f option` ?

Makefile
stefano-garzarella, line 1: nit: typo in the https://github.com/coconut-svsm/svsm/pull/792/commits/cb6f678878de88468ee3b9d75979934d24a65697 description: s/an/and

Makefile
```diff
@@ -187,9 +187,6 @@ bin/stage1-trampoline: stage1_elf_trampoline
 bin/svsm-test: stage1_elf_test
 	cp -f $(STAGE1_ELF) $@
 
-bin/svsm.bin: bin/svsm
```
stefano-garzarella, line 191: I can't figure out if these changes are related to this commit or if they just accidentally ended up here.
If they are related, maybe we should explain why it's no longer needed.

stefano-garzarella (commented):
I left some small comments, but overall LGTM.

A small observation related to the `--features` option.
I tried the following commands:
```
cargo xbuild -f svsm: configs/qemu-target.json
cargo xbuild -f foo: configs/qemu-target.json
cargo xbuild -f foo:foo configs/qemu-target.json
```

All of them will not fail. Is it okay or should we add some checks on packages?

stefano-garzarella:
> >   
> > All of them will not fail. Is it okay or should we add some checks on packages and empty features?
> 
> It makes sense to fail on empty features, but not sure whether we should fail on non-existing packages?

yeah, I see, if we don't want to fail, maybe we can print a warning that we didn't set the feature since the package doesn't exist. Just to be sure that the user is aware that what they specified in the cmdline is something wrong (e.g. prevent typos, etc.)

BTW it's not a strong opinion, up to you ;-)

xbuild/src/main.rs
joergroedel: Added, thanks.

Makefile
joergroedel: Fixed, thanks.

Makefile
joergroedel: The reason is that `svsm.bin` can't be built with `xbuild`, though `xbuild` is required to compile the SVSM kernel going forward (to generate `git_version.rs`). As `svsm.bin` is for non-IGVM boots only, which is not supported anymore, it is removed here. After this series it would be broken anyway.

xbuild/src/main.rs
```diff
@@ -190,7 +190,6 @@ impl ComponentConfig {
                 .clone()
                 .map(|feat| feat.split(',').map(|f| f.trim().to_string()).collect())
                 .unwrap_or_default();
-
```
stefano-garzarella: Something went wrong with this commit https://github.com/coconut-svsm/svsm/pull/792/commits/173ed1440802fc719f3c1891f461b611891eb370 since it's only removing a blank line.

Makefile
stefano-garzarella: Okay, I see now, thanks!

xbuild/src/version.rs
```diff
@@ -0,0 +1,50 @@
...
+        .map_err(|_| ())?;
+    if !output.status.success() {
+        return Err(());
+    }
+
+    let stdout = String::from_utf8(output.stdout).unwrap();
+    let mut lines = stdout.lines();
+    let first_line = lines.next().unwrap().trim();
+    // Strip leading `v` from git tag
+    let version = first_line.strip_prefix("v").unwrap_or(first_line);
```
stefano-garzarella: nit: This function returns a `Result` so what about avoiding `unwrap()`?

xbuild/src/features.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn create_from_args(args: &Args) -> Self {
+        let list = args
+            .features
+            .iter()
+            .map(|f| CmdLineFeature::create_from_raw(String::from(f)))
+            .collect();
+        Self { list }
+    }
+
+    pub fn check_empty_features(&self) {
```
stefano-garzarella: Following the next function, maybe we can call this `print_empry_features`.

stefano-garzarella (commented):
@joergroedel thanks! I left some minor comments, no need to address all of them, some are just suggestions while reading code.

xbuild/src/version.rs
joergroedel: A missing `v` prefix is not an error that needs to be reported. If the prefix isn't there the function should just return first line of stdout.

xbuild/src/features.rs
joergroedel: Yes, makes sense. I will change that.

xbuild/src/main.rs
joergroedel: Thanks, I will remove the patch.

xbuild/src/version.rs
stefano-garzarella: Sorry, I meant at line 24 and 26.

joergroedel: Okay, that makes sense. I updated to code.

PR #794, opened 2025-08-26
Update IGVM build and usage in INSTALL.md

RARelph:
IGVM repo now requires cbuild and installation of IGVM system-wide now requires su access to the rust toolchain, which is uncommon. Update our installation instructions to install cbuild for the local account and build and use IGVM from a local directory.


stefano-garzarella:
@RARelph I just merged #791 that did something similar (at least for part of it) of this PR. Can you rebase and update the commit/PR description, since the cbuild dep should be already fixed?

PR #798, opened 2025-09-01
kernel/protocol/attest: Add extended attestation support

IT302:
Add support for requesting an extended attestation package (report + manifest + certificate bundle) for SVSM_ATTEST_SERVICES and SVSM_ATTEST_SINGLE_SERVICE.

The certificate bundle is optional and may be supplied by the host. If unavailable, only the report and manifest are returned.

This enables consumers to obtain a self-contained evidence set without extra host round trips. Linux requests the extended attestation report by default.

### DIFF
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -11,6 +11,7 @@ extern crate alloc;
 use crate::address::{Address, PhysAddr};
 use crate::crypto::digest::{Algorithm, Sha512};
 use crate::error::{AttestError, SvsmError};
+use crate::greq::services::get_extended_report;
 use crate::greq::{
     pld_report::{SnpReportRequest, SnpReportResponse},
     services::get_regular_report,
@@ -25,6 +26,8 @@ use alloc::{boxed::Box, vec::Vec};
 use uuid::{uuid, Uuid};
 use zerocopy::{FromBytes, FromZeros, Immutable, IntoBytes, KnownLayout};
 
+use crate::sev::ghcb::GhcbError;
+
 pub const ATTEST_PROTOCOL_VERSION_MIN: u32 = 1;
 pub const ATTEST_PROTOCOL_VERSION_MAX: u32 = 1;
 
@@ -35,6 +38,18 @@ const SVSM_ATTEST_SINGLE_SERVICE: u32 = 1;
 
 #[cfg(all(feature = "vtpm", not(test)))]
 const SVSM_ATTEST_VTPM_GUID: Uuid = uuid!("c476f1eb-0123-45a5-9641-b4e7dde5bfe3");
+
+// According to
+// https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/drivers/virt/coco/sev-guest/sev-guest.c#L370
+// https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/include/linux/psp-sev.h#L17
+// the maximum size of the certificate buffer is 0x4000 bytes (16K).
+// Could not find the  maximum size of the certificate buffer in the
+// "Secure VM Service Module for SEV-SNP Guests, Revision 1.0",
+// "SEV Secure Nested Paging Firmware ABI Specification, Revision 1.58"
+// or "SEV-ES Guest-Hypervisor Communication Block Standardization, Revision 2.04"
+// If future specifications update the maximum certificate buffer size, update the
+// MAX_CERTIFICATE_SIZE constant below to match the new value.
+const MAX_CERTIFICATE_SIZE: usize = 0x4000;
 
 // Attest services operation structure, as defined in Table 11 of Secure VM Service Module for
 // SEV-SNP Guests 58019 Rev, 1.00 July 2023
@@ -115,6 +130,44 @@ impl AttestServicesOp {
         }
 
         Ok(MemoryRegion::new(gpa, size))
+    }
+
+    /// Returns an optional MemoryRegion describing the certificate buffer.
+    ///
+    /// Behaviour:
+    /// - certificate_size == 0 -> Ok(None)
+    ///   (caller did not request certificates / extended report)
+    /// - 0 < certificate_size <= MAX_CERTIFICATE_SIZE and certificate_gpa page aligned
+    ///   -> Ok(Some(MemoryRegion))
+    /// - Otherwise -> Err(SvsmReqError::invalid_parameter())
+    ///
+    /// Notes:
+    /// - Buffer may span multiple pages but must be physically contiguous.
+    /// - Buffer GPA must be page aligned.
+    /// - Size is capped at MAX_CERTIFICATE_SIZE (0x4000 bytes).
+    pub fn get_certificate_region(&self) -> Result<Option<MemoryRegion<PhysAddr>>, SvsmReqError> {
+        let size = usize::try_from(self.certificate_size)
+            .map_err(|_| SvsmReqError::invalid_parameter())?;
+
+        // If size is 0, the certificate buffer is not present.
+        // This is valid and shall not return an error.
+        // It is used to indicate that the user does not want an extended attestation
+        // that returns certificates.
+        if size == 0 {
+            return Ok(None);
+        }
+
+        // Ensure that size is not greater than MAX_CERTIFICATE_SIZE
+        if size > MAX_CERTIFICATE_SIZE {
+            return Err(SvsmReqError::invalid_parameter());
+        }
+
+        let gpa = PhysAddr::from(self.certificate_gpa);
+        if !gpa.is_page_aligned() {
+            return Err(SvsmReqError::invalid_parameter());
+        }
+
+        Ok(Some(MemoryRegion::new(gpa, size)))
     }
 }
 
@@ -237,6 +290,23 @@ impl AttestSingleServiceOp {
     pub fn get_guid(&self) -> uuid::Uuid {
         Uuid::from_bytes_le(self.guid)
     }
+
+    /// See [`AttestServicesOp::get_certificate_region`] for details.
+    pub fn get_certificate_region(&self) -> Result<Option<MemoryRegion<PhysAddr>>, SvsmReqError> {
+        self.op.get_certificate_region()
+    }
+
+    /// Returns true if an extended report is requested
+    /// Extended report is requested by providing a certificate buffer gpa and
+    /// providing a non-zero certificate buffer size.
+    /// See [`AttestServicesOp::get_certificate_region`] for details.
+    pub fn is_extended_report(&self) -> Result<bool, SvsmReqError> {
+        // If certificate buffer size is greater than 0, it is an extended report.
+        // else it is a regular report.
+        let size = usize::try_from(self.op.certificate_size)
+            .map_err(|_| SvsmReqError::invalid_parameter())?;
+        Ok(size > 0)
+    }
 }
 
 fn get_attestation_report(nonce: &[u8]) -> Result<Box<SnpReportResponse>, SvsmReqError> {
@@ -255,6 +325,35 @@ fn get_attestation_report(nonce: &[u8]) -> Result<Box<SnpReportResponse>, SvsmRe
     let _response_size = get_regular_report(resp_buffer)?;
 
     Ok(resp)
+}
+
+fn get_attestation_report_extended(
+    nonce: &[u8],
+) -> Result<(Box<SnpReportResponse>, Box<[u8; MAX_CERTIFICATE_SIZE]>), SvsmReqError> {
+    let mut resp = SnpReportResponse::new_box_zeroed()
+        .map_err(|_| SvsmReqError::FatalError(SvsmError::Mem))?;
+    let resp_buffer = resp.as_mut_bytes();
+    // Cast error is infallibly discarded.
+    let (report_req, _) = SnpReportRequest::mut_from_prefix(resp_buffer)
+        .map_err(|_| SvsmReqError::invalid_parameter())?;
+
+    // Extended report requires a certificate buffer.
+    // Allocate a zero initialized buffer for certificates.
+    // MAX_CERTIFICATE_SIZE is defined as 0x4000 bytes (16k).
+    let mut certs = <[u8; MAX_CERTIFICATE_SIZE]>::new_box_zeroed()
+        .map_err(|_| SvsmReqError::FatalError(SvsmError::Mem))?;
+    let certs_buffer = certs.as_mut_bytes();
+
+    // Zero initialized, so
+    // vmpl=0
+    // flags=1: Use VCEK.
+    report_req.user_data = nonce
+        .try_into()
+        .map_err(|_| SvsmReqError::invalid_parameter())?;
+
+    let _response_size = get_extended_report(resp_buffer, certs_buffer)?;
+
+    Ok((resp, certs))
 }
 
 fn write_report_and_manifest(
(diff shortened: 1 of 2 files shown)


stefano-garzarella:
@IT302 thanks for the PR. How can we test this PR and how this is related to #719 ?

IT302:
> @IT302 thanks for the PR. How can we test this PR and how this is related to #719 ?

This commit enables extended attestation by including (when present on the host) the additional certificate material needed to verify the report, as described in Section 7.2 of the “Secure VM Service Module for SEV-SNP Guests” specification. This is unrelated to issue #719 .

Current problem: The existing implementation ignores requests for extended attestation and always performs a standard attestation flow. That is technically incorrect: the request parameters should decide whether the SVSM returns only the basic attestation report and manifest (standard) or the report, manifest plus the host-provided certificate objects (extended).

What this PR changes:

Detects when the caller requests extended attestation.
If certificates are available on the host, they are copied into (and returned through) the certificate buffer provided.
Falls back gracefully: if the certificates are not present, the attestation still succeeds, but only the standard components are returned. The SVSM will log a warning in that case.

How to test with Linux TSM (example): (report is a configfs TSM report directory)

```
report=/sys/kernel/config/tsm/report/report0
mkdir -p  $report 
echo svsm > $report/service_provider 
echo c476f1eb-0123-45a5-9641-b4e7dde5bfe3 > $report/service_guid
dd if=/dev/urandom bs=64 count=1 of=$report/inblob
hexdump -C $report/outblob 
ls -l $report/
```

References
https://github.com/virtee/snpguest/blob/main/README.md#extended-attestation-flowchart

kernel/src/protocols/attest.rs
```diff
@@ -257,6 +327,35 @@ fn get_attestation_report(nonce: &[u8]) -> Result<Box<SnpReportResponse>, SvsmRe
...
+    // Extended report requires a certificate buffer.
+    // Allocate a zero initialized buffer for certificates.
+    // MAX_CERTIFICATE_SIZE is defined as 0x4000 bytes (16k).
+    let mut certs = <[u8; MAX_CERTIFICATE_SIZE]>::new_box_zeroed()
+        .map_err(|_| SvsmReqError::FatalError(SvsmError::Mem))?;
+    let certs_buffer = certs.as_mut_bytes();
+
+    // Zero initialized, so
+    // vmpl=0
+    // flags=1: Use VCEK.
```
stefano-garzarella: Why here we use flags=1 and in `get_attestation_report()` flags=0 ?

kernel/src/protocols/attest.rs
```diff
@@ -326,10 +477,36 @@ fn attest_single_service(
...
-
-    write_report_and_manifest(manifest, params, &ops.op, resp.report.as_bytes())
+    if ops.is_extended_report()? {
+        // If the report is an extended report, it means that the caller requested a certificate.
+        // Get extended attestation report from PSP with Sha512(nonce||manifest) as REPORT_DATA.
+        // Handle SvsmReqError::FatalError(SvsmError::Ghcb(GhcbError::VmgexitError(certs_buffer_size,psp_rc,)))
+        // from the PSP indicating that the certificate buffer is too small.
+        // The required size is in certs_buffer_size.
+        // Per spec, return required size in rdx register and raise an error.
+        let (resp, certs) = match get_attestation_report_extended(hash.as_slice()) {
```
stefano-garzarella: In the commit message we mention

> Add support for requesting an extended attestation package
    (report + manifest + certificate bundle) for SVSM_ATTEST_SERVICES and
    SVSM_ATTEST_SINGLE_SERVICE.

But we call `get_attestation_report_extended()` only in `attest_single_service()`, so how we support SVSM_ATTEST_SERVICES?

IT302:
Added code to support SVSM_ATTEST_SERVICES. Removed an erroneous comment regarding flags and refactored the implementation by extracting the logic into a separate function.

kernel/src/protocols/attest.rs
```diff
@@ -36,6 +39,18 @@ const SVSM_ATTEST_SINGLE_SERVICE: u32 = 1;
...
+// https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/drivers/virt/coco/sev-guest/sev-guest.c#L370
+// https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/include/linux/psp-sev.h#L17
+// the maximum size of the certificate buffer is 0x4000 bytes (16K).
+// Could not find the  maximum size of the certificate buffer in the
+// "Secure VM Service Module for SEV-SNP Guests, Revision 1.0",
+// "SEV Secure Nested Paging Firmware ABI Specification, Revision 1.58"
+// or "SEV-ES Guest-Hypervisor Communication Block Standardization, Revision 2.04"
+// If future specifications update the maximum certificate buffer size, update the
+// MAX_CERTIFICATE_SIZE constant below to match the new value.
+const MAX_CERTIFICATE_SIZE: usize = 0x4000;
```
stefano-garzarella, line 53: @tlendacky @joergroedel should we specify this into the spec?

IT302:
> > Added code to support SVSM_ATTEST_SERVICES. Removed an erroneous comment regarding flags and refactored the implementation by extracting the logic into a separate function.
> 
> @IT302 thanks for the update, and sorry for the long delay.
> 
> BTW how can I test separately SVSM_ATTEST_SERVICES and SVSM_ATTEST_SINGLE_SERVICE ?
> 
> IIUC [#798 (comment)](https://github.com/coconut-svsm/svsm/pull/798#issuecomment-3334456211) is for SVSM_ATTEST_SINGLE_SERVICE, right?
To test SVSM_ATTEST_SERVICES, use the above but set service-guid to null

```
report=/sys/kernel/config/tsm/report/report0
mkdir -p  $report 
echo svsm > $report/service_provider 
echo 00000000-0000-0000-0000-000000000000 > $report/service_guid
dd if=/dev/urandom bs=64 count=1 of=$report/inblob
hexdump -C $report/outblob 
hexdump -C  $report/manifestblob
```

kernel/src/protocols/attest.rs
```diff
@@ -313,6 +416,96 @@ fn write_report_and_manifest(
...
+        return Err(SvsmError::Attestation(AttestError::CertificateSize).into());
+    }
+
+    copy_slice_to_guest(certs, cert_region.start())?;
+
+    // Set certificate size in bytes in rdx register
+    params.rdx = certs
+        .len()
+        .try_into()
+        .map_err(|_| SvsmError::Attestation(AttestError::Certificate))?;
```
stefano-garzarella, line 464: In `write_report_and_manifest()` we still have:
```
    // Does not support certificate currently, so setting certificate size to 0
    params.rdx = 0;
```

This is a bit confusing now, instead of having a new write_cert() function, should we extend `write_report_and_manifest()` to support also the cert writing, so we can write `edx` just one time?

kernel/src/protocols/attest.rs
```diff
@@ -313,6 +416,96 @@ fn write_report_and_manifest(
...
+        return Ok(());
+    }
+
+    // Check that certificates fit in the buffer. If too large,
+    // return an error indicating the buffer is too small .
+    if certs.len() > cert_region.len() {
+        return Err(SvsmError::Attestation(AttestError::CertificateSize).into());
+    }
+
+    copy_slice_to_guest(certs, cert_region.start())?;
```
stefano-garzarella, line 458: IIUC here we are writing the entire buffer (`MAX_CERTIFICATE_SIZE` bytes) in the guest memory, even if the PSP will return less. Is that okay?

kernel/src/protocols/attest.rs
stefano-garzarella: Or maybe we can call `write_cert()` in `write_report_and_manifest()`, but set edx in one point.

stefano-garzarella:
@IT302 I'm not really an expert on this topic but I'd like to merge this PR.
Can you provide steps that I need to do to provide certificate bundle from the host and check in the guest they are the same.

Would be nice to have a `test_in_svsm` to test it automatically, but for now a step by step should be okay.

kernel/src/protocols/attest.rs
IT302: As far as I can tell, PSP does not provide the size of the certificates, see https://github.com/coconut-svsm/svsm/blob/346f4862d73ef40c9c9cf7c05cf2c444dff2be3a/kernel/src/greq/driver.rs#L306. PSP may not even tell the actual size as the certificates are injected by the host. Rather than adding complex parsing logic to the SVSM kernel to determine the exact certificate size, we return the entire backing buffer to the caller. This is the safer approach since the certificates originate from the host (outside the SVSM trust boundary), and parsing untrusted data in the kernel increases attack surface.

stefano-garzarella: Yeah, since we are zeroing all the buffer, it should be fine.

kernel/src/protocols/attest.rs
IT302: Good catch — this was left over from when we didn’t handle extended attestation, so we don’t technically need it anymore

kernel/src/protocols/attest.rs
```diff
@@ -307,12 +410,99 @@ fn write_report_and_manifest(
         .try_into()
         .map_err(|_| SvsmError::Attestation(AttestError::Manifest))?;
 
-    // Does not support certificate currently, so setting certificate size to 0
-    params.rdx = 0;
```
stefano-garzarella, line 311: Sorry, I didn't mean to remove this, but to set here (on in another single place) on the right value.

Now for example, if the guest ask for the standard report, where we write it?

IMO we should avoid to spread rdx write all around a code, but have a single point.

IT302: I understand what you meant originally, but it turns out we don’t actually need it anymore as the PR adds extended attestation handling.

Background:

Per the SVSM spec, a “standard” attestation means the host does not provide a certificate buffer. In that case, the SVSM should not set the certificate buffer length in RDX.
The original implementation did not distinguish between standard and extended attestations and always wrote RDX = 0 to indicate there is no certificate buffer.

stefano-garzarella: Okay, so from spec:

> Upon successful completion of the SNP attestation request, the attestation report will be copied
to the input attestation report buffer gPA, the services manifest will be copied to the input
services manifest buffer gPA, RCX will be set to the size of the services manifest and, if a
certificate data buffer was provided, the certificate data will be copied to the input certificate
data buffer gPA and RDX will be set to the size of the certificate data.

What you're saying makes sense to me, but just to double-check with @tlendacky and @joergroedel is it correct to not set RDX if the certificate data buffer was not provided?

tlendacky: > Okay, so from spec:
> 
> > Upon successful completion of the SNP attestation request, the attestation report will be copied
> > to the input attestation report buffer gPA, the services manifest will be copied to the input
> > services manifest buffer gPA, RCX will be set to the size of the services manifest and, if a
> > certificate data buffer was provided, the certificate data will be copied to the input certificate
> > data buffer gPA and RDX will be set to the size of the certificate data.
> 
> What you're saying makes sense to me, but just to double-check with @tlendacky and @joergroedel is it correct to not set RDX if the certificate data buffer was not provided?

Correct, if the input certificate data buffer address is zero, then RDX is not updated. The idea is to allow the SVSM to execute an SNP Guest Request NAE event when a certificate data buffer has not been supplied vs a SNP Extended Guest Request NAE event when a certificate data buffer has been supplied.

stefano-garzarella: @tlendacky thanks for confirming!

tlendacky: I should qualify that better. The spec has:

> The certificate data buffer is optional. Its presence is indicated by setting the certificate data
buffer size to a non-0 value. If the certificate data buffer length is non-0, the certificate data
buffer will be treated as physically contiguous in the guest address space if the buffer size is
greater than 4 KB.

So, if a non-zero certificate buffer length is set, then the extended guest request is used and RDX updated.

kernel/src/protocols/attest.rs
stefano-garzarella: I opened https://github.com/coconut-svsm/svsm/issues/940 to move the discussion there. I think this PR can go in the while.

kernel/src/protocols/attest.rs
```diff
@@ -307,12 +410,99 @@ fn write_report_and_manifest(
...
+        let (resp, certs) = match get_attestation_report_extended(hash) {
+            Ok((resp, certs)) => (resp, certs),
+            Err(SvsmReqError::FatalError(SvsmError::Ghcb(GhcbError::VmgexitError(
+                certs_buffer_size,
+                _psp_rc,
+            )))) => {
+                // The PSP returned an error code indicating that the buffer size is too small.
+                // The required size is in certs_buffer_size.
+                // Per spec, return required size in rdx and raise an error
+                params.rdx = certs_buffer_size;
```
stefano-garzarella: Are we sure about this?

I mean in https://github.com/coconut-svsm/svsm/blob/346f4862d73ef40c9c9cf7c05cf2c444dff2be3a/kernel/src/sev/ghcb.rs#L767C1-L774C10 we have:

>         // On error, RBX and exit_info_2 are returned for proper error handling.
>         // For an extended request, if the buffer provided is too small, the hypervisor
>         // will return in RBX the number of contiguous pages required
>         if sw_exit_info_2 != 0 {
>             return Err(
>                 GhcbError::VmgexitError(self.rbx.load(Ordering::Relaxed), sw_exit_info_2).into(),
>             );
>         }


So not sure if they are in pages and not in bytes.

stefano-garzarella (commented):
@IT302 LGTM and I successfully tested, I have just one small question about error path.

kernel/src/protocols/attest.rs
IT302: The GHCB spec (section 4.1.8.1) specifies that this value represents the number of pages, not bytes. See: https://docs.amd.com/v/u/en-US/56421

I followed the existing naming convention used here for readability: https://github.com/coconut-svsm/svsm/blob/346f4862d73ef40c9c9cf7c05cf2c444dff2be3a/kernel/src/greq/services.rs#L103

It appears the variable was mislabeled as "bytes" at some point, but I didn't want to refactor code I'm not very familiar with.

Looking at the GHCB implementation, it does appear to return pages (not bytes): https://github.com/coconut-svsm/svsm/blob/346f4862d73ef40c9c9cf7c05cf2c444dff2be3a/kernel/src/sev/ghcb.rs#L767-L774

stefano-garzarella: @IT302 yep, it's exactly my point, it seems they are pages, but here we are using as bytes.

stefano-garzarella: > The GHCB spec (section 4.1.8.1) specifies that this value represents the number of pages, not bytes. See: https://docs.amd.com/v/u/en-US/56421
> 
> I followed the existing naming convention used here for readability:
> 
> https://github.com/coconut-svsm/svsm/blob/346f4862d73ef40c9c9cf7c05cf2c444dff2be3a/kernel/src/greq/services.rs#L103
> 
> It appears the variable was mislabeled as "bytes" at some point, but I didn't want to refactor code I'm not very familiar with.

@cclaudio do you remember something about this?

> 
> Looking at the GHCB implementation, it does appear to return pages (not bytes):
> 
> https://github.com/coconut-svsm/svsm/blob/346f4862d73ef40c9c9cf7c05cf2c444dff2be3a/kernel/src/sev/ghcb.rs#L767-L774

IT302: I don't think we're actually using bytes here. The comments and naming in services.rs suggest bytes, but if you trace the code, it appears to return pages.

Looking at the GHCB implementation (ghcb.rs), it returns pages which then propagate up through services.rs . I don't see anywhere in this path where pages are converted to bytes.

stefano-garzarella: @IT302 I agree on that, we are using pages, and that is the problem IMO.
In the comment here we talk about bytes (maybe):

> Per spec, return required size in rdx and raise an error

And in the spec it's also highlighted that should be bytes:

> If a certificate data buffer was provided and if the size of the certificate data from the hypervisor
> exceeds the size of the supplied certificate data buffer, RCX will be set to the size of the services
> manifest and RDX will be set to the size of the certificate data (in bytes) and the call will return
> SVSM_ERR_INVALID_PARAMETER.

IT302: Oh, I see what you mean now. Let me fix my code to conform with the SVSM spec rather than the GHCB spec. For some reason I had it in my head that all the specs said pages.

stefano-garzarella: @IT302 thanks and sorry to not be clear from the beginning.

IT302: @stefano-garzarella Done!

IT302: @stefano-garzarella Hold on for a few minutes, I did not push the latest commit.

stefano-garzarella: @IT302 I can't see any difference, can you double-check if the PR is updated?

IT302: @stefano-garzarella  Pushed the right one just now.

kernel/src/protocols/attest.rs
```diff
@@ -307,12 +410,99 @@ fn write_report_and_manifest(
...
+        let (resp, certs) = match get_attestation_report_extended(hash) {
+            Ok((resp, certs)) => (resp, certs),
+            Err(SvsmReqError::FatalError(SvsmError::Ghcb(GhcbError::VmgexitError(
+                certs_buffer_size,
+                _psp_rc,
+            )))) => {
+                // The PSP returned an error code indicating that the buffer size is too small.
+                // The required size is in certs_buffer_size but specified as number of 4 KB pages.
+                // Per SVSM spec, return required size (in bytes) in rdx and raise an error
+                params.rdx = certs_buffer_size * 4096;
```
stefano-garzarella: What about using `PAGE_SHIFT` like in https://github.com/coconut-svsm/svsm/blob/346f4862d73ef40c9c9cf7c05cf2c444dff2be3a/kernel/src/greq/driver.rs#L132-L135

IT302: @stefano-garzarella Changed to explicit shift instead of relying on compiler.

Issue #797, opened 2025-09-01
igvm: update to 0.4.0


luigix25:
igvm and igvm_defs 0.4.0 have been released. IIUC this means we can drop the dependency `zerocopy07` (https://github.com/coconut-svsm/svsm/blob/main/igvmmeasure/Cargo.toml#L15-L16).

PR #799, opened 2025-09-01
Bump igvm and igvm_defs version to 0.4.0

luigix25:
`igvm` and `igvm_defs` crates have been updated to version 0.4.0. This allows us to drop the dependency on "zerocopy07" as `igvm_defs` is using the same version of zerocopy as svsm.

Fixes: #797

### DIFF
--- a/igvmbuilder/src/vmsa.rs
+++ b/igvmbuilder/src/vmsa.rs
@@ -8,7 +8,7 @@ use igvm::registers::{SegmentRegister, X86Register};
 use igvm::snp_defs::{SevFeatures, SevSelector, SevVmsa};
 use igvm::IgvmDirectiveHeader;
 use igvm_defs::IgvmNativeVpContextX64;
-use zerocopy07::FromZeroes;
+use zerocopy::FromZeros;
 
 use crate::cmd_options::{Hypervisor, SevExtraFeatures};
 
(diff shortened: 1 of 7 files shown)


igvmbuilder/src/vmsa.rs
```diff
@@ -66,7 +66,7 @@ pub fn construct_native_start_context(
     regs: &[X86Register],
     compatibility_mask: u32,
 ) -> IgvmDirectiveHeader {
-    let mut context_box = IgvmNativeVpContextX64::new_box_zeroed();
+    let mut context_box = IgvmNativeVpContextX64::new_box_zeroed().unwrap();
```
stefano-garzarella: nit: I'd like starting to use `.expect()` where we can't return errors. (e.g.  `.expect("Failed to allocate memory for context"), but we already used `unwrap()` in several places, so it's not a strong opinion for now

stefano-garzarella (dismissed):
LGTM, just a minor comment

igvmbuilder/src/vmsa.rs
```diff
@@ -186,7 +187,7 @@ pub fn construct_vmsa(
     extra_features: &Vec<SevExtraFeatures>,
     hypervisor: Hypervisor,
 ) -> IgvmDirectiveHeader {
-    let mut vmsa_box = SevVmsa::new_box_zeroed();
+    let mut vmsa_box = SevVmsa::new_box_zeroed().expect("Failed to allocate memory for CPU state");
```
stefano-garzarella:
```suggestion
    let mut vmsa_box = SevVmsa::new_box_zeroed().expect("Failed to allocate memory for VMSA");
```

stefano-garzarella (approved):

PR #800, opened 2025-09-03
Update IGVM build and usage in INSTALL.md

RARelph:
The new IGVM default system-wide install target now requires su access to the rust toolchain, which is uncommon. Update our installation instructions to build and use IGVM in a local directory.

### DIFF
--- a/Documentation/docs/installation/INSTALL.md
+++ b/Documentation/docs/installation/INSTALL.md
@@ -92,13 +92,12 @@ require `cargo-c` to build, install it with:
 cargo install cargo-c
 ```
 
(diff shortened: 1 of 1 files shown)


stefano-garzarella:
@RARelph out of curiosity, why you closed #795 and opened this one?

RARelph:
Yes, I've seen it done before... But maybe it's because of a rebase in between? I'm not sure. I know I can send up force pushes if the item is tip. Anyway, I'll figure it out next time. But let's get this in, please. It's just documentation for how to build IGVM, install it, and have QEMU use it without having to have rust installed for root. IGVM recently updated their makefile to require rust at install time, sadly.

PR #801, opened 2025-09-04
igvm_params: correctly make IGVM memory map visible to guest FW

msft-jlange:
There is existing code that populates an IGVM memory map for guest firmware, but it only make that memory map visible to guest firmware if the memory map is within the guest firmware GPA range specified in the IGVM file.  When the memory map is outside the firmware range, it must explicitly be made visible to the guest.

### DIFF
--- a/kernel/src/igvm_params.rs
+++ b/kernel/src/igvm_params.rs
@@ -338,8 +338,9 @@ impl IgvmParams<'_> {
         assert!(self.should_launch_fw());
 
         let mut regions = Vec::new();
+        let fw_info = &self.igvm_param_block.firmware;
 
-        if self.igvm_param_block.firmware.in_low_memory != 0 {
+        if fw_info.in_low_memory != 0 {
             // Add the lowmem region to the firmware region list so
             // permissions can be granted to the guest VMPL for that range.
             regions.push(MemoryRegion::from_addresses(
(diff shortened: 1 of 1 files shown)


kernel/src/igvm_params.rs
```diff
@@ -348,10 +349,28 @@ impl IgvmParams<'_> {
...
+
+        // If this firmware expects an IGVM memory map but the IGVM memory
+        // map is not within the firmware GPA range, then add the IGVM memory
+        // map to the set of firmware regions.
+        if fw_info.memory_map_page_count != 0 {
+            let map_region = MemoryRegion::new(
+                PhysAddr::from(fw_info.memory_map_page as u64 * PAGE_SIZE as u64),
+                fw_info.memory_map_page_count as usize * PAGE_SIZE,
+            );
+            if !fw_region.contains_region(&map_region) {
```
peterfang: How about something like the following to make it more generic? Technically speaking the IGVM memory map region shouldn't overlap with any of the fw regions. Of course this would require `fw_region` to be pushed into `regions` prior to the `memory_map_page_count` check.
```suggestion
            if !regions.iter().any(|mr| mr.contains_region(&map_region)) {
```

msft-jlange: Not a bad idea.

kernel/src/igvm_params.rs
```diff
@@ -348,10 +349,36 @@ impl IgvmParams<'_> {
...
+        // If this firmware expects an IGVM memory map but the IGVM memory
+        // map is not within any of the firmware GPA ranges, then add the IGVM
+        // memory map to the set of firmware regions.
+        if fw_info.memory_map_page_count != 0 {
+            let map_region = MemoryRegion::new(
+                PhysAddr::from(fw_info.memory_map_page as u64 * PAGE_SIZE as u64),
+                fw_info.memory_map_page_count as usize * PAGE_SIZE,
+            );
+            // Scan all the firmware regions to determine whether any of them
+            // contain the memor map.  If not, the memory map should be added
```
peterfang: Nit: typo (memory)

peterfang (dismissed):
LGTM, waiting on testing. Thanks!

peterfang (approved):

PR #806, opened 2025-09-10
CocoonFs based persistence

nicstange:
This PR integrates CocoonFs for SVSM persistence.

I presented about CocoonFs in one of our devel meetings, see the [minutes from 2025-07-23](https://github.com/coconut-svsm/governance/blob/main/Meetings/devel-call-2025-07-23.md). To recap, it's been designed specifically with the persistence needs of TEEs in mind, the main features distinguishing it from existing solutions are
- Authentication of the FS state as a whole with a single, global Merkle tree -- other FSes implementing that are typically read-only.
- Fresh and random IVs at file granularity generated for every new write.
- Robustness against service disruptions by means of a journal.
For more details, see [the cocoonfs-cli README](https://github.com/coconut-svsm/cocoon-tpm/tree/main/cocoonfs-cli), which provides some further pointers to e.g. the spec.

One remark ahead: I believe it is well understood, but better to make it explicit anyway: as the secret key is obtained from the attestation subsystem, and that secret is not authenticated in any way yet, the overall scheme in its current state is prone to MITM attacks and **totally insecure**.

This PR implements the bare persistence functionality only, which culminates in providing `persistence_write_inode_sync()` and `persistence_read_inode_sync()` for use from the other SVSM code parts. I opted against wiring up the vtpm in this PR, because
- It's big enough by itself already.
- There are some security implications which need to be understood and documented, mainly due to the vtpm's identity not necessarily being unique anymore. We might end up implementing rollback protection at some point, but that would require talking to the outside and we don't have that yet. Until then, we might perhaps enforce `TPM_SU_CLEAR` for `TPM2_Shutdown()` for ensuring that the null hierarchy gets wiped at least.
- Lastly, @osteffenrh and @stefano-garzarella  have some patches for implementing vtpm persistence, c.f. [Oliver's kvm-forum-2025 branch](https://github.com/osteffenrh/coconut-svsm/commits/kvm-forum-2025/). It should be straight-forward to port these to the `persistence_{read,write}_inode_sync()` interface introduced here, and can help with that after coordinating with them.

As there are no persistence users yet, I added a small demo reading and incrementing a "boot counter" at startup. That can be found in the last commit, and is supposed to get reverted one way or the other at some point.

## Testing
- Setup attestation -- it's currently the only way to get a secret into the SVSM. Refer to `Documentation/docs/developer/ATTESTATION.md` in the sources.
- Enable the `cocoonfs` Cargo feature. This will pull in the `virtio-drivers` feature.
- Prepare a CocoonFs filesystem image using `cocoonfs-cli` (find it at the link provided above, or simply at crates.io):
  ```cocoonfs -i cocoonfs.img -f write-mkfs-info-header -H sha2 -C aes -t 128 -I 'ddeeff' -s 8M```
  Note that this merely writes a CocoonFs "filesystem creation info header", which doesn't require access to the key. The actual filesystem formatting will get done when the SVSM first attempts to open the filesystem.
- Attach the filesystem to the SVSM via the qemu command line (taken from `scripts/launch_guest.sh` in the sources):
  - append `,x-svsm-virtio-mmio=on` to the `-machine` spec and add
  - `-global virtio-mmio.force-legacy=false`
  - `-drive file=<path>/cocoonfs.img,format=raw,if=none,id=svsm_storage,cache=none`
  - `-device virtio-blk-device,drive=svsm_storage`

If everything goes well, you should see the following message at the first boot:
```
[SVSM] persistent CocoonFs storage opened successfully
[SVSM] persistence demo: no boot counter found yet
[SVSM] persistence demo: successfully wrote updated boot counter
```
and e.g.
```
[SVSM] persistent CocoonFs storage opened successfully
[SVSM] persistence demo: boot counter read back is 1
[SVSM] persistence demo: successfully wrote updated boot counter
```
etc. in subsequent ones.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,6 +80,7 @@ bitflags = "2.6"
 clap = { version = "4.4.14", default-features = false }
 cocoon-tpm-crypto = { version = "0.1.4", default-features = false }
 cocoon-tpm-tpm2-interface = { version = "0.1.1", default-features = false }
+cocoon-tpm-utils-async = { version = "0.1.3", default-features = false }
 cocoon-tpm-utils-common = { version = "0.1.3", default-features = false }
 concat-kdf = "0.1.0"
 gdbstub = { version = "0.7.10", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,7 @@ bitfield-struct = "0.6.2"
 bitflags = "2.6"
 clap = { version = "4.4.14", default-features = false }
 cocoon-tpm-crypto = { version = "0.1.4", default-features = false }
+cocoon-tpm-storage = { version = "0.1.4", default-features = false }
 cocoon-tpm-tpm2-interface = { version = "0.1.1", default-features = false }
 cocoon-tpm-utils-async = { version = "0.1.3", default-features = false }
 cocoon-tpm-utils-common = { version = "0.1.3", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,9 +78,9 @@ bindgen = { version = "0.71.0", default-features = false }
 bitfield-struct = "0.6.2"
 bitflags = "2.6"
 clap = { version = "4.4.14", default-features = false }
-cocoon-tpm-crypto = { version = "0.1.0", default-features = false }
-cocoon-tpm-tpm2-interface = { version = "0.1.0", default-features = false }
-cocoon-tpm-utils-common = { version = "0.1.0", default-features = false }
+cocoon-tpm-crypto = { version = "0.1.4", default-features = false }
+cocoon-tpm-tpm2-interface = { version = "0.1.1", default-features = false }
+cocoon-tpm-utils-common = { version = "0.1.3", default-features = false }
 concat-kdf = "0.1.0"
 gdbstub = { version = "0.7.10", default-features = false }
 gdbstub_arch = { version = "0.3.3" }
--- a/Documentation/docs/developer/PERSISTENCE.md
+++ b/Documentation/docs/developer/PERSISTENCE.md
@@ -1,0 +1,100 @@
+Persistence
+===========
+
+The persistence subsystem enables the SVSM to securely preserve sensitive state
+across reboots, such as, but not limited to, the vTPM NV state or UEFI
+variables.
+
+
+Overview
+--------
+When persisting sensitive state to external, non-volatile storage from a TEE or
+the SVSM in particular, special security considerations apply: the untrusted
+virtualization host not only has full access to the external storage volume's
+contents, but is even capable of spoofing and altering IO requests over time,
+potentially enabling certain attacks on traditional block-level FDE schemes.
+
+The
+[CocoonFs](https://coconut-svsm.github.io/cocoon-tpm/cocoonfs/cocoonfs-format.html)
+filesystem format has been designed specifically with a TEE's security
+requirements in mind, and is currently the only filesystem supported by the
+SVSM. Support for other formats might perhaps get added in the future if
+there's a demand.
+
+The persistence storage is protected by a symmetric key received at startup from
+a KBS as the result of a successful remote attestation, therefore the
+persistence functionality requires a working [attestation
+setup](ATTESTATION.md).
+
+There are two alternative ways to initially provision a CocoonFs filesystem
+image with the [`cocoonfs-cli`](https://crates.io/crates/cocoonfs-cli) utility:
+without or with access to the key. If created without a key, a mere "filesystem
+creation header" would get written, the actual formatting would then
+subsequently take place on the first attempt to open the image from the SVSM,
+with the key obtained from attestation procedure. If created with a key, the
+filesystem will get formatted right away at `cocoonfs-cli` invocation time.
+
+At startup, the SVSM would iterate over all block devices attached to it and use
+the first one that can be opened successfully with the key provided from
+attestation for all its persistence needs. In particular:
+
+* At most one CocoonFs volume will actively be used from a given SVSM boot instance.
+* If a block device has a CocoonFs image formatted onto it, but its
+  authentication with the key obtained from attestation failed, it would get
+  skipped.
+* If no working CocoonFs image had been found yet, and the next block device has
+  a "CocoonFs creation header" on it, the filesystem would get formatted with
+  whatever key had been obtained from attestation and henceforth be used for the
+  persistence needs, irrespective of whether there's perhaps another CocoonFs
+  block device formatted with that same key later in the iteration list or not.
+
+Current development state and caveats
+-------------------------------------
+The persistence functionality is under development, the core subsystem itself
+has been implemented, but none of the components which could potentially make
+use of it have been adapted yet to actually do so.
+
+Furthermore, the exact details on how the symmetric key material used for
+securing the externally stored data is to get derived from the secret received
+from the KBS have not been finalized yet, so do not expect at this point that
+persistence volume images accessible from the SVSM at some code version will
+continue to work with past or future releases.
+
+Currently, the secret key material received from the KBS is not authenticated in
+any way, making the persistence subsystem prone to MITM attacks. The details on
+how to resolve this have not been settled yet, for reference and further details
+c.f. [this mail thread](https://lore.kernel.org/r/877bwxyiqv.fsf@).
+
+Setup/Testing
+-------------
+1. Enable the `persistence` Cargo feature. This will pull in the virtio-drivers feature.
+2. Setup attestation -- it's currently the only way to get a secret into the SVSM.
+3. Prepare a CocoonFs filesystem image using [`cocoonfs-cli`](https://crates.io/crates/cocoonfs-cli)
+   ```
+   cocoonfs -i cocoonfs.img -f write-mkfs-info-header -H sha2 -C aes -t 128 -I 'ddeeff' -s 8M
+   ```
+   Note that this merely writes a CocoonFs "filesystem creation info header",
+   which doesn't require access to the key. The actual filesystem formatting
+   will get done when the SVSM first attempts to open the filesystem.
+4. Attach the filesystem image to the SVSM via the qemu command line (taken from
+   `scripts/launch_guest.sh`): append `,x-svsm-virtio-mmio=on` to the `-machine`
+   spec and add
+   ```
+   -global virtio-mmio.force-legacy=false
+   -drive file=<path>/cocoonfs.img,format=raw,if=none,id=svsm_storage,cache=none
+   -device virtio-blk-device,drive=svsm_storage
+   ```
+
+If everything goes well, you should see the following message at the first boot:
+```
+[SVSM] persistent CocoonFs storage opened successfully
+[SVSM] persistence demo: no boot counter found yet
+[SVSM] persistence demo: successfully wrote updated boot counter
+```
+and e.g.
+```
+[SVSM] persistent CocoonFs storage opened successfully
+[SVSM] persistence demo: boot counter read back is 1
+[SVSM] persistence demo: successfully wrote updated boot counter
+```
+etc. in subsequent ones.
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -46,6 +46,7 @@ cocoon-tpm-crypto = { workspace = true, features = [
     "ecc_nist_p521",
 ]}
 cocoon-tpm-tpm2-interface = { workspace = true }
+cocoon-tpm-utils-async = { workspace = true }
 cocoon-tpm-utils-common = { workspace = true }
 concat-kdf = { workspace = true, optional = true }
 gdbstub = { workspace = true, optional = true }
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -95,7 +95,7 @@ attest = [
 attest-serial = ["attest"]
 
 default = []
-persistence = ["dep:cocoon-tpm-storage", "block", "virtio-drivers"]
+persistence = ["dep:cocoon-tpm-storage", "block", "virtio-drivers", "attest"]
 enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 vtpm = ["dep:libtcgtpm"]
 nosmep = []
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -44,9 +44,9 @@ cocoon-tpm-crypto = { workspace = true, features = [
     "ecc_nist_p256",
     "ecc_nist_p384",
     "ecc_nist_p521",
-], optional = true }
-cocoon-tpm-tpm2-interface = { workspace = true, optional = true }
-cocoon-tpm-utils-common = { workspace = true, optional = true }
+]}
+cocoon-tpm-tpm2-interface = { workspace = true }
+cocoon-tpm-utils-common = { workspace = true }
 concat-kdf = { workspace = true, optional = true }
 gdbstub = { workspace = true, optional = true }
 gdbstub_arch = { workspace = true, optional = true }
@@ -84,9 +84,6 @@ test.workspace = true
 attest = [
     "vsock",
     "dep:aes-kw",
-    "dep:cocoon-tpm-crypto",
-    "dep:cocoon-tpm-tpm2-interface",
-    "dep:cocoon-tpm-utils-common",
     "dep:concat-kdf",
     "dep:kbs-types",
     "dep:libaproxy",
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -45,6 +45,7 @@ cocoon-tpm-crypto = { workspace = true, features = [
     "ecc_nist_p384",
     "ecc_nist_p521",
 ]}
+cocoon-tpm-storage = { workspace = true, features = ["zeroize"], optional = true }
 cocoon-tpm-tpm2-interface = { workspace = true }
 cocoon-tpm-utils-async = { workspace = true }
 cocoon-tpm-utils-common = { workspace = true }
@@ -94,6 +95,7 @@ attest = [
 attest-serial = ["attest"]
 
 default = []
+persistence = ["dep:cocoon-tpm-storage", "block", "virtio-drivers"]
 enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 vtpm = ["dep:libtcgtpm"]
 nosmep = []
--- a/kernel/src/async.rs
+++ b/kernel/src/async.rs
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+// Copyright 2025 SUSE LLC
+// Author: Nicolai Stange <nstange@suse.de>
+
+//! Implementation of Rust `async` [`Future`] related functionality.
+
+use cocoon_tpm_utils_async as utils_async;
+
+use utils_async::sync_types;
+
+use crate::locking::{LockGuard, RWLock, ReadLockGuard, SpinLock, WriteLockGuard};
+
+impl<T: Send> sync_types::Lock<T> for SpinLock<T> {
+    type Guard<'a>
+        = LockGuard<'a, T>
+    where
+        T: 'a;
+
+    fn lock(&self) -> Self::Guard<'_> {
+        SpinLock::lock(self)
+    }
+}
+
+impl<T: Send> sync_types::ConstructibleLock<T> for SpinLock<T> {
+    fn get_mut(&mut self) -> &mut T {
+        SpinLock::get_mut(self)
+    }
+}
+
+impl<T: Send + Sync> sync_types::RwLock<T> for RWLock<T> {
+    type ReadGuard<'a>
+        = ReadLockGuard<'a, T>
+    where
+        T: 'a;
+    type WriteGuard<'a>
+        = WriteLockGuard<'a, T>
+    where
+        T: 'a;
+
+    fn read(&self) -> Self::ReadGuard<'_> {
+        RWLock::lock_read(self)
+    }
+
+    fn write(&self) -> Self::WriteGuard<'_> {
+        RWLock::lock_write(self)
+    }
+
+    fn get_mut(&mut self) -> &mut T {
+        RWLock::get_mut(self)
+    }
+}
+
+/// Implementation of the `cocoon-tpm-utils-async` crate's [`SyncTypes`](sync_types::SyncTypes)
+/// trait suitable for the SVSM kernel execution environment.
+#[derive(Debug)]
+pub struct SvsmSyncTypes;
+
+impl sync_types::SyncTypes for SvsmSyncTypes {
+    type Lock<T: Send> = SpinLock<T>;
+    type RwLock<T: Send + Sync> = RWLock<T>;
+    type SyncRcPtrFactory = sync_types::GenericArcFactory;
+}
--- a/kernel/src/async.rs
+++ b/kernel/src/async.rs
@@ -6,6 +6,7 @@
 
 use cocoon_tpm_utils_async as utils_async;
 
+use core::task;
 use utils_async::sync_types;
 
 use crate::locking::{LockGuard, RWLock, ReadLockGuard, SpinLock, WriteLockGuard};
@@ -59,4 +60,24 @@ impl sync_types::SyncTypes for SvsmSyncTypes {
     type Lock<T: Send> = SpinLock<T>;
     type RwLock<T: Send + Sync> = RWLock<T>;
     type SyncRcPtrFactory = sync_types::GenericArcFactory;
+}
+
+/// Busypoll an asynchronous task to completion.
+///
+/// The task to poll on is represented as a closure, `task`. Note that this scheme enables
+/// convenient polling on standard Rust [`Future`]s, as well as on `Future`-like types taking
+/// additional arguments for their `poll()`.
+///
+/// # Arguments:
+///
+/// * `task` -> The task to poll to completion.
+pub fn task_busypoll_to_completion<O, T: FnMut(&mut task::Context<'_>) -> task::Poll<O>>(
+    mut task: T,
+) -> O {
+    let mut ctx = task::Context::from_waker(task::Waker::noop());
+    loop {
+        if let task::Poll::Ready(result) = task(&mut ctx) {
+            return result;
+        }
+    }
 }
--- a/kernel/src/attest.rs
+++ b/kernel/src/attest.rs
@@ -8,7 +8,7 @@
 extern crate alloc;
 
 use crate::{
-    crypto::SecretSlice,
+    crypto::{SecretSlice, get_svsm_rng},
     error::SvsmError,
     greq::{pld_report::*, services::get_regular_report},
     io::{Read, Write},
@@ -21,15 +21,10 @@ use aes_gcm::{AeadInPlace, Aes256Gcm, KeyInit, Nonce, aead::generic_array::Gener
 use aes_kw::{KeyInit as _, KwAes256};
 use alloc::{string::ToString, vec::Vec};
 use cocoon_tpm_crypto::{
-    CryptoError, EmptyCryptoIoSlices,
+    CryptoError,
     ecc::{EccKey, curve::Curve, ecdh::ecdh_c_1_1_cdh_compute_z},
-    rng::{self, HashDrbg, RngCore as _, X86RdSeedRng},
 };
-use cocoon_tpm_tpm2_interface::{self as tpm2_interface, TpmEccCurve, TpmiAlgHash, TpmsEccPoint};
-use cocoon_tpm_utils_common::{
-    alloc::try_alloc_zeroizing_vec,
-    io_slices::{self, IoSlicesIterCommon as _},
-};
+use cocoon_tpm_tpm2_interface::{TpmEccCurve, TpmsEccPoint};
 use kbs_types::Tee;
 use libaproxy::*;
 use serde::Serialize;
@@ -332,27 +327,8 @@ impl From<AttestationError> for SvsmError {
 /// Generate a key used to establish a secure channel between the confidential guest and
 /// attestation server.
 fn sc_key_generate(curve: &Curve) -> Result<EccKey, CryptoError> {
-    let mut rng = {
-        let mut rdseed = X86RdSeedRng::instantiate().map_err(|_| CryptoError::RngFailure)?;
-        let mut hash_drbg_entropy =
-            try_alloc_zeroizing_vec(HashDrbg::min_seed_entropy_len(TpmiAlgHash::Sha256))?;
-
-        rdseed.generate::<_, EmptyCryptoIoSlices>(
-            io_slices::SingletonIoSliceMut::new(hash_drbg_entropy.as_mut_slice())
-                .map_infallible_err(),
-            None,
-        )?;
-
-        rng::HashDrbg::instantiate(
-            tpm2_interface::TpmiAlgHash::Sha256,
-            &hash_drbg_entropy,
-            None,
-            Some(b"SVSM attestation RNG"),
-        )
-    }?;
-
+    let mut rng = get_svsm_rng()?;
     let curve_ops = curve.curve_ops()?;
-
     EccKey::generate(&curve_ops, &mut rng, None)
 }
 
--- a/kernel/src/crypto/mod.rs
+++ b/kernel/src/crypto/mod.rs
@@ -154,6 +154,9 @@ pub mod digest {
 
 pub mod rustcrypto;
 
+mod rng;
+pub use rng::{SvsmRng, get_svsm_rng};
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/kernel/src/crypto/rng.rs
+++ b/kernel/src/crypto/rng.rs
@@ -1,0 +1,158 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+// Copyright 2025 SUSE LLC
+// Author: Nicolai Stange <nstange@suse.de>
+
+//! SVSM Random Number Generator (RNG) management.
+//!
+//! Whenever users need an cryptographically secure RNG, they should invoke [`get_svsm_rng()`] to
+//! obtain an exclusively owned [`SvsmRng`] instance.
+
+extern crate alloc;
+use alloc::boxed::Box;
+
+use crate::locking::SpinLock;
+
+use core::sync::atomic;
+
+use cocoon_tpm_crypto::{
+    CryptoError, EmptyCryptoIoSlices,
+    hash::hash_alg_digest_len,
+    rng::{self, HashDrbg, RngCore, X86RdSeedRng},
+};
+use cocoon_tpm_tpm2_interface::TpmiAlgHash;
+use cocoon_tpm_utils_common::{
+    alloc::box_try_new,
+    fixed_vec::{FixedVec, FixedVecMemoryAllocationFailure},
+    io_slices::{self, IoSlicesIterCommon as _},
+    zeroize,
+};
+
+/// Hash algorithm to be used for all NIST HashDrbg instantiations.
+// C.f. NIST SP 800-90A Rev. 1: the HashDrbg's security strength is equal to the underlying hash
+// algorithm's pre-image resistance. That is, SHA256 gives a security strength of 256 bits.
+const SVSM_RNG_DRBG_HASH_ALG: TpmiAlgHash = TpmiAlgHash::Sha256;
+
+/// Number of [`SVSM_RNG_POOL`] slots.
+const SVSM_RNG_POOL_SIZE: usize = 4;
+
+/// RNG instance pool
+///
+/// [`SvsmRng::drop()`] returns instances back into some free slot, if any. [`get_svsm_rng()`] tries
+/// to take a currently unused instance from the pool before attempting to instantiate one.
+#[allow(clippy::type_complexity)]
+static SVSM_RNG_POOL: SpinLock<
+    [Option<Box<rng::ChainedRng<rng::X86RdSeedRng, rng::HashDrbg>>>; SVSM_RNG_POOL_SIZE],
+> = SpinLock::new([None, None, None, None]);
+
+/// Counter used for the RNG instance's personalization.
+static SVSM_RNG_INSTANTIATION_ID: atomic::AtomicU64 = atomic::AtomicU64::new(0);
+
+/// Opaque type implementing [`RngCore`] as suitable for the SVSM environment and build
+/// configuration.
+///
+/// Instances of `SvsmRng` are to be obtained through [`get_svsm_rng()`].
+///
+/// No assumptions must be made about `SvsmRng`, other than instances thereof are small, typically
+/// of pointer size, and that it implements [`RngCore`].
+#[allow(missing_debug_implementations)]
+pub struct SvsmRng {
+    // Is never None while self is alive, the rng needs to get stored in an Option<> so that it can
+    // get returned back to the pool when dropped.
+    rng: Option<Box<rng::ChainedRng<X86RdSeedRng, rng::HashDrbg>>>,
+}
+
+impl SvsmRng {
+    fn instantiate() -> Result<Self, CryptoError> {
+        // The FixedVec internal representation is optimized for the case of power-of-two lengths,
+        // relative to a specified compile-time "base" value. In practice, the HashDrbg seed length
+        // is equal to the digest length for all known hashes, so use that for the FixedVec base. If
+        // not a power of two (e.g. SHA-1), the optimization wouldn't be possible anyway, so the
+        // value is irrelevant then.
+        const DRBG_HASH_ALG_DIGEST_LEN_LOG2: u32 =
+            hash_alg_digest_len(SVSM_RNG_DRBG_HASH_ALG).ilog2();
+        let mut hash_drbg_entropy = zeroize::Zeroizing::new(
+            FixedVec::<u8, DRBG_HASH_ALG_DIGEST_LEN_LOG2>::new_with_default(
+                HashDrbg::min_seed_entropy_len(SVSM_RNG_DRBG_HASH_ALG),
+            )
+            .map_err(|e| match e {
+                FixedVecMemoryAllocationFailure => CryptoError::MemoryAllocationFailure,
+            })?,
+        );
+
+        let mut rdseed_rng = X86RdSeedRng::instantiate().map_err(|_| CryptoError::RngFailure)?;
+        rdseed_rng.generate::<_, EmptyCryptoIoSlices>(
+            io_slices::SingletonIoSliceMut::new(hash_drbg_entropy.as_mut_slice())
+                .map_infallible_err(),
+            None,
+        )?;
+
+        let mut personalization = [0u8; 16];
+        personalization[..8].copy_from_slice(b"SVSM-RNG");
+        personalization[8..].copy_from_slice(
+            &SVSM_RNG_INSTANTIATION_ID
+                .fetch_add(1, atomic::Ordering::Relaxed)
+                .to_ne_bytes(),
+        );
+
+        let hash_drbg_rng = rng::HashDrbg::instantiate(
+            SVSM_RNG_DRBG_HASH_ALG,
+            &hash_drbg_entropy,
+            None,
+            Some(&personalization),
+        )?;
+
+        let chained_rng = rng::ChainedRng::chain(rdseed_rng, hash_drbg_rng);
+
+        Ok(Self {
+            rng: Some(box_try_new(chained_rng).map_err(|_| CryptoError::MemoryAllocationFailure)?),
+        })
+    }
+}
+
+impl rng::RngCore for SvsmRng {
+    fn generate<
+        'a,
+        'b,
+        OI: cocoon_tpm_crypto::CryptoWalkableIoSlicesMutIter<'a>,
+        AII: cocoon_tpm_crypto::CryptoPeekableIoSlicesIter<'b>,
+    >(
+        &mut self,
+        output: OI,
+        additional_input: Option<AII>,
+    ) -> Result<(), rng::RngGenerateError> {
+        let rng = match self.rng.as_mut() {
+            Some(rng) => rng,
+            None => {
+                return Err(rng::RngGenerateError::CryptoError(CryptoError::Internal));
+            }
+        };
+        rng.generate(output, additional_input)
+    }
+}
+
+impl Drop for SvsmRng {
+    fn drop(&mut self) {
+        if let Some(rng) = self.rng.take() {
+            // Return the rng instance in some free pool slot, if any.
+            let mut pool = SVSM_RNG_POOL.lock();
+            for pool_slot in pool.iter_mut() {
+                if pool_slot.is_none() {
+                    *pool_slot = Some(rng);
+                    break;
+                }
+            }
+        }
+    }
+}
+
+/// Obtain an exclusively owned [`SvsmRng`] instance.
+pub fn get_svsm_rng() -> Result<SvsmRng, CryptoError> {
+    let mut pool = SVSM_RNG_POOL.lock();
+    for pool_slot in pool.iter_mut() {
+        if let Some(rng) = pool_slot.take() {
+            return Ok(SvsmRng { rng: Some(rng) });
+        }
+    }
+
+    SvsmRng::instantiate()
+}
--- a/kernel/src/locking/rwlock.rs
+++ b/kernel/src/locking/rwlock.rs
@@ -499,6 +499,17 @@ impl<T: Send, I: IrqLocking> RawRWLock<T, I> {
     pub fn write_noblock(&self) -> RawWriteLockGuard<'_, T, I> {
         self.try_lock_write().expect("Detected potential deadlock")
     }
+
+    /// Returns a mutable reference to the underlying data.
+    ///
+    /// Since this call borrows the `RawRWLock` mutably, no actual locking needs to take place --
+    /// the mutable borrow statically guarantees no new locks can be acquired while this reference
+    /// exists.
+    pub fn get_mut(&mut self) -> &mut T {
+        // SAFETY: the returned reference carries an exclusive borrow on self,
+        // thereby establishing exclusive access.
+        unsafe { &mut *self.data.get() }
+    }
 }
 
 /// A lock can only be acquired for read access if its inner type implements
--- a/kernel/src/locking/rwlock.rs
+++ b/kernel/src/locking/rwlock.rs
@@ -582,6 +582,12 @@ impl<T: Send + Sync, I: IrqLocking> RawRWLock<T, I> {
     }
 }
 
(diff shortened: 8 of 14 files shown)


Cargo.toml
```diff
@@ -64,9 +64,9 @@ base64 = { version = "0.22.1", default-features = false }
 bitfield-struct = "0.6.2"
 bitflags = "2.6"
 clap = { version = "4.4.14", default-features = false }
-cocoon-tpm-crypto = { version = "0.1.0", default-features = false }
```
osteffenrh, line 81: nit: s/cargo.toml/Cargo.toml/ in message.

kernel/src/crypto/mod.rs
```diff
@@ -101,3 +101,6 @@ pub mod digest {
 // Crypto implementations supported. Only one of them must be compiled-in.
```
osteffenrh, line 153: Commit message, check sentence: "Implement a minimal for RNG pool for the SVSM."

kernel/src/attest.rs
```diff
@@ -243,27 +239,8 @@ impl From<AttestationError> for SvsmError {
...
-            tpm2_interface::TpmiAlgHash::Sha256,
-            &hash_drbg_entropy,
-            None,
-            Some(b"SVSM attestation RNG"),
-        )
-    }?;
-
+    let mut rng = get_svsm_rng()?;
     let curve_ops = curve.curve_ops()?;
-
```
osteffenrh, line 355: nit: unrelated change?

kernel/src/locking/rwlock.rs
```diff
@@ -305,6 +306,12 @@ impl<T: Send + Sync, I: IrqLocking> RawRWLock<T, I> {
     }
 }
 
+impl<T: Send, I: IrqLocking> From<T> for RawRWLock<T, I> {
+    fn from(value: T) -> Self {
+        RawRWLock::new(value)
```
osteffenrh: `Self::new()` ?

osteffenrh (changes requested):
Here are some initial comments. I'll do some more work later.

kernel/src/attest.rs
```diff
@@ -66,7 +67,7 @@ impl TryFrom<Tee> for AttestationDriver<'_> {
 
 impl AttestationDriver<'_> {
     /// Attest SVSM's launch state by communicating with the attestation proxy.
-    pub fn attest(&mut self) -> Result<Vec<u8>, SvsmError> {
+    pub fn attest(&mut self) -> Result<Zeroizing<Vec<u8>>, SvsmError> {
```
osteffenrh: Good idea to zero out the secrets after use.

We should keep in mind that `Zeroizing<Vec<>>` does not cover reallocations done inside the Vec. To make this more water-tight we could use something with a fixed size (slice?) or come up with our own data type for handling secrets.

nicstange: FWIW, cocoon-tpm-utils has a `FixedVec` (primarily motivated by reducing the memory overhead of the `struct Vec` itself, which is 3 usizes). That would work AFAICS, as it never reallocates by design. Shall I do the conversion in a separate (later) PR, or now? Note that the `Vec` use above shouldn't be problematic, as it's not getting resized.

osteffenrh: We can do that later since it is not a problem at the moment (no resizes), IMO.

tylerfanelli:
> > > > A higher level comment: Maybe we can agree on how to encode secrets and binary data across the project. Here hex is used, while the kbs uses base64 (also @tylerfanelli)
> > > 
> > > 
> > > The attestation subsystem uses base64 because that's what is returned by a standard KBS. I could modify the KBS `aproxy` module to convert everything to hex, however. Whichever is preferred.
> > 
> > 
> > I believe @osteffenrh's comment was exclusively about CLIs and user experience. That is, `kbs-test`'s `--secret`, which expects base64 encoding vs. `cocoonfs-cli`'s `-K` which takes the very same value in hex. For anything where humans are not involved, base64 is clearly more compact, so I would personally stick with that. For CLIs OTOH, my gut feeling is that specifying binary data as hex is more common in general, but I might be wrong about that. If nothing else, hex allows for figuring the key length more or less right from eyeing at the string. So I'm leaning towards hex, but would absolutely be open for using base64 in case that's more established in the CoCo ecosystem.
> 
> Yes, I meant the human-facing representations.

Oh, I see. Yes, I agree that hex encoding would be preferred.

Cargo.toml
nicstange: Fixed, thanks!

kernel/src/crypto/mod.rs
nicstange: Fixed, thanks!

kernel/src/locking/rwlock.rs
nicstange: Fixed, thanks!

kernel/src/attest.rs
nicstange: Not really IMO: before this change the rng instantiation code in `sc_key_generate()` had been non-trivial and it made sense to separate it visually from the rest. Now both the `rng` + `curve_ops` input parameter instantiations for the subsequent `EccKey::generate()` (missing from the diff context unfortunately) are mere one-liners and it would IMO look a bit odd to separate them by a blank line from `EccKey::generate()`. Do you agree or shall I still restore the blank line?

osteffenrh: Agreed. (can make as resolved.)

kernel/src/attest.rs
osteffenrh: Created issue #807 for that. (You can resolve this comment)

nicstange: Thanks!

kernel/src/attest.rs
nicstange: Will do, thanks!

kernel/src/persistence.rs
```diff
@@ -694,3 +697,216 @@ pub fn persistence_init(mut key: Zeroizing<Vec<u8>>) -> Result<(), SvsmError> {
...
+/// # Arguments:
+///
+/// * `inode` - Number of the inode to update.
+/// * `data` - The data to write to `inode`.
+/// * `issue_sync` - Whether or not to issue a sync request to the underlying storage after the
+///   write has completed. That's best effort though and relies on the host to behave well.
+#[allow(unused)]
+pub fn persistence_write_inode_sync(
+    inode: u32,
+    data: Zeroizing<Vec<u8>>,
```
osteffenrh: Why do we need the data to be written to be in a Zeroizing buffer?

kernel/src/persistence.rs
```diff
@@ -694,3 +697,216 @@ pub fn persistence_init(mut key: Zeroizing<Vec<u8>>) -> Result<(), SvsmError> {
...
+///
+/// In case the `inode` does not exist, `None` is returned, otherwise all of the inode's data
+/// is returned wrapped in `Some`. Note that a non-existing inode and an existing one with
+/// empty data are considered different.
+///
+/// # Arguments:
+///
+/// * `inode` - Number of the inode whose data to read.
+#[allow(unused)]
+pub fn persistence_read_inode_sync(inode: u32) -> Result<Option<Zeroizing<Vec<u8>>>, SvsmError> {
```
osteffenrh: Here we have the same situation as in the attestation code.
User might do something with the buffer that causes a reallocation with is not covered by `Zeroizing<>`. IMO using `Zeroizing<Vec<>>` suggests a false sense of security .

kernel/src/persistence.rs
nicstange: The line of reasoning was that data written is sensitive in general, and making the extern facing API take a `Zeroizing` kind of documents that if the CocoonFs somehow happened to drop the buffer in the course of the operation (which is not possible with the busy-polling executor), then it would still get cleared.

From a technical POV, it doesn't matter much in the end: `Zeroizing` is `DerefMut` and with that, the wrapped value can get `mem::take()`en out again. That is, I could make `persistence_write_inode_sync()` to wrap the data itself on entry and unwrap on return. I personally like it slightly better the way it's currently is, but I don't really have a strong opinion on it...

kernel/src/persistence.rs
nicstange: Yes, absolutely, the `zeroize` crate even has some docs about the caveats: https://docs.rs/zeroize/latest/zeroize/index.html#stackheap-zeroing-notes

We could perhaps implement a `FoolProofZeroizing`, implementing `Deref` and `DerefMut` only to the `Vec`'s slice, but not to the `Vec` itself, if desired.

kernel/src/persistence.rs
```diff
@@ -0,0 +1,564 @@
...
+        let io_block_size_log2 = driver.block_size_log2() as u32;
+        let io_block_size_128b_log2 = io_block_size_log2.saturating_sub(7);
+        Self {
+            driver,
+            io_block_size_128b_log2,
+        }
+    }
+}
+
+impl NvChip for SvsmBlockDriverNvChip {
```
stefano-garzarella: Is `NvChip` specific for TPM NvData, or it's a generic abstraction for any data?

kernel/src/persistence.rs
```diff
@@ -562,3 +573,124 @@ impl NvChipFuture<SvsmBlockDriverNvChip> for SvsmBlockDriverNvChipTrimFuture {
...
+///
+/// The first successfully opened CocoonFs instance, if any, will henceforth be used to serve all
+/// the SVSM's persistence related needs.
+///
+/// # Arguments:
+///
+/// * `key` - The CocoonFs root key used (indirectly) for authentication and encryption.
+pub fn persistence_init(mut key: Zeroizing<Vec<u8>>) -> Result<(), SvsmError> {
+    // Iterate through all Virtio block devices and see if there's a CocoonFs on it, either already
+    // formatted or one with with a mkfsinfo header, which will get formatted transparently at first
```
stefano-garzarella: This behaviour IMO should be configured somewhere (compile-time or run-time).

IIUC the KBS architecture (cc @tylerfanelli), everything works if the FS is already present and initialized.

kernel/src/persistence.rs
```diff
@@ -562,3 +573,124 @@ impl NvChipFuture<SvsmBlockDriverNvChip> for SvsmBlockDriverNvChipTrimFuture {
...
+///
+/// Iterate over available block devices and attempt to open a CocoonFs instance with the provided
+/// `key` on each, until one that works has been found. Any CocoonFs instances that appear to have a
+/// valid filesystem, header but that could not get opened successfully, e.g. because the
+/// authentication with `key` failed, are getting reported in the log and being skipped over. In
+/// case a CocoonFs "filesystem creation info header" is being encountered in the search, the
+/// filesystem is formatted for used with `key` in the course.
+///
+/// The first successfully opened CocoonFs instance, if any, will henceforth be used to serve all
+/// the SVSM's persistence related needs.
```
stefano-garzarella: Also this policy should be configurable IMO. E.g. the user can specify that all the device must be encrypted, etc.

kernel/src/persistence.rs
```diff
@@ -562,3 +573,124 @@ impl NvChipFuture<SvsmBlockDriverNvChip> for SvsmBlockDriverNvChipTrimFuture {
...
+                return Err(nvfs_error_to_svsm_error(e));
+            }
+        };
+    }
+
+    // No CocoonFs block device available.
+    log::info!("no persistent CocoonFs storage found");
+    SVSM_COCOONFS_INSTANCE
+        .init(None)
+        .expect("SVSM CocoonFs instance already initialized");
```
stefano-garzarella: Should we document this behavior in the function documentation?

Why not just returning an error and let the caller decide to panic or not?

kernel/src/persistence.rs
```diff
@@ -0,0 +1,964 @@
...
+    box_try_new(read_inode_fut).map_err(NvFsError::from)
+}
+
+/// Synchronously write data to an inode on persistent storage.
+///
+/// If the `inode` does not exist yet, it will get created. All of the inode's contents will get
+/// replaced with `data`.
+///
+/// `persistence_write_inode_sync()` assumes ownership of the `data` buffer for the duration of the
+/// operation. It gets returned back unmodified to the caller on success.
```
stefano-garzarella, line 895: IMO should we document also what happen in case of error, can the caller just re-do the same call, or it needs to abort the entire operation?

kernel/src/persistence.rs
```diff
@@ -5,20 +5,31 @@
...
-use alloc::boxed::Box;
+use alloc::{boxed::Box, vec::Vec};
 
-use core::{fmt::Debug, pin, task};
+use core::{fmt::Debug, future::Future, pin, task};
 
 use cocoon_tpm_storage::{
     chip::{NvChip, NvChipFuture, NvChipIoError, NvChipReadRequest, NvChipWriteRequest},
+    fs::{
+        cocoonfs::{CocoonFs, CocoonFsFormatError, CocoonFsOpenFsFuture},
```
stefano-garzarella: Can we rethink about naming a bit, for example here can we remove all the `CocoonFs` prefix?

kernel/src/persistence.rs
tylerfanelli: I don't understand the question. What exactly does KBS have to do with this?

kernel/src/persistence.rs
nicstange: It's the cocoon-tpm-storage crate's trait defining an interface to block devices, c.f.  the [docs at docs.rs](https://docs.rs/cocoon-tpm-storage/0.1.1/cocoon_tpm_storage/chip/trait.NvChip.html)

For more context, the cocoon-tpm-storage crate defines two traits: `NvChip` and `NvFs`.  `CocoonFs` is an implementation of the latter, built on the former.

kernel/src/persistence.rs
stefano-garzarella: I mean, here IIUC we are initializing the filesystem for the state during the boot, if it's not initialized. Since with Trustee we are not authenticating the attestation server, IIUC we should avoid that, to avoid that the host force it.

kernel/src/persistence.rs
nicstange: These are only imports of the implementation from the cocoon-tpm-storage crate, i.e. not visible outside the SVSM `persistence` module?

In the future, we might perhaps want to have some more, alternative filesystem implementations, both in the SVSM itself, or in the cocoon-tpm-storage crate. E.g. for the latter, I'm thinking about implementing an unprotected dummy FS for debugging purposes at some point.

Or are you proposing to `s/cocoonfs::CocoonFsFormatError/cocoonfs::FormatError/` and `s/cocoonfs::CocoonFsOpenFsFuture/cocoonfs::OpenFsFuture/`?

kernel/src/persistence.rs
stefano-garzarella: OKay, the name is a bit confusing IMO, it seems related to TPM's NvData.

kernel/src/persistence.rs
stefano-garzarella: > These are only imports of the implementation from the cocoon-tpm-storage crate, i.e. not visible outside the SVSM `persistence` module?

Yep, I know, but I don't see any PR in `cocoon-tpm`, so I'm partially reviewing it here.

> 
> In the future, we might perhaps want to have some more, alternative filesystem implementations, both in the SVSM itself, or in the cocoon-tpm-storage crate. E.g. for the latter, I'm thinking about implementing an unprotected dummy FS for debugging purposes at some point.
> 
> Or are you proposing to `s/cocoonfs::CocoonFsFormatError/cocoonfs::FormatError/` and `s/cocoonfs::CocoonFsOpenFsFuture/cocoonfs::OpenFsFuture/`?

Yep, this ^ :-)

kernel/src/persistence.rs
nicstange: Right. I think it would be best to document that `persistence_init()` must get invoked at most once.

Adding an extra error code for this logic error condition seems a bit too much to me, as there is (and only ever will be AFAICS) a single call site, namely the SVSM init code.

kernel/src/persistence.rs
nicstange: There's one condition -- concurrent attempts to commit transactions -- where it's reasonable to retry: [`NvFsError::Retry`](https://docs.rs/cocoon-tpm-storage/0.1.1/cocoon_tpm_storage/fs/enum.NvFsError.html), everything else is considered a "hard failure" like memory allocation or IO error.

Perhaps it would be best to handle the `NvFsError::Retry` case in `persistence_write_sync()` itself by looping over and starting the write operation from scratch?

kernel/src/persistence.rs
nicstange: If you mean we should support different filesystems for the SVSM persistence needs, then yes, but this is the first one supported, so there's only one currently.

If you mean we should support more than one CocoonFs volume, then probably yes, at some point. However, I would propose to tackle that incrementally later, because we would potentially need a means to obtain different keys from the KBS.

FWIW, the CocoonFs format would support such a scheme, because all volumes have a salt in their header, which is supposed to also serve as a "volume identifier".

So the overall scheme would look like something along the lines of
1.) Iterate all block devices, find CocoonFs headers, get the salt/id out of each.
2.) Invoke the attestation machinery with the salts/ids found, obtain the respective secrets from the KBS.
3.) Open the filesystems.

And then there would be some practical questions, like: "service XY" (TPM, UEFI store) wants to persist something, which of the available volumes would it choose for that?

kernel/src/persistence.rs
nicstange: I see. Problem is grep yields 886 occurences of "NvChip" in cocoon-tpm-storage, so unless you beat me to it, the lazy me would personally prefer to leave it as is?

kernel/src/persistence.rs
nicstange: I'll write a mail to the list for kicking off a discussion on this.

kernel/src/persistence.rs
stefano-garzarella: > And then there would be some practical questions, like: "service XY" (TPM, UEFI store) wants to persist something, which of the available volumes would it choose for that?

Right, so I was just saying that we should enforce this. Like: "for now SVSM supports only 1 virtio-blk device in the configuration that must have CocoonFS on top"

kernel/src/persistence.rs
stefano-garzarella: I usually prefer (where we can) to demand the panic to the caller. BTW not a strong opinion in this case (if we document it).

kernel/src/persistence.rs
stefano-garzarella: > There's one condition -- concurrent attempts to commit transactions -- where it's reasonable to retry: [`NvFsError::Retry`](https://docs.rs/cocoon-tpm-storage/0.1.1/cocoon_tpm_storage/fs/enum.NvFsError.html), everything else is considered a "hard failure" like memory allocation or IO error.
> 
> Perhaps it would be best to handle the `NvFsError::Retry` case in `persistence_write_sync()` itself by looping over and starting the write operation from scratch?

yeah, or document it, so the caller can do that.

kernel/src/persistence.rs
stefano-garzarella: As you prefer, it was just an observation :-)

osteffenrh: I am not so happy with the name. At least for me, NvChip is more confusing than helpful.
Can't we use some refactoring tool to do the rename?

nicstange: Ok then, I'll do a `s/NvChip/NvBlockDevice/` in the cocoon-tpm-storage crate. Any suggestions on how to rename
the `SvsmBlockDriverNvChip` introduced above? FWIW, that's the thing implementing the `NvChip`/`NvBlockDevice` trait for types implementing the SVSM's `BlockDriver` trait. Simply `s/SvsmBlockDriverNvChip/SvsmBlockDriverNvBlockDevice/`?

stefano-garzarella: @nicstange I'm a bit confused by all prefixes overall. I can understand that in C where we don't have namespaces, but in Rust, IMO we can relax it a bit. e.g. for ` SvsmBlockDriverNvBlockDevice` why it can be just `BlockDevice` ?
It's not even used outside `kernel/src/persistence.rs`, so why it's `pub` ?

kernel/src/persistence.rs
nicstange: Done throughout, resolving.

kernel/src/persistence.rs
nicstange: Renamed misleading `NvChip` to `NvBlkDev` in the cocoon-tpm-storage crate, and `SvsmBlockDriverNvBlockDevice` to `SvsmNvBlkDev` here in `persistence.rs`. Stripped all unneeded `pub` specifiers.

I believe everything's been addressed here, and mark as resolved in a day or two, unless someone objects.

kernel/src/persistence.rs
nicstange: Added `Documentation/docs/developer/PERSISTENCE.md`, explaining that behavior.

kernel/src/persistence.rs
nicstange: Opted for the documentation option.

kernel/src/persistence.rs
nicstange: `persistence_write_sync()` and `persistence_read_sync()` handle `NvFsError::Retry` transparently by themselves now. Any returned error is a "real" one. Amended the rustdoc comments.

Documentation/docs/developer/PERSISTENCE.md
```diff
@@ -0,0 +1,100 @@
...
+the first one that can be opened successfully with the key provided from
+attestation for all its persistence needs. In particular:
+
+* At most one CocoonFs volume will actively be used from a given SVSM boot instance.
+* If a block device has a CocoonFs image formatted onto it, but its
+  authentication with the key obtained from attestation failed, it would get
+  skipped.
+* If no working CocoonFs image had been found yet, and the next block device has
+  a "CocoonFs creation header" on it, the filesystem would get formatted with
+  whatever key had been obtained from attestion and henceforth be used for the
```
osteffenrh: s/attestion/attestation/

nicstange:
> For moving the convenience helpers into the cocoon-tpm-storage crate, I would not be against doing that in some generalized form (I'd like to keep the async executor out of that crate, so it would come in `Future` form).

When managing the UEFI var dir in a dedicated file, one would probably want to combine updates to that and to the some UEFI var files in a single CocoonFs transaction though, for atomicity.

kraxel:
> When managing the UEFI var dir in a dedicated file, one would probably want to combine updates to that and to the some UEFI var files in a single CocoonFs transaction though, for atomicity.

Goes into the 'nice to have' bucket, but it is not a hard requirement.

* When adding a variable I can write the variable file first and update the index second.
* When removing a variable I can update the index first and remove the variable file second.
* When updating a variable I do not need to touch the index in the first place.

That should make sure things are consistent for the variable store, even though there is the risk that it leaves variable files in the filesystem which are not referenced in the index in case only one of the two updates is successfully flushed to the filesystem for add/remove ops.

nicstange:
Changes:
- CocoonFs format changes:
  - Widen inode numbers from 32 to 64 bits.
  - Auxiliary filesystem metadata support. Arbitrary UUID-value pairs. Suggested in the context of attestation for storing wrapped keys.
  - Maintain a global update counter to support future rollback protection schemes - a counter is much more convenient to handle than those authentication tree root HMACs with no ordering relation.
- Update the deps.
- Rebase work, most notably
  - don't discover virtio block devices from `persistence.rs`, use the one installed in global `BLOCK_DEVICE` now.
  - slightly rework the `SvsmNvBlkDev` implementation wrapper to store a generic `D` capable of dereffing to `BlockDriver` in place of the former owned `Box<dyn BlockDriver>`. Needed because the `BLOCK_DEVICE` owns the `Box`, while the `SvsmNvBlkDev` needs to get passed around. The generic `D` will effectively always be a `&'static dyn BlockDriver` pointing to the `BLOCK_DEVICE`'s contents.
- Persistence initialization is now a split operation:
  1. Read the metadata. Provide it to the attestation (not implemented yet), and obtain the key back.
  2. With the key, and the already read metadata, open the filesystem.

Documentation/docs/developer/PERSISTENCE.md
nicstange: Applied, thanks!

kernel/src/persistence.rs
```diff
@@ -0,0 +1,613 @@
...
+
+/// Wrapper around [`BlockDriver`] implementors, itself implementing the `cocoon-tpm-storage`
+/// crate's [`NvBlkDev`] block device abstraction.
+struct SvsmNvBlkDev<D: 'static + ops::Deref<Target: BlockDriver> + Send + Sync + Unpin> {
+    driver: D,
+    io_block_size_128b_log2: u32,
+}
+
+impl<D: ops::Deref<Target: BlockDriver> + Send + Sync + Unpin> SvsmNvBlkDev<D> {
+    fn new(driver: D) -> Self {
```
luigix25, line 46: i ran `make clippy` on every single commit and it fails in "Add cocoon-tpm-storage dep and wire it up to SVSM's BlockDriver":

```
error: associated function `new` is never used
  --> kernel/src/persistence.rs:30:8
   |
29 | impl<D: ops::Deref<Target: BlockDriver> + Send + Sync + Unpin> SvsmNvBlkDev<D> {
   | ------------------------------------------------------------------------------ associated function in this implementation
30 |     fn new(driver: D) -> Self {
   |        ^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`
```

kernel/src/persistence.rs
```diff
@@ -0,0 +1,1123 @@
...
+                match FixedVec::new_with_default(1usize << (dev_io_block_size_128b_log2 + 7)) {
+                    Ok(zeroes_buffer) => zeroes_buffer,
+                    Err(_) => {
+                        return task::Poll::Ready(Err(NvBlkDevIoError::MemoryAllocationFailure));
+                    }
+                };
+        }
+
+        for i in 0..this.io_blocks_count {
+            let dev_block_id = (this.io_block_index + i) >> dev_io_block_size_128b_log2;
```
luigix25: claude is reporting that this shift is unnecessary:

```
In SvsmNvBlkDevTrimFuture::poll (line 608), io_block_index is the NvBlkDev io-block index — confirmed by the bounds check on line 590–591, which compares it directly against dev_io_blocks (the total io-block
  count from io_blocks()):

  } else if self.io_block_index > dev_io_blocks          // io_block_index in io-block units
      || dev_io_blocks - self.io_block_index < self.io_blocks_count

  But then line 608 applies the same shift that was correct for 128-byte-unit values:

  let dev_block_id = (this.io_block_index + i) >> dev_io_block_size_128b_log2;
  //                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^  already in io-block == dev-block units
  //                                               the shift makes no sense here

  Since one io-block already equals one dev-block, the correct formula is just:

  let dev_block_id = this.io_block_index + i
```

On a quick look it kind of makes sense, but I'm not 100% sure

kernel/src/persistence.rs
```diff
@@ -0,0 +1,1123 @@
...
+        }
+    };
+
+    boot_counter += 1;
+
+    // This splats on allocation failure, but it's only a demo.
+    let data = Zeroizing::new(boot_counter.to_le_bytes().to_vec());
+    match persistence_write_inode_sync(SvsmPersistenceStaticInode::Demo as u64, data, true) {
+        Ok(_) => log::info!("persistence demo: successfully wrote updated boot counter"),
+        Err(e) => log::error!("persistence demo: boot counter updating write failed: {e:?})"),
```
luigix25: nit: there is an extra `)`

Documentation/docs/developer/PERSISTENCE.md
```diff
@@ -0,0 +1,100 @@
...
+
+There are two alternative ways to initially provision a CocoonFs filesystem
+image with the [`cocoonfs-cli`](https://crates.io/crates/cocoonfs-cli) utility:
+without or with access to the key. If created without a key, a mere "filesystem
+creation header" would get written, the actual formatting would then
+subsequently take place on the first attempt to open the image from the SVSM,
+with the key obtained from attestation procedure. If created with a key, the
+filesystem will get formatted right away at `cocoonfs-cli` invocation time.
+
+At startup, the SVSM would iterate over all block devices attach to it and use
```
luigix25: /s/attach/attached?

kernel/src/crypto/rng.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Counter used for the RNG instance's personalization.
+static SVSM_RNG_INSTANTIATION_ID: atomic::AtomicU64 = atomic::AtomicU64::new(0);
+
+/// Opaque type implementing [`RngCore`] as suitable for the SVSM environment and build
+/// configuration.
+///
+/// Instances of `SvsmRng` are to be obtained through [`get_svsm_rng()`].
+///
+/// No assumptions must be made about `SvsmRng`, other than instances thereof are small, typically
+/// of pointer size), and that it implements [`RngCore`].
```
luigix25: nit: extra ), or missing (

kernel/src/persistence.rs
```diff
@@ -0,0 +1,1123 @@
...
+    }
+}
+
+impl<D: 'static + ops::Deref<Target: BlockDriver> + Send + Sync + Unpin> Debug for SvsmNvBlkDev<D> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("SvsmNvBlkDev").finish()
+    }
+}
+
+/// Convert a [`SvsmError` to a [`NvBlkDevIoError`].
```
luigix25: missing ]

kernel/src/persistence.rs
```diff
@@ -0,0 +1,1123 @@
...
+/// block devices via `persistence_discover()`, if any. That metadata is then served as input to the
+/// attestation procedure in order to obtain a key. Eventually, the bootstrap info and the key get
+/// passed to `persistence_init()` for the unlocking.
+///
+/// # See also:
+///
+/// * [`persistence_init()`]
+pub fn persistence_discover() -> Result<Option<PersistenceBootstrapInfo>, SvsmError> {
+    // Inspect the global BLOCK_DEVICE and see if there's a CocoonFs on it, either already formatted
+    // or one with with a mkfsinfo header, which will get formatted transparently at first
```
luigix25: extra with

kernel/src/persistence.rs
```diff
@@ -0,0 +1,1123 @@
...
+/// input to the attestation, and, once the key has been obtained, to eventually get passed to
+/// [`persistence_init()`] for the unlocking.
+#[allow(missing_debug_implementations)]
+pub struct PersistenceBootstrapInfo {
+    blkdev: SvsmNvBlkDev<&'static dyn BlockDriver>,
+    fs_metadata: cocoonfs::FsMetadata,
+}
+
+impl PersistenceBootstrapInfo {
+    /// Access the matadata.
```
luigix25: /s/matadata/metadata?

kernel/src/async.rs
```diff
@@ -0,0 +1,133 @@
...
+unsafe fn nop_raw_waker_wake(data: *const ()) {
+    // Do nothing -- it's a nop waker. No ressources are associated with it.
+    debug_assert!(data.is_null());
+}
+
+/// `wake_by_ref()` entry of the [`NOP_WAKER_VTABLE`].
+// SAFETY: The RawWakerVTable signature requires the function to be marked unsafe, there's nothing
+// actually unsafe about it -- data is always a null ptr and not accessed.
+unsafe fn nop_raw_waker_wake_by_ref(_data: *const ()) {
+    // Do nothing -- it's a nop waker.
```
luigix25: why not adding an assert like in the other cases?

kernel/src/locking/spinlock.rs
```diff
@@ -7,6 +7,7 @@
 use super::common::*;
 use crate::types::TPR_LOCK;
 use core::cell::UnsafeCell;
+use core::convert::From;
```
luigix25: i think this can be dropped

kernel/src/locking/rwlock.rs
```diff
@@ -7,6 +7,7 @@
 use super::common::*;
 use crate::types::TPR_LOCK;
 use core::cell::UnsafeCell;
+use core::convert::From;
```
luigix25: ditto

kernel/src/crypto/rng.rs
```diff
@@ -0,0 +1,147 @@
...
+pub struct SvsmRng {
+    // Is never None while self is alive, the rng needs to get stored in an Option<> so that it can
+    // get returned back to the pool when dropped.
+    rng: Option<Box<rng::ChainedRng<X86RdSeedRng, rng::HashDrbg>>>,
+}
+
+impl SvsmRng {
+    fn instantiate() -> Result<Self, CryptoError> {
+        let mut hash_drbg_entropy = zeroize::Zeroizing::new(
+            FixedVec::<u8, 5>::new_with_default(HashDrbg::min_seed_entropy_len(
```
luigix25: i think 5 is log2(32), I wonder if we should use a const

kernel/src/crypto/rng.rs
```diff
@@ -0,0 +1,147 @@
...
+// algorithm's pre-image resistance. That is, SHA256 gives a security strength of 256 bits.
+const SVSM_RNG_DRBG_HASH_ALG: TpmiAlgHash = TpmiAlgHash::Sha256;
+
+/// RNG instance pool
+///
+/// [`SvsmRng::drop()`] returns instances back into some free slot, if any. [`get_svsm_rng()`] tries
+/// to take a currently unused instance from the pool before attempting to instantiate one.
+#[allow(clippy::type_complexity)]
+static SVSM_RNG_POOL: SpinLock<
+    [Option<Box<rng::ChainedRng<rng::X86RdSeedRng, rng::HashDrbg>>>; 4],
```
luigix25: adding const like `SVSM_RNG_POOL_SIZE` here?

kernel/src/svsm.rs
```diff
@@ -548,11 +550,22 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
...
-        let mut proxy = AttestationDriver::try_from(Tee::Snp).unwrap();
-        let _data = proxy.attest().unwrap();
+        // Obtain the persistence metadata first. It's not the case yet,
+        // but it likely will be needed as input to the attestation.
+        #[cfg(feature = "cocoonfs")]
+        let persistence_bootstrap_info = persistence_discover().unwrap();
 
-        // Nothing to do with data at the moment, simply print a success message.
+        let mut proxy = AttestationDriver::try_from(Tee::Snp).unwrap();
+        let secret = proxy.attest().unwrap();
```
luigix25, line 562: what about using expect with an error message?

luigix25 (commented):
Before merging I suspect the demo part should be dropped.

*Huge* thanks for this work!

kernel/src/async.rs
nicstange: For context, the assert in the other cases is `debug_assert!(data.is_null())`.

For the answer: the other cases have that, because the correctness of their implementations (to do nothing) depends on the `data` argument being a null pointer -- they are required to manage the storage backing `data` one way or the other (clone, drop, consume). C.f. https://doc.rust-lang.org/std/task/struct.RawWakerVTable.html for reference. While it is true that `data` is null in the `wake_by_ref()` case also, nothing relies on it, and that's why I omitted the `assert` from there.

kernel/src/persistence.rs
nicstange: Annotated `fn new()` with a `allow(dead_code)`  and removed that again in the subsequent commit introducing an actual usage.

kernel/src/persistence.rs
nicstange: Claude was right, fixed. Thanks a lot for running it!

kernel/src/persistence.rs
nicstange: Fixed.

Documentation/docs/developer/PERSISTENCE.md
nicstange: Applied.

kernel/src/crypto/rng.rs
nicstange: Fixed.

kernel/src/persistence.rs
nicstange: Added.

kernel/src/persistence.rs
nicstange: Fixed.

kernel/src/persistence.rs
nicstange: Applied.

kernel/src/locking/spinlock.rs
nicstange: Indeed, dropped it.

kernel/src/locking/rwlock.rs
nicstange: Dropped as well.

kernel/src/crypto/rng.rs
nicstange: Introduced local `const DRBG_HASH_ALG_DIGEST_LEN_LOG2`, with a a value derived from the `const SVSM_RNG_DRBG_HASH_ALG`.

kernel/src/crypto/rng.rs
nicstange: Done.

kernel/src/svsm.rs
nicstange: Note that it's been like that before this PR and the attestation subsystem would have emitted some warnings on what went wrong in case the `unwrap()`  here panicked. This code here is tentative and will get reworked anyway, so I'd simply leave it as-is for now.

luigix25: sure, fine for me!

nicstange:
@luigix25 : Thanks a million for your thorough review, it's much appreciated!

> Before merging I suspect the demo part should be dropped.

Yes, or reverted after -- it's all contained in a single commit. Right now, the demo is the only persistence user, so I'll leave it around until right before the merge so that people can actually test/play with it.

Changes:
- Rebased, with conflicts in Cargo.toml only, due to reformatting.
- Applied suggestions by @luigix25:
  - Fixed various typos and grammar issues in docs and log messages.
  - Fixed a bug in the `TrimFuture`.
  - Dropped unnecessary `core::convert::From` imports.
  - Introduced two constants for code clarity: `DRBG_HASH_ALG_DIGEST_LEN_LOG2` (alongside an explaining comment) and `SVSM_RNG_POOL_SIZE`.

kernel/src/svsm.rs
```diff
@@ -551,11 +553,21 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
...
```
stefano-garzarella, line 562: IIUC as next step (I guess with @armenon-rh helps since one of his PRs) the `persistence_bootstrap_info` should be passed to the `proxy.attest()`, right?

kernel/Cargo.toml
```diff
@@ -95,7 +95,7 @@ attest = [
 attest-serial = ["attest"]
 
 default = []
-cocoonfs = ["dep:cocoon-tpm-storage", "block", "virtio-drivers"]
+cocoonfs = ["dep:cocoon-tpm-storage", "block", "virtio-drivers", "attest"]
```
stefano-garzarella: I'm confused here, if `cocoonfs` depends on `attest`, why in `kernel/src/svsm.rs` we have `#[cfg(not(feature = "cocoonfs"))]` inside the attestation block? (I mean here: https://github.com/coconut-svsm/svsm/pull/806/changes/c9c5317690a98c5225cc918b163c315e354e8ec4#diff-b3e68fb174065a98fa188eecccb858b06bb0a61f3fa8dc20a5194b7e6a4f9666R565)

kernel/src/svsm.rs
```diff
@@ -567,6 +567,7 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
         #[cfg(feature = "cocoonfs")]
         if let Some(persistence_bootstrap_info) = persistence_bootstrap_info {
             persistence_init(persistence_bootstrap_info, &secret).unwrap();
+            persistence_demo();
```
stefano-garzarella, line 571: I'm fine to keep the demo here, maybe just add a TODO comment to remove it, or a special tmp feature to enable/disable it (not strong opinion on that, I'm fine to keep it till we don't use it for attestation)

kernel/src/async.rs
```diff
@@ -60,3 +61,73 @@ impl sync_types::SyncTypes for SvsmSyncTypes {
...
+    nop_raw_waker_wake,
+    nop_raw_waker_wake_by_ref,
+    nop_raw_waker_drop,
+);
+
+/// `clone()` entry of the [`NOP_WAKER_VTABLE`].
+// SAFETY: The RawWakerVTable signature requires the function to be marked unsafe, there's nothing
+// actually unsafe about it -- data is always a null ptr and not accessed.
+unsafe fn nop_raw_waker_clone(data: *const ()) -> task::RawWaker {
+    // No ressources are associated with the nop waker.
```
stefano-garzarella: nit:
```suggestion
    // No resources are associated with the nop waker.
```

stefano-garzarella: Copied&Pasted in the next 2 functions as well

kernel/src/async.rs
```diff
@@ -60,3 +61,73 @@ impl sync_types::SyncTypes for SvsmSyncTypes {
...
+/// convenient polling on standard Rust [`Future`]s, as well as on `Future`-like types taking
+/// additional arguments for their `poll()`.
+///
+/// # Arguments:
+///
+/// * `task` -> The task to poll to completion.
+pub fn task_busypoll_to_completion<O, T: FnMut(&mut task::Context<'_>) -> task::Poll<O>>(
+    mut task: T,
+) -> O {
+    let raw_waker = nop_raw_waker_new();
```
stefano-garzarella: Can we use https://doc.rust-lang.org/core/task/struct.Waker.html#method.noop ?

IIUC it's stable since 1.85 and we are at 1.86, so should be fine and allow us to remove some code here. WDYT?

kernel/src/persistence.rs
```diff
@@ -817,3 +820,257 @@ pub fn persistence_init(
...
+/// retry received from the backing filesystem implementation are handled transparently within
+/// `persistence_write_inode_sync()` itself.
+///
+/// # Arguments:
+///
+/// * `inode` - Number of the inode to update.
+/// * `data` - The data to write to `inode`.
+/// * `issue_sync` - Whether or not to issue a sync request to the underlying storage after the
+///   write has completed. That's best effort though and relies on the host to behave well.
+#[allow(unused)]
```
stefano-garzarella, line 907: Could this be removed after next commit that adds the demo?

kernel/src/persistence.rs
```diff
@@ -817,3 +820,257 @@ pub fn persistence_init(
...
+/// empty data are considered different.
+///
+/// Any error propagated back to the caller indicates an actual problem -- in particular requests to
+/// retry received from the backing filesystem implementation are handled transparently within
+/// `persistence_read_inode_sync()` itself.
+///
+/// # Arguments:
+///
+/// * `inode` - Number of the inode whose data to read.
+#[allow(unused)]
```
stefano-garzarella, line 1034: Ditto.

kernel/src/persistence.rs
```diff
@@ -1074,3 +1074,49 @@ pub fn persistence_read_inode_sync(inode: u64) -> Result<Option<Zeroizing<Vec<u8
...
+            log::info!("persistence demo: boot counter read back is {boot_counter}");
+            boot_counter
+        }
+        None => {
+            log::info!("persistence demo: no boot counter found yet");
+            0
+        }
+    };
+
+    boot_counter += 1;
```
stefano-garzarella: nit (feel free to ignore, it's a demo xD):
```suggestion
    boot_counter = boot_counter.wrapping_add(1);
```

kernel/Cargo.toml
```diff
@@ -94,6 +95,7 @@ attest = [
 attest-serial = ["attest"]
 
 default = []
+cocoonfs = ["dep:cocoon-tpm-storage", "block", "virtio-drivers"]
```
stefano-garzarella: Not a strong opinion, but would it better to change the name of the feature, I mean something like `persistence` like the module we are adding. Anyway, `cocoonfs` is also fine, just something that came to mind while reviewing.

stefano-garzarella (commented):
@nicstange thanks for the huge work! Overall LGTM. I left some minor comments (mostly for my understanding) and some nits (feel free to ignore them and close those comments).

I think we are near to merge this :-)

stefano-garzarella:
> @luigix25 : Thanks a million for your thorough review, it's much appreciated!
> 
> > Before merging I suspect the demo part should be dropped.
> 
> Yes, or reverted after -- it's all contained in a single commit. Right now, the demo is the only persistence user, so I'll leave it around until right before the merge so that people can actually test/play with it.

I'm fine also to keep it while we support it in the attestation.

kernel/src/svsm.rs
nicstange: Correct. It provides access to the auxiliary fs metadata, which, for our purposes, is supposed to contain a wrapped key.

kernel/Cargo.toml
nicstange: The `#[cfg(not(feature = "cocoonfs"))]` branch is basically the original code retained, for the likely case someone would like to work on attestation without having the whole persistence thing setup/enabled.

stefano-garzarella: oh, I see now, my mind was thinking the opposite...

kernel/src/async.rs
nicstange: Hah! Nice, thanks for the pointer, will look into it.

kernel/src/persistence.rs
nicstange: Let's keep it until it's been decided whether the demo should get included or not.

stefano-garzarella: yeah, I saw that comment later. Makes sense to me!

kernel/src/svsm.rs
stefano-garzarella: of course, feel free to ignore this if you would like to remove the demo commit from this PR.

kernel/Cargo.toml
nicstange: Way back we discussed the possibility to support other filesystems in the future and I didn't want to reserve the generic "persistence" now. But either way is fine for me, awaiting instructions ;)

stefano-garzarella: IMHO we can have "persistence" that for now uses cocoonfs since it's the only one we support. When we will add a new one, we can maybe add more features to enable/disable each one.

kernel/src/svsm.rs
stefano-garzarella: @nicstange great, thanks for confirming!

kernel/src/svsm.rs
nicstange: I added a TODO comment right before the `persistent_demo()` invocation.

kernel/src/async.rs
nicstange: The code got removed now (in favor of `task::Waker::noop()`).

kernel/src/async.rs
nicstange: And it's working great!

kernel/src/persistence.rs
nicstange: Applied :)

kernel/Cargo.toml
nicstange: Renamed the "cocoonfs" Cargo feature to "persistence".

nicstange:
Latest changes: implemented suggestion by @stefano-garzarella:
- Renamed "cocoonfs" Cargo feature to "persistence".
- Dropped custom NOP `task::Waker` implementation in favor of `Waker::noop()`.
- Added TODO comment right before `persistence_demo()`invocation.
- Make demo `boot_counter` increment wrap on overflows :P

stefano-garzarella (approved):

kernel/src/async.rs
luigix25: ok thanks!

kernel/src/persistence.rs
osteffenrh: Let's leave it as is. We can still improve the zeroing later, if needed. This is not a big problem.

Thanks for your work!

osteffenrh (approved):

stefano-garzarella:
@nicstange is it expected that on the first boot it takes 4/5 seconds to open the cocoonfs storage, I'm seeing this:

```
[SVSM] attestation successful
# 4/5 seconds without any output
[SVSM] persistent CocoonFs storage opened successfully
[SVSM] persistence demo: no boot counter found yet
[SVSM] persistence demo: successfully wrote updated boot counter
```

From the second boot I see this without any relevant time between the attestation print and the cocoonfs prints:

```
[SVSM] attestation successful
[SVSM] persistent CocoonFs storage opened successfully
[SVSM] persistence demo: boot counter read back is 1
[SVSM] persistence demo: successfully wrote updated boot counter
```

Image created with `cocoonfs -i cocoonfs.img -f write-mkfs-info-header -H sha2 -C aes -t 128 -I 'ddeeff' -s 8M`

nicstange:
> @nicstange is it expected that on the first boot it takes 4/5 seconds to open the cocoonfs storage, I'm seeing this:
[..]
> Image created with `cocoonfs -i cocoonfs.img -f write-mkfs-info-header -H sha2 -C aes -t 128 -I 'ddeeff' -s 8M`

Yes, with a debug build at least (not sure about release builds). The cocoonfs-cli command above merely placed a "filesystem creation header" on the volume, a feature intended for keyless provisioning, and the actual formatting takes place at first opening time then.

If you don't want that behavior, you can also create the filesystem directly with cocoonfs-cli, via the 'mkfs' subcommand. But then you need to provide the key to it.

stefano-garzarella:
> > @nicstange is it expected that on the first boot it takes 4/5 seconds to open the cocoonfs storage, I'm seeing this:
> > [..]
> > Image created with `cocoonfs -i cocoonfs.img -f write-mkfs-info-header -H sha2 -C aes -t 128 -I 'ddeeff' -s 8M`
> 
> Yes, with a debug build at least (not sure about release builds). The cocoonfs-cli command above merely placed a "filesystem creation header" on the volume, a feature intended for keyless provisioning, and the actual formatting takes place at first opening time then.

Oh, I see, should we log that? I mean in SVSM having an `info` log about formatting a new image. (we can do also later in a followup)

nicstange:
> Oh, I see, should we log that? I mean in SVSM having an `info` log about formatting a new image. (we can do also later in a followup)

We can, the information whether or not the filesystem had already been formatted should be available in the `PersistenceBootstrapInfo` passed to the attestation step. I would prefer to do that as a follow-up though.

stefano-garzarella:
> > Oh, I see, should we log that? I mean in SVSM having an `info` log about formatting a new image. (we can do also later in a followup)
> 
> We can, the information whether or not the filesystem had already been formatted should be available in the `PersistenceBootstrapInfo` passed to the attestation step. I would prefer to do that as a follow-up though.

sure

joergroedel (approved):

PR #808, opened 2025-09-15
SVSM Reset/Reboot guest

RARelph:
I'm new to the code base and Rust, so I expect lots of comments. But please be kind.

This change, along with coconut-svsm/linux Pull Request #15, allows an SVSM Linux guest to do "sudo reboot". It also requires the changes to EDK2 to disable MTRR usage under SNP.
It works with or without the QEMU -no-reboot option, because QEMU isn't involved in the VMPL2 guest reboot process.

The change is 2 commits. The first is a separately developed run-length encoded bit map implementation, with lots of tests, examples, and a couple of thread-safe wrappers. The second commit is adding the new SVSM_REQ_CORE_REBOOT message and it's handler. The handler invalidates all the pages the guest validated using SVSM_REQ_CORE_PVALIDATE and then resets the guest vCPU.

That is probably incomplete for some use cases, but it seems to be enough to reboot an Ubuntu 24.04 guest repeatedly.


luigix25:
I still have to look at your PR in detail, for now I have just one stupid question: is there a specific reason why you decided to import the `rlebits` crate in the codebase and not just put it as a dependency in the `Cargo.toml` file? Thanks!

RARelph:
I asked Joerg what his preference was... and this is what he recommended. Since I wrote rlebits myself, it isn't available in crates.io or any other public repository.

RARelph:
Just figured out that I cannot add reviewers because I don't have write permissions for this repo. So... mentioning @msft-jlange here hoping for his comments.

kernel/src/platform/mod.rs
```diff
@@ -106,6 +106,11 @@ pub trait SvsmPlatform: Sync {
         Ok(())
     }
 
+    /// Relaunches guest boot firmware (part of a guest reboot)
+    fn relaunch_fw(&self) -> Result<(), SvsmError> {
+        Ok(())
```
msft-jlange: Shouldn't this return an error on platforms that don't support relaunch?

RARelph: I suppose it could return SvsmError::NotSupported. The caller is the core protocol reboot command. If it returns to the VMPL 2 call site at all, that's an error. The Linux code that sent a reboot command that unexpectedly returns just tries another method to reboot.

kernel/src/protocols/core.rs
```diff
@@ -467,6 +536,7 @@ pub fn core_protocol_request(request: u32, params: &mut RequestParams) -> Result
         SVSM_REQ_CORE_WITHDRAW_MEM => core_withdraw_mem(params),
         SVSM_REQ_CORE_QUERY_PROTOCOL => core_query_protocol(params),
         SVSM_REQ_CORE_CONFIGURE_VTOM => core_configure_vtom(params),
+        SVSM_REQ_CORE_REBOOT => core_reboot(params),
```
msft-jlange: This should be a separate protocol.  The core protocol is intended for those operations that are inherently part of what it means to be an SVSM, and must be supported by every entity that can support the SVSM protocols.  Reboot is an optional service that should not be assumed to be a part of every SVSM, and must be separately enumerable.  Separately enumerable functionality is the domain of other protocols.

kernel/src/platform/mod.rs
RARelph: Done.

kernel/src/protocols/core.rs
RARelph: I know I'm the newbee here, but your premise doesn't make sense to me. Maybe the current behavior of REBOOT isn't optimal, but I note that SVSM_REQ_CORE_CONFIGURE_VTOM is optional, merely requiring that the platform report availability, not necessarily functionality.

msft-jlange: The optional nature of vTOM is a host property, not an SVSM property.  On hosts that support vTOM, the SVSM is expected to permit the guest to configure it.  The fact that COCONUT-SVSM doesn't implement this is arguably a bug, but not an important one because Linux doesn't use vTOM so nobody has raised the issue.  Plus, memory configuration is core to the guest's ability to execute at all, which is another important attribute of the core protocol (like memory validation state, vCPU creation, and fundamental SVSM interaction).

Contrast this with reboot, where support is entirely within the SVSM, and which is not a fundamental behavior of how the guest can exist and execute.

RARelph: I'm not following... your first comment said core protocol is for those operations that are inherently part of what it means to be an SVSM. Your second comment, noting that the new reboot command is entirely within SVSM, seems at odds with that. And in between we note that VTOM is a "host" property (not platform or SVSM?), but part of SVSM Core Protocol.

I don't think it makes sense to create an entirely new protocol for one command that platforms could implement and we'd like all platforms that DO implement it to implement it the same way. Further, if we ever get to the point where the SVSM can be the handler of unintentional guest resets like triple faults, they would go through this same code. That said, I see how merely requiring availability reporting for new capabilities might lead to cramming everything in the core protocol.

Anyway, let's discuss this with more participants in tomorrow's meeting.

msft-jlange: VTOM configuration is a guest concept, which can only be performed by an SVSM, and is only relevant on hosts that support the use of VTOM.  VTOM bleeds everywhere.

What it means to be an SVSM is to provide the set of services that are required for fundamental execution of a guest.  If the SVSM can't provide the foundational set of guest services, then it can't be an SVSM.

There is no enumeration protocol for functions within the core protocol.  If the core protocol is support at version _n_, then all functions defined by version _n_ need to be supported.  There is no concept of optional services within a foundational protocol.  And that should continue to be the case.

There's nothing wrong with designing a protocol that supports only a single function.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+    InvalidValue,
+    RunsOutOfSpace,
+    InternalError,
+    #[cfg(feature = "thread-safe")]
+    LockPoisoned,
+}
+
+#[derive(Debug, Clone)]
+pub struct RleBits {
+    runs: [usize; 199], // Alternating run lengths, starting with 0s at index 0
```
nicstange: Do I get it right that this puts a static limit on the total number of runs possible?

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+            return None;
+        }
+
+        let mut start = 0usize;
+        for (i, &len) in self.runs.iter().enumerate() {
+            if index < start + len {
+                return Some((i % 2) as u8);
+            }
+            start += len;
+        }
```
nicstange: Any idea whether run insertion/removal or lookup is the more performance sensitive operation? Because if it was lookup, we might be better of storing the runs' absolute end address rather than the lengths and do a binary search here. The downside would be that insertion/removal would have to apply "correction offsets" to all runs in the tail following the modified position.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        None // Should NEVER happen
+    }
+
+    pub fn get_run(&self, n: usize) -> usize {
+        if n >= self.runs.len() {
+            return 0;
+        }
+        self.runs[n]
+    }
```
nicstange: The more "Rusty" way would be to implement the `Index` trait. Perhaps for a custom, private `usize` index wrapper like `RleBitsIndex` if you don't want to expose that to the outside.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    pub fn set(&mut self, index: usize, value: u8) -> Result<(), RleBitsError> {
```
nicstange: Make `value` a `bool`, to save you some domain check further below.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+        self.set_range(index, 1, value)
+    }
+
+    fn move_runs_up(&mut self, start: usize, count: usize) -> Result<(), RleBitsError> {
+        let len = self.runs.len();
+        if start + count >= len {
+            return Err(RleBitsError::RunsOutOfSpace);
+        }
+        let copy_len = len - count;
+        self.runs.copy_within(start..copy_len, start + count);
```
nicstange: The premise here is that the tail `count` entries or more are all unoccupied, right? Because if not, we would shift some out. Perhaps I'm missing something, but I cannot find anything in the code which would prevent this from happening though.

Perhaps it would be worthwhile to add something like a `runs_used_len: usize` to `struct RleBits` for the checking. As a side-effect, that would also enable us to avoid unconditionally moving the full tail around: right now, in e.g. the initial state of having one entry, inserting another one would move 199 - 1 usizes around, if I'm not mistaken.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+    pub fn set_range(
+        &mut self,
+        mut index: usize,
+        mut len: usize,
+        value: u8,
+    ) -> Result<(), RleBitsError> {
+        if (index.checked_add(len).is_none()) || (index + len > self.length) {
+            return Err(RleBitsError::RangeExceedsLength);
+        }
+        if value != 0 && value != 1 {
```
nicstange: If `value` was a bool, this check wouldn't be necessary.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+                            self.runs[i] -= len;
+                            return Ok(());
+                        }
+                        // index + len >= next...
+                        // We can squeeze out the current wrong-value run
+                        // since it is completely obliterated by the range
+                        let merged_bits = self.runs[i] + self.runs[i + 1];
+                        start -= self.runs[i - 1];
+                        self.runs[i - 1] += merged_bits;
+                        self.move_runs_down(i, 2);
```
nicstange: Moving in fixed steps makes the whole operation quadratic in the number of pre-existing overlapping runs. Not sure that's an actual concern for the use-case of tracking page validations, but perhaps it would be better to search for the end of overlapping runs once upfront and do a single `move_runs_down()` as appropriate.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+                        i -= 1; // repeat analysis of newly expanded prior run
+                        continue;
+                    }
+
+                    // Special case for index == start (== 0) && i == 0 && value != 0
+                    if i != 0 || value == 0 || start != 0 || index != 0 {
```
nicstange: This is trivially false as per the control logic: if here, then `i == 0`. It follows that `start == 0`. Per being here `index == start (== 0)`. Also, per being here, `i & 0x1 != value`, therefore, as `i == 0`, `value != 0`.  IMO, as a matter of personal taste, it would be better to have an `debug_assert!()` instead with an explaining comment in place of the error return.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+                flag = true;
+                runs = i;
+            }
+        }
+        assert!(length == self.length, "Run lengths don't add up!");
+        runs
+    }
+
+    #[cfg(any(feature = "std", test))]
+    pub fn dump_with<F>(&self, format_addr: F)
```
nicstange, line 276: Would it make sense to `impl Debug` instead?

rlebits/src/sync.rs
```diff
@@ -0,0 +1,142 @@
...
+/// **Performance characteristics:**
+/// - Read scalability: Poor (all reads serialize)
+/// - Write scalability: Poor (all writes serialize)  
+/// - Complexity: Simple
+/// - Best for: Low contention scenarios
+#[cfg(feature = "thread-safe")]
+#[derive(Debug)]
+pub struct ThreadSafeRleBits {
+    inner: Mutex<RleBits>,
+}
```
nicstange, line 21: A question: does this offer anything other than convenience over having users wrap the non-sync `RleBits` in the locking implementation of their choice?

rlebits/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+// 4. EventuallyConsistentRleBits - Thread-local with periodic sync
+let eventually_bits = EventuallyConsistentRleBits::new(1000);
+```
+
+**Performance characteristics** (from benchmarks):
+- `ShardedRleBits`: 6.3x faster than mutex (best for spatially separated access)
+- `RwLockRleBits`: 1.1x faster than mutex (best for read-heavy workloads)  
+- `ThreadSafeRleBits`: Baseline performance (best for simple use cases)
+- `EventuallyConsistentRleBits`: Fastest for thread-local access (requires std)
```
nicstange: I can't find `ThreadSafeRleBits` or `EventuallyConsistentRleBits` in the code, is this perhaps a leftover?

kernel/src/protocols/core.rs
```diff
@@ -457,6 +478,54 @@ fn core_remap_ca(params: &RequestParams) -> Result<(), SvsmReqError> {
...
+    // Clear GUEST_VALID page list
+    GUEST_VALID.reset();
+
+    flush_tlb_global_sync();
+
+    Ok(())
+}
+
+fn core_reboot(_params: &RequestParams) -> Result<(), SvsmReqError> {
+    invalidate_guest_pages()?;
```
nicstange: I'm wondering, is it possible to race with `core_pvalidate_one()` here?

kernel/src/protocols/core.rs
```diff
@@ -457,6 +478,54 @@ fn core_remap_ca(params: &RequestParams) -> Result<(), SvsmReqError> {
...
+    crate::platform::SVSM_PLATFORM.relaunch_fw()?;
```
nicstange: Certainly not in the scope of this PR, but mentioning it anyway so it won't get lost: relaunching the FW requires a reset of the TPM.

nicstange (commented):
Hi Richard,

first of all, thanks for all the nice work!

I went primarily through the `RleBits` and had a couple of questions and some suggestions, find them all as comments inline.

My primary concern is the static upper limit on the total number of runs. I'm not sure myself what could reasonably be done about it though. Perhaps we can just implement a scheme to double the capacity whenever we run out of free slots (and never shrink again).

Thanks!

Nicolai

rlebits/src/lib.rs
RARelph: Yes. When I started this, I was under the (perhaps mistaken) impression that dynamic memory allocation was to be avoided, if it was available at all. FWIW, booting Ubuntu Linux takes ~60 runs.

rlebits/src/lib.rs
RARelph: Lookup, at the moment, isn't done at all. There was a hope that we could lookup whether a page was already in the correct state and avoid doing the PVALIDATE at all, but that fails because of "implicit conversion" done by the guest without knowledge of the SVSM. So the present use is a ton of insertions, with very few removals, and no lookups. At reboot time, we use get_run(), invalidate all the validated pages, then reset the bitmap.

rlebits/src/lib.rs
RARelph: This function was an afterthought, really, and wasn't needed until I gave up on trying to maintain the bit map across a reboot. I'll look in to the Index trait.

rlebits/src/lib.rs
RARelph: There was an issue with using boolean... I don't recall what it is at the moment, but I'll look in to it again.

rlebits/src/lib.rs
RARelph: Not instead, certainly. What I was running in to was wanting to dump page addresses, rather than page numbers. So I wanted a dump function that printed out the values << 12. And in hex. But the test cases all 'think' in terms of simple indices. I didn't want 2 functions, so... this is what I came up with.

rlebits/src/sync.rs
RARelph: Nope. Well, other than a Rust-newbie like myself having to figure out what all the appropriate choices are for a no_std environment that didn't require re-working all the code that was already using the non-sync RleBits. Just use ThreadSafeRleBits to define the bit map instance and I was done.

rlebits/README.md
RARelph: ThreadSafeRleBits is used. EventuallConsistentRleBits is not. That's a leftover. I'll clean that up.

kernel/src/protocols/core.rs
RARelph: Maybe... At least in Linux, it seems all the APs are shut down before the BSP attempts to pull the reset line. But this may not be the case for other guests, or if one vCPU ends up doing something asynchronous, like a triple fault. Perhaps I should wrap it in with_lock<>. I'll look in to that.

kernel/src/protocols/core.rs
RARelph: Thanks. Honestly, I don't know how to tell whether secure boot is 'working' even on the initial boot.

rlebits/src/lib.rs
RARelph: Correct on your premise. It is an invariant throughout the code that the array is maintained with 0s only at the end (and at index 0, if the bitmap has a 1/true in the first bit.) I'm seeing code that will detect if we try to split a run near the end of the array, but I'm also not seeing at the moment how I detect a failure of shifting non-zero runs off the end.
And, yes, adding a 'runs used' value to the structure would allow move_runs_up() to be optimized a bit. I'll study this further.

RARelph: So thinking about the detection of throwing non-0 runs off the end, I think that could be fairly easily done by checking the last 'count' values in the runs array are 0 before doing the copy_within().

Note that move_runs_up() is always called with a count of 2, so this seems pretty cheap. But first I'm going to create a test case for this and study the optimization opportunity further. The copy_within() call is expensive, for sure. But so may be maintaining the runs used value. Maybe. (Or maybe not... Hmmm...)

rlebits/src/lib.rs
nicstange: I'm personally not aware of any such "avoid allocs" policy in the context of the SVSM.

It's not immediately obvious, but as long as the environment defines a [global allocator](https://doc.rust-lang.org/beta/std/alloc/index.html#the-global_allocator-attribute), which the SVSM does, then one can use core Rust containers like `Vec` etc, even with `no_std`:

```
extern crate alloc;
use alloc::vec::Vec;
```
for the import in your module (the `std::vec` is only a re-export of the `alloc:vec` AFAIK).

One thing to be a bit cautious about is that `Vec::resize()` panics on reallocation failure, which is something to avoid in the SVSM. Prevent `Vec::resize()` from doing reallocations itself in the first place by issuing a `Vec::try_reserve()` (or `Vec::try_reserve_extact()`) prior to it.

rlebits/src/lib.rs
nicstange: Ok, thanks for the insights. I think we (you) can mark this as resolved then.

rlebits/src/lib.rs
nicstange: FWIW, it's possible to convert back and forth between `u8` and `bool` in Rust as follows:

- `bool` -> `u8`: `flag as u8`
- `u8` -> bool: `value != 0`

rlebits/src/lib.rs
nicstange: AFAICS, the cost of `move_runs_{up,down}()` or rather their `copy_within()` doesn't depend so much on the `count`, but rather on the position within the `runs` array. What I mean is that e.g. `move_runs_up()` essentially translates to a
```
let mut i = 199 - count;
while i > start {
   i -= 1;
   self.runs[i + count] = self.runs[i];
}
```
That is, it boils down to `199 - (start + count)` memory loads and stores (vectorization and such set aside).

with a `runs_used`, one could initialize the `i` with that, and the number of word moves would be `runs - start` in total, which would be much better if `runs_used` was significantly less than the capacity of `199` on average. For maintaining `runs_used`: wouldn't that just come to increments and decrements respectively by `count` in `move_runs_{up,down}()`?

rlebits/src/lib.rs
nicstange: Ok, feel free to mark as "resolved".

rlebits/src/sync.rs
nicstange: I would suggest dropping the `ThreadSafeRleBits` then if somehow possible -- for dependency hygiene if nothing else: the dependency on the external `spin` crate wouldn't be needed anymore then.

Instead, you might want to consider using e.g. the SVSM `SpinLock` defined in `kernel/src/locking/spinlock.rs` directly in the actual reboot code, i.e. for `GUEST_VALID`.

kernel/src/protocols/core.rs
nicstange: The guest isn't trusted though. Perhaps I'm too paranoid, but maybe the guest and the hypervisor can cooperate on a plot to confuse the SVSM.

rlebits/src/lib.rs
RARelph: Sorry, I wasn't very clear. I meant that checking the last 'count' (which is always 2) values are 0 is pretty cheap. Yes, the cost of the memmove is dependent on the length.

I wrote a test that did trip the bug you discovered. I then got energetic and added a max_run value to the structure. That turned out to be more complicated than it seemed at first, dealing with all the end cases that just naturally fell out before. But it's done now and I don't think the overhead of maintaining the value is too onerous. It makes the code harder to follow, which is always a trade-off I'm sensitive to. But I'll let you be the judge.

rlebits/src/lib.rs
RARelph: Yup, understood. I did write it that way at first... now that I've got the max_run variable, I'll go look at that again.

rlebits/src/lib.rs
RARelph: Ya, given that the present use case is always setting or clearing a single bit, I'll leave that for the far-off future.

rlebits/src/lib.rs
RARelph: I thought of ripping it out altogether. I put it in early during development, and later looked at it and reached the same conclusion that it was provably false. I'm curious if the compiler is smart enough to have chucked it. Can I use debug_assert() in no_std code?

rlebits/README.md
RARelph: Thanks. I took out the reference to the shards implementation as well.

kernel/src/protocols/core.rs
RARelph: Fair enough. I'll take care of this.

rlebits/src/lib.rs
nicstange: Yes, `debug_assert()` is available with `no_std` (and it expands to nothing for release builds). I personally find such `debug_assert()`s helpful for understanding the code with complex control flows like it's the case here, so I would retain the assertion in some form or the other, but that's up to you of course.

kernel/src/protocols/core.rs
RARelph: @msft-jlange - I haven't forgotten about this. I'm working with @tlendacky to get the spec updated and putting reboot in a new protocol.

rlebits/src/lib.rs
RARelph: Still pending... Definitely something I want to do, but it'll probably take me most of a day.

rlebits/src/lib.rs
RARelph: Will respond tomorrow.

rlebits/src/lib.rs
RARelph: Done.

rlebits/src/sync.rs
RARelph: I'll add this to tomorrow's to do.

rlebits/README.md
RARelph: Done.

kernel/src/protocols/core.rs
RARelph: I think this is done.

RARelph:
Thanks, @nicstange, @msft-jlange. I very much appreciate the comments. I'm obviously not done, since the format checker isn't happy, and I've got several suggestions still to implement. But wanted to push up some of the simpler fixes and see if some of the comments can be closed.

rlebits/src/lib.rs
nicstange: Makes sense.

rlebits/src/lib.rs
nicstange: I though a bit more about this. As you explained elsewhere, the current usage scenario is to add/remove single page entries and one would hope that guests typically pvalidated their individual regions incrementally in linear order each, i.e. with no random jumping around in the common case. With this, the most prevalent scenario by far would be shoveling single bits at the boundary of two `RleBits` runs from one run to its neighbor, with no run insertions/removals involved. In this case, the dominating factor from a performance POV would be the linear search over the runs to find the ones to modify. If we were to track each run's associated absolute end address (exclusive; "address" == page index here) instead of the runs' lengths each, then we could do a binary search and I'm anticipating that this would result in a significant speed-up for the common case. FTR, I made a thinko when stating in my initial comment that the tail runs' need to get "correction offsets" applied upon removal/insertion -- that's of course not needed.

As this only affects the internal representation and can be considered an implementation detail, it's entirely up to you whether to pursue that at this point or somewhen later perhaps. If you opt for now, note that the `Rust` core library provides some quite convenient primitives for doing binary searches, e.g. the [`slice::binary_search()`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) family or [`slice::partition_point()`](https://doc.rust-lang.org/std/primitive.slice.html#method.partition_point), so you don't have to roll your own at least.

kernel/src/protocols/core.rs
```diff
@@ -315,6 +331,12 @@ fn core_pvalidate_one(entry: u64, flush: &mut bool) -> Result<(), SvsmReqError>
 
     drop(lock);
 
+    // Update bitmap...
+    let value = (valid == PvalidateOp::Valid) as u8;
+    GUEST_VALID
+        .set(page_index, value)
+        .map_err(|_| SvsmReqError::FatalError(SvsmError::Fault))?;
+
```
nicstange: As this can fail, I would do the update before the pvalidate in order to avoid inconsistencies between the actual and the tracked validation status. If the then subsequent pvalidate failed, the prior `RleBits` update can get reverted, which is always guaranteed to succeed if I'm not mistaken (if not you might opt to panic as far as I'm concerned).

Also, to avoid races, move the `RleBits` update to under `PVALIDATE_LOCK` critical section right above protecting the `pvalidate`. In order to avoid the costs of an extra locking operation, it might perhaps even be possible to merge `GUEST_VALID` into `PVALIDATE_LOCK`. That is, make `PVALIDATE_LOCK` to wrap the `RleBits` instance if possible.

For the record, the memory allocator, which might get invoked by `RleBits`, does not take `PVALIDATE_LOCK` itself AFAICS, so there shouldn't be any recursion deadlock issue at least.

rlebits/src/lib.rs
nicstange: > Sorry, I wasn't very clear. I meant that checking the last 'count' (which is always 2) values are 0 is pretty cheap. Yes, the cost of the memmove is dependent on the length.

Ah, I was so focused on those tail memmoves mentally that I misread your comment :). 


> I wrote a test that did trip the bug you discovered. I then got energetic and added a max_run value to the structure. That turned out to be more complicated than it seemed at first, dealing with all the end cases that just naturally fell out before. But it's done now and I don't think the overhead of maintaining the value is too onerous. It makes the code harder to follow, which is always a trade-off I'm sensitive to. But I'll let you be the judge.

Thanks a lot! I'll have a look in a couple of days again on this and the other changes once everything has settled a bit.

rlebits/src/lib.rs
RARelph: I wonder if implementing the Index trait might be confusing... should it return the value of the bit at the index (ie, what RleBits.get() returns) or the value of the runs array at the index (which is what get_run() returns?

rlebits/src/lib.rs
RARelph: This will go in the next push.

rlebits/src/sync.rs
RARelph: What I'm planning to do here is remove thread-safe from the default build in Cargo.toml and then look at the SVSM SpinLock. I'd like to retain the thread-safety code, even if SVSM doesn't use it. Out of curiousity, given that spin exists, what motivated creating SpinLock?

rlebits/src/lib.rs
nicstange: Ack. Please ignore my comment then.

kernel/src/protocols/core.rs
nicstange: Note to myself, the TCG reference implementation simulator's `_rpc__Signal_PowerOn()`  apparently has the ability to issue a `_TPM_init` signal, so it might serve as a basis for us.

stefano-garzarella: Maybe we can just call `TcgTpm::signal_poweron()` (https://github.com/coconut-svsm/svsm/blob/968a9aa9b3c33f2c5bc0ff9d9ed44eb7c4a8e816/kernel/src/vtpm/tcgtpm/mod.rs#L129)  with the `only_reset` set to true.

cc @jejb

nicstange: Just for reference, when claiming that the TPM ought to be reset, I was referring to 10.2.2 ("Initialization State") of the TPM 2.0 Library part 1 ("Architecture"), v184

"Regardless of how it is generated, _TPM_Init should coincide with a reset of the Roots of Trust for
Measurement for which the TPM is the Root of Trust for Reporting. For example, if the TPM is a component
on the PC’s motherboard, _TPM_Init should coincide with a reset of the processor and chipset. After
_TPM_Init is indicated, the RTM should begin executing the Core Root of Trust for Measurement. It should
not be possible to reset the TPM without resetting the RTM. It should not be possible to reset the RTM without
resetting the TPM."

nicstange:
> On Thu, 2025-10-02 at 01:28 -0700, Nicolai Stange wrote: [...]
> Just for reference, when claiming that the TPM ought to be reset, I was referring to 10.2.2 ("Initialization State") of the TPM 2.0 Library part 1 ("Architecture"), v184
> Just on this one point, if you allow the TPM to fully reset across guest reboot then all the PCRs go back to zero and all measurements are lost. That means that the SVSM itself may not measure anything into the TPM because those measurements would be lost but, because the SVSM isn't resetting, nothing would redo them. 

Thanks for the explanation! I'm wondering, in theory, couldn't the SVSM itself redo the measurements upon guest cold restart (assuming it keeps its event log around somewhere)?

rlebits/src/lib.rs
RARelph: Converted array to vec throughout RleBits. Enhanced move_runs_up to support resizing the vec. Test if the resize will work first (using try_reserve), and return an error if not.

rlebits/src/lib.rs
RARelph: I've added a max_run value to the structure, which holds the index of the last non-zero run in the vec. move_runs_up() takes advantage of this new value to reduce the range of copy_within.

rlebits/src/lib.rs
RARelph: Done.

rlebits/src/lib.rs
RARelph: Done.

rlebits/src/lib.rs
RARelph: I am going to look in to this further, so unresolved this conversation to keep it on my radar.

kernel/src/protocols/core.rs
RARelph: OK, this is where I expected it to be 3 weeks ago. I have the new protocol in place. Now to write up what we discussed last Wednesday, get Tom to update the spec accordingly, and make what the implementation match that.

kernel/src/protocols/core.rs
RARelph: I can rearrange the code to do the update of GUEST_VALID first and then undo it if there's an error during the PVALIDATE instruction execution. As for using PVALIDATE_LOCK... I note that core_pvalidate_one() only acquires the lock as a reader (PVALIDATE_LOCK.lock_read();). I presume that's sufficient because the microcode for PVALIDATE provides protection against multiple writers. But GUEST_VALID needs only one writer at a time. I can wrap RleBits with PVALIDATE_LOCK, but I think I'll have to upgrade the PVALIDATE_LOCK to a lock_write. What do you think?

rlebits/SYNC_STRATEGIES.md
```diff
@@ -0,0 +1,163 @@
+# Scalability Alternatives to Global Spin Mutex
+
+You're absolutely right that a global spin mutex isn't scalable! Here are several more scalable alternatives for the RleBits thread-safe implementation:
```
00xc: This looks like an AI answer, I don't think it belongs in the SVSM repo

rlebits/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+    InvalidValue,
+    RunsOutOfSpace,
+    InternalError,
+    #[cfg(feature = "thread-safe")]
+    LockPoisoned,
+}
+
+#[derive(Debug, Clone)]
+pub struct RleBits {
+    runs: Vec<usize>, // Alternating run lengths, starting with 0s at index 0
```
00xc: Typical Rust convention is to have these comments on top of the field they document. You should also likely use `///` instead so that the documentation is generated properly.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+            return None;
+        }
+
+        for (i, &next) in self.runs.iter().enumerate() {
+            if index < next {
+                return Some((i % 2) == 1);
+            }
+        }
+
+        debug_assert!(false, "RleBits::get() - fell through all runs");
```
00xc: You could also do the following instead of asserting `false`:

```rust
#[cfg(debug_assertions)]
unreachable!("RleBits::get() - fell through all runs");
```

rlebits/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    pub fn set_range(
+        &mut self,
+        mut index: usize,
+        mut len: usize,
+        value: bool,
+    ) -> Result<(), RleBitsError> {
+        if (index.checked_add(len).is_none()) || (index + len > self.limit) {
+            return Err(RleBitsError::RangeExceedsLength);
+        }
```
00xc, line 143: I think this should be `index + len >= self.limit` (greater or equal, not just greater).

You can also combine both checks for simplicity (using `<` here is correct):

```rust
if index.checked_add(len).filter(|end| end < self.limit).is_none() {
    return Err(RleBitsError::RangeExceedsLength);
}
```

rlebits/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Usage
+
+### Basic Usage (no_std)
+
+```rust
+use rlebits::RleBits;
+
+// Create a bit array that can hold values 0-99
+let mut bits = RleBits::new(100);
```
00xc: This is out of sync with the actual implementation, where `RleBits::new()` takes 2 arguments.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+pub struct RleBits {
+    runs: Vec<usize>, // Alternating run lengths, starting with 0s at index 0
+    limit: usize,     // Upper limit of values, exclusive (ie, 0-9 is represented by 10)
+    max_run: usize,   // Maximum runs index currently in use
+}
+
+impl RleBits {
+    pub fn new(limit: usize, size: usize) -> Self {
+        let mut runs = vec![0usize; size];
+        runs[0] = limit;
```
00xc, line 61: This will panic if `RleBits::new()` is called with `size = 0`. I suggest you prevent this condition here, since there is a similar issue in the `reset()` method. You can either ask for `size: NonZeroUsize` (preferred) or simply check that size is not zero and return an error otherwise.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+        RleBits {
+            runs,
+            limit,
+            max_run: 0,
+        }
+    }
+
+    pub fn reset(&mut self) {
+        self.runs[0] = self.limit;
```
00xc, line 70: As mentioned above, this will panic if `RleBits::new()` is called with `size = 0`.

rlebits/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+        assert!(self.runs[self.max_run] == self.limit, "Last run != limit!");
+        assert!(runs == self.max_run + 1, "Run count incorrect!");
+        runs
+    }
+
+    #[cfg(any(feature = "std", test))]
+    pub fn dump_with<F>(&self, format_addr: F)
+    where
+        F: Fn(usize) -> std::string::String,
+    {
```
00xc, line 279: `String` is not the only type that can be printed. The correct trait bound here would be to ask that `F` returns anything that implements `std::fmt::Display`:

```rust
pub fn dump_with<F, S>(&self, format_addr: F)
where
    F: Fn(usize) -> S,
    S: std::fmt::Display,
{
    ...
}
```

rlebits/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+proptest = "1.7.0"
+rand = "0.9.2"
+
+# Integration tests with required features
+[[test]]
+name = "integration_tests"
+path = "tests/integration_tests.rs"
+
+# Note: Use 'cargo test --features std,thread-safe' to run all tests including thread-safe tests
+# Or use 'cargo test-all' alias for convenience
```
00xc: I don't think this belongs here, better added in the README.

rlebits/src/lib.rs
RARelph: I prefer the present formulation. And I believe > is correct. For a limit of 100, index of 99 with a length of 1 is valid.

rlebits/SYNC_STRATEGIES.md
RARelph: Ya, AI generated this entire file and I haven't done a good job pruning it. I've fixed this, though I've left the file in place.

rlebits/src/lib.rs
RARelph: Thanks. I've done this.

rlebits/src/lib.rs
RARelph: Done.

rlebits/README.md
RARelph: Thanks. Fixed.

rlebits/src/lib.rs
RARelph: I chose to check for 0 and change a 0 to a 1.

rlebits/src/lib.rs
RARelph: Fixed by making 10 the minimum size in new.

rlebits/src/lib.rs
00xc: As it is the following code panics:

```rust
    let mut rle = RleBits::new(1, 1);
    rle.set(0, true);
```

When I applied the proposed changed it did not crash, but I think you're right and it just happens to avoid the panic due to a logic error.

The panic happens here:

https://github.com/coconut-svsm/svsm/blob/d0cc30d7441920e4100fdddb30f25f3a8dadcbb6/rlebits/src/lib.rs#L198

RARelph: Good catch. This is a special case (vec.len=1) in a special case (setting ALL bits) in an unavoidable special case (set bit 0 true, requiring a 0 length run[0], the only 'included' run allowed to be 0). Sigh. The code assumes the vector has at least 2 elements and doesn't go through move_runs_up() which would attempt to expand the vector. Given the intended use cases, I'm just going to set the minimum size for a new vector to a more rational value... like 10. OK?

rlebits/src/lib.rs
RARelph: Changing anything < 10 to 10.

rlebits/Cargo.toml
RARelph: Removed.

kernel/src/protocols/core.rs
RARelph: Per our discussions in the Wednesday meetings, this first implementation will not modify TPM state at all on reboot.

kernel/src/protocols/core.rs
RARelph: I have re-arranged the code as described, but left in place the separate lock on GUEST_VALID since it is different.

rlebits/src/sync.rs
RARelph: I've replaced the ThreadSafeRleBits with RWLock and adjusted the rlebits implementation to not include the thread-safe feature by default.

PR #810, opened 2025-09-17
task: thread start safety

msft-jlange:
Enforce safety rules to ensure that the startup parameter for a new kernel thread can be interpreted safely, and require `unsafe` when the compiler cannot enforce startup parameter safety on its own.

### DIFF
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -41,6 +41,46 @@ use super::task_mm::{TaskKernelMapping, TaskMM};
 
 pub const INITIAL_TASK_ID: u32 = 1;
 
+/// This trait is used to describe a type that can be used as the start
+/// parameter for a kernel thread.  The thread start path requires that all
+/// start paremeters be able to fit into a single register, which constrains
+/// them to `usize`.  Other types can be used as long as there is an
+/// appropriate mechanism to covert those types to and from the `usize` that
+/// must be passed via the parameter register.  This trait defines such a
+/// conversion mechanism.
+///
+/// # Safety
+///
+/// Implementations of this thread must ensure that the `to_usize` and
+/// `from_usize` methods correctly preserve the safety of the parameter as it
+/// transitions from its native type to and from a temporary `usize`
+/// representation.
+pub unsafe trait KernelThreadStartParameter {
+    fn to_usize(parameter: Self) -> usize;
+
+    /// # Safety
+    /// This function must only be called with a `usize` parameter that was
+    /// previously returned from a call to `to_usize()`.
+    unsafe fn from_usize(parameter: usize) -> Self;
+}
+
+// SAFETY: types that implement `From` for conversions to and from `usize` are
+// trivially convertible to and from `usize` and thus will always preserve
+// safety of the parameter during conversion.
+unsafe impl<T> KernelThreadStartParameter for T
+where
+    T: From<usize>,
+    usize: From<T>,
+{
+    fn to_usize(parameter: Self) -> usize {
+        usize::from(parameter)
+    }
+
+    unsafe fn from_usize(parameter: usize) -> Self {
+        Self::from(parameter)
+    }
+}
+
 #[derive(Debug)]
 pub struct KernelThreadStartInfo {
     entry: usize,
@@ -49,14 +89,10 @@ pub struct KernelThreadStartInfo {
 }
 
 impl KernelThreadStartInfo {
-    pub fn new<T>(entry: fn(T), start_parameter: T) -> Self
-    where
-        T: From<usize>,
-        usize: From<T>,
-    {
+    pub fn new<T: KernelThreadStartParameter>(entry: fn(T), start_parameter: T) -> Self {
         Self {
             entry: entry as usize,
-            start_parameter: usize::from(start_parameter),
+            start_parameter: T::to_usize(start_parameter),
             start_routine: run_kernel_task::<T> as usize,
         }
     }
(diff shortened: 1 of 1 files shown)


kernel/src/task/tasks.rs
```diff
@@ -851,13 +886,21 @@ unsafe fn setup_new_task_common(xsa_addr: u64) {
...
 
-fn run_kernel_task(entry: fn(usize), xsa_addr: u64, start_parameter: usize) {
+/// # Safety
+/// This function must only be invoked as a thread start function, and must
+/// be invoked according to the start parameter data type expected by the entry
+/// point.
+unsafe fn run_kernel_task<T>(entry: fn(T), xsa_addr: u64, start_parameter: usize)
+where
+    T: From<usize>,
+    usize: From<T>,
```
stefano-garzarella: Not sure if `usize: From<T>,` here is a leftover, can we remove it?

kernel/src/task/tasks.rs
```diff
@@ -41,6 +41,45 @@ use super::task_mm::{TaskKernelMapping, TaskMM};
...
+        Self {
+            entry: entry as usize,
+            start_parameter: usize::from(start_parameter),
+            start_routine: run_kernel_task::<T> as usize,
+        }
+    }
+
+    /// # Safety
+    /// The caller is required to provide a start parameter that is appropriate
+    /// to the entry point according to the safety requriements of the entry
```
stefano-garzarella: s/requriements/requirements

kernel/src/task/tasks.rs
msft-jlange: You're right; this is not strictly needed here because the bound is enforced in `KernelThreadStartInfo::new()`.  It's certainly harmless here, and it provides symmetry with `KernelThreadStartInfo::new()` since the two functions must go hand-in-hand.  But I can remove it here; the compiler will make sure that the code doesn't deteriorate beyond what is truly required.

kernel/src/task/tasks.rs
```diff
@@ -851,13 +886,20 @@ unsafe fn setup_new_task_common(xsa_addr: u64) {
...
+where
+    T: From<usize>,
+{
     // SAFETY: the save area address is provided by the context switch assembly
     // code.
     unsafe {
         setup_new_task_common(xsa_addr);
     }
-    entry(start_parameter);
+    entry(T::from(start_parameter));
```
joergroedel: I see the point, but the constraining it to `From(usize)` seems limiting. Can you elaborate on the actual parameters you plan to pass this way? I think it would be better to require `T: Copy`, consume the start parameter in `run_kernel_task()` and copy it to the new tasks stack. This makes the start data private to the task (not thread) and avoid passing global pointers under hood.

msft-jlange: First, I will say that the primary motivation for this change is not to expand the set of types that can be passed as a thread startup parameter, but rather to enable the use of `unsafe` thread start functions.  In my particular case, I want to declare a thread start function that takes a `*const Foo` which requires that the function be `unsafe` because it's going to cast that pointer and start using it.  This also imposes a concern that the relationship between a thread start routine and its startup parameter carries safety considerations, because the parameter must be correct for the invocation of the thread, and thus any code establishes that pairing must be `unsafe`.  We really don't want to require that every thread start routine be `unsafe`, nor do we want to require that every thread creation routine be `unsafe`, so we need a way to divide thread startup into safe/unsafe categories.

The creation of `KernelThreadStartInfo` provides a convenient way to encapsulate the safety considerations, because the relationship between start routine and startup parameter is established when the `KernelThreadStartInfo` is created, and the tuple can safely be passed along the creation chain without having to reexamine or reevaluate the safety at every step.

Once we have divided into safe/unsafe versions of thread startup, the next reasonable question to ask is what is the scope of thread creation that can be entirely safe.  I thought it would be useful to go beyond pure `usize`, which is where this came from.

We can't say that a thread start parameter can be any `T: Copy` type because a thread start parameter is limited to a single usize, because it is passed via a single register in the frame that invokes the start routine.  See the following code in `allocate_ktask_stack()`:
```
            // start argument parameter.
            (*task_context).regs.rdx = start_parameter;
```
If we were to support any `T: Copy` type then we would have to allocate enough space in the target thread's stack to hold the copy of the startup parameter, and we would have to manipulate the register contents of the target thread's stack frame to hold the data correctly.  This is delicate, because it relies on a lot of complex compiler register convention rules, and would be beyond the scope of this PR in any case.  Instead, my approach is to support any type that can trivially be converted to or from a register-sized value (i.e. `usize`), which is exactly why this code is written in this way, since `From` defines trivial conversions to and from a type.  In practice, this really means that only primitive types like `u32` or `isize` can be used; we could also extend to types like `PhysAddr` or `VirtAddr` but functions that rely on those are likely inherently unsafe anyway (due to the need to prove the validity of the address) so that wouldn't be especially useful.

Note that passing references as start parameters must also be illegal, because a reference implies a lifetime (which will be enforced by the borrow checker), but there is no way for the compiler to comprehend the lifetime of a new thread, so it cannot enforce lifetime rules on the reference that is passed across the thread boundary.  This, of course, would also assume that the reference is even valid across the thread boundary, which is not a safe assumption across kernel tasks (because task stacks are not visible outside of the owning task).

We could choose to go beyond trivial `usize` conversions by defining a new trait that describes how to convert to/from usize for the purpose of passing the start parameter (e.g. `trait KernelThreadStartParameter` with methods `to_usize()` and `from_usize()`).  Types that support trivial conversion to/from `usize` would just use the existing `From` implementation, but other types could implement custom logic that is known to be sound (it would be possible, for example, to pass `Box` or `Arc` objects as thread start parameters).  While this may be interesting, I feel it is beyond the scope of this PR, because it can be added as a future extension.

We could also choose to scale down, and say that `usize` is the only allowed type for safe thread startup parameters, and then we could get rid of the generic thread start and the `From`-based conversions.  I'm not sure that's any better than the code that I have written, but I can strip it down if you think it's preferable.

msft-jlange:
I went ahead and implemented the `KernelThreadStartParameter` trait as described in the comments above.  Although it doesn't do anything different than the previous draft of the code, defining that trait created a convenient place to include the commentary from this PR, so it seemed worthwhile to add for the sake of keeping the code well commented.

kernel/src/task/tasks.rs
joergroedel: Thanks for the explanations and updating the PR. Approved now.

luigix25:
This PR broke `test-in-svsm`. I opened #822 to fix it.

PR #819, opened 2025-09-24
scripts/show-stack-sizes: fix RUSTFLAGS variable

stefano-garzarella:
Commit 56c0aae6 ("build: do not hardcode host target triple") added a new section in the .cargo/config.toml file, moving rustflags from the generic `[build]` section to a target-specific section `[target.x86_64-unknown-none]`.

Update the show-stack-sizes.sh script to use correct target-specific path.

### DIFF
--- a/scripts/show-stack-sizes.sh
+++ b/scripts/show-stack-sizes.sh
@@ -1,4 +1,4 @@
-#!/bin/env sh
+#!/bin/bash
 # SPDX-License-Identifier: MIT OR Apache-2.0
 #
 # Author: Tom Dohrmann <erbse.13@gmx.de>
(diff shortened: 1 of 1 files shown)


joergroedel:
Thanks, this mostly fixes it for me. One problem remains, which is that is uses bash-specifics for testing the `RELEASE` variable. Can you also hardcode the script to `bash` instead of `sh`?

stefano-garzarella:
> Thanks, this mostly fixes it for me. One problem remains, which is that is uses bash-specifics for testing the `RELEASE` variable. Can you also hardcode the script to `bash` instead of `sh`?

@joergroedel good catch, I just added another commit to fix it to this PR.

joergroedel (approved):

PR #820, opened 2025-09-24
sev/secrets_page: clean up stack usage

msft-jlange:
Change various copy operations on the secrets page to use pointers directly instead of using copy operations.  This ensures that a non-optimizing compiler will not blow up the stack by creating page-sized stack locals for the sake of copying data.

### DIFF
--- a/kernel/src/sev/secrets_page.rs
+++ b/kernel/src/sev/secrets_page.rs
@@ -12,6 +12,7 @@ use crate::types::GUEST_VMPL;
 
 extern crate alloc;
 use alloc::boxed::Box;
+use core::ptr;
 
 pub const VMPCK_SIZE: usize = 32;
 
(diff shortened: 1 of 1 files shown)


kernel/src/sev/secrets_page.rs
```diff
@@ -87,12 +88,24 @@ impl SecretsPage {
...
         unsafe {
-            *to = *self;
+            to.copy_from(ptr::from_ref(self), 1);
         }
     }
 
     pub fn copy_for_vmpl(&self, vmpl: usize) -> Box<SecretsPage> {
-        let mut sp = Box::new(*self);
+        // SAFETY: the new box is uninitialized so data can be copied into it,
+        // which will complete the initiazation process and make it ready for
```
stefano-garzarella:
```suggestion
        // which will complete the initialization process and make it ready for
```

stefano-garzarella (approved):
Just a little typo in a comment, but the code LGTM. Tests passes with qemu (both test-in-svsm that linux boot). Thanks!

stefano-garzarella (approved):

PR #822, opened 2025-09-25
kernel/task: fix `start_kernel_task` parameters

luigix25:
Commit fd2152881e8c ("task: encapsulate thread start information for safety") modified the parameters for `start_kernel_task`. Two calls were not updated accordingly, breaking `test-in-svsm`.

Update them using `KernelThreadStartInfo`.

### DIFF
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -955,7 +955,7 @@ fn task_exit() {
 #[cfg(all(test, test_in_svsm))]
 mod tests {
     extern crate alloc;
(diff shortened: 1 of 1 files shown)


luigix25:
> @luigix25 oops, it seems something is missing when we install toolchains. If you want, feel free to move that patch to another PR, so we can discuss more if it makes sense or not, while we can fix quickly the main issue, so we can run `make test-in-svsm` on other PRs.

Yeah, I'm working on a fix, but for the meanwhile I'll remove the commit.

stefano-garzarella (approved):

PR #823, opened 2025-09-26
github/workflows: Build test images

luigix25:
Build test images in the CI to make sure that a PR doesn't break tests.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,7 +20,9 @@ jobs:
           submodules: recursive
 
       - name: Install rust toolchain
-        run: rustup toolchain install --profile minimal
+        run: |
+          rustup +nightly target add x86_64-unknown-none
+          rustup toolchain install --profile minimal
 
       - name: Add rust components
         run: rustup component add rustfmt rust-src clippy
@@ -39,6 +41,9 @@ jobs:
 
       - name: Build release image
         run: cargo xbuild --release ./configs/*
+
+      - name: Build test images (test-in-svsm)
+        run: make test-igvm
 
       - name: Run tests
         run: make test


.github/workflows/rust.yml
```diff
@@ -20,7 +20,9 @@ jobs:
           submodules: recursive
 
       - name: Install rust toolchain
-        run: rustup toolchain install --profile minimal
+        run: |
+          rustup +nightly target add x86_64-unknown-none
```
stefano-garzarella: Can we avoid this and use the `nightly-check` pipeline to run `make test-igvm` since we already install the nightly toolchain in that pipeline?

stefano-garzarella: Please also mention in the commit description that we need a nightly toochain (and maybe why).

luigix25: > Can we avoid this and use the `nightly-check` pipeline to run `make test-igvm` since we already install the nightly toolchain in that pipeline?

In that case I'd need to install the TPM build dependencies in that pipeline. Do you still prefer to move it to the other pipeline?

stefano-garzarella: mmm, good question, I'd say yes.
I mean, I like we have 2 different pipelines to not mix stable and nightly toolchains and if we need a nightly toolchain to build `test-igvm` IMO we should do that in `nightly-check`.

@joergroedel WDYT?

luigix25: not needed anymore, I am now using the qemu workflow.

Makefile
```diff
@@ -4,7 +4,15 @@ SVSM_ARGS += --features ${FEATURES}
...
 
+#We don't support virtio in nocc mode
+ifeq ($(NOCC),1)
+NOCC=--nocc
+FEATURES_TEST ?= vtpm
+else
+NOCC=
 FEATURES_TEST ?= vtpm,virtio-drivers
+endif
+
```
stefano-garzarella: Can we avoid this at all and do `make FEATURES_TEST="vtpm"` in .github/workflows/qemu.yml

kernel/src/cpu/vc.rs
```diff
@@ -338,11 +338,13 @@ mod tests {
     #[test]
     #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
     fn test_has_memory_encryption_info_cpuid() {
-        const CPUID_EXTENDED_FUNCTION_INFO: u32 = 0x8000_0000;
-        const CPUID_MEMORY_ENCRYPTION_INFO: u32 = 0x8000_001F;
-        // SAFETY: CPUID does never affect safety.
-        let extended_info = unsafe { __cpuid_count(CPUID_EXTENDED_FUNCTION_INFO, 0) };
-        assert!(extended_info.eax >= CPUID_MEMORY_ENCRYPTION_INFO);
+        if !is_test_platform_type(SvsmPlatformType::Native) {
```
stefano-garzarella: I'd prefer an early return, but I see we use `is_test_platform_type` mostly in this way, so doesn't metter.

.github/workflows/qemu.yml
```diff
@@ -108,14 +108,15 @@ jobs:
 
       - name: Install Rust x86_64-unknown-none
```
stefano-garzarella: Please update also the title here, and I'd copy what we do in https://github.com/coconut-svsm/svsm/blob/579de9f3ead1d7fc596965d818f465979e00bea1/.github/workflows/rust.yml#L69

.github/workflows/qemu.yml
stefano-garzarella, line 1: > Unfortunately virtio-drivers are not available for nocc.

I'd add the reason.
Also in this commit is not clear what it means, since we disabled it in the previous commit, but if we avoid the NOCC var, it makes sense here where we will use `make FEATURES_TEST="vtpm" ...`

Makefile
luigix25: yes, but in this way you can run `make test-in-svsm` locally on a non-snp machine.

kernel/src/cpu/vc.rs
luigix25: As you prefer. I tried to stick to what others did in the code

Makefile
stefano-garzarella: That we don't use in the CI, so I'd avoid adding new parameters when you have a way to do that.
Also, IIUC the missing part is just #775 that I hope we will merge soon.

stefano-garzarella:
> update: if some of the tests doesn't pass (and panics the kernel) the GH action will run until timeout (180s). Investigating..

@luigix25 IIRC @osteffenrh added the `timeout` in `scripts/test-qemu-nocc-boot.sh` because QEMU was never exiting with just the boot. But with `test-in-svsm` IIRC, QEMU will exit at the end of the tests, so maybe we don't need that script anymore. (And if we need, maybe we need to update the name)

luigix25:
> > update: if some of the tests doesn't pass (and panics the kernel) the GH action will run until timeout (180s). Investigating..
> 
> @luigix25 IIRC @osteffenrh added the `timeout` in `scripts/test-qemu-nocc-boot.sh` because QEMU was never exiting with just the boot. But with `test-in-svsm` IIRC, QEMU will exit at the end of the tests, so maybe we don't need that script anymore. (And if we need, maybe we need to update the name)

If everything goes well, then `test-in-svsm` exits correctly, but if something goes wrong it panics and qemu remains open. I'm modifying Oliver's script to detect a panic.

Makefile
luigix25: yep, that is the only missing part. I'll remove this part.

stefano-garzarella:
> > > > update: if some of the tests doesn't pass (and panics the kernel) the GH action will run until timeout (180s). Investigating..
> > > 
> > > 
> > > @luigix25 IIRC @osteffenrh added the `timeout` in `scripts/test-qemu-nocc-boot.sh` because QEMU was never exiting with just the boot. But with `test-in-svsm` IIRC, QEMU will exit at the end of the tests, so maybe we don't need that script anymore. (And if we need, maybe we need to update the name)
> > 
> > 
> > If everything goes well, then `test-in-svsm` exits correctly, but if something goes wrong it panics and qemu remains open. I'm modifying Oliver's script to detect a panic.
> 
> Strange, in the panic handler we have `platform::halt();` (
> 
> https://github.com/coconut-svsm/svsm/blob/579de9f3ead1d7fc596965d818f465979e00bea1/kernel/src/svsm.rs#L422
> ), so I'm expecting QEMU to exit on `hlt`, no?

no, so maybe we should call `testing:exit()`, we need to write in the `QEMU_EXIT_PORT`

luigix25:
> > 
> > Strange, in the panic handler we have `platform::halt();` (
> > https://github.com/coconut-svsm/svsm/blob/579de9f3ead1d7fc596965d818f465979e00bea1/kernel/src/svsm.rs#L422
> > 
> > ), so I'm expecting QEMU to exit on `hlt`, no?
> 
> no, so maybe we should call `testing:exit()`, we need to write in the `QEMU_EXIT_PORT`

Exactly, you have to write in the `QEMU_EXIT_PORT`. You mean calling it every time the kernel panics?

kernel/src/testing.rs
```diff
@@ -92,9 +94,19 @@ pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
...
-            .unwrap();
+
+        match SVSM_PLATFORM.platform_type() {
+            SvsmPlatformType::Snp => {
+                current_ghcb()
+                    .ioio_out(QEMU_EXIT_PORT, GHCBIOSize::Size32, 0)
+                    .unwrap();
+            }
+            SvsmPlatformType::Native => {
+                SVSM_PLATFORM.get_io_port().outl(QEMU_EXIT_PORT, 0);
```
osteffenrh: I know this value was not chosen in this PR, but I think we should use something else here, since this causes QEMU to exit with code `1`, which is also used for general (startup) errors. AFAIK `exit_code = (port_value << 1) | 1`. See https://os.phil-opp.com/testing/#exiting-qemu  (Not sure where the actual QEMU documentation for this is).
But we can address this is a separate PR (if we want to change it).

stefano-garzarella: I like the idea! In this way, we can also easily differentiate when we exit from the panic handler and when all tests completed successfully!

If we add the exit in the panic handler in this PR (I'm in favor of that), I think we should implement your suggestion too.

luigix25: I like this idea too! Thanks for the suggestion!

stefano-garzarella:
@luigix25 we just merged #775 so I think we can now rebase and enable also `virtio-drivers` feature, right?

luigix25:
> @luigix25 we just merged #775 so I think we can now rebase and enable also `virtio-drivers` feature, right?

Thanks! yep, I am already working on this and some minor fixes

kernel/src/utils/util.rs
```diff
@@ -86,6 +89,22 @@ pub unsafe fn zero_mem_region(start: VirtAddr, end: VirtAddr) {
     unsafe { start.as_mut_ptr::<u8>().write_bytes(0, count) }
 }
 
+pub const QEMU_EXIT_SUCCESS: u32 = 0x10;
+pub const QEMU_EXIT_FAIL: u32 = 0x0;
```
osteffenrh: This will still map to exit code 1 of QEMU, which is used by QEMU itself (for example if the command line is wrong).
I we use something else here (`0x11` maybe?) we can distinguish QEMU start failure from SVSM test failure, too.

kernel/src/utils/util.rs
```diff
@@ -86,6 +89,22 @@ pub unsafe fn zero_mem_region(start: VirtAddr, end: VirtAddr) {
     unsafe { start.as_mut_ptr::<u8>().write_bytes(0, count) }
 }
 
+pub const QEMU_EXIT_SUCCESS: u32 = 0x10;
+pub const QEMU_EXIT_FAIL: u32 = 0x11;
+
+#[cfg(test)]
+pub fn qemu_write_exit(value: u32) {
```
stefano-garzarella: What about using an `enum` as parameter, and here do the mapping to u32 values that QEMU expects?

Makefile
```diff
@@ -31,6 +31,12 @@ else ifeq ($(V), 2)
 CARGO_ARGS += -vv
 endif
 
+ifeq ($(NOCC),1)
+NOCC=--nocc
+else
+NOCC=
+endif
+
```
stefano-garzarella: Can we just add a parameter that we can eventually reuse in the future, instead of adding custom parameter for every new one.

I mean something like `CLIPPY_ARGS` or `FEATURES_TEST`, in this case we can name it something like `TEST_ARGS`.

scripts/test-in-svsm.sh
```diff
@@ -47,9 +47,37 @@ dd if=/dev/urandom of="$TEST_DIR/svsm_state.raw" bs=512 count=1024
 test_io $TEST_DIR/pipe.in $TEST_DIR/pipe.out &
 TEST_IO_PID=$!
 
-$SCRIPT_DIR/launch_guest.sh --igvm $SCRIPT_DIR/../bin/coconut-test-qemu.igvm \
+NOCC=""
```
stefano-garzarella: Same here, `LAUNCH_GUEST_ARGS`, so we can reuse it to add new parameter in the future.

scripts/test-in-svsm.sh
```diff
@@ -47,9 +47,37 @@ dd if=/dev/urandom of="$TEST_DIR/svsm_state.raw" bs=512 count=1024
...
+      ;;
+    *)
+      echo "Invalid parameter $1"
+      exit 1
+      ;;
+  esac
+done
+
+
+if $SCRIPT_DIR/launch_guest.sh --igvm $SCRIPT_DIR/../bin/coconut-test-qemu.igvm \
```
stefano-garzarella: Can we execute the command, then check `$?` ?

scripts/test-in-svsm.sh
```diff
@@ -47,9 +47,37 @@ dd if=/dev/urandom of="$TEST_DIR/svsm_state.raw" bs=512 count=1024
...
     --state "$TEST_DIR/svsm_state.raw" \
-    --unit-tests $TEST_DIR/pipe || true
+    --unit-tests $TEST_DIR/pipe \
+    $NOCC ; then
+    exit_value=0
+# SVSM returns 33 when everything goes well
```
stefano-garzarella: QEMU returns it IIUC, so maybe here we can write that `SVSM writes 33 to the QEMU tested exit port when all tests passed` (or something like that)

scripts/test-in-svsm.sh
stefano-garzarella: Also, what is the difference between this branch and the next one (33 exit value)?

.github/workflows/qemu.yml
```diff
@@ -106,20 +106,18 @@ jobs:
...
+      - name: Install latest nightly
+        run: rustup toolchain install nightly -t x86_64-unknown-none --profile minimal --force -c rustfmt
 
       - name: Install TPM 2.0 Reference Implementation build dependencies
         run: sudo apt install -y autoconf autoconf-archive pkg-config build-essential automake
 
-      - name: Build SVSM
-        run: cargo xbuild configs/qemu-target.json
+      - name: Build test
+        run: NOCC=1 make bin/coconut-test-qemu.igvm
```
stefano-garzarella: Why we need `NOCC=1` here?

scripts/test-in-svsm.sh
stefano-garzarella: > Also, what is the difference between this branch and the next one (33 exit value)?

I mean, I don't understand why for both cases we set `exit_value=0`, I'm missing something.

luigix25: > Can we execute the command, then check `$?` ?

Not really. QEMU returns a non-zero value for success and because of the `-e` option the script would stop executing.
`|| true` would overwrite the return value with 0.

So that `if` is a trick to execute the script and to check the result, The "side-effect" is an unused branch.
I really didn't want to disable `-e` and enable it again afterwards.

.github/workflows/qemu.yml
luigix25: oops, leftover

scripts/test-in-svsm.sh
stefano-garzarella: What about something like this?

```diff
diff --git a/scripts/test-in-svsm.sh b/scripts/test-in-svsm.sh
index b63aec0d..17b597d3 100755
--- a/scripts/test-in-svsm.sh
+++ b/scripts/test-in-svsm.sh
@@ -63,13 +63,13 @@ while [[ $# -gt 0 ]]; do
 done
 
 
-if $SCRIPT_DIR/launch_guest.sh --igvm $SCRIPT_DIR/../bin/coconut-test-qemu.igvm \
+$SCRIPT_DIR/launch_guest.sh --igvm $SCRIPT_DIR/../bin/coconut-test-qemu.igvm \
     --state "$TEST_DIR/svsm_state.raw" \
     --unit-tests $TEST_DIR/pipe \
-    $NOCC ; then
-    exit_value=0
+    $NOCC || svsm_exit_code=$?
+
 # SVSM returns 33 when everything goes well
-elif [ $? -eq 33 ]; then
+if [ $svsm_exit_code -eq 33 ]; then
     echo "All tests passed"
     exit_value=0
 else
```

luigix25: it works! Thanks for the hint!

kernel/src/testing.rs
```diff
@@ -90,12 +88,7 @@ pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
 }
 
 fn exit() -> ! {
-    if has_qemu_testdev() {
-        const QEMU_EXIT_PORT: u16 = 0xf4;
-        current_ghcb()
-            .ioio_out(QEMU_EXIT_PORT, GHCBIOSize::Size32, 0)
-            .unwrap();
-    }
+    qemu_write_exit(QEMUExitValue::Success);
```
stefano-garzarella, line 91: IIUC we are changing the behavior, before this patch we wrote 0x0, now we are writing 0x10, right?

Can you point this out in the commit message (also with the reason).

kernel/src/utils/util.rs
```diff
@@ -6,6 +6,9 @@
 
 use crate::address::{Address, VirtAddr};
 use crate::types::PAGE_SIZE;
+#[cfg(test)]
```
stefano-garzarella: Why we are adding this in `util.rs`?

Can we use some modules used just for testing?
Or maybe create a `test.rs` under `kernel/src/utils/`.

kernel/src/utils/util.rs
```diff
@@ -99,9 +105,21 @@ pub enum QEMUExitValue {
...
+                    .ioio_out(QEMU_EXIT_PORT, GHCBIOSize::Size32, value as u64)
+                    .unwrap();
+            }
+            SvsmPlatformType::Native => {
+                SVSM_PLATFORM
+                    .get_io_port()
+                    .outl(QEMU_EXIT_PORT, value as u32);
+            }
+
+            _ => {}
```
stefano-garzarella: Can we add all the matches (e.g. Tdx) and call `unimplemented!()` so we will remember to implement it if one day we will need it?

scripts/test-in-svsm.sh
```diff
@@ -47,9 +47,37 @@ dd if=/dev/urandom of="$TEST_DIR/svsm_state.raw" bs=512 count=1024
...
+
+
 $SCRIPT_DIR/launch_guest.sh --igvm $SCRIPT_DIR/../bin/coconut-test-qemu.igvm \
     --state "$TEST_DIR/svsm_state.raw" \
-    --unit-tests $TEST_DIR/pipe || true
+    --unit-tests $TEST_DIR/pipe \
+    $LAUNCH_GUEST_ARGS || svsm_exit_code=$?
+
+# SVSM writes 33 to the QEMU exit port when all tests passed
+if [ $svsm_exit_code -eq 33 ]; then
```
stefano-garzarella: SVSM currently writes 0x10, so it's not really why we are checking 33 here, can you explain it at least in the comment?

luigix25: qemu takes that value and returns `(value << 1) | 1` so 0x10 becomes 0x21.
I'll add a comment

kernel/src/testing.rs
luigix25: > IIUC we are changing the behavior, before this patch we wrote 0x0, now we are writing 0x10, right?

Correct. Writing 0 means that QEMU will return 1, that imho is weird for success. I'll explain it in the commit

stefano-garzarella: mmm, I don't see how 33 is less weird than 1 for sucess, but maybe I'm missing something...

luigix25: so, what do you suggest?
0 for success (that becomes 1)
1 for failure (that becomes 3) ?

stefano-garzarella: Okay, re-reading @osteffenrh motivation in https://github.com/coconut-svsm/svsm/pull/823#discussion_r2394017961 made it much more clear about the reason you're changing it.

LGTM (using something different from 0), but please add that motivation in the commit message.

stefano-garzarella: And maybe use a separate commit for that. IMHO it's strange that a commit that does a refactoring, changes also the behavior.

kernel/src/utils/util.rs
luigix25: `kernel/src/testing.rs` module is exported only when `test` and `test_in_svsm` are set. This is why I ended up using `util.rs`

`kernel/src/testutils.rs` is exported only when `test` is set.

So I see two options here: creating a `test.rs`, or always export `testutils.rs`. I'd go with the second one.

WDYT?

stefano-garzarella: So maybe you can use ` kernel/src/testing.rs`  and do:
```
    loop {
        debug_break();
        #[cfg(test)]
        qemu_write_exit(QEMUExitValue::Fail);
        platform::halt();
    }
```

Like we do in svsm_main() when calling test_main().

luigix25: This makes the stub unnecessary, much cleaner!

luigix25:
Addressed Stefano's comments:

- I moved `qemu_write_exit` to `kernel/src/testing.rs`
- Removed the stub, it is not necessary anymore.
- Refactored `qemu_write_exit`: there is no need to differentiate per platform anymore.
- Added comment explaining the QEMU value
- Introduced a commit for QEMU exit value

kernel/src/testing.rs
```diff
@@ -99,8 +99,8 @@ fn exit() -> ! {
 #[repr(u32)]
 #[derive(Debug)]
 pub enum QEMUExitValue {
-    Success = 0x0,
-    Fail = 0x1,
+    Success = 0x10,
```
stefano-garzarella: Please, move this change before the refactoring. I'd avoid adding new code (e.g.  `enum QEMUExitValue`) in a commit and change it in a following commit of the same PR.

kernel/src/testing.rs
```diff
@@ -108,8 +106,8 @@ pub enum QEMUExitValue {
 pub fn qemu_write_exit(value: QEMUExitValue) {
     if has_qemu_testdev() {
         const QEMU_EXIT_PORT: u16 = 0xf4;
-        current_ghcb()
-            .ioio_out(QEMU_EXIT_PORT, GHCBIOSize::Size32, value as u64)
-            .unwrap();
+        SVSM_PLATFORM
```
stefano-garzarella, line 109: Ditto, move this change before the refactoring.

kernel/src/svsm.rs
stefano-garzarella, line 1: Please fix https://github.com/coconut-svsm/svsm/pull/823/commits/f67c1e96236298f44688b63557f75878cfb3bfe6 description. From the title, it seems it does only the refactoring, while the real change here is the qemu_write_exit call the panic handler. I think the title should talk more about that than the refactoring. (Or split this in 2 commits, but I'm fine also with a single commit with a title fixed).

scripts/test-in-svsm.sh
```diff
@@ -47,9 +47,38 @@ dd if=/dev/urandom of="$TEST_DIR/svsm_state.raw" bs=512 count=1024
...
+
 $SCRIPT_DIR/launch_guest.sh --igvm $SCRIPT_DIR/../bin/coconut-test-qemu.igvm \
     --state "$TEST_DIR/svsm_state.raw" \
-    --unit-tests $TEST_DIR/pipe || true
+    --unit-tests $TEST_DIR/pipe \
+    $LAUNCH_GUEST_ARGS || svsm_exit_code=$?
+
+# SVSM writes 0x10 to the QEMU exit port when all tests passed.
+# This results in QEMU returning 0x21 ((0x10 << 1) | 1)
+if [ $svsm_exit_code -eq 33 ]; then
```
stefano-garzarella: What about using the hex value also in the code?
e.g. `if [[ $svsm_exit_code -eq 0x21 ]]; then`

.github/workflows/qemu.yml
```diff
@@ -106,20 +106,18 @@ jobs:
...
         run: sudo apt install -y autoconf autoconf-archive pkg-config build-essential automake
 
-      - name: Build SVSM
-        run: cargo xbuild configs/qemu-target.json
+      - name: Build test
+        run: make bin/coconut-test-qemu.igvm
 
       ########################
 
       - name: Run SVSM in QEMU
```
stefano-garzarella:
```suggestion
      - name: Run test-in-svsm in QEMU
```

kernel/src/svsm.rs
luigix25: I reworded the commit message, hope it's more clear now!

luigix25:
- Reordered commits according to Stefano's suggestions.
- Reworded a commit message
- Very minor changes to bash and yaml

kernel/src/testing.rs
```diff
@@ -87,10 +87,19 @@ pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
     exit();
 }
 
+#[repr(u32)]
+#[derive(Debug)]
+pub enum QEMUExitValue {
```
stefano-garzarella, line 101: Please move this to the next commit, here just fix the issue.
In general, we avoid mixing new code with fixes, to avoid problems with backports (it's true, we don't have stable branches yet, but it's good practice to follow).

stefano-garzarella (commented):
LGTM, just a minor comment.

stefano-garzarella (approved):
@joergroedel I think we can merge this, but I'll wait for your review.

Note the CI time is not affected by this PR, since running tests is just 2 sec (see `Run test-in-svsm in QEMU` in https://github.com/coconut-svsm/svsm/actions/runs/18306479849/job/52124795547?pr=823)

PR #824, opened 2025-09-26
igvmbuilder/ovmf: Add IGVM Memory Map metadata type

peterfang:
The IGVM Memory Map metadata type describes the starting address and the length of the memory region that holds the guest's IGVM memory map. SVSM is expected to provide the guest with this information before the guest launches. The guest's OVMF is expected to use this memory map, if it exists, as the source of its system memory map.

This allows SVSM to convey the guest's memory map through IGVM using both OVMF and IGVM firmware.

Also, introduce per-platform memory map in `IgvmParamBlockFwInfo` to support multi-platform IGVM files. This info is filled in by igvmbuilder as it parses the firmware. During SVSM runtime, the guest's IGVM memory map location is retrieved from `IgvmParamBlockFwInfo` based on the underlying platform.


kraxel:
Where are the corresponsing edk2 changes?

IMHO it does not make sense to add the igvm memory map to SEV metadata.  SEV and IGVM are independent from each other.

FYI: I have a edk2 PR pending which adds igvm parameter support (memory map and vcpu count) to ovmf:
https://github.com/tianocore/edk2/pull/11487
That uses a separate table (format is identical to the sev table though) for igvm support.

kraxel:
My PR was just merged into master.

I'm wondering whenever the coconut edk2 fork should be archived to avoid confusion?
I think there is little reason to /not/ work against upstream tianocore repo, or do we have important svsm changes which are not merged upstream?
@stefano-garzarella

kraxel:
Hmm, that coconut assumtion does not match the igvm spec.  parameter areas are page sized and page aligned, but one parameter area are can carry multiple parameters which do not have any aligment/size constrains.  And in fact ovmf actually supports two parameters in the parameter page: the memory map and the vcpu count.

I think coconut should be more flexible here and accept what the igvm spec allows.

peterfang:
> I think coconut should be more flexible here and accept what the igvm spec allows.

Makes sense - I'll resolve this in COCONUT rather than edk2. Thanks.

kernel/src/igvm_params.rs
```diff
@@ -199,47 +196,73 @@ impl IgvmParams<'_> {
...
-        if self.page_state_change_required() {
-            SVSM_PLATFORM.page_state_change(
-                mem_map_region,
-                PageSize::Regular,
-                PageStateChangeOp::Private,
-            )?;
-        }
+        let preval_count = self.igvm_param_block.firmware.prevalidated_count as usize;
+        if !self
+            .igvm_param_block
```
msft-jlange: The determination of whether the memory map needs to be validated is based on the construction of the firmware, i.e. it is known at the time the IGVM file is built.  Can we do this scan in the IGVM builder and simply encode the bool in the firmware parameters instead of having to do this work at runtime?

peterfang: This makes sense. I'll add a bool to `struct IgvmParamBlockFwInfo` so we don't have to do this at runtime.

msft-jlange (approved):

PR #826, opened 2025-09-29
mm: Allow RO regions to start on HugePage boundary

osteffenrh:
Remove an assert that forbids memory regions starting on a HugePage
boundary to be marked as read-only.

This restriction does not make sense and also does not match the comment
above it.

I am hitting this assert because, as it seems, the svsm binary size is reaching a critical size for me.

Adding one line of code (`log::info!()`) causes this:
```
[SVSM] CPU 2 is online
[SVSM] Launching AP 3 with APIC-ID 3
[SVSM] CPU 3 is online
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 2025.09-devel-14-ga6045d82+
[SVSM] ERROR: Info: panicked at kernel/src/mm/pagetable.rs:1246:9:
assertion failed: !region.start().is_aligned(usize::from(PageSize::Huge))
[SVSM] ---BACKTRACE---:
[SVSM]   [ffffff80000725be]
[SVSM]   [ffffff80000902cb]
[SVSM]   [ffffff8000090448]
[SVSM]   [ffffff800003ea5e]
[SVSM]   [ffffff8000060520]
[SVSM] ---END---
```
Taking away another call somewhere makes it work again.

`region.start()` = 0xffffff80003ff000 works, this one (1 page larger) not: 0xffffff8000400000

### DIFF
--- a/kernel/src/mm/pagetable.rs
+++ b/kernel/src/mm/pagetable.rs
@@ -1241,10 +1241,6 @@ impl PageTable {
         &mut self,
         region: MemoryRegion<VirtAddr>,
     ) -> Result<(), SvsmError> {
-        // Since 4k alignment is explicitly mentioned in the safety requirements,
-        // let's use a debug assert here.
-        debug_assert!(!region.start().is_aligned(usize::from(PageSize::Huge)));
-
         for page in region.iter_pages(PageSize::Regular) {
             match self.walk_addr(page) {
                 Mapping::Level0(entry) => {


p4zuu:
> @p4zuu, can this be removed, or what is the intention behind it?

The idea was to catch the cases where the `ro_after_init` section was mapped as a huge page. Since this was expected to be technically possible but unwanted and rare, we decided that a `debug_assert` would be sufficient. The code that makes the memory region read-only relies on page table walk that depends on the the page size, and since we only wanted to support this for 4k pages, we added a few checks to break before going over a page table walk with the wrong page size.

> region.start() = 0xffffff80003ff000 works, this one (1 page larger) not: 0xffffff8000400000

This is actually strange, and not the behavior we want. There is maybe a bug somewhere

osteffenrh:
> @osteffenrh You mean the debug_assert fails if the page is 2M aligned, is that correct?

Yes.

stefano-garzarella:
> > @p4zuu I found the assertion (and the comment) difficult to understand TBH. How can we prevent a 4k page (aligned to 4k) from also being aligned to 2M by chance?
> 
> There were many discussions about this in the original PR: [#655 (comment)](https://github.com/coconut-svsm/svsm/pull/655#discussion_r2099826657) The TLDR version is: we'll cross that bridge when we come to it (huge page boundary). I guess that means now :)

@p4zuu yeah, exactly, Thanks for clarifying it :-)

PR #528, opened 2024-11-21
Attestation driver and proxy (with KBS attestation)

tylerfanelli:
This series introduces the attestation module in SVSM as well as the attestation proxy to facilitate communication between SVSM and a remote server for TEE evidence evaluation.

### Abstract

To unlock early persistent state in SVSM (for example, persistent vTPM data), an attestation of the launch environment must take place to ensure that SVSM is running on trusted hardware and launched correctly. As SVSM does not have a network stack available, we have opted to introduce a proxy that would run on the host (for now, although this may change as other options are explored) and facilitate attestation between SVSM and a remote verification server such as KBS.

Much of what is implemented here is described in the [Early Attestation and Measurement Architecture in COCONUT SVSM document](https://docs.google.com/document/d/11ZsxP8jsviP3ddp9Hrn0rf6inttNw_Pbnz0psXlxlPs/edit?usp=sharing).

### Changes

There are three main modules introduced into SVSM with this series:

- `kernel/attest.rs`: Attestation driver that communicates with the proxy through the COM3 serial port. This will eventually communicate over another channel (virtio-vsock, etc..) when available.
- `aproxy`: Attestation proxy that runs on a host and facilitates communication with a remote attestation server.
- `libaproxy`: Shared types between `kernel/attest.rs` and `aproxy` for (de)serialization of data.

### Attestation

There exists two different phases of attestation:

- **Negotiation**: Attestation servers usually require that a client embed metadata into its attestation evidence. The negotiation phase (initiated by SVSM) defines what exactly should be included in TEE attestation evidence. For instance, a server may require that a client embed a nonce hash and public components of the TEE public key (used to encrypt secrets) into the attestation evidence to ensure it is fresh and legitimate. The negotiation phase returns the parameters that SVSM should include in its attestation evidence based on the underlying attestation server and protocol.
- **Attestation**: With all relevant data embedded in TEE evidence, SVSM sends its evidence to the remote server for evaluation. Upon successful attestation, the proxy will obtain an encrypted secret (only decryptable by SVSM's TEE key) for SVSM to use. For example, SVSM could use this secret to unlock encrypted persistent storage.

As there exists multiple protocols for TEE attestation, the proxy was built to be configurable to different protocols. As such, SVSM can be completely agnostic of the attestation protocol used. This is done by splitting the proxy up into two main components:

[...]

PR #828, opened 2025-09-30
attestation: Update kernel module and proxy for AES-GCM, other small changes

tylerfanelli:
This PR updates the attestation kernel module and proxy with two main modifications:
- Updating the kernel module and proxy to use AES-GCM as the standard encryption mechanism protecting secrets received into the SVSM kernel once successfully attested.
- Updating the proxy to use standard KBS 0.4.0 as used by the Trustee server.

The `kbs-test` server was also updated to emulate KBS 0.4.0 and thus can be used to test this series.

cc/ @stefano-garzarella @nicstange @berrange

### Start `kbs-test`
```
$ git clone https://github.com/tylerfanelli/kbs-test.git && cd kbs-test
$ git checkout main
$ cargo build
$ export SVSM=/path/to/svsm
$ MEASUREMENT="$(${SVSM}/bin/igvmmeasure --check-kvm ${SVSM}/bin/coconut-qemu.igvm measure -b)"
$ cargo run -- -m $MEASUREMENT
```

### Start the attestation proxy
```
$ git clone https://github.com/tylerfanelli/svsm.git && cd svsm
$ git checkout attest-update
$ make aproxy
$ bin/aproxy --protocol kbs --url http://0.0.0.0:8080 --unix /tmp/svsm-proxy.sock --force
```

### Run SVSM with attestation

```
$ FW_FILE=/path/to/OVMF.fd make FEATURES=attest 
$ ./scripts/launch_guest.sh --qemu /path/to/qemu-system-x86_64 --image /path/to/qcow2-file.qcow2 --aproxy /tmp/svsm-proxy.sock
```

### DIFF
--- a/Documentation/docs/developer/ATTESTATION.md
+++ b/Documentation/docs/developer/ATTESTATION.md
@@ -36,13 +36,168 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
 should include in its attestation evidence based on the underlying attestation
 server and protocol.
 
+In the negotiation phase, a `NegotiationRequest` is sent from SVSM to the proxy.
+An example `NegotiationRequest` is shown below.
+```json
+{
+    "version": [0,1,0],
+    "tee": "snp",
+}
+```
+
+- `version`: The SVSM attestation protocol version to use. In place to ensure
+backwards compatibility with updated protocol versions should modifications
+occur. The API follows [SemVer](https://semver.org/) and is represented as a
+tuple [MAJOR, MINOR, PATCH]. The current version is 0.1.0.
+- `tee`: The TEE hardware architecture that SVSM is running on.
+
+The proxy will then complete the negotiation phase with the remote attestation
+server and reply with a list of negotiation parameters that must be included in
+the attestation evidence.
+
+A `NegotiationResponse` is sent from the proxy to SVSM.
+An example `NegotiationResponse` is shown below.
+```json
+{
+    "challenge": "oFlY92ZdS3ymzxokYuDzxw==\",
+    "params": [
+                "EcPublicKeyBytes",
+                "Challenge"
+              ]
+}
+```
+- `challenge`: The challenge nonce returned by the remote attestation server
+that will likely need to be hashed into the attestation evidence to ensure
+freshness. Represented as variably-sized array of bytes.
+- `params`: The negotiation parameters. Each `NegotiationParam` represents some
+form of data that must be hashed into the attestation evidence. This hash will
+be reconstructed by the remote attestation server when the evidence is presented
+from SVSM.
+
+The valid negotiation parameters are as follows:
+- `Challenge`: The bytes represented in the `challenge`.
+- `EcPublicKeyBytes`: The byte buffers of the public key's x and y coordinates
+(in that order).
+
+SVSM can then collect the attestation evidence (with the negotiation parameters
+embedded within the report data) and continue to the attestation phase.
+
 ### Attestation Phase
 
-With all relevant data embedded in TEE evidence, SVSM sends its evidence to the
-remote server for evaluation. Upon successful attestation, the proxy will obtain
-an encrypted secret (only decryptable by SVSM's TEE private key) for SVSM to
-use. For example, SVSM could use this secret to unlock encrypted persistent
+With all relevant data embedded in the TEE evidence, SVSM sends the evidence to
+the remote server for evaluation. Upon successful attestation, the proxy will
+obtain an encrypted secret (only decryptable by SVSM's attestation private key)
+for SVSM to use. For example, SVSM could use this secret to unlock encrypted
 storage.
+
+In the attestation phase, an `AttestationRequest` is sent from SVSM to the proxy.
+An example `AttestationRequest` is shown below.
+```json
+{
+    "tee": "snp",
+    "evidence":{
+        "Snp":{
+            "report": "AwAAAAAAAAAAAAMAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAEAAAADAAAAAAAY0QUAAAAAAAAAAAAAAAAAAAAHZTCHj6
+                       luB8MACrYnlqNmLuBgdclkh4UqGAMdZfMHZ+uVHytFLnvJXLkPx3xMf8p
+                       B7faGTY+Mlwi96geujAbfPKa1C+9Kt7Hts/t0Vp+TKQachyjYCZLBhTV2
+                       f7zI05r0HlwonTiV/mva7ZwxvdGaAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAACcdt0ZPBJrpXFX6RTFjSItibRTVsXM
+                       Rn/pL/wbViPO6///////////////////////////////////////////A
+                       wAAAAAAF9EZAQEAAAAAAAAAAAAAAAAAAAAAAAAAAACd3vUWsdS5AAdRkM
+                       51mXGLWEdB7w3CkS2L4/AmeQQPCKQFLcC81HJoD2st3/01IHaqOwGz3Xd
+                       Lb57uqDPWYLxpAwAAAAAAGNEdNwEAHTcBAAMAAAAAABjRAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA+InMMSLac18ai1bi
+                       1kJZISAv0MXt7fBB9do732UwrcjUaxCNxZun8fsAWKXU1LPcAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAz+KTpyaINiKXdTAYhGTMl+g05It+QVEsrZ
+                       5Vc1LHpng+KO4uFLFw0PLiXbNTwGU8AAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+                       AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=\",
+            "certs_buf": null
+        }
+    },
+    "challenge": "oFlY92ZdS3ymzxokYuDzxw==\",
+    "key":{
+        "x": "AeFSAOU2/rtNV4VS0uTr5M729jAU1RIZ+p90kRRbIac6x56y40bG39gc+oxykTITDL
+              gQER79+kVluCC+Lt6QSfaU\",
+        "y": "AcBKVeKrw0p+7Nv/JuZMn+zuuQAhpSawoRq3g6Rhc9soXzsPlVLblIEw9muXSNUWbv
+              fNrLHrmw+qy8lR1o/Kvkys\"
+    }
+}
+```
+
+- `tee`: The TEE architecture that the evidence should be interpreted as.
+- `evidence`: The attestation evidence (i.e. report) from the TEE processor.
+Based on the underlying TEE architecture (SEV-SNP being represented in the
+example).
+
+Valid evidence formats are the following:
+- `Snp`: SEV-SNP
+    - `report`: Attestation report bytes.
+    - `certs_buf`: Optional byte buffer representing SEV-SNP certificate chain
+                   (ARK, ASK, VEK) set by the host hypervisor.
+
+- `challenge`: The original challenge nonce given in the `NegotiationResponse`.
+- `key`: The EC public key that will be used to encrypt secret payloads received
+from the remote server upon a successful attestation. Public key x and y
+coordinates supplied.
+
+The proxy will forward the evidence and metadata to the remote attestation
+server for evaluation. Upon successful attestation, the proxy should be able to
+retrieve some secret payload from the remote server. The proxy will retrieve
+this secret and reply to SVSM with an `AttestationResponse`.
+An example `AttestationResponse` is shown below.
+```json
+{
+    "success": true,
+    "secret": "aQ8Kt1EqRlj54Me3\",
+    "decryption":{
+        "epk":{
+            "x": "AeFSAOU2/rtNV4VS0uTr5M729jAU1RIZ+p90kRRbIac6x56y40bG39gc+oxykT
+                  ITDLgQER79+kVluCC+Lt6QSfaU\",
+            "y": "ASrjxo2XfkkkNTiq174Uhj++eNgwzt1iA/hRrYA/6tn8i4ZgPjIiBT0EybAvn8
+                  p3JQQmw6QKM38Ck5saMZcWF/4/\"
+        },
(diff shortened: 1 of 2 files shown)


nicstange:
> I think now is the time to add documentation/specifications for the protocol and finalize a version (if this is the final version).

As a disclaimer, I haven't looked into the details of this PR yet, but just to avoid unnecessary work I'm writing it now: one thing which we very likely need from the protocol and which wasn't implemented the time I last checked is some sort of authentication for the received key material. Otherwise an adversary would be able to impersonate the KBS and make the SVSM to protect its persisted state with a secret he controls. Using authenticated encryption for the key material is definitely a step in that direction, but not sufficient. I think a full solution would be outside the scope of this PR though, and would personally leave the protocol finalization for some later PR.

tylerfanelli:
> @tylerfanelli I took a quick look, and I also see some changes to the protocol used between SVSM and the proxy. Can you explain better why they are necessary?

The changes are as follows:
- Represent each byte field as a `Vec<u8>` instead of base64-encoded `String`s. This relieves the kernel module from needing to worry about base64 {de}serialization which should instead be up to the proxy backend.
- Split the challenge nonce from the usual `NegotiationParam`s. In KBS (and likely other protocols with nonce challenges), the respective nonce needs to be sent alongside the attestation evidence for validation. This allows the kernel module to identify and cache the nonce so it is able to send it alongside the evidence at a later point.
- Standardize around ECDH-ES+A256KW for secret decryption. We previously agreed that SVSM would only support EC keys, and this is an established, secure algorithm for fetching secrets that includes authentication.

> Are they generic enough to support any attestation protocol, or Trustee-specific?

They are generic. The only real change is that we now require ECDH-ES+A256KW for secret encryption/decryption.

> I think now is the time to add documentation/specifications for the protocol and finalize a version (if this is the final version).

I think this is much closer to the final version, and only needing one change. I can add some documentation to this PR.

stefano-garzarella:
> @stefano-garzarella To test, you can use the branch found in [coconut-svsm/kbs-test#6](https://github.com/coconut-svsm/kbs-test/pull/6).

@tylerfanelli thanks! BTW, can you post a changelog when you update the PR? It will help reviewers to understand the difference between versions.

luigix25:
I tested this PR running:
```
make aproxy
bin/aproxy --protocol kbs --url http://0.0.0.0:8080 --unix /tmp/svsm-proxy-luigi.sock --force
```

for kbs

```
cargo build
export SVSM=/path/to/svsm
MEASUREMENT="$(${SVSM}/bin/igvmmeasure --check-kvm ${SVSM}/bin/coconut-qemu.igvm measure -b)"
cargo run -- -m $MEASUREMENT
```

and svsm:

```
FW_FILE=/path/to/OVMF.fd make FEATURES=attest 
./scripts/launch_guest.sh --qemu /home/lleonard/qemu/build/qemu-system-x86_64 --image /home/lleonard/repo/snp-svsm-vtpm/images/fedora-luks.qcow2 --aproxy /tmp/svsm-proxy-luigi.sock
```

Everything run smoothly. Please check this is the correct way of running it, if so what about adding it in the PR description?

tylerfanelli:
> `aproxy/scr/main.rs`:
> 
> ```
> /// Supported servers include:
> /// kbs-test: https://github.com/tylerfanelli/kbs-test (for testing).
> #[clap(long = "protocol")]
> backend: backend::Protocol,
> ```
> 
> IIUC `kbs-test` has been replaced by `kbs`, also the repo url is now different.

Thanks. I've opted to just remove that comment altogether, as I feel an explanation of supported protocols/servers belongs in the documentation instead.

kernel/src/attest.rs
```diff
@@ -57,7 +56,7 @@ impl TryFrom<Tee> for AttestationDriver<'_> {
             _ => return Err(AttestationError::UnsupportedTee.into()),
         }
 
-        let curve = Curve::new(TpmEccCurve::NistP521).map_err(AttestationError::Crypto)?;
+        let curve = Curve::new(TpmEccCurve::NistP256).map_err(AttestationError::Crypto)?;
```
nicstange: Is this change intentional? For reference, c.f. https://github.com/coconut-svsm/svsm/pull/528#discussion_r2163317299 .

kernel/src/attest.rs
```diff
@@ -120,45 +117,65 @@ impl AttestationDriver<'_> {
...
+            .decrypt_in_place_detached(
+                Nonce::from_slice(&decryption.iv),
+                &decryption.aad,
+                secret,
+                GenericArray::from_slice(&decryption.tag),
+            )
+            .map_err(AttestationError::SecretDecrypt)?;
+
+        Ok(())
     }
```
nicstange, line 181: Ok, so just to summarize and confirm I'm getting it right:
- The "secret" is protected by the "Content Encryption Key" (CEK) with AES-GCM.
- The CEK is sent by the KBS alongside the secret, protected with AES-KW.
- The Key Encryption Key (KEK) used for AES-KWing the CEK is derived from the ECDH key exchange.

Correct?

The fact that you're recoding the KDF key derivation part in a very specific way makes me wonder whether you're following some specific standard (some RFC or alike) here? If yes, a reference as a comment in the code would probably be helpful.

If not, I'd have two questions:
1. What's the AES-KW detour good for, i.e. why not simply use the derived key from ECDH directly as the CEK?
2. I would have to recheck more carefully, but it appears to me that the concat-kdf you're using now is almost equivalent to the TCG `KDFe()`, the one which is used internally by `ecdh_c_1e_1s_cdh_party_v_key_gen()` replaced by this change: the only difference is that
    - the `KDFe()` takes a `str` for the label/usage, whereas the `kdm` from above is an `[u8]`.
    - and the replaced `ecdh_c_1e_1s_cdh_party_v_key_gen()` sets `KDFe()`'s `PartyUInfo`/`PartyVInfo` to the two parties' public keys respectively.

    So in case you're not following some specific standard's rules here, I would be wondering what motivated the change to open-code the secret derivation?

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,111 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
+```
+NegotiationRequest {
+    version: "0.1.0",
+    tee: Tee::Snp,      // AMD SEV-SNP architecture.
+}
+```
+
+- `version`: The SVSM attestation protocol version to use. In place to ensure
+backwards compatibility with updated protocol versions should modifications
+occur.
```
berrange: This doesn't explain how back compatibility works, or any rules about the version number scheme. Does it make sense to store version as an opaque string, vs splitting its components ? Should we document that it is "semver" or if not, then explicit document compatibility rules for bumping each component ?

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,111 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
+An example `NegotiationResponse` is shown below.
+```
+NegotiationResponse {
+    challenge: [0, 1, 2, 3],
+    params: [
+                NegotiationParam::EcPublicKeyBytes,
+                NegotiationParam::Bytes([4, 5, 6, 7]),
+                NegotiationParam::Challenge,
+            ],
+}
```
berrange, line 67: This is describing the attestation process in terms of Rust data structures, but on the wire, IIUC, it is a JSON encoding.

If this document is intended to act as a specification to the communications protocol, would it be better to describe stuff in terms of the JSON format on the wire, so it is independent of the current rust impl choices ?

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,111 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
 storage.
 
+In the attestation phase, an `AttestationRequest` is sent from SVSM to the proxy.
+An example `AttestationRequest` is shown below.
+```
+AttestationRequest {
+    evidence: AttestationEvidence::Snp {
+        report: [0, 1, 2, 3],
+        certs_buf: None
+    },
```
berrange: Thinking ahead to a future where SVSM is ported to more than just Sev-SNP, I presume the format of the 'evidence' field here must be a match for the 'tee' valid previously supplied in the 'NegotiationRequest' struct ? Should we perhaps document that this must match ? Or should the data be self-describing, such that 'tee' is repeated on this message too.  This is where I feel it might be better to describe the protocol in terms of the wire format, as it would better illustrate that from the wire data alone, you can't tell what TEE this evidence is representing.

kernel/src/attest.rs
nicstange: > The fact that you're recoding the KDF key derivation part in a very specific way makes me wonder whether you're following some specific standard (some RFC or alike) here? If yes, a reference as a comment in the code would probably be helpful.

Figured it out myself: https://datatracker.ietf.org/doc/html/rfc7518#section-4.6.2
(In general, I think it would help readers to sprinkle some more RFC references around in the code, the JWE format is another example).

The question what the benefits of using "ECDH-ES+A256KW" over "ECDH-ES" in "Direct Key Agreement mode" are still stands -- FTR, I don't question it really, but I'm curious.

kernel/src/attest.rs
tylerfanelli: This is mainly to bring the secret decryption in line with the Trustee server, which uses P256 by default. I could explore adding P521 support to trustee if we want to keep it at that length?

tylerfanelli: @nicstange I instead opted to [add P-521 key support in Trustee](https://github.com/confidential-containers/trustee/pull/1003). I'll revert this change to use P-521.

tylerfanelli:
Changes since last push:
- Revert to EC P-521 instead of P-256 for stronger security.

kernel/src/attest.rs
tylerfanelli: You can test by pulling the [latest changes of kbs-test](https://github.com/coconut-svsm/kbs-test/pull/6).

Documentation/docs/developer/ATTESTATION.md
tylerfanelli: Yes, it would probably be better to describe in terms of JSON. Will make the change.

Documentation/docs/developer/ATTESTATION.md
tylerfanelli: > Does it make sense to store version as an opaque string, vs splitting its components ?

What is the benefit of splitting the components?

> Should we document that it is "semver" or if not, then explicit document compatibility rules for bumping each component ?

The intention is to use semver. I'll document that.

Documentation/docs/developer/ATTESTATION.md
tylerfanelli: I think it would be most clear to make the evidence self describing by adding a `tee` indicator within the JSON. I've added this and updated the documentation.

Documentation/docs/developer/ATTESTATION.md
berrange: > > Does it make sense to store version as an opaque string, vs splitting its components ?
> 
> What is the benefit of splitting the components?

It enforces the data model that we expect, as a set of three numbers, instead of allowing arbitrary strings as a version.

tylerfanelli: Sure, it can switch to that.

Documentation/docs/developer/ATTESTATION.md
tylerfanelli: See https://github.com/coconut-svsm/svsm/pull/828/commits/a8243386b8b18c275940833a6312297bc6cd5301.

Documentation/docs/developer/ATTESTATION.md
tylerfanelli: See https://github.com/coconut-svsm/svsm/commit/a8243386b8b18c275940833a6312297bc6cd5301.

kernel/src/attest.rs
tylerfanelli: From what I understand, key wrapping provides some very minor benefits such as working more nicely with key management systems and providing some stronger security. [NIST SP-800-38F](https://csrc.nist.gov/publications/detail/sp/800-38f/final) says the following:

_KW, KWP, and TKW were designed to protect the confidentiality and the authenticity/integrity of cryptographic keys. Each provides an option for protecting keys in a manner that is distinct from the methods that protect general data. Segregating keys from general data can provide an extra layer of protection_.

It's just one more level of indirection in a sense, instead of using the CEK directly, you use it to unwrap another key and use that.

tylerfanelli: Also, I've added the RFC references around in various places the keys are used.

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,151 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
+{
+    "version": [0,1,0],
+    "tee": "snp",
+}
+```
+
+- `version`: The SVSM attestation protocol version to use. In place to ensure
+backwards compatibility with updated protocol versions should modifications
+occur. The API follows [SemVer](https://semver.org/) and is represented as a
+tuple [MAJOR, MINOR, PATCH].
```
berrange: Should probably also note what the current version of the spec is here.

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,151 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
+    "challenge": [93,11,16,123,114,198,71,58,163,70,55,25,15,84,35,45],
+    "params": [
+                "EcPublicKeyBytes",
+                "Challenge"
+              ]
+}
+```
+- `challenge`: The challenge nonce returned by the remote attestation server
+that will likely need to be hashed into the attestation evidence to ensure
+freshness.
```
berrange: Please document that this is representing an array of bytes. Also mention if this is a fixed size array, or can vary .

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,151 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
+- `params`: The negotiation parameters. Each `NegotiationParam` represents some
+form of data that must be hashed into the attestation evidence. This hash will
+be reconstructed by the remote attestation server when the evidence is presented
+from SVSM.
```
berrange, line 75: Lets document what known value params currently are defined and what each means.

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,151 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
+              205,58,142,200,48,36,182,221,138,88,75,49,205,13,151,18,189,127,
+              112,217,20,38,149,19,232,114,155,125]
+    }
+}
+```
+
+- `tee`: The TEE architecture that the evidence should be interpreted as.
+- `evidence`: The attestation evidence (i.e. report) from the TEE processor.
+Based on the underlying TEE architecture (SEV-SNP being represented in the
+example).
```
berrange, line 144: We should also document the TEE specific data that's expected here. It isn't clear what format 'certs_buf' would be in particular.

The SNP report will be fairly large, and its making me wonder if we're taking the right approach by representing binary data by using an array of integers, as opposed to perhaps using a base64 encoded string. Taking the 'challenge' data - this is 16 bytes, but as an array of integers this examples is taking 54 bytes, and worst case would be 400% size penalty, compared to base64 which is approx a 30% encoding penalty.

tylerfanelli: I originally removed the encoding to avoid having the kernel worry about base64. However, I was able to move encoding for *all* byte buffers to the libaproxy crate and do encoding/decoding transparently. See https://github.com/coconut-svsm/svsm/pull/828/commits/f7d9b3b964d03b55f14a983c904ae5cd078ee073.

I've updated the documentation to reflect the new JSON objects passed with base64 encoding.

stefano-garzarella:
@tylerfanelli CI is failing, can you take a look?

tylerfanelli:
> @tylerfanelli CI is failing, can you take a look?

This is an issue with the aes_gcm crate. See [here](https://github.com/RustCrypto/traits/issues/2036).

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -36,14 +36,169 @@ is fresh and legitimate. The negotiation phase returns the parameters that SVSM
...
+
+Valid evidence formats are the following:
+- `Snp`: SEV-SNP
+    - `report`: Attestation report bytes.
+    - `certs_buf`: Optional byte buffer representing SEV-SNP certificate chain
+                   (ARK, ASK, VEK) set by the host hypervisor.
```
berrange, line 150: So just to clarify, this 'certs_buf' field would be a single string, with all the certs concatenated into the string, as opposed to an array of strings one per cert ?

tylerfanelli: Yes, that's because the actual buffer can really contain any certificates the host deems necessary. If there was a defined set of certificates we know would be in the buffer (ARK, ASK, VCEK for instance), we could parse it.

Perhaps at a later point I can add parsing for certs that are commonly included? Thinking of the chip endorsement key (VCEK, or VLEK on clouds).

berrange: I think it is fine to have one string with all the certs concatenated - i just wanted to clarify that since the example shows `certs_buf: null` and thus is slightly ambiguous what's expected for the data here.

kernel/src/attest.rs
stefano-garzarella: @nicstange can we resolve this conversation (to merge the PR)?

IIUC this is pre-existing, so maybe better to discuss in an issue and eventually provide different ways to do that to cover more scenarios.

Documentation/docs/developer/ATTESTATION.md
stefano-garzarella: @berrange can I resolve this conversation (to merge the PR), or we should do some followup?

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -107,7 +262,7 @@ SEV-SNP machine with an SVSM-enabled kernel.
     cd kbs-test
     MEASUREMENT="$(${SVSM}/bin/igvmmeasure --check-kvm ${SVSM}/bin/coconut-qemu.igvm measure -b)"
     HEX_EXPECTED_MEASUREMENT="$(echo $MEASUREMENT | xxd -p)"
-    cargo run -- --measurement $HEX_EXPECTED_MEASUREMENT
+    cargo run -- --measurement $HEX_EXPECTED_MEASUREMENT --secret $HEX_SECRET
```
stefano-garzarella:
```suggestion
    cargo run -- --measurement $MEASUREMENT --secret $HEX_SECRET
```

And we can remove the `HEX_EXPECTED_MEASUREMENT="$(echo $MEASUREMENT | xxd -p)"` line.

stefano-garzarella (changes requested):
@tylerfanelli great, LGTM, I just left a minor comment in the documentation that differs from the cover letter of this PR, indeed following it, I'm able to test this correctly.

Documentation/docs/developer/ATTESTATION.md
tylerfanelli: Yes, I've made the change.

PR #830, opened 2025-10-02
libtcgtpm: bump TPM reference implementation to v184

stefano-garzarella:
TCG released a new v184 version TPM reference implementation

Unfortunately, during the v184 release they accidentally reverted our change to disable pthread. They were quick to cherry pick it back, so now on top of v184 tag there is the commit bc29a21 ("Re-merge "Allow disabling the pthread support check"") that we can use in our `libtcgtpm`.

Update also TpmConfiguration with the upstream one; disable SIMULATION and ALLOW_FORCE_FAILURE_MODE as before. CRYPTO_LIB_REPORTING now depends on SIMULATION, so we don't need to explicitly disable it anymore. CC_Vendor_TCG_Test is also disabled to avoid build issue.

So this is the diff of our configuration and the default one available upstream:
```
    $ diff -ru libtcgtpm/deps/tpm-20-ref/TPMCmd/TpmConfiguration libtcgtpm/deps/TpmConfiguration
    diff --color -ru libtcgtpm/deps/tpm-20-ref/TPMCmd/TpmConfiguration/TpmConfiguration/TpmBuildSwitches.h libtcgtpm/deps/TpmConfiguration/TpmConfiguration/TpmBuildSwitches.h
    --- libtcgtpm/deps/tpm-20-ref/TPMCmd/TpmConfiguration/TpmConfiguration/TpmBuildSwitches.h	2025-10-02 09:33:46.352717543 +0200
    +++ libtcgtpm/deps/TpmConfiguration/TpmConfiguration/TpmBuildSwitches.h	2025-10-02 12:06:27.662628186 +0200
    @@ -49,7 +49,7 @@
     // macros will make sure that it is set correctly. A simulated TPM would include a
     // Virtual TPM. The interfaces for a Virtual TPM should be modified from the standard
     // ones in the Simulator project.
    -#define SIMULATION                  YES
    +#define SIMULATION                  NO

     // ENABLE_TPM_DEBUG_PRINT enables arbitrary string printing.
     // enables the TPM_DEBUG_PRINT macro to route debugging strings
    @@ -102,7 +102,7 @@
     // ability of the platform library to force failure mode while leaving the rest
     // of the TPM behavior alone.  Useful for testing when the full set of options
     // controlled by SIMULATION may not be desired.
    -#define ALLOW_FORCE_FAILURE_MODE    YES
    +#define ALLOW_FORCE_FAILURE_MODE    NO

     ////////////////////////////////////////////////////////////////
     // Internal checks
[...]

PR #832, opened 2025-10-06
libtcgtpm: use cmake instead of autotools to build tpm-20-ref

stefano-garzarella:
The next TCG TPM reference implementation release (v185) will remove
the support for autotools [1]. Currently, v184 supports both, so let's
switch to cmake to be prepared.
    
The main difference is that dependency on TpmBigNum and Math_Ossl was
previously specified in the autotools files when producing libtpm.a and
libplatform.a, but now it is only specified for the final product
(e.g. the simulator in their project). Therefore, we must explicitly
build and link them when producing libtcgtpm.a.
    
[1] https://github.com/TrustedComputingGroup/TPM/pull/6#issuecomment-3357499732

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -28,7 +28,7 @@ jobs:
             override: true
 
       - name: Install TPM 2.0 Reference Implementation build dependencies
-        run: sudo apt install -y autoconf autoconf-archive pkg-config build-essential automake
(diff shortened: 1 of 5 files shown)


stefano-garzarella:
Marked this draft since it depends on #830

osteffenrh (approved):

joergroedel (approved):

PR #835, opened 2025-10-14
Clippy: enable lints on stage1, verus_stub and virtio-drivers

luigix25:
stage1, verus _stub and virtio-drivers do not have the `[lints]` entry in the Cargo.toml file.
This means that running `make clippy` does not show all the warnings/error stated in the main Cargo.toml.

Add it and fix all the errors generated.

### DIFF
--- a/stage1/Cargo.toml
+++ b/stage1/Cargo.toml
@@ -15,3 +15,6 @@ load-stage2 = []
 
 [dependencies]
 bootlib.workspace = true
+
+[lints]
+workspace = true
--- a/verus_stub/Cargo.toml
+++ b/verus_stub/Cargo.toml
@@ -12,3 +12,6 @@ verus_macro_stub = { workspace = true, optional = true}
 [features]
 default = ["dep:verus_macro_stub"]
 disable = ["builtin_macros", "vstd"]
+
+[lints]
+workspace = true
--- a/virtio-drivers/Cargo.toml
+++ b/virtio-drivers/Cargo.toml
@@ -22,3 +22,6 @@ zerocopy = { workspace = true, features = ["derive"] }
 [features]
 default = ["alloc"]
 alloc = ["zerocopy/alloc"]
+
+[lints]
+workspace = true


virtio-drivers/src/device/blk.rs
```diff
@@ -267,9 +267,10 @@ impl<H: Hal, T: Transport> VirtIOBlk<H, T> {
             reserved: 0,
             sector: block_id as u64,
         };
-        let token = self
-            .queue
-            .add(&[req.as_bytes()], &mut [buf, resp.as_mut_bytes()])?;
+        let token = unsafe {
```
stefano-garzarella, line 272: New unsafe blocks should have the safety documentation.

virtio-drivers/src/device/blk.rs
```diff
@@ -267,6 +267,7 @@ impl<H: Hal, T: Transport> VirtIOBlk<H, T> {
             reserved: 0,
             sector: block_id as u64,
         };
+        // SAFETY: Safe because queue is properly initialized
```
stefano-garzarella: Sorry, I don't understand this safety comment.

The `VirtQueue::add()` docs is the following:
```
    /// Add buffers to the virtqueue, return a token.
    ///
    /// The buffers must not be empty.
    ///
    /// Ref: linux virtio_ring.c virtqueue_add
    ///
    /// # Safety
    ///
    /// The input and output buffers must remain valid and not be accessed until a call to
    /// `pop_used` with the returned token succeeds.
```

So, I don't see the relationship with the queue initialized properly, can you elaborate a bit?

virtio-drivers/src/device/blk.rs
stefano-garzarella: Okay, I see you added them in the next commit, but IMO we should avoid that. Every new unsafe block should be documented when introduced. Maybe there is a specific reason I can't see, so why you decided to split in 2 commits?

virtio-drivers/src/device/blk.rs
stefano-garzarella: Ditto, for the other comments in this file.

virtio-drivers/src/volatile.rs
```diff
@@ -39,19 +39,23 @@ pub trait VolatileReadable<T> {
 
 impl<T: FromBytes + Immutable> VolatileReadable<T> for *const ReadOnly<T> {
     unsafe fn vread_hal<H: Hal>(self) -> T {
+        // SAFETY: we trust the caller that the address is non-null, properly aligned and readable
```
stefano-garzarella: Which address ? `self` ?

virtio-drivers/src/volatile.rs
```diff
@@ -39,19 +39,23 @@ pub trait VolatileReadable<T> {
 
 impl<T: FromBytes + Immutable> VolatileReadable<T> for *const ReadOnly<T> {
```
stefano-garzarella, line 48: Should we add safety comments also in the `VolatileReadable` trait? not for `clippy` but more for us developers.

stage1/Cargo.toml
stefano-garzarella, line 1: Please move commit https://github.com/coconut-svsm/svsm/pull/835/commits/76004c6977faf3dd385b61e0f79be5ef3dc974f1 as last step. To avoid breaking the bisection, it is good practice to ensure that every commit succeeds with `make` (or `make clippy`, etc.). So first fix all the code, and then enable the lints.

Also the description is a bit lacking. These are not workspaces, but crates that are part of the main workspace, and here we are telling clippy to use the lints enabled for the workspace in these crates as well, so I would write something like this:

```
stage1,verus_stub,virtio-drivers: inherit the lints of the workspace

When these crates were introduced, we forgot to enable the lints
inherited from the workspace. Enable them for consistency.
```

virtio-drivers/src/device/blk.rs
luigix25: No specific reason, I just batched all the comments in the last commit.
I'll move the comments related to the new unsafe blocks in the same commit that adds them.

virtio-drivers/src/volatile.rs
luigix25: Yes! I modify the comment from `address` to `self` so it's clearer.

virtio-drivers/src/device/blk.rs
luigix25: IIUC `VirtQueue::add()` is unsafe because of the two unsafe blocks used.

My understanding, looking at the comments of those unsafe blocks, is that `self.avail` is guaranteed to be aligned, dereferenceable and initialized. 

This is because the `VirtQueue` struct has been initialized properly. 

This is why I wrote that comment, I hope it's clearer now

stefano-garzarella: Honestly, I'm more confused now.

This is the block change I'm commenting:
```diff
@@ -266,9 +267,11 @@ impl<H: Hal, T: Transport> VirtIOBlk<H, T> {
             reserved: 0,
             sector: block_id as u64,
         };
-        let token = self
-            .queue
-            .add(&[req.as_bytes()], &mut [buf, resp.as_mut_bytes()])?;
+        // SAFETY: Safe because queue is properly initialized
+        let token = unsafe {
+            self.queue
+                .add(&[req.as_bytes()], &mut [buf, resp.as_mut_bytes()])?
+        };
         if self.queue.should_notify() {
             self.transport.notify(QUEUE);
         }
```

I thought we need the `unsafe` block because we are calling `self.queue.add()` which is unsafe. Looking at `VirtQueue::add()` documentation, we have:
```
    /// # Safety
    ///
    /// The input and output buffers must remain valid and not be accessed until a call to
    /// `pop_used` with the returned token succeeds.
```

So, I still don't understand the safety comment we are adding. I think here we should ensure that the buffers are valid and untouched until we call `pop_used`, no?

virtio-drivers/src/volatile.rs
stefano-garzarella: BTW rather than “we trust the caller,” we are delegating to the caller to respect this,

virtio-drivers/src/device/blk.rs
luigix25: Yes, you are right. I was just considering the `unsafe` blocks in the `add` function.

virtio-drivers/src/volatile.rs
```diff
@@ -39,24 +43,32 @@ pub trait VolatileReadable<T> {
...
+        // SAFETY: we delegate the caller that self is non-null, properly aligned and readable
+        unsafe { self.read_volatile().0 }
     }
 }
 
 /// A trait implemented by MMIO registers which may be written to.
+///
+/// # Safety
+///
+/// Caller is responsibile of passing a non-null, aligned and writable address
```
stefano-garzarella: This should go on top of the `vwrite_hal` and `vwrite` functions IMO, which are unsafe, but again `address` is not really clear here.

virtio-drivers/src/volatile.rs
```diff
@@ -33,53 +33,81 @@ impl<T: FromBytes + IntoBytes + Immutable> Volatile<T> {
...
     }
 }
 
 /// A trait implemented by MMIO registers which may be written to.
 pub trait VolatileWritable<T> {
     /// Performs a volatile write to the MMIO register.
+    ///
+    /// # Safety
+    ///
+    /// Caller is responsibile of passing a non-null, aligned and writable `self`
```
stefano-garzarella:
```suggestion
    /// Caller is responsible of passing a non-null, aligned and writable `self`
```

Also in other place in ths file.

virtio-drivers/src/device/blk.rs
```diff
@@ -288,8 +292,12 @@ impl<H: Hal, T: Transport> VirtIOBlk<H, T> {
...
         resp: &mut BlkResp,
     ) -> Result<()> {
-        self.queue
-            .pop_used(token, &[req.as_bytes()], &mut [buf, resp.as_mut_bytes()])?;
+        // SAFETY: The caller is responsible for ensuring that `req` and `resp` are the same ones passed
+        // to `read_blocks_nb` when it returned the token.
+        unsafe {
+            self.queue
+                .pop_used(token, &[req.as_bytes()], &mut [buf, resp.as_mut_bytes()])?;
+        };
```
stefano-garzarella:
```suggestion
        }
```

Also in other place, it's redundant

luigix25: It would be nice to have in `cargo fmt` such option, not sure it exists.

luigix25:
- Fixed some typos
- Removed unnecessary semicolons

virtio-drivers/src/device/blk.rs
stefano-garzarella: This should make `clippy` unhappy:

```
index 9a1248ea..8777e635 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,6 +133,7 @@ string_to_string = "warn"
 suboptimal_flops = "warn"
 undocumented_unsafe_blocks = "warn"
 unnecessary_box_returns = "warn"
+unnecessary_semicolon = "warn"
```

Maybe there is something else we can enable in `rustfmt.toml` to have `cargo fmt` to check that, but I'm not sure it's a `cargo fmt` work.

luigix25: that should work too! Should I pick it up in this PR?

stefano-garzarella: > that should work too! Should I pick it up in this PR?

nope, if we want to enable it, it should be in another PR IMO.

PR #844, opened 2025-10-29
libtcgtpm: Makefile fixes

osteffenrh:
Some fixes for the Makefile of the TPM library, plus
some convenience targets for easier debugging.

### DIFF
--- a/libtcgtpm/Makefile
+++ b/libtcgtpm/Makefile
@@ -197,3 +197,10 @@ distclean: clean
 	rm -rf $(TCGTPM_BUILD_DIR)
 
 .PHONY: all clean distclean
+
+.PHONY: libcrt configure_openssl configure_tcgtpm libcrypto tcgtpm
+libcrt: $(LIBCRT)
+configure_openssl: $(OPENSSL_MAKEFILE)
+libcrypto: $(LIBCRYPTO)
+configure_tcgtpm: $(TCGTPM_MAKEFILE)
+tcgtpm: $(OUT_DIR)/libtcgtpm.a


libtcgtpm/Makefile
```diff
@@ -35,6 +35,8 @@ LIBS = $(LIBCRT) $(LIBCRYPTO) $(LIBBNMATH) $(LIBTPMBIGNUM) $(LIBTPM) $(LIBPLATFO
 
 OUT_DIR ?= $(CWD)
 
+.NOTPARALLEL:
```
stefano-garzarella, line 41: Can we add a comment here like we have in the commit description?

libtcgtpm/Makefile
```diff
@@ -45,6 +45,13 @@ $(OUT_DIR)/libtcgtpm.a: $(LIBS)
 
 .PHONY: $(LIBCRT)
 
+.PHONY: libcrt configure_openssl configure_tcgtpm libcrypto tcgtpm
+libcrt: $(LIBCRT)
+configure_openssl: $(OPENSSL_MAKEFILE)
+libcrypto: $(LIBCRYPTO)
```
stefano-garzarella: what about `configure_libcrypto` ?
or `openssl` instead of `libcrypto`, I mean I'd use the same for both, but not a strong opinion.

osteffenrh: or rather rename libcrypto to openssl in the whole file?
Most things use openssl in the name, for example OPENSSL_MAKEFILE etc...

stefano-garzarella: yeah, +1 for s/libcrypto/openssl

osteffenrh: maybe better leave it as is. we configure and build the whole openssl, but we only consume libcrypto.a. That's why it looks a bit confusing. But renaming would not make it much better, IMO.

stefano-garzarella: I see, agree!

libtcgtpm/Makefile
```diff
@@ -48,6 +48,13 @@ $(OUT_DIR)/libtcgtpm.a: $(LIBS)
 
 .PHONY: $(LIBCRT)
 
+.PHONY: libcrt configure_openssl configure_tcgtpm libcrypto tcgtpm
```
stefano-garzarella: Sorry, I noticed this only now, can we avoid putting this new block between `.PHONY: $(LIBCRT)` and `$(LIBCRT):` ?

You can merge the .PHONY targets or add the entire new hunk before the one already there, as you prefer.

libtcgtpm/log_new
stefano-garzarella, line 1: mmm, new file added by mistake?

stefano-garzarella: I guess the entire last commit is in by mistake, right?

osteffenrh: yeah.

luigix25 (approved):

stefano-garzarella (approved):

Issue #838, opened 2025-10-23
libtcgtpm uses pthread


peterfang:
I'm running into this issue after pulling https://github.com/coconut-svsm/svsm/pull/832:

```
  (cd /home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/tpm-20-ref/TPMCmd/build && \
        cmake \
                -DCMAKE_C_FLAGS="-g -O0 -DDEBUG=YES -static -nostdinc -fno-stack-protector -fPIE -mno-sse -mno-red-zone -DFILE_BACKED_NV=NO -I/home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/li
bcrt/include -I/home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/openssl/include" \
                -DOPENSSL_CRYPTO_LIBRARY="/home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/openssl/libcrypto.a" \
                -DOPENSSL_INCLUDE_DIR="/home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/openssl/include" \
                -Duser_TpmConfiguration_Dir="/home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/TpmConfiguration" \
                ..)
  -- The C compiler identification is GNU 11.4.0
  -- The CXX compiler identification is GNU 11.4.0
  -- Detecting C compiler ABI info
  -- Detecting C compiler ABI info - done
  -- Check for working C compiler: /usr/bin/cc - skipped
  -- Detecting C compile features
  -- Detecting C compile features - done
  -- Detecting CXX compiler ABI info
  -- Detecting CXX compiler ABI info - done
  -- Check for working CXX compiler: /usr/bin/c++ - skipped
  -- Detecting CXX compile features
  -- Detecting CXX compile features - done
  -- Looking for pthread.h
  -- Looking for pthread.h - not found
  -- Could NOT find Threads (missing: Threads_FOUND)
  -- Found OpenSSL: /home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/openssl/libcrypto.a (found version "3.5.0")
  -- Could NOT find Threads (missing: Threads_FOUND)
  -- Found OpenSSL: /home/pfang/local/svsm/coconut-svsm/libtcgtpm/deps/openssl/libcrypto.a (found version "3.5.0") found components: Crypto
  -- Configuring done
  -- Generating done

.....

[...]

PR #845, opened 2025-10-30
libtcgtpm: Use embedded cmake build

osteffenrh:
The `FindOpenSSL.cmake` script pulls in`FindThreads.cmake` if
building for Linux (possibly to satisfy OpenSSL's dependencies).
We do not use threads, so this is not required.

Set `CMAKE_SYSTEM_NAME` to "Generic", indicating an embedded style build,
so that FindThreads is not called.

See also: <https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html>

The difference can be observed by running cmake with debug output by adding

```
--debug-output --trace-source=FindThreads.cmake
```

to the cmake call in the Makefile.


osteffenrh:
see https://github.com/coconut-svsm/svsm/issues/838

stefano-garzarella:
@osteffenrh great, this works also with the buggy `cmake` (3.28.3) shipped by Ubuntu 22.04, so can you add `Closes #838` in the commit message?

PR #846, opened 2025-11-04
github/workflows: Add qemu.yml SHA to QEMU cache key

luigix25:
When the QEMU config changes a new build is required. Currently the cache key for the QEMU build does not include any reference to the configuration.

Add it, so any change to qemu.yml will trigger a new build.

### DIFF
--- a/.github/workflows/qemu.yml
+++ b/.github/workflows/qemu.yml
@@ -32,12 +32,17 @@ jobs:
           submodules: true
           fetch-depth: 1
 
-      - name: Get QEMU form cache
+      - name: Calculate SHA for qemu.yml
+        run: |
+          QEMU_YML_SHA=$(shasum .github/workflows/qemu.yml | awk '{print $1}')
+          echo "QEMU_YML_SHA=${QEMU_YML_SHA}" >> "$GITHUB_ENV"
+
+      - name: Get QEMU from cache
         id: qemu_cache
         uses: actions/cache@v4
         with:
           path: tools
-          key: qemu-${{ env.QEMU_SHA }}--rust-${{ env.RUST_VERSION_IGVM }}--igvm-${{ env.IGVM_REF }}
+          key: qemu-${{ env.QEMU_SHA }}--config-${{ env.QEMU_YML_SHA }}
 
       - if: ${{ steps.qemu_cache.outputs.cache-hit != 'true' }}
         name: Enable Ubuntu source repositories


.github/workflows/qemu.yml
```diff
@@ -32,12 +32,17 @@ jobs:
...
+          CONFIG_SHA=$(shasum .github/workflows/qemu.yml | awk '{print $1}')
+          echo "CONFIG_SHA=${CONFIG_SHA}" >> "$GITHUB_ENV"
+
+      - name: Get QEMU from cache
         id: qemu_cache
         uses: actions/cache@v4
         with:
           path: tools
-          key: qemu-${{ env.QEMU_SHA }}--rust-${{ env.RUST_VERSION_IGVM }}--igvm-${{ env.IGVM_REF }}
+          key: qemu-${{ env.QEMU_SHA }}--rust-${{ env.RUST_VERSION_IGVM }}--igvm-${{ env.IGVM_REF }}--config-${{ env.CONFIG_SHA }}
```
stefano-garzarella: Since `RUST_VERSION_IGVM` and `IGVM_REF` are defined in this file (`.github/workflows/qemu.yml`), the new CONFIG_SHA should already catch any changes of that values, so can we remove them from the key?

luigix25: yes, we can remove them from the key.

stefano-garzarella: okay, in that case, maybe we can rename the var to something like QEMU_WORKFLOW_SHA or QEMU_YML_SHA to make it clear that it covers all changes in the workflow

.github/workflows/qemu.yml
```diff
@@ -21,23 +21,23 @@ jobs:
...
+          QEMU_YML_SHA=$(shasum .github/workflows/qemu.yml | awk '{print $1}')
+          echo "QEMU_YML_SHA=${QEMU_YML_SHA}" >> "$GITHUB_ENV"
+
+      - name: Get QEMU from cache
         id: qemu_cache
         uses: actions/cache@v4
         with:
           path: tools
-          key: qemu-${{ env.QEMU_SHA }}--rust-${{ env.RUST_VERSION_IGVM }}--igvm-${{ env.IGVM_REF }}
+          key: qemu--config-${{ env.QEMU_YML_SHA }}
```
stefano-garzarella: mmm, why you removed QEMU_SHA ?
IIUC it is needed to check if we pushed new commit in our fork.

.github/workflows/qemu.yml
```diff
@@ -21,23 +21,23 @@ jobs:
...
-          echo "QEMU_SHA=${QEMU_SHA}" >> "$GITHUB_ENV"
-
       - name: Checkout
         uses: actions/checkout@v3
         with:
           submodules: true
           fetch-depth: 1
 
-      - name: Get QEMU form cache
+      - name: Calculate SHA for QEMU configuration
```
stefano-garzarella: I'd also update this

.github/workflows/qemu.yml
luigix25: oops, yes you are right!

stefano-garzarella:
LGTM, but I'd like an opinion from @osteffenrh that introduced the qemu cache.

PR #850, opened 2025-11-05
apic: Add ICR read for APIC

MelodyHuibo:
During the guest boot, the ICR register is accessed. Handle this read in the SVSM properly in order to allow the guest to continue booting successfully instead of halting midway indefinitely.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -104,6 +104,7 @@ pub struct ApicIcr {
 #[expect(missing_copy_implementations)]
 #[derive(Default, Debug)]
 pub struct LocalApic {
+    icr: ApicIcr,
     irr: [u32; 8],
     allowed_irr: [u32; 8],
     isr_stack_index: usize,
@@ -120,6 +121,7 @@ pub struct LocalApic {
 impl LocalApic {
     pub const fn new() -> Self {
         Self {
+            icr: ApicIcr::new(),
             irr: [0; 8],
             allowed_irr: [0; 8],
             isr_stack_index: 0,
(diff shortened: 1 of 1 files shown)


kernel/src/cpu/apic.rs
```diff
@@ -615,9 +618,16 @@ impl LocalApic {
         }
     }
 
+    fn handle_icr_read(&self) -> u64 {
+        let icr: u64 = self.icr.into();
+        icr
```
msft-jlange:
```suggestion
        self.icr.into()
```
No reason to have extra lines to do the same thing.

kernel/src/cpu/apic.rs
```diff
@@ -615,9 +618,16 @@ impl LocalApic {
...
+    }
+
     fn handle_icr_write(&mut self, value: u64) -> Result<(), SvsmError> {
         let icr = ApicIcr::from(value);
 
+        self.icr = icr;
```
00xc: Perhaps we want to do this **after** checking that the ICR delivery mode is valid below? Otherwise I would add a comment explicitly stating why we want to keep an invalid value in `self.icr` for a later read. I might have missed the discussion, but I think the new SVSM spec draft does not specify if a write with an unsupported delivery mode should go through (I would guess it does not).

msft-jlange: I strongly agree with @00xc's comment - I should have made the same point myself.

kernel/src/cpu/apic.rs
MelodyHuibo: ok

kernel/src/cpu/apic.rs
MelodyHuibo: That is a good point, I agree. I will do that.

joergroedel (approved):

PR #853, opened 2025-11-06
libtcgtpm: build everything in OUT_DIR and use CARGO_MAKEFLAGS or NUM_JOBS

stefano-garzarella:
This PR should fix the annoying issue we have with `libtcgtpm`, as we were compiling several things in-tree.

Let's make better use of Cargo `OUT_DIR` to build everything under `target/...`.

While we're at it, let's improve the `clean` target and make better use of other Cargo parameters for parallelism, such as `NUM_JOBS` and `CARGO_MAKEFLAGS`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-*.o
-*.bin
-*.elf
 *.swp
 *~
 *.igvm
--- a/libtcgtpm/deps/libcrt/.gitignore
+++ b/libtcgtpm/deps/libcrt/.gitignore
@@ -1,1 +1,0 @@
-libcrt.a


stefano-garzarella:
@osteffenrh oh, I see, what about something like this:
```diff
diff --git a/libtcgtpm/Makefile b/libtcgtpm/Makefile
index 5c7e1a43..2d3e0034 100644
--- a/libtcgtpm/Makefile
+++ b/libtcgtpm/Makefile
@@ -24,6 +24,7 @@ TCGTPM_SRC_DIR = $(DEPS_DIR)/tpm-20-ref/TPMCmd
 
 ROOT_DIR := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
 OUT_DIR ?= $(ROOT_DIR)
+override OUT_DIR := $(abspath $(OUT_DIR))
 
 LIBCRT_BUILD_DIR = $(OUT_DIR)/libcrt-build
 OPENSSL_BUILD_DIR = $(OUT_DIR)/openssl-build
```

osteffenrh:
> @osteffenrh oh, I see, what about something like this:
> 
> ```diff
> diff --git a/libtcgtpm/Makefile b/libtcgtpm/Makefile
> index 5c7e1a43..2d3e0034 100644
> --- a/libtcgtpm/Makefile
> +++ b/libtcgtpm/Makefile
> @@ -24,6 +24,7 @@ TCGTPM_SRC_DIR = $(DEPS_DIR)/tpm-20-ref/TPMCmd
>  
>  ROOT_DIR := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
>  OUT_DIR ?= $(ROOT_DIR)
> +override OUT_DIR := $(abspath $(OUT_DIR))
>  
>  LIBCRT_BUILD_DIR = $(OUT_DIR)/libcrt-build
>  OPENSSL_BUILD_DIR = $(OUT_DIR)/openssl-build
> ```

That should do it.

osteffenrh (approved):

.gitignore
```diff
@@ -1,6 +1,3 @@
-*.o
-*.bin
-*.elf
 *.swp
 *~
 *.igvm
```
msft-jlange, line 3: I think we can remove `*.igvm` as well (all IGVM files are in `bin` which is already excluded) but that can happen in a separate PR.

msft-jlange:
The recursive clean was the missing piece.  With that, I can build successfully, including after executing `make clean` and building again.  I think this change is great!

.gitignore
stefano-garzarella: good point! I'll do that in the follow-up PR where I'll move all the binaries under `/bin/` so remove more lines

joergroedel (approved):

PR #855, opened 2025-11-09
apic: Clear flags in the doorbell page after all information is in IRR

MelodyHuibo:
When the SVSM is consuming the information from the doorbell page and before it is saved in its IRR, a VMEXIT can happen.

The hypervisor would raise a #HV leading to SVSM executing its #HV handler when another interrupt delivery is attempted.

This would lead to losing the consumed information before the SVSM has managed to save the interrupt into its IRR, which will cause an interrupt to get lost eventually.

To avoid this, clear all of flags only after interrupt information is added into the IRR.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -13,6 +13,7 @@ use crate::error::SvsmError;
 use crate::mm::GuestPtr;
 use crate::platform::guest_cpu::GuestCpuState;
 use crate::requests::SvsmCaa;
+use crate::sev::hv_doorbell::HVDoorbellFlags;
 use crate::sev::hv_doorbell::HVExtIntStatus;
 use crate::types::GUEST_VMPL;
 
@@ -698,7 +699,7 @@ impl LocalApic {
 
     fn consume_host_interrupts(&mut self) {
         let hv_doorbell = this_cpu().hv_doorbell().unwrap();
-        let vmpl_event_mask = hv_doorbell.per_vmpl_events.swap(0, Ordering::Relaxed);
+        let vmpl_event_mask = hv_doorbell.per_vmpl_events.load(Ordering::Acquire);
         // Ignore events other than for the guest VMPL.
         if vmpl_event_mask & (1 << (GUEST_VMPL - 1)) == 0 {
             return;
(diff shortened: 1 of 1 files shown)


msft-jlange:
I don't understand the problem, so I cannot understand why this fix is proposed.

The SVSM does not rely on #HV for correct processing of guest interrupts; it only relies on #HV as a mechanism to get control when the host presents an interrupt at the time the guest is executing.  The #HV in this case does not have an associated SVSM interrupt associated with it, so when the SVSM's #HV handler runs, it will observe that it has nothing to do, and will clear `NoFurtherSignal`.  The SVSM itself will loop correctly when `NoFurtherSignal` is cleared so that there is no possibility of an interrupt to the SVSM itself being lost.

Once the SVSM gets control (whether for #HV or for any other reason), it will invoke the APIC emulation code to scan the guest interrupt information in the doorbell page to decide whether a guest interrupt is pending.  It does this with interrupts disabled, so any #HV delivery at this point will not affect the flow (it will only prevent the guest from being reentered, and will instead cause the SVSM to run again instead of entering the guest).  When the APIC emulation code scans the doorbell page, the first thing it does is to clear the per-VMPL events flag - this ensures that if another guest interrupt is raised after this - whether before or after the APIC emulation code scans the IRR - then the host will generate another #HV, and thus the host will not miss any notification that there is further processing to perform.  Note that clearing the per-VMPL events flag is atomic with respect to scanning it, so there is no possibility that anything can be missed.

The process of scanning the IRR is complex and involves different paths depending on whether multiple vectors are present, but everything begins with first performing an atomic swap to clear the flag state in the doorbell page before any processing begins.  This ensures that if the host proceeds to present another interrupt, it will perform full injection processing (first setting the single vector and/or the multiple vectors flag, then setting the per-VMPL flag, then scheduling #HV if necessary).  The APIC emulation code does not loop until the multiple vector bitmap is empty, but it doesn't have to, because clearing the flags means that any further attempt by the host to present an interrupt will raise another #HV, causing the SVSM to process the IRR on the next attempt to reenter the guest.

Based on this, I cannot see any window in which any interrupt can be lost.  If you do, please describe the exact situation that concerns you, including both the point in the SVSM flow where the host can inject the interrupt and how you think it is lost at that point.

PR #858, opened 2025-11-12
QEMU: Get MADT via IVGM parameter

osteffenrh:
Get the MADT from IGVM parameters supplied by QEMU.
This requires these QEMU changes: https://github.com/coconut-svsm/qemu/pull/27
Then we can remove the separate code-paths for QEMU that read the MADT via fw_cfg
and do what all other supported hypervisors do.

### DIFF
--- a/kernel/src/igvm_params.rs
+++ b/kernel/src/igvm_params.rs
@@ -36,7 +36,7 @@ pub struct IgvmParams<'a> {
     igvm_param_block: &'a IgvmParamBlock,
     igvm_param_page: &'a IgvmParamPage,
     igvm_memory_map: &'a IgvmMemoryMap,
-    igvm_madt: Option<&'a [u8]>,
+    igvm_madt: &'a [u8],
     igvm_guest_context: Option<&'a IgvmGuestContext>,
 }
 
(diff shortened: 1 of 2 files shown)


kernel/src/igvm_params.rs
```diff
@@ -273,10 +273,14 @@ impl IgvmParams<'_> {
     pub fn load_cpu_info(&self) -> Result<Option<Vec<ACPICPUInfo>>, SvsmError> {
         match self.igvm_madt {
             Some(madt_data) => {
```
luigix25: Should we mark madt as always present and drop `Some`?

luigix25 (commented):
Maybe we can also drop fw_cfg support for madt?

kernel/src/igvm_params.rs
osteffenrh: I would like to. Question is, if it is better to leave it around for a moment until everyone has updated their qemu?

luigix25: mmh, IIRC if you have the MADT directive in the IGVM file, and QEMU does not support it, it will stop booting.

osteffenrh: Yeah. I guess there is no need to keep the fw_cfg path around.

igvmbuilder/src/gpa_map.rs
```diff
@@ -147,10 +147,7 @@ impl GpaMap {
 
         let igvm_param_block = GpaRange::new_page(kernel_fs.get_end())?;
         let general_params = GpaRange::new_page(igvm_param_block.get_end())?;
-        let madt_size = match options.hypervisor {
-            Hypervisor::HyperV | Hypervisor::Vanadium => PAGE_SIZE_4K,
-            Hypervisor::Qemu => 0,
-        };
+        let madt_size = PAGE_SIZE_4K;
```
luigix25: I think we can remove this variable

joergroedel (dismissed):

joergroedel:
Approved. We need to send a note to the mailing list when this is merged to let everyone know they need to update their QEMU.

luigix25 (commented):
Some more cleaning can be done in `igvm_params.rs`

```        
let madt = if param_block.madt_size != 0 {
```

because now `madt_size` is always 4K.

Same for `igvm_builder.rs`
```
if self.gpa_map.madt.get_size() != 0 {
```

Thanks for the great work!

igvmbuilder/src/gpa_map.rs
osteffenrh: Done.

luigix25 (approved):
tested and works as expected!

peterfang (approved):
Great work! Thanks!

PR #860, opened 2025-11-14
kernel: refactor init functions for safety

msft-jlange:
Move the bulk of initialization logic into separate functions instead of invoking the scheduler or idle loop as callouts from initialization logic.  This ensures that all data used by core initialization logic is limited in scope and cannot live across scheduling boundaries.  This will make it easier in the future to control lifetimes of initialization data.

### DIFF
--- a/kernel/src/cpu/shadow_stack.rs
+++ b/kernel/src/cpu/shadow_stack.rs
@@ -71,10 +71,8 @@ pub fn shadow_stack_info() {
 /// to set up the shadow stack to return from this code.
 #[macro_export]
 macro_rules! enable_shadow_stacks {
-    ($bsp_percpu:ident) => {{
+    ($token_addr:ident) => {{
         use core::arch::asm;
-
-        let token_addr = $bsp_percpu.get_top_of_shadow_stack().unwrap();
 
         // SAFETY: This assembly enables shadow-stacks and does not impact Rust
         // memory safety.
@@ -89,8 +87,8 @@ macro_rules! enable_shadow_stacks {
                 in("ecx") S_CET,
                 in("edx") 0,
                 in("eax") SCetFlags::SH_STK_EN.bits() | SCetFlags::WR_SHSTK_EN.bits(),
-                token_addr = in(reg) token_addr.bits(),
-                token_val = in(reg) token_addr.bits() + 8 + MODE_64BIT,
+                token_addr = in(reg) $token_addr.bits(),
+                token_val = in(reg) $token_addr.bits() + 8 + MODE_64BIT,
                 options(nostack, readonly),
             );
         }
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -102,9 +102,9 @@ global_asm!(
 
         /*
          * Make sure (%rsp + 8) is 16b-aligned when control is transferred
-         * to svsm_start as required by the C calling convention for x86-64.
+         * to svsm_entry as required by the C calling convention for x86-64.
          */
-        call svsm_start
+        call svsm_entry
         int3
 
         .bss
@@ -169,8 +169,7 @@ fn init_cpuid_table(addr: VirtAddr) {
     register_cpuid_table(&CPUID_PAGE);
 }
 
-#[no_mangle]
-extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: usize) -> ! {
+fn svsm_start(li: &KernelLaunchInfo, vb_addr: usize) -> Option<VirtAddr> {
     let launch_info: KernelLaunchInfo = *li;
     init_platform_type(launch_info.platform_type);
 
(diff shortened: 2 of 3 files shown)


kernel/src/cpu/shadow_stack.rs
```diff
@@ -44,10 +49,16 @@ pub fn is_cet_ss_supported() -> bool {
...
+    IS_CET_ENABLED.load(Ordering::Relaxed)
+}
+
 pub fn shadow_stack_info() {
     log::info!(
         "Kernel shadow stacks {}",
-        match is_cet_ss_supported() {
+        match is_cet_ss_enabled() {
             true => "enabled",
             false => "not supported",
```
stefano-garzarella, line 63: Should we update this log message?

I mean something like this:
```suggestion
            false => "not enabled",
```

kernel/src/svsm.rs
```diff
@@ -305,9 +317,7 @@ extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: usize) -> ! {
     unreachable!("SVSM entry point terminated unexpectedly");
 }
 
-pub fn svsm_main(cpu_index: usize) {
-    debug_assert_eq!(cpu_index, 0);
-
```
stefano-garzarella, line 310: Is removing the assert intentional? In that case, should we mention the reason in the commit description? (asking because it's not really clear to me the reason, but maybe I'm missing something)

kernel/src/svsm.rs
```diff
@@ -393,8 +403,11 @@ pub fn svsm_main(cpu_index: usize) {
         )
         .expect("Failed to launch request loop task");
     }
+}
 
-    cpu_idle_loop(cpu_index);
+pub fn svsm_main(_context: usize) {
+    svsm_init();
+    cpu_idle_loop(0);
```
stefano-garzarella, line 411: Ditto, is using always 0 here intentional ? (I mean, with the assert is clear that it can't be !=0)

kernel/src/cpu/shadow_stack.rs
msft-jlange: This output was always a little confusing, because the only choices were "enabled" and "not supported" (nothing about "not supported").  I would prefer not to dive into the matrix of possible combinations in this PR; we can make more sense of this output (which is really just for debugging) in a future PR.

kernel/src/svsm.rs
msft-jlange: There is no longer anything to assert.  The launch parameter to `svsm_main()` is now unused.  I would have thought that would be obvious because of the underscore in the name.  Plus, the name change from `cpu_index` to `context` should suggest that this no longer represents a CPU index and therefore it has no meaning.

kernel/src/svsm.rs
msft-jlange: I've added a comment which hopefully makes this more clear.

stefano-garzarella: Thanks for that. I mean, it was already clear in the code, what is not clear to me, is why we need to do that in this PR.

IIUC the caller is already doing something like that (using `svsm_main` in  `setup_idle_task()` only for the bsp), so IMHO was better the old behavior, because the assert would allow us to prevent to use `svsm_main` as idle task for other cpu, but maybe I'm missing something.

msft-jlange: I have a future change in which `svsm_main()` needs initialization parameters that are unrelated to the CPU index, so `svsm_main()` will have to have a different calling convention than `cpu_idle_loop()`.  That change is primarily about memory management so I wanted to keep it in a separate PR from this one.

stefano-garzarella: okay, I see now. In that case I would probably have changed everything in the patch where we change the calling convention, because that would make the reason clear.

That said, now I understand the point, thanks :-)

msft-jlange: Well, the real reason wasn't the change in calling convention but the upcoming change to pass unique parameters to the init function.  I hadn't thought of this change at the time I changed the calling convention.  And I prefer to keep PRs small when possible which is why I'm not including this change in the much bigger PR that overhauls init parameters.

joergroedel (approved):

kernel/src/svsm.rs
```diff
@@ -296,6 +291,23 @@ extern "C" fn svsm_start(li: &KernelLaunchInfo, vb_addr: usize) -> ! {
 
     sse_init();
 
+    bsp_percpu
+        .get_top_of_shadow_stack()
+        .unwrap_or(VirtAddr::from(0u64))
```
peterfang: Can we just return `bsp_percpu.get_top_of_shadow_stack()`? I feel that's a bit clearer than defaulting it to 0. `allocate_init_shadow_stack()` is already gated by `is_cet_ss_supported()`.

msft-jlange: I've made that change and the corresponding change to the caller of `svsm_start()`.  Hopefully this looks better to you, because I'm not sure there are further changes that could improve it.

peterfang (approved):

PR #861, opened 2025-11-14
kernel: refactor request processing loop

msft-jlange:
Start guest processing using a dedicated startup routine rather than having the per-CPU start routine (per-thread) also be the initial task start routine.

### DIFF
--- a/kernel/src/requests.rs
+++ b/kernel/src/requests.rs
@@ -87,23 +87,27 @@ fn request_loop_once(
     }
 }
 
-pub fn request_loop_main(cpu_index: usize) {
-    log::info!("Launching request-processing task on CPU {}", cpu_index);
+pub fn request_loop_start(_: usize) {
+    // This should always be started on the BSP.
+    debug_assert_eq!(this_cpu().get_cpu_index(), 0);
 
-    if cpu_index != 0 {
-        // Send this task to the correct CPU.
-        set_affinity(cpu_index);
-    } else {
-        // When starting the request loop on the BSP, start an additional
-        // request loop task for each other processor in the system.
-        let cpu_count = PERCPU_AREAS.len();
-        for task_index in 1..cpu_count {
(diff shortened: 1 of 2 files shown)


kernel/src/requests.rs
```diff
@@ -87,23 +87,24 @@ fn request_loop_once(
...
+        start_kernel_thread(KernelThreadStartInfo::new(request_loop_main, task_index))
+            .expect("Failed to launch request loop thread");
     }
 
-    debug_assert_eq!(cpu_index, this_cpu().get_cpu_index());
+    // Enter the main processing loop for the BSP.
+    request_loop_main(0);
+}
+
+pub fn request_loop_main(cpu_index: usize) {
```
peterfang: Now that `request_loop_start()` is the start routine, can we make `request_loop_main()` a private function?

kernel/src/requests.rs
```diff
@@ -87,23 +87,24 @@ fn request_loop_once(
...
+pub fn request_loop_start(_: usize) {
+    // Start an additional request loop task for each other processor in the
+    // system.
+    let cpu_count = PERCPU_AREAS.len();
+    for task_index in 1..cpu_count {
+        start_kernel_thread(KernelThreadStartInfo::new(request_loop_main, task_index))
+            .expect("Failed to launch request loop thread");
     }
 
-    debug_assert_eq!(cpu_index, this_cpu().get_cpu_index());
```
peterfang, line 106: How about we keep this `debug_assert_eq!()` to ensure the thread is in an expected state?

joergroedel (approved):
Approved with the changes request by @peterfang.

Testing passed on my side, including `test-in-svsm`

kernel/src/requests.rs
msft-jlange: I'm bringing back the assert to verify that the request task is started on the BSP (processor 0).  All of the other set their own affinity so they will be known to be correct, but the first thread in the task would benefit from an assert as you say.

PR #865, opened 2025-11-18
cpu/vc: fix tests if `TSC_AUX` MSR is not intercepted (+ small `#VC` handler fix)

00xc:
* Fix a copy + paste mistake in the `#VC` handler causing incorrect handling of `RDTSCP`.
* Fix `#VC` tests if `TSC_AUX` is not intercepted, which causes access to the MSR to not generate a `#VC`, which in turn results in the GHCB not being modified. Thus, do not use `verify_ghcb_gets_altered()` in that case.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -329,6 +329,7 @@ fn vc_decoding_needed(error_code: usize) -> bool {
 mod tests {
     use super::*;
     use crate::cpu::msr::{rdtsc, rdtscp, read_msr, write_msr, RdtscpOut};
+    use crate::locking::SpinLock;
     use crate::sev::ghcb::GHCB;
     use crate::sev::utils::{get_dr7, raw_vmmcall, set_dr7};
     use bootlib::platform::SvsmPlatformType;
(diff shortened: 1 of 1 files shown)


kernel/src/cpu/vc.rs
```diff
@@ -661,15 +664,49 @@ mod tests {
...
+        // Read cached value or initialize it by probing the MSR
+        *TSC_AUX_INTERCEPTED.lock().get_or_insert_with(|| {
+            fill_ghcb_with_test_data();
+            let _ = read_msr(MSR_TSC_AUX);
+            verify_ghcb_was_altered()
+        })
+    }
+
+    fn test_read_tsc_aux() -> u64 {
+        if tsc_aux_intercepted() {
```
msft-jlange: I don't really understand the point of this test after your change.  First, you read the MSR and use `verify_ghcb_was_altered()` to determine whether the MSR is intercepted, and if so, you read the MSR again and execute `verify_ghcb_was_altered()` a second time.  So really, all you are testing is whether the intercept, if it occurs, occurs twice in a row.  We don't apply that philosophy to any other #VC testing.  If we're willing to relax the test to permit the RDMSR not to raise #VC, then why bother to execute the RDMSR a second time?

00xc: > If we're willing to relax the test to permit the RDMSR not to raise #VC, then why bother to execute the RDMSR a second time?

We still test that the MSR does reflect the value that RDTSC / RDSTCP sees. I guess we could alternatively check at the beginning of the test and simply not go forward if we detect that the MSR is not intercepted.

msft-jlange: I think it would be more clear to do what you say: test for TSC_AUX intercept at the beginning of `test_wrmsr_tsc_aux()` and only move forward with the read/write tests if the MSR is intercepted.  Otherwise, I worry that everybody who reviews these tests will have the same confusion I did.

00xc: Done

00xc: I kept the rdstcp test as before (just inlined a function) since we probably still want to test that the instruction works properly

msft-jlange (approved):

PR #868, opened 2025-11-19
verify: Upgrade verus to official release-2025-11-16

ziqiaozhou:
* Update verus builtin crate names
* Update some annotations since latest verus handles const items in a slightly different ways.
* Add verus_verify for all Impl item since it is ignored if not marked as verus_verify in latest verus.
* Update tool to download prebuilt verus

### DIFF
--- a/scripts/vinstall.sh
+++ b/scripts/vinstall.sh
@@ -5,12 +5,12 @@
 #
 # Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
 # A script to install verus tools
-VERISMO_REV=ffee0c9
-VERUS_RUST_VERSION=1.91
+VERISMO_REV=4f504b7
+VERUS_RUST_VERSION=1.91.0
 
 # Install x86_64-unknown-none target for verus-compatible Rust version
 export RUSTUP_TOOLCHAIN=$VERUS_RUST_VERSION
-rustup target add x86_64-unknown-none
+rustup target add x86_64-unknown-none --toolchain $RUSTUP_TOOLCHAIN
 # Install the verus toolchain
 cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV cargo-v
 # verus-rustc as a wrapper to call verus with proper rustc flags.


scripts/vinstall.sh
```diff
@@ -5,27 +5,18 @@
...
-export RUSTUP_TOOLCHAIN=`grep '^channel' $VERUS_DIR/rust-toolchain.toml | sed -E 's/.*"([0-9]+\.[0-9]+\.[0-9]+)".*/\1/'`
-RUSTUP_TARGET="x86_64-unknown-none"
-rustup target add "$RUSTUP_TARGET"
+VERISMO_REV=ffee0c9
+VERUS_RUST_VERSION=1.91
 
+# Install x86_64-unknown-none target for verus-compatible Rust version
+export RUSTUP_TOOLCHAIN=$VERUS_RUST_VERSION
+rustup toolchain install $RUSTUP_TOOLCHAIN
+rustup target add x86_64-unknown-none
```
stefano-garzarella: What about this?

```suggestion
rustup target add x86_64-unknown-none --toolchain $RUSTUP_TOOLCHAIN
```

scripts/vinstall.sh
stefano-garzarella, line 1: Can you rewrite a bit the commit, I mean something like this:
```
scripts/vinstall: install rust toolchain needed by verus

Add verus-compatible rust toolchain install in vinstall.sh, because ....
```

scripts/vinstall.sh
```diff
@@ -5,27 +5,18 @@
...
 # A script to install verus tools
-VERISMO_REV=7545c6c
-VERUS_DIR=`cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "builtin_macros") | .targets[].src_path'| xargs dirname | xargs -I{} realpath {}/../../../`
-
-# Choose the Rust toolchain version that is compatible with verus.
-export RUSTUP_TOOLCHAIN=`grep '^channel' $VERUS_DIR/rust-toolchain.toml | sed -E 's/.*"([0-9]+\.[0-9]+\.[0-9]+)".*/\1/'`
-RUSTUP_TARGET="x86_64-unknown-none"
-rustup target add "$RUSTUP_TARGET"
+VERISMO_REV=ffee0c9
+VERUS_RUST_VERSION=1.91
```
stefano-garzarella:
```suggestion
VERUS_RUST_VERSION=1.91.0
```

This fixes the panic that I have, but please next time test the script in a clean environment (e.g. using containers), to be sure everything it's fine.

Also, maybe in the CI you should remove the GH step to install the toolchain, and use just this script, so we are sure we are replicating the same steps in the CI and locally

stefano-garzarella: > Also, maybe in the CI you should remove the GH step to install the toolchain, and use just this script, so we are sure we are replicating the same steps in the CI and locally

Oh, I see you don't have it here https://github.com/coconut-svsm/svsm/pull/873 great, that CI should have found this issue while https://github.com/MSRSSP/svsm/commit/f4091f64efb0e90d1e84bf3da2e23a618888f571 hide that issue on your side

ziqiaozhou: Yes. Just to ensure it works well in a clean env and in the CI #873, I tried it at https://github.com/MSRSSP/svsm/pull/21 where I use verus-workflow branch as main branch in my repo. Since this PR is from a branch with verus keyworkd, it will trigger the workflow.

If #873 got merged into main, svsm repo should see "Verification" workflow under actions and can trigger it manually, like https://github.com/MSRSSP/svsm/actions/runs/19556358785

scripts/vinstall.sh
stefano-garzarella: @ziqiaozhou I re-opened this, since I don't see this addressed. It's a minor thing, but it's what we usually do in this repo, so we can identify fast what subsystem a commit affect. That will be nice for backports and bisection.

ziqiaozhou: I forgot to update the first line of comment. Thanks!

ziqiaozhou:
> > I think it is due to missing rust toolchain 1.91-x86_64-unknown-linux-gnu (if you run verus directly, it should throw out that error).
> 
> edit: ~How can I do that?~
> 
> ```
> $ verus --version
> verus: required rust toolchain 1.91.0-x86_64-unknown-linux-gnu not found
> ```
> 
> The `VERUS_RUST_VERSION` is wrong, see [#868 (comment)](https://github.com/coconut-svsm/svsm/pull/868#discussion_r2545073553)
> 
> > The upated version now will install verus from a prebuilt based on your os and arch since building verus is slow. Due to that reason, the script will not automatically download Rust toolchain 1.91. I updated it to install the toolchain automatically.
> 
> edit: ~I still have the panic here :-(~ fixed with [#868 (comment)](https://github.com/coconut-svsm/svsm/pull/868#discussion_r2545073553)
> 
> BTW it's strange to have a panic in that case, why not an error?

Although it may not matter in the updated vinstall.sh. I updated the cargo-v tool version to propagate proper error from verus tool. This is a good suggestion.

PR #871, opened 2025-11-19
task/schedule: fix set affinity to self

msft-jlange:
Fix the case where a thread attempts to set affinity to the current CPU. In such a case, sending an IPI and waiting for its completion will not work, but in such a case the thread affinity is already correct so no work needs to be performed.

### DIFF
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -399,19 +399,23 @@ pub fn go_idle() {
 }
 
 pub fn set_affinity(cpu_index: usize) {
-    // Mark the current task as blocked so it is not scheduled again.
-    let task = this_cpu().current_task();
-    task.set_task_blocked();
+    // Affinity signaling is only required if the target CPU is not the current
+    // CPU.
+    if cpu_index != this_cpu().get_cpu_index() {
+        // Mark the current task as blocked so it is not scheduled again.
+        let task = this_cpu().current_task();
+        task.set_task_blocked();
 
-    // Mark this task as the task pending an affinity change.
-    let mut runqueue = this_cpu().runqueue().lock_write();
-    assert!(runqueue.set_affinity.is_none());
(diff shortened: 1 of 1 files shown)


kernel/src/task/schedule.rs
```diff
@@ -399,19 +399,23 @@ pub fn go_idle() {
...
+        // Mark this task as the task pending an affinity change.
+        let mut runqueue = this_cpu().runqueue().lock_write();
+        assert!(runqueue.set_affinity.is_none());
+        runqueue.set_affinity = Some((task, cpu_index));
+        drop(runqueue);
+
+        // Find another task to run.  The scheduler will complete the affinity
+        // change once a new task has been selected on this processor.
+        schedule();
+    }
```
stefano-garzarella, line 418: nit: I'm a huge fun of early return that in this case will make the diff also smaller, but not a strong opinion in this case, I see value also on not doing that, so up to you.

```suggestion
    // Affinity signaling is only required if the target CPU is not the current
    // CPU.
    if cpu_index == this_cpu().get_cpu_index() {
        return;
    }
    
    // Mark the current task as blocked so it is not scheduled again.
    let task = this_cpu().current_task();
    task.set_task_blocked();

    // Mark this task as the task pending an affinity change.
    let mut runqueue = this_cpu().runqueue().lock_write();
    assert!(runqueue.set_affinity.is_none());
    runqueue.set_affinity = Some((task, cpu_index));
    drop(runqueue);

    // Find another task to run.  The scheduler will complete the affinity
    // change once a new task has been selected on this processor.
    schedule();
```

stefano-garzarella (approved):

kernel/src/task/schedule.rs
msft-jlange: I'm generally not a huge fan of early return, though maybe I could be convinced someday.  Until then, I'd like to leave this code as it is.

PR #872, opened 2025-11-19
test-in-svsm: improve task testing

msft-jlange:
Add a test for `set_affinity()`, which also requires that the main test loop run in a task other than the idle task.

This change also suppresses request loop processing and user init when running in test mode, since those are not part of the test framework.


msft-jlange:
This requires #871 to merge first since that fix is required for the affinity test to pass.

PR #873, opened 2025-11-19
Add manually triggered workflow for verification.

ziqiaozhou:
This PR will add a Verification workflow that can be triggered manually to help developers to check the verification result without installing tool in their local machine.

Although it is not ready to run the CI for all PR reviews, we should want to run the CI manually periodically to capture the broken proofs.

It will make it easier for me to view failure logs and allows the Verus team to identify issues more quickly if their assistance is needed to fix the tool.

As I mentioned at #874, this PR is to add the workflow to capture failed verification.

Update on the trigger condition:

* If the PR is merging from a branch to main with verus keyword, it will trigger Verification workflow.
* Otherwise, you can only trigger Verification workflow manually for a specific branch after the workflow got merged into the that branch.

### DIFF
--- a/Documentation/docs/developer/VERIFICATION.md
+++ b/Documentation/docs/developer/VERIFICATION.md
@@ -4,7 +4,28 @@ Formal verification is done via [Verus](https://github.com/verus-lang/verus).
 To execute verification, ensure you have set up the necessary tools to run
 `cargo verify`.
 
-## Setup
+We provide two options to run verification: GitHub workflow and local build.
+
+## Verification in a Remote Branch via GitHub workflow
+
+When submitting a change related to verification, developers can verify their
+modifications by running the [Verification
+workflow](../../../.github/workflows/manual-verify.yml). This can be done by
+triggering the `Verification` workflow for a specific branch under Actions.
+
+At the moment, only admins can manually trigger this workflow within the
+coconut-svsm organization, since verification is still an experimental feature.
+
+Developers can trigger the workflow from their **fork** (e.g.,
+https://github.com/USERNAME/svsm/actions/workflows/manual-verify.yml), as long
+as both the main branch and the target branch in the fork contain the workflow
+file. Running workflow in a fork avoids consuming CI resources from the
+organization. After running the workflow on their fork, developers can include
+the verification results in their pull request.
+
+## Verification in Local Build
+
+### Setup
 
 Run the following commands to install Verus tools.
 
@@ -21,8 +42,6 @@ cd svsm
   versions, so we may occasionally upgrade the toolchain to support those
   features, which does not guarantee it is the exact version defined by
   `rust-toolchain.toml`
-
-## Build
 
 ### Build svsm with verification
 


stefano-garzarella:
@ziqiaozhou how can I trigger it on this PR for example? I don't see anything new in the `Checks` tab.

ziqiaozhou:
> @ziqiaozhou thanks, but my question was more on the previous version (without the conditions).
> 
> How can I start it manually? maybe there were some missing configuration or if you can explain where I need to go to start it. I don't think we should add the conditions, but just the manual start (if the TCG will agree on that).

If this PR got merged into default branch in the repo, svsm repo should see "Verification" workflow under actions and you can click it manually, like https://github.com/MSRSSP/svsm/actions/runs/19556358785. I do not the permission to do that from my side for coconut-svsm, but I guess you should be able to see
Verification workflor if run 
`gh workflow list`
and then use 
`gh workflow run Verification`

Without merging it into default branch, you may not be able to find that choice under Action in the github web. That is why I shared https://github.com/MSRSSP/svsm/actions/workflows/manual-verify.yml here to give you an idea how it works. https://github.com/MSRSSP/svsm/actions/runs/19556358785 is the action I manually trigger for a branch that contains manual-verify.yml.

Screenshot 2025-11-21 at 10 55 14 AM

stefano-garzarella:
@ziqiaozhou or the idea is to allow the submitter to run the pipeline on their fork and not on the PR?
(maybe in this way we also avoid burning the organization CI cycles)

Documentation/docs/developer/VERIFICATION.md
```diff
@@ -101,6 +101,23 @@ and to write specification and proofs in ghost mode.
   are tightly related to the project code, we may still put them under svsm/kernel with
   `#[verus_verify(rlimit=x)]`
 
+### Run GitHub workflow when submitting changes
```
stefano-garzarella: What about moving this section near `Build svsm with verification` and propose these steps as alternative to the local verification?

Documentation/docs/developer/VERIFICATION.md
```diff
@@ -101,6 +101,23 @@ and to write specification and proofs in ghost mode.
...
+
+When submitting a change, developers should verify their modifications by
```
stefano-garzarella: Do you mean a change to the code or to the verification?

Because if we mean the code, I would perhaps be more relaxed using “can” instead of “should”

Documentation/docs/developer/VERIFICATION.md
```diff
@@ -101,6 +101,23 @@ and to write specification and proofs in ghost mode.
...
+running the [Verification
+workflow](../../../.github/workflows/manual-verify.yml). This can be done by
+triggering the `Verification` workflow for a specific branch under
+[Actions](https://github.com/coconut-svsm/svsm/actions).
```
stefano-garzarella: I'd avoid to link the community repo here, but just use an example of personal repo / fork (e.g. https://github.com/USER/svsm/actions)

scripts/vinstall.sh
```diff
@@ -17,5 +18,5 @@ cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV car
 cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV verus-rustc
 cargo v install-verus
 
-# verus formatter
-cargo install --git https://github.com/verus-lang/verusfmt --rev v0.5.7
+# Download prebuilt verus formatter since the build is slow.
+curl --proto '=https' --tlsv1.2 -LsSf https://github.com/verus-lang/verusfmt/releases/download/$VERUSFMT_REV/verusfmt-installer.sh | sh
```
stefano-garzarella: Can we support both (e.g. adding a parameter to choose to download the binary instead of build it).

Also this will download and install it every time (which could be okay for the CI). But for local dev, can we check if we really need to install it? (e.g. running `verusfmt --version` and check the version)

Documentation/docs/developer/VERIFICATION.md
ziqiaozhou: That sounds more reasonable. I’ve now provided two verification options—GitHub workflow and local build—in two sections positioned close to each other for easier comparison.

scripts/vinstall.sh
ziqiaozhou: I provide --prebuilt-verusfmt option to allow the workflow to use prebuilt.

stefano-garzarella (approved):
@ziqiaozhou thanks for this!

PR #877, opened 2025-11-20
repo: reorganize top-level structure

00xc:
Reorganize the top-level structure of the repo.

* Move the following into `tools/`:
    * igvmbuilder
    * igvmmeasure
    * aproxy
    * ~~xbuild~~
    * gen_meta / print-meta

* ~~Move the following into `lib/`:~~
    * ~~libtcgtpm~~
    * ~~bootlib~~
    * ~~libaproxy~~
    * ~~cpuarch~~
    * ~~elf~~
    * ~~syscall~~
    * ~~virtio-drivers~~
    * ~~test~~
    * ~~verification crates (verify_external, verify_proof, verus_macro_stub, verus_stub)~~

* Untouched:
    * SVSM core (kernel, stage1, user)
    * scripts
    * configs
    * Documentation
    * fuzz
    * release (not sure it belongs in lib/)
    * packit (both a tool and a library, not sure where it should go)


00xc:
> @00xc in the last TSC meeting we discussed a bit the repo reorganization, and we are not sure about `lib/`, we would like to discuss that a bit more (e.g lib/ vs topics vs user/kernel, etc.)
> 
> So our suggestion for now is to remove that changes from this PR and keep only the moves of tools, except for `xbuild`, which we consider to be more relevant to the build system of the entire project, so we thnk it is better that it lives in the root.

Okay, although xbuild does sound like a tool to me 😄

stefano-garzarella:
> Okay, although xbuild does sound like a tool to me 😄

Yeah, I see your point, but it's kind of special tool used as build system for svsm. Organizing a repo, like naming, is not easy xD I'll open a PR, and we can continue the discussion there.

Thanks for this work!

PR #878, opened 2025-11-20
QEMU launcher: add --snapshot [on|off]

osteffenrh:
Add a new command line option to the QEMU launcher:
`--snapshot [on|off]` to enable/disable writes to the disk image.

This can be useful to install missing packages or change configs
in the image, while launching it with the same machine configuration.

### DIFF
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -98,11 +98,11 @@ while [[ $# -gt 0 ]]; do
 done
 
 # Split the QEMU version number so we can specify the correct parameters
-QEMU_VERSION=`$QEMU --version | grep -Po '(?<=version )[^ ]+'`
+QEMU_VERSION=$($QEMU --version | grep -Po '(?<=version )[^ ]+')
 QEMU_MAJOR=${QEMU_VERSION%%.*}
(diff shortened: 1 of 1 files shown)


scripts/launch_guest.sh
```diff
@@ -71,6 +72,22 @@ while [[ $# -gt 0 ]]; do
...
+          SNAPSHOT="on"
+          ;;
+        off)
+          SNAPSHOT="off"
+          ;;
+        *)
+          echo "ERROR: Invalid option for --snapshot: $2" >&2
+          exit 1
+          ;;
+      esac
```
stefano-garzarella: What about removing the parsing here, since QEMU will already do that.

I mean this:
```suggestion
      SNAPSHOT="$2"
```

With that applied, QEMU will print:
```
$ ./scripts/launch_guest.sh --image /tmp/image --snapshot INVALID
...

qemu-system-x86_64: -drive file=/tmp/image,if=none,id=disk0,format=qcow2,snapshot=INVALID: Parameter 'snapshot' expects 'on' or 'off'
```

osteffenrh: Of course.
Done!

PR #879, opened 2025-11-20
Introduce vsock support for attestation

luigix25:
This PR introduces the vsock transport for the attestation.

Vsock is more flexible than the serial port as it has the concept of port, and we can reuse vsock for anything else.

With this PR SVSM tries to perform attestation first with vsock, if it fails it falls back using the serial port.
No change intended to the attestation's logic, I am just adding a new transport type.

Default vsock port for attestation is `1995`. 

To compile it you need to enable the feature `attest` and  `vsock`.  To run it follow the instructions available in the Documentation.

### DIFF
--- a/Documentation/docs/developer/ATTESTATION.md
+++ b/Documentation/docs/developer/ATTESTATION.md
@@ -10,6 +10,7 @@
     - [Frontend](#frontend)
     - [Backend](#backend)
     - [Host Proxy Diagram](#host-proxy-diagram)
+  - [Transport Methods](#transport-methods)
   - [Try for yourself](#try-for-yourself)
 <!--toc:end-->
 
@@ -242,6 +243,15 @@ launching the proxy. The supported backend attestation protocols include:
                              protocol)
 ```
 
+## Transport Methods
+
+SVSM communicates with the attestation proxy using one of two transport methods:
+
+- **vsock**: When the `vsock` feature is enabled, SVSM will first try to use vsock for communication with the host proxy
+  using port `1995`. If it fails, SVSM will try again using the serial port.
+
+- **Serial port**: If vsock is not available, SVSM uses the COM3 serial port for communication with the attestation proxy.
+
 ## Try for yourself
 
 Please keep in mind that the attestation services in SVSM are **experimental**
@@ -270,11 +280,33 @@ SEV-SNP machine with an SVSM-enabled kernel.
     ```shell
     git clone https://github.com/coconut-svsm/svsm.git
     # ... build OVMF, qemu, SVSM IGVM, etc...
-    FW_FILE=... make FEATURES=attest
+    FW_FILE=... make FEATURES=attest                       # serial transport
+    FW_FILE=... make FEATURES=attest,vsock,virtio-drivers  # vsock transport (with serial fallback)
     ```
 
 3. Run the proxy on the host
 
+    The proxy configuration depends on the transport mechanism:
+
+    Common parameters:
+
+    * `--url http://0.0.0.0:8080`: The attestation server is running at `http://0.0.0.0:8080`.
+    * `--protocol kbs`: The attestation server communicates via the KBS
+        protocol, configure the backend to use the KBS protocol.
+
+    **vsock parameters**
+    ```shell
+    cd svsm
+    make aproxy
+    bin/aproxy --protocol kbs \
+               --url http://0.0.0.0:8080 \
+               --vsock
+    ```
+
+    * `--vsock`: Listen for messages from SVSM on a vsock port (default: `1995`), which is
+      the port used by SVSM for attestation.
+
+    **Serial Port parameters**
     ```shell
     cd svsm
     make aproxy
@@ -283,28 +315,33 @@ SEV-SNP machine with an SVSM-enabled kernel.
                --unix /tmp/svsm-proxy.sock \
                --force
     ```
-    This runs the proxy with the following specified in the arguments:
 
-     * `--url http://0.0.0.0:8080`: The attestation server is running at
-       `http://0.0.0.0:8080`.
-     * `--protocol kbs`: The attestation server communicates via the KBS
-       protocol, configure the backend to use the KBS protocol.
-     * `--unix /tmp/svsm-proxy.sock`: Listen for messages from SVSM on a socket
-       created in file `/tmp/svsm-proxy-sock`.
-     * `--force`: Remove the `/tmp/svsm-proxy.sock` file (if it already exists)
-       before creating the socket.
+    * `--unix /tmp/svsm-proxy.sock`: Listen for messages from SVSM on a socket
+      created in file `/tmp/svsm-proxy-sock`.
+    * `--force`: Remove the `/tmp/svsm-proxy.sock` file (if it already exists)
+      before creating the socket.
 
 4. Run a guest with SVSM
 
-    Initially, SVSM communicates over the COM3 serial port. The attestation proxy
-    socket will need to be available in the correct `-serial` argument position to
-    ensure it communicates with the right socket.
+    SVSM will use vsock for communication if the feature is enabled, otherwise it
+    falls back to the COM3 serial port (see [Transport Methods](#transport-methods)).
+    The attestation proxy will need to be configured correctly to ensure proper communication
+    according to the transport used.
 
-    ```shell
-    ./scripts/launch_guest.sh --qemu $QEMU \
-                              --image $QCOW2 \
-                              --aproxy /tmp/svsm-proxy.sock
-    ```
+    * **vsock**
+      ```shell
+      ./scripts/launch_guest.sh --qemu $QEMU \
+                                --image $QCOW2 \
+                                --vsock 3
+      ```
+
+    * **Serial Port**
+      ```shell
+      ./scripts/launch_guest.sh --qemu $QEMU \
+                                --image $QCOW2 \
+                                --aproxy /tmp/svsm-proxy.sock
+      ```
+
     If successful, you should be able to find a message indicating a successful
     attestation within the SVSM boot logs.
 
--- a/Documentation/docs/developer/ATTESTATION.md
+++ b/Documentation/docs/developer/ATTESTATION.md
@@ -287,7 +287,7 @@ SEV-SNP machine with an SVSM-enabled kernel.
 
      * `--url http://0.0.0.0:8080`: The attestation server is running at
        `http://0.0.0.0:8080`.
-     * `--protocol kbs-test`: The attestation server communicates via the KBS
+     * `--protocol kbs`: The attestation server communicates via the KBS
        protocol, configure the backend to use the KBS protocol.
      * `--unix /tmp/svsm-proxy.sock`: Listen for messages from SVSM on a socket
        created in file `/tmp/svsm-proxy-sock`.
--- a/Documentation/docs/developer/ATTESTATION.md
+++ b/Documentation/docs/developer/ATTESTATION.md
@@ -10,6 +10,7 @@
     - [Frontend](#frontend)
     - [Backend](#backend)
     - [Host Proxy Diagram](#host-proxy-diagram)
+  - [Transport Methods](#transport-methods)
   - [Try for yourself](#try-for-yourself)
 <!--toc:end-->
 
@@ -242,6 +243,15 @@ launching the proxy. The supported backend attestation protocols include:
                              protocol)
 ```
 
+## Transport Methods
+
+SVSM communicates with the attestation proxy using one of two transport methods:
+
+- **vsock**: When the `vsock` feature is enabled, SVSM will first try to use vsock for communication with the host proxy
+  using port `1995`. If it fails, SVSM will try again using the serial port.
+
+- **Serial port**: If vsock is not available, SVSM uses the COM3 serial port for communication with the attestation proxy.
+
 ## Try for yourself
 
 Please keep in mind that the attestation services in SVSM are **experimental**
@@ -270,11 +280,33 @@ SEV-SNP machine with an SVSM-enabled kernel.
     ```shell
     git clone https://github.com/coconut-svsm/svsm.git
     # ... build OVMF, qemu, SVSM IGVM, etc...
-    FW_FILE=... make FEATURES=attest
+    FW_FILE=... make FEATURES=attest                       # serial transport
+    FW_FILE=... make FEATURES=attest,vsock,virtio-drivers  # vsock transport (with serial fallback)
     ```
 
 3. Run the proxy on the host
 
+    The proxy configuration depends on the transport mechanism:
+
+    Common parameters:
+
+    * `--url http://0.0.0.0:8080`: The attestation server is running at `http://0.0.0.0:8080`.
+    * `--protocol kbs`: The attestation server communicates via the KBS
+        protocol, configure the backend to use the KBS protocol.
+
+    **vsock parameters**
+    ```shell
+    cd svsm
+    make aproxy
+    bin/aproxy --protocol kbs \
+               --url http://0.0.0.0:8080 \
+               --vsock
+    ```
+
+    * `--vsock`: Listen for messages from SVSM on a vsock port (default: `1995`), which is
+      the port used by SVSM for attestation.
+
+    **Serial Port parameters**
     ```shell
     cd svsm
     make aproxy
@@ -283,28 +315,33 @@ SEV-SNP machine with an SVSM-enabled kernel.
                --unix /tmp/svsm-proxy.sock \
                --force
     ```
-    This runs the proxy with the following specified in the arguments:
 
-     * `--url http://0.0.0.0:8080`: The attestation server is running at
-       `http://0.0.0.0:8080`.
-     * `--protocol kbs`: The attestation server communicates via the KBS
-       protocol, configure the backend to use the KBS protocol.
-     * `--unix /tmp/svsm-proxy.sock`: Listen for messages from SVSM on a socket
-       created in file `/tmp/svsm-proxy.sock`.
-     * `--force`: Remove the `/tmp/svsm-proxy.sock` file (if it already exists)
-       before creating the socket.
+    * `--unix /tmp/svsm-proxy.sock`: Listen for messages from SVSM on a socket
+      created in file `/tmp/svsm-proxy.sock`.
+    * `--force`: Remove the `/tmp/svsm-proxy.sock` file (if it already exists)
+      before creating the socket.
 
 4. Run a guest with SVSM
 
-    Initially, SVSM communicates over the COM3 serial port. The attestation proxy
-    socket will need to be available in the correct `-serial` argument position to
-    ensure it communicates with the right socket.
+    SVSM will use vsock for communication if the feature is enabled, otherwise it
+    falls back to the COM3 serial port (see [Transport Methods](#transport-methods)).
+    The attestation proxy will need to be configured correctly to ensure proper communication
+    according to the transport used.
 
-    ```shell
-    ./scripts/launch_guest.sh --qemu $QEMU \
-                              --image $QCOW2 \
-                              --aproxy /tmp/svsm-proxy.sock
-    ```
+    * **vsock**
+      ```shell
+      ./scripts/launch_guest.sh --qemu $QEMU \
+                                --image $QCOW2 \
+                                --vsock 3
+      ```
+
+    * **Serial Port**
+      ```shell
+      ./scripts/launch_guest.sh --qemu $QEMU \
+                                --image $QCOW2 \
+                                --aproxy /tmp/svsm-proxy.sock
+      ```
+
     If successful, you should be able to find a message indicating a successful
     attestation within the SVSM boot logs.
 
--- a/kernel/src/attest.rs
+++ b/kernel/src/attest.rs
@@ -33,11 +33,35 @@ use serde::Serialize;
 use sha2::{Digest, Sha512};
 use zerocopy::{FromBytes, IntoBytes};
 
+enum Transport<'a> {
+    Serial(SerialPort<'a>),
+}
+
+impl Read for Transport<'_> {
+    type Err = SvsmError;
+
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Err> {
+        match self {
+            Transport::Serial(serial) => serial.read(buf),
+        }
+    }
+}
+
+impl Write for Transport<'_> {
+    type Err = SvsmError;
+
+    fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Err> {
+        match self {
+            Transport::Serial(serial) => serial.write(buf),
+        }
+    }
+}
+
 /// The attestation driver that communicates with the proxy via some communication channel (serial
 /// port, virtio-vsock, etc...).
 #[allow(missing_debug_implementations)]
 pub struct AttestationDriver<'a> {
-    sp: SerialPort<'a>,
+    transport: Transport<'a>,
     tee: Tee,
     ecc: EccKey,
 }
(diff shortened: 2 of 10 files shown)


luigix25:
~Depends on #766~

tools/aproxy/src/main.rs
```diff
@@ -26,32 +27,46 @@ struct Args {
...
+
+    /// vsock port used by SVSM for attestation
+    #[clap(long, conflicts_with_all = ["unix", "force"], required_unless_present("unix"))]
+    vsock_port: Option<u32>,
 
     /// Force Unix domain socket removal before bind
-    #[clap(long, short, default_value_t = false)]
+    #[clap(long, short, requires = "unix", default_value_t = false)]
     force: bool,
 }
```
00xc: Just as a suggestion, instead of doing this dance (which can become hard to read, specially if we decide to add more transports), I'd suggest using [clap groups](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#argument-relations).

So something like:

```rust
#[clap(long, group = "transport", required = true)]
unix: Option<String>,

#[clap(long, group = "transport", required = true, conflicts_with = "force")]
vsock_port: Option<u32>,

#[clap(long, short, requires = "unix", default_value_t = false)]
force: bool,
```

This way only one of the options in the "transport" group can be specified.

Ideally, just setting `requires = "unix"` on `force` should be enough, but due to a bug in clap (see clap-rs/clap#4707) we have to manually add `conflicts_with = "force"` to `vsock_port` so that `--vsock-port $port --force` is rejected.

luigix25: I didn't know about clap groups, thanks!

luigix25: I guess there is some kind of bug in the clap help message.

With your suggested changes
```shell
bin/aproxy --vsock-port 1234 
error: the following required arguments were not provided:
  --url <URL>
  --protocol <BACKEND>
  --unix <UNIX>
```

Without 

```shell
bin/aproxy_old --vsock-port 1234         
error: the following required arguments were not provided:
  --url <URL>
  --protocol <BACKEND>
```

First one doesn't make too much sense. as `vsock-port` and `unix` cannot coexist.

Clap doesn't allow to use them together, but the help message is wrong

00xc: Right, it seems groups have even more bugs :( feel free to discard my suggestion, we can make it work later if we find a better way

luigix25:
- Rebased to latest main to pickup vsock changes.
- Added `read_exact` and `write_all` in the Read/Write trait

kernel/src/error.rs
```diff
@@ -148,6 +148,8 @@ pub enum SvsmError {
     /// Errors related to vsock.
     #[cfg(feature = "vsock")]
     Vsock(VsockError),
+    /// Unexpected end of input.
```
stefano-garzarella: EoI or EoF ?

stefano-garzarella: Also not sure if we should introduce IO errors.

tools/aproxy/src/main.rs
```diff
@@ -24,11 +25,19 @@ struct Args {
...
-    #[clap(long)]
-    unix: String,
+    #[clap(
+        long,
+        conflicts_with = "vsock_port",
+        required_unless_present("vsock_port")
+    )]
+    unix: Option<String>,
+
+    /// vsock port used by SVSM for attestation
```
stefano-garzarella: Can we be a bit more precise here?

IIUC this is the port where aproxy is listening and SVSM will connect, right?

So maybe `vsock listening port where SVSM will connect` something like that. We don't need `for attestation` is implicit since this is the attestation proxy...

tools/aproxy/src/main.rs
```diff
@@ -24,11 +25,19 @@ struct Args {
...
+    #[clap(long, conflicts_with_all = ["unix", "force"], required_unless_present("unix"))]
+    vsock_port: Option<u32>,
```
stefano-garzarella: Should we add a default also here? I mean the kernel uses 1995 as default, why not having the same here?

maybe the param can be `--vsock PORT` where PORT is optional and default to 1995. Is it doable with clap?

kernel/src/io.rs
```diff
@@ -70,11 +70,35 @@ pub trait Read {
     type Err: Into<SvsmError>;
 
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Err>;
+
+    fn read_exact(&mut self, buf: &mut [u8]) -> Result<(), SvsmError> {
```
stefano-garzarella: Can we use `Self::Err` like for other API ?

kernel/src/io.rs
```diff
@@ -70,11 +70,35 @@ pub trait Read {
...
+    }
 }
 
 /// Generic Write trait to be implemented over any transport channel when writing multiple bytes.
 pub trait Write {
     type Err: Into<SvsmError>;
 
     fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Err>;
+
+    fn write_all(&mut self, buf: &[u8]) -> Result<(), SvsmError> {
```
stefano-garzarella: Ditto about the error

tools/aproxy/src/main.rs
```diff
@@ -43,22 +52,24 @@ enum ArgsBackend {
...
-    }
-
-    let listener = UnixListener::bind(args.unix).context("unable to bind to UNIX socket")?;
+    if let Some(port) = args.vsock_port {
+        let listener = VsockListener::bind(&VsockAddr::new(VMADDR_CID_ANY, port))
+            .context("bind and listen failed")?;
+        for stream in listener.incoming() {
+            let mut stream = stream.context("Failed to connect to the vsock socket")?;
+            let mut http_client = backend::HttpClient::new(args.url.clone(), args.backend.into())?;
+            attest::attest(&mut stream, &mut http_client)?;
```
stefano-garzarella: Can we introduce an helper for this to avoid code duplication?

kernel/src/attest.rs
```diff
@@ -34,21 +37,59 @@ use sha2::{Digest, Sha512};
...
+    Vsock(VsockStream),
     Serial(SerialPort<'a>),
 }
 
 impl Transport<'_> {
     fn write(&mut self, buf: &[u8]) -> Result<usize, SvsmError> {
         match self {
+            #[cfg(feature = "vsock")]
+            Transport::Vsock(vsock) => {
+                vsock.write_all(buf)?;
```
stefano-garzarella: mmm, it seems strange to me call write_all() inside a write().
IMO we should implement/use write_all() also for the serial and call this method write_all()

kernel/src/attest.rs
```diff
@@ -34,21 +37,59 @@ use sha2::{Digest, Sha512};
...
+            }
             Transport::Serial(serial) => serial.write(buf),
         }
     }
 
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, SvsmError> {
         match self {
+            #[cfg(feature = "vsock")]
+            Transport::Vsock(vsock) => {
+                vsock.read_exact(buf)?;
```
stefano-garzarella: Ditto about read/read_exact()

kernel/src/attest.rs
```diff
@@ -34,21 +37,59 @@ use sha2::{Digest, Sha512};
...
         }
     }
+
+    #[cfg(feature = "vsock")]
+    fn new() -> Self {
+        match VsockStream::connect(ATTEST_DEFAULT_VSOCK_PORT, VMADDR_CID_HOST) {
+            Ok(value) => Transport::Vsock(value),
+            Err(e) => {
+                log::warn!(
+                    "Vsock error: {e:?} occurred when attempting to start attestation on port {ATTEST_DEFAULT_VSOCK_PORT}. Attempting to start again using the serial port.",
```
stefano-garzarella: the message is unclear, we should make more clear that the connection is failing. In this way it seems that we are listening.

kernel/src/svsm.rs
```diff
@@ -543,6 +543,9 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
         panic!("Failed to prepare guest FW: {e:#?}");
     }
 
+    #[cfg(feature = "virtio-drivers")]
+    initialize_virtio_mmio(&boot_params).expect("Failed to initialize virtio-mmio drivers");
+
```
stefano-garzarella, line 548: Why this change? This should at least mentioned in the commit description

kernel/src/types.rs
```diff
@@ -99,3 +99,6 @@ pub const TPR_NORMAL: usize = 0;
 pub const TPR_LOCK: usize = 2;
 pub const TPR_SYNCH: usize = 13;
 pub const TPR_IPI: usize = 14;
+
+pub const ATTEST_DEFAULT_VSOCK_PORT: u32 = 1995;
+pub const ATTEST_DEFAULT_SERIAL_IO_ADDR: u16 = 0x3e8; // COM3
```
stefano-garzarella: not sure if this is the best place, but I need to think more about this. Any other option?

stefano-garzarella (commented):
Did a very quick review and left some comments. Thanks

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -283,28 +315,33 @@ SEV-SNP machine with an SVSM-enabled kernel.
                --unix /tmp/svsm-proxy.sock \
                --force
     ```
-    This runs the proxy with the following specified in the arguments:
 
-     * `--url http://0.0.0.0:8080`: The attestation server is running at
-       `http://0.0.0.0:8080`.
-     * `--protocol kbs-test`: The attestation server communicates via the KBS
```
stefano-garzarella: is it okay to change this from `kbs-test` to `kbs` ?

luigix25: I don't think `kbs-test` is supported anymore in aproxy

kernel/src/svsm.rs
luigix25: attestation (may) require vsock, so we need to initialize it before it. Will add it to the description

Documentation/docs/developer/ATTESTATION.md
stefano-garzarella: So, maybe better to fix it in a separate commit or at least mention it in the commit.
I thought was a rebase issue.

tools/aproxy/src/main.rs
luigix25: good question, but I don't think it's easily doable: 

The problem is that `VsockListener` and `UnixListener` return two different types and there is no "Stream" Trait. The only option is to create something similar to what I did with `Transport`. wdyt?

Without any changes I think we could create a function to group these two calls
```            
let mut http_client = backend::HttpClient::new(args.url.clone(), args.backend.into())?;
attest::attest(&mut stream, &mut http_client)?;
```

I don't know it it's worth it or not

tools/aproxy/src/main.rs
luigix25: will do.

kernel/src/error.rs
luigix25: `read_exact` should return an error and I don't think we already have an "appropriate" error here, this is why I added a new one.

kernel/src/types.rs
luigix25: very good question. Maybe within `attest.rs` itself? I don't expect these consts be used anywhere else

tools/aproxy/src/main.rs
stefano-garzarella: > I don't know it it's worth it or not

Maybe no.

Or we can try a macro. Anyway, feel free to ignore if it will be too complex, we can do it later.

kernel/src/error.rs
stefano-garzarella: yes, that part is clear, my 2 questions are:
- why calling it Eof and have the description talking about `end of input` ?
- should we introduce `IO` or `InputOutput` errors like for example we did for `Vsock`, `Apic`, `Attestation`, etc. and move `Eof` under that type ?

kernel/src/types.rs
stefano-garzarella: yep, maybe that place is better. those are more default values than types, but I hope we can remove those with the device tree, so let's move them to attest.rs for now.

tools/aproxy/src/main.rs
luigix25: Ok, I'll keep this open for now so we can get more feedbacks.

kernel/src/error.rs
luigix25: >why calling it Eof and have the description talking about end of input ?

at the beginning it was end of input, I then changed to EoF but forgot to update the doc.

>should we introduce IO or InputOutput errors like for example we did for Vsock, Apic, Attestation, etc. and move Eof under that type ?

Right

tools/aproxy/src/main.rs
luigix25: I ended up using a macro. closing this

luigix25:
addressed @stefano-garzarella's comments
aproxy:
- Parameter renamed `--vsock`
- Some changes to clap. Hope it looks cleaner now. Unfortunately we sill need `conflicts_with = "vsock"` [@00xc]
- using a macro to simplify the code

svsm:
- Now using read_exact/write_all everywhere.
- Introduced IoError
- read_exact/write_all now use Self::Err

Improved commit messages.

kernel/src/attest.rs
```diff
@@ -184,7 +213,7 @@ impl AttestationDriver<'_> {
     fn read(&mut self) -> Result<Vec<u8>, AttestationError> {
         let len = {
             let mut bytes = [0u8; 8];
-            self.sp
+            self.transport
                 .read(&mut bytes)
                 .or(Err(AttestationError::ProxyRead))?;
```
stefano-garzarella, line 252: all of this `.or()` are pre-existings, so not an issue for this PR, but I'd like to avoid it and maybe extend `ProxyRead`/`ProxyWrite` to wrap the source error

kernel/src/attest.rs
```diff
@@ -52,11 +62,41 @@ impl Write for Transport<'_> {
...
+impl Transport<'_> {
+    #[cfg(feature = "vsock")]
+    fn new() -> Self {
+        match VsockStream::connect(ATTEST_DEFAULT_VSOCK_PORT, VMADDR_CID_HOST) {
+            Ok(value) => Transport::Vsock(value),
+            Err(e) => {
+                log::warn!(
+                    "Failed to connect to attestation proxy on vsock port \
+                     {ATTEST_DEFAULT_VSOCK_PORT}: {e:?}. \
+                     Attempting to start again using the serial port.",
```
stefano-garzarella: nit: what about:
`Falling back to serial port transport.`

tools/aproxy/src/main.rs
```diff
@@ -40,26 +46,30 @@ enum ArgsBackend {
...
 
+macro_rules! accept_loop {
+    ($listener:expr, $url:expr, $backend:expr) => {
+        for stream in $listener.incoming() {
+            let mut stream = stream.context("Failed to accept connection")?;
+            let mut http_client = backend::HttpClient::new($url.clone(), $backend.into())?;
+            attest::attest(&mut stream, &mut http_client)?;
+        }
+    };
+}
```
stefano-garzarella: What about avoiding the macro in this way:
```diff
@@ -46,14 +50,17 @@ enum ArgsBackend {
     Kbs,
 }
 
-macro_rules! accept_loop {
-    ($listener:expr, $url:expr, $backend:expr) => {
-        for stream in $listener.incoming() {
-            let mut stream = stream.context("Failed to accept connection")?;
-            let mut http_client = backend::HttpClient::new($url.clone(), $backend.into())?;
-            attest::attest(&mut stream, &mut http_client)?;
-        }
-    };
+fn accept_loop<S: Read + Write>(
+    incoming: impl Iterator<Item = io::Result<S>>,
+    url: &str,
+    backend: ArgsBackend,
+) -> anyhow::Result<()> {
+    for stream in incoming {
+        let mut stream = stream.context("Failed to accept connection")?;
+        let mut http_client = backend::HttpClient::new(url.to_string(), backend.into())?;
+        attest::attest(&mut stream, &mut http_client)?;
+    }
+    Ok(())
 }
 
 fn main() -> anyhow::Result<()> {
@@ -62,14 +69,14 @@ fn main() -> anyhow::Result<()> {
     if let Some(port) = args.vsock {
         let listener = VsockListener::bind(&VsockAddr::new(VMADDR_CID_ANY, port))
             .context("bind and listen failed")?;
-        accept_loop!(listener, args.url, args.backend);
+        accept_loop(listener.incoming(), &args.url, args.backend)?;
     } else if let Some(unix) = args.unix {
         if args.force {
             let _ = fs::remove_file(&unix);
         }
 
         let listener = UnixListener::bind(unix).context("unable to bind to UNIX socket")?;
-        accept_loop!(listener, args.url, args.backend);
+        accept_loop(listener.incoming(), &args.url, args.backend)?;
     }
 
     Ok(())
```

kernel/src/attest.rs
```diff
@@ -33,7 +35,13 @@ use serde::Serialize;
 use sha2::{Digest, Sha512};
 use zerocopy::{FromBytes, IntoBytes};
 
+#[cfg(feature = "vsock")]
+const ATTEST_DEFAULT_VSOCK_PORT: u32 = 1995;
```
stefano-garzarella: What about defining this in `libaproxy` so both this code and `aproxy` can use the same constant?

Documentation/docs/developer/ATTESTATION.md
stefano-garzarella: Since we are here, can we replace also `--backend` with `--protocol` ? (if you want to do in another PR, it's also fine)

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -283,28 +315,33 @@ SEV-SNP machine with an SVSM-enabled kernel.
...
-     * `--url http://0.0.0.0:8080`: The attestation server is running at
-       `http://0.0.0.0:8080`.
-     * `--protocol kbs`: The attestation server communicates via the KBS
-       protocol, configure the backend to use the KBS protocol.
-     * `--unix /tmp/svsm-proxy.sock`: Listen for messages from SVSM on a socket
-       created in file `/tmp/svsm-proxy-sock`.
-     * `--force`: Remove the `/tmp/svsm-proxy.sock` file (if it already exists)
-       before creating the socket.
+    * `--unix /tmp/svsm-proxy.sock`: Listen for messages from SVSM on a socket
+      created in file `/tmp/svsm-proxy-sock`.
```
stefano-garzarella: Pre-existing, but since you are touching this code, this should be `/tmp/svsm-proxy.sock`

stefano-garzarella (commented):
Overall LGTM, just some minor comments.

Documentation/docs/developer/ATTESTATION.md
luigix25: yep, I'll open another PR fixing all these small problems!

kernel/src/attest.rs
luigix25: makes sense, yes!

Documentation/docs/developer/ATTESTATION.md
luigix25: opened #1046

Documentation/docs/developer/ATTESTATION.md
luigix25: See #1046

tools/aproxy/src/main.rs
luigix25: nice, thanks!

kernel/src/attest.rs
luigix25: this makes sense. Unfortunately clap requires a string for `default_missing_value`, which is the only place where this value is used in aproxy. 

I see two options here:
- defining a "companion" string like `ATTEST_DEFAULT_VSOCK_PORT_STR`
- Use `const_format` to convert the u32 to a string at compile time

I have a slight preference over the second one.

luigix25:
Addressed all @stefano-garzarella's comments:
- Moved ATTEST_DEFAULT_VSOCK_PORT to libaproxy so it consumed by aproxy and kernel
- minor changes to aproxy: using a function instead of a macro for stream handling
- rebased to latest main

tools/aproxy/src/main.rs
```diff
@@ -10,14 +10,18 @@ mod backend;
 
 use anyhow::Context;
 use clap::{Parser, ValueEnum};
+use const_format::formatcp;
+use libaproxy::ATTEST_DEFAULT_VSOCK_PORT;
```
stefano-garzarella, line 14: This is added later by another commit...

kernel/src/attest.rs
```diff
@@ -70,11 +108,6 @@ impl TryFrom<Tee> for AttestationDriver<'_> {
     type Error = SvsmError;
 
     fn try_from(tee: Tee) -> Result<Self, Self::Error> {
-        // TODO: Make the IO port configurable/discoverable for other transport mechanisms such as
```
stefano-garzarella, line 49: Please don't remove this TODO, maybe move this near the new `ATTEST_DEFAULT_SERIAL_IO_ADDR` declaration.

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -283,28 +315,33 @@ SEV-SNP machine with an SVSM-enabled kernel.
...
+      ./scripts/launch_guest.sh --qemu $QEMU \
+                                --image $QCOW2 \
+                                --vsock 3
+      ```
+
+    * **Serial Port**
+      ```shell
+      ./scripts/launch_guest.sh --qemu $QEMU \
+                                --image $QCOW2 \
+                                --aproxy /tmp/svsm-proxy.sock
```
stefano-garzarella, line 342: maybe we should rename `--aproxy` to `aproxy-serial` or extend it to support both serial or vsock, anyway we can fix it later.

stefano-garzarella (changes requested):
We should fix the bisection breakage in commit 4, the rest LGTM. I left some minor comments, the one related to the script can be fixed later.

kernel/src/attest.rs
stefano-garzarella: I'm not sure that adding a new dependency just for this is a good idea.

stefano-garzarella: BTW not a strong opinion on my side

Documentation/docs/developer/ATTESTATION.md
luigix25: >extend it to support both serial or vsock, anyway we can fix it later

what do you mean? for vsock we just need to attach the device (`--vsock cid`)

stefano-garzarella: I mean that `--aproxy vsock:3` can attach a vsock device with CID = 3, similar to `--aproxy serial:/path/to/unix` that can attach a serial port with the unix backend. IMO is more uniform.

tools/aproxy/src/main.rs
luigix25: oops, moving it to `aproxy: enable vsock for attestation`

Documentation/docs/developer/ATTESTATION.md
luigix25: uh ok, will do in a followup PR.

kernel/src/attest.rs
luigix25: yeah, it's not super nice, but the alternatives are not nice either. We can change this later on if this becomes a packaging problem.

kernel/src/attest.rs
```diff
@@ -33,11 +35,73 @@ use serde::Serialize;
 use sha2::{Digest, Sha512};
 use zerocopy::{FromBytes, IntoBytes};
 
+// TODO: Make the IO port configurable/discoverable for other transport mechanisms such as
+// virtio-vsock
```
stefano-garzarella: I'd remove the vsock mention here and also I'd mention that maybe we can remove the support for serial port at all

stefano-garzarella (approved):

stefano-garzarella:
@tylerfanelli can you take a look?

tylerfanelli (approved):
LGTM, thanks!

PR #880, opened 2025-11-21
kernel: implement boot heap to improve memory management

msft-jlange:
This change permits stage2 to allocate data within the heap that is managed by the kernel after the kernel starts.  This simplifies the allocation of some boot data (like IGVM parameters) and also permits boot data (like IGVM parameters) to be freed after the boot process is complete rather than keeping the memory around forever.

### DIFF
--- a/kernel/src/igvm_params.rs
+++ b/kernel/src/igvm_params.rs
@@ -10,16 +10,18 @@ use crate::acpi::tables::{load_acpi_cpu_info, ACPICPUInfo, ACPITable};
 use crate::address::{Address, PhysAddr, VirtAddr};
 use crate::cpu::efer::EFERFlags;
 use crate::error::SvsmError;
+use crate::mm::alloc::free_multiple_pages;
 use crate::mm::{GuestPtr, PerCPUPageMappingGuard, PAGE_SIZE};
 use crate::platform::{PageStateChangeOp, PageValidateOp, SevFWMetaData, SVSM_PLATFORM};
 use crate::types::PageSize;
-use crate::utils::MemoryRegion;
+use crate::utils::{round_to_pages, MemoryRegion};
 use alloc::vec::Vec;
 use cpuarch::vmsa::VMSA;
 
 use bootlib::igvm_params::{IgvmGuestContext, IgvmParamBlock, IgvmParamPage};
 use bootlib::kernel_launch::LOWMEM_END;
 use core::mem::size_of;
+use core::ops::Deref;
 use core::slice;
 use igvm_defs::{IgvmEnvironmentInfo, MemoryMapEntryType, IGVM_VHS_MEMORY_MAP_ENTRY};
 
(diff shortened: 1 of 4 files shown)


msft-jlange:
> @msft-jlange I didn't take a look of the code, but from the CI it seems some `mm::alloc` tests are failing.

These tests pass now as a result of #882.

kernel/src/igvm_params.rs
```diff
@@ -467,3 +472,39 @@ impl IgvmParams<'_> {
...
+    type Target = IgvmParams<'a>;
+    fn deref(&self) -> &IgvmParams<'a> {
+        &self.igvm_params
+    }
+}
+
+impl Drop for IgvmBox<'_> {
+    fn drop(&mut self) {
+        let page_count =
+            (self.igvm_params.igvm_param_block.param_area_size as usize).div_ceil(PAGE_SIZE);
```
peterfang: This can be replaced with `align_up(param_area_size, PAGE_SIZE) / PAGE_SIZE` which is optimized for power-of-2 calculations.

kernel/src/mm/alloc.rs
```diff
@@ -1788,6 +1795,14 @@ pub fn free_page(vaddr: VirtAddr) {
     ROOT_MEM.lock().free_page(vaddr)
 }
 
+/// Free multiple pages at the given virtual address.
+pub fn free_multiple_pages(vaddr: VirtAddr, count: usize) {
```
peterfang, line 1799: In other OS code bases I usually see a name like `free_pages()`, and the second argument already suggests that it can be more than one page. I'm raising this because kernel MM code can be shared by all kernel components.

kernel/src/stage2.rs
```diff
@@ -52,13 +52,71 @@ extern "C" {
...
+    pub fn page_count(&self) -> usize {
+        self.page_count
+    }
+
+    pub fn next_free(&self) -> usize {
+        self.next_free
+    }
+
+    pub fn allocate(&mut self, size: usize) -> Result<(VirtAddr, PhysAddr), SvsmError> {
+        let page_count = size.div_ceil(PAGE_SIZE);
```
peterfang: The disassembly of this looks kind of intricate. We can leverage `align_up()` for this.

kernel/src/mm/alloc.rs
msft-jlange: I'd like to defer this change because the page heap logic will be largely rewritten once we move away from the direct memory map, and this will end up being part of that overhaul.

kernel/src/igvm_params.rs
msft-jlange: I added a `round_to_pages()` function which essentially does what you recommended, and which can be used in multiple places.

peterfang:
> > Overall this is a great change and all of my comments are pretty minor. I didn't immediately find a place where boot data like IGVM parameters is freed after boot is complete. Is that in an upcoming PR or did I just not look hard enough?
> 
> The IGVM parameters will be freed when the `IgvmBox` is dropped. That will happen when `svsm_init()` returns (this is one of the reasons for my earlier PR to split `svsm_main()` and `svsm_init()` - so that boot-time data would be dropped before boot is complete).

Ah got it. As long as IGVM parameters are only used for `svsm_init()` purposes this looks good then. Thanks.

peterfang (approved):

PR #881, opened 2025-11-21
cpu: Make sure interrupts do not disappear

MelodyHuibo:
When the guest is processing an interrupt, an intercept can happen which will stop the injection and set EXITINTINFO.V = 1.

That then means that the interrupt needs to be reinjected. Usually, when the hypervisor sees the valid bit, it will re-inject the interrupt.

However, with Alternate Injection, the hypervisor can not re-inject the interrupt into the guest directly but the SVSM needs to do that.

At the same time, the busy bit gets set so clear it too.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -1060,6 +1060,27 @@ impl PerCpu {
         }
     }
 
+    pub fn ai_handle_intercepts(&self, vmsa: &mut VMSA) {
+        let use_alternate_injection = SVSM_PLATFORM.query_apic_registration_state();
+        let g_eii = vmsa.guest_exitintinfo;
+
+        if !use_alternate_injection {
+            return;
+        }
+
+        // Re-inject events when intercept happens
+        if g_eii.valid() {
+            vmsa.event_inj = g_eii;
+        }
+
+        // Clear busy bit
+        let mut vintr_ctrl = vmsa.vintr_ctrl;
+        if vintr_ctrl.busy() {
+            vintr_ctrl.set_busy(false);
+            vmsa.vintr_ctrl = vintr_ctrl;
+        }
+    }
+
     pub fn use_apic_emulation(&self) -> bool {
         self.guest_apic().is_some()
     }
--- a/cpuarch/src/vmsa.rs
+++ b/cpuarch/src/vmsa.rs
@@ -78,7 +78,7 @@ pub struct VIntrCtrl {
     pub v_intr_vector: u8,
     #[bits(23)]
     _rsvd_62_40: u32,
-    busy: bool,
+    pub busy: bool,
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
--- a/kernel/src/vmm/execloop.rs
+++ b/kernel/src/vmm/execloop.rs
@@ -139,6 +139,8 @@ pub fn enter_guest(mut regs: &[GuestRegister]) -> GuestExitMessage {
             // Clear EFER.SVME in guest VMSA.
             vmsa.disable();
 
+            cpu.ai_handle_intercepts(vmsa);
+
             if let Some(msg) = get_svsm_request_message(vmsa_ref.deref_mut()) {
                 return msg;
             }


joergroedel (approved):
Testing passed and changes look good to me. But waiting for @msft-jlange as the SME for final approval before merging.

kernel/src/cpu/percpu.rs
```diff
@@ -1055,8 +1055,22 @@ impl PerCpu {
     }
 
     pub fn update_apic_emulation(&self, vmsa: &mut VMSA, caa: Option<NonNull<SvsmCaa>>) {
+        let vintr_ctrl = vmsa.vintr_ctrl;
+        let g_eii = vmsa.guest_exitintinfo;
```
msft-jlange, line 1065: The guest EXITINTINFO can be set in cases other than event injection.  There may be cases where a nested page fault may arise during event delivery where the host can resolve the nested page fault by itself without requiring further guest intervention.  These can happen on systems that don't have alternate injection configured at all.  For this reason, handling of EXITINTINFO should not be placed behind any code that checks for APIC emulation.  The right way to handle this would be to check EXITINTINFO on every transition from the guest into the SVSM (not on the return-to-guest path) and not gated behind APIC emulation support.  This would be in `execloop::enter_guest()` after the call to `vmsa.disable()` and before the call to `get_svsm_request_message()`.

MelodyHuibo: The EXITINTINFO value in the VMSA is only of matter when running with Alternate Injection or with Secure AVIC. However, when running with Secure AVIC, the VMSA EXITINTINFO information is copied to the VMSA EVENTINJ field on VMEXIT.  
In: APM Volume 2 15.36.21.2: "On a VMEXIT in Secure AVIC mode, the guest virtual interrupt state (VIntrCtrl bits 15:0) and the Secure AVIC Control MSR are saved in the VMSA. Additionally, EXITINTINFO is written to the VMSA EVENTINJ field if EXITINTINFO[VECTOR] is not equal to 29 (#VC). This causes the processor to re-inject the interrupted event automatically on the next VMRUN." 
So it is not required to be checked by the SVSM. If the guest is not using either Alternate Injection or Secure AVIC, it is the VMCB fields that are consulted.
Therefore, the only path that needs to be concerned with EXITINTINFO is the Alternate Injection path. So I think current design is more reasonable.

msft-jlange: > The EXITINTINFO value in the VMSA is only of matter when running with Alternate Injection or with Secure AVIC.

That's a great point.  However, it is still the case that the processing that occurs following a guest exit might need to change based on whether there was a pending interrupt.  For that reason, I still maintain that the most correct place to examine EXITINTINFO is immediately upon exit from the guest, before any further processing occurs, so that none of the SVSM processing can generate a state that cannot be reconciled by the time we are ready to reenter the guest.

MelodyHuibo: @msft-jlange Hi Jon, I just updated the PR, let me know if this is what you had in mind. I will finalize it then

kernel/src/cpu/percpu.rs
```diff
@@ -1060,6 +1060,29 @@ impl PerCpu {
...
+        let use_alternate_injection = SVSM_PLATFORM.query_apic_registration_state();
+        let g_eii = vmsa.guest_exitintinfo;
+        let vintr_ctrl = vmsa.vintr_ctrl;
+
+        if !use_alternate_injection {
+            return;
+        }
+
+        if let Some(mut apic) = self.guest_apic_mut() {
+            if g_eii.valid() && g_eii.event_type() == VmsaEventType::Interrupt {
```
msft-jlange: If the pending interrupt type is not `Interrupt`, you still need to reinject the event.  Consider the case that #NPF occurs during an attempt to dispatch an exception.  In this case, the processor will set the Busy bit as well as making EXITINTINFO valid.  But the exception must not be lost in this case; it must be reinjected.  So the correct behavior is to move the pending event into the event inject field regardless of its type.

kernel/src/cpu/apic.rs
```diff
@@ -280,6 +284,12 @@ impl LocalApic {
         self.update_required = true;
     }
 
+    pub fn reinject_interrupt<T: GuestCpuState>(&mut self, cpu_state: &mut T, irq: u8) {
+        // Reinject the interrupt not injected by the guest because of the interception
+        cpu_state.queue_interrupt(irq);
+        self.interrupt_queued = true;
```
msft-jlange: This logic shouldn't be necessary.  Instead of reinjecting the interrupt through the APIC, you can simply copy EXITINTINFO into the EVENTINJ field.  That will preserve the pending interrupt state, and from the point of view of the APIC emulator, it will simply observe that a previously presented interrupt has not yet been processed by the guest and thus it will leave the APIC state alone (exactly as if the SVSM is reentered after interrupt injection without the guest executing at all).

kernel/src/cpu/percpu.rs
```diff
@@ -1073,6 +1060,30 @@ impl PerCpu {
...
+        if !use_alternate_injection {
+            return;
+        }
+
+        // Re-inject events when intercept happens
+        if g_eii.valid() {
+            let event_inj: u64 = g_eii.into();
+
+            // Clear the reserved field
+            vmsa.event_inj = VmsaEventInject::from(event_inj & !(0x7ffff << 12));
```
msft-jlange: Can you explain the reason for the bit manipulation here?  The processor architecture specifically defines EVENT_INJ and EXITINTINFO to be the same format, and the COCONUT sources define both of those to be `VmsaEventInject`.  It should be possible to simply copy the EXITINTINFO directly into the EVENT_INJ field.

MelodyHuibo: Hi Jon, these bits are reserved bits in EVENT_INJ, in the APM these bits are marked as "reserved, SBZ", The definition shown in APM for SBZ: 

Should be zero. An attempt by software to set an SBZ bit to 1 results in undefined behavior. So I just tried to leave the reserved bits as 0, not bring any unexpected undefined behavior, do this make sense?

msft-jlange: Yes, but my point is that the hardware will set those bits to zero in the EXITINTINFO, and thus there is no reason for you to force them to zero here.  My question is why you expect the hardware to present EXITINTINFO with non-zero SBZ bits?

MelodyHuibo: Oh, I don't expect the hardware to present with non-zero SBZ bits, I am just being overly cautious, perhaps a bit paranoid. :) If you think I should just copy all of bits directly, that's certainly a lot cleaner.

msft-jlange: If you're going to be paranoid, your code needs to have a clear comment about why it's appropriate to be paranoid.  In this case, I don't think being paranoid is appropriate because the hardware is designed to provide the information correctly.

MelodyHuibo: OK, thank you, I will remove it. :)

PR #883, opened 2025-11-24
kernel: initialize CPUID/secrets only on SNP platforms

msft-jlange:
The CPUID and secrets page are only used on SNP platforms, so there is no need to initialize them on other platforms.  This also ensures they are not attempted to be used with uninitialized contents.

### DIFF
--- a/kernel/src/sev/secrets_page.rs
+++ b/kernel/src/sev/secrets_page.rs
@@ -6,13 +6,16 @@
 
 use crate::address::VirtAddr;
 use crate::locking::{RWLock, ReadLockGuard, WriteLockGuard};
+use crate::mm::PAGE_SIZE;
 use crate::protocols::core::CORE_PROTOCOL_VERSION_MAX;
 use crate::sev::vmsa::VMPL_MAX;
 use crate::types::GUEST_VMPL;
+use crate::utils::zero_mem_region;
 
 extern crate alloc;
 use alloc::boxed::Box;
 use core::ptr;
+use core::sync::atomic::{AtomicBool, Ordering};
 
 pub const VMPCK_SIZE: usize = 32;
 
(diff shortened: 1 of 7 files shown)


kernel/src/stage2.rs
```diff
@@ -523,8 +538,14 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
         kernel_fs_end: u64::from(launch_info.kernel_fs_end),
         stage2_start: 0x800000u64,
         stage2_end: launch_info.stage2_end as u64,
-        cpuid_page: launch_info.cpuid_page as u64,
-        secrets_page: launch_info.secrets_page as u64,
+        cpuid_page: match cpuid_page {
+            None => 0,
+            Some(cpuid_vaddr) => u64::from(cpuid_vaddr),
+        },
```
00xc: Just as a suggestion, this can be made a bit more concise:

```suggestion
        cpuid_page: cpuid_page.unwrap_or_else(VirtAddr::null).into(),
```

Alternatively, `.unwrap_or_default()` can be used instead of `unwrap_or_else(...)`. `u64::from` can be used instead of `.into()` as well.

Same thing applies to `secrets_page` below.

kernel/src/sev/secrets_page.rs
```diff
@@ -141,11 +144,36 @@ impl Default for SecretsPage {
 }
 
 static SECRETS_PAGE: RWLock<SecretsPage> = RWLock::new(SecretsPage::new());
+static SECRETS_PAGE_INITIALIZED: AtomicBool = AtomicBool::new(false);
```
00xc, line 147: Perhaps you could use `ImmutAfterInitCell<RWLock<SecretsPage>>`, which will already tell you if it is initialized, instead of introducing a new atomic variable.

msft-jlange: This variable is going to go away in a future PR once the secrets page becomes a heap allocation instead of a static.  At that time, `SECRETS_PAGE` will become `RwLock<Option<&'static SecretsPage>>` which is compatible with the read/write lock functions here, so everything will look clean.  I don't think it's worth over-optimizing the code in this PR since it's going to be replaced so soon.

peterfang (approved):

PR #887, opened 2025-11-26
console: simplify initialization

00xc:
* Use `ImmutAfterInitCell` instead of keeping track of initialization state manually.
* Remove a level of indirection in the `Console` struct.

### DIFF
--- a/kernel/src/serial.rs
+++ b/kernel/src/serial.rs
@@ -4,11 +4,10 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
-use super::io::{IOPort, DEFAULT_IO_DRIVER};
+use super::io::IOPort;
 use crate::{error::SvsmError, io};
 use core::fmt::Debug;
 
-pub const SERIAL_PORT: u16 = 0x3f8;
 const BAUD: u32 = 9600;
 const DLAB: u8 = 0x80;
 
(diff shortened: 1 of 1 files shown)


kernel/src/console.rs
```diff
@@ -26,48 +27,37 @@ impl fmt::Write for Console {
...
+            Self::Serial(serial) => {
+                serial.write(buffer).unwrap();
+            }
         }
     }
 }
 
-static WRITER: SpinLock<Console> = SpinLock::new(Console {
-    writer: &DEFAULT_SERIAL_PORT,
-});
```
peterfang, line 38: Should we just remove `DEFAULT_SERIAL_PORT` altogether? `NativePlatform` already sets up a native console using `DEFAULT_IO_DRIVER`.

00xc: Applied your suggestion, thanks!

kernel/src/serial.rs
```diff
@@ -116,5 +116,3 @@ impl io::Write for SerialPort<'_> {
         Ok(buf.len())
     }
 }
-
-pub static DEFAULT_SERIAL_PORT: SerialPort<'_> = SerialPort::new(&DEFAULT_IO_DRIVER, SERIAL_PORT);
```
stefano-garzarella, line 120: nit: `SERIAL_PORT` constant seems unused now, can we remove it?

joergroedel: Yes, `SERIAL_PORT` can be removed as well.

joergroedel (commented):
Except the tiny comment from Stefano, this looks good to me.

PR #888, opened 2025-11-26
kernel: allocate boot data from kernel heap

msft-jlange:
Now that stage2 is able to allocate data in the kernel's page heap, data used at boot time can be moved from global variables into the heap.  This includes the launch information, the BSP init stack, the CPUID page, and the secrets page.  Data which is not required after boot (the BSP init stack and the launch information) can be freed when it is no longer needed.

### DIFF
--- a/kernel/src/sev/secrets_page.rs
+++ b/kernel/src/sev/secrets_page.rs
@@ -6,16 +6,15 @@
 
 use crate::address::VirtAddr;
 use crate::locking::{RWLock, ReadLockGuard, WriteLockGuard};
-use crate::mm::PAGE_SIZE;
 use crate::protocols::core::CORE_PROTOCOL_VERSION_MAX;
 use crate::sev::vmsa::VMPL_MAX;
 use crate::types::GUEST_VMPL;
-use crate::utils::zero_mem_region;
+use crate::utils::immut_after_init::ImmutAfterInitCell;
 
 extern crate alloc;
 use alloc::boxed::Box;
+use core::ops::{Deref, DerefMut};
 use core::ptr;
-use core::sync::atomic::{AtomicBool, Ordering};
 
 pub const VMPCK_SIZE: usize = 32;
 
@@ -143,37 +142,46 @@ impl Default for SecretsPage {
     }
 }
 
-static SECRETS_PAGE: RWLock<SecretsPage> = RWLock::new(SecretsPage::new());
-static SECRETS_PAGE_INITIALIZED: AtomicBool = AtomicBool::new(false);
+#[derive(Debug)]
+pub struct SecretsPageRef {
+    secrets_page: &'static mut SecretsPage,
+}
 
-pub fn secrets_page() -> Option<ReadLockGuard<'static, SecretsPage>> {
-    let guard = SECRETS_PAGE.lock_read();
-    if SECRETS_PAGE_INITIALIZED.load(Ordering::Relaxed) {
-        Some(guard)
-    } else {
-        None
+// TODO - eliminate the use of Deref/DerefMut and perform this work during
+// lock acquisition by using the map() methods on the lock guards.
+impl Deref for SecretsPageRef {
+    type Target = SecretsPage;
+    fn deref(&self) -> &SecretsPage {
+        self.secrets_page
     }
 }
 
-pub fn secrets_page_mut() -> Option<WriteLockGuard<'static, SecretsPage>> {
-    let guard = SECRETS_PAGE.lock_write();
-    if SECRETS_PAGE_INITIALIZED.load(Ordering::Relaxed) {
-        Some(guard)
-    } else {
-        None
+impl DerefMut for SecretsPageRef {
+    fn deref_mut(&mut self) -> &mut SecretsPage {
+        self.secrets_page
     }
+}
+
+static SECRETS_PAGE: ImmutAfterInitCell<RWLock<SecretsPageRef>> = ImmutAfterInitCell::uninit();
+
+pub fn secrets_page() -> Option<ReadLockGuard<'static, SecretsPageRef>> {
+    SECRETS_PAGE.try_get_inner().ok().map(RWLock::lock_read)
+}
+
+pub fn secrets_page_mut() -> Option<WriteLockGuard<'static, SecretsPageRef>> {
+    SECRETS_PAGE.try_get_inner().ok().map(RWLock::lock_write)
 }
 
 /// # Safety
 /// The caller is required to supply a valid virtual address that points to a
-/// secrets page.
-pub unsafe fn initialize_secrets_page(secrets_page_vaddr: VirtAddr) {
-    let mut guard = SECRETS_PAGE.lock_write();
-    assert!(!SECRETS_PAGE_INITIALIZED.swap(true, Ordering::Relaxed));
+/// secrets page that will remain allocated in the static lifetime.
+pub unsafe fn initialize_secrets_page(addr: VirtAddr) {
     // SAFETY: the caller takes responsibility for the correctness of the
     // virtual address.
     unsafe {
-        guard.copy_from(secrets_page_vaddr);
-        zero_mem_region(secrets_page_vaddr, secrets_page_vaddr + PAGE_SIZE);
+        let secrets_page = addr.aligned_mut::<SecretsPage>().unwrap();
+        SECRETS_PAGE
+            .init(RWLock::new(SecretsPageRef { secrets_page }))
+            .expect("Failed to initialize secrets page");
     }
 }
--- a/kernel/src/stage2.rs
+++ b/kernel/src/stage2.rs
@@ -46,7 +46,7 @@ use svsm::platform::{
     SvsmPlatformCell,
 };
 use svsm::types::{PageSize, PAGE_SIZE};
-use svsm::utils::{round_to_pages, MemoryRegion};
+use svsm::utils::{round_to_pages, zero_mem_region, MemoryRegion};
 
 use release::COCONUT_VERSION;
 
@@ -580,6 +580,18 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
     // Determine whether this platforms uses a secrets pgae.
     let secrets_page = stage2_platform.get_secrets_page(launch_info);
 
+    // Copy the secrets page into the kernel address space if required.
+    let kernel_secrets_page = secrets_page.map(|secrets_addr| {
+        // SAFETY: the secrets page address is assumed to have been correctly
+        // configured in the stage2 image if it is present at all.
+        unsafe {
+            let new_vaddr = copy_page_to_kernel(secrets_addr, &mut kernel_heap)
+                .expect("Failed to copy secrets page");
+            zero_mem_region(secrets_addr, secrets_addr + PAGE_SIZE);
+            new_vaddr
+        }
+    });
+
     // Determine whether use of interrupts on the SVSM should be suppressed.
     // This is required when running SNP under KVM/QEMU.
     let suppress_svsm_interrupts = match platform_type {
@@ -609,7 +621,7 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
         stage2_start: 0x800000u64,
         stage2_end: launch_info.stage2_end as u64,
         cpuid_page: u64::from(kernel_cpuid_page.unwrap_or(VirtAddr::null())),
-        secrets_page: u64::from(secrets_page.unwrap_or(VirtAddr::null())),
+        secrets_page: u64::from(kernel_secrets_page.unwrap_or(VirtAddr::null())),
         stage2_igvm_params_phys_addr: u64::from(launch_info.igvm_params),
         stage2_igvm_params_size: igvm_params.size() as u64,
         igvm_params_virt_addr: u64::from(igvm_vaddr),
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -207,8 +207,8 @@ unsafe fn svsm_start(li: *const KernelLaunchInfo) -> Option<VirtAddr> {
     if launch_info.secrets_page != 0 {
         let secrets_page_virt = VirtAddr::from(launch_info.secrets_page);
 
-        // SAFETY: the secrets page address directly comes from IGVM.  Its address
-        // is trusted if it is non-zero.
+        // SAFETY: the secrets page address was allocated by stage 2 in the kernel
+        // heap and the address is trusted if it is non-zero.
         unsafe {
             initialize_secrets_page(secrets_page_virt);
         }


kernel/src/stage2.rs
```diff
@@ -581,7 +588,16 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
...
+    check_launch_info(&kernel_launch_info);
+
+    // SAFETY: the virtual address of the allocated block is known to be usable
+    // and is known to be uninitialized data which can be filled with the
+    // computed launch information.
+    unsafe {
+        let kernel_launch_block =
+            &mut *launch_info_vaddr.as_mut_ptr::<MaybeUninit<KernelLaunchInfo>>();
+        kernel_launch_block.write(kernel_launch_info);
+    };
```
00xc, line 642: So in this case you did the right thing, but in `load_igvm_params()` there was a bug introduced (see #891) when using the `VirtAddr` returned from the allocator. I think we should have safe methods for the stage 2 heap to avoid all this unsafe dance with `MaybeUninit`.

My suggestion would be:

```rust
fn allocate_with<T>(&mut self, val: T) -> Result<&mut T, SvsmError> {
    let (vaddr, _) = self.allocate(size_of::<T>());
    // SAFETY: the destination is exclusively reserved by the allocator.
    let dst = unsafe { &mut *vaddr.as_mut_ptr::<MaybeUninit<T>>() };
    Ok(dst.write(val))
}
```

And something similar for slices (for `load_igvm_params()`). This would constrain the unsafety within the allocator's internal implementation.

msft-jlange: I completely agree that stage2 needs to be better about using `MaybeUninit`.  I'm going to be completely rewriting the boot code in the next month or two to move most of it out of stage2 and into the IGVM builder, so I'd like to defer this suggestion to that rewrite, where I will plan to solve the problem.

kernel/src/boot_stage2.rs
```diff
@@ -353,8 +353,8 @@ global_asm!(
 global_asm!(
     r#"
         .globl bsp_stack
-        bsp_stack:
+        bsp_stack: .quad 0
         .globl bsp_stack_end
-        bsp_stack_end:
+        bsp_stack_end: .quad 0
```
peterfang: Nit: Since these two symbols are just here to appease the linker and have no actual use, we can just add `.quad 0` to the end and save a little space

```
        .globl bsp_stack
        bsp_stack:
        .globl bsp_stack_end
        bsp_stack_end:
        .quad 0
```

kernel/src/cpu/percpu.rs
```diff
@@ -827,16 +827,34 @@ impl PerCpu {
...
+    pub fn setup_idle_task(&self) -> Result<(), SvsmError> {
+        self.setup_idle_task_internal(KernelThreadStartInfo::new(
+            cpu_idle_loop,
+            self.shared.cpu_index,
+        ))
+    }
+
+    /// # Safety
+    /// The caller is required to pass a start parameter that is appropriate
+    /// for the specified entry point.
```
peterfang: The safety comment looks a bit confusing at first. It doesn't explain why passing a `usize` as the start parameter can be unsafe and requires the reader to look at `svsm_main()` to figure that out. Is there a way to make this clearer?

kernel/src/sev/secrets_page.rs
```diff
@@ -143,37 +141,68 @@ impl Default for SecretsPage {
...
-    } else {
-        None
+    // Clippy wants to turn the match statement below into a map call, but
+    // doing so doesn't wokr without violating either copy or borrow rules,
+    // since the result of the match isn't the contents of the option but
+    // is the lock guard itself.
+    #[allow(clippy::manual_map)]
+    match guard.secrets_page {
+        None => None,
+        Some(_) => Some(guard),
```
peterfang: How about just using `is_some()`?
```suggestion
    if guard.secrets_page.is_some() {
        Some(guard)
    } else {
        None
    }
```

kernel/src/sev/secrets_page.rs
```diff
@@ -143,37 +141,68 @@ impl Default for SecretsPage {
...
```
peterfang: Same as above

kernel/src/stage2.rs
```diff
@@ -285,16 +309,8 @@ unsafe fn map_and_validate(
     unsafe {
         platform.validate_virtual_page_range(vregion, PageValidateOp::Validate)?;
     }
-    valid_bitmap_set_valid_range(paddr, paddr + vregion.len());
```
peterfang, line 288: I'm a bit hesitant about this. While from a stage2/kernel split point of view it makes sense to assume that the SVSM kernel area is properly validated when stage2 is done, in practice we rely on the actual layout in `svsm.lds` to ensure that only designated areas are valid. Just pointing out that this change does take away that accuracy we get from having a valid bitmap in stage2.

kernel/src/stage2.rs
```diff
@@ -527,35 +540,74 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
...
+        cpuid_page.map(|cpuid_addr| {
+            copy_page_to_kernel(cpuid_addr, &mut kernel_heap).expect("Failed to copy CPUID page")
+        })
+    };
+
     // Determine whether this platforms uses a secrets pgae.
     let secrets_page = stage2_platform.get_secrets_page(launch_info);
 
+    // Copy the secrets page into the kernel address space if required.
+    let kernel_secrets_page = if let Some(secrets_page_vaddr) = secrets_page {
```
peterfang: Since `kernel_cpuid_page` is assigned through `map()`, maybe use `map()` here as well?

kernel/src/svsm.rs
```diff
@@ -323,21 +339,33 @@ extern "C" fn svsm_entry(li: &KernelLaunchInfo, vb_addr: usize) -> ! {
...
 }
 
-fn svsm_init() {
+fn free_init_bsp_stack() {
+    // SAFETY: the stack base and limit addresses were initialized when the
+    // kernel was first started.
+    let (stack_base, stack_end) =
+        unsafe { (VirtAddr::from(bsp_stack), VirtAddr::from(bsp_stack_end)) };
+
+    let stack_pages = (stack_end - stack_base).div_ceil(PAGE_SIZE);
```
peterfang: `round_to_pages()` is better optimized here

kernel/src/cpu/percpu.rs
msft-jlange: I will change this function to take a `KernelThreadStartInfo` which would require the caller to supply a safety comment with the information you describe.  That makes the safety comment specific to `svsm_main()` and should make the safety conditions much more obvious.

kernel/src/sev/secrets_page.rs
msft-jlange: I find that `match` is clearer than `if is_some()`/`else`.  @00xc, do you have an opinion?  I believe you've suggested before that `match` is preferable in cases like this.

kernel/src/stage2.rs
msft-jlange: The layout described in the linker file pertains only to the virtual address space, while the validation bitmap being constructed here pertains only to the physical space.  The loader in stage2 guarantees that all physical pages in the kernel range are validated for one reason or another (either as part of loading the kernel ELF or as part of the heap construction).  Can you describe a theoretical case where the loader would somehow leave some physical pages behind?

kernel/src/svsm.rs
msft-jlange: I think I wrote this commit before I saw your similar comments in my last PR.

kernel/src/stage2.rs
peterfang: Ah I see. You're right about the fact that we keep allocating kernel physical pages continuously in stage2, so there shouldn't be page gaps between segments and the kernel heap. As long as this allocation strategy persists, it's safe to assume all of the pages that the kernel should use are validated. Thanks for pointing this out.

msft-jlange: > As long as this allocation strategy persists, it's safe to assume all of the pages that the kernel should use are validated.

This allocation strategy is a key design point of the kernel memory area setup, so I believe we can safely continue to rely on it.

kernel/src/sev/secrets_page.rs
00xc: I don't see a particular reason to do either in this particular case. But the truth is that this pattern looks pretty awkward, since we're returning an `Option` of a guard containing a `SecretsPageRef`, which itself is an `Option` over a reference, so you have 2 unwraps on each access (one implicit via `Deref` and another one explicitly). 

Since you only use `None` to keep track initialization state, you could simply have an `ImmutAfterInitCell<RwLock<SecretsPageRef>>`. You can then remove the `Option` in `SecretsPageRef`, and this function is just:

```rust
pub fn secrets_page() -> Option<ReadLockGuard<'static, SecretsPageRef>> {
    SECRETS_PAGE.try_get_inner().ok().map(RWLock::lock_read)
}
```

...and similarly for `secrets_page_mut()`.

You can also return a `Result` instead of an `Option`, so that if the caller unwraps the proper error is displayed.

kernel/src/sev/secrets_page.rs
```diff
@@ -143,37 +141,68 @@ impl Default for SecretsPage {
...
-    assert!(!SECRETS_PAGE_INITIALIZED.swap(true, Ordering::Relaxed));
+/// secrets page that will remain allocated in the static lifetime.
+pub unsafe fn initialize_secrets_page(addr: VirtAddr) {
     // SAFETY: the caller takes responsibility for the correctness of the
     // virtual address.
-    unsafe {
-        guard.copy_from(secrets_page_vaddr);
-        zero_mem_region(secrets_page_vaddr, secrets_page_vaddr + PAGE_SIZE);
-    }
+    let secrets_page = unsafe { &mut *addr.as_mut_ptr::<SecretsPage>() };
```
00xc: You can do `addr.aligned_mut::<'static, SecretsPage>()`, which will perform a NULL and alignment check for you. You can actually omit the generics in your current code I think, the compiler can guess them.

kernel/src/sev/secrets_page.rs
00xc: I also think we need to change `SecretsPageRef` to store a `NonNull` instead of an actual reference. Even if we prevent write access to read side users, since they would not be able to use `DerefMut`, we would technically be creating two mutable references to the same location, which is undefined behavior.

msft-jlange: > I also think we need to change `SecretsPageRef` to store a `NonNull` instead of an actual reference. Even if we prevent write access to read side users, since they would not be able to use `DerefMut`, we would technically be creating two mutable references to the same location, which is undefined behavior.

I'd like to postpone this work for the moment.  `RWLock::lock_read()` requires `Sync` but `NonNull` is not `Sync` so fixing the alias problem you describe requires some more unsafe glue.  We should consider implementing a wrapper type for `NonNull` that resolve the `Sync` problem because I expect this to come up in more places than just the secrets page, and that goes well beyond the scope of this PR.

00xc: It's preferable to have the unsafe impl of Sync though. Two mutable references to the same location is considered UB even if they are not dereferenced. The unsafe impl is easy to reason about, since SecretsPageRef does not have any interior mutability.

msft-jlange: Yes, but what I'm saying is that fixing this correctly will really involve library code that is usable outside of the secrets page, and that generalization goes beyond the scope of this PR.  That's why I'd rather solve the general problem of mutable aliasing separately.

00xc: The point is that we cannot merge this code because it technically contains undefined behavior. For now we should have a workaround (unsafe impl of Send and Sync for SecretsPageRef), which we can later improve with a more general solution.

msft-jlange: My argument is this cannot be the only place in our code base where we have this problem, so merging this would not make our code base any more unsound than it already is.  If we are already obligated to have a pending work item to fix a persistent broken pattern, then it is much more efficient to fix it holistically then to have a patchwork of one-off fixes that have to be cleaned up over time.

However, I can't find any examples at the moment where we have other dual references like this, so I can't substantiate the "not making it any worse" position.

As I sketch this out, there's a lot I don't like about how this is turning out, so I'm going to plan to put in `TODO` comments as a reminder that this needs to be fixed.  It would be fantastic if you could turn your attention to that at some point after this PR merges so we can make things a lot cleaner.

00xc: I am fine with adding a `TODO`, but let's add it on the unsafe impl, so that it becomes a "clean this up" note instead of a "fix this bug", and you or I can come later and apply a more general solution. Having a cleaner commit history is important, but I also think it is lower on the priority list than having a bug-free codebase. Just because there may be other instances of this issue does not justify introducing a new one.

msft-jlange: The `TODO` is on the definition of `SecretsPageRef` with a note that we shouldn't have to have a type-specific object defined this way in order to satisfy the safety properties.  But yes, the spirit is "clean this up" and not "fix this bug".

00xc: I now think I was misreading this. So having an actual reference is fine, the lock grants a reference *to the stored mutable reference*, which grants all the protection we need.

kernel/src/svsm.rs
```diff
@@ -172,16 +173,21 @@ fn init_cpuid_table(addr: VirtAddr) {
...
+unsafe fn svsm_start(li: *const KernelLaunchInfo, vb_addr: usize) -> Option<VirtAddr> {
+    // SAFETY: the caller guarantees the correctness of the launch info
+    // pointer.
+    let launch_info = unsafe { ScopedRef::<KernelLaunchInfo>::new(li).unwrap() };
     init_platform_type(launch_info.platform_type);
 
     let vb_ptr = core::ptr::NonNull::new(VirtAddr::new(vb_addr).as_mut_ptr::<u64>()).unwrap();
 
-    mapping_info_init(&launch_info);
+    mapping_info_init(launch_info.as_ref());
```
00xc: Since `ScopedRef` implements `Deref` you can still use `&launch_info`. This applies to other uses you have below.

kernel/src/sev/secrets_page.rs
```diff
@@ -143,37 +142,55 @@ impl Default for SecretsPage {
...
+    fn deref_mut(&mut self) -> &mut SecretsPage {
+        let secrets_ptr: *const SecretsPage = self.secrets_page;
+        let secrets_mut = secrets_ptr as *mut SecretsPage;
+        // SAFETY: The secrets page is held within a lock scope that guarantees
+        // that the lock is held exclusively as long as a mutable
+        // SecretsPageRef exists.  Therefore, the reference can safely be
+        // cast from a shared borrow to a mutable borrow here.
+        unsafe { secrets_mut.as_mut().unwrap() }
     }
 }
```
00xc, line 163: You cannot cast a `&T` to a `&mut T`, this is definitely UB. If you run this sample code under Miri you will see an undefined behavior error:

<https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=515a907b68c1f19319f96900749468b2>

I think you had it right originally: store a `&'static mut SecretsPage` in `SecretsPageRef`, and implement `Deref` and `DerefMut` normally. What I got wrong in my original comment is that this never creates two mutable references: the lock grants references `SecretsPageRef`, which itself contains a mutable reference, but it is never duplicated. An immutable reference to a mutable reference is fine.

00xc: Also note the solution you already had (what I propose again) does not have any unsafe code if I'm not mistaken.

msft-jlange: I was worried that storing a `&mut SecretsPage` would be UB if multiple dereferences could occur - my reasoning was that having a `&mut SecretsPage` exist at the same time as multiple `&SecretsPage` dereferences would violate safety.  But it sounds like you're saying that the mere existence of that mutable reference isn't a problem even if it coexists with multiple other references because the lock usage protects everything.

00xc: `&mut SecretsPage` always exists within `SecretsPageRef`, but it will not accessible mutably via a `&SecretsPageRef`, only via `&mut SecretsPageRef`.

2 read guards will, in effect, create 2 immutable references to a *single* mutable reference, which does not grant write access. This is because `&mut T` is not `Copy`, so you cannot move it out of `&(&mut T)`.

kernel/src/sev/secrets_page.rs
```diff
@@ -143,37 +142,49 @@ impl Default for SecretsPage {
...
 
-static SECRETS_PAGE: RWLock<SecretsPage> = RWLock::new(SecretsPage::new());
-static SECRETS_PAGE_INITIALIZED: AtomicBool = AtomicBool::new(false);
+#[derive(Debug)]
+pub struct SecretsPageRef {
+    // This is stored as a shared reference instead of a mutable reference
```
00xc: This comment is now outdated

kernel/src/sev/secrets_page.rs
```diff
@@ -143,37 +142,49 @@ impl Default for SecretsPage {
...
+    secrets_page: &'static mut SecretsPage,
+}
 
-pub fn secrets_page() -> Option<ReadLockGuard<'static, SecretsPage>> {
-    let guard = SECRETS_PAGE.lock_read();
-    if SECRETS_PAGE_INITIALIZED.load(Ordering::Relaxed) {
-        Some(guard)
-    } else {
-        None
+// TODO - eliminate the use of Deref/DerefMut and perform this work during
```
00xc, line 150: I think this comment is not needed anymore, right? There is no unsafe here anyway.

msft-jlange: I would like to eliminate `Deref` and `DerefMut` as the comment suggests because it's better to construct the correct type guard at the time the lock guard is generated than to have to call `Deref`/`DerefMut` every time it's used.  This becomes possible when `map()` is implemented on the lock guards but not before.  So I'd like to keep this `TODO` around to make sure this change can happen once #906 merges.

kernel/src/sev/secrets_page.rs
msft-jlange: That's a helpful explanation, thanks!

peterfang (approved):

joergroedel (approved):

PR #890, opened 2025-11-27
cpu/percpu: (partially) harden against reentrancy issues

00xc:
Harden several fields in the `PerCpu` against potential reentrancy issues by making them `Sync`. This includes protecting against reentrancy in the same call chain and in interrupt context.

This does not fix any actual issues but makes it harder to introduce new ones in the future. This also does not completely remove all potential issues, but covers the easiest parts of the `PerCpu` to replace.

This PR also introduces a new method for `ImmutAfterInitCell` to ensure cleaner error paths during initialization of several parts of the `PerCpu`.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -164,15 +164,15 @@ impl PerCpuAreas {
 
 #[derive(Debug)]
 struct IstStacks {
-    double_fault_stack: Cell<Option<VirtAddr>>,
-    double_fault_shadow_stack: Cell<Option<VirtAddr>>,
+    double_fault_stack: ImmutAfterInitCell<VirtAddr>,
+    double_fault_shadow_stack: ImmutAfterInitCell<VirtAddr>,
 }
 
 impl IstStacks {
     const fn new() -> Self {
         IstStacks {
-            double_fault_stack: Cell::new(None),
-            double_fault_shadow_stack: Cell::new(None),
+            double_fault_stack: ImmutAfterInitCell::uninit(),
+            double_fault_shadow_stack: ImmutAfterInitCell::uninit(),
         }
     }
 }
@@ -586,11 +586,15 @@ impl PerCpu {
     }
 
     pub fn get_top_of_df_stack(&self) -> Option<VirtAddr> {
-        self.ist.double_fault_stack.get()
+        self.ist.double_fault_stack.try_get_inner().ok().copied()
     }
 
     pub fn get_top_of_df_shadow_stack(&self) -> Option<VirtAddr> {
-        self.ist.double_fault_shadow_stack.get()
+        self.ist
+            .double_fault_shadow_stack
+            .try_get_inner()
+            .ok()
+            .copied()
     }
 
     pub fn get_current_stack(&self) -> MemoryRegion<VirtAddr> {
@@ -671,18 +675,16 @@ impl PerCpu {
     }
 
     fn allocate_ist_stacks(&self) -> Result<(), SvsmError> {
-        let double_fault_stack = self.allocate_stack(SVSM_STACK_IST_DF_BASE)?;
-        self.ist.double_fault_stack.set(Some(double_fault_stack));
-
+        self.ist
+            .double_fault_stack
+            .try_init_from_fn(|| self.allocate_stack(SVSM_STACK_IST_DF_BASE))?;
         Ok(())
     }
 
     fn allocate_isst_shadow_stacks(&self) -> Result<(), SvsmError> {
-        let double_fault_shadow_stack =
-            self.allocate_shadow_stack(SVSM_SHADOW_STACK_ISST_DF_BASE, ShadowStackInit::Exception)?;
-        self.ist
-            .double_fault_shadow_stack
-            .set(Some(double_fault_shadow_stack));
+        self.ist.double_fault_shadow_stack.try_init_from_fn(|| {
+            self.allocate_shadow_stack(SVSM_SHADOW_STACK_ISST_DF_BASE, ShadowStackInit::Exception)
+        })?;
 
         Ok(())
     }


kernel/src/utils/immut_after_init.rs
```diff
@@ -140,6 +154,50 @@ impl<T> ImmutAfterInitCell<T> {
...
+    where
+        F: FnOnce() -> Result<T, E>,
+    {
+        self.try_init()?;
+        let val = match f() {
+            Ok(val) => val,
+            Err(e) => {
+                // The function failed, so back off. The state should have
+                // remained as in-progress meanwhile.
+                assert_eq!(
```
msft-jlange: Why is this an assert?  There is no way for the `self.init` field to change on another thread while its value is `IMMUT_INIT_IN_PROGRESS`.  The closure `f` in this case cannot reach the `self.init` field so it cannot cause the value to change, so the assert does not appear to be required for safety.  By comparison, the function `complete_init` does not contain an assert and it has the same properties with respect to `self.init` as this function.

00xc: Precisely to make sure that the invariant is held, otherwise there is surely a bug. The reason `complete_init()` does not have that assert is, well, that I did not write it, as far as I remember. Would you agree to convert it to a `debug_assert`?

msft-jlange: My primary concern is that `complete_init()` and `try_init_from_fn()` have different assert behaviors.  Since they have identical concerns with respect to the correctness of the initialization guard, either both should assert or neither should assert.  As you say, you did not write `complete_init()` (I did) but I would think that in the interest of consistent safety, you would want to ensure that your change is consistent with the rest of the object's behavior - and since you believe in the value of the assert, you would add the assert to `complete_init()` for consistency.

Since the condition here is internal to an object with a small implementation, and since the condition is not subject to influence by users of the object, I think it is more appropriate to use a debug assert, since the invariant can never be violated if `ImmutAfterInitCell` is implemented correctly (which we intend to guarantee).

kernel/src/cpu/percpu.rs
```diff
@@ -532,15 +533,14 @@ impl PerCpu {
 
     /// Sets up the CPU-local GHCB page.
     pub fn setup_ghcb(&self) -> Result<(), SvsmError> {
-        let page = GhcbPage::new()?;
         self.ghcb
-            .set(page)
-            .expect("Attempted to reinitialize the GHCB");
+            .try_init_from_fn(GhcbPage::new)
```
msft-jlange: Can you explain why your version is better than `self.ghcb.init(page).expect(...)`?  I find it really confusing to see a `?` after a call to `expect()`.  The previous version had an explicit error check after the GHCB allocation, which made the potential error cases much more obvious.  Your version may be slightly more concise, but I strongly favor clarity over conciseness.

kernel/src/cpu/percpu.rs
```diff
@@ -651,15 +656,15 @@ impl PerCpu {
     }
 
     fn allocate_init_shadow_stack(&self) -> Result<(), SvsmError> {
-        let init_stack =
-            Some(self.allocate_shadow_stack(SVSM_SHADOW_STACKS_INIT_TASK, ShadowStackInit::Init)?);
-        self.init_shadow_stack.set(init_stack);
+        self.init_shadow_stack.try_init_from_fn(|| {
+            self.allocate_shadow_stack(SVSM_SHADOW_STACKS_INIT_TASK, ShadowStackInit::Init)
+        })??;
```
msft-jlange: The `??` syntax is especially mind-boggling, and I prefer the old syntax for the reasons described above.

kernel/src/cpu/percpu.rs
```diff
@@ -651,15 +656,15 @@ impl PerCpu {
...
         Ok(())
     }
 
     fn allocate_context_switch_stack(&self) -> Result<(), SvsmError> {
-        let cs_stack = Some(self.allocate_stack(SVSM_CONTEXT_SWITCH_STACK)?);
-        self.context_switch_stack.set(cs_stack);
+        self.context_switch_stack
+            .try_init_from_fn(|| self.allocate_stack(SVSM_CONTEXT_SWITCH_STACK))??;
```
msft-jlange: `??` again here.

kernel/src/cpu/percpu.rs
```diff
@@ -672,18 +677,16 @@ impl PerCpu {
     }
 
     fn allocate_ist_stacks(&self) -> Result<(), SvsmError> {
-        let double_fault_stack = self.allocate_stack(SVSM_STACK_IST_DF_BASE)?;
-        self.ist.double_fault_stack.set(Some(double_fault_stack));
-
+        self.ist
+            .double_fault_stack
+            .try_init_from_fn(|| self.allocate_stack(SVSM_STACK_IST_DF_BASE))??;
```
msft-jlange: `??` again here.

kernel/src/cpu/percpu.rs
```diff
@@ -672,18 +677,16 @@ impl PerCpu {
...
 
     fn allocate_isst_shadow_stacks(&self) -> Result<(), SvsmError> {
-        let double_fault_shadow_stack =
-            self.allocate_shadow_stack(SVSM_SHADOW_STACK_ISST_DF_BASE, ShadowStackInit::Exception)?;
-        self.ist
-            .double_fault_shadow_stack
-            .set(Some(double_fault_shadow_stack));
+        self.ist.double_fault_shadow_stack.try_init_from_fn(|| {
+            self.allocate_shadow_stack(SVSM_SHADOW_STACK_ISST_DF_BASE, ShadowStackInit::Exception)
+        })??;
```
msft-jlange: `??` again here.

kernel/src/cpu/percpu.rs
00xc: Because there are two distinct things that can fail here: the allocation of the `GHCB` page allocation and the initialization of the `ImmutAfterInitCell` (because it was somehow already initialized).

If you read the docs for `try_init_from_fn`, it  states that it returns a `Result<Result<&T, E>, ImmutAfterInitError>`. That is, an outer result corresponding to the initialization of the cell, and then an inner for whatever the closure returned. The first `expect()` effectively unwraps the outer `Result`, and the `?` returns the potential error in the inner `Result`.

My change keeps the behavior we already had: if `GhcbPage::new()` fails, we return the error, and if initialization fails, we panic. The only difference is that `GhcbPage::new()` is now called if and only if initialization of the cell would succeed in the first place.

msft-jlange: You've explained how and why it works, which I understand, but you didn't address my question about why this is better and/or more readable.  I still maintain that your change, which is functionally correct and preserves existing behavior, and also more concise, is less understandable.

00xc: As I said, it ensures that `GhcbPage::new()` is only called when initialization would succeed. I have not claimed anywhere that this is more readable.

It prevents potential side effects on error paths. For example, in `PerCpu::setup_hv_doorbell()`, calling `allocate_hv_doorbell_page()` will register the new doorbell page via the GHCB, but if initialization of the cell fails, the page will be dropped without being unregistered. There are several more examples of side effects in PerCpu methods, like `allocate_init_shadow_stack()`, `allocate_context_switch_stack()`, etc.

You may also look at `PerCpu::alloc_svsm_vmsa()`, which is effectively cleaner now, as it was attempting to reproduce the pattern I have introduced.

kernel/src/utils/immut_after_init.rs
```diff
@@ -140,6 +161,55 @@ impl<T> ImmutAfterInitCell<T> {
...
+    ///
+    /// # Returns
+    ///
+    /// * `Ok(Ok(&T))` if both initialization and the callback suceeded.
+    /// * `Ok(Err(E))` if the cell could have been initialized, but the callback
+    ///   failed. The cell is left uninitialized in this case.
+    /// * `Err(ImmutAfterInitErr)`` if the cell was already initialized.
+    pub fn try_init_from_fn<F, E>(&self, f: F) -> ImmutAfterInitResult<Result<&T, E>>
+    where
+        F: FnOnce() -> Result<T, E>,
```
msft-jlange: Part of your change is to provide a conversion from `ImmutAfterInitResult` to `SvsmError`, which is excellent for consistency.  Now that we've done that, can we change this function so it returns `Result<&T, SvsmError>` and to change `F: FnOnce() -> Result<T, SvsmError>`?  If we do this, then any error result from a call to `f()` can be directly passed back as the error result of `try_init_from_fn` without requiring nesting of results, and that would greatly enhance readability.

kernel/src/cpu/percpu.rs
msft-jlange: > As I said, it ensures that `GhcbPage::new()` is only called when initialization would succeed. I have not claimed anywhere that this is more readable.

There are two important points behind this statement.  One is a point you're making: it is advantageous to ensure that no `T` is generated unless initialization is guaranteed to succeed.  The other is a point I'm making: readability is of critical importance for maintainability, and we should work hard to avoid changes that regress readability unless there is no better alternative.

In the case of `setup_ghcb()`, there is no particular problem with allocating a GHCB page and then abandoning it if initialization fails, particularly because we're going to panic if initialization fails and therefore the memory leak is irrelevant.  The same is really true for `setup_hv_doorbell()`; you are correct that the existing code will register the doorbell page before we know whether initialization will succeed, but that becomes moot as soon as initialization fails because we are going to panic immediately.  So while your desire to avoid unnecessary allocation in cases where initialization cannot succeed is a good desire, it is of no practical value in the two cases we're citing here.  There may be value elsewhere, but I haven't looked.

If we think this pattern is one that must be supported, please consider my suggestion posted elsewhere where we can flatten the construction error and the initialization error into the same result.  This loses a small amount of fidelity (such as the case where the construction function may return an ImmutAfterInitError) but dramatically improves readability, and I think the tradeoff is a net positive.

PR #892, opened 2025-12-01
cpu/idt: fix incorrect RBX read/write

00xc:
Fix a likely copy-paste error when reading or writing the RBX register in `X86ExceptionContext`, which caused incorrect instruction emulation and unstable #VC tests due to unexpected register clobbering.


kernel/src/cpu/idt/common.rs
```diff
@@ -149,7 +149,7 @@ impl InsnMachineCtx for X86ExceptionContext {
             Register::Rax => self.regs.rax = val,
             Register::Rdx => self.regs.rdx = val,
             Register::Rcx => self.regs.rcx = val,
-            Register::Rbx => self.regs.rdx = val,
+            Register::Rbx => self.regs.rbx = val,
```
stefano-garzarella, line 152: It seems we have similar issue also on `read_reg()` (line 126 - I'm always asking why GH don't allow us to comment on other lines...)

00xc: Ah nice catch, somehow I managed to miss that one myself.

PR #894, opened 2025-12-02
scripts/launch_guest.sh: use dynamic reduced-phys-bits from sev-caps

YvanY0:
Update launch_guest.sh to dynamically query the reduced-phys-bits value
from sev-caps instead of using a hardcoded value of 1.

The script now calls sev-caps which returns both cbit and
reduced-phys-bits values. This value is then passed to QEMU's sev-snp-guest
object configuration.

This change allows QEMU to use the correct reduced-phys-bits value for
different AMD processor configurations, improving compatibility across
various hardware.

Signed-off-by: Yihuang Yu <yihyu@redhat.com>

### DIFF
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -11,7 +11,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 : "${QEMU:=qemu-system-x86_64}"
 : "${IGVM:=$SCRIPT_DIR/../bin/coconut-qemu.igvm}"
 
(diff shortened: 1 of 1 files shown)


stefano-garzarella (approved):

stefano-garzarella:
@tlendacky oh I see, should QEMU or KVM complain about it?

I just checked and also on my case I have `-object sev-snp-guest,id=cgs0,cbitpos=51,reduced-phys-bits=5` and everything was fine.

PaulYuuu:
OK, thanks @tlendacky for your explanation, I should open an issue to discuss this first, my fault.

It's nice to have doc to highlight this.

PR #897, opened 2025-12-02
 sev: require the debug register virtualization feature

msft-jlange:
The debug registers require protection from host interference and must be swapped during guest entry/exit, so require the SEV "debug register virtualization" feature flag.  All SNP-capable platforms support debug register swap.


tlendacky:
Qemu currently doesn't have a way to set this feature, so you'll likely break SVSM boot without a corresponding Qemu or KVM update to set the bit in the BSP VMSA.

PR #898, opened 2025-12-02
Fix broken proofs due to NO_PAGE and metadata_ptr change in allocator

ziqiaozhou:
The old proof is broken due to inconsistant assumptions between proof code and implementation.

The original allocator proof has an assumption about the valid range of the page number, that is inconsistent with the latest change.

This PR fixed the proof by updating the specification related to when a page number is considered as OOB indicator.

In addition, I fixed 1 implementation bug in 740c443e0d3cec94b0b7a4e72134735bf2ef2f88 where it should check NO_PAGE instead of 0.

Verification results:
https://github.com/ziqiaozhou/svsm/actions/runs/19877910936
https://github.com/ziqiaozhou/svsm/actions/runs/19877914684

Minors:

9e36ebf88cd736c46fa7c4f3f84cb08d32e59c37 fixes a minor problem in Cargo.toml where the verus version was accidentally upgraded to a newer version.

Last commit removes an unnecessary masking, which is not a bug and does not affect verification result and so is optional.

### DIFF
--- a/kernel/src/mm/alloc.verus.rs
+++ b/kernel/src/mm/alloc.verus.rs
@@ -115,7 +115,14 @@ impl HeapMemoryRegion {
         &&& self@.info_ptr_exposed@ == self@.mr_map@.provenance
         &&& self.map() == self@.mr_map@.map
         &&& self.map().wf()
-        &&& self.metadata_addr@ == self@.base_ptr()@.addr
+        &&& self.wf_metadata_addr()
+    }
+
+    // Metadata pointer invariants
+    spec fn wf_metadata_addr(&self) -> bool {
+        &&& self.metadata_addr@ == self@.mr_map@.metadata_addr()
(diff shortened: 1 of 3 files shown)


Cargo.toml
stefano-garzarella, line 1: nit: Is the TODO in the commit https://github.com/coconut-svsm/svsm/pull/898/commits/17a0d9ad551b1462374a56ff4d231c0c9b9ff053 message supposed to be there?

stefano-garzarella (commented):
LGTM too, I'm just unsure if the TODO message is a leftover or not, but really a minor thing. Thanks!

Cargo.toml
ziqiaozhou: That is a reason why I choose not to update the version after 2025-11-16 in this PR.

stefano-garzarella: Ah, I see, thanks!

stefano-garzarella (approved):

PR #899, opened 2025-12-03
task: implement idle halt correctly

msft-jlange:
When there is no task to run on a given CPU, the scheduler must go idle.  This must be done correctly with respect to interrupt dispatch, since receipt of interrupts may change the scheduler state.  If the idle thread commits to going idle but a scheduler interrupt arrives on the idle thread before it goes idle, then the idle thread must be prepared to detect this case and reenter the scheduler.  This change ensures that idle evaluation is done with interrupts disabled so that interrupts that affect scheduler state will always cause wake from halt instead of being lost.

### DIFF
--- a/kernel/src/platform/mod.rs
+++ b/kernel/src/platform/mod.rs
@@ -90,12 +90,7 @@ pub trait SvsmPlatform: Sync {
     /// serves as proof that interrupts have been correctly disabled so that
     /// interrupt state can be correctly manipulated as required by the idle
     /// halt action.
-    fn idle_halt(&self, _irq_guard: &IrqGuard) {
-        // SAFETY: executing HLT in assembly is always safe.
-        unsafe {
-            asm!("sti", "hlt", "cli");
-        }
-    }
+    fn idle_halt(&self, _irq_guard: &IrqGuard);
 
     /// Performs basic early initialization of the runtime environment.
     fn env_setup(&mut self, debug_serial_port: u16, vtom: usize) -> Result<(), SvsmError>;
--- a/kernel/src/cpu/idt/svsm_entry.S
+++ b/kernel/src/cpu/idt/svsm_entry.S
@@ -117,12 +117,20 @@ asm_entry_hv:
 	pushq	%rax
 	pushq	%rbx
 	pushq	%rcx
+	// Fetch RIP from the trap frame.
+	movq	({EXCEP_RIP_OFF} - {EXCEP_RCX_OFF})(%rsp), %rax
 	// Check whether interrupts were enabled at the time of #HV.  If so,
 	// commit to processing all #HV events immediately.
 	testl	${IF}, ({EXCEP_FLAGS_OFF} - {EXCEP_RCX_OFF})(%rsp)
 	jnz	continue_hv
-	// Fetch RIP from the trap frame.
-	movq	({EXCEP_RIP_OFF} - {EXCEP_RCX_OFF})(%rsp), %rax
+	// Check whether the trap RIP is within any code that may require
+	// RIP redirection
+	leaq	hv_redirect_window_start(%rip), %rbx
+	leaq	hv_redirect_window_end(%rip), %rcx
+	cmp	%rbx, %rax
+	jb	hv_no_rip_redirect
+	cmp	%rcx, %rax
+	jae	hv_no_rip_redirect
 	// Check whether the trap RIP is within the guest VMPL return window.
 	leaq	switch_vmpl_window_start(%rip), %rbx
 	leaq	switch_vmpl_window_end(%rip), %rcx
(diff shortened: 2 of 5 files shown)


kernel/src/tdx/tdcall.rs
```diff
@@ -452,18 +452,25 @@ pub fn tdvmcall_wrmsr(msr: u32, value: u64) {
     debug_assert!(tdx_result(ret).is_ok());
 }
 
-pub fn tdvmcall_halt() {
+pub fn tdvmcall_halt(enable_interrupts: bool) {
     let pass_regs = (1 << 10) | (1 << 11) | (1 << 12);
+    let interrupts_blocked: u64 = if enable_interrupts { 0 } else { 1 };
```
00xc: Maybe this makes it more clear that we are negating what the caller passed:

```rust
let interrupts_blocked = u64::from(!enable_interrupts);
```

Although you could invert the meaning of the bool parameter to avoid doing the negation in the first place.

kernel/src/platform/mod.rs
```diff
@@ -343,3 +355,19 @@ pub fn halt() {
         SvsmPlatformType::Tdp => TdpPlatform::halt(),
     }
 }
+
+pub fn idle_halt() {
+    // This function should always be called with interrupts disabled.
+    // Failure to do so does not affect safety, but may affect interrupt
+    // delivery.
+    debug_assert!(!irqs_enabled());
```
00xc: For symmetry, shouldn't we have a similar `debug_assert` in `halt()` above?

kernel/src/tdx/tdcall.rs
msft-jlange: I agree that your version reads better.  I'll make the change.

I will say that the GHCI specification is vague on the expected values here.  Today, it seems to want zero/non-zero, but in the future, it may be the case that a specific bit position is used to indicate that interrupts are blocked, and if that does happen, then we will need an explicit `if` to select the correct bit.  Until then, I like your version better.

kernel/src/platform/mod.rs
msft-jlange: I don't think that's necessary.  The `idle_halt()` method must explicitly enable interrupts in a particular way (in order to ensure that wake from halt works as expected) and will also explicitly disable interrupts before returning to the caller.  Because interrupt enable state is manipulated in this way, we need to assert that interrupts are disabled upon entry to ensure that the behavior is as expected.  The `halt()` method makes no changes to interrupt state at all, so the caller should be allowed to come in with any interrupt posture it wants.

I haven't yet reviewed your PR #901, but I suspect that once a formal termination method is defined, then `halt()` can go away, at which point this becomes moot anyway.  I would certainly like to get rid of generic uses of halt and instead divide those use cases into idle halt (this PR) and forceful shutdown (which is what I think you're trying to accomplish in your PR).

00xc: > The halt() method makes no changes to interrupt state at all, so the caller should be allowed to come in with any interrupt posture it wants.

I agree, but in this PR you precisely changed the docstring for `SvsmPlatform::halt()` to say:
> Halts the system **with interrupts disabled** as required by the platform.

Maybe you can reword that to state what you're saying here (the caller may call into it with any interrupt state).

kernel/src/platform/mod.rs
```diff
@@ -82,6 +84,19 @@ pub trait SvsmPlatform: Sync {
         }
     }
 
+    /// Halts the system with interrupts enabled as required by the platform.
+    fn idle_halt(&self) {
```
00xc: Could we make it so that the caller must pass in the corresponding IRQ guard? This would enforce the IRQ state at compile time and allow removing the `debug_assert`.

kernel/src/tdx/tdcall.rs
```diff
@@ -452,18 +452,25 @@ pub fn tdvmcall_wrmsr(msr: u32, value: u64) {
     debug_assert!(tdx_result(ret).is_ok());
 }
 
-pub fn tdvmcall_halt() {
+pub fn tdvmcall_halt(enable_interrupts: bool) {
```
joergroedel: The `bool` makes the code hard to read at the places where this is used, as it is not obvious what `true` vs. `false` changes in the function. Can we make this a bit more clear, e.g. by changing it to an `enum` with a descriptive name or by having a separate function `tdvmcall_halt_irq_enable()` variant of the function?

kernel/src/platform/mod.rs
msft-jlange: Yes, I like that better.  There should never be a case where we need to execute idle halt without first holding an `IrqGuard`.

joergroedel (approved):

PR #900, opened 2025-12-03
kernel/igvm_params: Allow unaligned guest memory map

peterfang:
The IGVM memory map does not need to be page-aligned or page-sized. Don't make assumptions about alignment when writing to the memory map. Pre-validation of the memory map is still done on page boundaries.

Fixes: dd438cacf ("bootlib/igvm_params: Use byte-level IGVM memory map")

### DIFF
--- a/kernel/src/config.rs
+++ b/kernel/src/config.rs
@@ -15,44 +15,6 @@ use crate::utils::MemoryRegion;
 use alloc::vec::Vec;
 use cpuarch::vmsa::VMSA;
 
-fn check_ovmf_regions(
-    flash_regions: &[MemoryRegion<PhysAddr>],
-    kernel_region: &MemoryRegion<PhysAddr>,
-) {
-    let flash_range = {
-        let one_gib = 1024 * 1024 * 1024usize;
-        let start = PhysAddr::from(3 * one_gib);
-        MemoryRegion::new(start, one_gib)
-    };
-
-    // Sanity-check flash regions.
-    for region in flash_regions.iter() {
-        // Make sure that the regions are between 3GiB and 4GiB.
-        if !region.overlap(&flash_range) {
-            panic!("flash region in unexpected region");
-        }
-
-        // Make sure that no regions overlap with the kernel.
-        if region.overlap(kernel_region) {
-            panic!("flash region overlaps with kernel");
-        }
-    }
-
-    // Make sure that regions don't overlap.
-    for (i, outer) in flash_regions.iter().enumerate() {
-        for inner in flash_regions[..i].iter() {
-            if outer.overlap(inner) {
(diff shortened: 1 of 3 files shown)


luigix25:
The problem is unrelated, it's a grub bug. Using a F43 image it boots successfully!

So this PR fixes the issue!

kernel/src/config.rs
```diff
@@ -16,41 +16,40 @@ use alloc::vec::Vec;
 use cpuarch::vmsa::VMSA;
 
 fn check_ovmf_regions(
```
msft-jlange: As far as I can tell, the inputs to this function are determined statically from the IGVM file and are not based on any configuration detected at runtime (e.g. firmware config).  If the inputs are going to be the same on every boot from a given file, then can we move this validation into the IGVM file builder and not have to do any of this checking at runtime?

kernel/src/igvm_params.rs
```diff
@@ -177,17 +177,21 @@ impl IgvmParams<'_> {
         // If the parameters do not include a guest memory map area, then no
         // work is required.
         let fw_info = &self.igvm_param_block.firmware;
-        if fw_info.memory_map_size == 0 {
+        let mem_map_size = if fw_info.memory_map_size == 0 {
             return Ok(());
-        }
+        } else {
```
msft-jlange: If you're going to return immediately when `fw_info.memory_map_size` is zero, then why use an `else` clause here?  It seems much cleaner just to return when possible and let the rest of the logic proceed in a straight line without having to use if/let or let/if syntax.

peterfang: I agree, I'll fix this

kernel/src/config.rs
peterfang: Good idea. I'll rework this. Thanks

kernel/src/igvm_params.rs
```diff
@@ -180,14 +180,17 @@ impl IgvmParams<'_> {
         if fw_info.memory_map_size == 0 {
             return Ok(());
         }
+        let mem_map_size = fw_info.memory_map_size as usize;
 
         // Map the guest memory map area into the address space.
         let mem_map_gpa = PhysAddr::from(fw_info.memory_map_address as u64);
-        let mem_map_region = MemoryRegion::new(mem_map_gpa, fw_info.memory_map_size as usize);
+        // Guest memory map may not be page-aligned or page-sized.
```
msft-jlange:
```suggestion
        // Guest memory map might not be page-aligned or page-sized.
```
Using the words "may not" implies that it is not allowed (that is how most standards use the term "may not"), but you're saying that the memory map might or might not be aligned, so we should make that a little more clear.

PR #901, opened 2025-12-03
platform: add termination method

00xc:
* Add a proper type to represent the TDVMCALL register bitmap.
* Add TDCALL implementation for ReportFatalError
* Add a platform hook to terminate the guest:
  * Native: simply halt in a loop
  * SNP: request exit through MSR protocol
  * TDX: issue corresponding TDCALL
* Cleanup the exit path for in-SVSM tests.
* Request termination from panic handler in stage2 and main SVSM kernel.

Note that the exit can still be intercepted if the GDB stub is enabled, as the SVSM panic handler calls `debug_break()` before attempting to terminate.

### DIFF
--- a/kernel/src/stage2.rs
+++ b/kernel/src/stage2.rs
@@ -624,7 +624,5 @@ fn panic(info: &PanicInfo<'_>) -> ! {
 
     print_stack(3);
 
-    loop {
-        platform::halt();
-    }
+    platform::terminate();
 }
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -457,11 +457,10 @@ fn panic(info: &PanicInfo<'_>) -> ! {
 
     debug_break();
 
-    // If we are running tests, notify qemu. Otherwise, simply halt.
+    // If we are running tests, notify qemu. Otherwise, simply
+    // terminate the guest.
     #[cfg(all(test, test_in_svsm))]
     crate::testing::exit(crate::testing::QEMUExitValue::Fail);
     #[cfg(any(not(test), not(test_in_svsm)))]
-    loop {
-        svsm::platform::halt();
-    }
+    svsm::platform::terminate();
 }


kernel/src/platform/mod.rs
```diff
@@ -233,6 +234,17 @@ pub trait SvsmPlatform: Sync {
         vaddr: VirtAddr,
         data: &mut [MaybeUninit<u8>],
     ) -> Result<(), SvsmError>;
+
+    /// Terminates the guest.
+    fn terminate() -> !
```
msft-jlange, line 248: Since every platform appears to have a unique implementation of this function, can we move this logic out of the trait object and into the respective platform objects?

msft-jlange: Sorry, what I meant was to remove it from the trait altogether.  Implementations can live in the respective platform objects and not in the trait implementations for those objects.  This gets rid of the `Self: Sized` constraint and makes the code cleaner.  Compare this to what I did with `idle_halt()` in #899.

00xc: I'm not fully convinced. If this is something that we expect all platforms to implement, what is the advantage besides removing those two lines? Conceptually it makes sense to have it as part of the trait, no?

msft-jlange: I see two possible values for placing something into a trait, at least with respect to the platform trait as we have implemented it (i.e. dynamic dispatch only).  One is for methods that can be dispatched through the trait object; this only applies to methods that take `&self`.  The other is for methods that have a default implementation that can be used by more than one specialization of the trait (this is how `halt()` is written today).  What you're building here doesn't fall into either of those categories.  Because the public exposure of termination is `platform::terminate()`, and because that function uses `match` to specialize its behavior based on platform type, there is no chance that we will fail to provide some sort of implementation for each defined platform type.  If we'll get it right regardless of how the per-platform code is written, then why not pick the version that has the cleanest syntax?

00xc: The way I see it, a trait is first and foremost and interface to which different types adhere to. I see it as self-documenting, the fact that `SvsmPlatform` has a `terminate()` method. If you do `cargo doc` and navigate to `SvsmPlatform` you'll see the set of things a "platform" must be able to do.

msft-jlange: I see your point, and I'm remembering that I had originally intended to implement `halt()` as a `&self` method so that it was available through dynamic dispatch just like all other methods, but I couldn't do that because `halt()` is needed from panic flows before the global `SVSM_PLATFORM` is established, and that's how we wound up with the weird `platform::halt()` that we have.  Your termination routine fits the same pattern as that halt case, so it makes sense for it to follow the same pattern.

kernel/src/tdx/tdcall.rs
```diff
@@ -476,6 +477,36 @@ pub fn tdvmcall_halt() {
...
+}
+
+pub fn tdvmcall_report_fatal_error() -> ! {
+    // Default error code (0)
+    let r12 = FatalErrorR12::new();
+
+    // The call is not supposed to return, so guard against
+    // a potential return with an infinite loop
+    loop {
+        // Set RCX to 0, indicating there is no valid data in
```
msft-jlange: This won't work.  You have to include r11 and r12 in the register mask at a minimum or else the host won't see their values.  The GHCI specification also requires that r10 must pass the value zero (indicating that the remainder of the registers conform to the standard GHCI TDVMCALL specification) and thus r10 must also be included in the set of registers to pass.

00xc: I added a proper bitfield struct to represent these registers and fixed the call according to your comment, thanks

kernel/src/platform/mod.rs
```diff
@@ -343,3 +348,12 @@ pub fn halt() {
...
 }
+
+/// Terminates the guest with a platform-specific mechanism
+pub fn terminate() -> ! {
+    match *SVSM_PLATFORM_TYPE {
+        SvsmPlatformType::Native => NativePlatform::terminate(),
+        SvsmPlatformType::Snp => SnpPlatform::terminate(),
+        SvsmPlatformType::Tdp => TdpPlatform::terminate(),
+    }
+}
```
joergroedel, line 369: I think this PR is a good opportunity to also clean up where the code loops around the platform specific termination calls. I think we should consolidate the loop in this function and remove the loops further down the call-stack. Basically let this function be a loop around the platform-specific terminate calls and halt.

kernel/src/tdx/tdcall.rs
```diff
@@ -476,16 +499,47 @@ pub fn tdvmcall_halt() {
...
+    gpa_valid: bool,
+}
+
+pub fn tdvmcall_report_fatal_error() -> ! {
+    let pass_regs = TdxRegs::R10 | TdxRegs::R11 | TdxRegs::R12;
+    // Default error code (0)
+    let r12 = FatalErrorR12::new();
+
+    // The call is not supposed to return, so guard against
+    // a potential return with an infinite loop
```
msft-jlange: I think it would be helpful to point out that the reason there is a potential return is if the untrusted host refuses to cooperate.  Otherwise, it's strange to read a comment that says "this isn't supposed to happen but it might" - that sounds like paranoid programming.  Explaining the role of the untrusted host clarifies why something that isn't supposed to happen might actually happen.

kernel/src/platform/mod.rs
00xc: My concern with adding the loop here is that one may still call e.g. `SnpPlatform::terminate()` directly (for whatever reason), and that function signature will not have a `-> !` return type. We cannot have a `-> !` return signature for the platform methods and loop in the standalone `terminate()` function because clippy will complain (since looping over a function that does not return makes no sense).

I'd rather enforce this at the platform method, to ensure that no callers outside the platform module can do the wrong thing.

00xc: It is also a bit better in terms of self documentation for the `SvsmPlatform` trait to have a `fn terminate() -> !` method.

00xc: On second thought, it might be safer to do this at the lowest level possible (MSR VMGEXIT / TDCALL), that way the caller can never do the wrong thing, since the low level functions are `pub`.

kernel/src/tdx/tdcall.rs
00xc: Done. Also added the equivalent comment for SNP.

kernel/src/platform/mod.rs
00xc: For now leaving I am leaving this as it is, with the exception that I have moved the IRQ masking from `request_termination_msr()` into the platform method for symmetry, but I have left the loops at the lowest level to ensure that no callers can do the wrong thing.

kernel/src/sev/msr_protocol.rs
```diff
@@ -218,14 +217,13 @@ pub unsafe fn invalidate_page_msr(addr: PhysAddr) -> Result<(), GhcbMsrError> {
 pub fn request_termination_msr() -> ! {
     let info: u64 = GHCBMsr::TERM_REQ;
 
-    // Since this processor is destined for a fatal termination, there is
-    // no reason to preserve interrupt state.  Interrupts can be disabled
-    // outright prior to shutdown.
-    raw_irqs_disable();
     // SAFETY: Requesting termination doesn't break memory safety.
     unsafe { write_msr(SEV_GHCB, info) };
```
msft-jlange: Might be slightly better to move the MSR write into the loop.  A non-cooperative host can be non-cooperative for many reasons, but if one of those is a bug where the host fails to observe the GHCB MSR in certain cases, and then corrupts it for some reason, rewriting the MSR on every loop iteration might help work around such a host bug.

msft-jlange (approved):

peterfang (approved):
The TDX part looks good, thanks!

joergroedel (approved):

PR #903, opened 2025-12-04
locking/rwlock: fix Send/Sync on reader/writer locks

msft-jlange:
Fix `RWLock<T>` to be `Sync` only when `T` is `Sync`, and to be `Send` only when `T` is `Send`.  These constraints are required because reader/writer locks permit multiple shared accesses to the inner type.  Consider, for example, a lock `lock: RWLock<Cell<T>>`.  `Cell<T>` is inherently not `Sync`, and if the lock were to blindly implement `Sync`, then it would be possible for multiple threads to have access to `lock` and to call `lock.lock_read().set()` simultaneously to modify the contents of the cell, which would result in a clear conflict.  Because of this, the lock itself can only be `Sync` if its inner type is `Sync`.  The same holds true for `Send`.

### DIFF
--- a/kernel/src/locking/rwlock.rs
+++ b/kernel/src/locking/rwlock.rs
@@ -100,10 +100,16 @@ pub struct RawRWLock<T, I> {
     phantom: PhantomData<fn(I)>,
 }
 
-// SAFETY: All well-formed locks are `Send`.
-unsafe impl<T, I> Send for RawRWLock<T, I> {}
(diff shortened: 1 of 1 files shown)


msft-jlange:
I had forgotten that `lock_read()` is only implemented at all when `T: Send + Sync` so the scenario described above cannot result in a problem because `lock_read()` is unavailable in the above example.

PR #904, opened 2025-12-04
docs: fix aproxy parameter

luigix25:
aproxy parameter `kbs-test` is not available anymore. Update it to `kbs`.


tylerfanelli (approved):
Thanks!

PR #906, opened 2025-12-04
locking/rwlock: add guard mapping methods

00xc:
~~Fully harden the PerCpu structure against undefined behavior when used in a reentrant manner. Our approach before was not correct, since `Cell` and `RefCell` do not guarantee atomicity when keeping track of reads and writes, meaning that a preempted context may observe an inconsistent state, leading to undefined behavior.~~

~~The approach taken is to replace the aforementioned interior mutability types with read-write locks. Thus, whenever an invalid borrow is attempted (e.g. a mutable borrow while an immutable borrow is alive), it can be detected by the same mechanism that is used to prevent the same problem with threads. To avoid deadlocks, we introduce non-blocking methods to `RWLock`, and specific, panicking shorthands to avoid introducing too many calls to `unwrap()`.~~

Add methods to manipulate read and write locking guards, similar to those available for `RefCell`-related guards. This will help in a future change where `RefCells` are replaced by `RWLocks` to keep users mostly unbothered.

This PR does the following:
* Add non-blocking methods to `RWLock`.
* Add `map()`, `map_split()` and `filter_map()` methods to locking guards, similar to those for `Ref` and `RefMut`.
~~* Cleanup some access patterns into `PerCpu` fields.~~
~~* Replace remaining `Cell`s and `RefCell`s in the `PerCpu` with `RWLock`.~~
~~* Cleanup trait bounds on several `PerCpu` fields.~~
~~* Assert at compile time that `PerCpu` is `Sync`.~~

### DIFF
--- a/kernel/src/locking/rwlock.rs
+++ b/kernel/src/locking/rwlock.rs
@@ -384,6 +384,11 @@ impl<T: Send, I: IrqLocking> RawRWLock<T, I> {
         }
     }
 
+    /// Returns a pointer into the underlying data.
+    pub const fn as_ptr(&self) -> *mut T {
+        self.data.get()
+    }
+
     /// This function is used to wait until all writers have finished their
     /// operations and retrieve the current state of the [`RWLock`].
     ///


kernel/src/locking/rwlock.rs
```diff
@@ -52,17 +145,119 @@ pub type ReadLockGuardAnyTpr<'a, T, const TPR: usize> =
 pub struct RawWriteLockGuard<'a, T, I> {
     /// Reference to the associated `AtomicU64` in the [`RWLock`]
     rwlock: &'a AtomicU64,
-    /// Reference to the protected data (mutable)
-    data: &'a mut T,
+    /// Pointer to the protected data
+    data: NonNull<T>,
```
msft-jlange, line 154: You've changed this from a reference to a pointer without explanation.  While I know that the standard implementation of `RefCell` does this as well, it's not at all clear - especially here - why that change is needed.  Can you add a comment for why this should be a pointer instead of a reference?

kernel/src/locking/rwlock.rs
```diff
@@ -17,12 +19,101 @@ use core::sync::atomic::{AtomicU64, Ordering};
 pub struct RawReadLockGuard<'a, T, I> {
     /// Reference to the associated `AtomicU64` in the [`RWLock`]
     rwlock: &'a AtomicU64,
-    /// Reference to the protected data
-    data: &'a T,
+    /// Pointer to the protected data
+    data: NonNull<T>,
```
msft-jlange, line 26: Same as the pointer comment in `RawWriteLockGuard`.

kernel/src/locking/rwlock.rs
```diff
@@ -52,17 +145,119 @@ pub type ReadLockGuardAnyTpr<'a, T, const TPR: usize> =
...
+        )
+    }
+}
+
 /// Implements the behavior of the [`WriteLockGuard`] when it is dropped
 impl<T, I> Drop for RawWriteLockGuard<'_, T, I> {
     fn drop(&mut self) {
-        // There are no readers - safe to just set lock to 0
-        self.rwlock.store(0, Ordering::Release);
+        self.rwlock.fetch_sub(1 << RW_BITS, Ordering::Release);
```
msft-jlange:
```suggestion
        self.rwlock.fetch_sub(compose_val(0, 1), Ordering::Release);
```
This makes the manipulation consistent with how the write lock is acquired.

kernel/src/locking/rwlock.rs
```diff
@@ -286,24 +542,57 @@ impl<T: Send + Sync, I: IrqLocking> RawRWLock<T, I> {
...
+    /// could not be granted at this time, `None` is returned.
+    pub fn try_lock_read(&self) -> Option<RawReadLockGuard<'_, T, I>> {
+        let irq_state = I::acquire_lock();
+        let val = self.check_for_writers()?;
+        let (readers, _) = split_val(val);
+        let new_val = compose_val(readers + 1, 0);
+
+        self.rwlock
+            .compare_exchange(val, new_val, Ordering::Acquire, Ordering::Relaxed)
+            .ok()?;
```
msft-jlange, line 551: This doesn't handle the case where an additional reader arrives after the `check_for_writers()` completes.  The arrival of another reader does not prevent acquisition of the lock and should not return an error.  I believe the only way to solve this is with a CAS loop.

kernel/src/locking/rwlock.rs
```diff
@@ -249,12 +466,51 @@ impl<T: Send, I: IrqLocking> RawRWLock<T, I> {
...
+    pub fn try_lock_write(&self) -> Option<RawWriteLockGuard<'_, T, I>> {
+        let irq_state = I::acquire_lock();
+
+        // Check in a single go if there are any readers or writers
+        let val = self.rwlock.load(Ordering::Relaxed);
+        let (readers, writers) = split_val(val);
+        if readers != 0 || writers != 0 {
+            return None;
+        }
+
```
msft-jlange:
```suggestion
        let val = compose_val(0, 0);
```

There's no reason to read prior to the compare/exchange if we're only going to accept one possible previous value.  The compare/exchange itself will provide the necessary enforcement.

kernel/src/locking/rwlock.rs
00xc: This is briefly explained in the commit message. I can add a comment in the code as well. In short, this is the problem I was facing when using a reference, when implementing `map()`:

```
  --> kernel/src/locking/rwlock.rs:44:9
   |
43 |           let value = f(&*orig);
   |                           ---- `orig` is borrowed here
44 | /         RawReadLockGuard {
45 | |             rwlock,
46 | |             data: value,
47 | |             _irq_state,
48 | |         }
   | |_________^ returns a value referencing data owned by the current function
```

Apparently this is also to prevent `noalias` violations at the LLVM level. I don't fully understand this part, but I assume the standard library developers do.

kernel/src/locking/rwlock.rs
```diff
@@ -52,17 +148,122 @@ pub type ReadLockGuardAnyTpr<'a, T, const TPR: usize> =
...
+        // Move the original IRQ state out of drop.
+        // SAFETY: we are really reading from a reference, so the source
+        // pointer is safe. The original guard is behind `ManuallyDrop`,
+        // so only the copy we just make will invoke drop.
+        let _irq_state = unsafe { core::ptr::read(&orig._irq_state) };
+
+        // Increase the writer count by 1.
+        let rwlock = orig.rwlock;
+        rwlock.fetch_add(1 << RW_BITS, Ordering::Release);
```
msft-jlange:
```suggestion
        rwlock.fetch_add(compose_val(0, 1));
```

kernel/src/locking/rwlock.rs
```diff
@@ -17,12 +19,104 @@ use core::sync::atomic::{AtomicU64, Ordering};
...
+
+        // Move the original IRQ state out of drop.
+        // SAFETY: we are really reading from a reference, so the source
+        // pointer is safe. The original guard is behind `ManuallyDrop`,
+        // so only the copy we just make will invoke drop.
+        let _irq_state = unsafe { core::ptr::read(&orig._irq_state) };
+
+        // Increase the reader count by 1.
+        let rwlock = orig.rwlock;
+        rwlock.fetch_add(compose_val(1, 0), Ordering::Release);
```
msft-jlange, line 100: I like how you are using `compose_val()` to describe the adjustment to the reader count - it is more clear than simply adding 1, which would have the same effect.  The existing implementation of `RawReadLockGuard::drop()` (which you haven't modified) still uses `fetch_sub(1)`, and it would be ideal if you could change it to conform to the use of `compose_val()` that is now used everywhere.

kernel/src/locking/rwlock.rs
```diff
@@ -52,17 +148,122 @@ pub type ReadLockGuardAnyTpr<'a, T, const TPR: usize> =
...
+        Ok(RawWriteLockGuard {
+            rwlock,
+            data: NonNull::from(value),
+            _variance: PhantomData,
+            _irq_state,
+        })
+    }
+}
+
+impl<'a, T, I: IrqLocking> RawWriteLockGuard<'a, T, I> {
```
msft-jlange: I don't understand why `map_split()` requires different bounds on `I` than the other methods.  In practice, because of the way locks are defined, `I` must always be `IrqLocking`, so why not just add that bound to all `impl` blocks and avoid having to create separate blocks for different methods?

If there is a good reason for the reason you've created the split that you have, can you add a comment to that effect?  From reading this code, I have no idea why you would exclude the `IrqBound` for the `map()` and `filter_map()` methods, so I have no way to assess whether future modifications are correctly expressing the bounds that they require.

00xc: As a general rule, in Rust you typically impose as least constraints as possible on impl blocks. In the case of `map_split()` we need to acquire a second guard, so the trait bound needs to be there to call the right method. `map()` does not require anything of the sort, so it does not have it. I can add it for symmetry.

PR #907, opened 2025-12-04
kernel/mm/alloc:  Remove dead code allowance

n-ramacciotti:
This PR introduces some cleanup and a configuration to enable SVSM ALLOCATOR only when needed, avoiding the need to use `#[allow(dead_code)]`.

When running unit tests outside of SVSM (e.g., `make test`), the crate is compiled twice (this can be seen with verbose mode): once with the test flag for running tests, and once without it for rustdoc. In both cases, the target is the host machine. In the second compilation, ALLOCATOR is never used, triggering an unused warning.

The fix conditionally compiles ALLOCATOR only when needed.

### DIFF
--- a/kernel/src/mm/alloc.rs
+++ b/kernel/src/mm/alloc.rs
@@ -2376,7 +2376,7 @@ unsafe impl GlobalAlloc for SvsmAllocator {
     }
 }
 
-#[cfg_attr(any(target_os = "none"), global_allocator)]
+#[cfg_attr(target_os = "none", global_allocator)]
 #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


stefano-garzarella (approved):
@n-ramacciotti thanks!

Not for this PR, but I wonder why fuzzing and testing use `ALLOCATOR`. Maybe tests (outside svsm) could have an `SvsmAllocator` in `TestRootMem` and use that, but anyway, this fix LGTM.

00xc (approved):

luigix25 (approved):

joergroedel (approved):

PR #909, opened 2025-12-05
Makefile: build test kernels via xbuild

00xc:
Allow building test SVSM kernels (`coconut-test-*.igvm`) in `xbuild`, instead of having the corresponding logic embedded in the Makefile, by adding specific JSON recipes.

In order to do this, rewire feature handling in `xbuild`, so that it can consume features even when building a component from the Makefile. The idea is to forward features properly when this call chain happens:

```
FEATURES_TEST=x,y,z make test-in-svsm
    cargo xbuild --feature x,y,z configs/qemu-test-target.json
        FEATURES_TEST=x,y,z make bin/test-kernel.elf
            cargo test (...) --features=x,y,z
```

`bin/test-kernel.elf` is still built from the Makefile for now, since it requires a special command line invocation, and moving that logic into `xbuild` would introduce too much complexity. Thus, we have `xbuild` call back into the Makefile to build it (just like it currently does for the TDX stage 1).

There is also a modification made to the Github pipeline, since it invokes xbuild on every recipe under `configs/*` with the stable toolchain, and the new recipes require the nightly toolchain. This keeps the existing behavior, where only non-testing kernels are built.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,14 +31,14 @@ jobs:
       - name: Install TPM 2.0 Reference Implementation build dependencies
         run: sudo apt install -y build-essential cmake pkg-config
 
-      - name: Build debug image
-        run: cargo xbuild ./configs/*
+      - name: Build (non-test) debug images
+        run: cargo xbuild ./configs/*.json
 
-      - name: Build debug image with all features
-        run: cargo xbuild --all-features ./configs/*
+      - name: Build (non-test) debug images with all features
+        run: cargo xbuild --all-features ./configs/*.json
 
-      - name: Build release image
-        run: cargo xbuild --release ./configs/*
+      - name: Build (non-test) release images
+        run: cargo xbuild --release ./configs/*.json
 
       - name: Run tests
         run: make test


.github/workflows/rust.yml
```diff
@@ -31,14 +31,20 @@ jobs:
...
-
-      - name: Build debug image with all features
-        run: cargo xbuild --all-features ./configs/*
-
-      - name: Build release image
-        run: cargo xbuild --release ./configs/*
+      - name: Build (non-test) debug images
+        run: |
+          shopt -s extglob
+          cargo xbuild ./configs/!(*-test-*.json)
```
stefano-garzarella: What about putting test configurations under `/configs/tests/` ?

configs/test/vanadium-test-target.json
```diff
@@ -0,0 +1,35 @@
+{
+    "igvm": {
+        "vanadium": {
+            "output": "coconut-vanadium.igvm",
```
stefano-garzarella: I guess this should be `coconut-test-vanadium.igvm`

.github/workflows/rust.yml
00xc: Yeah, that seems more robust

Makefile
```diff
@@ -49,7 +50,7 @@ else
 BUILD_FW =
 endif
 
-IGVM_FILES = bin/coconut-qemu.igvm bin/coconut-hyperv.igvm bin/coconut-vanadium.igvm
+IGVM_FILES = bin/coconut-qemu.igvm bin/coconut-hyperv.igvm bin/coconut-vanadium.igvm bin/coconut-test-qemu.igvm bin/coconut-test-hyperv.igvm bin/coconut-test-vanadium.igvm
```
stefano-garzarella: Is this change really needed?

I mean now `make igvm` (and `make all`) will also build igvm for tests, while we didn't before.
We already have `test-igvm` to build them. 

Oh, maybe this is need for `.PHONY`, right?
Maybe better to add a new variable (e.g. `IGVM_TEST_FILES`) for that. WDYT?

00xc: Thanks, done

Issue #448, opened 2024-08-28
Kernel security features


p4zuu:
I think we can start tracking down the security features we could add to the COCONUT kernel to improve its security:

- [ ] KASLR
- [ ] Read-only GDT and IDT
- [ ] SMEP and SMAP
- [ ] Heap hardening

This list aims to be as complete as possible, so please comment or reach me if you want to add or modify elements from the list, or would like to assign yourself to a task :)

Freax13:
Shadow stacks :)

00xc:
We could look into LLVM's [KCFI](https://doc.rust-lang.org/stable/unstable-book/compiler-flags/sanitizer.html#kernelcontrolflowintegrity).

Also we could perhaps consider a KPTI-like mechanism depending on our threat model regarding userspace.

Freax13:
Rust has support for [KASAN](https://doc.rust-lang.org/stable/unstable-book/compiler-flags/sanitizer.html#kerneladdresssanitizer).

p4zuu:
> Rust has (unstable) support for [KASAN](https://doc.rust-lang.org/stable/unstable-book/compiler-flags/sanitizer.html#kerneladdresssanitizer).

KASAN would be very useful in testing env indeed, I'll track it here too. @00xc already had a look into KASAN but struggled a bit with the linker as far as I remember. In any cases, we would still have to write the KASAN handlers ourselves, right?

Freax13:
I have a working implementation of KASAN in [mushroom](https://github.com/Freax13/mushroom/blob/main/tee/kernel/src/sanitize.rs).

> In any cases, we would still have to write the KASAN handlers ourselves, right?

Yes, there's an interface that has to implemented by the kernel.

00xc:
Other than the linking issues I had some boot hangs. I managed to get a working build by disabling stack instrumentation. I did not start the actual implementation though.

Freax13:
We could extend our elf loader to support RELRO.

Freax13:
I'd like to take a stab at shadow stacks.

nagyation:
I am interested in implementing "RELRO support for the ELF loader", Have this already been discussed somewhere or assigned to someone already?

00xc:
> I am interested in implementing "RELRO support for the ELF loader", Have this already been discussed somewhere or assigned to someone already?

I don't think so. Feel free to join the community call on Wednesday and bring up the topic. I'm sure others will have comments.

PR #910, opened 2025-12-05
svsm: Implement KASLR

vmezzela:
This PR implement **KASLR** for the svsm (feature mentioned in #448).

With those changes, `stage2` loads the kernel ELF with a random offset after `SVSM_GLOBAL_BASE`  within a range that does not overlap with other regions in the virtual address space.

With the code as-is, we randomize at most 26 bits [1].  A reorganization of the address space layout would be required if higher randomization needs to be pursued

The last missing piece is to check hardware support for `rdrandr`. Before proceeding further, I am opening this Draft PR to ensure that the maintainers agree the implementation is heading in the right direction.

---
- [1] There are 256GB (2<sup>38</sup>) between `SVSM_GLOBAL_BASE` and `SVSM_GLOBAL_END`, but the code needs to be 4KB aligned, which eats 12bits. Hence 2<sup>26</sup>.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,6 +4,7 @@
 rustflags = [
 	"-C", "force-frame-pointers",
 	"-C", "linker-flavor=ld",
+	"-C", "relocation-model=pie",
 	"--cfg", "aes_force_soft",
 	"--cfg", "polyval_force_soft",
 ]
--- a/elf/src/header.rs
+++ b/elf/src/header.rs
@@ -65,6 +65,7 @@ impl Elf64Hdr {
     const ELFOSABI_GNU: Elf64char = 3;
 
     const ET_EXEC: Elf64Half = 2;
+    const ET_DYN: Elf64Half = 3;
 
     const EM_X86_64: Elf64Half = 62;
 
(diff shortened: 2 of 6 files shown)


msft-jlange:
This conflicts with boot refactoring work that is underway right now.  I would like to revisit this after the boot refactoring is complete.

msft-jlange:
The existing stage2 will be eliminated in the next few months, and the ASLR design here cannot be adapted to the new boot loader model (in which the ELF loader runs completely within the IGVM builder).  We will need a new design that builds relocation tables to be used both for the page tables constructed by the IGVM builder as well as for applying base address relocations.  Once the boot loader changes have merged, we can talk about a new approach.  I have built exactly this type of ASLR model in the past and would be happy to offer advice once we have the framework to support it.

vmezzela:
Hi @msft-jlange, I see that the now boot loader work is completed. Is KASLR still something that would be good to have? If so I'd be happy to implement it.

00xc:
> Hi @msft-jlange, I see that the now boot loader work is completed. Is KASLR still something that would be good to have? If so I'd be happy to implement it.

I think so, all of the bootloader changes have been merged.

PR #911, opened 2025-12-05
sev/vmsa: fix `VMSA::set_tpr()`

00xc:
The method was making a copy of the field and updating the copy, but it never copied the updated value back into the structure, rendering the method a no-op.


msft-jlange (approved):

stefano-garzarella (approved):
Thanks!

Maybe we should consider adding some tests (in another PR) for this e.g.
```rust
#[test]
fn test_vmsa_tpr_roundtrip() {
    let mut vmsa = VMSA::default();
    vmsa.set_tpr(0x80);
    assert_eq!(vmsa.get_tpr(), 0x80);
}
```

PR #912, opened 2025-12-08
Remove duplicated specs for From/Into/Integer from latest verus's vstd lib

ziqiaozhou:
When marking From/Into trait impl as #[verus_verify], verus' vstd lib provides a postcondition to check the trait implementation. Thus, we do not need to maintain the spec for From/Into/Integer traits and can remove annotations #[verus_spec(...)] on `fn from(..)`.

For example, verus forces us to implement trait FromSpecImpl and do the check for the From trait implementation to ensure it follows [the spec defined in vstd lib](https://github.com/verus-lang/verus/blob/main/source/vstd/std_specs/convert.rs).


stefano-garzarella (approved):
LGTM and verification tests pass. Thanks!

PR #913, opened 2025-12-09
userspace: Introduce test in svsm and Allocator

n-ramacciotti:
This PR introduces heap allocation support in userspace and enables running tests in userspace mode within SVSM.

The heap is a single 64KiB static block initialized by `userlib` at startup before calling the `main` function. Atomic state operations ensure that only a single initialization can occur.
The PR implements the `GlobalAlloc` trait by exploiting `AllocBlock` from coconut-alloc, which supports lockless allocations up to 32KiB using atomic operations.

The PR enables running tests in userspace by:
- Adding a custom test runner in `userlib` via the `test_runner` feature, which is included in `userinit` for the `x86_64-unknown-none` target during tests
- Updating the qemu test configuration to build userspace binaries via Makefile  with the `LINK_TEST` environment variable
- Moving filesystem initialization after kernel tests to ensure an empty filesystem for tests that require it.
- Catching test failures in userspace via the exit syscall.

## 
This PR is marked as draft as it depends on the changes from:
- [#901](https://github.com/coconut-svsm/svsm/pull/901) and [#909](https://github.com/coconut-svsm/svsm/pull/909) from SVSM: these simplify test termination and userspace test creation respectively.
- [#4](https://github.com/coconut-svsm/coconut-alloc/pull/4) from coconut-alloc: this removes an unstable feature not available in the current toolchain (1.86.0) and adds a new utility function required in this PR

### DIFF
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -34,7 +34,7 @@ use svsm::cpu::sse::sse_init;
 use svsm::debug::gdbstub::svsm_gdbstub::{debug_break, gdbstub_start};
 use svsm::debug::stacktrace::print_stack;
 use svsm::enable_shadow_stacks;
-use svsm::fs::{initialize_fs, populate_ram_fs};
+use svsm::fs::initialize_fs;
 use svsm::hyperv::hyperv_setup;
 use svsm::igvm_params::IgvmBox;
 use svsm::kernel_region::new_kernel_region;
@@ -348,9 +348,6 @@ fn svsm_init() {
 
     init_memory_map(&config, &LAUNCH_INFO).expect("Failed to init guest memory map");
 
-    populate_ram_fs(LAUNCH_INFO.kernel_fs_start, LAUNCH_INFO.kernel_fs_end)
-        .expect("Failed to unpack FS archive");
-
     init_capabilities();
 
     let cpus = config.load_cpu_info().expect("Failed to load ACPI tables");
@@ -402,8 +399,12 @@ fn svsm_init() {
     #[cfg(not(test))]
     {
         use svsm::fs::opendir;
+        use svsm::fs::populate_ram_fs;
         use svsm::requests::request_loop_start;
         use svsm::task::exec_user;
+
+        populate_ram_fs(LAUNCH_INFO.kernel_fs_start, LAUNCH_INFO.kernel_fs_end)
+            .expect("Failed to unpack FS archive");
 
         match exec_user("/init", opendir("/").expect("Failed to find FS root")) {
             Ok(_) => (),
@@ -431,6 +432,26 @@ pub fn svsm_main(_context: usize) {
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
     crate::test_main();
+
+    #[cfg(all(test, test_in_svsm))]
+    {
+        use crate::testing::QEMUExitValue;
+        use svsm::fs::opendir;
+        use svsm::fs::populate_ram_fs;
+        use svsm::task::exec_user;
+
+        populate_ram_fs(LAUNCH_INFO.kernel_fs_start, LAUNCH_INFO.kernel_fs_end)
+            .expect("Failed to unpack FS archive");
+
+        match exec_user("/init", opendir("/").expect("Failed to find FS root")) {
+            Ok(_) => (),
+            Err(e) => log::info!("Failed to launch /init: {e:?}"),
+        }
+
+        log::info!("All tests passed!");
+
+        crate::testing::exit(QEMUExitValue::Success);
+    }
 }
 
 #[panic_handler]
--- a/kernel/src/syscall/class0.rs
+++ b/kernel/src/syscall/class0.rs
@@ -14,6 +14,13 @@ use core::ffi::c_char;
 use syscall::SysCallError;
 
 pub fn sys_exit(exit_code: u32) -> ! {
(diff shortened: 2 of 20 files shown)


kernel/src/testing.rs
```diff
@@ -81,9 +81,9 @@ pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
         (test_case.testfn.0)();
     }
 
-    info!("All tests passed!");
+    info!("Kernel tests passed!");
 
-    exit(QEMUExitValue::Success);
+    //exit(QEMUExitValue::Success);
```
luigix25: Probably a leftover from development

kernel/src/svsm.rs
```diff
@@ -348,9 +348,6 @@ fn svsm_init() {
 
     init_memory_map(&config, &LAUNCH_INFO).expect("Failed to init guest memory map");
 
-    populate_ram_fs(LAUNCH_INFO.kernel_fs_start, LAUNCH_INFO.kernel_fs_end)
```
luigix25, line 351: Maybe I'm missing something, but I don't understand the reason behind this change.

user/lib/src/alloc.rs
```diff
@@ -1,9 +1,17 @@
 use coconut_alloc::AllocBlock;
-use core::alloc::{GlobalAlloc, Layout};
+#[cfg(any(test, all(not(test), target_os = "none")))]
+use core::alloc::GlobalAlloc;
+use core::alloc::Layout;
+#[cfg(any(test, all(not(test), target_os = "none")))]
```
luigix25: what about
```rust
#[cfg(any(test, all(not(test), target_os = "none")))]
use core::{alloc::GlobalAlloc,ptr}
```
?

(untested)

kernel/src/syscall/class0.rs
```diff
@@ -14,6 +14,13 @@ use core::ffi::c_char;
 use syscall::SysCallError;
 
 pub fn sys_exit(exit_code: u32) -> ! {
+    #[cfg(all(test, test_in_svsm))]
+    {
+        if exit_code != 0 {
```
luigix25, line 19: Can the exit code be anything other than 0? IIUC, a test fails because it hits an assert, which causes the system to panic.

luigix25 (commented):
quick review, thanks for working on this!

kernel/src/testing.rs
n-ramacciotti: Yes, exactly.

user/lib/src/alloc.rs
n-ramacciotti: Yes, it makes sense. Thanks

kernel/src/syscall/class0.rs
n-ramacciotti: When a test fails, the panic handler in `userlib` calls `exit(!0)`

kernel/src/svsm.rs
n-ramacciotti: Some tests need the filesystem to be empty, for example  `create_dir()` in `kernel/src/fs/filesystem.rs`. Having the filesystem already populated would make those tests fail. That’s why the call to `populate_ram_fs` has been moved to a different place.

kernel/src/syscall/class0.rs
luigix25: Oh, you are right!

kernel/src/svsm.rs
stefano-garzarella: > Some tests need the filesystem to be empty, for example `create_dir()` in `kernel/src/fs/filesystem.rs`. Having the filesystem already populated would make those tests fail. That’s why the call to `populate_ram_fs` has been moved to a different place.

This is exactly the things that should be mentioned in the commit description. Please pay attention on describing "why" a patch/change is needed.

luigix25: > Some tests need the filesystem to be empty, for example `create_dir()` in `kernel/src/fs/filesystem.rs`. Having the filesystem already populated would make those tests fail. That’s why the call to `populate_ram_fs` has been moved to a different place.

Mmh, could this issue be unrelated to your PR? because currently FS is populated before test are run.

update:
I _think_ the RamFs has always been empty during tests until now

n-ramacciotti: > update: I _think_ the RamFs has always been empty during tests until now

Yes, that’s correct.

kernel/src/syscall/class0.rs
luigix25: I was thinking if we should move this code to the panic handler, just like we do for the kernel.

luigix25:
@n-ramacciotti Don't forget to add an SPDX header to new files.

kernel/src/syscall/class0.rs
n-ramacciotti: I'm not sure this is doable. The code interacts with qemu and halts the platform, using `SVSM_PLATFORM`, which is not available in `userlib`.

00xc:
Maybe you could separate out this into (a) the introduction of a userspace allocator and (b) userspace test enablement? It's easier for review, and it seems to me that (a) is not really dependent on #901 and #909

n-ramacciotti:
Closing this PR in favor of the other two PRs that resulted from the split  ([#926](https://github.com/coconut-svsm/svsm/pull/926) and [#927](https://github.com/coconut-svsm/svsm/pull/927)).
The third PR will be added later.

PR #915, opened 2025-12-13
sev/utils: Use iterator combinators for pvalidate and RMP revoke paths

v-thakkar:
Switch pvalidate_range_4k() and rmp_revoke_guest_access() to iterator- based try_for_each for cleaner, idiomatic Rust.

No functional changes.

### DIFF
--- a/kernel/src/sev/utils.rs
+++ b/kernel/src/sev/utils.rs
@@ -39,6 +39,16 @@ impl SevSnpError {
             | Self::FAIL_SIZEMISMATCH(ret) => *ret,
         }
     }
+
+    fn with_addr(self, addr: VirtAddr) -> Self {
+        let vaddr = addr.bits() as u64;
+        match self {
+            Self::FAIL_INPUT(_) => Self::FAIL_INPUT(vaddr),
+            Self::FAIL_PERMISSION(_) => Self::FAIL_PERMISSION(vaddr),
+            Self::FAIL_SIZEMISMATCH(_) => Self::FAIL_SIZEMISMATCH(vaddr),
+            Self::FAIL_UNCHANGED(_) => Self::FAIL_UNCHANGED(vaddr),
+        }
+    }
 }
 
 impl fmt::Display for SevSnpError {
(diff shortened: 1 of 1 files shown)


00xc:
> In the case of validation, a failure to validate always indicates either a security attack from the host or a security-critical error in guest-side logic. In both cases, it is most appropriate to panic once a failure is detected so that the security posture does not erode further.

When servicing `SVSM_CORE_PVALIDATE` guest requests via the core protocol, the spec says we are to return an error if `PVALIDATE` or `RMPADJUST` fail, so I do not think it is true that panicking is always the right choice.

> Beyond that, reverting the validation state (which is not possible on TDX at all) is no more guaranteed to succeed than the operation that failed in the first place, so reversion can never be relied upon.

We can detect if the reversion fails. This is what we do in `SharedBox` actually: on error we first try to revert the state, and if that fails then we panic, which I think is a good middle ground.

I think at least we should let the caller know how far the operation got through if there was an error, and let them take the right action based on this information.

kernel/src/sev/utils.rs
```diff
@@ -154,9 +156,9 @@ pub unsafe fn pvalidate(
 
     match ret {
         0 if changed => Ok(()),
-        0 if !changed => Err(SevSnpError::FAIL_UNCHANGED(0x10).into()),
-        1 => Err(SevSnpError::FAIL_INPUT(ret).into()),
-        6 => Err(SevSnpError::FAIL_SIZEMISMATCH(ret).into()),
+        0 if !changed => Err(SevSnpError::FAIL_UNCHANGED(vaddr.bits() as u64).into()),
+        1 => Err(SevSnpError::FAIL_INPUT(vaddr.bits() as u64).into()),
+        6 => Err(SevSnpError::FAIL_SIZEMISMATCH(vaddr.bits() as u64).into()),
```
00xc: Thinking more about it, this will mess up how the SVSM handles guest requests. When a core protocol request is made, these values are used to construct the returned error to the guest. For example, when servicing a `SVSM_CORE_PVALIDATE` request, the value contained in each of these variants, in the way it is passed back to the guest, is interpreted as the value in `RAX` after the `PVALIDATE` call. So returning the failing address needs to be done at the caller of this function, not here.

v-thakkar: Ah, I see. That makes sense. I've pushed the changes at the callers of pvalidate now. Thanks

joergroedel: Works now, thanks.

PR #916, opened 2025-12-15
cpu/idt: clean up vector handling

msft-jlange:
Remove exception handlers for exceptions that can never be raised, and ensure that exceptions generated from user mode will terminate the user task instead of causing a system panic.

### DIFF
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -14,8 +14,8 @@ use super::super::x86::apic_eoi;
 use super::common::{
     user_mode, IdtEntry, IdtEventType, PageFaultError, AC_VECTOR, BP_VECTOR, BR_VECTOR, CP_VECTOR,
     DB_VECTOR, DE_VECTOR, DF_VECTOR, GP_VECTOR, HV_VECTOR, IDT, INT_INJ_VECTOR, IPI_VECTOR,
-    MCE_VECTOR, MF_VECTOR, NMI_VECTOR, NM_VECTOR, NP_VECTOR, OF_VECTOR, PF_VECTOR, SS_VECTOR,
-    SX_VECTOR, TS_VECTOR, UD_VECTOR, VC_VECTOR, VE_VECTOR, XF_VECTOR,
+    MCE_VECTOR, MF_VECTOR, NMI_VECTOR, OF_VECTOR, PF_VECTOR, SS_VECTOR, UD_VECTOR, VC_VECTOR,
+    VE_VECTOR, XF_VECTOR,
 };
 use crate::address::VirtAddr;
 use crate::cpu::irq_state::{raw_get_tpr, raw_set_tpr, tpr_from_vector};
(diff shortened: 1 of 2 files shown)


kernel/src/cpu/idt/svsm.rs
```diff
@@ -174,6 +164,19 @@ pub fn idt_init() -> Result<(), SvsmError> {
...
 }
 
+// General task termination handler
+#[no_mangle]
+extern "C" fn ex_handler_terminate(ctx: &mut X86ExceptionContext, vector: usize) {
+    if user_mode(ctx) {
+        // Terminate the current task if it is a user-mode task.
+        terminate();
+    } else {
+        // Kernel-mode tasks should never cause a termination class execption,
```
00xc:
```suggestion
        // Kernel-mode tasks should never cause a termination class exception,
```

kernel/src/cpu/idt/svsm.rs
```diff
@@ -174,6 +174,19 @@ pub fn idt_init() -> Result<(), SvsmError> {
     GLOBAL_IDT.init(idt).map_err(|_| SvsmError::PlatformInit)
 }
 
+// General task termination handler
+#[no_mangle]
+extern "C" fn ex_handler_terminate(ctx: &mut X86ExceptionContext, vector: usize) {
+    if user_mode(ctx) {
+        // Terminate the current task if it is a user-mode task.
+        terminate();
```
joergroedel, line 179: It is certainly worth printing a message here to let the user/developer know what happened.

joergroedel: Ideally including the user-mode RIP which caused the exception.

kernel/src/cpu/idt/svsm.rs
```diff
@@ -93,11 +88,7 @@ fn init_idt_exceptions(idt: &mut IDT<'_>) {
     idt.set_entry(OF_VECTOR, IdtEntry::entry(asm_entry_of));
     idt.set_entry(BR_VECTOR, IdtEntry::entry(asm_entry_br));
     idt.set_entry(UD_VECTOR, IdtEntry::entry(asm_entry_ud));
-    idt.set_entry(NM_VECTOR, IdtEntry::entry(asm_entry_nm));
     idt.set_entry(DF_VECTOR, IdtEntry::entry(asm_entry_df));
-    idt.set_entry(TS_VECTOR, IdtEntry::entry(asm_entry_ts));
-    idt.set_entry(NP_VECTOR, IdtEntry::entry(asm_entry_np));
-    idt.set_entry(SS_VECTOR, IdtEntry::entry(asm_entry_ss));
```
peterfang, line 100: Just curious - Why remove #SS here and make it fall back to #GP? I think a bad/buggy user process could trigger this?

msft-jlange: You're right.  I had assumed that a user #SS could simply converted to #GP and presented that way, but it turns out #SS is a contributory exception so if #SS raises #GP, then we have a double fault, which would be bad.  I'll put #SS back as a terminating event.

PR #917, opened 2025-12-15
platform: mark all page validation operations as unsafe

00xc:
Page validation is always an unsafe operation, as it can easily break memory safety. We already consider functions like `SvsmPlatform::validate_virtual_page_range()` unsafe, so complete this by marking all related functions as unsafe as well, and propagate the unsafety across callers and callees.

### DIFF
--- a/kernel/src/svsm_paging.rs
+++ b/kernel/src/svsm_paging.rs
@@ -68,6 +68,11 @@ pub fn init_page_table(
     Ok(pgtable)
 }
 
+/// Invalidates the given memory region using a platform-specific method.
+///
+/// While this cannot generate undefined behavior, the caller must be careful
+/// not to invalidate memory that will be used later, as accesing it will lead
+/// to an immediate panic.
 fn invalidate_boot_memory_region(
     platform: &dyn SvsmPlatform,
     config: &SvsmConfig<'_>,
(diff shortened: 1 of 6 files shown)


00xc:
> Based on the code that you are modifying here and the comments you are adding, it appears that the fundamental safety issue is not based on page validation, but based on the fact that arbitrary accesses to physical memory are unsafe because there can be no knowledge in the language about what references may exist to that physical memory before it is modified. This is certainly true regardless of whether the validation state of memory is changing. Would it be better to retitle the PR and to possibly expand its scope to ensure that all operations on raw physical addresses are marked `unsafe`?

I think this needs to be evaluated on a case-by-case basis. For example, we can check that a physical address belongs to the guest (`valid_phys_address()`), so mapping and accessing that memory via `GuestPtr` is safe, but changing its state is not, as we could trigger a double page validation, which is an issue in SEV-SNP.

I would like to keep the scope of this PR as it is, and we can open further PRs for other areas and evaluate more properly.

kernel/src/svsm_paging.rs
```diff
@@ -64,7 +64,11 @@ pub fn init_page_table(
     Ok(pgtable)
 }
 
-fn invalidate_boot_memory_region(
+/// # Safety
+///
+/// The caller must ensure that the given physical memory is not used again.
+/// See safety considerations for [`SvsmPlatform::validate_physical_page_range`].
+unsafe fn invalidate_boot_memory_region(
```
msft-jlange: It's not clear to me why this is unsafe.  As I wrote earlier, invalidating memory can only introduce faults and cannot cause data corruption.

PR #922, opened 2025-12-17
mm/guestmem: fix `GuestPtr` trait bounds & add `try_read()`

00xc:
Currently `GuestPtr` requires that the underlying `T` is `Copy`, but that does not provide the right safety guarantees. Instead, do not require anything for the type itself, but mandate `FromBytes` for reads, and `IntoBytes` for writes. These are stronger guarantees than T, since they prevent undefined behavior if an invalid representation of T is read, and prevent writing out padding data, potentially leaking information.

This change includes deriving `FromBytes` and `IntoBytes` for the required types, as well as adding explicit padding if needed.

### DIFF
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -486,6 +486,21 @@ impl<T> UserPtr<T> {
         unsafe { self.guest_ptr.read() }
     }
 
+    /// Attempts to read the `T` behind the pointer, verifying it has a valid
+    /// representation in the process (see [`GuestPtr::try_read`]).
+    #[inline]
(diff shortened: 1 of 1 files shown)


00xc:
> It's easy to envision a future where we define a protocol that encodes enums, bitfields, or other data that is not guaranteed to be convertible from an arbitrary byte stream. Should we consider bounding guest memory with `TryFromBytes` instead of `FromBytes` so that we can correctly handle such types when reading guest memory?

Yes. I intentionally kept this PR small since I want it merged to rebase some other work on top, but this is doable. I just have not found the need in the codebase yet.

joergroedel (approved):

PR #925, opened 2025-12-19
virtio: centralize mmio slot handling

luigix25:
Currently the (only) virtio driver does the slot probing by itself. This is problematic when there are multiple drivers in the system.

Introduce `MMIOSlot`. This structure is in charge of probing all the slots and enumerate them.

As of now, this only supports `virtio-mmio` slots with _one_ device per type.

### DIFF
--- a/kernel/src/block/virtio_blk.rs
+++ b/kernel/src/block/virtio_blk.rs
@@ -6,12 +6,14 @@
 // Author: Oliver Steffen <osteffen@redhat.com>
 
 use super::api::BlockDriver;
-use crate::address::PhysAddr;
-use crate::block::BlockDeviceError;
+use crate::block::{BlockDeviceError, BLOCK_DEVICE};
 use crate::error::SvsmError;
 use crate::types::PAGE_SIZE;
 use crate::virtio::devices::VirtIOBlkDevice;
+use crate::virtio::mmio::{MmioSlot, MmioSlots};
 use virtio_drivers::device::blk::SECTOR_SIZE;
+use virtio_drivers::transport::DeviceType::Block;
+
 extern crate alloc;
 use alloc::boxed::Box;
 pub struct VirtIOBlkDriver(Box<VirtIOBlkDevice>);
@@ -23,8 +25,8 @@ impl core::fmt::Debug for VirtIOBlkDriver {
 }
 
 impl VirtIOBlkDriver {
-    pub fn new(mmio_base: PhysAddr) -> Result<Self, SvsmError> {
-        Ok(VirtIOBlkDriver(VirtIOBlkDevice::new(mmio_base)?))
+    pub fn new(slot: MmioSlot) -> Result<Self, SvsmError> {
+        Ok(VirtIOBlkDriver(VirtIOBlkDevice::new(slot)?))
     }
 }
 
@@ -68,28 +70,43 @@ impl BlockDriver for VirtIOBlkDriver {
     }
 }
 
+/// Initializes the global block device subsystem with a VirtIO block driver.
+///
+/// This function searches for a virtio-blk device in the MMIO slots list.
+/// If discovered, the first virtio-blk device will be initialized and
+/// registered as the global block device.
+/// **Only one block device is supported**
+///
+/// # Arguments
+///
+/// * `slots` - The virtio MMIO slots list
+///
+/// # Returns
+///
+/// * Returns Ok() if:
+///     * The driver is correctly initialized
+///     * No virtio-block devices are found
+/// * Returns an error if:
+///     * The driver initialization fails
+///     * The global block device has already been initialized
+pub fn initialize_block(slots: &mut MmioSlots) -> Result<(), SvsmError> {
+    let Some(slot) = slots.pop_slot(Block) else {
+        return Ok(());
+    };
+
+    let driver = VirtIOBlkDriver::new(slot)?;
+
+    BLOCK_DEVICE.init(Box::new(driver))?;
+
+    Ok(())
+}
+
 #[cfg(all(test, test_in_svsm))]
 mod tests {
-    use crate::{
-        address::PhysAddr, fw_cfg::FwCfg, platform::SVSM_PLATFORM, testutils::has_test_iorequests,
-    };
+    use crate::testutils::has_test_iorequests;
     use core::cmp::min;
     extern crate alloc;
     use super::*;
-
-    /// Find the first virtio-blk device in the hardware-info list
-    fn get_blk_device() -> VirtIOBlkDriver {
-        let cfg = FwCfg::new(SVSM_PLATFORM.get_io_port());
-
-        let dev = cfg
-            .get_virtio_mmio_addresses()
-            .unwrap_or_default()
-            .iter()
-            .find_map(|a| VirtIOBlkDriver::new(PhysAddr::from(*a)).ok())
-            .expect("No virtio-blk device found");
-
-        dev
-    }
 
     /// Get the sha256 sum of the disk image from the host (see `scripts/test-in-svsm.sh`)
     fn get_image_hash_from_host() -> Option<[u8; 32]> {
@@ -147,7 +164,10 @@ mod tests {
         use alloc::vec;
         use sha2::{Digest, Sha256};
         assert!(sectors_at_once > 0);
-        let blk = get_blk_device();
+
+        let blk = BLOCK_DEVICE
+            .try_get_inner()
+            .expect("Block driver not found");
 
         let expected_hash = get_image_hash_from_host().unwrap();
 
@@ -211,7 +231,9 @@ mod tests {
 
         assert!(sectors_at_once > 0);
 
-        let blk = get_blk_device();
+        let blk = BLOCK_DEVICE
+            .try_get_inner()
+            .expect("Block driver not found");
 
         let n_sectors = blk.size() / SECTOR_SIZE;
         let mut buffer = vec![0u8; sectors_at_once * SECTOR_SIZE];
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -35,6 +35,8 @@ use svsm::cpu::sse::sse_init;
 use svsm::debug::gdbstub::svsm_gdbstub::{debug_break, gdbstub_start};
 use svsm::debug::stacktrace::print_stack;
 use svsm::enable_shadow_stacks;
+#[cfg(feature = "virtio-drivers")]
+use svsm::error::SvsmError;
 use svsm::fs::{initialize_fs, populate_ram_fs};
 use svsm::hyperv::hyperv_setup;
 use svsm::igvm_params::IgvmBox;
@@ -55,6 +57,8 @@ use svsm::svsm_paging::{
 use svsm::task::{schedule_init, start_kernel_task, KernelThreadStartInfo};
 use svsm::types::PAGE_SIZE;
 use svsm::utils::{round_to_pages, MemoryRegion, ScopedRef};
+#[cfg(feature = "virtio-drivers")]
+use svsm::virtio::probe_mmio_slots;
 #[cfg(all(feature = "vtpm", not(test)))]
 use svsm::vtpm::vtpm_init;
 
@@ -149,6 +153,25 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
         PhysAddr::from(launch_info.heap_area_phys_start),
     );
     init_kernel_mapping_info(kernel_mapping, None);
+}
+
+/// Probes for VirtIO MMIO devices and initializes them.
+///
+/// # Returns
+///
+/// Returns Ok if initialization is successful or no virtio devices are found
+/// Returns an error when a virtio device is found but its driver initialization fails.
+#[cfg(feature = "virtio-drivers")]
+fn initialize_virtio_mmio() -> Result<(), SvsmError> {
+    let mut slots = probe_mmio_slots();
+
+    #[cfg(feature = "block")]
+    {
+        use svsm::block::virtio_blk::initialize_block;
+        initialize_block(&mut slots)?;
+    }
+
+    Ok(())
 }
 
 /// # Panics
@@ -410,6 +433,9 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
     vtpm_init().expect("vTPM failed to initialize");
 
     virt_log_usage();
+
+    #[cfg(feature = "virtio-drivers")]
+    initialize_virtio_mmio().expect("Failed to initialize virtio-mmio drivers");
 
     if let Err(e) = SVSM_PLATFORM.launch_fw(&config) {
         panic!("Failed to launch FW: {e:?}");
--- a/kernel/src/virtio/devices.rs
+++ b/kernel/src/virtio/devices.rs
@@ -8,18 +8,14 @@
 use super::hal::*;
 extern crate alloc;
 use alloc::boxed::Box;
-use core::ptr::NonNull;
 use virtio_drivers::device::blk::VirtIOBlk;
-use virtio_drivers::transport::mmio::{MmioError, MmioTransport};
-use virtio_drivers::transport::{DeviceType, Transport};
-use virtio_drivers::PAGE_SIZE;
+use virtio_drivers::transport::mmio::MmioTransport;
 
 use super::error::*;
-use crate::address::PhysAddr;
 use crate::error::SvsmError;
 use crate::locking::SpinLock;
-use crate::mm::global_memory::{map_global_range_4k_shared, GlobalRangeGuard};
-use crate::mm::pagetable::PTEntryFlags;
+use crate::mm::global_memory::GlobalRangeGuard;
+use crate::virtio::mmio::MmioSlot;
 
 pub struct VirtIOBlkDevice {
     pub device: SpinLock<VirtIOBlk<SvsmHal, MmioTransport<SvsmHal>>>,
@@ -33,33 +29,12 @@ impl core::fmt::Debug for VirtIOBlkDevice {
 }
 
 impl VirtIOBlkDevice {
-    pub fn new(mmio_base: PhysAddr) -> Result<Box<Self>, SvsmError> {
-        virtio_init();
-
-        let mem = map_global_range_4k_shared(mmio_base, PAGE_SIZE, PTEntryFlags::data())?;
-
-        // Not expected to fail, because mem exists.
-        let header = NonNull::new(mem.addr().as_mut_ptr()).unwrap();
-
-        // SAFETY: `header` is the MMIO config area; we have to trust the content is valid.
-        let transport = unsafe {
-            // TODO: Use more detailed error types ?
-            MmioTransport::<SvsmHal>::new(header).map_err(|e| match e {
-                MmioError::BadMagic(_) => VirtioError::InvalidDevice,
-                MmioError::UnsupportedVersion(_) => VirtioError::InvalidDevice,
-                MmioError::ZeroDeviceId => VirtioError::InvalidDevice,
-            })?
-        };
-
-        if transport.device_type() != DeviceType::Block {
-            return Err(VirtioError::InvalidDeviceType)?;
-        }
-
-        let blk = VirtIOBlk::new(transport).map_err(|_| VirtioError::InvalidDevice)?;
+    pub fn new(slot: MmioSlot) -> Result<Box<Self>, SvsmError> {
+        let blk = VirtIOBlk::new(slot.transport).map_err(|_| VirtioError::InvalidDevice)?;
 
         Ok(Box::new(VirtIOBlkDevice {
             device: SpinLock::new(blk),
-            _mmio_space: mem,
+            _mmio_space: slot.mmio_range,
         }))
     }
 }


kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,116 @@
...
+pub unsafe fn probe_mmio_slots() -> Option<MMIOSlot> {
+    virtio_init();
+
+    let cfg = FwCfg::new(SVSM_PLATFORM.get_io_port());
+
+    let Ok(dev) = cfg.get_virtio_mmio_addresses() else {
+        return None;
+    };
+
+    let mut slots: Vec<MMIOSlotType> = Vec::<MMIOSlotType>::with_capacity(dev.len());
```
00xc: The generic `MMIOSlotType` does not need to be specified here, at the very least not twice. You probably had to do this before you had the loop below written. Once you add the `.push()` call the compiler is able to figure out the type, so you can simply do:

```rust
let mut slots = Vec::with_capacity(dev.len());
```

Or if you prefer something more verbose, leave just one of the type hints.

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// The caller must ensure:
+/// * Single-threaded execution
+/// * Called exactly once during boot
+pub unsafe fn probe_mmio_slots() -> Option<MMIOSlot> {
+    virtio_init();
+
+    let cfg = FwCfg::new(SVSM_PLATFORM.get_io_port());
+
+    let Ok(dev) = cfg.get_virtio_mmio_addresses() else {
```
00xc, line 54: Or simply:

```rust
let dev = cfg.get_virtio_mmio_addresses().ok()?;
```

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,116 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `virtio_type` - The virtio device type to search for
+    ///
+    /// # Returns
+    ///
+    /// * `Some(MMIOSlotType)` - The first matching slot
+    /// * `None` - No slot matching the requested device type was found
+    pub fn get_slot(&mut self, virtio_type: DeviceType) -> Option<MMIOSlotType> {
```
00xc: Maybe it would be better to name it `pop_slot()` if it consumes the item.

kernel/src/block/mod.rs
```diff
@@ -10,3 +10,11 @@ pub mod error;
 pub mod virtio_blk;
 
 pub use error::BlockDeviceError;
+
+extern crate alloc;
+use crate::{block::api::BlockDriver, utils::immut_after_init::ImmutAfterInitCell};
+use alloc::boxed::Box;
+
+// This variable is only used by tests for now.
```
00xc: Then we can have the dead code allowance gated behind the test feature gate.

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,116 @@
...
+};
+
+#[derive(Debug)]
+pub struct MMIOSlotType {
+    pub mmio_range: GlobalRangeGuard,
+    pub transport: MmioTransport<SvsmHal>,
+}
+
+#[derive(Debug)]
+pub struct MMIOSlot {
```
00xc: Maybe better named `MmioSlots` (plural)

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,116 @@
...
+    address::PhysAddr,
+    fw_cfg::FwCfg,
+    mm::{map_global_range_4k_shared, pagetable::PTEntryFlags, GlobalRangeGuard},
+    platform::SVSM_PLATFORM,
+    types::PAGE_SIZE,
+    virtio::hal::{virtio_init, SvsmHal},
+};
+
+#[derive(Debug)]
+pub struct MMIOSlotType {
```
00xc: Maybe better named `MmioSlot`

kernel/src/virtio/slot.rs
luigix25: thanks!

kernel/src/virtio/slot.rs
luigix25: right, it's much cleaner!

kernel/src/virtio/slot.rs
luigix25: makes sense!

kernel/src/block/mod.rs
luigix25: I agree with you, but unfortunately it's not so easy. I'd have to put quite a few `#[cfg(all(test, test_in_svsm))]` around the code because of unused imports. I don't know if there are better ways.

I was also wondering if we should add a new `block` feature.

kernel/src/svsm.rs
```diff
@@ -151,6 +153,27 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
...
+        log::warn!("MmioSlots: No slots found");
+        return;
+    };
+
+    #[cfg(feature = "block")]
+    {
+        use svsm::block::virtio_blk::initialize_block;
+
+        let Some(block_slot) = slots.pop_slot(virtio_drivers::transport::DeviceType::Block) else {
+            log::warn!("MmioSlots: no virtio-blk device found");
```
stefano-garzarella: why a warning here?
Is it always expected to have a virtio-blk device?

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        let Ok(mem) = map_global_range_4k_shared(phys_addr, PAGE_SIZE, PTEntryFlags::data()) else {
+            log::warn!("MmioSlots: Failed to map MMIO region at {:x}", addr);
+            continue;
+        };
+
+        // Not expected to fail, because mem exists.
+        let header = NonNull::new(mem.addr().as_mut_ptr()).unwrap();
+
+        // SAFETY: `header` is the MMIO config area; we have to trust the content is valid.
```
stefano-garzarella: this is the safety requirement:
```
    /// # Safety
    /// `header` must point to a properly aligned valid VirtIO MMIO region, which must remain valid
    /// for the lifetime of the transport that is returned.
```

So I don't understand this comment

kernel/src/block/mod.rs
```diff
@@ -10,3 +10,9 @@ pub mod error;
 pub mod virtio_blk;
 
 pub use error::BlockDeviceError;
+
+extern crate alloc;
+use crate::{block::api::BlockDriver, utils::immut_after_init::ImmutAfterInitCell};
+use alloc::boxed::Box;
+
+static BLOCK_DEVICE: ImmutAfterInitCell<Box<dyn BlockDriver>> = ImmutAfterInitCell::uninit();
```
stefano-garzarella, line 19: Please add a comment here to say that we currently support only 1 block device.

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,112 @@
...
+/// # Safety
+///
+/// Must be called on the bootstrap processor (BSP) before other CPUs are started.
+/// This function is not thread-safe and performs global device enumeration that
+/// must only happen once during system initialization.
+///
+/// The caller must ensure:
+/// * Single-threaded execution
+/// * Called exactly once during boot
+pub unsafe fn probe_mmio_slots() -> Option<MmioSlots> {
```
stefano-garzarella: Why an Option? Can be just an empty Vec ?

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        let slot_type = MmioSlot {
+            mmio_range: mem,
+            transport,
+        };
+
+        slots.push(slot_type);
+    }
+
+    Some(MmioSlots { slots })
```
stefano-garzarella: Indeed, it seems we always return something, I can't see where we return None (and it makes sense to me)

kernel/src/block/virtio_blk.rs
```diff
@@ -147,11 +142,14 @@ mod tests {
...
-        let blk = get_blk_device();
+
+        let blk_driver = BLOCK_DEVICE
+            .try_get_inner()
+            .expect("Block driver not found");
 
         let expected_hash = get_image_hash_from_host().unwrap();
 
-        let n_sectors = blk.size() / SECTOR_SIZE;
+        let n_sectors = blk_driver.size() / SECTOR_SIZE;
```
stefano-garzarella: I'd avoid this renaming in this commit to keep it as small as possible.

kernel/src/svsm.rs
```diff
@@ -151,6 +153,27 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
     init_kernel_mapping_info(kernel_mapping, None);
 }
 
+#[cfg(feature = "virtio-drivers")]
+fn initialize_virtio_mmio() {
+    // SAFETY: Called during svsm_init() that runs on the BSP
+    let Some(mut slots) = (unsafe { probe_mmio_slots() }) else {
+        log::warn!("MmioSlots: No slots found");
```
stefano-garzarella: Again, why a warning here?

Also, I'm not sure if we never return `None`.

kernel/src/svsm.rs
stefano-garzarella: Maybe instead of warning, I'll print info when we find them.

kernel/src/virtio/mmio.rs
```diff
@@ -0,0 +1,112 @@
...
+        };
+
+        slots.push(slot_type);
+    }
+
+    Some(MmioSlots { slots })
+}
+
+impl MmioSlots {
+    /// Retrieves and removes the first available MMIO slot matching the specified device type.
```
stefano-garzarella, line 99: From https://doc.rust-lang.org/std/vec/struct.Vec.html#method.pop_if

> Removes and returns the last element from a vector if the predicate returns true ...

So, are you sure this comment is true?
Maybe a test will be nice to have.

kernel/src/virtio/slot.rs
luigix25: the only case when it returns none is when `get_virtio_mmio_addresses` fails.

stefano-garzarella: So, we are returning `None` if `get_virtio_mmio_addresses` fails, but `Some(empty vector)` if we are not able to map none of them?

This is really confusing IMO and at least should be documented.
Why not returning an error? (like `get_virtio_mmio_addresses` )

luigix25: > So, we are returning `None` if `get_virtio_mmio_addresses` fails, but `Some(empty vector)` is we are not able to map none of them?

correct

> 
> This is really confusing IMO and at least should be documented. Why not returning an error? (like `get_virtio_mmio_addresses` )

Your idea is to return the probed slots or an error in all other cases (failure or no slots found)?

kernel/src/virtio/mmio.rs
luigix25: oops, good catch!

kernel/src/virtio/slot.rs
stefano-garzarella: I'd just like to be consistent. I don't see any value on using `Option` with `Vec` or at least I can't see it in this use case.

kernel/src/svsm.rs
luigix25: > Maybe instead of warning, I'll print info when we find them.

this already happens in `probe_mmio_slots` when a slot is successfully probed. 
I'll switch the log to info from debug.

I'll also drop the warning

kernel/src/virtio/slot.rs
luigix25: This is pre-existent. I took that code (and comment) from `kernel/src/virtio/devices.rs`.

stefano-garzarella: Okay, but I still don't understand it. IMO we should improve it and match the safety requirement.

kernel/src/virtio/mmio.rs
```diff
@@ -0,0 +1,121 @@
...
+        };
+
+        // Not expected to fail, because mem exists.
+        let header = NonNull::new(mem.addr().as_mut_ptr()).unwrap();
+
+        // SAFETY: we are trusting QEMU to provide us an address that points to a valid virtio-MMIO region.
+        // `map_global_range_4k_shared` guarantees us proper address alignment.
+        // The memory region has the same lifetime of the MmioSlot structure which will be consumed by the driver.
+        let Ok(transport) = (unsafe { MmioTransport::<SvsmHal>::new(header) }) else {
+            log::debug!("MmioSlots: {:x} empty", addr);
```
stefano-garzarella, line 77: Why at line 67 we print a warning but here is a debug message?

kernel/src/virtio/mmio.rs
```diff
@@ -0,0 +1,121 @@
...
+/// This function queries the fw_cfg interface to discover virtio-MMIO device
+/// addresses and maps their MMIO regions.
+///
+/// # Returns
+///
+/// * `MmioSlots` - Collection of discovered VirtIO devices.
+///
+/// # Safety
+///
+/// Must be called on the bootstrap processor (BSP) before other CPUs are started.
```
stefano-garzarella: I don't understand why this is unsafe if called twice. What will happen?

Usually safety comments are about memory-safety guarantees (https://doc.rust-lang.org/reference/unsafety.html)

kernel/src/virtio/slot.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        let Ok(mem) = map_global_range_4k_shared(phys_addr, PAGE_SIZE, PTEntryFlags::data()) else {
+            log::warn!("MmioSlots: Failed to map MMIO region at {:x}", addr);
+            continue;
+        };
+
+        // Not expected to fail, because mem exists.
+        let header = NonNull::new(mem.addr().as_mut_ptr()).unwrap();
+
+        // SAFETY: we are trusting QEMU to provide us an address that points to a valid virtio-MMIO region.
```
stefano-garzarella: Why we need to trust QEMU? I mean it seems we do checks, so IIUC we just need to be sure that the area is aligned (and we check it as you wrote) and the lifetime (you wrote also that), so I don't get in which term we should trust QEMU, what it can do wrongly?

kernel/src/block/api.rs
stefano-garzarella, line 1: > Send + Sync traits are required for concurrency handling.

This is not clear to me, mainly IIUC this is required by ImmutAfterInitCell()  no?

kernel/src/virtio/mmio.rs
luigix25: well, it would read again _all_ the slots from the FwCfg interface. So the same memory areas would be mapped again. I'm not sure what would happen if you try to instantiate a second driver on the same virtio-device, but we agree that this must not happen.

kernel/src/virtio/slot.rs
luigix25: right, we have to trust the addresses it gives us, but if the address is "wrong", we handle it. I'll modify the comment.

kernel/src/block/api.rs
luigix25: yes exactly!

kernel/src/virtio/mmio.rs
stefano-garzarella: Okay, but what Rust memory-safety guarantees we are violating if that happen?
This is what a Safety section should talk about.

kernel/src/block/api.rs
stefano-garzarella: Okay, so why not just writing that `ImmutAfterInitCell` requires that?

kernel/src/virtio/mmio.rs
luigix25: I guess none. I'll remove the `unsafe`.
Should I leave the comments? (without the safety keyword)

kernel/src/virtio/mmio.rs
luigix25: if we fail to map an MMIO region there must be something wrong about page tables, therefore a warning.
Knowing that an mmio-slot slot is empty could be useful for debugging. 
I can drop that if you think it's unnecessary

stefano-garzarella: I was thinking if having a working too here makes sense, since IIUC the VMM is telling us that there is a virtio-mmio device there, but we are not able to recognize it, so IMO makes sense to have a warning too, maybe a bit more descriptive.

kernel/src/virtio/mmio.rs
stefano-garzarella, line 1: What about calling this file `mmio.rs`?

kernel/src/block/mod.rs
stefano-garzarella: I think we should also document this in Documentation/ somewhere, but we can do later maybe with #806

kernel/src/virtio/mmio.rs
stefano-garzarella: Really, I don't see any issue on running probe_mmio_slots() concurrently and also twice or more, so I don't understand what you are worried about for both "must":

> /// The caller must ensure:
> /// * Single-threaded execution
> /// * Called exactly once during boot

The main actions done here are:
- `map_global_range_4k_shared()` will map the same phys_addr on multiple virtual address (yeah, it's not great, but what will be the issue?)
- `MmioTransport::<SvsmHal>::new()` does only reads, so again what will be the issue?

I agree that doesn't make sense to call it twice, but for sure nothing wrong can happen, or I may miss something.

stefano-garzarella: okay, talking with @osteffenrh I get better the original issue:

> it hangs if trying to probe/initialize a transport on the same mmio config window where already a driver is running on

IIUC this doesn't have anything to do with rust memory-safety, but more with qemu, virtio, or the architecture in general. So, I agree we should mention that in the comment, but explaining the original issue, otherwise is not really clear what we are preventing.
In addition, if you prefer, we can enforce it in the code and panic or error on subsequent calls, but it's not a strong request on my side, just an idea if you're worried about that.

kernel/src/virtio/mmio.rs
luigix25: note that, at the moment, QEMU advertises _all_ slots, regardless if they are used by a device or not. Things will change with the device tree.

kernel/src/virtio/mmio.rs
luigix25: yep this is not about memory safety, it's about not messing up the system. I'll write a comment about that.

kernel/src/virtio/mmio.rs
stefano-garzarella: aaaa, okay, maybe worth a comment there, it's not clear at all that empty is expected, but yeah in that case should not be a warning at all, debug is fine.

kernel/src/block/virtio_blk.rs
```diff
@@ -68,29 +68,24 @@ impl BlockDriver for VirtIOBlkDriver {
...
+    let Ok(driver) = VirtIOBlkDriver::new(slot) else {
+        log::error!("Failed to initialize virtio-blk driver");
+        return;
+    };
+
+    BLOCK_DEVICE
+        .init(Box::new(driver))
+        .expect("virtio-block driver already initialized");
+}
+
```
stefano-garzarella: What about returning an error here? (+ adding documentation especially on when it can fail)

In general IMO we should try to avoid as much as possible to panic in internal modules and demand it to the caller.

kernel/src/block/virtio_blk.rs
stefano-garzarella, line 1: > virtio: integrate MmioSlot infrastructure with block device subsystem
>
> Introduce `initialize_block` that uses the MmioSlot structure to handle
> virtio-blk initialization.
> 
> Note that this happens in the bootstrap processor, so no
> concurrency is possible.
> 
> Switch to the same initialization function in virtio-blk tests.
> 
> Signed-off-by: Luigi Leonardi <leonardi@redhat.com>

IMO this commit message should also mention that now svsm is going to panic if built with `virtio-drivers` and `block` features enabled, since we are expecting at least one device exposed by the VMM.

Also please re-check it about the concurrency, etc.

kernel/src/virtio/devices.rs
```diff
@@ -7,19 +7,14 @@
...
-use crate::address::PhysAddr;
 use crate::error::SvsmError;
 use crate::locking::SpinLock;
-use crate::mm::global_memory::{map_global_range_4k_shared, GlobalRangeGuard};
-use crate::mm::pagetable::PTEntryFlags;
+use crate::mm::global_memory::GlobalRangeGuard;
+use crate::virtio::mmio::MmioSlot;
+use alloc::boxed::Box;
+use virtio_drivers::device::blk::VirtIOBlk;
+use virtio_drivers::transport::mmio::MmioTransport;
```
stefano-garzarella: why moving this include? 

Please keep the diff minimal and also keep the internal crate + external crates separated like it was before.

kernel/src/virtio/mmio.rs
```diff
@@ -0,0 +1,116 @@
...
+/// addresses and maps their MMIO regions.
+///
+/// # Returns
+///
+/// * `MmioSlots` - Collection of discovered VirtIO devices.
+///
+/// Must be called on the bootstrap processor (BSP) before other CPUs are started.
+/// This function is not thread-safe and performs global device enumeration that
+/// must only happen once during system initialization. Otherwise, if the same slot
+/// is probed and/or used by two different drivers, this can cause unexpected behaviours.
```
stefano-garzarella: This part is confusing, what about replacing in this way?

```suggestion
/// Probes and enumerates all virtio-MMIO devices available in the system.
///
/// This function queries the fw_cfg interface to discover virtio-MMIO device
/// addresses and maps their MMIO regions.
///
/// # Usage
///
/// This function is typically called once during early system initialization
/// to discover all available virtio devices. Each slot in the returned collection
/// should be consumed exactly once by calling [`MmioSlots::pop_slot`], as each
/// slot represents exclusive ownership of a device's MMIO region.
///
/// # Returns
///
/// Returns an [`MmioSlots`] collection containing all discovered virtio-MMIO devices.
/// Returns an empty collection if no devices are found or if the fw_cfg interface
/// is unavailable.
```

kernel/src/svsm.rs
```diff
@@ -151,6 +153,22 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
...
+fn initialize_virtio_mmio() {
+    let mut slots = probe_mmio_slots();
+
+    #[cfg(feature = "block")]
+    {
+        use svsm::block::virtio_blk::initialize_block;
+
+        let Some(block_slot) = slots.pop_slot(virtio_drivers::transport::DeviceType::Block) else {
+            return;
+        };
```
stefano-garzarella: I'd move this into the `initialize_block()`, so if in the future we will handle multiple block devices, we can do everything there, in the block module.

kernel/src/block/virtio_blk.rs
luigix25: >svsm is going to panic if built with virtio-drivers and block features enabled, since we are expecting at least one >device exposed by the VMM.

SVSM will not panic if the VMM exposes no block devices or slots. `initialize_block` (that can cause a panic) will never run if no block slots are found.

>Also please re-check it about the concurrency, etc.

will do

kernel/src/virtio/mmio.rs
luigix25: thanks!

kernel/src/block/virtio_blk.rs
stefano-garzarella: > > svsm is going to panic if built with virtio-drivers and block features enabled, since we are expecting at least one >device exposed by the VMM.
> 
> SVSM will not panic if the VMM exposes no block devices or slots. `initialize_block` will never run if no block slots are found.

Okay, I see it now, this behavior should be documented IMO.

> 
> > Also please re-check it about the concurrency, etc.
> 
> will do

thanks

luigix25:
v4:
- addressed @stefano-garzarella comments.
- Better error handling
- Removed unnecessary changes
- Better documentation
- block driver now consumes the slot

kernel/src/block/virtio_blk.rs
```diff
@@ -68,29 +68,43 @@ impl BlockDriver for VirtIOBlkDriver {
     }
 }
 
+/// Initializes the global block device subsystem with a VirtIO block driver.
+///
+/// This function initializes a VirtIO block driver, using the first virtio-blk
+/// device from the MMIO slots list, and registers it as the global block device.
+/// **Only one block device is supported**
```
stefano-garzarella:
```suggestion
/// This function searches for a virtio-blk device in the MMIO slots list.
/// If discovered, the first virtio-blk device will be initialized and 
/// registered as the global block device.
/// **Only one block device is supported**
```

kernel/src/block/virtio_blk.rs
```diff
@@ -68,29 +68,43 @@ impl BlockDriver for VirtIOBlkDriver {
...
+/// # Returns
+///
+/// * Returns Ok() if:
+///     * The driver is correctly initialized
+///     * No virtio-block devices are found
+/// * Returns an error if:
+///     * The driver initialization fails
+///     * The global block device has already been initialized
+pub fn initialize_block(slots: &mut MmioSlots) -> Result<(), SvsmError> {
+    let Some(slot) = slots.pop_slot(virtio_drivers::transport::DeviceType::Block) else {
```
stefano-garzarella: nit: we can add `use virtio_drivers::transport::DeviceType::Block` on top like we do for other `virtio_drivers` stuff

stefano-garzarella (approved):

PR #926, opened 2025-12-19
userspace: Introduce test in svsm

n-ramacciotti:
The PR enables running tests in svsm userspace by:
- Adding a custom test runner in `userlib` via the `test_runner` feature, which ~~is included in `userinit`~~ can be included in userspace binaries ~~(like in `userinit` in the `[testing]` commits)~~ for the `x86_64-unknown-none` target during tests.
- The `test-runner` features enables `test` crate in `userlib` only for `x86_64-unknown-none` target. For other targets, the feature is included but the optional `test` dependency is ignored.
- ~~Updating the ~~qemu~~ test configurations to build userspace binaries via Makefile  with the `LINK_TEST` environment variable.~~
- ~~user binaries are compiled in a secondary `makefile` in the `user` folder. These rules are called by the root makefile and can have `FEATURES_TEST` AND `USERMODULE_FEATURES_TEST` customizable (see https://github.com/coconut-svsm/svsm/pull/926#issuecomment-4712818365)~~
- ~~User-space binaries can be compiled via `Makefile`, updating the tests configurations to include `test_in_svsm` configuration with `LINK_TEST` (like in `[testing]` commits)~~
- ~~Moving filesystem initialization after kernel tests to ensure an empty filesystem for tests that require it.~~
- Changing some filesystem tests to account for an already populated filesystem.
- Catching test failures in userspace via the exit syscall.

## 
~~This PR is marked as draft as it depends on the changes from:~~
- ~~[#909](https://github.com/coconut-svsm/svsm/pull/909): this simplify test creation with xbuild~~

These changes were originally included in [#913](https://github.com/coconut-svsm/svsm/pull/913), which also introduced a userspace allocator, but as suggested, I split them into multiple PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2143,6 +2143,7 @@ dependencies = [
 name = "userinit"
 version = "0.1.0"
 dependencies = [
+ "test",
  "userlib",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2151,6 +2151,7 @@ name = "userlib"
 version = "0.1.0"
 dependencies = [
  "syscall",
+ "test",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2330,6 +2330,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6f5e870be6c3b371b77fe0ee0bafb859fa4964b4404c27de1d380043c4dda20"
 
 [[package]]
+name = "uapi_tester"
+version = "0.1.0"
+dependencies = [
+ "test",
+ "userlib",
+]
+
+[[package]]
 name = "ucs2"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2427,6 +2427,7 @@ name = "userlib"
 version = "0.1.0"
 dependencies = [
  "syscall",
+ "test",
 ]
 
 [[package]]
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ bin/svsm-kernel.elf: bin
 	cargo build --package svsm --bin svsm ${CARGO_ARGS} ${SVSM_ARGS} --target=x86_64-unknown-none
 	objcopy -O elf64-x86-64 ${OBJCOPY_ELF_ARGS} ${SVSM_KERNEL_ELF} $@
 
-TEST_IN_SVSM_MODULES = svsm
+TEST_IN_SVSM_MODULES = svsm userinit
 TEST_IN_SVSM_TARGETS = $(TEST_IN_SVSM_MODULES:%=bin/test-%.elf)
 # Root of SVSM
 MAKEFILE_DIR = $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,13 @@ endif
 
 FEATURES_TEST ?= vtpm,virtio-drivers,block
 SVSM_ARGS_TEST += --no-default-features
+MODULE_ARGS_TEST += --no-default-features
 ifneq ($(FEATURES_TEST),)
 SVSM_ARGS_TEST += --features ${FEATURES_TEST}
 XBUILD_ARGS_TEST += --feature ${FEATURES_TEST}
+	ifneq ($(origin FEATURES_TEST),file)
+		MODULE_ARGS_TEST += --features ${FEATURES_TEST}
+	endif
 endif
 
 TEST_ARGS ?=
--- a/Makefile
+++ b/Makefile
@@ -156,14 +156,7 @@ bin/svsm-kernel.elf: bin
 	cargo build --package svsm --bin svsm ${CARGO_ARGS} ${SVSM_ARGS} --target=x86_64-unknown-none
 	objcopy -O elf64-x86-64 ${OBJCOPY_ELF_ARGS} ${SVSM_KERNEL_ELF} $@
 
-bin/test-kernel.elf: bin
-# RUSTDOC=true removes doctests, which is necessary as they do not work with
-# custom test runners. See https://github.com/coconut-svsm/svsm/issues/705.
-	RUSTDOC=true LINK_TEST=1 cargo +nightly test --package svsm ${CARGO_ARGS} ${SVSM_ARGS_TEST} \
-		--target=x86_64-unknown-none \
-		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../$@"]'
-
-TEST_IN_SVSM_MODULES =
+TEST_IN_SVSM_MODULES = svsm
 TEST_IN_SVSM_TARGETS = $(TEST_IN_SVSM_MODULES:%=bin/test-%.elf)
 # Root of SVSM
 MAKEFILE_DIR = $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,19 @@ bin/test-kernel.elf: bin
 		--target=x86_64-unknown-none \
 		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../$@"]'
 
+TEST_IN_SVSM_MODULES =
+TEST_IN_SVSM_TARGETS = $(TEST_IN_SVSM_MODULES:%=bin/test-%.elf)
+# Root of SVSM
+MAKEFILE_DIR = $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
+$(TEST_IN_SVSM_TARGETS): bin/test-%.elf: bin
+# RUSTDOC=true removes doctests, which is necessary as they do not work with
+# custom test runners. See https://github.com/coconut-svsm/svsm/issues/705.
+	RUSTDOC=true LINK_TEST=1 cargo +nightly test --package $* \
+		$(CARGO_ARGS) ${MODULE_ARGS_TEST} \
+		--target=x86_64-unknown-none \
+		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 $(MAKEFILE_DIR)/$@"]'
+
 ${FS_BIN}: bin
 ifneq ($(FS_FILE), none)
 	cp -f $(FS_FILE) ${FS_BIN}
@@ -193,4 +206,4 @@ clean:
 
 distclean: clean
 
-.PHONY: test miri clean clippy bin/stage2.bin bin/svsm-kernel.elf bin/test-kernel.elf stage1_elf_trampoline distclean $(APROXYBIN) $(IGVM_FILES) $(IGVM_TEST_FILES)
+.PHONY: test miri clean clippy bin/stage2.bin bin/svsm-kernel.elf bin/test-kernel.elf stage1_elf_trampoline distclean $(APROXYBIN) $(IGVM_FILES) $(IGVM_TEST_FILES) $(TEST_IN_SVSM_TARGETS)
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ endif
 	touch ${FS_BIN}
 
 clippy:
-	${CARGO} clippy ${CLIPPY_OPTIONS} --workspace --exclude svsm --exclude stage1 --exclude svsm-fuzz -- ${CLIPPY_ARGS}
+	${CARGO} clippy ${CLIPPY_OPTIONS} --workspace --exclude svsm --exclude stage1 --exclude svsm-fuzz --exclude uapi_tester -- ${CLIPPY_ARGS}
 	RUSTFLAGS="--cfg fuzzing" ${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm-fuzz -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm --target x86_64-unknown-none -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package bldr --target x86_64-unknown-none -- ${CLIPPY_ARGS}
--- a/configs/test/qemu-test-target.json
+++ b/configs/test/qemu-test-target.json
@@ -28,6 +28,12 @@
     },
     "firmware": {},
     "fs": {
-        "modules": {}
+        "modules": {
+            "userinit": {
+                "path": "/init",
+                "type": "make",
+                "output_file": "bin/test-userinit.elf"
+            }
+        }
     }
 }
--- a/configs/test/qemu-test-target.json
+++ b/configs/test/qemu-test-target.json
@@ -13,7 +13,7 @@
     "kernel": {
         "svsm": {
             "type": "make",
-            "output_file": "bin/test-kernel.elf"
+            "output_file": "bin/test-svsm.elf"
         },
         "stage2": {
             "manifest": "kernel/Cargo.toml",
--- a/configs/test/qemu-test-target.json
+++ b/configs/test/qemu-test-target.json
@@ -33,6 +33,17 @@
     },
     "firmware": {},
     "fs": {
-        "modules": {}
+        "modules": {
+            "uapi_tester": {
+                "path": "/test/uapi_tester",
+                "type": "test",
+                "toolchain": "nightly",
+                "no_default_features": true,
+                "env": {
+                    "RUSTDOC": "true",
+                    "LINK_TEST": "1"
+                }
+            }
+        }
     }
 }
--- a/kernel/src/fs/filesystem.rs
+++ b/kernel/src/fs/filesystem.rs
@@ -931,7 +931,7 @@ mod tests {
 
         // Check if it appears in the listing
         let root_list = list_dir("").unwrap();
-        assert_eq!(root_list, [FileName::from("test1")]);
+        assert!(root_list.contains(&FileName::from("test1")));
 
         // Try again - should succeed now
         create("test1/file1").unwrap();
@@ -950,7 +950,7 @@ mod tests {
 
         // Check if it appears in the listing
         let root_list = list_dir("").unwrap();
-        assert_eq!(root_list, [FileName::from("test1")]);
+        assert!(root_list.contains(&FileName::from("test1")));
 
         // Try creating again as file - should fail
         create("test1").unwrap_err();
@@ -966,7 +966,8 @@ mod tests {
 
         // Check if it is removed from the listing
         let root_list = list_dir("").unwrap();
-        assert_eq!(root_list, [FileName::from("test2")]);
+        assert!(!root_list.contains(&FileName::from("test1")));
+        assert!(root_list.contains(&FileName::from("test2")));
 
         // Cleanup
         rmdir("test2").unwrap();
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -613,7 +613,34 @@ pub unsafe fn svsm_main(li: usize) {
 
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::{QEMUExitValue, exit};
+    use svsm::fs::{list_dir, opendir};
+    use svsm::task::exec_user;
+
     crate::test_main();
+
+    // What is a good way of identifying all
+    // userspace binaries? What about
+    // a file which lists all the modules that
+    // should be tested
+    let files = list_dir("").unwrap();
+    log::info!("Files: {:?}", files);
+    // Should I move this logic in a userspace
+    // binary (`test-runner`)?
+    for file in files {
+        // TODO: A waiting mechanism should be added
+        match exec_user(file.as_ref(), opendir("/").unwrap()) {
+            Ok(_) => (),
+            Err(e) => {
+                log::info!("Failed to launch {file}: {e:?}");
+                exit(QEMUExitValue::Fail)
+            }
+        }
+    }
+
+    log::info!("All tests passed!");
+
+    exit(QEMUExitValue::Success);
 }
 
 #[panic_handler]
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -623,11 +623,16 @@ pub unsafe fn svsm_main(li: usize) {
 
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::user_tests_in_svsm;
     use crate::testing::{QEMUExitValue, exit};
 
     // This function calls `svsm_kernel_test_runner` to sequentially run all
     // tests in the kernel crate collected during compilation.
     crate::kernel_tests_in_svsm();
+
+    user_tests_in_svsm();
+
+    log::info!("All tests passed!");
 
     exit(QEMUExitValue::Success);
 }
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -623,7 +623,9 @@ pub unsafe fn svsm_main(li: usize) {
 
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
-    crate::test_main();
+    // This function calls `svsm_kernel_test_runner` to sequentially run all
+    // tests in the kernel crate collected during compilation.
+    crate::kernel_tests_in_svsm();
 }
 
 #[panic_handler]
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -623,9 +623,13 @@ pub unsafe fn svsm_main(li: usize) {
 
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::{QEMUExitValue, exit};
+
     // This function calls `svsm_kernel_test_runner` to sequentially run all
     // tests in the kernel crate collected during compilation.
     crate::kernel_tests_in_svsm();
+
+    exit(QEMUExitValue::Success);
 }
 
 #[panic_handler]
--- a/kernel/src/syscall/class0.rs
+++ b/kernel/src/syscall/class0.rs
@@ -14,6 +14,13 @@ use core::ffi::c_char;
 use syscall::SysCallError;
 
 pub fn sys_exit(exit_code: u32) -> ! {
+    #[cfg(all(test, test_in_svsm))]
+    {
+        if exit_code != 0 {
+            use crate::testing::QEMUExitValue;
+            crate::testing::exit(QEMUExitValue::Fail);
+        }
+    }
     log::info!(
         "Terminating task {}, exit_code {exit_code}",
         current_task().get_task_name()
--- a/kernel/src/testing.rs
+++ b/kernel/src/testing.rs
@@ -84,9 +84,7 @@ pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
         (test_case.testfn.0)();
     }
 
-    info!("All tests passed!");
-
-    exit(QEMUExitValue::Success);
+    info!("Kernel tests passed!");
 }
 
 /// Exits the in-SVSM test session with the given exit value.
--- a/kernel/src/testing.rs
+++ b/kernel/src/testing.rs
@@ -1,15 +1,20 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 //
 
+extern crate alloc;
+
 use log::info;
 use test::ShouldPanic;
 
+use crate::fs::{list_dir, opendir};
+use crate::task::{TaskExitStatus, exec_user, wait_for_termination};
 use crate::{
     locking::{LockGuard, SpinLock},
     platform::SVSM_PLATFORM,
     serial::SerialPort,
     testutils::has_qemu_testdev,
 };
+use alloc::format;
 
 #[macro_export]
 macro_rules! assert_eq_warn {
@@ -87,6 +92,28 @@ pub fn svsm_kernel_test_runner(test_cases: &[&test::TestDescAndFn]) {
     }
 
     info!("Kernel tests in svsm passed!");
+}
+
+/// Executes the user binaries listed in `/test` dir sequentially, checks their exit
+/// value and panics if the binary exits with a code different from zero.
+/// If the directory is not present, it is assumed that there are no tests to run.
+pub fn user_tests_in_svsm() {
+    let Ok(files) = list_dir("/test") else {
+        log::info!("/test dir not found, skipping user tests in svsm");
+        return;
+    };
+
+    for file in files {
+        log::info!("Running tests in svsm for {file}");
+        let task = exec_user(&format!("test/{file}"), opendir("/").unwrap()).unwrap();
+        wait_for_termination(task.clone());
+        let exit_status = task.get_exit_status().unwrap();
+        assert_eq!(
+            exit_status,
+            TaskExitStatus::Exited(0),
+            "{file} tests failed with exit status {exit_status:?}"
+        );
+    }
 }
 
 /// Exits the in-SVSM test session with the given exit value.
--- a/kernel/src/testing.rs
+++ b/kernel/src/testing.rs
@@ -62,7 +62,7 @@ pub fn svsm_test_io() -> LockGuard<'static, Option<SerialPort<'static>>> {
     sp
 }
 
-pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
+pub fn svsm_kernel_test_runner(test_cases: &[&test::TestDescAndFn]) {
     info!("running {} tests", test_cases.len());
     for mut test_case in test_cases.iter().copied().copied() {
         if test_case.desc.should_panic == ShouldPanic::Yes {
--- a/kernel/src/testing.rs
+++ b/kernel/src/testing.rs
@@ -86,9 +86,7 @@ pub fn svsm_kernel_test_runner(test_cases: &[&test::TestDescAndFn]) {
         (test_case.testfn.0)();
     }
 
-    info!("All tests passed!");
-
-    exit(QEMUExitValue::Success);
+    info!("Kernel tests in svsm passed!");
 }
 
 /// Exits the in-SVSM test session with the given exit value.
--- a/user/init/Cargo.toml
+++ b/user/init/Cargo.toml
@@ -9,8 +9,16 @@ path = "src/main.rs"
 test = false
 doctest = false
 
+[lib]
+test = true
+doctest = true
+
 [dependencies]
 userlib.workspace = true
+
+[target."x86_64-unknown-none".dev-dependencies]
+test.workspace = true
+userlib = { workspace = true, features = ["test_runner"] }
 
 [lints]
 workspace = true
--- a/user/init/build.rs
+++ b/user/init/build.rs
@@ -2,6 +2,23 @@
 //
 
 fn main() {
-    println!("cargo:rustc-link-arg=-Tuser/lib/module.lds");
-    println!("cargo:rustc-link-arg=-no-pie");
+    // Extra cfgs
+    println!("cargo::rustc-check-cfg=cfg(test_in_svsm)");
+
+    // Userinit
+    println!("cargo:rustc-link-arg-bin=userinit=-Tuser/lib/module.lds");
+    println!("cargo:rustc-link-arg-bin=userinit=-no-pie");
+    println!("cargo:rustc-link-arg-bin=userinit=-nostdlib");
+
+    // Extra linker args for tests.
+    println!("cargo:rerun-if-env-changed=LINK_TEST");
+    if std::env::var("LINK_TEST").is_ok() {
+        println!("cargo:rustc-cfg=test_in_svsm");
+        println!("cargo:rustc-link-arg=-nostdlib");
+        println!("cargo:rustc-link-arg=-Tuser/lib/module.lds");
+        println!("cargo:rustc-link-arg=-no-pie");
+    }
+
+    println!("cargo:rerun-if-changed=user/lib/module.lds");
+    println!("cargo:rerun-if-changed=build.rs");
 }
--- a/user/init/src/main.rs
+++ b/user/init/src/main.rs
@@ -7,8 +7,16 @@ use userlib::*;
 
 declare_main!(main);
 
+#[cfg(not(test))]
 fn main() -> u32 {
     println!("COCONUT-SVSM init process starting");
+
+    0
+}
+
+#[cfg(test)]
+fn main() -> u32 {
+    crate::userspace_test_main();
 
     0
 }
--- a/user/init/src/main.rs
+++ b/user/init/src/main.rs
@@ -1,44 +1,49 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 //
-
-#![no_std]
-#![no_main]
+#![cfg_attr(not(test), no_std)]
+#![cfg_attr(not(test), no_main)]
 
 use userlib::*;
-
-use core::ptr::{addr_of, addr_of_mut};
-
-static mut SOME_BSS_DATA: [u64; 128] = [0; 128];
-static mut SOME_DATA: [u64; 128] = [0x01; 128];
-static SOME_RO_DATA: [u64; 128] = [0xee; 128];
-
-fn check(arr: &[u64; 128], val: u64) {
-    for v in arr.iter() {
-        if *v != val {
-            panic!("Unexpected array value");
-        }
-    }
-}
-
-fn write(arr: &mut [u64; 128], val: u64) {
-    for v in arr.iter_mut() {
-        *v = val;
-    }
-}
 
 declare_main!(main);
 
 fn main() -> u32 {
     println!("COCONUT-SVSM init process starting");
 
-    // SAFETY: Single-threaded process, so no data races. Safe to access global
-    // mutable data.
-    unsafe {
-        write(&mut *addr_of_mut!(SOME_DATA), 0xcc);
-        write(&mut *addr_of_mut!(SOME_BSS_DATA), 0xaa);
-        check(&*addr_of!(SOME_DATA), 0xccu64);
-        check(&*addr_of!(SOME_RO_DATA), 0xeeu64);
-        check(&*addr_of!(SOME_BSS_DATA), 0xaa);
-    }
     0
+}
+
+#[cfg(test)]
+mod tests {
+    use core::ptr::{addr_of, addr_of_mut};
+
+    static mut SOME_BSS_DATA: [u64; 128] = [0; 128];
+    static mut SOME_DATA: [u64; 128] = [0x01; 128];
+    static SOME_RO_DATA: [u64; 128] = [0xee; 128];
+
+    fn check(arr: &[u64; 128], val: u64) {
+        for v in arr.iter() {
+            assert_eq!(*v, val, "Unexpected array value");
+        }
+    }
+
+    fn write(arr: &mut [u64; 128], val: u64) {
+        for v in arr.iter_mut() {
+            *v = val;
+        }
+    }
+
+    #[test]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_memory_check() {
+        // SAFETY: Single-threaded process, so no data races. Safe to access global
+        // mutable data.
+        unsafe {
+            write(&mut *addr_of_mut!(SOME_DATA), 0xcc);
+            write(&mut *addr_of_mut!(SOME_BSS_DATA), 0xaa);
+            check(&*addr_of!(SOME_DATA), 0xccu64);
+            check(&*addr_of!(SOME_RO_DATA), 0xeeu64);
+            check(&*addr_of!(SOME_BSS_DATA), 0xaa);
+        }
+    }
 }
--- a/user/init/src/main.rs
+++ b/user/init/src/main.rs
@@ -6,39 +6,10 @@
 
 use userlib::*;
 
-use core::ptr::{addr_of, addr_of_mut};
-
-static mut SOME_BSS_DATA: [u64; 128] = [0; 128];
-static mut SOME_DATA: [u64; 128] = [0x01; 128];
-static SOME_RO_DATA: [u64; 128] = [0xee; 128];
-
-fn check(arr: &[u64; 128], val: u64) {
-    for v in arr.iter() {
-        if *v != val {
-            panic!("Unexpected array value");
-        }
-    }
-}
-
-fn write(arr: &mut [u64; 128], val: u64) {
-    for v in arr.iter_mut() {
-        *v = val;
-    }
-}
-
 declare_main!(main);
 
 fn main() -> u32 {
     println!("COCONUT-SVSM init process starting");
 
-    // SAFETY: Single-threaded process, so no data races. Safe to access global
-    // mutable data.
-    unsafe {
-        write(&mut *addr_of_mut!(SOME_DATA), 0xcc);
-        write(&mut *addr_of_mut!(SOME_BSS_DATA), 0xaa);
-        check(&*addr_of!(SOME_DATA), 0xccu64);
-        check(&*addr_of!(SOME_RO_DATA), 0xeeu64);
-        check(&*addr_of!(SOME_BSS_DATA), 0xaa);
-    }
     0
 }
--- a/user/lib/src/testing.rs
+++ b/user/lib/src/testing.rs
@@ -1,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+
+use crate::{console_print, print, println};
+use test::ShouldPanic;
+
+pub fn svsm_userspace_test_runner(test_cases: &[&test::TestDescAndFn]) {
+    println!("running {} user tests", test_cases.len());
+    for mut test_case in test_cases.iter().copied().copied() {
+        if test_case.desc.should_panic == ShouldPanic::Yes {
+            test_case.desc.ignore = true;
+            test_case
+                .desc
+                .ignore_message
+                .get_or_insert("#[should_panic] not supported");
+        }
+
+        if test_case.desc.ignore {
+            if let Some(message) = test_case.desc.ignore_message {
+                println!("test {} ... ignored, {message}", test_case.desc.name.0);
+            } else {
+                println!("test {} ... ignored", test_case.desc.name.0);
+            }
+            continue;
+        }
+
+        println!("test {} ...", test_case.desc.name.0);
+        (test_case.testfn.0)();
+    }
+
+    println!("Userspace tests passed!");
+}
--- a/user/lib/src/testing.rs
+++ b/user/lib/src/testing.rs
@@ -1,0 +1,32 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+
+use crate::println;
+use test::ShouldPanic;
+
+pub fn svsm_usermodule_test_runner(test_cases: &[&test::TestDescAndFn]) {
+    println!("running {} user tests", test_cases.len());
+    for mut test_case in test_cases.iter().copied().copied() {
+        if test_case.desc.should_panic == ShouldPanic::Yes {
+            test_case.desc.ignore = true;
+            test_case
+                .desc
+                .ignore_message
+                .get_or_insert("#[should_panic] not supported");
+        }
+
+        if test_case.desc.ignore {
+            if let Some(message) = test_case.desc.ignore_message {
+                println!("test {} ... ignored, {message}", test_case.desc.name.0);
+            } else {
+                println!("test {} ... ignored", test_case.desc.name.0);
+            }
+            continue;
+        }
+
+        println!("test {} ...", test_case.desc.name.0);
+        (test_case.testfn.0)();
+    }
+
+    println!("Usermodule tests in svsm passed!");
+}
--- a/user/uapi_tester/Cargo.toml
+++ b/user/uapi_tester/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "uapi_tester"
+version = "0.1.0"
+edition.workspace = true
+
+[[bin]]
+name = "uapi_tester"
+path = "src/main.rs"
+test = false
+doctest = false
+
+[lib]
+test = true
+doctest = true
+
+[dependencies]
+userlib.workspace = true
+
+[target."x86_64-unknown-none".dev-dependencies]
+test.workspace = true
+userlib = { workspace = true, features = ["test_runner"] }
+
+[lints]
+workspace = true
--- a/user/uapi_tester/src/main.rs
+++ b/user/uapi_tester/src/main.rs
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+
+#![cfg_attr(not(test), no_std)]
+#![cfg_attr(not(test), no_main)]
+
+#[cfg(not(test))]
+compile_error!(
+    "This crate is only meant to be used for testing purposes. \
+    It should not be built outside of a test context."
+);
+
+use userlib::*;
+
+#[cfg(test)]
+declare_main!(main);
+
+#[cfg(test)]
+fn main() -> u32 {
+    crate::usermodule_tests_in_svsm();
+    0
+}
+
+#[cfg(test)]
+mod tests {
+    static mut SOME_BSS_DATA: [u64; 128] = [0; 128];
+    static mut SOME_DATA: [u64; 128] = [0x01; 128];
+    static SOME_RO_DATA: [u64; 128] = [0xee; 128];
+
+    fn check(arr: &[u64; 128], val: u64) {
+        for v in arr.iter() {
+            assert_eq!(*v, val, "Unexpected array value");
+        }
+    }
+
+    fn write(arr: &mut [u64; 128], val: u64) {
+        for v in arr.iter_mut() {
+            *v = val;
+        }
+    }
+
+    #[test]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_memory_check() {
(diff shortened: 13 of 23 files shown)


kernel/src/svsm.rs
```diff
@@ -470,9 +483,33 @@ pub unsafe fn svsm_main(li: usize) {
     cpu_idle_loop(0);
 }
 
+#[cfg(all(test, test_in_svsm))]
+static KERNEL_FS_START: ImmutAfterInitCell<u64> = ImmutAfterInitCell::uninit();
+#[cfg(all(test, test_in_svsm))]
+static KERNEL_FS_END: ImmutAfterInitCell<u64> = ImmutAfterInitCell::uninit();
+
```
stefano-garzarella: Instead of re-introducing global variables, can we just revert commit https://github.com/coconut-svsm/svsm/pull/926/commits/7cda2a2be7becb768c91a934b2b7b3f3139deaa9 and call `populate_ram_fs()` in `svsm_init()` for `not(test)` and `test_in_svsm`?

user/init/Cargo.toml
```diff
@@ -9,6 +9,10 @@ path = "src/main.rs"
 test = false
 doctest = false
 
+[lib]
+test = true
+doctest = true
+
```
stefano-garzarella: Why this is needed?
Should we mention in the commit description?

Also is commit https://github.com/coconut-svsm/svsm/pull/926/commits/954f011403c3ab9a4c3ab7548026b0919f4a3cf0 the right one?

Cargo.lock
stefano-garzarella, line 1: Should this change be moved where we update Cargo.toml (commit https://github.com/coconut-svsm/svsm/pull/926/commits/b6caa8479f99cf12536d171e4c6032e43ca2ff92) ?

stefano-garzarella (commented):
Thanks for this!

BTW do we need any changes in the CI to run these tests, or they are already covered?

kernel/src/svsm.rs
luigix25: Another option would be to modify the tests, so that they don't expect an empty FS anymore. However, I'm not sure how doable it is.

stefano-garzarella: > Another option would be to modify the tests, so that they don't expect an empty FS anymore. However, I'm not sure how doable it is.

Yeah, agree! If it's easily doable, it's even better, so we don't need extra cfg in our codebase.

luigix25: Yep, it's an easy fix.

You can do something like this:

Replace

```rust
assert_eq!(root_list, [FileName::from("test1")]);
```

with

```rust
assert!(root_list.contains(&FileName::from("test1")));
```

there are multiple pieces to change, this is just a suggestion.

user/init/Cargo.toml
n-ramacciotti: > Why this is needed? Should we mention in the commit description?

By default these values are already set to true, so this section can be removed. It is here for clarity and for consistency with `kernel/Cargo.toml`, which contains the same section.

> Also is commit [954f011](https://github.com/coconut-svsm/svsm/commit/954f011403c3ab9a4c3ab7548026b0919f4a3cf0) the right one?

You're right, I made a mistake when splitting commits. 
It should have been introduced with `userinit/src/lib.rs` in  [b6caa84](https://github.com/coconut-svsm/svsm/pull/926/commits/b6caa8479f99cf12536d171e4c6032e43ca2ff92)

Cargo.lock
n-ramacciotti: Yes, that makes sense

kernel/src/svsm.rs
n-ramacciotti: > Another option would be to modify the tests, so that they don't expect an empty FS anymore. However, I'm not sure how doable it is.

The filesystem tests are run both inside and outside SVSM. Another option would be to disable the tests in SVSM using something like:
```
#[cfg_attr(test_in_svsm, ignore = "Can only be run outside guest")]
```

However, it probably makes more sense to leave the tests enabled and follow your suggestion, as it keeps things cleaner.

configs/test/qemu-test-target.json
```diff
@@ -28,6 +28,12 @@
     },
     "firmware": {},
     "fs": {
-        "modules": {}
+        "modules": {
```
luigix25, line 36: I see that you didn't change hyperv and vanadium targets. Is there a reason for that?

kernel/src/svsm.rs
```diff
@@ -472,6 +472,22 @@ pub unsafe fn svsm_main(li: usize) {
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
     crate::test_main();
+
+    #[cfg(test_in_svsm)]
```
luigix25: I'm not a test expert, but why is this needed?

kernel/src/svsm.rs
```diff
@@ -472,6 +472,22 @@ pub unsafe fn svsm_main(li: usize) {
...
+    {
+        use crate::testing::QEMUExitValue;
+        use svsm::fs::opendir;
+        use svsm::task::exec_user;
+
+        match exec_user("/init", opendir("/").expect("Failed to find FS root")) {
```
luigix25: `exec_user` is now executed regardless of tests, can we avoid code duplication somehow?

user/init/src/main.rs
```diff
@@ -1,41 +1,53 @@
...
-    // SAFETY: Single-threaded process, so no data races. Safe to access global
-    // mutable data.
-    unsafe {
-        write(&mut *addr_of_mut!(SOME_DATA), 0xcc);
-        write(&mut *addr_of_mut!(SOME_BSS_DATA), 0xaa);
-        check(&*addr_of!(SOME_DATA), 0xccu64);
-        check(&*addr_of!(SOME_RO_DATA), 0xeeu64);
-        check(&*addr_of!(SOME_BSS_DATA), 0xaa);
+    #[test]
+    fn test_memory_check() {
```
luigix25: IIUC this test can only be run in a guest right? if so I think you need to add 
`#[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]`

configs/test/qemu-test-target.json
n-ramacciotti: There’s no particular reason. I didn’t test those targets and omitted them, but I don’t see any specific reason not to update them as well. I’ll include those changes too.

kernel/src/svsm.rs
n-ramacciotti: You’re right, it’s actually not needed. `test_in_svsm_task` is only included when running tests inside SVSM, so the additional cfg is redundant and can be removed.

For clarity, an alternative could be to change the initial `#[cfg(test)]` to `#[cfg(all(test, test_in_svsm))]`, but this would be redundant as well.

user/init/src/main.rs
n-ramacciotti: The `tests` module is included only with both `test` and `test_in_svsm` enabled. I can change it to be included only for `test` and apply your suggestion. Thanks.

kernel/src/svsm.rs
n-ramacciotti: In this configuration, user tests start after kernel tests. 
It is possible to move `exec_user` and `testing::exit()` out of `test_in_svsm_task` and to place them before the `#[cfg(not(test))]`, removing the duplication from that section

Makefile
```diff
@@ -154,6 +154,11 @@ bin/test-kernel.elf: bin
 		--target=x86_64-unknown-none \
 		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../$@"]'
 
+bin/test-userinit.elf: bin
+	RUSTDOC=true LINK_TEST=1 cargo +nightly test --package userinit ${CARGO_ARGS} \
+        --target=x86_64-unknown-none \
+		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../../$@"]'
```
stefano-garzarella: Please use tabs here. Should not be a problem since it's a continuation after `\` but yeah, Makefile can be picky with spaces vs tabs.

Makefile
```diff
@@ -154,6 +154,11 @@ bin/test-kernel.elf: bin
 		--target=x86_64-unknown-none \
 		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../$@"]'
 
+bin/test-userinit.elf: bin
+	RUSTDOC=true LINK_TEST=1 cargo +nightly test --package userinit ${CARGO_ARGS} \
```
stefano-garzarella: Is `RUSTDOC=true` really needed also here?

user/init/build.rs
```diff
@@ -1,4 +1,21 @@
 fn main() {
-    println!("cargo:rustc-link-arg=-Tuser/lib/module.lds");
-    println!("cargo:rustc-link-arg=-no-pie");
+    // Extra cfgs
+    println!("cargo::rustc-check-cfg=cfg(test_in_svsm)");
```
stefano-garzarella: is `::` intentional here in `cargo::r...`?

I think it works, but since we are using a single `:` in all other cases, maybe we can uniform the file.

Makefile
```diff
@@ -198,7 +198,8 @@ clippy:
 	RUSTFLAGS="--cfg fuzzing" cargo clippy ${CLIPPY_OPTIONS} --all-features --package svsm-fuzz -- ${CLIPPY_ARGS}
 	cargo clippy ${CLIPPY_OPTIONS} --all-features --package svsm --target x86_64-unknown-none -- ${CLIPPY_ARGS}
 	cargo clippy ${CLIPPY_OPTIONS} --all-features --package stage1 --target x86_64-unknown-none -- ${CLIPPY_ARGS} ${STAGE1_RUSTC_ARGS}
-	cargo clippy ${CLIPPY_OPTIONS} --all-features --workspace --tests --exclude packit -- ${CLIPPY_ARGS}
+	cargo clippy ${CLIPPY_OPTIONS} --all-features --workspace --tests --exclude packit --exclude userlib -- ${CLIPPY_ARGS}
+	cargo clippy ${CLIPPY_OPTIONS} --package userlib --tests -- ${CLIPPY_ARGS}
```
stefano-garzarella: Thanks for mentioning in the commit:

> Exclude the new `test_runner` feature from clippy checks since
> it is only used for `test_in_svsm`, not for regular tests that
> use `--all-features`.

But I don't understand why it's related to `--all-features`, what was the issue?

kernel/src/fs/filesystem.rs
```diff
@@ -966,7 +966,7 @@ mod tests {
 
         // Check if it is removed from the listing
         let root_list = list_dir("").unwrap();
-        assert_eq!(root_list, [FileName::from("test2")]);
+        assert!(!root_list.contains(&FileName::from("test1")));
```
stefano-garzarella, line 969: To be fair, I think we should also check that `test2` is still there

kernel/src/svsm.rs
```diff
@@ -497,7 +497,20 @@ pub unsafe fn svsm_main(li: usize) {
...
     crate::test_main();
+
+    match exec_user("/init", opendir("/").expect("Failed to find FS root")) {
+        Ok(_) => (),
+        Err(e) => log::info!("Failed to launch /init: {e:?}"),
+    }
+
+    log::info!("All tests passed!");
+
+    crate::testing::exit(QEMUExitValue::Success);
```
stefano-garzarella: What about adding `exit` in the `use` on top?

I mean `use crate::testing::{exit, QEMUExitValue}`

kernel/src/svsm.rs
```diff
@@ -497,7 +497,20 @@ pub unsafe fn svsm_main(li: usize) {
...
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::QEMUExitValue;
+    use svsm::fs::opendir;
+    use svsm::task::exec_user;
+
     crate::test_main();
+
+    match exec_user("/init", opendir("/").expect("Failed to find FS root")) {
+        Ok(_) => (),
+        Err(e) => log::info!("Failed to launch /init: {e:?}"),
```
stefano-garzarella: So, if we fail to launch `/init` we just log it, but exit with `QEMUExitValue::Success`, is that on purpose ?

kernel/src/syscall/class0.rs
```diff
@@ -14,6 +14,13 @@ use core::ffi::c_char;
 use syscall::SysCallError;
 
 pub fn sys_exit(exit_code: u32) -> ! {
+    #[cfg(all(test, test_in_svsm))]
+    {
+        if exit_code != 0 {
+            use crate::testing::QEMUExitValue;
+            crate::testing::exit(QEMUExitValue::Fail);
+        }
+    }
```
stefano-garzarella: Here we should add a comment, and I'm not sure this specific code should go in the sys calls. @joergroedel WDYT?

Do we have any other less invasive alternative?

kernel/src/svsm.rs
```diff
@@ -497,7 +497,20 @@ pub unsafe fn svsm_main(li: usize) {
...
+    }
+
+    log::info!("All tests passed!");
```
stefano-garzarella, line 635: What about `Userspace tests passed` ?

kernel/src/svsm.rs
```diff
@@ -497,7 +497,20 @@ pub unsafe fn svsm_main(li: usize) {
 
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::QEMUExitValue;
+    use svsm::fs::opendir;
+    use svsm::task::exec_user;
+
     crate::test_main();
+
+    match exec_user("/init", opendir("/").expect("Failed to find FS root")) {
```
stefano-garzarella: Are we sure after `exec_user` returns, the user process is ended?

I'm asking because it's returning the task_id, so not sure if we should add some other API to wait the completion. In that case maybe we can also retrieve the return value without the changes in `exit()` syscall.

user/init/src/main.rs
stefano-garzarella, line 1:
```
user/init: Add memory tests

Move memory checks in a dedicated test function

Signed-off-by: Nicola Ramacciotti <niko.ramak@gmail.com>
```

IMO we should describe a bit better this commit https://github.com/coconut-svsm/svsm/pull/926/changes/6346327986d5bac8f800a0489a15655688d0cda1 , especially that now the production binary is not running tests anymore (not sure why we did it, but we should at least make it sure other people will discovery it from the git history)

Makefile
n-ramacciotti: It is not needed at this time. I can remove it.

user/init/build.rs
n-ramacciotti: > is `::` intentional here in `cargo::r...`?

I used the same pattern as in `kernel/build.rs`.

> I think it works, but since we are using a single `:` in all other cases, maybe we can uniform the file.

IIUC, from [doc](https://doc.rust-lang.org/cargo/reference/build-scripts.html):
```
MSRV: 1.77 is required for cargo::KEY=VALUE syntax. To support older versions, use the cargo:KEY=VALUE syntax.
```
The one with `::` is the new syntax. It should work both ways. 

Which version would you prefer to use for unifying the file?

Makefile
n-ramacciotti: Using clippy with `--tests` and `--all-features` creates a conflict on the test crate. Given that the `test_runner` feature is enabled, `test` from the workspace is imported instead of the standard one and the following error occurs:
```
$ cargo clippy --all-features --package userlib --tests
    Checking test v0.1.0 (...) 
    ...
    Checking userlib v0.1.0 (...)
error[E0425]: cannot find function `test_main_static` in crate `test`
```

I had to disable `--all-features` to avoid this issue.

kernel/src/svsm.rs
n-ramacciotti: I'll change this behavior to exit with a failure

kernel/src/svsm.rs
n-ramacciotti: This is related to all tests, both kernel and userspace. The `Userspace tests passed` message is printed by the userspace runner in `user/lib/src/testing.rs`.

kernel/src/svsm.rs
n-ramacciotti: > Are we sure after `exec_user` returns, the user process is ended?
>
> I'm asking because it's returning the task_id, so not sure if we should add some other API to wait the completion.

Yes, a waiting mechanism should be added . I noticed that a  `WaitQueue` is defined in `kernel/src/task/waiting.rs`, but it is never used in the code. Maybe it could be used as a starting point.

> In that case maybe we can also retrieve the return value without the changes in `exit()` syscall.

Right

msft-jlange: I understand why we want to invoke the user-mode test harness from a test-in-svsm kernel, but we shouldn't name the test process `init` because `init` is already a separate concept.  It would be much better to have a dedicated user-mode binary that acts as a test runner rather than borrowing an existing binary.

user/init/build.rs
msft-jlange: See my other comment about moving the test harness out of `init` into a dedicated binary.

kernel/src/svsm.rs
n-ramacciotti: My initial idea was that each user-mode binary would have to use `userlib::testing::svsm_userspace_test_runner` and  it would be compiled with `test_in_svsm`. 
I can then introduce a user-space binary (for example, `test-init`), which calls the `exec` syscall for all the user-space binary.

I'm not sure I fully understand your point. Could you please explain that in more detail?

Are you referring to a single user-mode runner which  groups and invokes tests from all user-mode binaries?

stefano-garzarella: > I can then introduce a user-space binary (for example, `test-init`),

Yep, this one, but I would not use `init` in the name, since it's not related to the `init` process, but it's just a test runner, so maybe `test-runner`.

So, instead of having this in `user/init/src/main.rs` :
```
fn main() -> u32 {
    #[cfg(test)]
    {
        crate::userspace_test_main();
    }

    #[cfg(not(test))]
    {
        println!("COCONUT-SVSM init process starting");
    }
    0
}
```

We should add a new `user/test-runner/src/main.rs` with:
```
fn main() -> u32 {
     crate::userspace_test_main();

    0
}
```

Or something like that.

Then, in `test_in_svsm_task()`, instead of using `exec_user(“/init”, ...)`, use `exec_user(“/test-runner”,...)`, which makes much more sense.

n-ramacciotti: > Yep, this one, but I would not use `init` in the name, since it's not related to the `init` process, but it's just a test runner, so maybe `test-runner`.

It makes sense

> We should add a new `user/test-runner/src/main.rs` with:
> 
> ```
> fn main() -> u32 {
>      crate::userspace_test_main();
> 
>     0
> }
> ```
> 
> Or something like that.

IIUC, `test-runner` could be a "simple" loop on each user binary.
In this case `crate::userspace_test_main()` and the custom runner
`userlib::testing::svsm_userspace_test_runner` aren't needed in this
crate.

> So, instead of having this in `user/init/src/main.rs` :
> 
> ```
> fn main() -> u32 {
>     #[cfg(test)]
>     {
>         crate::userspace_test_main();
>     }
> 
>     #[cfg(not(test))]
>     {
>         println!("COCONUT-SVSM init process starting");
>     }
>     0
> }
> ```
> 

If I want to run tests inside `init`, like the `test_memory_check` I introduce in [6346327](https://github.com/coconut-svsm/svsm/commit/6346327986d5bac8f800a0489a15655688d0cda1), shouldn't I have the above configuration in `init`?

stefano-garzarella: > If I want to run tests inside `init`, like the `test_memory_check` I introduce in [6346327](https://github.com/coconut-svsm/svsm/commit/6346327986d5bac8f800a0489a15655688d0cda1), shouldn't I have the above configuration in `init`?

Why that test should be in init ?

stefano-garzarella: > IIUC, `test-runner` could be a "simple" loop on each user binary.
> In this case `crate::userspace_test_main()` and the custom runner
> `userlib::testing::svsm_userspace_test_runner` aren't needed in this
> crate.

I haven't looked into the details, but it might make sense.

n-ramacciotti: > Why that test should be in init ?

The current behavior of `init`(see [e8d571c](https://github.com/coconut-svsm/svsm/commit/e8d571c18b06bcce144744fffcff50e28e20c483)) is to perform some checks that I moved into `test_memory_check`. I don't think this check will be necessary in the future, it was just to show the behavior of the PR.

stefano-garzarella: > > Why that test should be in init ?
> 
> The current behavior of `init`(see [e8d571c](https://github.com/coconut-svsm/svsm/commit/e8d571c18b06bcce144744fffcff50e28e20c483)) is to perform some checks that I moved into `test_memory_check`. I don't think this check will be necessary in the future, it was just to show the behavior of the PR.

I think we can remove them, or leave them as they are now or move in some other place, but IMO init's main should not have any `#[cfg(test)]`

n-ramacciotti: > IIUC, `test-runner` could be a "simple" loop on each user binary.

In this case, could we loop directly  into `test_in_svsm_task` instead of using an additional binary (`test-runner`)?

stefano-garzarella: > > IIUC, `test-runner` could be a "simple" loop on each user binary.
> 
> In this case, could we loop directly into `test_in_svsm_task` instead of using an additional binary (`test-runner`)?

No idea, if you think it's doable, let's try that and we can check how the result is.

n-ramacciotti: > No idea, if you think it's doable, let's try that and we can check how the result is.

I can try, but currently there is still the problem that you highlighted in [this discussion](https://github.com/coconut-svsm/svsm/pull/926#discussion_r2848300621).

However, as there are currently no userspace modules to test, I can upload the structure for now.

n-ramacciotti:
V7:
- Rebased on the latest main.
- Introduced second main in `init` for testing purposes only, which calls the custom test harness.
  This is to avoid changing the normal main.

user/init/src/main.rs
```diff
@@ -7,12 +7,20 @@ use userlib::*;
 
 declare_main!(main);
 
+#[cfg(not(test))]
```
n-ramacciotti: ~~This commit will be squashed. It has been split here only for ease of viewing.~~

Another possibility consists in moving the test harness directly inside `declare_main` macro:

```
macro_rules! declare_main {
 // ...
 #[cfg(not(test))]
 {
       let main_fn: fn() -> u32 = $path;
       let ret = main_fn();
       exit(ret);
 }
  #[cfg(test)]
  {
        crate::userspace_test_main();
       exit(0)
  }
```

However, this requires:
- `#[allow(clippy::crate_in_macro_def)]` when defining the macro
- `#[allow(dead_code)]` for main function, as it is not used when testing

n-ramacciotti:
v8:
 - Moved the optional dependency `test` in  `userlib` under the target `x86_64-unknown-none`. This resolves [this discussion](https://github.com/coconut-svsm/svsm/pull/926#discussion_r2847940734). The feature `test_runner` still enables the optional dependency, but only when compiling for `x86_64-unkwown-none`, otherwise cargo ignores it.

n-ramacciotti:
v9:
- Added to `Makefile` a static pattern rule for user binaries instead of repeating the same command for each one of them

n-ramacciotti:
v10:
- rebased on latest main
- Moved `bin/test-kernel.elf` in the new static pattern rule. (and renamed to bin/test-svsm.elf)
- Changed default features assignment when directly launching `make bin/test-module.elf`

n-ramacciotti:
v11:
- rebased on latest main (with support for `wait_for_termination`)
- improvements to commit messages

Todo:
- [ ] Save and retrieve the exit code to determine whether the test was a failure or success 
  instead of using the exit syscall directly  (see [d3499a0](https://github.com/coconut-svsm/svsm/pull/926/changes/d3499a0bcd83cadba586747edb159e4321a1839d))
I think that the `save` part should be addressed in a different PR.  I'm not sure if it makes sense to mark this PR as ready for review without those changes. Perhaps the above commit could be reverted in another separate PR if this one can be considered ready [@stefano-garzarella]

user/lib/src/testing.rs
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+
+use crate::{console_print, print, println};
```
stefano-garzarella: is `console_print` really used?

kernel/src/svsm.rs
```diff
@@ -618,7 +618,35 @@ pub unsafe fn svsm_main(li: usize) {
 
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::{QEMUExitValue, exit};
+    use svsm::fs::{list_dir, opendir};
+    use svsm::task::{exec_user, wait_for_termination};
+
     crate::test_main();
+
+    // What is a good way of identifying all
```
stefano-garzarella: Please prefix these questions with something like TODO: or RFC:

kernel/src/svsm.rs
```diff
@@ -618,7 +618,35 @@ pub unsafe fn svsm_main(li: usize) {
...
+    // userspace binaries? What about
+    // a file which lists all the modules that
+    // should be tested
+    let files = list_dir("").unwrap();
+    log::info!("Files: {:?}", files);
+    // Should I move this logic in a userspace
```
stefano-garzarella: As we discussed, moving in userspace will require to add syscall to export wait_for_termination behavior, right?

kernel/src/svsm.rs
```diff
@@ -618,7 +618,35 @@ pub unsafe fn svsm_main(li: usize) {
...
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::{QEMUExitValue, exit};
+    use svsm::fs::{list_dir, opendir};
+    use svsm::task::{exec_user, wait_for_termination};
+
     crate::test_main();
+
+    // What is a good way of identifying all
+    // userspace binaries? What about
```
stefano-garzarella: How do we create that file?
I'd like to keep it as simple as possible.

kernel/src/svsm.rs
```diff
@@ -618,7 +618,35 @@ pub unsafe fn svsm_main(li: usize) {
...
+    log::info!("Files: {:?}", files);
+    // Should I move this logic in a userspace
+    // binary (`test-runner`)?
+    for file in files {
+        match exec_user(file.as_ref(), opendir("/").unwrap()) {
+            Ok(task) => {
+                wait_for_termination(task);
+            }
+            Err(e) => {
+                log::info!("Failed to launch {file}: {e:?}");
```
stefano-garzarella: maybe error instead of info?

kernel/src/svsm.rs
```diff
@@ -618,7 +618,35 @@ pub unsafe fn svsm_main(li: usize) {
 
 #[cfg(test)]
 fn test_in_svsm_task(_context: usize) {
+    use crate::testing::{QEMUExitValue, exit};
+    use svsm::fs::{list_dir, opendir};
+    use svsm::task::{exec_user, wait_for_termination};
+
     crate::test_main();
```
stefano-garzarella: I spent quite some time to understand this.

What about renaming this in kernel_tests() and adding userspace_tests() with the code below?

kernel/src/syscall/class0.rs
```diff
@@ -15,6 +15,15 @@ use core::ffi::c_char;
...
 
 pub fn sys_exit(exit_code: u32) -> ! {
+    #[cfg(all(test, test_in_svsm))]
+    {
+        // TODO: instead of exiting here, we should save the exit
+        // code and decide after wait_for_termination is called
+        if exit_code != 0 {
+            use crate::testing::QEMUExitValue;
+            crate::testing::exit(QEMUExitValue::Fail);
+        }
```
stefano-garzarella: yep, that would be much better IMO.

Makefile
```diff
@@ -6,9 +6,13 @@ endif
 
 FEATURES_TEST ?= vtpm,virtio-drivers,block,vsock
 SVSM_ARGS_TEST += --no-default-features
+MODULE_ARGS_TEST += --no-default-features
```
stefano-garzarella: I'm really confused by MODULE_ARGS_TEST and changes in the Makefile to build userspace modules. I think we should try to avoid this hack in some better way, but I need to think more for suggestions.

user/init/src/main.rs
```diff
@@ -42,3 +43,10 @@ fn main() -> u32 {
     }
     0
 }
+
+#[cfg(test)]
+fn main() -> u32 {
```
stefano-garzarella: IIRC we discussed this a bit, instead of adding this in user/init/src/main.rs why not creating something different, like user/test_runner/src/main.rs

init.rs should only be an init (to be extended later when we will have an userspace)

stefano-garzarella (commented):
I left some comments after a first look. Thanks for this work!

user/lib/src/testing.rs
n-ramacciotti: When I tried to compile without it, I got the following error message:
```
error[E0425]: cannot find function `console_print` in this scope
  --> user/lib/src/console.rs:38:27
   |
38 |         ($($arg:tt)*) => (console_print(format_args!($($arg)*)))
   |                           ^^^^^^^^^^^^^ not found in this scope
   |
  ::: user/lib/src/testing.rs:7:5
   |
7  |     println!("running {} user tests", test_cases.len());
   |     --------------------------------------------------- in this macro invocation
   |
   = note: this error originates in the macro `print` which comes from the expansion of the macro `println` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this function through its public re-export
  --> user/lib/src/testing.rs:3:1
   |
3  + use crate::console_print;
   |

```

kernel/src/svsm.rs
n-ramacciotti: Yes, it would require a new system call. For now, it can be maintained in the kernel. I will remove the comment

kernel/src/svsm.rs
n-ramacciotti: My idea was to modify xbuild by adding a new field to the recipe:
```
/// Components for the filesystem image.
#[derive(Debug, Clone, Deserialize)]
pub struct FsConfig {
    modules: HashMap<String, ComponentConfig>,
    ##
    files : ..
    ##
}
```

I should look into these more closely, though.

kernel/src/svsm.rs
n-ramacciotti: > What about renaming this in kernel_tests() 

Sure, it makes much sense!

> adding userspace_tests() with the code below?

Do you mean another function available during testing which wraps the loop? it makes sense. I will add it

kernel/src/syscall/class0.rs
n-ramacciotti: I'm working on saving the `exit_code`. It will be a separate PR

user/init/src/main.rs
n-ramacciotti: In my view, the loop in `test_in_svsm_task` loops over each userbinary compiled for testing mode. Each binary has to have its runner (imported from `userlib`) which will run the tests of that binary only.

Each userbinary will have:
```rust
#![cfg_attr(
    all(test, test_in_svsm),
    no_main,
    feature(custom_test_frameworks),
    test_runner(userlib::testing::svsm_userspace_test_runner),
    reexport_test_harness_main = "userspace_test_main"
)]
```

and will call `userspace_test_main` somewhere (in this case a main function available with #[cfg(test)].

When compiled for testing, `init` will just run the test found in the `init` crate tagged with:
```rust
```

Currently, only the 'init' module is available. However, when other modules become available, 'init' will not interact with them.

I don't understand what the purpose of a `test_runner` crate is in this scenario. @stefano-garzarella could you please elaborate  further.

stefano-garzarella: okay, maybe I expressed myself wrong, but in this case those tests are not really related to init right (also because our init is not doing anythin)?
(I know it's pre-existing)

So maybe better to add a new crate just for testing userspace support of svsm (that maybe we can extend for example to test syscall, etc.)

kernel/src/syscall/class0.rs
stefano-garzarella: great, thanks!

kernel/src/svsm.rs
stefano-garzarella: > > What about renaming this in kernel_tests()
> 
> Sure, it makes much sense!
> 
> > adding userspace_tests() with the code below?
> 
> Do you mean another function available during testing which wraps the loop? it makes sense. I will add it

yep

user/lib/src/testing.rs
stefano-garzarella: What about this?

```diff
diff --git a/user/lib/src/console.rs b/user/lib/src/console.rs
index a79c56ba..a04c2b2e 100644
--- a/user/lib/src/console.rs
+++ b/user/lib/src/console.rs
@@ -35,7 +35,7 @@ pub fn console_print(args: fmt::Arguments<'_>) {
 
 #[macro_export]
 macro_rules! print {
-        ($($arg:tt)*) => (console_print(format_args!($($arg)*)))
+        ($($arg:tt)*) => ($crate::console_print(format_args!($($arg)*)))
 }
 
 #[macro_export]
```

user/init/src/main.rs
n-ramacciotti: >  those tests are not really related to init right 

Right, those tests have nothing to do with `init`. Is it possible that `init` will  run related tests? In that case it should follow this structure

> add a new crate just for testing userspace support of svsm

Do you mean a dummy crate that contains only tests (for that crate), which are called by the (userlib) runner when the crate is compiled for testing and it is not used in normal builds?

user/lib/src/testing.rs
n-ramacciotti: Right! I will add a new commit for this

Makefile
n-ramacciotti: I agree that this is not the best solution.

For building userspace modules in a test environment, it is needed to follow the same pattern that is used for `make bin/test-kernel.elf`:

```makefile
bin/test-kernel.elf: bin
# RUSTDOC=true removes doctests, which is necessary as they do not work with
# custom test runners. See https://github.com/coconut-svsm/svsm/issues/705.
	RUSTDOC=true LINK_TEST=1 cargo +nightly test --package svsm \
	       ${CARGO_ARGS} \
	       ${SVSM_ARGS_TEST} \
		--target=x86_64-unknown-none \
		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../$@"]'
```

It should be something like:
```makefile
bin/test-userinit.elf: bin
# RUSTDOC=true removes doctests, which is necessary as they do not work with
# custom test runners. See https://github.com/coconut-svsm/svsm/issues/705.
	RUSTDOC=true LINK_TEST=1 cargo +nightly test --package userinit \
	        ${CARGO_ARGS} \
	         ${SVSM_ARGS_TEST} \
		--target=x86_64-unknown-none \
		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../../$@"]'
```

If I use `xbuild` (`make bin/coconut-qemu-test.igvm`) there is no problem as `xbuild` can sort the feature for module (svsm:attest, userinit:featureX) from a single `FEATURES=svsm:attest,userinit:featurex`.

The problem is there when using `make bin/test-userinit.elf` directly. As it would take `FEATURES_TEST` which contains features not related to `userinit` (vtpm, etc..).

This requires to specify a different `SVSM_ARGS_TEST`, lets say `USERINIT_ARGS_TEST`, but this would create a problem when using `make bin/coconut-test-qemu.igvm` as here `xbuild` should specify two different `FEATURES_SVSM` and `FEATURES_USERINIT` and this would expand as the number of usermodules increases.

Also all the default features should all be passed to `xbuild` like `SVSM_ARGS_TEST` now.

I added `MODULE_ARGS_TEST` which simply ignores `FEATURES_TEST` when calling `make bin/test-userinit.elf` directly.

A better solution would be to avoid to call back into the `makefile`, as @00xc suggested in #909:
`
bin/test-kernel.elf is still built from the Makefile for now, since it requires a special command line invocation, and moving that logic into xbuild would introduce too much complexity. Thus, we have xbuild call back into the Makefile to build it 
`

n-ramacciotti:
v12:
- Addressed most of @stefano-garzarella's comments:
  - add `$crate` in macro definitions (`print!` and `println!`)
  - Changed println in userspace test runner from `Userspace tests passed` to `Usermodule tests passed`
    as the runner only runs tests for a single module at a time
  - Rename kernel test harness from `test_main` to `kernel_tests_in_svsm`
  - Introduce `userspace_tests_in_svsm` to wrap the loop on the `exec` of userbinaries
  - Removed comment on moving the `exec` loop to userspace
  - Move `exit` from QEMU in a separate commit.
  - Prefix a comment in the code with [RFC]

TODO
- [ ] I'm working on saving the `exit_code`

n-ramacciotti:
CI is failing [1], I don't know why (yet), but at least now it fails in the right place. (before it failed after the print that all kernel tests succeeded, which was not true [2] as this is a failure in a kernel test)

@vsntk18 it's seems that `wait_for_termination` should be placed if we want to be sure that when  `start_kernel_thread` returns the task is terminated before moving. (as we discussed in [3]). 
I mentioned you also because, IIUC, this was a test you introduced and maybe you can have an idea on why it's failing. 

I will look into it in the next few days.

[1] https://github.com/coconut-svsm/svsm/actions/runs/27468738903/job/81195646363?pr=926
[2] https://github.com/coconut-svsm/svsm/actions/runs/27468172423/job/81194123020
[3] https://github.com/coconut-svsm/svsm/pull/52#discussion_r3396964529

vsntk18:
@n-ramacciotti  This test depends on the order of activation of tasks.  I will take a look on Monday.

n-ramacciotti:
> @n-ramacciotti This test depends on the order of activation of tasks. I will take a look on Monday.

@vsntk18  After some testing[1], I think a possible scheduling order is as follows: when the `SVSM test task` starts `task1` in `test_fpu_context_switch`, `task1` is scheduled. However, when `task1` starts `task2`, the `SVSM test task` is scheduled and continues until it finishes (printing `All kernel tests passed!`, which is not true). Then `task2` is scheduled until it finishes, after which `task1` resumes and fails the assertion.

I still don't know why the assertion fails. I don't think this part is related to task activation order, perhaps on the order of `task1` and `task2`, but we have to bu sure that `SVSM test task` wait that `task1` in `task2` terminate (so `task1` or `svsm test task` must wait `task2`). I will open a PR for this where we can discuss this better. (see #1115)

[1] https://github.com/n-ramacciotti/svsm/actions/runs/27494307809/job/81265414543?pr=22

Makefile
stefano-garzarella: `SVSM_ARGS_TEST` contains args to build SVSM's tests, indeed, for example, for the `test` target, we use it only for `svsm` packages, so we should not use it in `bin/test-userinit.elf` target IMO:

```
test:
	cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --package svsm
	cargo test ${CARGO_ARGS} --workspace --exclude svsm
```

stefano-garzarella: > A better solution would be to avoid to call back into the `makefile`, as @00xc suggested in #909:
> `bin/test-kernel.elf is still built from the Makefile for now, since it requires a special command line invocation, and moving that logic into xbuild would introduce too much complexity. Thus, we have xbuild call back into the Makefile to build it `

yep, agree on this, but if it's too complicated and the problem is just `FEATURES_TEST` default for a specific target, what about this:
```
bin/test-userinit.elf: FEATURES_TEST=
```

IMO we should keep target-svsm and target-user* separated.

user/init/src/main.rs
stefano-garzarella: > Do you mean a dummy crate that contains only tests (for that crate), which are called by the (userlib) runner when the crate is compiled for testing and it is not used in normal builds?

yep, something like that.

n-ramacciotti: Ok, I will move the tests from `init` into the new crate. 

However, it is still possible that the init function will have tests in the future and will have to follow this structure. (I'll drop this for now in its crate.)

stefano-garzarella: > However, it is still possible that the init function will have tests in the future and will have to follow this structure. (I'll drop this for now in its crate.)

yeah, absolutely!

Makefile
n-ramacciotti: > yep, agree on this, but if it's too complicated and the problem is just `FEATURES_TEST` default for a specific target, what about this:
> 
> ```
> bin/test-userinit.elf: FEATURES_TEST=
> ```

IIUC, this is a similar solution to the `MODULE_ARGS_TEST`, but in this case `FEATURE_TEST` would always be empty while with `MODULE_ARGS_TEST` you could still specify them (`FEATURES_TEST=.. make bin/test-userinit.elf`)


In this case `SVSM_ARGS_TEST` is used only for `make test` and it contains (by default) the same features as `make bin/test-kernel.elf`

n-ramacciotti: > IMO we should keep target-svsm and target-user* separated.

We can separate them, but then we would have to specify a new variable for each one of them.
(only `USERINIT_ARGS_TEST` for now) and also `xbuild` should be updated accordingly (to set that value) as it will be possible to have different feature for each usermodule and a single `USERMODULE_ARGS_TEST` would not help


We should also concatenate all these variables into `--feature` of `xbuild`

stefano-garzarella: > > yep, agree on this, but if it's too complicated and the problem is just `FEATURES_TEST` default for a specific target, what about this:
> > ```
> > bin/test-userinit.elf: FEATURES_TEST=
> > ```
> 
> IIUC, this is a similar solution to the `MODULE_ARGS_TEST`, but in this case `FEATURE_TEST` would always be empty while with `MODULE_ARGS_TEST` you could still specify them (`FEATURES_TEST=.. make bin/test-userinit.elf`)
> 
> In this case `SVSM_ARGS_TEST` is used only for `make test` and it contains (by default) the same features as `make bin/test-kernel.elf`

Sorry, I didn't get.

Did you try it?

You can still specify `FEATURE_TEST=foo make` and `FEATURE_TEST` will be set to `foo`.

n-ramacciotti:
v13:
- addressed @stefano-garzarella's comments:
  - added a new dummy crate for testing:
    - I copy paste `init`, so there might be some leftover. I pushed to understand if this is the desired approach
    - I don't like `userdummy_test` as the name of the crate, but neither `test_runner` as IMHO is misleading.
- dropped `test_fpu_context_switch` (see #1115 to understand why) 

Now QEMU in CI works!

Makefile
stefano-garzarella: > > IMO we should keep target-svsm and target-user* separated.
> 
> We can separate them, but then we would have to specify a new variable for each one of them. (only `USERINIT_ARGS_TEST` for now) and also `xbuild` should be updated accordingly (to set that value) as it will be possible to have different feature for each usermodule and a single `USERMODULE_ARGS_TEST` would not help
> 
> We should also concatenate all these variables into `--feature` of `xbuild`

Yes, or use 2 different makefiles, one for kernel and one for userspace.

n-ramacciotti:
v14:
- Addressed @stefano-garzarella's comment:
  - Introduced `user/Makefile`, it is now possible to:
    - (a) specify `FEATURES_TEST=svsm:feature1,userdummy:feature2 make test-in-svsm` (from cli)
    - (b) specify `FEATURES_TEST=feature2 make bin/test-userdummy.elf` (from cli)
    - (c) specify default features in `USERDUMMY_FEATURES_TEST ?= features2`  in user makefile
    - (d) specify `feature2` in config recipe
  
 In all these cases, `feature2` will be assigned to `userdummy`. 
 
 These methods can  be used together (tested locally), with the exception of :
 - (b) and (a) as they are both from cli. (Even though `xbuild` will internally call `make bin/test-userdummy.elf` passing its FEATURES_TEST)
 - (b) and (d) as recipe is not used on `bin/test-userdummy.elf` on its own.

n-ramacciotti:
v15:
 - Rebased on latest main
 - Removed [hack] commit that commented out `test_fpu_context_switch`: In this moment the runs after `kernel tests passed`, but for the moment it doesn't fail. (#1115 for that fix)
 - used `::` instead of `:` in `dummy/build.rs` as [1]:
 `
  MSRV: 1.77 is required for cargo::KEY=VALUE syntax. To support older versions, use the cargo:KEY=VALUE syntax.
 `

[1] https://doc.rust-lang.org/cargo/reference/build-scripts.html

todo:
- [x] The only part missing now is retrieving the exit code. Hopefully, tomorrow I'll post an RFC patch for that
- [x] Find a good name for `userdummy`: this is **just a placeholder name**. The crate will be used to test userspace in general (syscall included).  IMHO, `test_runner` is not a good name as it seems that it is the runner, while in reality the runner is in `userlib/testing`. In fact, every other userspace crate will be able to be compiled in userspace testing mode (following the conventional structure `test-in-svsm` which for the moment is only in `userdummy`)

[@stefano-garzarella]

luigix25:
why is this still marked as draft?

stefano-garzarella:
> why is this still marked as draft?

From https://github.com/coconut-svsm/svsm/pull/926#issuecomment-4718693013

> todo:
> 
>     * [ ]  The only part missing now is retrieving the exit code. Hopefully, tomorrow I'll post an RFC patch for that

n-ramacciotti:
> why is this still marked as draft?

The only part missing now is retrieving the exit code. Hopefully, tomorrow I'll post an RFC patch for that
(I updated the patch update description)

Technically, this works, but it is better to avoid the changes in `sys_exit`

Makefile
00xc: > A better solution would be to avoid to call back into the makefile, as @00xc suggested in https://github.com/coconut-svsm/svsm/pull/909:
bin/test-kernel.elf is still built from the Makefile for now, since it requires a special command line invocation, and moving that logic into xbuild would introduce too much complexity. Thus, we have xbuild call back into the Makefile to build it

I am actually working on moving everything into xbuild, not sure if it will work but let's see

n-ramacciotti: > I am actually working on moving everything into xbuild, not sure if it will work but let's see

@00xc Great! 

About `xbuild`: https://github.com/coconut-svsm/svsm/pull/926#discussion_r3404368449

kernel/src/svsm.rs
n-ramacciotti: Does it make sense or values hardcoded in the code would be enough? Maybe this could be addressed in a future PR, using hardcoded values for now

00xc: > My idea was to modify xbuild by adding a new field to the recipe:
> 
> ```
> /// Components for the filesystem image.
> #[derive(Debug, Clone, Deserialize)]
> pub struct FsConfig {
>     modules: HashMap<String, ComponentConfig>,
>     ##
>     files : ..
>     ##
> }
> ```
> 
> I should look into these more closely, though.

Could we just place test binaries in a dedicated path? We already treat `/init` specially, we could have `/test/*` with all test binaries.

Makefile
00xc: > > I am actually working on moving everything into xbuild, not sure if it will work but let's see
> 
> @00xc Great!

Please take a look at #1127

> About `xbuild`: [#926 (comment)](https://github.com/coconut-svsm/svsm/pull/926#discussion_r3404368449)

Added a comment there

kernel/src/svsm.rs
n-ramacciotti: Oh right! That's even better. Thanks for the suggestion!

n-ramacciotti:
v16:
- rebased on latest main and removed user makefile as now `test` build are handled directly by `xbuild`
- the todos from v15 are still relevant https://github.com/coconut-svsm/svsm/pull/926#issuecomment-4718693013

n-ramacciotti:
v17:
- Retrieved `exit_status` and exit base on the result
- Changed name from `userdummy` to `kapi_tester`

Now it's ready for review!

kernel/src/svsm.rs
```diff
@@ -623,7 +623,43 @@ pub unsafe fn svsm_main(li: usize) {
...
+
+#[cfg(test)]
+fn user_tests_in_svsm() -> Result<(), SvsmError> {
+    use alloc::format;
+    use svsm::fs::{list_dir, opendir};
+    use svsm::task::TaskExitStatus;
+    use svsm::task::{exec_user, wait_for_termination};
+
+    let files = list_dir("/test")?;
+    log::info!("Files: {files:?}");
```
luigix25: probably this can be removed

user/kapi_tester/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+#[cfg(not(test))]
+compile_error!(
+    "This crate is only meant to be used for testing purposes. \
+    It should not be built outside of a test context."
+);
+
+#[cfg(not(test))]
+fn main() -> u32 {
+    0
+}
```
luigix25: can we drop this main?

luigix25 (commented):
just nits, rest LGTM!

kernel/src/svsm.rs
n-ramacciotti: Yes, thanks!

n-ramacciotti:
v18:
- addressed @luigix25's comments:
  - removed unnecessary main for non test builds
  - removed `log::info`

kernel/src/svsm.rs
```diff
@@ -629,9 +629,36 @@ fn test_in_svsm_task(_context: usize) {
...
+    use svsm::task::TaskExitStatus;
+    use svsm::task::{exec_user, wait_for_termination};
+
+    for file in list_dir("/test")? {
+        log::info!("Running tests in svsm for {file}");
+        let task = exec_user(&format!("test/{file}"), opendir("/")?)?;
+        wait_for_termination(task.clone());
+        let exit_status = task.get_exit_status()?;
+        if exit_status != TaskExitStatus::Exited(0) {
+            return Err(SvsmError::Test);
```
stefano-garzarella: TBH I don't like to add a new `SvsmError::Test` just for test. IMO we should avoid that.

Also I think it's useful to print the TaskExitStatus value we received for debugging purpose.

What about the following options:
1. move `exit(QEMUExitValue::Fail)` in `user_tests_in_svsm()` , use `unwrap()` instead of `?` (the panic handler should already call `exit(QEMUExitValue::Fail)` ) and remove the return value from user_tests_in_svsm() like we do for kernel_tests_in_svsm()
2. return another value already defined like `SvsmError::Fault`

I prefer 1, it's perfectly fine to panic in a test (calling unwrap, etc.) in a case of failure. I also think we don't even need to move `exit(QEMUExitValue::Fail)`, just add an assert on the `exit_status`.

kernel/src/svsm.rs
```diff
@@ -629,9 +629,36 @@ fn test_in_svsm_task(_context: usize) {
...
 }
 
+#[cfg(test)]
+fn user_tests_in_svsm() -> Result<(), SvsmError> {
+    use alloc::format;
+    use svsm::fs::{list_dir, opendir};
+    use svsm::task::TaskExitStatus;
+    use svsm::task::{exec_user, wait_for_termination};
+
+    for file in list_dir("/test")? {
```
stefano-garzarella: if the `/test` is missing, we are going to fail, right?

Is this intentional, or we should treat is like we don't have any test to run?

user/kapi_tester/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "kapi_tester"
```
stefano-garzarella: `kapi` is confusing as name IMO.
In Linux we use the opposite (e.g. `uapi` to refer to the API exposed by the kernel to the userspace).

kernel/src/svsm.rs
```diff
@@ -629,9 +629,36 @@ fn test_in_svsm_task(_context: usize) {
...
+        exit(QEMUExitValue::Fail);
+    }
+
+    log::info!("All tests passed!");
+
     exit(QEMUExitValue::Success);
 }
 
+#[cfg(test)]
+fn user_tests_in_svsm() -> Result<(), SvsmError> {
```
stefano-garzarella: What about moving this in `kernel/src/testing.rs` ?

kernel/src/svsm.rs
n-ramacciotti: > we should treat is like we don't have any test to run?

Right, it's better to assume that there are no tests

user/kapi_tester/Cargo.toml
n-ramacciotti: I will change it to `uapi`

kernel/src/svsm.rs
n-ramacciotti: Makes sense!

kernel/src/svsm.rs
n-ramacciotti: I'll change the code to use option 1 with the assertion

n-ramacciotti:
v19:
- addressed @stefano-garzarella's comments:
  - moved `user_tests_in_svsm` in `testing.rs`
  - changed `kapi` to `uapi`
  - Assumed that if `/test` is not available there are no tests to run
  - introduced assertion in `user_test_in_svsm` and removed return value

kernel/src/testing.rs
```diff
@@ -86,9 +91,28 @@ pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
...
-    info!("All tests passed!");
+    info!("Kernel tests in svsm passed!");
+}
 
-    exit(QEMUExitValue::Success);
+/// Executes the user binaries listed in `/test` dir sequentially, checks their exit
+/// value and panics if the binary exits with a code different from zero.
+/// If the directory is not present, it is assumed that there are no tests to run.
+pub fn user_tests_in_svsm() {
+    if let Ok(files) = list_dir("/test") {
```
stefano-garzarella: nit: What about this? (I found it more readable, but not a strong opinion)

```suggestion
    let Ok(files) = list_dir("/test") else {
        log::info!("/test dir not found, skipping user tests in svsm");
        return;
      };

      for file in files {
```

Makefile
```diff
@@ -157,13 +157,14 @@ endif
...
 clippy:
-	${CARGO} clippy ${CLIPPY_OPTIONS} --workspace --exclude svsm --exclude stage1 --exclude svsm-fuzz -- ${CLIPPY_ARGS}
+	${CARGO} clippy ${CLIPPY_OPTIONS} --workspace --exclude svsm --exclude stage1 --exclude svsm-fuzz --exclude uapi_tester -- ${CLIPPY_ARGS}
 	RUSTFLAGS="--cfg fuzzing" ${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm-fuzz -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm --target x86_64-unknown-none -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package bldr --target x86_64-unknown-none -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package stage1 --target x86_64-unknown-none -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --workspace --tests --exclude svsm -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm --tests -- ${CLIPPY_ARGS}
+	${CARGO} clippy ${CLIPPY_OPTIONS} --package uapi_tester --tests -- ${CLIPPY_ARGS}
```
stefano-garzarella: Do we really need this?
I mean my doubt if this is already covered by line 165

user/uapi_tester/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_memory_check() {
+        // SAFETY: Single-threaded process, so no data races. Safe to access global
+        // mutable data.
+        unsafe {
+            write(&mut *addr_of_mut!(SOME_DATA), 0xcc);
+            write(&mut *addr_of_mut!(SOME_BSS_DATA), 0xaa);
+            check(&*addr_of!(SOME_DATA), 0xccu64);
+            check(&*addr_of!(SOME_RO_DATA), 0xeeu64);
+            check(&*addr_of!(SOME_BSS_DATA), 0xaa);
```
stefano-garzarella: nit: not strong opinion, and also not sure if it makes the code better, but should we take this opportunity to replace `addr_of` and `addr_of_mut` ?

See https://doc.rust-lang.org/core/ptr/macro.addr_of.html
> addr_of!(expr) is equivalent to &raw const expr. The macro is soft-deprecated; use &raw const instead.

user/uapi_tester/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[[bin]]
+name = "uapi_tester"
+path = "src/main.rs"
+test = false
+doctest = false
+
+[lib]
+test = true
+doctest = true
```
stefano-garzarella, line 14: Not sure if it's a good idea. IIRC `doctest`s use std, but maybe I'm missing something and there is a reason to set this to true.

stefano-garzarella (commented):
LGTM, mostly nit comments plus the makefile rule which I'm not sure.

user/uapi_tester/Cargo.toml
n-ramacciotti: From an old conversation in this PR (https://github.com/coconut-svsm/svsm/pull/926#discussion_r2640448089): By default these values are already set to true, so this section can be removed. It is here for clarity and for consistency with `kernel/Cargo.toml`, which contains the same section.

Makefile
n-ramacciotti: Oh, right!

n-ramacciotti:
v20:
- addressed @stefano-garzarella's comments:
  - Introduced early return in `user_tests_in_svsm`
  - Removed redundant rule in `make clippy`
  - Replaced `addr_of` and `addr_of_mut` with `&raw const` and `&raw mut`. I introduced some unnecessary parentheses, that, IMO, improves readability.

user/uapi_tester/Cargo.toml
stefano-garzarella: @n-ramacciotti I see, if we have in `kernel/Cargo.toml` let's leave them, we can figure out later if there are issues.

stefano-garzarella (approved):

joergroedel (approved):

PR #927, opened 2025-12-19
userspace: introduce coconut-alloc as allocator

n-ramacciotti:
This PR introduces heap allocation support in userspace.

The heap is a single 64KiB static block initialized by `userlib` at startup before calling the `main` function. Atomic state operations ensure that only a single initialization can occur.
The PR implements the `GlobalAlloc` trait by exploiting `AllocBlock` from [coconut-alloc](https://github.com/coconut-svsm/coconut-alloc), which supports lockless allocations up to 32KiB using atomic operations.

##
This PR is marked as draft as it depends on the changes from:
- [#4](https://github.com/coconut-svsm/coconut-alloc/pull/4) from coconut-alloc: this ~~removes an unstable feature not available in the current toolchain (1.86.0) and~~ adds a new utility function required in this PR
- ~~[#937](https://github.com/coconut-svsm/svsm/pull/937): this updates rust toolchain to 1.87.0  as required by coconut-alloc~~

These changes were originally included in [#913](https://github.com/coconut-svsm/svsm/pull/913), which also introduced test in userspace svsm, but as suggested, I split them into multiple PRs.

### DIFF
--- a/user/init/src/main.rs
+++ b/user/init/src/main.rs
@@ -2,6 +2,8 @@
 #![no_main]
 
 use userlib::*;
+extern crate alloc;
+use alloc::vec::Vec;
 
 use core::ptr::{addr_of, addr_of_mut};
 
@@ -37,5 +39,13 @@ fn main() -> u32 {
         check(&*addr_of!(SOME_RO_DATA), 0xeeu64);
         check(&*addr_of!(SOME_BSS_DATA), 0xaa);
     }
+
+    let mut v: Vec<u64> = Vec::new();
+    for i in 0..256 {
+        v.push(i);
+    }
+
+    println!("Vector length: {}", v.len());
+
     0
 }


user/lib/src/alloc.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+const MAX_ALLOC_SIZE: usize = 32 * 1024; // 32 KiB
+const UNINITIALIZED: usize = 0;
+const INITIALIZING: usize = 1;
+const INITIALIZED: usize = 2;
+
+// One time initialization of a static mutable reference:
+// https://doc.rust-lang.org/edition-guide/rust-2024/static-mut-references.html#no_std-one-time-initialization
+static mut HEAP: AllocBlock = AllocBlock::new();
```
joergroedel: This makes the heap be pre-allocated in the data or bss section. COCONUT has an mmap syscall which can map anonymous memory to the process address space, can you try to use this instead?

user/lib/src/alloc.rs
```diff
@@ -0,0 +1,172 @@
...
+static STATE_INITIALIZED: AtomicUsize = AtomicUsize::new(UNINITIALIZED);
```
joergroedel: This seems to be overly complicated. Heap initialization happens when the process runs only a single thread, before even `main` is started. Also I wonder if a `OnceCell` would be a more idiomatic solution to tracking initialization state.

user/lib/src/alloc.rs
```diff
@@ -0,0 +1,172 @@
...
+        AllocBlock::initialize(heap);
+    }
+    STATE_INITIALIZED.store(INITIALIZED, Ordering::SeqCst);
+    Ok(())
+}
+
+fn get_global_heap() -> Result<&'static AllocBlock, AllocError> {
+    // fixme: this check can be removed.
+    if STATE_INITIALIZED.load(Ordering::SeqCst) != INITIALIZED {
+        println!("Heap not initialized");
```
joergroedel: No need for all these `println!` messages.

user/lib/src/alloc.rs
```diff
@@ -0,0 +1,172 @@
...
+// AllocBlock uses atomic operations internally to handle concurrent alloc/free safely.
+unsafe impl GlobalAlloc for SvsmUserAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        if layout.size() == 0 || layout.size() > MAX_ALLOC_SIZE {
+            return ptr::null_mut();
+        }
+
+        let Ok(heap) = get_global_heap() else {
+            return ptr::null_mut();
+        };
```
joergroedel, line 69: I am fine with just panic'ing when code tries to allocate something while heap is not initialized.

joergroedel (commented):
Thanks for your contribution, this looks very promising and I am happy to see progress on the user-mode allocator implementation. Left some comments to consider.

user/lib/src/alloc.rs
n-ramacciotti: Thank you for the feedback. I have looked into SVSM, but I cannot find a  `mmap` syscall. I did find a `mmap_user()` function that allows mapping anonymous memory for the process. Please let me know if I’m wrong.

If  `mmap` syscall is indeed not available, I was wondering if the following alternative approach would be acceptable:

- Map the heap directly in `exec_user()` at a fixed and known address.

Otherwise, if you prefer to implement a proper `mmap` syscall, I would be happy to help with that.

user/lib/src/alloc.rs
n-ramacciotti: > This seems to be overly complicated.  

I agree with that.

>Also I wonder if a `OnceCell` would be a more idiomatic solution to tracking initialization state.

I'll start looking into a better solution

user/lib/src/alloc.rs
joergroedel: HEy @n-ramacciotti, I thought `MMAP` support was already in the upstream project, but apparently it was only in one of my local branches. I re-based the code and pushed it as a draft PR in #945. Please have a look and check if it works for you.

PR #928, opened 2025-12-19
kernel: move page table construction into stage2

msft-jlange:
Now that stage2 has the ability to allocate memory within the kernel heap, the kernel page tables configured by stage2 can be allocated directly from the kernel heap, avoiding the need to build them in stage2 and rebuild them in the kernel.  This also eliminates the duplicated page table construction code and the bugs that arise when these two code paths diverge.

### DIFF
--- a/kernel/src/stage2.rs
+++ b/kernel/src/stage2.rs
@@ -353,13 +353,6 @@ fn setup_kernel_page_tables<'a>(heap: &mut KernelHeap) -> Result<KernelPageTable
     };
     *current_paging_root.entry_mut(pml4e_index) = *pml4e;
 
-    // Also place a mapping to low memory in the new kernel page table so it
-    // can access stage2 pages during the early part of boot.
-    paging_root.entry_mut(0).set_unrestricted(
-        make_private_address(current_paging_root.entry_mut(0).address()),
-        pxe_flags,
-    );
-
     // Allocate a new page to use as a page directory table page.  This will
     // span 1 GB of address space, which is the maximum that will ever be
     // addressible during stage 2 execution.


msft-jlange:
This boots correctly on Hyper-V, both native and SNP, both with and without vTOM, so I will probably need some help identifying what's going wrong on the QEMU configuration.

peterfang:
If this is more urgent, I can also test it after this gets merged.

kernel/src/mm/ptguards.rs
```diff
@@ -33,6 +33,8 @@ impl PerCPUPageMappingGuard {
     /// * `paddr_start` - The starting physical address of the range.
     /// * `paddr_end` - The ending physical address of the range.
     /// * `alignment` - The desired alignment for the mapping.
+    /// * `shared` - Indicates whether the address describes a shared page or a
```
00xc: Maybe it should be an enum for readability, something like `Visibility::Shared` / `Visibility::Private` (although in the long run this should be a different type like I proposed in #751).

msft-jlange: Done properly, this should align with `platform::PageValidateOp` to minimize the number of distinct enums, or to align with the memory management primitives you allude to - and both of those are beyond the scope of what this PR aims to achieve.

It looks like this change will ultimately not be necessary for this PR to work correctly, so I will likely just remove this altogether, but I need to do some more work to confirm that this will be possible.

kernel/src/mm/transition_pgtable.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl TransitionPageTable {
+    /// # Safety
+    /// This must only be called during early boot when the SIPI stub page
+    /// table is known not to be used for any other purpose.
+    pub unsafe fn new() -> Result<Self, SvsmError> {
+        // Obtain a virtual address mapping the current page tables so they
+        // can be examined.
+        let current_pt_root =
+            // SAFETY: the current paging root can always be safely mapepd for
```
peterfang: typo: mapped

kernel/src/mm/transition_pgtable.rs
```diff
@@ -0,0 +1,94 @@
...
+            paddr = paddr + PAGE_SIZE_1G;
+        }
+
+        // Set the first entry in the root page to point to the allocated
+        // page table page.
+        paging_root.as_mut()[0].set(make_private_address(pdp_paddr), pd_flags);
+
+        // Copy the kernel entry from the current page table into the stub
+        // page table.
+        paging_root.as_mut()[PGTABLE_LVL3_IDX_SHARED] =
```
peterfang: `.as_mut()` is not needed here

kernel/src/mm/transition_pgtable.rs
```diff
@@ -0,0 +1,94 @@
...
+        let pt_flags = pd_flags | PTEntryFlags::HUGE;
+        let mut paddr = make_private_address(PhysAddr::new(0));
+        for index in 0..3 {
+            pdp_root[index].set(paddr, pt_flags);
+            paddr = paddr + PAGE_SIZE_1G;
+        }
+
+        // Set the first entry in the root page to point to the allocated
+        // page table page.
+        paging_root.as_mut()[0].set(make_private_address(pdp_paddr), pd_flags);
```
peterfang: `.as_mut()` is not needed here

kernel/src/mm/transition_pgtable.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        // Copy the kernel entry from the current page table into the stub
+        // page table.
+        paging_root.as_mut()[PGTABLE_LVL3_IDX_SHARED] =
+            current_pt_root.as_ref()[PGTABLE_LVL3_IDX_SHARED];
```
peterfang: `.as_mut()` and `.as_ref()` are not needed here

kernel/src/mm/pagetable.rs
```diff
@@ -389,20 +396,30 @@ impl PTEntry {
...
+    /// Allocates a zeroed pagetable page and returns a `PageBox` containing
+    /// the allocation.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SvsmError`] if the page cannot be allocated.
+    pub fn alloc_box() -> Result<PageBox<Self>, SvsmError> {
+        PageBox::try_new_zeroed()
+    }
+
```
peterfang, line 414: Is this for an upcoming change? I don't see it used anywhere else

kernel/src/stage2.rs
```diff
@@ -88,11 +103,56 @@ impl KernelHeap {
...
+        self.virt_base.unwrap_or(self.local_virt_base) + offset
+    }
+
+    pub fn remap_memory(
+        &mut self,
+        vaddr: VirtAddr,
+        page_tables: &mut KernelPageTables<'_>,
+    ) -> Result<(), SvsmError> {
+        // Map the heap physical memory span at the requested address in the
+        // kernel page tables.  All memory has already been validated, so only
```
peterfang, line 121: Since there's an assumption here (All memory has already been validated), do we want to make this function `unsafe`? I understand that it's only used by stage2 so this is not a strong opinion

kernel/src/mm/transition_pgtable.rs
```diff
@@ -0,0 +1,94 @@
...
+        let current_pt_root =
+            // SAFETY: the current paging root can always be safely mapepd for
+            // reading.
+            unsafe { PerCPUMapping::<PTPage>::create(make_private_address(read_cr3()))? };
+
+        // Map two pages to be used as initial page tables: one as a paging
+        // root and one as the initial page directory page.  These are known
+        // to be usable because low memory was accepted by stage2.
+        let paging_root_paddr = PhysAddr::new(SIPI_STUB_PT_GPA as usize);
+        let pdp_paddr = paging_root_paddr + PAGE_SIZE;
```
peterfang, line 50: Do we need to use `make_private_address()` here?

kernel/src/stage2.rs
```diff
@@ -101,9 +161,11 @@ impl KernelHeap {
         }
 
         // Calculate the allocation base based on the current position within
-        // the heap.
+        // the heap.  If no heap virtual address has been configured yet,
+        // then
```
peterfang: Looks like something is missing here

kernel/src/stage2.rs
```diff
@@ -88,11 +103,56 @@ impl KernelHeap {
...
+
+        Ok(())
+    }
+
     pub fn allocate(&mut self, size: usize) -> Result<(VirtAddr, PhysAddr), SvsmError> {
         let page_count = round_to_pages(size);
         self.allocate_pages(page_count)
     }
 
+    pub fn allocate_zeroed(&mut self, size: usize) -> Result<(VirtAddr, PhysAddr), SvsmError> {
```
peterfang, line 145: Do we actually need to do zeroing? `prepare_heap()` ensures that the kernel heap area is all-zero after page validation and there is no `free()` here to mess things up

kernel/src/stage2.rs
```diff
@@ -112,6 +174,208 @@ impl KernelHeap {
...
+    let pml4e = paging_root.entry_mut(pml4e_index);
+    pml4e.set_unrestricted(make_private_address(pdpt_paddr), pxe_flags);
+
+    // Set the PML4E in the current page table as well so the
+    // kernel address space is also visible in the current address space.
+    // SAFETY: the physical address of the current paging root is known to be
+    // identity-mapped in the current address space and therefore that address
+    // can be used to obtain a page table view.
+    let mut current_paging_root = unsafe {
+        let vaddr = VirtAddr::new(u64::from(this_cpu().get_pgtable().cr3_value()) as usize);
```
peterfang: There's a shorter way to do this:
```suggestion
        let vaddr = VirtAddr::new(this_cpu().get_pgtable().cr3_value().bits());
```

kernel/src/stage2.rs
```diff
@@ -342,6 +614,22 @@ unsafe fn load_elf_segment(
...
+fn count_elf_pages(elf: &Elf64File<'_>) -> usize {
+    // Enumerate the segments of this ELF file to count the total amount of
+    // physical memory required.
+    let mut page_count: usize = 0;
+    let vaddr_alloc_info = elf.image_load_vaddr_alloc_info();
+    let vaddr_alloc_base = vaddr_alloc_info.range.vaddr_begin;
+    for segment in elf.image_load_segment_iter(vaddr_alloc_base) {
+        let segment_size =
+            (segment.vaddr_range.vaddr_end - segment.vaddr_range.vaddr_begin) as usize;
+        page_count += segment_size.div_ceil(PAGE_SIZE);
```
peterfang:
```suggestion
        page_count += round_to_pages(segment_size);
```

kernel/src/stage2.rs
```diff
@@ -342,6 +614,22 @@ unsafe fn load_elf_segment(
...
+/// Calculates the number of physical pages required to load an ELF file.
+fn count_elf_pages(elf: &Elf64File<'_>) -> usize {
+    // Enumerate the segments of this ELF file to count the total amount of
+    // physical memory required.
+    let mut page_count: usize = 0;
+    let vaddr_alloc_info = elf.image_load_vaddr_alloc_info();
+    let vaddr_alloc_base = vaddr_alloc_info.range.vaddr_begin;
+    for segment in elf.image_load_segment_iter(vaddr_alloc_base) {
+        let segment_size =
+            (segment.vaddr_range.vaddr_end - segment.vaddr_range.vaddr_begin) as usize;
```
peterfang, line 626: The check in `load_elf_segment()` can be moved here to ensure the ELF file is sane:

```
    // All ELF segments should be aligned to the page size. If not, there's
    // the risk of pvalidating a page twice, bail out if so. Note that the
    // ELF reading code had already verified that the individual segments,
    // with bounds specified as in the ELF file, are non-overlapping.
    if !segment_start.is_page_aligned() {
        return Err(SvsmError::Elf(ElfError::UnalignedSegmentAddress));
    }
```

kernel/src/stage2.rs
```diff
@@ -454,35 +754,37 @@ fn load_igvm_params(
...
-    loaded_kernel_pregion: MemoryRegion<PhysAddr>,
-    heap_base_vaddr: VirtAddr,
+    kernel_page_count: usize,
     platform: &dyn SvsmPlatform,
     config: &SvsmConfig<'_>,
 ) -> Result<KernelHeap, SvsmError> {
-    // Heap starts after kernel
-    let heap_pstart = loaded_kernel_pregion.end();
+    let kernel_size = kernel_page_count * PAGE_SIZE;
+    let heap_pstart = kernel_region.start() + kernel_size;
```
peterfang, line 766: Is `heap_pstart` just `kernel_region.start()`? I'm confused by the calculation below because it looks like `kernel_size` got taken out of `kernel_region` twice

kernel/src/svsm.rs
```diff
@@ -257,26 +264,16 @@ unsafe fn svsm_start(li: *const KernelLaunchInfo) -> Option<VirtAddr> {
...
-    let kernel_elf_buf = unsafe { slice::from_raw_parts(kernel_elf_buf_ptr, kernel_elf_len) };
-    let kernel_elf = match elf::Elf64File::read(kernel_elf_buf) {
-        Ok(kernel_elf) => kernel_elf,
-        Err(e) => panic!("error reading kernel ELF: {}", e),
+    paging_init(platform, false).expect("Failed to initialize paging");
+
+    // SAFETY: the current page table was allocated by stage2 from the kernel
+    // heap and therefore it can be built into a PageBox.
+    let init_pgtable: PageBox<PageTable> = unsafe {
+        let page_table_ptr = (launch_info.kernel_page_table_vaddr as usize) as *mut PageTable;
```
peterfang, line 272: Is `launch_info.kernel_page_table_vaddr` needed here? We can just call `phys_to_virt(current_cr3)`

kernel/src/stage2.rs
msft-jlange: If the memory hasn't been validated, then the worst that will happen is an exception and not corruption, so I don't see this as having the potential to violate memory safety.

kernel/src/mm/transition_pgtable.rs
msft-jlange: It appears you are correct.  These went in originally because the compiler was generating errors, but perhaps this is fixed with the 2024 edition.

kernel/src/mm/pagetable.rs
msft-jlange: It's used in the function `alloc()`.
```
let page = Self::alloc_box()?;
```

kernel/src/stage2.rs
msft-jlange: On SNP, page validation does not zero the contents of pages.  Zeroing is only performed when the guest OS requests PVALIDATE in the lower VMPL, but not when the SVSM validates its own pages.

This function will be removed in the near future so it's not worth considering optimizations.

kernel/src/stage2.rs
msft-jlange: Perhaps, but all of the ELF loading code will be removed from stage2 in my next PR, so is your suggestion really worth it?  This PR tries to minimize the overall code delta for logic that's destined for removal.

kernel/src/stage2.rs
msft-jlange: The term "kernel" is ambiguous throughout this logic.  "Kernel region" refers to the GPA range that is under exclusive control of the SVSM.  "Kernel size" refers to the size of the loaded kernel ELF, which is only a portion of the kernel region.  The kernel region is effectively divided into two portions: the kernel ELF and the page heap, in that order.  So this is calculating the base physical address of the heap as the first page within the kernel region that follows the loaded kernel image.  This is basically the same logic that the old code used, but it's now done based on kernel page count (which is needed for other reasons) instead of a snapshot of the physical address span consumed by the loaded kernel.

This code is also destined to be removed from stage2 in the next PR.

kernel/src/svsm.rs
msft-jlange: Using `phys_to_virt()` relies on `FIXED_MAPPING`, which is going to go away at some point once we remove the direct physical map of kernel memory (removing the direct map is desirable for many reasons).  Thus we should not be taking any new dependencies on it continuing to work.

kernel/src/mm/transition_pgtable.rs
msft-jlange: The existing convention is that all physical addresses exist without regard to private/shared bits, and the private/shared bits are only included at the time a mapping is created.  In this case, these physical addresses are in the domain of non-decorated addresses, and the per-CPU mapping that is created for them is the place where the PTEs are marked as private.

peterfang: Ah, I see. Thanks for the explanation

kernel/src/stage2.rs
peterfang: But here `heap_pregion`'s starting address is `heap_pstart + kernel_size`, which is essentially `kernel_region.start() + kernel_size + kernel_size` and this confuses me. `heap_size` is `kernel_region.end() - heap_pstart - kernel_size - ...` which is essentially `kernel_region.end() - kernel_region.start() - kernel_size - kernel_size - ...` right?

msft-jlange: I see what you mean - I misunderstood your comment because it was associated with the lines above, which appear correct, but you're pointing out that the construction of `heap_pregion` below is where the double-add occurs.  I agree that the accounting is wrong and needs to be fixed.

peterfang: Thanks for confirming that. I think `heap_size` has the same problem - there's no need to do `.and_then(|r| r.checked_sub(kernel_size))`

msft-jlange: Agreed.

PR #929, opened 2025-12-22
vpu/vc: panic on a page not validated #VC

msft-jlange:
The SVSM is designed to validate all pages before attempting to use any of them.  If a #VC exception occurs due to a page not being validated, then the system should panic.  This must be checked before the system attempts to emulate the faulting instruction, since the emulation attempt may be incorrect if trying to read through any non-validated pages.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -30,6 +30,7 @@ pub const SVM_EXIT_CPUID: usize = 0x72;
 pub const SVM_EXIT_IOIO: usize = 0x7b;
 pub const SVM_EXIT_MSR: usize = 0x7c;
 pub const SVM_EXIT_RDTSCP: usize = 0x87;
+pub const SVM_EXIT_PAGE_NOT_VALIDATED: usize = 0x404;
 pub const X86_TRAP_DB: usize = 0x01;
 pub const X86_TRAP: usize = SVM_EXIT_EXCP_BASE + X86_TRAP_DB;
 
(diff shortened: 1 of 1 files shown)


kernel/src/cpu/vc.rs
```diff
@@ -118,6 +119,14 @@ pub fn stage2_handle_vc_exception_no_ghcb(ctx: &mut X86ExceptionContext) -> Resu
 pub fn stage2_handle_vc_exception(ctx: &mut X86ExceptionContext) -> Result<(), SvsmError> {
     let err = ctx.error_code;
 
+    // Panic if the #VC is due to an access to a non-validated page.  The SVSM
+    // kernel will always validate pages prior to any attempt to access
+    // them with C=1, so any such exception indicates either a security failure
+    // or a memory access bug, both of which are fatal.  If the validation
+    // behavior changes in the future, this exit code must be handled specially
+    // rather than attempting to interpret the instruction.
+    assert!(err != SVM_EXIT_PAGE_NOT_VALIDATED);
```
00xc, line 128: We should probably also do this in `stage2_handle_vc_exception_no_ghcb()`. But to avoid having the comment triplicated, maybe we should have this moved into a standalone function.

msft-jlange: Stage2 will be going away soon and when it does, this assert will belong in the single #VC handler that exists instead of requiring a function callout.  I would prefer to leave this as it is to avoid having to rearrange it again once stage2 is removed.  And since #VC doesn't occur in practice in the current stage2 before the GHCB is configured, we're really no worse off by not having the assert there as well in the short term.

stefano-garzarella: @00xc are you okay with resolving this conversation?

stefano-garzarella (approved):

PR #930, opened 2025-12-22
SVSM: add support for symbolized stacktraces

00xc:
Add support to resolve symbol addresses during stack traces.

Before (by adding a panic to `test_nop()`):

```
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 2025.12-devel-36-g34216465+
[SVSM] ERROR: Info: panicked at kernel/src/lib.rs:53:17:
explicit panic
[SVSM] ---BACKTRACE---:
[SVSM]   [ffffff800004bf7d]
[SVSM]   [ffffff8000041aa6]
[SVSM]   [ffffff8000006880]
[SVSM] ---END---
```

With this PR:

```
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 2025.12-devel-45-gdaaf19f6+
[SVSM] ERROR: Info: panicked at kernel/src/lib.rs:53:17:
explicit panic
[SVSM] ---BACKTRACE---:
[SVSM]   [ffffff8000065afd] <svsm::test_nop::{closure#0} as core::ops::function::FnOnce<()>>::call_once+0x1d
[SVSM]   [ffffff800002d896] svsm::testing::svsm_test_runner+0xf6
[SVSM]   [ffffff8000006710] svsm::task::tasks::task_exit+0x0
[SVSM] ---END---
```

## Implementation

The implementation relies on the `.symtab` and `.strtab` sections in the kernel ELF, which are currently not included in the binary we embed in the IGVM file. When adding these sections (by passing the right flag to objcopy), the size of the IGVM file increases by around 20% to 25%, so this is only done for debug builds for now.

The stage2 attempts to parse these sections out of the ELF binary. If present, it heap-allocates a region for each, which is passed into the main SVSM kernel via `KernelLaunchInfo`.
When a stacktrace is performed, this information is queried. If no such information is found, the stacktrace proceeds normally, printing `??` instead of the symbol name.

### DIFF
--- a/kernel/src/debug/stacktrace.rs
+++ b/kernel/src/debug/stacktrace.rs
@@ -8,6 +8,7 @@ use crate::{
     address::{Address, VirtAddr},
     cpu::idt::common::{X86ExceptionContext, is_exception_handler_return_site},
     cpu::percpu::try_this_cpu,
+    debug::symbols::resolve_symbol,
     mm::{STACK_SIZE, STACK_TOTAL_SIZE, SVSM_CONTEXT_SWITCH_STACK, SVSM_STACK_IST_DF_BASE},
     utils::MemoryRegion,
 };
@@ -304,9 +305,19 @@ fn print_stack_frame(frame: StackFrame) {
     if !frame.is_aligned {
         let _ = annotations.write_char('#');
     }
+
     let space = if annotations.is_empty() { "" } else { " " };
 
-    log::info!("  [{:016x}]{}{}", frame.rip, space, annotations.as_str());
+    if let Some(sym) = resolve_symbol(frame.rip) {
+        log::info!(
+            "  [{:016x}]{}{} {sym}",
+            frame.rip,
+            space,
+            annotations.as_str(),
+        );
+    } else {
+        log::info!("  [{:016x}]{}{} ??", frame.rip, space, annotations.as_str());
+    }
 }
 
 pub fn print_stack(skip: usize) {


bootlib/src/symbols.rs
```diff
@@ -0,0 +1,28 @@
+use zerocopy::{FromBytes, Immutable, IntoBytes};
```
msft-jlange, line 7: Please add header comments that include license information at a minimum.

kernel/Cargo.toml
```diff
@@ -66,6 +66,7 @@ vstd = { workspace = true, optional = true}
 verify_proof = { workspace = true, optional = true}
 verify_external = { workspace = true, optional = true}
 verus_stub = { workspace = true }
+rustc-demangle = "0.1.26"
```
msft-jlange: Can you move this dependency into the workspace and inherit here?  That way, if we ever decide to consume the demangler in other crates, we will have a consistent version.  This appears to be the pattern with all of our other depdencies, as it looks to me like every single dependency in the kernel is marked `workspace = true`.

00xc: Done. I also moved the addition and use of this crate into the previous commit, since I implemented `Display` for `ResolvedSym`. The demangling is now invoked in the `Display` implementation, instead of doing so from the stacktrace code.

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+/// from the `KernelHeap`, each containing the information for the `.symtab` and
+/// `.strtab` respectively.
+pub fn load_kernel_symbols(
+    elf: &elf::Elf64File<'_>,
+    heap: &mut KernelHeap,
+) -> (MemoryRegion<VirtAddr>, MemoryRegion<VirtAddr>) {
+    let empty = MemoryRegion::new(VirtAddr::null(), 0);
+
+    let Some((symtab, strtab)) = elf.symtab.as_ref().zip(elf.strtab.as_ref()) else {
+        log::info!("No kernel symbols!");
```
peterfang, line 30: Should this be printed only if `cfg!(debug_assertions)`?

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+    Ok((r1, r2))
+}
+
+/// A type that allows progressively reserving a dynamically-sized portion of
+/// memory.
+struct PageCursor<'a> {
+    region: MemoryRegion<VirtAddr>,
+    pos: usize,
+    heap: &'a mut KernelHeap,
+}
```
peterfang, line 78: This looks like kernel memory infrastructure. Can we move this to somewhere that's closer to `KernelHeap`?

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /// Write some bytes into the cursor, allocating more memory in the process
+    /// if needed.
+    fn write(&mut self, val: &[u8]) -> Result<(), SvsmError> {
+        let new_pos = self.pos + val.len();
+        if let Some(need) = new_pos.checked_sub(self.region.len()) {
+            self.get_pages(need)?;
+        };
```
peterfang, line 96: I feel like `checked_sub()` can be avoided here? `self.region.is_empty()` is just a special case (see the other comment). In other cases we just need something that looks like `if starting_address + val.len() > self.region.end()`

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        let dst = self.region.start() + self.pos;
```
peterfang, line 98: Can we just cache the current starting address instead of doing this calculation every time? This is guaranteed to be contiguous across calls

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Extends the cursor's allocation by requesting pages to satisfy the
+    /// requested additional length.
+    fn get_pages(&mut self, len: usize) -> Result<(), SvsmError> {
+        let pages = round_to_pages(len);
+        let page = self.heap.allocate_pages(pages)?.0;
+        let region = MemoryRegion::new(page, pages * PAGE_SIZE);
+        if self.region.is_empty() {
+            self.region = region;
+            return Ok(());
+        }
```
peterfang, line 123: `self.region.is_empty()` means it's special (first-time allocation). Should we make this explicit by making `self.region` an `Option` or something?

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+fn allocate_kernel_symbols(
+    symtab: &Elf64Symtab<'_>,
+    strtab: &Elf64Strtab<'_>,
+    heap: &mut KernelHeap,
+) -> Result<(MemoryRegion<VirtAddr>, MemoryRegion<VirtAddr>), SvsmError> {
+    let mut dst1 = PageCursor::new(heap);
+    for i in 0..symtab.syms_num() {
+        let sym = symtab.read_sym(i)?;
```
peterfang, line 47: Is it more convenient to just add an iterator implementation for `Elf64Symtab`? Not sure if there are other uses of `syms_num`

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+        // `KSym` stores the size in a u32 since there should be no symbols
+        // spanning more than 4GiB, and this helps shave off 8 bytes from
+        // the struct (there is one per defined symbol, so this adds up to a
+        // reduction of ~100KiB in memory footprint as of writing). Thus, in
+        // the strange case that there is a symbol of this size, simply use the
+        // maximum value of a u32.
+        let size = u32::try_from(sym.st_size).unwrap_or(u32::MAX);
```
peterfang: This is a very weird case so perhaps it deserves a `log::warn!` in addition to making it `u32::MAX`

kernel/src/stage2_syms.rs
```diff
@@ -0,0 +1,124 @@
...
+        dst1.write(KSym::new(sym.st_value as usize, size, sym.st_name).as_bytes())?;
+    }
+    let r1 = MemoryRegion::new(dst1.region.start(), dst1.pos);
```
peterfang: `PageCursor` can implement a function that returns this so we can hide some of its details

bootlib/src/kernel_launch.rs
```diff
@@ -50,6 +50,10 @@ pub struct KernelLaunchInfo {
     pub stage2_igvm_params_phys_addr: u64,
     pub stage2_igvm_params_size: u64,
     pub igvm_params_virt_addr: u64,
+    pub kernel_symtab_start: *const KSym,
+    pub kernel_symtab_len: u64,
+    pub kernel_strtab_start: *const u8,
+    pub kernel_strtab_len: u64,
```
peterfang, line 56: @msft-jlange There are no other raw pointers in `KernelLaunchInfo`. Are you okay with adding these?

kernel/src/debug/symbols.rs
```diff
@@ -0,0 +1,123 @@
...
+                return Some(ResolvedSym::new(name, offset));
+            }
+
+            let Some(diff) = addr.checked_sub(sym.addr) else {
+                continue;
+            };
+            if diff < min_diff {
+                min_diff = diff;
+                let candidate = ResolvedSym::new(self.get_sym_name(sym), addr - sym.addr);
+                nearest_candidate = Some(candidate);
```
peterfang: We can do this after the loop ends to avoid unnecessary `get_sym_name()` calls

kernel/src/debug/symbols.rs
```diff
@@ -0,0 +1,123 @@
...
+        Self { symtab, strtab }
+    }
+
+    /// Gets the name for the given symbol as a C string.
+    fn get_sym_name(&self, sym: &KSym) -> &'static CStr {
+        let slice = &self.strtab[sym.name as usize..];
+        CStr::from_bytes_until_nul(slice).unwrap_or_default()
+    }
+
+    fn resolve_symbol(&self, addr: VirtAddr) -> Option<ResolvedSym> {
```
peterfang, line 63: Will this function ever return `None`?

kernel/src/debug/symbols.rs
```diff
@@ -0,0 +1,123 @@
...
+    // Slices must be aligned to the type of the item they refer to. The strtab
+    // is a byte slice, so it has no alignment requirements, but the symtab
+    // slice has `KSym` items, so check.
+    assert!(symtab_start.is_aligned());
+
+    // SAFETY: we trust the `KernelLaunchInfo` passed by stage2 to contain valid addresses.
+    let (symtab, strtab) = unsafe {
+        (
+            slice::from_raw_parts(symtab_start, li.kernel_symtab_len as usize),
+            slice::from_raw_parts(strtab_start, li.kernel_strtab_len as usize),
```
peterfang, line 111: I think the second argument should be the number of elements

kernel/src/debug/symbols.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// not allocate. The returned type implements [`Display`](fmt::Display),
+    /// so it can be printed in a straightforward way.
+    pub fn demangled_name(&self) -> Demangle<'_> {
+        demangle(self.name.to_str().unwrap_or_default())
+    }
+}
+
+impl fmt::Display for ResolvedSym {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:#}+0x{:x}", self.demangled_name(), self.off)
```
peterfang: We can just use `{:#x}` for `self.off`

kernel/src/stage2_syms.rs
00xc: In the future, we may enable symbols via a specific switch (instead of only including them in debug builds). This is a one-time print so I don't see a downside for keeping it.

kernel/src/stage2_syms.rs
00xc: Yes, but then after the `if` you suggest we need to do the subtraction to compute how much extra space we need anyway. This does both the check and the subtraction in a single step.

kernel/src/stage2_syms.rs
00xc: Do you mean storing `dst` in `self`? This cannot be done as we need to remember the actual beginning of the memory region in order to pass it later from stage2.

kernel/src/stage2_syms.rs
00xc: This would move complexity into the `write()` method, as then we need to handle the special case where the region is `None`, since we cannot do `self.region.len()` directly. The way it is now, the region starts with a length of zero, always triggering the check which ends up calling into `get_pages()`.

This also handles better the (unlikely) case that the ELF sections are present but empty, since we do not have to do any `unwrap()`s.

bootlib/src/kernel_launch.rs
00xc: Raw pointers are FFI-safe, so this should be fine.

kernel/src/stage2_syms.rs
00xc: In the future, this code will live in igvmbuilder, meaning it will be refactored out. Keeping the new code in a separate file will help with the refactor (as the only user of this `PageCursor` type will disappear from stage 2).

kernel/src/stage2_syms.rs
00xc: Adding a single `syms_num()` method and reusing the already existing `read_sym()` seems less intrusive and more convenient to me. If you have a strong opinion on this I can add an iterator.

kernel/src/debug/symbols.rs
00xc: Yes, if `addr` is below the address of any known symbol.

kernel/src/debug/symbols.rs
00xc: Done. I actually fixed this in stage2, so that the value passed reflects the number of items.

bootlib/src/kernel_launch.rs
msft-jlange: Yes, raw pointers are fine.  Someday we should convert a bunch of these other `u64` values to raw pointer types so their intended use is more explicit, and starting that trend now is a good thing.

kernel/src/stage2_syms.rs
msft-jlange: Yes, I asked for this code to be in a separate file so that it is easier to refactor.  I've already split the stage2 heap construction code into a separate library and keeping the new symbol code in a separate file will make the merge resolution infinitely easier.

kernel/src/stage2_syms.rs
peterfang: Good point. I guess obsessing over the cost of `checked_sub()` really isn't worth the extra effort

kernel/src/stage2_syms.rs
peterfang: We could keep `region` and still cache the current `dst` instead of using `pos` right? And you can check it against `self.region.end()` to see if more pages are needed. I'm just wondering if caching the actual address instead of using a position is simpler.

kernel/src/stage2_syms.rs
peterfang: OK, that makes sense

kernel/src/stage2_syms.rs
peterfang: This is probably an optimization that we don't need desperately. We can do it later if there's an obvious gain from using an iterator.

kernel/src/stage2_syms.rs
00xc: If you make `self.pos` a `VirtAddr` then you need to specially handle the case where `self.pos` is NULL, so you have to set it to the beginning of the region once you allocate it for the first time. It also complicates the `checked_sub()` check. It is better to keep an offset into the region, since 0 is always a valid offset.

peterfang: A quick rewrite proved that using `VirtAddr` actually yields the same amount of code. So not an improvement.

kernel/src/stage2_syms.rs
peterfang: What if need = 0? I think the code works in this case but that's because it's able to jump through all the hoops in `get_pages()`. What concerns me especially is that `self.heap.allocate_pages(0)` will actually give you an address even though we get an empty `MemoryRegion` in the end. And if this happens while `self.region` is already empty, then `self.region.start()` will happily leak an unallocated address. Let's just call `get_pages()` when we actually need it.

00xc: There is no "leak" per-se, because the address is not consumed by the allocator. On our side, getting back an empty region is a no-op:

```rust
if self.region.is_empty() {
    // self.region will still be empty, so it will be overwritten
    // next time we check
    self.region = region
    return Ok(());
}
```

or:

```rust
// region.len() is 0, so this is a no-op
self.region = self.region.expand(region.len())
```

And the `checked_sub()` check we do guarantees that we do not overflow the region, even if its length is 0. In other words, having an empty region is a completely valid state at all stages, even when passing it into the main kernel.

I will add the check for readability, but this `PageCursor` code will not live long anyway, since Jon will remove it with his upcoming changes.

peterfang: That's great, thanks!

PR #932, opened 2025-12-22
debug/stacktrace: do not allocate when printing a stacktrace

00xc:
Do not dynamically allocate a string when printing the stack trace information. When the SVSM is panicking the amount of available memory might be low, or the heap might not be set up, which will cause a panic during the initial panic, exhausting the stack.
    
Instead, use a statically allocated buffer to dynamically construct the required annotations.

### DIFF
--- a/kernel/src/debug/stacktrace.rs
+++ b/kernel/src/debug/stacktrace.rs
@@ -4,8 +4,6 @@
 //
 // Author: Nicolai Stange <nstange@suse.de>
 
-extern crate alloc;
-
 use crate::{
     address::{Address, VirtAddr},
     cpu::idt::common::{is_exception_handler_return_site, X86ExceptionContext},
@@ -13,9 +11,12 @@ use crate::{
     mm::{STACK_SIZE, STACK_TOTAL_SIZE, SVSM_CONTEXT_SWITCH_STACK, SVSM_STACK_IST_DF_BASE},
     utils::MemoryRegion,
 };
-use alloc::format;
 use bootlib::kernel_launch::{STAGE2_STACK, STAGE2_STACK_END};
-use core::{arch::asm, mem};
+use core::{
+    arch::asm,
+    fmt::{self, Write},
+    mem,
+};
 
 extern "C" {
     static bsp_stack: u64;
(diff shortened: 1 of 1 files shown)


kernel/src/debug/stacktrace.rs
```diff
@@ -244,26 +246,75 @@ impl Iterator for StackUnwinder {
...
+        self.pos = 0;
+    }
+}
+
+impl<const N: usize> fmt::Write for StacktraceBuf<N> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        // This should never overflow, but this is not a performance-critical
+        // path, and we really want to avoid panicking (if we got here we are
+        // already panicking in the first place), so check just in case.
+        let new_pos = self.pos.checked_add(s.len()).ok_or(fmt::Error)?;
```
osteffenrh, line 283: Just a thought: Why not copy as much into the buffer as we can and indicate truncation somehow (`...` at the end or similar)?
That might still be more useful than just printing nothing/the error message.

But I agree that this case is very unlikely to happen, since 512 bytes seems enough.

osteffenrh (approved):

kernel/src/debug/stacktrace.rs
```diff
@@ -244,26 +246,75 @@ impl Iterator for StackUnwinder {
...
+        buf.write_str(if annotated { "#" } else { " #" })?;
     }
-    log::info!("{}", msg);
+    log::info!("{}", buf.as_str());
+    Ok(())
 }
 
 pub fn print_stack(skip: usize) {
     let unwinder = StackUnwinder::unwind_this_cpu();
+    let mut buf = STACKTRACE_BUF.lock();
```
osteffenrh: Are we sure that everything called after taking the lock is panic-free? Otherwise we might end up in a deadlock

00xc: As far as I can tell, it is, but I don't think there's much we can do. The only thing I can think of to avoid taking a lock is to have a per-CPU buffer, but we are not guaranteed to have the `PerCpu` set up when we panic.

kernel/src/debug/stacktrace.rs
00xc: Okay, I modified the caller of `print_stack_frame()` to print the buffer even if writing fails, and add an indication if the message was truncated.

kernel/src/debug/stacktrace.rs
osteffenrh: Yeah, true. Ok.

kernel/src/debug/stacktrace.rs
osteffenrh: Thanks! :-D

kernel/src/debug/stacktrace.rs
msft-jlange: Can you use try-acquire semantics on this lock?  If the lock is already held, then you can fall back to a constant string.  This will ensure that there is no deadlock while not requiring any global resources to make forward progress.  Using a constant string will result in a loss of crash information but seeing a panic with a constant string is better than seeing nothing at all.

msft-jlange: You can also consider using an atomic static to record which CPU presently holds the global lock.  If it is not the current CPU, then you can just wait and let the panicking CPU proceed on its own.  If it is the current CPU (or if the current CPU is unknown) then you can proceed with a constant string.  This would ensure that the fallback to a constant string would only occur if there is reentrancy on the current CPU or if the panic is so early that the `PerCpu` has not yet been initialized and thus reentrancy cannot reliably be detected.

00xc: I finally opted to remove the spinlock altogether. The amount of dynamic elements in the formatted string is low. The main thing is adding a single space before the single-character annotations, which simply can be constructed on the stack for now.

kernel/src/debug/stacktrace.rs
```diff
@@ -244,18 +245,64 @@ impl Iterator for StackUnwinder {
...
+    let mut annotations = StacktraceBuf::<2>::new();
 
     if frame.is_exception_frame {
-        msg.push_str(" @");
-        annotated = true;
+        let _ = annotations.write_char('@');
     }
-    if !frame.is_aligned {
-        msg.push_str(if annotated { "#" } else { " #" });
+    if frame.is_aligned {
```
peterfang:
```suggestion
    if !frame.is_aligned {
```
The `#` marker is for unaligned stack frames

peterfang (approved):

PR #933, opened 2026-01-04
tools/igvmbuilder: fix highest VTL for native builds

msft-jlange:
Support for multiple VTLs requires VSM, so IGVM files that expect to launch in VTL 2 should always specify VSM as the platform type.  Native platform configuration should specify zero as the highest VTL.


peterfang (approved):
LGTM. `highest_vtl` isn't used in SVSM yet so there shouldn't be an impact.

PR #934, opened 2026-01-05
repo: update to Rust 2024 edition

00xc:
Rust 2024 edition was released with version 1.85 of the language, which means it is available to us. So as to not fall behind, update the edition of different components in the repo to the latest version.

Some of the changes that affect us are:
  * [Temporary](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-if-let-scope.html) and [tail](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-tail-expr-scope.html) scope duration changes
  * [`gen` is now a reserved keyword](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/gen-keyword.html)
  * [`extern` blocks must always be marked unsafe](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-attributes.html)
  * [`no_mangle` must always be marked unsafe](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-attributes.html)
  * [Formatting updates](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/rustfmt.html)

The full notes of the edition change can be consulted in the official Rust documentation site:

<https://doc.rust-lang.org/nightly/edition-guide/rust-2024/index.html>

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,1 @@
+style_edition = "2024"
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -30,7 +30,7 @@ RET=0
 for file in `git diff --diff-filter=d --name-only --staged`; do
     ext=${file##*.}
     if [ "$ext" == "rs" ]; then
-        rustfmt --check --edition 2021 $file > /dev/null 2>&1
+        rustfmt --check $file > /dev/null 2>&1
         if [ "$?" == "1" ]; then
             echo "$file needs rustfmt checking"
             RET=1


msft-jlange:
It looks like the new Rust edition changes the default sort order used by the formatter for `use` statements, and this causes a huge amount of churn in the code, which will prompt numerous conflicts with questionable benefit.  Is it possible to define some parameter in `Cargo.toml` to stick with the old sort order so that we can avoid requiring formatting changes to just about every single file?

00xc:
> It looks like the new Rust edition changes the default sort order used by the formatter for `use` statements, and this causes a huge amount of churn in the code, which will prompt numerous conflicts with questionable benefit. Is it possible to define some parameter in `Cargo.toml` to stick with the old sort order so that we can avoid requiring formatting changes to just about every single file?

The style edition can be specified in `rustfmt.toml`, but that is only used when invoking `rustfmt` directly, not `cargo fmt`, so there is no straightforward way to do this. Since January is typically low-traffic in terms of contributions, I think this might be a good time to do this.

msft-jlange:
> > It looks like the new Rust edition changes the default sort order used by the formatter for `use` statements, and this causes a huge amount of churn in the code, which will prompt numerous conflicts with questionable benefit. Is it possible to define some parameter in `Cargo.toml` to stick with the old sort order so that we can avoid requiring formatting changes to just about every single file?
> 
> The style edition can be specified in `rustfmt.toml`, but that is only used when invoking `rustfmt` directly, not `cargo fmt`, so there is no straightforward way to do this.

On top of that, the research that I did suggests that the sort order used by rustfmt is not configurable through `rustfmt.toml`.  The only thing we could influence is whether any sorting of import symbols is done at all - not how they are sorted.  So either we have to turn off sorting or we have to live with the changes.  Neither of those options is particularly great.

> Since January is typically low-traffic in terms of contributions, I think this might be a good time to do this.

I have a big stack of PRs coming so I'm not sure your assessment is entirely accurate - at least this year - but I don't think any other time would be any better.

The number one source of conflicts I have when rebasing is due to changes to imports, which is something that I wish I didn't have to manually take care of - this is something that smart tooling could do effortlessly.  It would be nice if there was some better solution here.

PR #935, opened 2026-01-05
tdp: start APs without going through stage2

msft-jlange:
AP startup can be simplified by using a protocol shared between stage1 and the SIPI stub that is used by native platforms.  This eliminates any dependency on stage2 for processor start, which will permit processor startup to occur without relying on stage2 page tables, and as well as the ultimate elimination of stage2.


kernel/src/platform/tdp.rs
```diff
@@ -153,6 +141,38 @@ impl SvsmPlatform for TdpPlatform {
...
+            } else {
+                self.validate_physical_page_range(
+                    MemoryRegion::new(PhysAddr::from(0u64), SIPI_STUB_GPA as usize),
+                    PageValidateOp::Validate,
+                )?;
+                if addr as usize > SIPI_STUB_GPA as usize + PAGE_SIZE {
+                    let paddr_start = PhysAddr::from(SIPI_STUB_GPA as usize + PAGE_SIZE);
+                    let paddr_end = PhysAddr::from(addr);
+                    self.validate_physical_page_range(
+                        MemoryRegion::new(paddr_start, paddr_end - paddr_start),
```
peterfang: Nit: `MemoryRegion::from_addresses()` accepts `paddr_end` directly

kernel/src/boot_stage2.rs
```diff
@@ -246,28 +236,6 @@ global_asm!(
...
-        /* vCPU index */
-        cmpl %esi, (%eax)
-        pause
-        jne .Lap_wait_for_signal
-
-        /* See smp.rs for prerequisites. */
-        movl ${AP_CTXT_ADDR}, %edi
-        movl ${AP_STARTUP_RIP_ADDR}, %eax
-        movq (%rax), %rax
-        jmp *%rax
```
peterfang, line 268: The entire `.Lskip_paging_setup` path can also be removed, as well as the use of `ap_flag` in `kernel/src/stage2.rs`

kernel/src/platform/mod.rs
```diff
@@ -166,6 +166,21 @@ pub trait SvsmPlatform: Sync {
     /// platform.
     fn get_io_port(&self) -> &'static dyn IOPort;
 
+    /// Validates low memory below the specified physical address.
+    /// # Safety
+    /// The caller is required to ensure that it is safe to validate low
+    /// memory.
+    unsafe fn validate_low_memory(&self, addr: u64) -> Result<(), SvsmError> {
```
peterfang: I'm curious about whether this could be implemented in `validate_virtual_page_range()`?

kernel/src/platform/tdp.rs
```diff
@@ -247,18 +267,31 @@ impl SvsmPlatform for TdpPlatform {
...
-        // transition_cr3 is not needed since all TD APs are using the stage2
-        // page table set up by the BSP.
-        context_mapping.write(create_ap_start_context(&context, 0));
+        context_mapping.write(create_ap_start_context(&context, self.transition_cr3));
+
+        // Map the reset page to write the VP index of the CPU being started.
+        // This will release the target CPU from its initial spin loop and
+        // permit it to jump into the SIPI stub.  The VP index gate is at an
+        // offset of 0x800 bytes into the reset page located at 0xFFFFF000.
+        let vp_index_pa = PhysAddr::new(0xFFFFF800);
```
peterfang, line 285: How was this address selected? From what I can tell `reset.S` starts at 0xfffffe00 and there could be data/text before that. If the rest of the stuff in stage1 will be removed later, can we choose a safer address in the interim?

kernel/src/platform/mod.rs
msft-jlange: There's an important difference here.  The function `validate_virtual_page_range()` is expected to validate the entirety of its argument range, and this function is expected to execute on all of memory throughout the execution of the kernel since memory state may change many times after boot.  On the other hand, `validate_low_memory()` is expected to run only once on boot to validate memory below the argument address with the exception of any memory that has been reserved by the platform (in the case of TDX, that's the SIPI stub).  I can add comments to this effect to make this more clear, but these differences in behavior require the functions to be separate.

kernel/src/platform/tdp.rs
msft-jlange: The entirety of the page at FFFFFxxx is reserved for use by stage1.  The code in `reset.S` extends from E00-FFF and the code in `stage1.rs` is less than 512 bytes, so it does not extend beyond 1FF.  That leaves a huge hole in the middle.

I'm not too worried about the precise extents of this code because stage1 is going to be replaced in the near future, and the VP index gate will be moved at that time based on the structure of the new boot loader.

peterfang: OK, just wanted to point that out so that we have a point of reference for the future. Since stage1 will be replaced soon this should be ok.

peterfang (approved):
LGTM now, thanks!

PR #936, opened 2026-01-08
verification: add missing SPDX header

luigix25:
`verus_stub` and `verus_macro_stub` are missing the SPDX header. Add it.

I'm the original author of those files.

### DIFF
--- a/verification/verus_macro_stub/src/lib.rs
+++ b/verification/verus_macro_stub/src/lib.rs
@@ -1,3 +1,9 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2025 Red Hat, Inc.
+//
+// Author: Luigi Leonardi <leonardi@redhat.com>
+
 extern crate proc_macro;
 use proc_macro::TokenStream;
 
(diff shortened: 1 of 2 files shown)


luigix25:
I think it could be a good idea to add a CI check for SPDX for all newly added files.

verification/verus_macro_stub/src/lib.rs
```diff
@@ -1,3 +1,9 @@
+// SPDX-License-Identifier: MIT
```
stefano-garzarella: in `verification/` all of them are `MIT OR Apache-2.0`, should we use the same?

stefano-garzarella:
> I think it could be a good idea to add a CI check for SPDX for all newly added files.

yeah, maybe we should add some script ore re-use some gh action in the marketplace.

verification/verus_macro_stub/src/lib.rs
luigix25: yeah, makes sense.

luigix25:
> > I think it could be a good idea to add a CI check for SPDX for all newly added files.
> 
> yeah, maybe we should add some script ore re-use some gh action in the marketplace.

I'll do it in a different PR

luigix25:
v2:
- Changed license from `MIT` to `MIT OR Apache-2.0` (@stefano-garzarella)

PR #937, opened 2026-01-08
repo: update minimum Rust version to 1.87.0

n-ramacciotti:
The [coconut-alloc](https://github.com/coconut-svsm/coconut-alloc) crate requires Rust [1.87.0](https://github.com/rust-lang/rust/releases/tag/1.87.0) for a feature stabilized in this release.

Update toolchain configuration as suggested in  [#924](https://github.com/coconut-svsm/svsm/pull/924)

Fix clippy warnings

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -2,7 +2,6 @@
 name = "svsm"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.86.0"
 
 [[bin]]
 name = "stage2"
--- a/kernel/src/protocols/vtpm.rs
+++ b/kernel/src/protocols/vtpm.rs
@@ -60,7 +60,7 @@ fn vtpm_platform_commands_supported_bitmap() -> u64 {
 
 fn is_vtpm_platform_command_supported(cmd: TpmPlatformCommand) -> bool {
     let vtpm = vtpm_get_locked();
-    vtpm.get_supported_commands().iter().any(|x| *x == cmd)
+    vtpm.get_supported_commands().contains(&cmd)
 }
 
 const SEND_COMMAND_REQ_INBUF_SIZE: usize = PAGE_SIZE - 9;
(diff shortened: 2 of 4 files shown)


kernel/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "svsm"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.86.0"
+rust-version = "1.87.0"
```
msft-jlange: Is there a reason this has an explicit version number instead of just relying on the root `rust-toolchain.toml` in the workspace?  I worry about the potential for different sub-crates to diverge in toolchain version and this could cause huge headaches.

kernel/src/protocols/vtpm.rs
stefano-garzarella, line 1: > kernel: Apply clippy suggestion
> 
> Refactor code as recommended by clippy

nit: I'd specify that it's clippy from 1.87

> 
> Signed-off-by: Nicola Ramacciotti <niko.ramak@gmail.com>

kernel/Cargo.toml
stefano-garzarella: IIUC this is the minimal version a crate claims to be compiled, while `rust-toolchain.toml` is to specify the toolchain to be used to build the whole workspace (and I guess it should satisfy all the crates min versions).

IIRC we added `rust-toolchain.toml` later,  and I think this field is optional, so yeah I think we could remove them till we need a more fine-grained rust versions (e.g. if we publish something on crates.io).

@msft-jlange should we do that in a separate PR or here since we are touching it? (not a strong opinion on my side)

msft-jlange: Changing the sub-crate files now and then removing the same lines in a different PR just seems like extra work.  If we're going to remove them, why not just remove them now?  I suppose we could remove those in a separate PR that merges before this one, but I don't think it's worth the effort.

luigix25 (approved):
LGTM! Thanks

stefano-garzarella (approved):

joergroedel (approved):

PR #943, opened 2026-01-19
platform: initialize CPUID & secrets pages via platform code

00xc:
The CPUID page and secrets page are specific to SEV-SNP, so do not open-code their initialization in main SVSM code, and move it into platform code. In order to do this, add a new hook for early SVSM-specific environment setup, mimicking the `env_setup()` and `env_setup_late()` pair of hooks (since the initialization is only done for the main SVSM and not stage 2)

In order to have cleaner error handling, make both `init_cpuid_table()` and `initialize_secrets_page()` return errors instead of implicitly panicking, so that the platform hook can bubble the errors up to the main SVSM function, having a single point of panic.

Note that, for SEV-SNP, the CPUID and secrets pages must always be present, so the NULL checks are no longer necessary, as the two downstream functions will perform both a NULL and alignment check, returning an error appropriately.

### DIFF
--- a/kernel/src/platform/mod.rs
+++ b/kernel/src/platform/mod.rs
@@ -36,7 +36,7 @@ use crate::types::PageSize;
 use crate::utils::MemoryRegion;
 use crate::utils::immut_after_init::ImmutAfterInitCell;
 
-use bootlib::kernel_launch::Stage2LaunchInfo;
+use bootlib::kernel_launch::{KernelLaunchInfo, Stage2LaunchInfo};
 use bootlib::platform::SvsmPlatformType;
 
 static SVSM_PLATFORM_TYPE: ImmutAfterInitCell<SvsmPlatformType> = ImmutAfterInitCell::uninit();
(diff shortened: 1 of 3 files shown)


kernel/src/cpu/cpuid.rs
```diff
@@ -15,10 +16,28 @@ static CPUID_PAGE: ImmutAfterInitCell<&SnpCpuidTable> = ImmutAfterInitCell::unin
...
-        .init(table)
-        .expect("Could not initialize CPUID page");
+/// Initialize the CPUID table page for the main SVSM kernel.
+pub fn init_svsm_cpuid_table(addr: VirtAddr) -> Result<(), SvsmError> {
+    // SAFETY: this is called from the main function for the SVSM and no other
+    // CPUs have been brought up, so the pointer cannot be aliased.
+    // `aligned_mut()` will check alignment for us.
+    let table = unsafe {
+        addr.aligned_mut::<SnpCpuidTable>()
+            .ok_or(SvsmError::InvalidCpuidPage)
```
msft-jlange, line 26: The kernel copy of the CPUID page is allocated from the kernel page heap so it's guaranteed to page-aligned.  Stage2 use of the CPUID page will be replaced in the near future with completely different code.  For these reasons, I think it's better just to `unwrap()` instead of defining a new error value since the error will never happen in practice.

kernel/src/error.rs
```diff
@@ -94,14 +94,18 @@ pub enum SvsmError {
...
     Insn(InsnError),
     /// Invalid address, usually provided by the guest
     InvalidAddress,
     /// Error reported when convert a usize to Bytes
     InvalidBytes,
+    /// Error reported when the CPUID page is not present or improperly
+    /// aligned.
+    InvalidCpuidPage,
+    /// Error reported when there is no secrets page set up, or its
+    /// address is not properly aligned
```
msft-jlange, line 107: The secrets page is either allocated in the IGVM file or in the kernel heap, and both of those guarantee page alignment.  This means there is no possibility that the secrets page is unaligned, except for a fundamental bug.  The only error worth detecting is a missing secrets page, and if an unaligned secrets page is detected, we can simply panic as an indication of an error in the loader.

PR #944, opened 2026-01-21
Fix nightly warnings

joergroedel:
The nightly compiler used for building the tests started to emit a number of warnings. Even worse, the stable and nightly toolchains disagree on the safety of `__cpuid_count()`. Fix or silence all the warnings for now.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -362,6 +362,9 @@ mod tests {
 
         const CPUID_EXTENDED_FUNCTION_INFO: u32 = 0x8000_0000;
         const CPUID_MEMORY_ENCRYPTION_INFO: u32 = 0x8000_001F;
+        // Stable clippy and the nightly rust compiler disagree on the safety
+        // of __cpuid_count(). Silence both until this is resolved.
+        #[allow(unused_unsafe)]
         // SAFETY: CPUID does never affect safety.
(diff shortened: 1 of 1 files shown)


kernel/src/cpu/vc.rs
```diff
@@ -363,6 +363,7 @@ mod tests {
         const CPUID_EXTENDED_FUNCTION_INFO: u32 = 0x8000_0000;
         const CPUID_MEMORY_ENCRYPTION_INFO: u32 = 0x8000_001F;
         // SAFETY: CPUID does never affect safety.
+        #[allow(unused_unsafe)]
```
stefano-garzarella, line 367: nit: I'd extend the safety comment adding the rustc version where it's not `unsafe` anymore, so in the future we can revalue it (ditto for the next hunk).

joergroedel: It is actually clippy in stable rust which still complains about missing `unsafe`, the rust compiler is fine. I can add a comment for that. The nightly rust compiler just starts to complain about the "unecessary" unsafe.

stefano-garzarella: ok, I see now. Up to you for the comment.

joergroedel: Added a comment and pushed an updated branch.

PR #945, opened 2026-01-21
Support MMAP and MUNMAP system calls

joergroedel:
This PR implements the MMAP and MUNMAP system calls and foundations for MRESIZE. The system calls do not have Posix semantics yet. In fact, partial unmaps are not supported. A `VMM` can only be unmapped as a whole for now.

The foundations for MRESIZE are also implemented, but that support is not finished yet. Therefore this PR is marked as Draft.

### DIFF
--- a/syscall/src/class0.rs
+++ b/syscall/src/class0.rs
@@ -1,11 +1,13 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 //
 // Copyright (c) 2024 Intel Corporation.
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
 //
 // Author: Chuanxiao Dong <chuanxiao.dong@intel.com>
+//         Joerg Roedel <joerg.roedel@amd.com>
 
-use super::call::{SysCallError, syscall1, syscall3};
-use super::{SYS_EXEC, SYS_EXIT};
+use super::call::{SysCallError, syscall1, syscall2, syscall3, syscall5};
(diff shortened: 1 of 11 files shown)


syscall/src/class0.rs
```diff
@@ -20,6 +20,92 @@ pub fn exit(code: u32) -> ! {
...
+    handle: Option<&ObjHandle>,
+    addr: usize,
+    offset: u64,
+    length: u64,
+    flags: MMFlags,
+) -> Result<usize, SysCallError> {
+    let obj: u64 = if let Some(fd) = handle {
+        fd.id().into()
+    } else {
+        0u64
```
n-ramacciotti, line 54: With None (Anonymous mapping) here 0 is assigned and it is later used in `kernel/src/syscall/class0.rs`:
```
pub fn sys_mmap(
..
let opt = obj_get(obj_id.into()).ok();
let mmap_addr = if let Some(obj) = opt {
 //map with obj
}else{
 //anonymous mapping
}
..
```

But `obj_id` 0 is already used in `kernel/src/task/tasks.rs`:
```
fn task_attach_console() {
    let file_handle = stdout_open();
    let obj_handle = ObjHandle::new(0);
    current_task()
        .add_obj_at(file_handle, obj_handle)
        .expect("Failed to attach console");
}
```

So, `sys_mmap` finds something when looking for object 0 and it enters the wrong branch.

### Workaround for testing with coconut-alloc
Changing obj_id to another value (like 1) makes the mmap work correctly when testing in the userspace allocator PR.

This in `syscall/src/console.rs` should be changed as well:
```
static CONSOLE_HANDLE: FsObjHandle = FsObjHandle::new(ObjHandle::new(0));
```

PR #946, opened 2026-01-22
SEV TLB flush fixes

joergroedel:
Make sure the SVSM flushes nested translations on RMP changes and remove the unnecessary use of per-ASID flushing.


kernel/src/sev/tlb.rs
```diff
@@ -47,7 +48,7 @@ pub fn flush_tlb_sync() {
 }
 
 pub fn flush_tlb_global() {
-    let rax: u64 = INVLPGB_VALID_ASID | INVLPGB_VALID_GLOBAL;
+    let rax: u64 = INVLPGB_VALID_GLOBAL | INVLPGB_VALID_NESTED;
```
msft-jlange, line 51: `INVLPGB_VALID_NESTED` only applies when INVLPGB is executed by the hypervisor.  Guest paging management cannot affect any nested translations because the guest cannot modify the state of the nested page tables.

PR #948, opened 2026-01-26
virtio/mmio: move fw_cfg check into probe_mmio_slots()

stefano-garzarella:
Move the fw_cfg availability check from svsm_init() into probe_mmio_slots(), which is the only place where fw_cfg is actually needed. Pass `SvsmConfig` as a parameter to probe_mmio_slots() to perform this check.

This prepares for future work where virtio MMIO addresses may be discovered via device tree instead of fw_cfg. By passing `SvsmConfig`, the probe function can internally decide how to discover devices without callers needing to know the details.

Also derive Default for MmioSlots to simplify returning empty slot collections.

### DIFF
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -162,13 +162,12 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
 /// Returns Ok if initialization is successful or no virtio devices are found
 /// Returns an error when a virtio device is found but its driver initialization fails.
 #[cfg(feature = "virtio-drivers")]
-fn initialize_virtio_mmio() -> Result<(), SvsmError> {
-
+fn initialize_virtio_mmio(_config: &SvsmConfig<'_>) -> Result<(), SvsmError> {
     #[cfg(feature = "block")]
     {
         use svsm::block::virtio_blk::initialize_block;
 
-        let mut slots = probe_mmio_slots();
+        let mut slots = probe_mmio_slots(_config);
         initialize_block(&mut slots)?;
     }
 
@@ -436,11 +435,7 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
     virt_log_usage();
 
     #[cfg(feature = "virtio-drivers")]
-    if config.has_fw_cfg_port() {
-        // Virtio cannot exist if there is no fw_cfg, so do not bother to
-        // attempt initialization if it is not present.
-        initialize_virtio_mmio().expect("Failed to initialize virtio-mmio drivers");
-    }
+    initialize_virtio_mmio(&config).expect("Failed to initialize virtio-mmio drivers");
 
     if let Err(e) = SVSM_PLATFORM.launch_fw(&config) {
         panic!("Failed to launch FW: {e:?}");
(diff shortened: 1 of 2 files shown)


kernel/src/svsm.rs
```diff
@@ -162,8 +162,8 @@ fn mapping_info_init(launch_info: &KernelLaunchInfo) {
 /// Returns Ok if initialization is successful or no virtio devices are found
 /// Returns an error when a virtio device is found but its driver initialization fails.
 #[cfg(feature = "virtio-drivers")]
-fn initialize_virtio_mmio() -> Result<(), SvsmError> {
-    let mut slots = probe_mmio_slots();
+fn initialize_virtio_mmio(config: &SvsmConfig<'_>) -> Result<(), SvsmError> {
+    let mut slots = probe_mmio_slots(config);
```
msft-jlange: Shouldn't this probe depend on feature `block`?  It appears that when compiling with `virtio-drivers` but not `block`, then the `slots` variable is unreferenced, which would cause a compiler warning.

luigix25: FWIW clippy is not complaining, but you have a good point. With any new virtio driver (like#766), this probe would then depend on `block`, `vsock` or `somethingelse`. 

A quick workaround could be to rename `slots` to `_slots`. WDYT?

msft-jlange: > A quick workaround could be to rename `slots` to `_slots`. WDYT?

That would result in unnecessary work probing the slots and allocating memory, which would then simply be freed without ever using it.  It would be better to move the probe under feature `block` to avoid the unnecessary work.  This would require `config` to be renamed `_config`.

stefano-garzarella: > FWIW clippy is not complaining, but you have a good point. With any new virtio driver (like#766), this probe would then depend on `block`, `vsock` or `somethingelse`.

@luigix25 `make clippy` uses `--all-features` but if you enable only `virtio-drivers` you will have:

```
$ cargo clippy --features virtio-drivers --package svsm --target x86_64-unknown-none 
   Compiling svsm v0.1.0 (/home/stefano/repos/coconut/svsm/kernel)
warning: unused variable: `slots`
   --> kernel/src/svsm.rs:166:13
    |
166 |     let mut slots = probe_mmio_slots(config);
    |             ^^^^^ help: if this is intentional, prefix it with an underscore: `_slots`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: variable does not need to be mutable
   --> kernel/src/svsm.rs:166:9
    |
166 |     let mut slots = probe_mmio_slots(config);
    |         ----^^^^^
    |         |
    |         help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default
```

I think we should start to use `cargo-all-features` like we did in rust-vmm to prevent these issues: https://github.com/rust-vmm/rust-vmm-ci/pull/192

> > A quick workaround could be to rename `slots` to `_slots`. WDYT?
> 
> That would result in unnecessary work probing the slots and allocating memory, which would then simply be freed without ever using it. It would be better to move the probe under feature `block` to avoid the unnecessary work. This would require `config` to be renamed `_config`.

@msft-jlange okay, let's to that for now, and with #766 we can play a bit more with features to run the probe on more cases (e.g. for vsock feature if any or in any case)

PR #951, opened 2026-01-29
Verify ACPI table checksum

osteffenrh:
Calculate the checksum when parsing ACPI tables and print a warning if
it does not match (is == 0).

Note that our QEMU fork still does not provide the correct checksum, while the upstream
master branch does. This will change when we rebase to the latest release soon.


msft-jlange:
What is the benefit of this change?  Are we seeing boot failures as a result of corrupt ACPI tables?  Do we think that checksum verification makes a meaningful difference in the stability of the boot process?

From the standpoint of confidential computing, the checksum is no more trusted than any of the rest of the contents, so it doesn't provide any particular security benefit.

PR #953, opened 2026-02-03
stage2: move kernel image layout code into a separate library

msft-jlange:
Move the logic that constructs the kernel memory map out of stage2 and into a separate library.  This enables future scenarios in which the kernel memory area can be populated outside of stage2, such as in image generation tools.

### DIFF
--- a/boot/bootimg/src/lib.rs
+++ b/boot/bootimg/src/lib.rs
@@ -22,6 +22,8 @@ use crate::symbols::load_kernel_symbols;
 pub use crate::defs::*;
 
 use ::elf::Elf64File;
+use bootdefs::kernel_launch::INITIAL_KERNEL_STACK_WORDS;
+use bootdefs::kernel_launch::InitialKernelStack;
 use bootdefs::kernel_launch::KernelLaunchInfo;
 use igvm_defs::PAGE_SIZE_4K;
 use zerocopy::IntoBytes;
@@ -60,13 +62,11 @@ where
         vmsa_in_kernel_heap,
     )?;
 
-    // Allocate 32 KB for the initial stack.  This must come at the base of
-    // the kernel heap so it is preceded by a guard page.  These pages must
-    // be added as empty pages so they are validated for immediate use.
+    // Alloacte 32 KB for the intiial stack.  This must come at the base of
+    // the kernel heap so it is preceded by a guard page.
     let stack_page_count = 8;
     let stack_size = stack_page_count * PAGE_SIZE_4K;
     let (initial_stack_paddr, initial_stack_base) = kernel_heap.allocate_pages(stack_page_count)?;
-    add_page_data(initial_stack_paddr, None, stack_size)?;
 
     // Initialize the page tables that will be used for mapping kernel data.
     let mut kernel_page_tables =
@@ -142,6 +142,23 @@ where
     };
     add_page_contents(add_page_data, launch_info_paddr, launch_info.as_bytes())?;
 
+    // Now add the contents of the initial stack.  All of the stack pages will
+    // be added as zero-filled pages except the final page, which holds the
+    // launch state.
+    let stack_zeroed_size = stack_size - PAGE_SIZE_4K;
+    add_page_data(initial_stack_paddr, None, stack_zeroed_size)?;
+    let initial_stack_data = InitialKernelStack {
+        _reserved: [0; 512 - INITIAL_KERNEL_STACK_WORDS],
+        paging_root,
+        launch_info_vaddr,
+        stack_limit: initial_stack_base,
+    };
+    add_page_data(
+        initial_stack_paddr + stack_zeroed_size,
+        Some(initial_stack_data.as_bytes()),
+        PAGE_SIZE_4K,
+    )?;
+
     let info = BootImageInfo {
         boot_params_paddr,
         cpuid_paddr,
@@ -149,12 +166,12 @@ where
         kernel_pdpt_paddr: kernel_page_tables.kernel_pdpt_paddr(),
         kernel_pml4e_index: kernel_page_tables.kernel_pml4e_index(),
         total_pt_pages,
+        kernel_page_tables_base: paging_root,
         context: BootImageContext {
             entry_point: kernel_entry,
-            initial_stack: initial_stack_base + (stack_page_count * PAGE_SIZE_4K),
-            initial_stack_base,
             paging_root,
-            launch_info_vaddr,
+            initial_stack: initial_stack_base + (stack_page_count * PAGE_SIZE_4K)
+                - (INITIAL_KERNEL_STACK_WORDS * 8) as u64,
         },
     };
 
--- a/kernel/src/stage2.rs
+++ b/kernel/src/stage2.rs
@@ -22,7 +22,7 @@ use bootimg::BootImageError;
 use bootimg::BootImageInfo;
 use bootimg::BootImageParams;
 use bootimg::prepare_boot_image;
-use core::arch::asm;
+use core::arch::global_asm;
 use core::mem;
 use core::panic::PanicInfo;
 use core::slice;
@@ -67,6 +67,7 @@ use release::COCONUT_VERSION;
 
 unsafe extern "C" {
     static mut pgtable: PageTable;
+    fn switch_to_kernel(entry: u64, initial_stack: u64, platform_type: u64) -> !;
 }
 
 #[derive(Debug)]
@@ -359,7 +360,7 @@ fn prepare_kernel_image(
     // is only necessary if there is a non-zero confidentiailty mask.
     if private_pte_mask() != 0 {
         for pt_index in 0..boot_image_info.total_pt_pages {
-            let paddr = boot_image_info.context.paging_root + (pt_index * PAGE_SIZE as u64);
+            let paddr = boot_image_info.kernel_page_tables_base + (pt_index * PAGE_SIZE as u64);
             // SAFETY: the boot image has been fully mapped and the address
             // translation from physical to virtual is known to be correct by
             // construction.
@@ -517,18 +518,31 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
         // Shut down the PerCpu instance
         shutdown_percpu();
 
-        asm!("jmp *%rax",
-             in("rax") boot_image_info.context.entry_point,
-             in("rdi") boot_image_info.context.launch_info_vaddr,
-             in("rsi") platform_type as u64,
-             in("rdx") boot_image_info.context.initial_stack,
-             in("rcx") boot_image_info.context.initial_stack_base,
-             in("r8") boot_image_info.context.paging_root,
-             options(att_syntax))
+        switch_to_kernel(
+            boot_image_info.context.entry_point,
+            boot_image_info.context.initial_stack,
+            platform_type as u64,
+        );
     };
-
-    unreachable!("Road ends here!");
 }
+
+global_asm!(
+    r#"
+        .globl switch_to_kernel
+        switch_to_kernel:
+
+        /* Switch to the kernel stack. */
+        movq %rsi, %rsp
+
+        /* Load the platform type into rax as expected by the kernel */
+        movq %rdx, %rax
+
+        /* Enter the kernel. */
+        push %rdi
+        ret
+        "#,
+    options(att_syntax)
+);
 
 #[panic_handler]
 fn panic(info: &PanicInfo<'_>) -> ! {
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -102,33 +102,29 @@ global_asm!(
 
         .globl startup_64
     startup_64:
-        /*
-         * Register conventions upon entry:
-         * RDI = launch info block
-         * RSI = platform type
-         * RDX = initial stack
-         * RCX = limit of initial stack
-         * R8 = kernel page table root
-         *
-         * Setup stack.
-         *
-         * The initial stack is always mapped across all page tables because it
-         * uses the shared PML4E, making it accessible after switching to the
-         * first task's page table.
-         *
-         * See switch_to() for details.
-         */
-        movq %rdx, %rsp
-        leaq bsp_stack_end(%rip), %rdx
-        movq %rsp, (%rdx)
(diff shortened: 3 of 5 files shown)


kernel/src/svsm.rs
```diff
@@ -138,10 +141,32 @@ global_asm!(
...
+/// The launch info block must correctly specify the initial state of the
+/// heap.
+unsafe fn memory_init(launch_info: &KernelLaunchInfo, platform: &dyn SvsmPlatform) {
+    // Unallocated hep memory has not already been accepted so it must be
+    // accepted here.
+    let heap_vaddr = VirtAddr::from(launch_info.heap_area_virt_start);
+    let heap_allocated = launch_info.heap_area_allocated as usize;
+    let heap_length = launch_info.heap_area_page_count as usize;
+    if launch_info.heap_area_allocated < launch_info.heap_area_page_count {
+        // SAFETY: the launch info is assumed to correctly
```
stefano-garzarella: I guess this comment was truncated by mistake.

kernel/src/mm/validate.rs
```diff
@@ -206,3 +213,61 @@ impl ValidBitmap {
...
+            let (paddr, page_size): (u64, usize) = match frame {
+                PageFrame::Size2M(pa) => (pa.into(), PAGE_SIZE_2M),
+                PageFrame::Size4K(pa) => (pa.into(), PAGE_SIZE),
+                // Other page sizes are not supported here.
+                _ => return Err(SvsmError::Mem),
+            };
+            // When encountering a discontinuity in physical addresses,
+            // process the old range before proceeding.
+            if (paddr != paddr_next) && (paddr_len != 0) {
+                let paddr_start = PhysAddr::new((paddr - paddr_len as u64) as usize);
```
stefano-garzarella: IIUC the "old range" is tracked by `paddr_next`, so should we use it here?

I mean `let paddr_start = PhysAddr::new((paddr_next - paddr_len as u64) as usize);`

boot/bootimg/src/symbols.rs
```diff
@@ -0,0 +1,137 @@
...
+        // Allocate a page from the kernel heap to hold this data and write it
+        // to the image.
+        let (paddr, vaddr) = self.heap.allocate_pages(1)?;
+
+        // Write the data to the allocated physical address.
+        (self.add_page_data)(paddr, Some(&self.data), PAGE_SIZE_4K)?;
+
+        // Expand the span of the allocation based on the returned virtual
+        // address.
+        self.span.length = vaddr + PAGE_SIZE_4K - self.span.start;
```
stefano-garzarella: I'm a bit confused here, IIUC `self.span.start` is always 0, so I don't get this.

Also should return early if `self.data` is empty?

Cargo.toml
```diff
@@ -44,7 +44,8 @@ edition = "2024"
 
 [workspace.dependencies]
 # internal library crates
-bootlib = { path = "bootlib" }
+bootdefs = { path = "boot/defs" }
+bootimg = { path = "boot/bootimg" }
```
stefano-garzarella, line 51: Should we add them in `members` too?

kernel/src/mm/validate.rs
msft-jlange: Good catch!

boot/bootimg/src/symbols.rs
msft-jlange: The `self.span` field was an adaptation of the old version of this code when it was in `stage2_symbols.rs` but you're right; the code has changed enough that the span tracking is no longer useful.  I'll modify this to track only the length instead of a span.

Cargo.toml
msft-jlange: Yes, we should.

boot/bootimg/src/symbols.rs
msft-jlange: Actually, the span is still needed, but what's missing is the initialization of `span.start`.  That's what needs to be fixed.

kernel/src/svsm.rs
```diff
@@ -138,10 +141,33 @@ global_asm!(
     options(att_syntax)
 );
 
-pub fn memory_init(launch_info: &KernelLaunchInfo) {
+/// # Safety
+/// The launch info block must correctly specify the initial state of the
+/// heap.
+unsafe fn memory_init(launch_info: &KernelLaunchInfo, platform: &dyn SvsmPlatform) {
+    // Unallocated hep memory has not already been accepted so it must be
```
stefano-garzarella: nit:
```suggestion
    // Unallocated heap memory has not already been accepted so it must be
```

kernel/src/mm/validate.rs
```diff
@@ -206,3 +213,61 @@ impl ValidBitmap {
...
+                // Other page sizes are not supported here.
+                _ => return Err(SvsmError::Mem),
+            };
+            // When encountering a discontinuity in physical addresses,
+            // process the old range before proceeding.
+            if (paddr != paddr_next) && (paddr_len != 0) {
+                let paddr_start = PhysAddr::new((paddr_next - paddr_len as u64) as usize);
+                platform.page_state_change(
+                    MemoryRegion::new(paddr_start, paddr_len),
+                    PageSize::Huge,
```
stefano-garzarella, line 247: Why using always `PageSize::Huge` ?

IIUC this is pre-existing, but it's more a comment for me to learn a bit more about this code.

bootlib/src/igvm_params.rs
```diff
@@ -151,10 +152,6 @@ pub struct IgvmParamBlock {
     /// The guest physical address of the base of the stage1 bootloader
     pub stage1_base: u64,
 
-    /// The amount of space that must be reserved at the base of the kernel
-    /// memory region (e.g. for VMSA contents).
-    pub kernel_reserved_size: u32,
-
```
stefano-garzarella, line 157: IIUC we are breaking compatibility with previous IGVM binary files, so doesn't matter to keep here 4 reserved bytes, right?

If my reasoning is true, what about mentioning that in the commit description?

boot/bootimg/src/symbols.rs
```diff
@@ -0,0 +1,159 @@
...
+                .extend_from_slice(&val[offset..offset + bytes_available]);
+
+            // Flush the full buffer and prepare to consume more data.
+            self.flush()?;
+            offset += bytes_available;
+        }
+    }
+
+    /// Returns the contiguous memory span that was allocated by the
+    /// PageCursor`.
```
stefano-garzarella: nit:
```suggestion
    /// `PageCursor`.
```

boot/bootimg/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    // is chosen to serve as a guard page for the initial stack.
+    let mut kernel_heap = create_kernel_heap(
+        boot_image_params.kernel_region_start,
+        boot_image_params.kernel_region_page_count,
+        &kernel_elf_sizes,
+        PAGE_SIZE_4K,
+        vmsa_in_kernel_heap,
+    )?;
+
+    // Allocate 32 KB for the intiial stack.  This must come at the base of
```
stefano-garzarella: nit:
```suggestion
    // Allocate 32 KB for the initial stack.  This must come at the base of
```

boot/bootimg/src/page_tables.rs
```diff
@@ -0,0 +1,308 @@
...
+
+#[derive(Debug)]
+struct PageTableContents {
+    allocation: Box<[MaybeUninit<X64Pte>]>,
+}
+
+impl PageTableContents {
+    fn new(count: usize) -> Self {
+        // Allcoation of a zeroed slice is unstable in the current version of
+        // the toolchain, so instead, an unintialized alloation is made and
```
stefano-garzarella: nit:
```suggestion
        // Allocation of a zeroed slice is unstable in the current version of
        // the toolchain, so instead, an uninitialized allocation is made and
```

kernel/src/mm/validate.rs
msft-jlange: This is based on the GHCB protocol (it's ignored for GHCI).  This is an instruction to the GHCB handling logic that the request should be made to the host using large pages where possible.  Without passing `PageSize::Huge`, the GHCB request will be made exclusively with small pages.

I doubt the GHCB code really should behave this way; it should always form the request using the 2 MB pages wherever possible.  There is nothing in the GHCB specification that says that the behavior of the page state change operation should be different between 2 MB and 4 KB pages; the distinction is only used for encoding.  This change to the GHCB protocol is outside the scope of this PR.

bootlib/src/igvm_params.rs
msft-jlange: There is no cross-version compatibility concern between the kernel and the IGVM format.  The kernel doesn't load IGVM files; the kernel is embedded in an IGVM file.  The only requirement is that the kernel image and the IGVM file builder used to build that kernel image must come from the same build, but that's how our build system works.  There is no interaction between IGVM files of one version and kernel images of another.

boot/bootimg/src/elf.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    ElfSizes {
+        virt_base: vaddr_alloc_base,
+        virt_len: vaddr_alloc_info.range.vaddr_end - vaddr_alloc_base,
+        phys_size: page_count,
+    }
+}
+
+/// Loads a single ELF segment and returns its virtual memory region.
```
peterfang: This function now returns the segment length

boot/bootimg/src/elf.rs
```diff
@@ -0,0 +1,130 @@
...
+use elf::Elf64File;
+use igvm_defs::PAGE_SIZE_4K;
+
+pub struct ElfSizes {
+    pub virt_base: u64,
+    pub virt_len: u64,
+    pub phys_size: u64,
+}
+
+pub fn get_elf_sizes(elf: &Elf64File<'_>) -> ElfSizes {
```
peterfang, line 26: This function needs a comment

boot/bootimg/src/elf.rs
```diff
@@ -0,0 +1,130 @@
...
+    add_page_data: &mut F,
+) -> Result<u64, BootImageError>
+where
+    F: FnMut(u64, Option<&[u8]>, u64) -> Result<(), BootImageError>,
+{
+    let vaddr_alloc_info = elf.image_load_vaddr_alloc_info();
+    let vaddr_alloc_base = vaddr_alloc_info.range.vaddr_begin;
+
+    // Map and populate the SVSM kernel ELF's PT_LOAD segments.
+    let mut phys_addr = paddr_base;
```
peterfang, line 103: We can make sure `paddr_base` is page-aligned first, since all segments should have a page-aligned starting address

boot/bootimg/src/elf.rs
```diff
@@ -0,0 +1,130 @@
...
+use crate::round_to_pages;
+
+use core::slice;
+use elf::Elf64File;
+use igvm_defs::PAGE_SIZE_4K;
+
+pub struct ElfSizes {
+    pub virt_base: u64,
+    pub virt_len: u64,
+    pub phys_size: u64,
```
peterfang: Can we call this something like `phys_page_count`? I almost got confused when I was looking at how kernel size was determined

kernel/src/stage2.rs
```diff
@@ -197,185 +93,96 @@ impl KernelPageTablePage<'_> {
...
+    ) -> Result<(), BootImageError> {
+        let total_size = total_size as usize;
+        assert_eq!((total_size & (PAGE_SIZE - 1)), 0);
+        // Create a page mapping at the next available virtual address.
+        // Since the stage2 page tables will be discarded upon entry to the
+        // kernel, the specific mapping address doesn't matter.  The addresses
+        // are used in sequence to avoid the need for any TLB flush after
+        // the address has been used.
+        let map_region = MemoryRegion::new(self.phys_to_virt(paddr), total_size);
+        map_page_range(map_region, PhysAddr::new(paddr as usize)).map_err(|e| {
```
peterfang:
```suggestion
        map_page_range(map_region, PhysAddr::from(paddr)).map_err(|e| {
```

kernel/src/stage2.rs
```diff
@@ -197,185 +93,96 @@ impl KernelPageTablePage<'_> {
...
-        // each address is in the kernel address space and has been mapped for
-        // the first time.  This guarantees that the virtual address region is
-        // not already in use and thus can safely be validated here
+        paddr: u64,
+        data: Option<&[u8]>,
+        total_size: u64,
+    ) -> Result<(), BootImageError> {
+        let total_size = total_size as usize;
+        assert_eq!((total_size & (PAGE_SIZE - 1)), 0);
+        // Create a page mapping at the next available virtual address.
```
peterfang: "next available" here means "using the same offset from the base of the kernel region"?

kernel/src/stage2.rs
```diff
@@ -812,142 +456,52 @@ pub extern "C" fn stage2_main(launch_info: &Stage2LaunchInfo) -> ! {
...
+    // Set the PML4E of the new kernel page tables in the current page table so
+    // the kernel address space is also visible in the current address space.
+    // SAFETY: the physical address of the current paging root is known to be
+    // identity-mapped in the current address space and therefore that address
+    // can be used to obtain a page table view.
     unsafe {
-        let kernel_launch_block =
-            &mut *launch_info_vaddr.as_mut_ptr::<MaybeUninit<KernelLaunchInfo>>();
-        kernel_launch_block.write(kernel_launch_info);
+        let vaddr = VirtAddr::new(u64::from(this_cpu().get_pgtable().cr3_value()) as usize);
```
peterfang:
```suggestion
        let vaddr = VirtAddr::new(u64::from(this_cpu().get_pgtable().cr3_value().bits()));
```

msft-jlange: That change doesn't work, but this does:
```
        let vaddr = VirtAddr::from(this_cpu().get_pgtable().cr3_value().bits());
```
which is cleaner than the original.

kernel/src/stage2.rs
msft-jlange: That's a stale comment from an earlier version.  I've cleaned it up to indicate that the mapping address is a local VA based on the specified physical address.

joergroedel (approved):
Testing on KVM with SNP passed.

PR #956, opened 2026-02-04
Update bytes to 1.11.1 and time to 0.3.47

stefano-garzarella:
`cargo audit` reports a security advisor about our `bytes` and `time` versions:

    Crate:     bytes
    Version:   1.11.0
    Title:     Integer overflow in `BytesMut::reserve`
    Date:      2026-02-03
    ID:        RUSTSEC-2026-0007
    URL:       https://github.com/advisories/GHSA-434x-w66g-qw3r
    Solution:  Upgrade to >=1.11.1
    Dependency tree:
    bytes 1.11.0
    ├── tower-http 0.6.7
    │   └── reqwest 0.12.24
    │       └── aproxy 0.1.0
    ├── reqwest 0.12.24
    ├── hyper-util 0.1.18
    │   └── reqwest 0.12.24
    ├── hyper 1.8.1
    │   ├── reqwest 0.12.24
    │   └── hyper-util 0.1.18
    ├── http-body-util 0.1.3
    │   └── reqwest 0.12.24
    ├── http-body 1.0.1
    │   ├── tower-http 0.6.7
    │   ├── reqwest 0.12.24
    │   ├── hyper-util 0.1.18
    │   ├── hyper 1.8.1
    │   └── http-body-util 0.1.3
    └── http 1.4.0
        ├── tower-http 0.6.7
        ├── reqwest 0.12.24
        ├── hyper-util 0.1.18
        ├── hyper 1.8.1
        ├── http-body-util 0.1.3
        └── http-body 1.0.1

    Crate:     time
    Version:   0.3.44
    Title:     Denial of Service via Stack Exhaustion
    Date:      2026-02-05
    ID:        RUSTSEC-2026-0009
    URL:       https://rustsec.org/advisories/RUSTSEC-2026-0009
    Severity:  6.8 (medium)
    Solution:  Upgrade to >=0.3.47
    Dependency tree:
    time 0.3.44
    ├── cookie_store 0.21.1
    │   └── reqwest 0.12.24
    │       └── aproxy 0.1.0
    └── cookie 0.18.1
        ├── reqwest 0.12.24
        └── cookie_store 0.21.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1334,9 +1334,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
 
 [[package]]
 name = "num-traits"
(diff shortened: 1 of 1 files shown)


msft-jlange:
Our `Cargo.toml` doesn't specifically request the `bytes` crate at all, so it seems that the version of `bytes` is being specified by whichever crate imports it.  Isn't that where the fix needs to be applied?  If not, can we insert a directive into `Cargo.toml` to demand the newer version?

My concern with this PR as it stands is that it should always be possible to delete `Cargo.lock` and have cargo regenerate it from scratch.  If we have to embed security-critical version information in the lock file, then we lose this property.  I think we need to make sure that the minimum version can be specified in some more durable file.

stefano-garzarella:
> It seems `time 0.3.47` requires `rustc 1.88.0`. @msft-jlange @peterfang @joergroedel that dependency is used only by the attestation proxy (aproxy), do you think it's okay to update our toolchain?

Rust toolchain updated with #968, so this needs to be rebased on that

PR #957, opened 2026-02-04
igvmbuilder: require tdx-stage1 for TDP

msft-jlange:
The TDP platform requires the presence of stage1 and will not boot otherwise.  The IGVM builder should generate an error if the TDP platform is requested but stage1 is not specified.

### DIFF
--- a/tools/igvmbuilder/src/gpa_map.rs
+++ b/tools/igvmbuilder/src/gpa_map.rs
@@ -105,6 +105,9 @@ impl GpaMap {
                 return Err("TDP platform must be specified when using --tdx-stage1".into());
             }
         } else {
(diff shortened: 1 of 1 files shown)


tools/igvmbuilder/src/gpa_map.rs
```diff
@@ -105,6 +105,9 @@ impl GpaMap {
                 return Err("TDP platform must be specified when using --tdx-stage1".into());
             }
         } else {
+            if COMPATIBILITY_MASK.contains(TDP_COMPATIBILITY_MASK) && options.tdx_stage1.is_none() {
```
peterfang: `options.tdx_stage1` is always `None` here, so we can drop the `is_none()` check

PR #958, opened 2026-02-05
insn_decode: clean up `InsnMachineCtx` / `InsnMachineMem`

00xc:
* Clean up the trait bounds for `InsnMachineMem`. The trait-associated type is always a generic, so simply make the trait generic.
* Clean up the returned value from `InsnMachineCtx::map_linear_addr()`. To allow for different types of returned values, use a trait-associated type.

### DIFF
--- a/kernel/src/insn_decode/decode.rs
+++ b/kernel/src/insn_decode/decode.rs
@@ -39,8 +39,6 @@
 // https://github.com/projectacrn/acrn-hypervisor/blob/master/hypervisor/
 // arch/x86/guest/instr_emul.c
 
-extern crate alloc;
-
 use super::insn::{DecodedInsn, Immediate, MAX_INSN_SIZE, Operand};
 use super::opcode::{OpCodeClass, OpCodeDesc, OpCodeFlags};
 use super::{InsnError, Register, SegRegister};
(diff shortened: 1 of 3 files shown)


msft-jlange (approved):

kernel/src/insn_decode/decode.rs
```diff
@@ -181,12 +180,12 @@ pub trait InsnMachineCtx: core::fmt::Debug {
     ///
     /// A `Result` containing a boxed trait object representing the mapped
     /// memory, or an `InsnError` if mapping fails.
```
stefano-garzarella: This documentation seems outdated now.

kernel/src/insn_decode/insn.rs
```diff
@@ -174,12 +171,15 @@ pub mod test_utils {
         }
     }
 
+    #[derive(Debug, Clone, Copy)]
     #[cfg_attr(not(test), expect(dead_code))]
-    struct TestMem<T> {
+    pub struct TestMem<T> {
```
stefano-garzarella, line 176: Why this needs to be public now?

stefano-garzarella: Aaaa, I see now `type Ptr<T: FromBytes + IntoBytes> = TestMem<T>;`

00xc: Yeah I should have clarified that perhaps, but since it's test code I didn't think it was worth it.

stefano-garzarella: yeah, it's fine ;-) thanks!

Issue #949, opened 2026-01-26
CI: use cargo-all-features


stefano-garzarella:
> FWIW clippy is not complaining, but you have a good point. With any new virtio driver (like#766), this probe would then depend on `block`, `vsock` or `somethingelse`.

@luigix25 `make clippy` uses `--all-features` but if you enable only `virtio-drivers` you will have:

```
$ cargo clippy --features virtio-drivers --package svsm --target x86_64-unknown-none 
   Compiling svsm v0.1.0 (/home/stefano/repos/coconut/svsm/kernel)
warning: unused variable: `slots`
   --> kernel/src/svsm.rs:166:13
    |
166 |     let mut slots = probe_mmio_slots(config);
    |             ^^^^^ help: if this is intentional, prefix it with an underscore: `_slots`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: variable does not need to be mutable
   --> kernel/src/svsm.rs:166:9
    |
166 |     let mut slots = probe_mmio_slots(config);
    |         ----^^^^^
    |         |
    |         help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default
```

I think we should start to use `cargo-all-features` like we did in rust-vmm to prevent these issues: https://github.com/rust-vmm/rust-vmm-ci/pull/192

_Originally posted by @stefano-garzarella in https://github.com/coconut-svsm/svsm/pull/948#discussion_r2728250172_

PR #959, opened 2026-02-05
CI: use cargo-hack to run clippy with feature combinations (plus fixes)

stefano-garzarella:
From https://github.com/taiki-e/cargo-hack: "Cargo subcommand to provide various options useful for testing and continuous integration."
    
Let's use cargo-hack with `--each-feature` in the Makefile's `clippy` target to check better our features and prevent issues:
```
--each-feature
    Perform for each feature of the package.
    
    This also includes runs with just --no-default-features flag, and
    default features.

    When this flag is not used together with --exclude-features (--skip)
    and --include-features and there are multiple features, this also
    includes runs with just --all-features flag.
```
    
In this way, each crate will be checked with all features enabled, no features enabled, default features, and then each individual feature. Using `--feature-powerset`, we could also try all combinations, but the CI time could easily explode, so for now, let's make a best effort.

Use it in our CI, but it's also available locally running `make clippy CARGO_HACK=1`.

Note: the first 5 patches are fixes for issues found by running  `make clippy CARGO_HACK=1`.

Closes: #949

### DIFF
--- a/Documentation/docs/developer/CONTRIBUTING.md
+++ b/Documentation/docs/developer/CONTRIBUTING.md
@@ -52,6 +52,30 @@ from the projects root directory.
 For detailed instructions on documentation guidelines please have a look at
 [RUSTDOC-GUIDELINES.md](RUSTDOC-GUIDELINES.md).
 
+Linting
+-------
+
+In addition to ```rustfmt```, it is recommended to run ```clippy``` before
+submitting changes to catch common mistakes and improve code quality.
+The Makefile provides a ```clippy``` target that runs it with the
+appropriate target and options for each crate:
+
+```
+make clippy
+```
+
+To check all feature combinations using ```cargo-hack```:
+
+```
+make clippy CARGO_HACK=1
+```
+
+This requires ```cargo-hack``` to be installed:
+
+```
+cargo install cargo-hack
+```
+
 Fuzzing
 -------
 


virtio-drivers/src/hal/fake.rs
```diff
@@ -6,12 +6,8 @@
 #![allow(missing_docs)]
 
 use crate::{BufferDirection, Hal, PAGE_SIZE, PhysAddr};
-use alloc::alloc::{alloc_zeroed, dealloc, handle_alloc_error};
-use core::{
-    alloc::Layout,
-    ptr::{self, NonNull},
-};
-use zerocopy::FromZeros;
+use std::alloc::{alloc_zeroed, dealloc, handle_alloc_error, Layout};
```
stefano-garzarella: @osteffenrh can you check commit https://github.com/coconut-svsm/svsm/pull/959/changes/ed7d477520bea332cd54968f0143c1d191ccd312 ?

I don't know if there were a specific reason to use `alloc` in FakeHal, so if we really need it, we should run test only with `alloc` feature

verification/verus_stub/src/lib.rs
```diff
@@ -12,5 +12,5 @@ pub use verus_builtin_macros::*;
 #[cfg(feature = "disable")]
 pub use vstd::prelude::*;
 
-#[cfg(not(feature = "disable"))]
+#[cfg(all(not(feature = "disable"), feature = "verus_macro_stub"))]
```
stefano-garzarella: @luigix25 can you check commit https://github.com/coconut-svsm/svsm/pull/959/changes/13167bbb6518011d1cb3b1a1ba819a523648b100

I honestly don't remember why we did this trick with `disable`, but as an alternative to this patch maybe we can have `verus_macro_stub` dependency not optional (it is an internal crate and also stub, so should be okay) and avoid this change here. I mean something like this:
```diff
--- a/verification/verus_stub/Cargo.toml
+++ b/verification/verus_stub/Cargo.toml
@@ -7,10 +7,10 @@ edition.workspace = true
 
 verus_builtin_macros = { workspace = true, optional = true }
 vstd = { workspace = true, optional = true }
-verus_macro_stub = { workspace = true, optional = true}
+verus_macro_stub = { workspace = true }
 
 [features]
-default = ["dep:verus_macro_stub"]
+default = []
 disable = ["verus_builtin_macros", "vstd"]
 
 [lints]
```

Now that I'm writing this comment, I like more this other option xD

virtio-drivers/src/hal/fake.rs
stefano-garzarella, line 1: @osteffenrh can you check commit https://github.com/coconut-svsm/svsm/pull/959/changes/194a45ee38a1f0e497478b9b06a796cdafe6a2d5 ?

I don't know if there were a specific reason to use `alloc` in FakeHal, so if we really need it, we should run test only with `alloc` feature

verification/verus_stub/src/lib.rs
luigix25: yeah, I prefer this last solution, much cleaner!

stefano-garzarella: done

virtio-drivers/src/hal/fake.rs
osteffenrh: Should be ok.

stefano-garzarella: @osteffenrh thanks for checking!

stefano-garzarella:
Discussed in the TSC meeting: we agreed on keeping the feature in the makefile but for now let's skip the CI to avoid to consume too much credits.

I'll update the PR.

PR #960, opened 2026-02-08
block: remove unnecessary Box wrapper

luigix25:
`VirtIOBlkDevice` is always wrapped in a Box. This is not necessary anymore, as the `BLOCK_DEVICE` struct is already wrapped in a Box.

### DIFF
--- a/kernel/src/block/virtio_blk.rs
+++ b/kernel/src/block/virtio_blk.rs
@@ -16,7 +16,7 @@ use virtio_drivers::transport::DeviceType::Block;
 
 extern crate alloc;
 use alloc::boxed::Box;
-pub struct VirtIOBlkDriver(Box<VirtIOBlkDevice>);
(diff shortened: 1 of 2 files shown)


luigix25:
Good point, you are right!
will do, thanks

kernel/src/block/virtio_blk.rs
```diff
@@ -25,8 +25,8 @@ impl core::fmt::Debug for VirtIOBlkDriver {
 }
 
 impl VirtIOBlkDriver {
-    pub fn new(slot: MmioSlot) -> Result<Self, SvsmError> {
-        Ok(VirtIOBlkDriver(VirtIOBlkDevice::new(slot)?))
+    pub fn new(slot: MmioSlot) -> Result<Box<Self>, SvsmError> {
```
stefano-garzarella, line 28: Please, can you mention also this change in the commit description and why we did it?

stefano-garzarella (approved):

PR #961, opened 2026-02-10
Update OpenSSL to 3.5.5

osteffenrh:
Update the OpenSSl submodule used by the TPM library, to the latest LTS version 3.5.5.
This requires stubbing `strpbrk()` in the C library.

Also disable additional unused algorithms in the OpenSSL/libcrypto build config.

### DIFF
--- a/libtcgtpm/Makefile
+++ b/libtcgtpm/Makefile
@@ -120,7 +120,6 @@ $(OPENSSL_MAKEFILE):
 			no-filenames \
 			no-gost \
 			no-http  \
-			no-hw \
 			no-idea \
 			no-ktls \
 			no-makedepend \


libtcgtpm/deps/libcrt/include/libcrt.h
```diff
@@ -402,3 +402,6 @@ typedef struct {
 } fd_set;
 
 #endif
+
+
+char* strpbrk (const char  *s, const char  *accept);
```
stefano-garzarella: This should be moved before the `#endif`

libtcgtpm/Makefile
```diff
@@ -106,11 +106,13 @@ $(OPENSSL_MAKEFILE):
 			no-dso \
 			no-dtls \
 			no-dtls1 \
+			no-dtls1-method \
 			no-dtls1_2 \
 			no-dtls1_2-method \
-			no-dtls1-method \
```
stefano-garzarella, line 111: nit: please mention the re-ordering in the commit message (it's fine to keep in this commit)

libtcgtpm/deps/libcrt/src/stub.c
```diff
@@ -194,3 +194,11 @@ time_t mktime(struct tm *tm) {
     NOT_IMPLEMENTED;
     return 0;
 }
+
+// string.h
+
+char *strpbrk (const char  *s, const char  *accept)
```
stefano-garzarella: nit:
```suggestion
char *strpbrk(const char *s, const char *accept)
```

libtcgtpm/deps/libcrt/include/libcrt.h
stefano-garzarella: nit:
```suggestion
char *strpbrk(const char *s, const char *accept);
```

libtcgtpm/Makefile
osteffenrh: Done

libtcgtpm/deps/libcrt/include/libcrt.h
osteffenrh: Done

libtcgtpm/Makefile
stefano-garzarella: Since we are touching this list, what about removing `no-hw`?

Running `make -C libtcgtpm distclean configure_openssl` I have the following warning:
```
***** Deprecated options: no-hw
```

osteffenrh: That has been there forever, but has no effect anymore (since 2019, https://github.com/openssl/openssl/commit/469ce8ff48ef06b2e508d0c06a42ec86379b0032). I'll remove that option.

stefano-garzarella (approved):

PR #968, opened 2026-02-16
Update Rust toolchain to 1.88.0  

stefano-garzarella:
Bump the Rust toolchain from 1.87.0 to 1.88.0, mainly for #956

Fixing all new clippy lints (`uninlined_format_args`, `borrow_as_ptr`, `manual_dangling_ptr`) and removing the deprecated `clippy::match_on_vec_items` lint. While investigating the `borrow_as_ptr` lint, a UB in `GDT::drop()` was discovered and fixed where `sgdt` was writing through an immutable reference.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.87.0"
+channel = "1.88.0"
 targets = [ "x86_64-unknown-none" ]


kernel/src/virtio/mmio.rs
```diff
@@ -77,15 +77,11 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
...
+            log::debug!("MmioSlots: {addr:x} empty");
             continue;
         };
 
-        log::info!(
-            "MmioSlots: Found {:?} at {:x}",
-            transport.device_type(),
-            addr
-        );
+        log::info!("MmioSlots: Found {:?} at {addr:x}", transport.device_type(),);
```
msft-jlange: Why is there an extra comma before the `)`?  Is this required somehow?  Shouldn't `rustfmt` remove it?

msft-jlange: Note that this occurs several times in this PR, and it's not clear to me whether this needs to be cleaned up before the PR merges.

stefano-garzarella: Yeah, let's clean it up. `cargo fmt` or `clippy` doesn't complain, but I agree with you, they are not needed.

PR #970, opened 2026-02-18
stage1: remove all non-TDX logic from stage1

msft-jlange:
The stage1 binary is now used only for TDP startup, and no longer has any relevance to SNP or native platforms.  All non-TDP code can be removed.


stage1/src/stage1.rs
```diff
@@ -7,231 +7,62 @@
...
+
+        /* Until the SIPI stub is consolidated with the boot loader stages,
+         * retain hard-coded constants in the assembly code.  These will be
+         * replaced once the SIPI stub is overhauled. */
+
+        /* Load the GDT from the SIPI stub. */
+        lgdt    [0xF01A]
+
+        /* Perform a far jump to the SIPI entry point. */
+        Ljmpl   $8, $0xF040
```
peterfang: Lowercase for consistency
```suggestion
        ljmpl   $8, $0xF040
```

stage1/src/stage1.rs
```diff
@@ -7,231 +7,62 @@
...
-
-    kernel_fs_bin:
-        .incbin "bin/svsm-fs.bin"
-    kernel_fs_bin_end:
-
-        .align 4
-    stage2_size:
-        .long   stage2_bin_end - stage2_bin
-    "#,
+        xorl    %esi, %esi
```
peterfang: Nit: no need to clear `%esi` since it must be 0 (passed the `testl` check)

tools/igvmbuilder/src/context.rs
```diff
@@ -340,3 +349,44 @@ pub fn construct_vmsa(
         vmsa: vmsa_box,
     }
 }
+
+pub fn construct_stage1_image(
+    stage1_path: &String,
+    gpa: u64,
+    regs: &[X86Register],
+    compatibility_mask: u32,
+) -> Result<IgvmDirectiveHeader, Box<dyn Error>> {
```
peterfang, line 358: I find it a bit odd that this is in `vmsa.rs`. Is this going to be moved somewhere else anytime soon?

tools/igvmbuilder/src/vmsa.rs
```diff
@@ -340,3 +349,44 @@ pub fn construct_vmsa(
...
+            tdp_context.rsp = *r as u32;
+        }
+    }
+
+    // Load the stage1 image.  It must not be larger than one page, and it must
+    // be at least as large as the context..
+    let mut stage1_image = fs::read(stage1_path)?;
+    if stage1_image.len() > PAGE_SIZE_4K as usize {
+        return Err("Stage1 image is too big".into());
+    }
```
peterfang: I think the stage1 image needs to be exactly one page, because it needs to cover the reset vector and also have the startup context located at exactly `0xFFFFF000` (as hardcoded in `stage1.rs`)

stage1/stage1.lds
peterfang: This block can be removed since there are no more `.startup.*` sections

peterfang: This can be removed as well

stage1/src/stage1.rs
```diff
@@ -7,231 +7,62 @@
...
 
-// Reset vector
-global_asm!(include_str!("reset.S"), options(att_syntax));
-
 global_asm!(
     r#"
-        .section ".startup.text","ax"
         .code32
+        .section .text
+        nop
```
peterfang: I don't think this is needed

stage1/src/stage1.rs
msft-jlange: The mixed-case version was prevalent at one point but I agree that all lower case looks better.

tools/igvmbuilder/src/context.rs
msft-jlange: After this change, the role of stage1 is really to be the TDX equivalent of start context (like the VMSA or the native context).  The code here inserts the context information (in the form of `TdpStartContext`) into the stage1 image so that execution can begin from the correct IGVM-generated context.  This is why this code is in the same place as the SNP and native context builders.  I'll plan to rename this file so that it applies to start contexts generally and not just VMSA.

PR #971, opened 2026-02-19
cross compile svsm on my macbook

kraxel:
Series of patches which allow me to cross compile svsm on my macbook.

 * This has some hacks which I do not consider mergeable.
 * This is not complete, I could get only parts of libtcgtpmto  actually compile.

So this is more an RfC.   Check the commit messages for details.

Hints how to solve the remaining issues or how to get rid of the hacks are welcome.


kraxel:
Obsoleted by https://github.com/coconut-svsm/svsm/pull/1106 + https://github.com/coconut-svsm/svsm/pull/981

Issue #885, opened 2025-11-25
Multiple broken tests under Miri


00xc:
Running all tests with Miri enabled:

```shell
MIRIFLAGS="-Zmiri-permissive-provenance" cargo +nightly miri test --workspace
```

The following generate Miri errors:
* `insn_decode::insn::tests::test_decode_ins_u8`
* `insn_decode::insn::tests::test_decode_ins_u16`
* `insn_decode::insn::tests::test_decode_ins_u32`
* `insn_decode::insn::tests::test_decode_mov_imm_to_reg`
* `insn_decode::insn::tests::test_decode_mov_reg_to_moffset`
* `insn_decode::insn::tests::test_decode_mov_reg_to_rm`
* `queue::tests::add_buffers_indirect`

The following get stuck in an infinite loop (or run extremely slowly):
* `device::blk::tests::device_id`
* `device::blk::tests::flush`
* `device::blk::tests::read`
* `device::blk::tests::write`

00xc:
Yes, I think I came to the conclusion that the secondary thread spawned in the test is not really executed by miri somehow, causing the main thread to block.

PR #974, opened 2026-02-20
kernel: fix tests under Miri

00xc:
Fix the core kernel tests mentioned in #885 when running under Miri. The first class of failures was due to using `as_ptr()` instead of `as_mut_ptr()`. The second class was more subtle, as invalid aliasing was introduced during instruction decoding in the `TestCtx` implementation.

Also, hide a `GuestPtr` test from Miri, since it uses inline assembly.

Note this does not fix the tests for virtio-drivers.

### DIFF
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -831,6 +831,7 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


joergroedel:
The miri fix for the instruction decoder broke the insn fuzzer, can you have a look? I am reverting that change for the upcoming release.

PR #977, opened 2026-02-25
fuzz: insn: provide valid pointer in TestCtx

00xc:
TestCtx was recently changed to take a pointer instead of storing a value, so feed in a valid pointer to prevent a NULL pointer dereference (which cannot happen under regular usage anyway).

Fixes: e29346da78fb ("kernel/insn_decode: tests: fix Miri errors")


joergroedel (approved):
This fixes the issue, thanks for the quick response.

PR #978, opened 2026-02-25
global_asm: ensure code is marked as `.text`

msft-jlange:
The `global_asm` macro doesn't provide any guarantees about where generated code is placed; it just inserts assembly code into the LLVM codegen stream, and the linker will place the assembly code into whichever section happens to be active at the time it encounters the assembly.  If that's a data section, then the code will wind up in a non-executable data area.  All assembly code must be explicitly marked with the correct section identifier (e.g. `.text`) to ensure that it will be correctly placed.

This change also makes the section choice for all assembly code explicit, placing global assembly into the section called `.text` instead of the current code section (which might be differently named based on compiler output).

### DIFF
--- a/kernel/src/cpu/idt/stage2.rs
+++ b/kernel/src/cpu/idt/stage2.rs
@@ -107,7 +107,7 @@ unsafe extern "C" {
 
 global_asm!(
     r#"
-        .text
+        .section .text
 
     generic_idt_handler_return:
         addq    $8, %rsp /* Skip ssp */
--- a/kernel/src/boot_stage2.rs
+++ b/kernel/src/boot_stage2.rs
@@ -17,7 +17,7 @@ use svsm::{
 
 global_asm!(
     r#"
-        .text
+        .section .text
         .section ".startup.text","ax"
         .code32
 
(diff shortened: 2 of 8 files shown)


kernel/src/cpu/idt/stage2.rs
```diff
@@ -107,7 +107,7 @@ unsafe extern "C" {
 
 global_asm!(
     r#"
-        .text
+        .section .text
```
peterfang, line 111: Any particular reason for not just sticking with the shorthand `.text`?

msft-jlange: According to Copilot, `.text` means the current default code section (which might have a different name, like `.text._RINvNtCs4kz9x7u7Q3_4test4main`), while `.section .text` always explicitly means the section named ".text".  When pressed, the argument fell apart a little bit, so it's possible the Copilot was hallucinating, but it is clear that the LLVM assembler only documents `.section` while `.text` comes from the GNU assembler, and it's reasonable to believe that they may have different effects on the output.  I favor explicit declarations to avoid confusion.

I made this change after I first pushed the PR, and the commit comment was updated but the PR comment wasn't, so I've updated the PR comment to match, which should hopefully make this a little more clear.

peterfang: Good point. Given the surprising linker behavior we observed here, this is probably a good idea.

PR #981, opened 2026-02-26
switch to rust linker

kraxel:
With this it is possible to cross-compile most¹ of svsm, because the rust linker is (unlike the binutils linker) not limited to the native architecture.  Tested by building svsm on Fedora/aarch64.

¹ The exception is the C code in libtcgtpm, i.e. the vtpm feature.

### DIFF
--- a/boot/bldr/src/bldr.lds
+++ b/boot/bldr/src/bldr.lds
@@ -31,7 +31,8 @@ SECTIONS
 		exception_table_end = .;
 	}
 	. = ALIGN(16);
-	.data : { *(.data) }
+	.data : { *(.data) *(.data.*) }
+	.got : { *(.got) *(.got.*) }
 	edata = .;
 	. = ALIGN(16);
 	.bss : {
@@ -45,7 +46,7 @@ SECTIONS
 	 * builder does not have to parse the ELF file to know how much space
 	 * to reserve for BSS. */
 	. = ALIGN(16);
-	.rodata : { *(.rodata) }
+	.rodata : { *(.rodata) *(.rodata.*) }
 }
 
 ENTRY(base_entry)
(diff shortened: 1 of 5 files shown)


luigix25:
Definitely not a linker expert, but tested on SNP and everything works smoothly!

boot/bldr/src/bldr.lds
```diff
@@ -31,7 +31,8 @@ SECTIONS
 		exception_table_end = .;
 	}
 	. = ALIGN(16);
-	.data : { *(.data) }
+	.data : { *(.data) *(.data.*) }
+	.got : { *(.got) *(.got.*) }
```
stefano-garzarella, line 35: Should we specify an alignment like for other sections?

I'm not sure; I just noticed that we do this for all the other sections, and I was wondering if it makes sense to do it for got as well.

stefano-garzarella (approved):
I can't see anything wrong except for the comment, which I think is more a nit, but I'd like another look from @joergroedel @peterfang @msft-jlange

boot/bldr/src/bldr.lds
kraxel: .got (aka global offset table) is data too.  Alignment should not be needed.  Also shouldn't hurt, so of this is preferred for consistency I can add it.

stefano-garzarella: I don't have a strong opinion on that TBH.
I just noticed the difference while reviewing, but I don't feel expert enough to suggest adding it or not.

joergroedel: I agree that alignment for the global offset table is not needed.

joergroedel (approved):

PR #992, opened 2026-03-05
igvmmeasure: transform crate into an library

kraxel:


tools/igvmmeasure/src/bin/igvmmeasure.rs
```diff
@@ -10,20 +10,14 @@ use std::fs::{self, File};
...
 use zerocopy::IntoBytes;
 
-use crate::id_block::SevIdBlockBuilder;
-
-mod cmd_options;
-mod id_block;
-mod igvm_measure;
-mod page_info;
-mod utils;
+use igvmmeasure::cmd_options::{CmdOptions, Commands, Platform};
```
00xc: There is no need to move this file to `src/bin` (and I'd say it's an atypical pattern). You should be able to do the conversion by updating Cargo.toml to:

```toml
[package]
name = "igvmmeasure"
version = "0.1.0"
edition.workspace = true

[[bin]]
name = "igvmmeasure"
path = "src/main.rs"
```

kraxel: > it's an atypical pattern

Hmm?  atypical?  It's cargo default behavior ...

The separation of library code and tools into different directories makes sense to me.  But, yes, it is not the only way to do it.  Matter of personal preference ...

stefano-garzarella:
@kraxel overall LGTM but can we improve a bit the commit descriptions by explaining the reasons behind these changes (and maybe even in the PR description). This is more to help newcomers.

For example, why removing `cmd_options` module? (I'm fine with it, just not fully clear to me the reason)

kraxel:
> @kraxel overall LGTM but can we improve a bit the commit descriptions by explaining the reasons behind these changes (and maybe even in the PR description). This is more to help newcomers.

The idea is (was?) to have a crate doing the mesurement calculations which can also be used elsewhere. 

With igvm working on that (see https://github.com/microsoft/igvm/issues/115, also community call two days ago (notes not yet online)) I'm not sure any more this is useful.  It is pretty clear now that this most likely would be some temporary solution only.

Opinions?

> For example, why removing `cmd_options` module? (I'm fine with it, just not fully clear to me the reason)

It just doesn't make sense to have that in the library crate as it is specific to the tool.

stefano-garzarella:
> > @kraxel overall LGTM but can we improve a bit the commit descriptions by explaining the reasons behind these changes (and maybe even in the PR description). This is more to help newcomers.
> 
> The idea is (was?) to have a crate doing the mesurement calculations which can also be used elsewhere.
> 
> With igvm working on that (see [microsoft/igvm#115](https://github.com/microsoft/igvm/issues/115), also community call two days ago (notes not yet online)) I'm not sure any more this is useful. It is pretty clear now that this most likely would be some temporary solution only.
> 
> Opinions?

IIUC it will take a while, so since we have this already ready, I'm fine to merge this, but I see that maybe we will remove all of these changes soon, so really up to you.

> 
> > For example, why removing `cmd_options` module? (I'm fine with it, just not fully clear to me the reason)
> 
> It just doesn't make sense to have that in the library crate as it is specific to the tool.

Yeah, I see, worth mentioning in the commit description IMO.

stefano-garzarella:
> > > @kraxel overall LGTM but can we improve a bit the commit descriptions by explaining the reasons behind these changes (and maybe even in the PR description). This is more to help newcomers.
> > 
> > 
> > The idea is (was?) to have a crate doing the mesurement calculations which can also be used elsewhere.
> > With igvm working on that (see [microsoft/igvm#115](https://github.com/microsoft/igvm/issues/115), also community call two days ago (notes not yet online)) I'm not sure any more this is useful. It is pretty clear now that this most likely would be some temporary solution only.
> > Opinions?
> 
> IIUC it will take a while, so since we have this already ready, I'm fine to merge this, but I see that maybe we will remove all of these changes soon, so really up to you.

I'll raise the point in the next TSC meeting (if it makes sense to continue to spend time on our tools or not).

stefano-garzarella:
@kraxel we discussed this in the last TSC. Since the work is already done, we are fine to merge this if it can help you in some way. But in long term we see the current code as temporary solution, so more in maintenance mode, as you also pointed out.

kraxel:
I have no code (yet) calculating measurements, and probably will just wait until the igvm crate supports this instead of going for a temporary solution.  So closing this PR.

PR #997, opened 2026-03-12
kernel: mm/vm: multiple cleanups and improvements

00xc:
This PR introduces two main improvements: 

#### 1. Removing the `Mapping` struct from the `VMR` code

The `Mapping` type is a wrapper of a `RWLock`, itself wrapping a boxed, dynamically dispatched `VirtualMapping` object. When a `Mapping` is inserted into a `VMR`, it is itself wrapped into an `Arc`, to keep track of its reference count. This introduces a double heap indirection that is simply not necessary.

There are no mutable users of `VirtualMapping` objects, and the trait requires `Sync`, so the `RWLock` is unnecessary overhead, and it can be removed. In the future, mappings that require mutation to service a `VirtualMapping` call should implement their own internal synchronization mechanism. This allows immutable `VirtualMapping` objects to have no synchronization overhead.

Moreover, `Arc` supports dyn-dispatch, so we can remove the `Box` as well. This leaves us with just an `Arc<dyn VirtualMapping>`. I left an alias of `Mapping = Arc<dyn VirtualMapping>` to reduce the diff size.

#### 2. Enforcing `VMR_GRANULE` size in `VMR::new()`

The `VMR` type expects to manage a section of the virtual address space aligned to `VMR_GRANULE`, and with the size being a multiple of `VMR_GRANULE`, although it does not make any checks in this regard. 

The task memory management code misuses `VMR` by creating `VMR`s with only subregions of the task address space, in order to keep a non-overlapping virtual address space across CPUs. Fix this by creating the `VMR` with the full task address space, and then inserting `VMReserved` mappings to prevent allocating addresses outside the intended subregion. Enforce this in `VMR::new()` to prevent future invalid uses.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -1161,7 +1161,8 @@ impl PerCpu {
     }
 
     pub fn handle_pf(&self, vaddr: VirtAddr, write: bool) -> Result<(), SvsmError> {
-        self.vm_range.handle_page_fault(vaddr, write)
+        let mut pgtable = self.get_pgtable();
+        self.vm_range.handle_page_fault(&mut pgtable, vaddr, write)
     }
 
     pub fn schedule_init(&self) -> TaskPointer {
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -762,13 +762,23 @@ impl PerCpu {
     }
 
     fn initialize_vm_ranges(&self) -> Result<(), SvsmError> {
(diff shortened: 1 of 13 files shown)


joergroedel (commented):
For some reason this breaks user-mode:

```
[SVSM] ERROR: Unexpected user-mode page-fault at RIP 0x0000000000010210 CR2: 0x0000000000010210 error code: 0x0000000000000014 - Terminating task
```

Otherwise I think the changes make sense. There is no real reason to keep the generic lock at the `Mapping` level. Mappings which need locks can implement them internally (or already do).

joergroedel (approved):
Testing passes now.

PR #998, opened 2026-03-12
xbuild: use packit as a library

00xc:
~As of right now, we have a git submodule with the packit utility because we use it as a regular binary from xbuild. Now that packit has a higher level to copy files into an archive, simply use it as a library instead. This removes the need for the git submodule, as the dependency can now be specified in Cargo.toml.~

*NOTE: kept the submodule, as discussed in the SVSM community call.*

As of right now, we build packit as a binary from xbuild in order to create the filesystem image to be embedded into the IGVM file. However, a recent change in packit simplifies using it as a Rust library, so use it as such to better integrate it into the codebase.

This unfortunately introduces issues with workspace-level feature unification, causing in some cases for packit to be linked with the combination of all of its features used in the workspace, which includes the "std" feature that xbuild needs, causing issues for the SVSM kernel. For now the only workaround in cargo is to use the `--package` option and adding `--exclude svsm` when using `--workspace`.

On the other hand, this lets us better specify that `SVSM_ARGS_TEST` in the Makefile should only be passed to the SVSM kernel target, and not the rest.

While we are at it, use workspace-level dependencies for xbuild.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,11 +116,14 @@ bin/coconut-test-vanadium.igvm:
 	cargo xbuild $(XBUILD_ARGS_TEST) ./configs/test/vanadium-test-target.json
 
 test:
-	cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --workspace
+	cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --package svsm
+	cargo test ${CARGO_ARGS} --workspace --exclude svsm
 
 miri:
 	MIRIFLAGS=-Zmiri-permissive-provenance \
-		cargo +nightly miri test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --workspace
+		cargo +nightly miri test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --package svsm
+	MIRIFLAGS=-Zmiri-permissive-provenance \
+		cargo +nightly miri test ${CARGO_ARGS} --workspace --exclude svsm
 
 test-igvm: $(IGVM_TEST_FILES)
 
@@ -176,7 +179,8 @@ clippy:
 	RUSTFLAGS="--cfg fuzzing" ${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm-fuzz -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm --target x86_64-unknown-none -- ${CLIPPY_ARGS}
 	${CARGO} clippy ${CLIPPY_OPTIONS} --package stage1 --target x86_64-unknown-none -- ${CLIPPY_ARGS} ${STAGE1_RUSTC_ARGS}
-	${CARGO} clippy ${CLIPPY_OPTIONS} --workspace --tests --exclude packit -- ${CLIPPY_ARGS}
+	${CARGO} clippy ${CLIPPY_OPTIONS} --workspace --tests --exclude svsm -- ${CLIPPY_ARGS}
+	${CARGO} clippy ${CLIPPY_OPTIONS} --package svsm ${SVSM_ARGS_TEST} --tests -- ${CLIPPY_ARGS}
 
 clean:
 	cargo clean
--- a/libtcgtpm/Makefile
+++ b/libtcgtpm/Makefile
@@ -6,6 +6,9 @@ OPENSSL_CONFIG_TYPE = --debug
 TCGTPM_CFLAGS = -g -O0 -DDEBUG=YES
 endif
 
+# Allow user to not link in libcrt
+USE_LIBCRT ?= 1
+
 ROOT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 DEPS_DIR := $(ROOT_DIR)/deps
 OUT_DIR ?= $(ROOT_DIR)
@@ -31,7 +34,11 @@ LIBCRT_BUILD_DIR = $(OUT_DIR)/libcrt-build
 OPENSSL_BUILD_DIR = $(OUT_DIR)/openssl-build
 TCGTPM_BUILD_DIR = $(OUT_DIR)/tcgtpm-build
 
+ifeq ($(USE_LIBCRT),1)
 LIBCRT = $(LIBCRT_BUILD_DIR)/libcrt.a
+else
+LIBCRT =
+endif
 LIBCRYPTO = $(OPENSSL_BUILD_DIR)/libcrypto.a
 
 LIBTPM_TARGET = Tpm_CoreLib
@@ -50,6 +57,17 @@ OPENSSL_MAKEFILE = $(OPENSSL_BUILD_DIR)/Makefile
 TCGTPM_MAKEFILE = $(TCGTPM_BUILD_DIR)/Makefile
 
 LIBS = $(LIBCRT) $(LIBCRYPTO) $(LIBBNMATH) $(LIBTPMBIGNUM) $(LIBTPM) $(LIBPLATFORM)
+
+ifeq ($(USE_LIBCRT),1)
+LIBCRT_INCLUDE = -I$(LIBCRT_SRC_DIR)/include
+LIBCRT_LINK = -Wl,rpath=$(LIBCRT_BUILD_DIR) -lcrt
+TCGTPM_CFLAGS += -static -nostdinc -fno-stack-protector -mno-sse -mno-red-zone
+OPENSSL_CFLAGS = -static -nostdinc -nostdlib -fno-stack-protector -mno-red-zone
+else
+LIBCRT_INCLUDE =
+LIBCRT_LINK =
+OPENSSL_CFLAGS =
+endif
 
 # Avoid running jobs of this Makefile in parallel to avoid possible conflicts
 # when building the TPM libraries, which use the same Makefile and subdirectory.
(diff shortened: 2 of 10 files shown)


msft-jlange:
We had a discussion among the TSC today about how to decide when it is preferable to use git submodules and when it is preferable to use github commit references in `Cargo.toml`.  The tradeoff between the two approaches comes down to the process that will be required when changes need to be made.  When modifications to a project require extensive changes to the SVSM code base - either because the project and the SVSM code are interdependent, or because the project's level of maturity is such that future changes are likely to be extensive - there is a strong advantage to keeping the project bound into the SVSM code base as a submodule.

Consider a case where changes to a separate project need to be made but cannot be made without making corresponding changes to SVSM because the SVSM changes are the only way to validate the changes being made to the separate project.  There is no way to submit a PR to the separate project without first testing it in SVSM, so there must be a way to test local changes by consuming the project in an SVSM build.  This requires not only enlisting in the project locally (which is easy) but also modifying the SVSM `Cargo.toml` to point at the local enlistment instead of github.  Then the SVSM changes will have to be developed on top of that, and there is a high probability that those changes will end up forming the basis for a future SVSM PR that consumes the updates to the project.  Naturally, that PR must exclude the local `Cargo.toml` change or else the build will break.  While this can certainly be accomplished as part of a development workflow, it requires manual steps that can easily be overlooked or performed incorrectly.

Contrast this with a git submodule, where the location of the separate project's source code is at a consistent location in all clones of the SVSM.  Development on both sides can be performed in place without requiring any customization of the local environment.  When development is complete and the PRs are ready to submit, the PR to the separate project can be submitted first, and once it has merged, the PR to the SVSM project can be submitted without further modification, since it will embed the correct commit hash for the submodule, and all other clones will be able to pick up the correct version.

The downside to the git submodule is that all SVSM clones must also embed a clone of the separate project, but when significant future development can be anticipated for the reasons stated above (maturity and/or interdependence), downside is outweighed by the more robust development process.

The view here is that the packit project is highly interdependent because its semantics must match the filesystem semantics of the SVSM kernel, and that future expansion of user-mode functionality is likely to require significant changes to those filesystem semantics - meaning that the packit project is also not sufficiently mature.  For these reasons, packit sources should remain tightly integrated in the form of a submodule.

There is an argument to be made that packit should not be a separate repo but should just be part of the SVSM repo, at least until the point that it has stabilized to the point that it can be spun out of the SVSM repo - and perhaps that should have been the approach from the beginning.  Either way, maintaining it as a project that is integrated directly via a github commit hash is the least desirable option.  If packit ever does reach a level of maturity and stability that it it makes the most sense to maintain it as a separate project, then it should be referenced as a github commit, but it's not there yet.

Makefile
```diff
@@ -116,11 +116,14 @@ bin/coconut-test-vanadium.igvm:
 	cargo xbuild $(XBUILD_ARGS_TEST) ./configs/test/vanadium-test-target.json
 
 test:
-	cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --workspace
+	cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --package svsm
+	cargo test ${CARGO_ARGS} --workspace --exclude svsm
```
stefano-garzarella, line 120: With this change, if I clean everything and run `make test` I have the same issue of the CI:

```
git clean -xfd
git submodule foreach --recursive git clean -xfd
git submodule sync
git submodule update
make test
...
   Doc-tests svsm

running 1 test
test kernel/src/mm/pagebox.rs - mm::pagebox::PageBox (line 38) - compile fail ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s

error: doctest failed, to rerun pass `-p svsm --doc`
make: *** [Makefile:119: test] Error 101
```

The strange thing is that if I swap the steps, everything is working, so not sure if it is something related to dependencies that `--workspace` build for docs.
I mean the following changes works, but I'd like to understand first why (maybe tomorrow):
```diff
diff --git a/Makefile b/Makefile
index 58832ee1..f6af6ddf 100644
--- a/Makefile
+++ b/Makefile
@@ -116,8 +116,8 @@ bin/coconut-test-vanadium.igvm:
        cargo xbuild $(XBUILD_ARGS_TEST) ./configs/test/vanadium-test-target.json
 
 test:
-       cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --package svsm
        cargo test ${CARGO_ARGS} --workspace --exclude svsm
+       cargo test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --package svsm
 
 miri:
        MIRIFLAGS=-Zmiri-permissive-provenance \
```

libtcgtpm/deps/openssl_svsm.conf
```diff
@@ -18,7 +18,7 @@ my %targets = (
                                           # exception 6 (invalid opcode) usually in:
                                           # vtpm_init()->libtpm:manufacture()->libcrypto:BN_CTX_new()
                                           release => "-O0"),
-                                   "-fPIE -m64 -nostdinc -nostdlib -static -fno-stack-protector -mno-red-zone")),
```
stefano-garzarella, line 21: Why this is needed?

Maybe we should cover it in the commit description.
e.g. `-mno-red-zone` was to fix https://github.com/coconut-svsm/svsm/issues/404

stefano-garzarella: Ah, ok, they are moved in the makefile.

libtcgtpm/Makefile
```diff
@@ -51,6 +58,16 @@ TCGTPM_MAKEFILE = $(TCGTPM_BUILD_DIR)/Makefile
 
 LIBS = $(LIBCRT) $(LIBCRYPTO) $(LIBBNMATH) $(LIBTPMBIGNUM) $(LIBTPM) $(LIBPLATFORM)
 
+ifeq ($(USE_LIBCRT),1)
+LIBCRT_INCLUDE = -I$(LIBCRT_SRC_DIR)/include
+LIBCRT_LINK = -Wl,rpath=$(LIBCRT_BUILD_DIR) -lcrt
+TCGTPM_CFLAGS += -static -nostdinc -fno-stack-protector -mno-sse -mno-red-zone
+OPENSSL_CFLAGS = -static -nostdinc -nostdlib -fno-stack-protector -mno-red-zone
+else
+LIBCRT_INCLUDE =
```
stefano-garzarella, line 67: Should we define an empty `OPENSSL_CFLAGS` here, or avoid defining all of them, I think it can also work. In any case I'd try to be consistent.

xbuild/src/fs.rs
```diff
@@ -57,19 +58,12 @@ impl FsConfig {
             }
             dst.push(dst_file);
             comp.config.objcopy.copy(&bin, &dst, args)?;
+
+            let filename = dst_file.to_str().expect("invalid file name");
```
stefano-garzarella: Why panic here instead of propagating the error?

Makefile
```diff
@@ -116,11 +116,14 @@ bin/coconut-test-vanadium.igvm:
...
 
 miri:
 	MIRIFLAGS=-Zmiri-permissive-provenance \
-		cargo +nightly miri test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --workspace
+		cargo +nightly miri test ${CARGO_ARGS} ${SVSM_ARGS_TEST} --package svsm
+	MIRIFLAGS=-Zmiri-permissive-provenance \
+		cargo +nightly miri test ${CARGO_ARGS} --workspace
```
stefano-garzarella: Should we exclude `svsm` here?

stefano-garzarella (commented):
I left some comments, but overall LGTM!

Makefile
stefano-garzarella: Yep, I just tried inverting the steps (because otherwise it will take too long) and it's failing:
```
$ make miri   
MIRIFLAGS=-Zmiri-permissive-provenance \
	cargo +nightly miri test  --workspace
error[E0204]: the trait `core::marker::Copy` cannot be implemented for this type
  --> kernel/src/fs/api.rs:23:10
   |
23 | #[derive(Copy, Clone, Debug, Default)]
   |          ^^^^
...
37 |     PackIt(PackItError),
   |            ----------- this field does not implement `core::marker::Copy`

error[E0277]: the trait bound `packit::PackItError: core::clone::Clone` is not satisfied
   --> kernel/src/fs/api.rs:37:12
    |
 23 | #[derive(Copy, Clone, Debug, Default)]
    |                ----- in this derive macro expansion
...
 37 |     PackIt(PackItError),
    |            ^^^^^^^^^^^ the trait `core::clone::Clone` is not implemented for `packit::PackItError`
    |
note: required by a bound in `core::clone::AssertParamIsClone`
   --> /home/stefano/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/clone.rs:357:34
    |
357 | pub struct AssertParamIsClone<T: Clone + PointeeSized> {
    |                                  ^^^^^ required by this bound in `AssertParamIsClone`

Some errors have detailed explanations: E0204, E0277.
For more information about an error, try `rustc --explain E0204`.
error: could not compile `svsm` (lib) due to 2 previous errors
make: *** [Makefile:123: miri] Error 101
```

libtcgtpm/deps/openssl_svsm.conf
00xc: Yes, exactly, these flags do not make sense if we are building for userspace.

stefano-garzarella (approved):

PR #1006, opened 2026-03-18
virtio: add device tree support

### LINKED ISSUE #833: Add Device Tree support
QEMU and SVSM are lacking device tree support for all the devices (ie virtio) attached.

luigix25:
Add device tree support for device enumeration.

This allows us to drop the fw_cfg interface used up until now.

~QEMU PR: https://github.com/coconut-svsm/qemu/pull/31~ Merged

The device tree is only used when building with the `virtio-drivers` features enabled. Since SVSM is looking for virtio-mmio devices, if the device tree is empty, nothing happens.

Fixes: #833


stefano-garzarella:
@luigix25 CI is failing, is that expected?

Also should we mark this as draft till we merge QEMU changes?

luigix25:
> @luigix25 CI is failing, is that expected?
> 
> Also should we mark this as draft till we merge QEMU changes?

@stefano-garzarella Yes, CI failure is expected as we need some QEMU changes. I'll mark it as draft for now.

stefano-garzarella:
> > @luigix25 CI is failing, is that expected?
> > Also should we mark this as draft till we merge QEMU changes?
> 
> @stefano-garzarella Yes, CI failure is expected as we need some QEMU changes. I'll mark it as draft for now.

mmm, but it seems a build issue when building tests (https://github.com/coconut-svsm/svsm/actions/runs/23251322695/job/67593804340?pr=1006):

```
[33](https://github.com/coconut-svsm/svsm/actions/runs/23251322695/job/67593804340?pr=1006#step:10:334)
   Compiling igvmbuilder v0.1.0 (/home/runner/work/svsm/svsm/tools/igvmbuilder)
error[E0063]: missing fields `dt_offset` and `dt_size` in initializer of `BootParamBlock`
   --> tools/igvmbuilder/src/igvm_builder.rs:258:12
    |
258 |         Ok(BootParamBlock {
    |            ^^^^^^^^^^^^^^ missing `dt_offset` and `dt_size`

For more information about this error, try `rustc --explain E0063`.
error: could not compile `igvmbuilder` (bin "igvmbuilder") due to 1 previous error
```

Are you sure it's related to QEMU?

luigix25:
- Rebased to latest main

@joergroedel @stefano-garzarella @osteffenrh PTAL :)

stefano-garzarella:
IIUC we still have a "[hack]" commit , so better to mark as draft till we can't remove it in order to avoid merging this by mistake.

luigix25:
> IIUC we still have a "[hack]" commit , so better to mark as draft till we can't remove it in order to avoid merging this by mistake.

yeah right, that commit switches the CI to use my qemu fork that has DT support.

kernel/src/virtio/mmio.rs
stefano-garzarella: Discussed in a private chat, but adding here to not forget.

DT is now carrying the MMIO area size, we should use it.

kernel/src/fw_cfg.rs
stefano-garzarella, line 1: do we still need the fw_cfg ?

kernel/src/types.rs
```diff
@@ -48,11 +48,14 @@ pub const SVSM_CS_ATTRIBUTES: u16 = 0xa09b;
 pub const SVSM_DS_ATTRIBUTES: u16 = 0xc093;
 pub const SVSM_TR_ATTRIBUTES: u16 = 0x89;
 
+/// VMPL level SVSM will be executed at.
+pub const SVSM_VMPL: usize = 1;
```
stefano-garzarella: mmm, is this related ?

kernel/src/types.rs
```diff
@@ -48,11 +48,14 @@ pub const SVSM_CS_ATTRIBUTES: u16 = 0xa09b;
...
 /// VMPL level the guest OS will be executed at.
 /// Keep VMPL 1 for the SVSM and execute the OS at VMPL-2. This leaves VMPL-3
 /// free for the OS to use in the future.
 pub const GUEST_VMPL: usize = 2;
 
+const _: () = assert!(SVSM_VMPL > 0 && SVSM_VMPL < GUEST_VMPL);
```
stefano-garzarella: ditto

tools/igvmbuilder/src/igvm_builder.rs
```diff
@@ -488,6 +497,18 @@ impl IgvmBuilder {
...
             initial_data: vec![],
         });
+        if self
+            .gpa_map
+            .boot_param_layout
+            .get_param_size(BootParamType::DeviceTree)
+            != 0
+        {
+            self.directives.push(IgvmDirectiveHeader::ParameterArea {
+                number_of_bytes: PAGE_SIZE_4K,
```
stefano-garzarella: Why this is hard coded? Also why that value.

tools/igvmbuilder/src/igvm_builder.rs
```diff
@@ -59,7 +59,8 @@ pub const ANY_NATIVE_COMPATIBILITY_MASK: u32 = NATIVE_COMPATIBILITY_MASK | VSM_C
 const IGVM_GENERAL_PARAMS_PA: u32 = 0;
 const IGVM_MEMORY_MAP_PA: u32 = 1;
 const IGVM_MADT_PA: u32 = 2;
-const IGVM_PARAMETER_COUNT: u32 = 3;
+const IGVM_DT_PA: u32 = 3;
+const IGVM_PARAMETER_COUNT: u32 = 4;
```
stefano-garzarella, line 62: So now this is 4 for all hypervisor, but DT is used only with QEMU. Is this okay?

tools/igvmbuilder/src/boot_params.rs
```diff
@@ -37,11 +40,18 @@ impl BootParamLayout {
         let general_param_offset = page_size + guest_context_size;
         let madt_offset = general_param_offset + page_size;
         let memory_map_offset = madt_offset + page_size;
-        let total_size = memory_map_offset + page_size;
+        let (device_tree_offset, device_tree_size) = if include_device_tree {
```
stefano-garzarella: Can we just add this for all hypervisors? What will be the issue?

kernel/src/virtio/mmio.rs
```diff
@@ -29,9 +28,28 @@ pub struct MmioSlots {
     slots: Vec<MmioSlot>,
 }
 
+/// Parses the device tree to extract base addresses of virtio-MMIO devices
+/// assigned to the SVSM plane.
+fn get_virtio_mmio_addresses(device_tree: Fdt<'_>) -> Vec<u64> {
+    device_tree
+        .find_all_nodes("/virtio_mmio")
```
stefano-garzarella: Should this be documented somewhere ?

I mean this is a contract with the VMM that should use exactly `virtio_mmio` and not `virtio-mmio` or `virtio mmio` for example.

kernel/src/virtio/mmio.rs
```diff
@@ -29,9 +28,28 @@ pub struct MmioSlots {
...
+        .filter(|dev| {
+            dev.compatible()
+                .is_some_and(|c| c.all().any(|c| c == "virtio,mmio"))
+        })
+        .filter(|dev| {
+            dev.property("plane")
```
stefano-garzarella: Okay, I think all of this must be documented somewhere, and also mention if are optional or not. E.g. why we need plane here? Can the VMM just expose the device for SVSM plane or there is an use case where we need to expose also devices for other planes ?

kernel/src/virtio/mmio.rs
```diff
@@ -44,22 +62,24 @@ pub struct MmioSlots {
...
-/// Returns an empty collection if no devices are found or if the fw_cfg interface
+/// Returns an empty collection if no devices are found or if the device tree
 /// is unavailable.
 pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
-    // Virtio MMIO addresses are discovered via fw_cfg, so skip probing
+    // Virtio MMIO addresses are discovered via device tree, so skip probing
     // if it is not present.
-    if !boot_params.has_fw_cfg_port() {
+    let Some(device_tree) = boot_params.get_device_tree() else {
+        log::warn!("MmioSlots: No device tree found.");
```
stefano-garzarella: Should this be really a warning ?

stefano-garzarella (changes requested):
The big missing part here IMO is documentation about what SVSM expects in the device tree. I mean something like `Documentation/devicetree` in Linux. (e.g. `Documentation/devicetree/bindings/virtio/mmio.yaml`, etc.

kernel/src/virtio/mmio.rs
luigix25: what do you suggest?

kernel/src/types.rs
luigix25: if you look in the `get_virtio_mmio_addresses` function, I need to check if a device is meant for SVSM or the guest. I didn't want to hardcode a magic number, so I added a new const

stefano-garzarella: ah, this is not mentioned at all in the commit description. IMO better to split in another commit.

Also, are you sure SVSM is running in VMPL1? Isn't VMPL0?

kernel/src/virtio/mmio.rs
```diff
@@ -29,9 +28,28 @@ pub struct MmioSlots {
...
+    device_tree
+        .find_all_nodes("/virtio_mmio")
+        .filter(|dev| {
+            dev.compatible()
+                .is_some_and(|c| c.all().any(|c| c == "virtio,mmio"))
+        })
+        .filter(|dev| {
+            dev.property("plane")
+                .and_then(|p| p.as_usize())
+                .is_some_and(|plane| plane == SVSM_VMPL)
```
stefano-garzarella: Is this expected also with `native` platform where we don't have VMPLs ?

kernel/src/virtio/mmio.rs
stefano-garzarella: In the previous code, if fw_cfg didn't provide any slot we didn't print anything, so I'm asking why now this need to be a warning or not an info or debug ?
Why we are worried about this?

kernel/src/types.rs
stefano-garzarella: that said, I think we can just use the already defined `VMPL0`, since SVSM can only run on it. while for the guest was a choice to select 2, but it can be 1 or 3, so for that reason we added a const to keep track of it.

kernel/src/virtio/mmio.rs
luigix25: currently the plane property is fake. It's just a value that is passed via the CLI (see launch_guest) so even on native platform we can have "planes".

tools/igvmbuilder/src/igvm_builder.rs
luigix25: good question. I think 4k should be more than enough for a full device tree. Currently DT size is around ~1k.

This is why I chose this value. Do you have something else in mind?

kernel/src/virtio/mmio.rs
stefano-garzarella: okay, but in the future how we plan to handle it? Should we mark `plane` as optional? Or should we ask to have it in any case set to 0 to assign a device to svsm?

We can fix later, but we need to add a big TODO here. (I prefer to define this now TBH)

tools/igvmbuilder/src/igvm_builder.rs
luigix25: For qemu IIRC this is not a problem. We had this "issue" in the past with MADT support.
I can't talk for hyper-v or vanadium. To be safe I should add an if, so it's consistent for all HVs

tools/igvmbuilder/src/igvm_builder.rs
stefano-garzarella: why need to be a fixed size ?

kernel/src/virtio/mmio.rs
luigix25: this is a _very_ good question, I think it depends on how the planes will be integrated in QEMU. IIRC (maybe @joergroedel can correct me) in the qemu branch you can still override the planes property, even in native mode.

tools/igvmbuilder/src/igvm_builder.rs
stefano-garzarella: For example for MADT we have:
```
            number_of_bytes: self
                .gpa_map
                .boot_param_layout
                .get_param_size(BootParamType::Madt) as u64,

```

Why we can't do the same here?

luigix25: IIUC this is how igvm works, you tell the HV how much memory you want to allocate, and the HV will do that for you. Disclaimer: I'm not an igvm expert

luigix25: and on top of that igvm-builder doesn't know how big the DT will be

luigix25: > For example for MADT we have:
> 
> ```
>             number_of_bytes: self
>                 .gpa_map
>                 .boot_param_layout
>                 .get_param_size(BootParamType::Madt) as u64,
> ```
> 
> Why we can't do the same here?

oh yes absolutely! But if you look closely, you have to hardcode the amount of memory somewhere

stefano-garzarella: I'm not an expert too, but I don't understand the difference with MADT TBH

luigix25: MADT is hardcoded to 4k as well.

`get_param_size` returns 4K for MADT.

DT is (or at least should be) disabled for all HVs but qemu. So by default size is 0

stefano-garzarella: Ah, I see now `BootParamLayout::new()`, but now in this case we have this info spread in 2 places, and we don't check in any place that they are the same... there we use `page_size` independently of what we define in `number_of_bytes`, or I'm wrong? (I don't know this code)

stefano-garzarella: IIRC OpenHCL already used DT in IGVM, can we check how much space they reserve?

TBH 4K doesn't seem too much

luigix25: > Ah, I see now `BootParamLayout::new()`, but now in this case we have this info spread in 2 places, and we don't check in any place that they are the same... there we use `page_size` independently of what we define in `number_of_bytes`, or I'm wrong? (I don't know this code)

yep, this is a problem in the code (probably due to a rebase). I'll use `get_param_size`

stefano-garzarella: So, going back, exactly because `get_param_size` can return 0 or 4k, should we use it here like we do for all other members ?

luigix25: yes!

kernel/src/virtio/mmio.rs
luigix25: The idea is that the HV exposes the full DT, SVSM gets what it's on its plane, removes those entries and forwards the DT to the firmware. 

(see https://github.com/coconut-svsm/qemu/pull/31#issuecomment-4132410994)

Note that edk2 is not able to consume a DT from igvm yet. So this is a long term plan

kernel/src/virtio/mmio.rs
luigix25: right, I can drop the log, like it was before.

kernel/src/virtio/mmio.rs
luigix25: mmh good point. What about removing the filter on /virtio_mmio and using just the compatible part? 
I'm using the linux compatible convention: https://www.kernel.org/doc/Documentation/devicetree/bindings/virtio/mmio.txt

stefano-garzarella: I don't know TBH. My point is more related to the documentation.

luigix25: ok, i'll write something in the doc. I'll also drop the filtering on `/virtio_mmio`, it was not necessary and may cause problems.

kernel/src/fw_cfg.rs
luigix25: not really! After this PR fw_cfg can be dropped entirely.
I think there is some leftover code for MADT that is not used anymore.

stefano-garzarella: cool, so let's add a commit to remove it completely

kernel/src/virtio/mmio.rs
stefano-garzarella: OKay, now it makes sense. Sorry, I was missing this point, I thought we needed DT just for SVSM.

kernel/src/fw_cfg.rs
luigix25: yeah i'll open a new PR for that, as fw_cfg MADT support was already removed some time ago.
Then this PR will remove the remaining parts.

stefano-garzarella: yep, even better!

luigix25: See here #1090!

tools/igvmbuilder/src/boot_params.rs
luigix25: very good question: here i set the device tree size. Later on, using the size I add the device tree directive into the igvm file.

Although there are optional and mandatory directives, QEMU considers them all as mandatory. This means that it will exits if it can't handle it properly.

I'm already working on a fix for QEMU.

I'm not sure if the other HVs are affected by the same bug.

luigix25:
- Now reading size from the device tree
- Now using coconut-qemu as it has device tree support
- Fixed filtering on the device tree properties
- Splitted work across more commits
- Removed fw_cfg entirely
- Rebased to latest main

TODO:

- [ ] Add some doc
- [ ] Fix parameter count / Enable DT for all HVs?

Notes:
- Commit 1 is from #1167 adding it here because they conflict

kernel/src/virtio/mmio.rs
luigix25: This is still an open question, coconut-qemu by default assigns all devices to plane 0, so they 'belong' to SVSM. In native mode this works fine as is.

Without the planes check we could use upstream QEMU (using microvm) to run test-in-svsm as it provides a device tree, but without the `plane` property.

tools/igvmbuilder/src/igvm_builder.rs
luigix25: Still an open question: it's not an easy fix. Let's keep this open for now for other reviewers.

kernel/src/types.rs
luigix25: you were right, SVSM runs at VMPL0. Not sure what const are you talking about though. 

I moved that change to a different commit.

stefano-garzarella: > Not sure what const are you talking about though.

I was talking about `VMPL0` constant.

stefano-garzarella: > > Not sure what const are you talking about though.
> 
> I was talking about `VMPL0` constant.

ah, you were referring to `that reason we added a const to keep track of it`. Let me rephrase: SVSM can only run at VMPL = 0, while the guest can run at levels 1 through 3. So we added GUEST_VMPL to keep track of which level we use for the guest (maybe one day it can be a runtime parameter), but for SVSM it will always be 0. That said I'm also fine with adding a new const if you think it improves the readability.

tools/igvmbuilder/src/boot_params.rs
stefano-garzarella: IIUC this is related to https://github.com/coconut-svsm/svsm/pull/1164#issuecomment-4978675817
Right?

luigix25: yep, correct!

kernel/src/virtio/mmio.rs
stefano-garzarella: Okay, this must be documented in the DT documentation we are going to add, so other hypervisor can do the same.

tools/igvmbuilder/src/igvm_builder.rs
luigix25: @msft-jlange probably you can help on this: if we set the device tree only for QEMU, "IGVM_PARAMETER_COUNT" in this wasy, it's wrong.

Does HV tolerate this or it would throw an error? IIRC we had the same problem when MADT via igvm was not available in QEMU but available for other HVs.

kernel/src/virtio/mmio.rs
luigix25: added some doc.

luigix25:
- Rebased to latest main
- Added documentation for the device tree
- Removed a TODO about mmio size (see https://github.com/coconut-svsm/svsm/pull/1167) 

TODO:
- [ ] Fix parameter count / Enable DT for all HVs?

luigix25:
Changes: 
- Plane property is only when PLANES are available in qemu.
- Fixed page boundary check

tools/igvmbuilder/src/gpa_map.rs
```diff
@@ -153,7 +153,8 @@ impl GpaMap {
             Some(fw) => fw.get_guest_context().is_some(),
             None => false,
         };
-        let boot_param_layout = BootParamLayout::new(include_guest_context);
+        let include_device_tree = matches!(options.hypervisor, Hypervisor::Qemu);
```
stefano-garzarella: IIUC from @msft-jlange and @AdamCDunlap we can expect both hyperv and vanadium be able to support it with an empty DT, so maybe we can remove this at all

scripts/launch_guest.sh
stefano-garzarella: Pre-existing, but now that we introduced `SVSM_VMPL=0`, should we introduce also GUEST_VMPL and use it here?

scripts/launch_guest.sh
```diff
@@ -55,7 +61,9 @@ while [[ $# -gt 0 ]]; do
     --state)
       VIRTIO=1
       STATE_DEVICE+="-drive file=$2,format=raw,if=none,id=svsm_storage,cache=none "
-      STATE_DEVICE+="-device virtio-blk-device,drive=svsm_storage "
+      STATE_DEVICE+="-device virtio-blk-device,drive=svsm_storage"
+      [ -n "$QEMU_PLANES" ] && STATE_DEVICE+=",plane=$SVSM_VMPL"
+      STATE_DEVICE+=" "
```
stefano-garzarella: Do we really need this?

scripts/launch_guest.sh
```diff
@@ -90,7 +98,9 @@ while [[ $# -gt 0 ]]; do
       ;;
     --vsock)
       VIRTIO=1
-      VSOCK_DEVICE="-device vhost-vsock-device,guest-cid=$2 "
+      VSOCK_DEVICE="-device vhost-vsock-device,guest-cid=$2"
+      [ -n "$QEMU_PLANES" ] && VSOCK_DEVICE+=",plane=$SVSM_VMPL"
+      VSOCK_DEVICE+=" "
```
stefano-garzarella: Ditto.

kernel/src/virtio/mmio.rs
```diff
@@ -117,7 +117,7 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
             continue;
         }
 
-        if phys_addr.crosses_page(core::mem::size_of::<VirtIOHeader>()) {
```
stefano-garzarella, line 83: Should we squash this fix when we introduce the discovery via device-tree? (for bisectability)

scripts/launch_guest.sh
```diff
@@ -35,6 +35,12 @@ STATE_DEVICE=""
 VSOCK_DEVICE=""
 VIRTIO=0
 
+# Must match SVSM_VMPL in kernel/src/types.rs
+SVSM_VMPL=0
+
+# Check for support for device-plane property
+QEMU_PLANES=$($QEMU -h | grep -o "device-plane=" || true)
```
stefano-garzarella: What happen here if `--qemu` is used to specify a different exec ?

scripts/launch_guest.sh
stefano-garzarella, line 1: Can we move changes to this script in a separate commit (that comes before this) or it will not work ?

kernel/src/virtio/mmio.rs
```diff
@@ -99,17 +136,17 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
         // SAFETY: The address is valid, mapped by `map_global_range_4k_shared`, and verified
         // to be VirtIOHeader-aligned by the guard above.
         // The memory region has the same lifetime of the MmioSlot structure which will be consumed by the driver.
-        // TODO: Currently the hypervisor does not advertise the size of the mmio space (header + config space).
-        //       The size will be provided by the device tree. For now, let's just make sure we don't go past
-        //       the boundaries of the allocated space.
-        let Ok(transport) = (unsafe { MmioTransport::new(header, mem.size() - page_offset) })
-        else {
+        let Ok(transport) = (unsafe { MmioTransport::new(header, slot.size) }) else {
```
stefano-garzarella, line 133: Should we consider `page_offset` too ? IIRC we fixed in some recent PR and maybe forgot to update here.

Documentation/docs/developer/design/DEVICE-TREE.md
```diff
@@ -0,0 +1,89 @@
...
+
+Follows the [Devicetree Specification v0.4 §3.2](https://github.com/devicetree-org/devicetree-specification/releases/tag/v0.4)
+with the addition of the `device-plane` property.
+
+| Property         | Type   | Required | Description                                |
+|------------------|--------|----------|--------------------------------------------|
+| `compatible`     | string | yes      | Must be `"coconut,svsm"`.                  |
+| `#address-cells` | u32    | yes      | Number of cells to encode child addresses. |
+| `#size-cells`    | u32    | yes      | Number of cells to encode child sizes.     |
+| `device-plane`   | u32    | yes      | The VMPL the guest OS runs at.             |
```
stefano-garzarella: Is `device-plane` supported/used right now?

kernel/src/virtio/mmio.rs
```diff
@@ -117,7 +117,7 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
             continue;
         }
 
-        if phys_addr.crosses_page(core::mem::size_of::<VirtIOHeader>()) {
+        if phys_addr.crosses_page(slot.size) {
             log::warn!("MmioSlots: MMIO device header at {phys_addr:x} crosses a page boundary");
```
stefano-garzarella: Should we update this log message since now we consider the entire MMIO device region ?

stefano-garzarella (commented):
I left some comments, but overall LGTM!

scripts/launch_guest.sh
stefano-garzarella: I mean `STATE_DEVICE+=" "` IMO we can remove it, since we add the space later when we use the variable.

Documentation/docs/developer/design/DEVICE-TREE.md
luigix25: is not used, same goes for interrupts.

kernel/src/virtio/mmio.rs
luigix25: good point, yes

scripts/launch_guest.sh
luigix25: oops it would not work, will fix it.

kernel/src/virtio/mmio.rs
luigix25: If you are referring to `- page_offset` it should not be considered! That was added because we always mapped a whole page, so if the address started at page + offset, we could go out of bounds.

Now the hypervisor tells us how big the slot is and we are sure that we don't go out of bounds because of the check:

```rust
if phys_addr.crosses_page(slot.size) {
```

Documentation/docs/developer/design/DEVICE-TREE.md
stefano-garzarella: so I guess we should introduce them only when supported, WDYT?

scripts/launch_guest.sh
luigix25: it will not work. I was thinking about using `[ -n "$SNP_GUEST" ] && [ -n "$QEMU_PLANES" ]` for consistency in the script. WDYT?

kernel/src/virtio/mmio.rs
stefano-garzarella: Oh yeah, I was confused by the start address where we stil use it. LGTM!

kernel/src/virtio/mmio.rs
luigix25: I'm fine with squashing, that said, probably I'm missing something, because I don't see why bisectability is a problem here.

Documentation/docs/developer/design/DEVICE-TREE.md
luigix25: ok, will drop them

scripts/launch_guest.sh
luigix25: yes, why not!

stefano-garzarella: What about adding an `if` block later in the script to fix all the parameters, so it's more clear what we need to add for SNP_GUEST and QEMU_PLANES. I mean combine changes to MACHINE, STATE_DEVICE, etc. in a single place, instead of spread them all around the file.

Not a strong opinion, just to avoid having `[ -n "$SNP_GUEST" ] && [ -n "$QEMU_PLANES" ] ` all over the script.

kernel/src/virtio/mmio.rs
stefano-garzarella: I'm missing something too, why this need to be fixed later and not when we start to support device tree ?

stefano-garzarella: IIUC we are doing this change because the region can cross the page boundary when using the DT, so if it happens, this commit will work okay, while when we will run the commit where we introduced the DT will fail, or am I missing something?

luigix25: ok right, i'll squash it

luigix25:
@stefano-garzarella 

- Marked the device tree as mandatory for all HVs
- Removed `interrupts` and `device-plane` property from the device tree spec as they are not used
- launch script now checks only in one place for the planes support
- introduced GUEST_VMPL in the launch_script
- squashed the fix of the boundary check

stefano-garzarella (approved):

PR #1012, opened 2026-03-24
docs: move debugging instructions in a dedicated file

n-ramacciotti:
Move debug information in a dedicated file (`DEBUGGING.md) in the developer folder.

Add new information and instructions to launch a debug session.

Avoid tracking `.gdb_history`

Fix a typo in a gdb command in the documentation.

More information are available in the commits description.

### DIFF
--- a/Documentation/docs/developer/DEBUGGING.md
+++ b/Documentation/docs/developer/DEBUGGING.md
@@ -46,8 +46,11 @@ virtual machine is started. You can then connect GDB to the waiting SVSM using
 the command, replacing `[n]` with the correct device index:
 
 ```shell
-sudo gdb --ex "target extended-remote /dev/pts/[n]`
+sudo gdb --ex "target extended-remote /dev/pts/[n]"
 ```
+
+**Note**: If you don't have acces to confidential hardware, you don't need to use
+```sudo```.
 
 If you have the source code available on the host system then you can add the
 debug symbols and use source-level debugging:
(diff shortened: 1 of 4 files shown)


Documentation/docs/developer/DEBUGGING.md
```diff
@@ -0,0 +1,71 @@
...
+[SVSM] ***********************************
+```
+
+Debug breakpoint can be inserted using the `debug_break` function defined in
+[svsm_gdbstub](../../rustdoc/svsm/debug/gdbstub/svsm_gdbstub) module.
+Make sure that feature `enable-gdb` is enabled, otherwise a call to that function
+will not have any effect.
+
+**Note**: Currently, there is a call to `debug_break` in the `svsm_init` function in
+[svsm.rs](https://github.com/coconut-svsm/svsm/blob/main/kernel/src/svsm.rs), but it is
```
n-ramacciotti, line 36: Is there a better way to highlight `svsm_init`?

00xc: You can link to a particular line in a particular commit, but it will get outdated at some point, so I'd say this is the best way to do it.

00xc (approved):

luigix25 (approved):

Documentation/docs/developer/DEBUGGING.md
n-ramacciotti: That makes sense. Thanks! I'll keep this approach and mark this conversation as solved

joergroedel (approved):
Make much sense, thanks.

PR #1028, opened 2026-04-01
kernel: cpu/tlb: enable partial per-CPU TLB flushes

00xc:
Now that partial SMP TLB flushes are enabled using `TlbFlushScope`, update the per-CPU TLB flush methods to use the type as well. This also includes adding `TlbFlushScope` constructor methods for convenience.

### DIFF
--- a/kernel/src/cpu/tlb.rs
+++ b/kernel/src/cpu/tlb.rs
@@ -95,8 +95,8 @@ impl TlbFlushScope {
     /// Flushes all the entries in the TLB for the current CPU.
     fn flush_percpu_all(&self) {
         match self.global {
-            true => flush_tlb_global_percpu(),
-            false => flush_tlb_percpu(),
+            true => __flush_tlb_global_percpu(),
+            false => __flush_tlb_percpu(),
         }
     }
 
@@ -147,6 +147,26 @@ pub fn flush_tlb_global_sync_page(vaddr: VirtAddr, pgsize: PageSize) {
 }
 
 pub fn flush_tlb_global_percpu() {
+    TlbFlushScope::all().with_global(true).flush_percpu();
+}
+
+pub fn flush_tlb_global_percpu_range(region: MemoryRegion<VirtAddr>, pgsize: PageSize) {
+    TlbFlushScope::range(region, pgsize)
+        .with_global(true)
+        .flush_percpu();
+}
+
+pub fn flush_tlb_global_percpu_page(vaddr: VirtAddr, pgsize: PageSize) {
+    TlbFlushScope::page(vaddr, pgsize)
+        .with_global(true)
+        .flush_percpu();
+}
+
+pub fn flush_tlb_percpu() {
+    TlbFlushScope::all().with_global(false).flush_percpu();
+}
+
+fn __flush_tlb_global_percpu() {
     let cr4 = read_cr4();
 
     // SAFETY: we are not changing any execution-state relevant flags
@@ -156,7 +176,7 @@ pub fn flush_tlb_global_percpu() {
     }
 }
 
-pub fn flush_tlb_percpu() {
+fn __flush_tlb_percpu() {
     // SAFETY: reloading CR3 with its current value is always safe.
     unsafe {
         write_cr3(read_cr3());
--- a/kernel/src/cpu/tlb.rs
+++ b/kernel/src/cpu/tlb.rs
@@ -43,6 +43,36 @@ pub struct TlbFlushScope {
 }
 
 impl TlbFlushScope {
+    /// Creates a TLB flush scope for all all addresses (including global).
+    pub const fn all() -> Self {
+        Self {
+            global: true,
+            range: TlbFlushRange::All,
+        }
+    }
+
+    /// Updates the scope to include or exclude global pages, as indicated.
+    pub const fn with_global(mut self, global: bool) -> Self {
+        self.global = global;
+        self
+    }
+
+    /// Creates a new TLB flush scope, including only a range of virtual addresses
+    /// (which may be global).
+    pub const fn range(region: MemoryRegion<VirtAddr>, pgsize: PageSize) -> Self {
+        Self {
+            global: true,
+            range: TlbFlushRange::Range { region, pgsize },
+        }
+    }
+
+    /// Creates a new TLB flush scope, including only a single page (which may be
+    /// global).
+    pub fn page(va: VirtAddr, pgsize: PageSize) -> Self {
+        let region = MemoryRegion::new(va, usize::from(pgsize));
+        Self::range(region, pgsize)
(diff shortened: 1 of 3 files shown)


kernel/src/cpu/tlb.rs
```diff
@@ -43,6 +43,34 @@ pub struct TlbFlushScope {
...
+    pub const fn with_range(mut self, region: MemoryRegion<VirtAddr>, pgsize: PageSize) -> Self {
+        self.range = TlbFlushRange::Range { region, pgsize };
+        self
+    }
+
+    /// Updates the flush scope to only include a singe page.
+    pub fn with_page(self, va: VirtAddr, pgsize: PageSize) -> Self {
+        let region = MemoryRegion::new(va, usize::from(pgsize));
+        self.with_range(region, pgsize)
+    }
```
joergroedel: This interface is problematic, the chaining hints that ranges are additive when they are actually replacing. These should be separate constructors, which also avoids setting self.range twice for every creation.

kernel/src/cpu/tlb.rs
```diff
@@ -43,6 +43,34 @@ pub struct TlbFlushScope {
...
+            global: true,
+            range: TlbFlushRange::All,
+        }
+    }
+
+    /// Updates the scope to include or exclude global pages, as indicated.
+    pub const fn with_global(mut self, global: bool) -> Self {
+        self.global = global;
+        self
+    }
```
joergroedel, line 58: How about initializing `self.global` to `false` and make this method set it to `true`? This better matches the method name and saves the parameter.

joergroedel (commented):
Overall I am fine with the changes. The interface needs improvements, I left some comments.

kernel/src/cpu/tlb.rs
00xc: Global enabled seems to be the safer default, no?

00xc: I believe we discussed this in a call a few weeks ago, marking this as resolved. Having the possibility to explicitly set global to true or false seems a bit more flexible and readable, so I'll keep it this way unless there are more comments.

00xc:
Rebased on main

msft-jlange:
This change has me wondering a bit how TLB flush list structures should be managed in light of the sorts of changes that we should expect in the future.  There are two points in particular.

First, as COCONUT develops more paravisor support, I expect the `TlbFlushRange` structure to be used in more places than simply the existing TLB flush code.  In particular, there will be cases where the SVSM has to flush individual addresses of a lower privilege level, such as when implementing the Hyper-V address range flushes.  These operations, at least on TDX, require use of TDX module functions specific to TDX Partitioning and do not overlap the TLB flush ISA at all (contrast this with SNP where both same-VMPL and cross-VMPL flushes simply use INVLPGB).  So there will be value in making full use of the `TlbFlushRange` structure separately from the `TlbFlushScope` structure.

In addition, we will be adding a third type of addresses to the `TlbFlushRange` structure which is a list of addresses rather than simply a single contiguous range (it will mostly likely be a list of run-length-encoded ranges for ease of mapping both to the Hyper-V address range list structure and the TDX Partitioning TLB flush list structure, both of which use identical RLE encodings for address lists).  That means there will end up being an additional type of construction of `TlbFlushRange`, which I expect will be embodied as `TlbFlushRange::push(&mut self, vaddr: VirtAddr)` so that code that loops over virtual address ranges can push individual addresses based on dynamic conditions (such as whether PTEs in a range are all valid, or have been accessed).

Taken together, I wonder whether the future direction will be to separate `TlbFlushRange` and `TlbFlushScope` such that either `TlbFlushScope` includes a `&TlbFlushRange` instead of embedding an actual `TlbFlushRange` object, especially to faciliate the dyamic, iterative construction of a list of addresses.  If we do that, then methods like `TlbFlushScope::page()` and `TlbFlushScope::range()` become less useful, and becomes replaced with patterns that construct the range first and the scope later.  In fact, if the semantics of the scope are limited to the set of processors involved and/or the global state, and the semantics of the range are the list of addresses involved, then they are truly orthogonal and the preferred pattern may well be to have all of the TLB flush functions take both a `TlbFlushRange` and `TlbFlushScope` as separate parameters instead of trying to fuse them together.

All of this is somewhat hypothetical because the code doesn't exist yet, but if there's a strong potential for moving TLB flush in this direction - as I think there will be - I wonder whether we should be moving the code in the opposite direction in the short term, only to rewrite it again later.

00xc:
> In addition, we will be adding a third type of addresses to the TlbFlushRange structure which is a list of addresses rather than simply a single contiguous range (it will mostly likely be a list of run-length-encoded ranges for ease of mapping both to the Hyper-V address range list structure and the TDX Partitioning TLB flush list structure, both of which use identical RLE encodings for address lists). That means there will end up being an additional type of construction of TlbFlushRange, which I expect will be embodied as TlbFlushRange::push(&mut self, vaddr: VirtAddr) so that code that loops over virtual address ranges can push individual addresses based on dynamic conditions (such as whether PTEs in a range are all valid, or have been accessed).

That can be hidden behind new constructors/methods. I also assume this has to do with the IPI buffer concern you mentioned on my previous PR. The point here is that the internal details of `TlbFlushRange` can be changed while we keep the amount of users low, which PR does by keeping it all behind the public functions in `kernel/src/cpu/tlb.rs`.

> Taken together, I wonder whether the future direction will be to separate `TlbFlushRange` and `TlbFlushScope` such that either `TlbFlushScope` includes a `&TlbFlushRange` instead of embedding an actual `TlbFlushRange` object, especially to faciliate the dyamic, iterative construction of a list of addresses. If we do that, then methods like `TlbFlushScope::page()` and `TlbFlushScope::range()` become less useful, and becomes replaced with patterns that construct the range first and the scope later. In fact, if the semantics of the scope are limited to the set of processors involved and/or the global state, and the semantics of the range are the list of addresses involved, then they are truly orthogonal and the preferred pattern may well be to have all of the TLB flush functions take both a `TlbFlushRange` and `TlbFlushScope` as separate parameters instead of trying to fuse them together.

Sure, which is why for now all of this is hidden behind public functions and I did not inline the constructors into the callers. That way we can change them later if code evolves in that direction if necessary.

> All of this is somewhat hypothetical because the code doesn't exist yet, but if there's a strong potential for moving TLB flush in this direction - as I think there will be - I wonder whether we should be moving the code in the opposite direction in the short term, only to rewrite it again later.

The code can easily be pulled out. `TlbFlushScope` is a wrapper over a `TlbFlushRange` with an added boolean to distinguish whether global addresses should be flushed. And even then that could be embeddeded in `TlbFlushScope`, I just didn't do it in order to avoid repeating the field across enum variants.

msft-jlange (approved):

PR #1029, opened 2026-04-02
bldr: introduce a new boot loader

msft-jlange:
Since the kernel boot image is fully prepared in the IGVM file builder (or nearly so), the requirements for a boot loader have been dramatically reduced.  This PR introduces a new boot loader which performs the necessary finalization of the boot image based on runtime configuration when necessary (such as modification of page tables for the SNP C-bit).  Because this boot loader is so simple, it does not rely on any of the code built into the kernel, so it can eliminate the need to share complex kernel code across multiple binary components.

In the future, this boot loader can be expanded to perform additional tasks (such as randomization of the kernel address space) but even that additional logic can be built using either small, dedicated code or small, shared components rather than having to rely on the complexity of the kernel logic that initializes a full runtime environment to support boot loader functionality.

### DIFF
--- a/boot/bldr/src/main.rs
+++ b/boot/bldr/src/main.rs
@@ -1,0 +1,345 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) Microsoft Corporation
+//
+// Author: Jon Lange <jlange@microsoft.com>
+
+#![no_std]
+#![no_main]
+
+mod temp_maps;
+
+use crate::temp_maps::TempMappings;
+use bootdefs::kernel_launch::BldrLaunchInfo;
+use bootdefs::kernel_launch::KernelLaunchInfo;
+use bootdefs::platform::SvsmPlatformType;
+use core::alloc::GlobalAlloc;
+use core::alloc::Layout;
+use core::arch::global_asm;
+use core::mem::offset_of;
+use core::slice;
+use cpuarch::sev_status::MSR_SEV_STATUS;
+use cpuarch::sev_status::SEVStatusFlags;
+use cpuarch::x86::CR0Flags;
+use cpuarch::x86::EFERFlags;
+use cpuarch::x86::MSR_EFER;
+
+global_asm!(
+    r#"
+        .section .text
+        .section .startup.text,"ax"
+        .code32
+
+        .globl startup_32
+        startup_32:
+
+        /* Upon entry, ESI holds the high 32 bits of VTOM. */
+
+        /* Save pointer to startup structure in EBP */
+        movl %esp, %ebp
+
+        /* Save the platform type for future use.  The platform type is loaded
+         * in EAX upon entry. */
+        movl %eax, {PLATFORM}(%ebp)
+
+        /* Check to see whether this is an SNP system.  If not, no page table
+         * manipulation is required. */
+        cmpl $1, %eax
+        jnz page_tables_ready
+
+        /* Check to see whether vTOM is active on this system.  If so, no page
+         * table manipulation is necessary */
+        movl ${MSR_SEV_STATUS}, %ecx
+        rdmsr
+        testl ${VTOM}, %eax
+        jnz page_tables_ready
+
+        /* Locate the CPUID page */
+        movl {CPUID_PAGE}(%ebp), %edx
+
+        /* Determine the number of entries and the address of the first
+         * entry.  The contents of the CPUID page have been validated by
+         * the SNP firmware so the data can be trusted directly. */
+        movl (%edx), %ecx
+        leal 16(%edx), %edx
+
+    10:
+        /* Check to see whether this entry is the extended SEV leaf.  If not,
+         * keep searching */
+        cmpl $0x8000001F, (%edx)
+        jz 11f
+        leal 48(%edx), %edx
+        dec %ecx
+        jnz 10b
+        ud2
+
+    11:
+        /* The C-bit position is encoded as the low 6 bits of the EBX field. */
+        mov 28(%edx), %eax
+        andl $0x3F, %eax
+
+        /* Save the C-bit position for later use. */
+        movl %eax, {C_BIT_POS}(%ebp)
+
+        /* Obtain the bounds of the page tables. */
+        movl {PT_START}(%ebp), %ecx
+        movl {PT_END}(%ebp), %edx
+
+        /* Insert the C-bit into each valid PTE. */
+    1:
+        testl $1, (%ecx)
+        jz 2f
+        bts %eax, (%ecx)
+    2:
+        addl $8, %ecx
+        cmp %edx, %ecx
+        jb 1b
+
+    page_tables_ready:
+        /* The page tables are now configured, so enable long mode. */
+        movl ${MSR_EFER}, %ecx
+        rdmsr
+        movl %eax, %ebx
+
+        /* Include both EFER.LME and EFER.NXE since the page tables will use
+         * the NX bit. */
+        orl $({EFER_LME} | {EFER_NXE}), %eax
+
+        /* Don't write the MSR if it is already correct. */
+        cmpl %eax, %ebx
+        jz 3f
+        wrmsr
+    3:
+        /* Enable paging so that long mode can be activated. */
+        movl {PT_ROOT}(%ebp), %edx
+        movl %edx, %cr3
+        movl %cr0, %eax
+        bts ${PG_SHIFT}, %eax
+        movl %eax, %cr0
+
+        /* Establish the correct GDT and jump to the 64-bit entry point. */
+        movl $gdt64_desc, %eax
+        lgdt (%eax)
+        ljmpl $0x8, $startup_64
+
+        .code64
+
+    startup_64:
+        /* Reload the data segments with 64bit descriptors. */
+        movw $0x10, %ax
+        movw %ax, %ds
+        movw %ax, %es
+        movw %ax, %fs
+        movw %ax, %gs
+        movw %ax, %ss
+
+        /*
+         * Follow the C calling convention for x86-64:
+         *
+         * - Pass &BldrLaunchInfo as the first argument (%rdi)
+         * - Pass VTOM as the second argument (%rsi)
+         * - Make sure (%rsp + 8) is 16b-aligned when control is transferred
+         *   to stage2_main
+         */
+        movl %ebp, %edi
+        shlq $32, %rsi
+        andq $~0xf, %rsp
+
+        /* Mark the next stack frame as the bottom frame */
+        xor %rbp, %rbp
+
+        call bldr_main
+
+        .globl switch_to_kernel
+    switch_to_kernel:
+        /* Switch to the kernel stack. */
+        movq %rsi, %rsp
+
+        /* Load the platform type into rax as expected by the kernel */
+        movq %rdx, %rax
+
+        /* Enter the kernel. */
+        push %rdi
+        ret
+
+        .globl pvalidate_one
+    pvalidate_one:
+        movq %rdi, %rax
+        xorl %ecx, %ecx
+        movl $1, %edx
+        pvalidate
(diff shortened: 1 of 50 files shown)


tools/igvmbuilder/src/initial_stack.rs
```diff
@@ -44,26 +44,70 @@ impl Stage2Stack {
...
+            kernel_entry: boot_image_info.context.entry_point,
+            kernel_stack: boot_image_info.context.initial_stack,
+            kernel_launch_info: boot_image_info.kernel_launch_info,
+            kernel_pt_vaddr: boot_image_info.kernel_page_tables_virt_base,
+            kernel_pt_count: boot_image_info.total_pt_pages,
+            kernel_pdpt_paddr: boot_image_info.kernel_pdpt_paddr,
+            kernel_pml4e_index: boot_image_info.kernel_pml4e_index,
+            page_table_start: gpa_map.init_page_tables.get_start() as u32,
+            page_table_end: gpa_map.init_page_tables.get_end() as u32,
+            page_table_root: gpa_map.init_page_tables.get_start() as u32,
```
peterfang: `page_table_start` and `page_table_root` appear to be the same (and therefore `PT_START` and `PT_ROOT` in bldr)?

boot/bldr/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+use core::mem::offset_of;
+use core::slice;
+use cpuarch::sev_status::MSR_SEV_STATUS;
+use cpuarch::sev_status::SEVStatusFlags;
+use cpuarch::x86::EFERFlags;
+use cpuarch::x86::MSR_EFER;
+
+global_asm!(
+    r#"
+        .text
```
peterfang: We discussed using `.section .text` instead [here](https://github.com/coconut-svsm/svsm/pull/978#discussion_r2857015973) so might as well continue doing it

boot/bldr/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+
+        /* Don't write the MSR if EFER_LME is already set.  This is required
+         * for certain versions of the TDX module. */
+        testl ${EFER_LME}, %eax
+        jnz 3f
+
+        /* Include NXE as well. */
+        orl $({EFER_LME} | {EFER_NXE}), %eax
+        wrmsr
+    3:
```
peterfang: To add more clarity, we can add a comment saying EFER.NXE is always expected to be set on TDX platforms. In boot_stage2, we test EFER for `EFER_LME | EFER_NXE`, and that's also an option for here

boot/bldr/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+        /* Enable paging so that long mode can be activated. */
+        movl {PT_ROOT}(%ebp), %edx
+        movl %edx, %cr3
+        movl %cr0, %eax
+        bts $31, %eax
```
peterfang: You can use `cpuarch::x86::CR0Flags::PG.bits()` here

boot/bldr/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+        xorl %ecx, %ecx
+        movl $1, %edx
+        pvalidate
+        jb 1f
+        ret
+    1:
+        ud2
+
+        .data
+        .align 16
```
peterfang, line 177: Nit: since there are no system descriptors in this GDT, `.align 8` should work just as well

boot/bldr/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+    gdt64:
+        .quad 0
+        .quad 0x00af9a000000ffff /* 64 bit code segment */
+        .quad 0x00cf92000000ffff /* 64 bit data segment */
+    gdt64_end:
+
+    gdt64_desc:
+        .word gdt64_end - gdt64 - 1
+        .quad gdt64
+
```
peterfang: Nit: extra empty line

joergroedel (commented):
Booting on native works.
Loading on SNP fails with an error from SNP_LAUNCH_UPDATE, probably a problem with the IGVM file:
```
qemu-system-x86_64: SNP_LAUNCH_UPDATE ret=-5 fw_error=0 ''
```

msft-jlange:
> Booting on native works. Loading on SNP fails with an error from SNP_LAUNCH_UPDATE, probably a problem with the IGVM file:
> 
> ```
> qemu-system-x86_64: SNP_LAUNCH_UPDATE ret=-5 fw_error=0 ''
> ```

@joergroedel I'm going to need a little bit of help from you to debug this.  I don't see any issues on Hyper-V, and there's not enough information here for me to have any clue what's going on.  It would be nice if the error message included a GPA so we could know which page has an issue, and the "fw_error=0" suggests to me that the PSP command didn't actually return an error code so it's as if something is going wrong within QEMU itself.  Perhaps you can get better diagnostic information about what actually failed.

tools/igvmbuilder/src/initial_stack.rs
msft-jlange: This is how we happen to lay out the page tables today (where the paging root is the first of the page table pages) but I don't want to require that as part of the architecture.  We can maintain more independence across the different components if the boot loader makes no assumptions about where the paging root sits within the set of page table pages.

boot/bldr/src/main.rs
msft-jlange: It would, but if we ever decide to add a descriptor, we will probably want 16-byte alignment.  At this point, I think 8 bytes of padding space is really no concern, so I'd rather keep it for general consistency and future-proofing unless we have a good reason to change it.

boot/bldr/src/main.rs
msft-jlange: Not sure I follow.  `PG.bits()` evaluates to `0x8000_0000` but what I need is `31`.  The use of 31 as a constant is present in the current boot_stage2 so I'm just maintaining the existing pattern.

peterfang: `cpuarch::x86::CR0Flags::PG.bits().trailing_zeros()` should work. Since you already did the work of moving more things into cpuarch, and EFER_* and CR4_PAE are now being used, this could help with code readability. Not a requirement if you feel strongly against it

msft-jlange: That does work, and I agree that it makes the intention much more clear.

boot/bldr/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+    2:
+        addl $8, %ecx
+        cmp %edx, %ecx
+        jb 1b
+
+    page_tables_ready:
+        /* The page tables are now configured, so enable long mode. */
+        movl ${MSR_EFER}, %ecx
+        rdmsr
+        movl %eax, %esi
```
joergroedel: This corrupts `%ESI` which carries `VTOM`, no?

joergroedel:
This still fails with the SNP_LAUNCH_UPDATE error:

```
char device redirected to /dev/pts/4 (label serial1)
qemu-system-x86_64: SNP_LAUNCH_UPDATE ret=-5 fw_error=0 ''
```

Tracing the high-level ioctls shows it fails here:

```
 qemu-system-x86-17818 [013] .....  1434.925821: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000298 type: 4 len: 0 ret: 0
 qemu-system-x86-17818 [013] .....  1434.926753: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000297 type: 4 len: 0 ret: 0
 qemu-system-x86-17818 [013] .....  1434.927841: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000296 type: 4 len: 0 ret: 0
 qemu-system-x86-17818 [013] .....  1434.929760: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x12 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [013] .....  1434.935759: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x806 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [013] .....  1434.936824: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x807 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [013] .....  1434.938808: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x808 type: 6 len: 0 ret: 0
 qemu-system-x86-17818 [013] .....  1434.952189: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x811 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.008046: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x100000 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.743214: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000277 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.760908: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000286 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.766766: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x800028b type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.774854: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000292 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.775940: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000293 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.777094: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000295 type: 1 len: 0 ret: 0
 qemu-system-x86-17818 [008] .....  1436.777101: bprint:               sev_mem_enc_ioctl: SNP_LAUNCH_UPDATE: gfn: 0x8000294 type: 1 len: 4096 ret: -5
```

Hope that helps in debugging this, I can do more tracing if needed.

joergroedel:
I believe the Linux kernels `guest_memfd` returns an error when a page is deployed twice, might this be the issue here?

boot/bldr/src/main.rs
msft-jlange: You are correct.  This is fixed now.

msft-jlange:
> I believe the Linux kernels `guest_memfd` returns an error when a page is deployed twice, might this be the issue here?

In the trace you pasted here, there is no page number listed twice, so that can't be the issue.  The `len: 4096` seems suspicious because every other page has `len: 0`.  However, the QEMU IGVM that I build looks nothing like the one you've pasted here.  I can't fully replicate what you have because I don't have an OVMF FW image to embed in the IGVM, but my page list looks totally different.  For example, I see a set of zero-filled pages from 0x800 through 0x805 as well as data-filled pages from 0x809-0x810.  I don't have a page at 4 GB, and the pages I see above 512 GB start at 0x8000000 and continue sequentially.  So I think we're going to have to dig into your IGVM generation to figure out what is going into it,.

joergroedel (approved):
Thanks for the update, testing now passes on my side.

peterfang:
The latest revision appears to have many additional changes, including a new file `temp_maps.rs`. Does this need some more review?

msft-jlange:
> The latest revision appears to have many additional changes, including a new file `temp_maps.rs`. Does this need some more review?

@peterfang that's up to you.  It couldn't hurt.

PR #1030, opened 2026-04-02
kernel: centralize CPUID feature discovery

00xc:
The SVSM uses CPUID in multiple places to discover certain features about the environment it runs in (like CET support, FPU features or hypervisor enumeration). The use of CPUID is not consistent, as sometimes the actual instruction is used, but other times the platform abstraction is used.

This PR centralizes the use of CPUID in the following manner:
1. A new universal `cpuid()` function is added, using the `SvsmPlatform` as a backend.
2. Users of CPUID are converted to the new `cpuid()` function. On SEV-SNP in particular this reduces the amount of #VC exceptions that need to be handled.
3. Centralize all CPUID discovery into a lookup table. Items in the lookup table are cached, so CPUID usage is drastically reduced in the long run. This makes uses of CPUID much cleaner, and removes the need for manual, ad-hoc caching for the values queried from CPUID.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -8,7 +8,7 @@ use super::idt::common::X86ExceptionContext;
 use crate::address::Address;
 use crate::address::VirtAddr;
 use crate::cpu::X86GeneralRegs;
-use crate::cpu::cpuid::{CpuidLeaf, cpuid_table_raw};
+use crate::cpu::cpuid::cpuid_table_raw;
 use crate::cpu::percpu::current_ghcb;
 use crate::cpu::percpu::this_cpu;
 use crate::debug::gdbstub::svsm_gdbstub::handle_debug_exception;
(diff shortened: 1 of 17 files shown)


kernel/src/cpu/vc.rs
```diff
@@ -248,26 +248,22 @@ fn handle_cpuid(ctx: &mut X86ExceptionContext) -> Result<(), SvsmError> {
...
 fn snp_cpuid(ctx: &mut X86ExceptionContext) -> Result<(), SvsmError> {
-    let mut leaf = CpuidLeaf::new(ctx.regs.rax as u32, ctx.regs.rcx as u32);
-    let xcr0_in = if leaf.cpuid_fn == 0xD && (leaf.cpuid_subfn == 1 || leaf.cpuid_subfn == 0) {
+    let cpuid_fn = ctx.regs.rax as u32;
+    let cpuid_subfn = ctx.regs.rcx as u32;
+    let xcr0_in = if cpuid_fn == 0xD && (cpuid_subfn == 1 || cpuid_subfn == 0) {
         1
     } else {
         0
     };
```
joergroedel, line 257: Not introduced by this PR, but I think this is a bug. The real `XCR0` value should be used here. If the given value is not in the CPUID table, SVSM should forward the request to the hypervisor. Otherwise the is the risk of allocating a too small XSAVE area and data corruption on each task switch.

joergroedel: Created issue #1057 to track this problem.

joergroedel (approved):
Thanks, changes look good to me. Testing also passed.

PR #1032, opened 2026-04-03
github/workflows: Unify signedOff and spdx checks

n-ramacciotti:
The `signed-off` and `spdx` workflows perform related actions and have the same setup.

Combining them in a single workflow reduces the number of redundant actions and optimizes the CI.

~~Is it better to keep the checks separated to make it easier to see if there is a problem in both of them? With these changes, if the first check fails the second one will be skipped.~~

Introduce ~~'continue-on-error'~~ `!cancelled`  to test both 'signed-off' and 'spdx' even if the first of the two steps fails


.github/workflows/pr-compliance.yml
```diff
@@ -20,9 +20,20 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
-
-      - name: Check commits
+
```
luigix25: trailing whitespace

n-ramacciotti: Yes, thanks!

n-ramacciotti:
v3:
- Used `!cancelled` instead of `continue-on-error`:
  - With this option, the step `spdx` (the second step) will always run unless it is cancelled. This happens even in case that `signed-off` (the first of the two steps) fails. 
  - This will also highlight better when both steps fail, as there will be a red `x` on both of them. With `continue-on-error` only one of the two would have the red sign. 
  - There is no need for an additional step in the workflow.

.github/workflows/pr-compliance.yml
```diff
@@ -0,0 +1,43 @@
...
+          git fetch
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+
+      - name: Signed-off-by
+        run: |
+          ${GITHUB_WORKSPACE}/scripts/check-signed-off.sh \
+            ${{ github.event.pull_request.base.sha }} \
+            ${{ github.event.pull_request.head.sha }}
+
+      - name : SPDX Headers
```
stefano-garzarella: nit:

```suggestion
      - name: Check SPDX Headers
```

.github/workflows/pr-compliance.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Fetch base branch
+        run: |
+          git fetch
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+
+      - name: Signed-off-by
```
stefano-garzarella: nit:

```suggestion
      - name: Check Signed-off-by
```

stefano-garzarella (changes requested):
Please squash the second commit into the first, I don't see any good reason to keep them separated. The rest LGTM, I left very nit comments, feel free to ignore them.

stefano-garzarella (approved):

PR #1033, opened 2026-04-05
kernel: cpu/cpuid: traverse CPUID page using an iterator

00xc:
The amount of items in the CPUID page is indicated by the count field, so use it to take that amount of items from an iterator and simplify cpuid_table_raw() and dump_cpuid_table(). No functional change.

### DIFF
--- a/kernel/src/cpu/cpuid.rs
+++ b/kernel/src/cpu/cpuid.rs
@@ -36,7 +36,7 @@ pub unsafe fn init_cpuid_table(addr: VirtAddr) {
             .expect("Misaligned SNP CPUID table address")
     };
 
-    for func in table.func.iter_mut().take(table.count as usize) {
+    for func in table.entries_mut() {
         if func.eax_in == 0x8000001f {
             func.eax_out |= 1 << 28;
         }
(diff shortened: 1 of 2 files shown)


kernel/src/cpu/cpuid.rs
```diff
@@ -131,24 +131,17 @@ impl CpuidResult {
...
-                edx: CPUID_PAGE.func[i].edx_out,
-            });
-        }
-    }
-
-    None
+    CPUID_PAGE
+        .func
+        .iter()
+        .take(CPUID_PAGE.count as usize)
```
joergroedel: Instead of enforcing array bounds with `take()`, please create a slice first and iterate over that. The slice creation will make the `unsafe` nature of this operation explicit and the use more safe.

00xc: Not sure I follow, what is the unsafe part in this operation? It seems more natural to create an iterator and limit it by the amount of populated entries.

joergroedel: The unsafe part is that the code needs to trust the `count` value to be memory-safe.

00xc: Precisely, even if `count` is invalid, using it with `take()` won't cause any issues. Creating a slice with an invalid `count` OTOH may panic.

joergroedel: A panic is not a memory-safety violation, silently using a wrong `count` is, no? Also, always having to use `take()` when iterating over the CPUID page is error-prone.

00xc: No, the CPUID page is zero-initialized, so this cannot result in any memory safety issue. `take()` just truncates an existing iterator to yield at most X items.

00xc: IOW, [`take()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take) cannot go out of bounds (otherwise it would need to be `unsafe`). If the iterator is shorter than the amount of specified items it simply shortcuts early.

joergroedel (approved):

PR #1034, opened 2026-04-07
libtcgtpm: remove nested archive libtcgtpm.a

kraxel:
The rust linker can not handle nested archives, i.e. archives containing
other archives instead of object files only.  The libtcgtpm.a archive is
such a nested archive, lets remove it.

This is one step toward being able to use the rust linker to build svsm.

### DIFF
--- a/libtcgtpm/Makefile
+++ b/libtcgtpm/Makefile
@@ -56,8 +56,6 @@ LIBBNMATH = $(TCGTPM_BUILD_DIR)/cryptolib_Ossl/lib$(LIBBNMATH_TARGET).a
 OPENSSL_MAKEFILE = $(OPENSSL_BUILD_DIR)/Makefile
 TCGTPM_MAKEFILE = $(TCGTPM_BUILD_DIR)/Makefile
 
-LIBS = $(LIBCRT) $(LIBCRYPTO) $(LIBBNMATH) $(LIBTPMBIGNUM) $(LIBTPM) $(LIBPLATFORM)
-
 ifeq ($(USE_LIBCRT),1)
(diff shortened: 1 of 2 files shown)


osteffenrh (approved):

stefano-garzarella:
cool, thanks for this. I'm curious if this could also be related to what we were seeing in #998 (CC @00xc )

libtcgtpm/Makefile
stefano-garzarella: I think we can remove this line now.

stefano-garzarella: Ditto, and maybe even the `tcgtpm` in .PHONY.

Or we should leave `$(OUT_DIR)/libtcgtpm.a` for some other user case (not sure really).

stefano-garzarella (commented):
LGTM, just some minor cleanup.

Just one question, what about squashing all of them in a single commit?
I don't see a real gain of having these changes separated, but maybe I'm missing something.

stefano-garzarella:
Just a note, I think there is a little conflict with #998, so maybe this PR needs to be rebased after merging it, but should be an easy conflict to resolve.

kraxel:
> Just one question, what about squashing all of them in a single commit? I don't see a real gain of having these changes separated, but maybe I'm missing something.

I've moved them out one by one and committed each working version separately, mostly out of the habit to prefer many small commits.  I can squash them all together, there is no technical reason to keep them separate.

stefano-garzarella:
@kraxel I just merged #998, can you rebase and fix the conflict?

stefano-garzarella (dismissed):

joergroedel (approved):
Testing succeeded.

stefano-garzarella (approved):

Issue #827, opened 2025-09-29
Archive EDK2 fork


stefano-garzarella:
_Originally posted by @kraxel in https://github.com/coconut-svsm/svsm/issues/824#issuecomment-3347369063_

> My PR was just merged into master.
> 
> I'm wondering whenever the coconut edk2 fork should be archived to avoid confusion?
> I think there is little reason to /not/ work against upstream tianocore repo, or do we have important svsm changes which are not merged upstream?
> @stefano-garzarella

stefano-garzarella:
@kraxel I'm moving the discussion here:

> My PR was just merged into master.

Great!

> 
> I'm wondering whenever the coconut edk2 fork should be archived to avoid confusion? I think there is little reason to /not/ work against upstream tianocore repo, or do we have important svsm changes which are not merged upstream? @stefano-garzarella

I think so. @osteffenrh do you remember if we have anything else?

I'll bring this to the TSC meeting tomorrow and if no one will complain, I'll update our documentation (to point to upstream commit) and archive our fork like we did for https://github.com/coconut-svsm/ms-tpm-20-ref.

Not sure if we want to keep it for future development (e.g. `reboot` work, but in that case maybe it's just Linux to be modified). In that case we should just rebase `svsm` on current upstream master.

luigix25:
FYI @RARelph patches are being reviewed in upstream edk2 (https://github.com/tianocore/edk2/pull/11536)

[OvmfPkg/X64: add opt/org.tianocore/UninstallMemAttrProtocol support](https://github.com/tianocore/edk2/commit/2abf30d65811118f5a17f292aeb5b6e12c29a808) is not really needed to boot a CVM.

So the only extra commit we have is [OvmfPkg/AmdSevDxe: Shim Reboot workaround](https://github.com/tianocore/edk2/commit/f9a36abb6f69852944c04e210127bddd778ad6f2), not really sure about this, though.

stefano-garzarella:
@kraxel IIRC we don't need our fork anymore, do you think we can archive it?

kraxel:
I think that should be ok now with the grub issues fixed since a year so there isn't any non-upstream patch we need left.

stefano-garzarella:
@kraxel thanks for checking, I archived it!

PR #1035, opened 2026-04-07
docs/INSTALL: update guest image requirement

### LINKED ISSUE #712: vtpm: update guest image requirements
The SVSM TPM driver is now merged upstream in the Linus' tree.
It will be part of `Linux v6.16-rc1`, so we should update our documentation and also https://github.com/coconut-svsm/linux

- [ ] Update https://github.com/coconut-svsm/svsm/blob/main/Documentation/docs/installation/INSTALL.md#preparing-the-guest-image
  - [ ] Upstream kernel should now contain all we need for the guest (from `Linux v6.16-rc1`)
  - [ ] `CONFIG_TCG_SVSM` is the new config option
- [ ] Backport the driver in https://github.com/coconut-svsm/linux or rebase it to `Linux v6.16`

stefano-garzarella:
The SVSM TPM driver has been merged upstream and released with Linux v6.16. Update the guest image documentation to reflect that a custom kernel is no longer needed and reference the new CONFIG_TCG_SVSM option.

Closes: #712

### DIFF
--- a/Documentation/docs/installation/INSTALL.md
+++ b/Documentation/docs/installation/INSTALL.md
@@ -5,7 +5,6 @@ Installation of the COCONUT-SVSM requires some components that are not
 upstream in their respective repositories yet:
 
 * Linux host kernel with SVSM support
-* Linux guest kernel with SVSM support
 * EDK2 with SVSM support
 * A modified QEMU which supports launching guests configured using IGVM
 * The SVSM source-code repository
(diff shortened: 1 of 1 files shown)


osteffenrh (approved):

Documentation/docs/installation/INSTALL.md
```diff
@@ -5,7 +5,6 @@ Installation of the COCONUT-SVSM requires some components that are not
 upstream in their respective repositories yet:
 
 * Linux host kernel with SVSM support
-* Linux guest kernel with SVSM support
 * EDK2 with SVSM support
```
luigix25, line 8: while we are here, I think this can be dropped as well. We can use vanilla edk2, all of our downstream patches have been merged

stefano-garzarella: okay, but it isn't a guest image requirement, it's host requirement so better in another PR.
Should we also remove the entire `Building the guest firmware` section or move it in some other file in Documentation/docs/developer/?

luigix25: good point, i guess every distribution ships an ovmf firmware binary. I _think_ we can drop it, but not a strong opinion.

stefano-garzarella: From https://github.com/coconut-svsm/svsm/issues/827 it seems we still want to keep the fork around, so IMO we should also keep the instructions around (at least for developers).
So I agree we can remove instructions, etc. from INSTALL.md, but IMO we should keep it somewhere else till we will decide to archive our edk2 fork.

joergroedel (approved):

PR #1002, opened 2026-03-15
attest: add attestation testing with kbs-test integration

nihalxkumar:
<details>
<summary>This PR adds tests for both attestation surfaces.</summary>

SVSM exposes two distinct attestation surfaces: the pre-boot attestation feature (the attest Rust feature flag), which runs before the guest OS starts and uses a KBS to retrieve secrets, and the in-guest attestation
protocol, through which a VMPL0 guest can request a hardware attestation
report via SVSM.

</details>

For the in-guest attestation protocol the wire format of AttestServicesOp and AttestSingleServiceOp is fixed by the SVSM spec (Tables 11 and 13 of "Secure VM Service Module for SEV-SNP Guests, Rev 1.00"). A silent field reorder or padding change would break the binary interface with the guest, so the layout is now enforced with compile-time `const` assertions, following the pattern established in `greq/msg.rs`. Runtime tests cover reserved-field rejection, certificate region boundary enforcement, GUID table serialization, and that unknown request codes return `UNSUPPORTED_PROTOCOL`.

For pre-boot attestation, the `hash()` function feeds `NegotiationParams` into SHA-512 in the order dictated by the
server's `NegotiationResponse`. Because the server controls which params contribute and in what sequence, ordering sensitivity is a security property.
 Tests verify that `Challenge → EcPublicKeyBytes` and `EcPublicKeyBytes → Challenge` produce the expected individual digests and that swapping the order produces a distinct output.

The third commit adds a test script that exercises the pre-boot KBS flow end-to-end: it starts `kbs-test` and
`aproxy`, launches the SVSM test image built with the attest feature, and checks that "attestation successful" appears in the output. This requires a real SNP-capable machine and is not wired into the standard QEMU CI
workflow for that reason.

Related: https://github.com/coconut-svsm/svsm/issues/773

PR #1036, opened 2026-04-07
github/workflows: drop explicit checkout ref in PR workflows

stefano-garzarella:
The SPDX and SignedOff workflows override the checkout ref to `github.event.pull_request.head.sha`. This breaks PRs whose branches were created before the scripts used by the pipelines were added to main. For example the CI of PR #1002 is failing to execute the SPDX pipeline [1] with the following:

    scripts/check-spdx.sh: No such file or directory
    Process completed with exit code 127.

For `pull_request` events, GitHub runs workflows from a merge commit of the PR head into the target branch [2]:

    "GITHUB_SHA for this event is the last merge commit on the
     pull request merge branch."

So the workflow file from main is picked up and executed, but the checkout then switches to the PR head, where the scripts do not exist.

Drop the `ref:` override to let actions/checkout use its default, which checks out the merge commit. Both the scripts (from main) and the PR changes are available there. The SPDX script reads files from the working tree and the SignedOff script only inspects git objects, so both work correctly with the merge commit.

[1] https://github.com/coconut-svsm/svsm/actions/runs/23894438075/job/69675546900
[2] https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

### DIFF
--- a/.github/workflows/pr-compliance.yml
+++ b/.github/workflows/pr-compliance.yml
@@ -18,7 +18,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
 
       - name: Fetch base branch


00xc (approved):
Can't really remember if there was a good reason to add this in the first place, so LGTM

luigix25 (approved):

joergroedel (approved):

PR #1038, opened 2026-04-15
[verification] Support Verus's verita test via cargo-verus.

ziqiaozhou:
Update to latest verus to use cargo-verus.
Enable including SVSM in Verus Verita tests through cargo-verus. Helps catch Verus breaking changes that impact SVSM verification.

Changes due to Verus updates:
Use new style of specification for TryFrom in updated Verus. Change referece comparison to value comparison due to changed mut ref feature in Verus.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -14,5 +14,5 @@ rustflags = [
 ]
 
 [alias]
-verify = "v build --features verus"
+verify = "verus focus --features verus"
 xbuild = "run --package xbuild --release --"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,10 +99,10 @@ zerocopy = { version = "0.8.2", features = ["derive"] }
 embedded-io = { version = "0.6.1" }
 
 # Verus repos
-verus_builtin = { version = "=0.0.0-2025-12-07-0054", default-features = false }
-verus_builtin_macros = { version = "=0.0.0-2025-12-07-0054", features = ["vpanic"], default-features = false }
-verus_state_machines_macros = { version = "=0.0.0-2025-11-23-0053", default-features = false }
-vstd = { version = "0.0.0-2025-12-07-0054", features = ["alloc", "allow_panic"], default-features = false }
+verus_builtin = { version = "=0.0.0-2026-04-12-0118", default-features = false }
+verus_builtin_macros = { version = "=0.0.0-2026-04-12-0118", features = ["vpanic"], default-features = false }
+verus_state_machines_macros = { version = "=0.0.0-2026-04-05-0114", default-features = false }
+vstd = { version = "=0.0.0-2026-04-12-0118", features = ["alloc", "allow_panic"], default-features = false }
 
 # Verification libs
 verify_proof = { path = "verification/verify_proof", default-features = false  }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,10 +101,10 @@ zerocopy = { version = "0.8.2", features = ["derive"] }
 embedded-io = { version = "0.6.1" }
 
 # Verus repos
-verus_builtin = { version = "=0.0.0-2025-12-07-0054", default-features = false }
-verus_builtin_macros = { version = "=0.0.0-2025-12-07-0054", features = ["vpanic"], default-features = false }
-verus_state_machines_macros = { version = "=0.0.0-2025-11-23-0053", default-features = false }
-vstd = { version = "0.0.0-2025-12-07-0054", features = ["alloc", "allow_panic"], default-features = false }
+verus_builtin = { version = "=0.0.0-2026-04-12-0118", default-features = false }
+verus_builtin_macros = { version = "=0.0.0-2026-04-12-0118", features = ["vpanic"], default-features = false }
+verus_state_machines_macros = { version = "=0.0.0-2026-04-05-0114", default-features = false }
+vstd = { version = "=0.0.0-2026-04-12-0118", features = ["alloc", "allow_panic"], default-features = false }
 
 # Verification libs
 verify_proof = { path = "verification/verify_proof", default-features = false  }
--- a/scripts/vinstall.sh
+++ b/scripts/vinstall.sh
@@ -5,14 +5,55 @@
 #
 # Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
 # A script to install verus tools
-VERISMO_REV=1840262abbcf32f0bd5e622e47fb18d7dfaca795
+set -e
+trap 'echo "Error at line $LINENO: $BASH_COMMAND"' ERR
+VERUS_RELEASE=0.2026.04.12.f1166c4 # May need to update if vstd or other verus library changes
 VERUS_RUST_VERSION=1.94.0
 VERUSFMT_REV=beff2fa686d856d5e60df368fd027d94ead11ac5 # v0.5.7
 
 # Install x86_64-unknown-none target for verus-compatible Rust version
 export RUSTUP_TOOLCHAIN=$VERUS_RUST_VERSION
 rustup target add x86_64-unknown-none --toolchain $RUSTUP_TOOLCHAIN
-# Install the verus toolchain
-cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV cargo-v
-cargo v install-verus
+
+# Install verusfmt
 cargo install --git https://github.com/verus-lang/verusfmt  --rev $VERUSFMT_REV
+
+# Install verus toolchain
+# Verus cannot be installed via cargo and its build is slow, so we download the prebuilt binaries
+VERUS_ASSETS=(
+    "verus"
+    "rust_verify"
+    "z3"
+    "cargo-verus"
+    "verus-root"
+)
+
+# Detect platform
+ARCH=$(uname -m)
+OS=$(uname -s)
+case "$ARCH" in
+    x86_64)        PLATFORM_ARCH="x86" ;;
+    aarch64|arm64) PLATFORM_ARCH="arm64" ;;
+    *) echo "Error: unsupported architecture: $ARCH" >&2; exit 1 ;;
+esac
+case "$OS" in
+    Linux)  PLATFORM_OS="linux" ;;
+    Darwin) PLATFORM_OS="macos" ;;
+    *) echo "Error: unsupported OS: $OS" >&2; exit 1 ;;
+esac
+
+PLATFORM="${PLATFORM_ARCH}-${PLATFORM_OS}"
+ZIPFILE="verus-${VERUS_RELEASE}-${PLATFORM}.zip"
+DOWNLOAD_URL="https://github.com/verus-lang/verus/releases/download/release/${VERUS_RELEASE}/${ZIPFILE}"
+TMPDIR=$(mktemp -d)
+
+# Download Verus prebuilt into a tmp folder
+curl -sfL "$DOWNLOAD_URL" -o "$TMPDIR/$ZIPFILE"
+unzip -q "$TMPDIR/$ZIPFILE" -d "$TMPDIR"
+
+# Move the extracted Verus assets to the final directory
+for asset in "${VERUS_ASSETS[@]}"; do
+    echo "Installing $asset to ~/.cargo/bin/"
+    mv "$TMPDIR/verus-$PLATFORM/$asset" ~/.cargo/bin/
+done
+rm -rf "$TMPDIR"
--- a/scripts/vinstall.sh
+++ b/scripts/vinstall.sh
@@ -5,8 +5,8 @@
 #
 # Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
 # A script to install verus tools
-VERISMO_REV=4f504b72ddd7a6d5b194b65db159e55a41831ab9
-VERUS_RUST_VERSION=1.91.0
+VERISMO_REV=1840262abbcf32f0bd5e622e47fb18d7dfaca795
+VERUS_RUST_VERSION=1.94.0
 VERUSFMT_REV=beff2fa686d856d5e60df368fd027d94ead11ac5 # v0.5.7
 
 # Install x86_64-unknown-none target for verus-compatible Rust version
@@ -14,8 +14,5 @@ export RUSTUP_TOOLCHAIN=$VERUS_RUST_VERSION
 rustup target add x86_64-unknown-none --toolchain $RUSTUP_TOOLCHAIN
 # Install the verus toolchain
 cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV cargo-v
-# verus-rustc as a wrapper to call verus with proper rustc flags.
-cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV verus-rustc
 cargo v install-verus
-
 cargo install --git https://github.com/verus-lang/verusfmt  --rev $VERUSFMT_REV
--- a/scripts/vinstall.sh
+++ b/scripts/vinstall.sh
@@ -5,14 +5,136 @@
 #
 # Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
 # A script to install verus tools
-VERISMO_REV=1840262abbcf32f0bd5e622e47fb18d7dfaca795
+set -e
+trap 'echo "Error at line $LINENO: $BASH_COMMAND"' ERR
+
+# Verus release version and commit hash
+VERUS_VERSION=0.2026.04.12.f1166c4
+VERUS_REV=f1166c42c3decd42c1cca2916ef2880d27cfb7d9
 VERUS_RUST_VERSION=1.94.0
+
+# Verusfmt version and commit hash
+VERUSFMT_VERSION=v0.5.7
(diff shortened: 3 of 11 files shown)


.cargo/config.toml
```diff
@@ -14,5 +14,5 @@ rustflags = [
 ]
 
 [alias]
-verify = "v build --features verus"
+verify = "verus focus --features verus"
```
stefano-garzarella, line 17: in `.github/workflows/manual-verify.yml` we added a workaround since there is an issue in cargo-v.
See https://github.com/coconut-svsm/svsm/pull/964

Now that we are not using cargo-v anymore, should we revert that commit  (or extend the check).

stefano-garzarella, line 17: I think we need to update `Documentation/docs/developer/VERIFICATION.md` where we have references to `cargo-v`

Cargo.toml
```diff
@@ -98,10 +98,10 @@ uuid = { version = "1.6.1", default-features = false }
 zerocopy = { version = "0.8.2", features = ["derive"] }
 
 # Verus repos
-verus_builtin = { version = "=0.0.0-2025-12-07-0054", default-features = false }
```
stefano-garzarella, line 104: Why dropping the `=` in the version?
Is it okay for us to start using potential new versions now?

scripts/vinstall.sh
```diff
@@ -5,17 +5,14 @@
...
-VERUS_RUST_VERSION=1.91.0
+VERISMO_REV=1840262abbcf32f0bd5e622e47fb18d7dfaca795
+VERUS_RUST_VERSION=1.94.0
 VERUSFMT_REV=beff2fa686d856d5e60df368fd027d94ead11ac5 # v0.5.7
 
 # Install x86_64-unknown-none target for verus-compatible Rust version
 export RUSTUP_TOOLCHAIN=$VERUS_RUST_VERSION
 rustup target add x86_64-unknown-none --toolchain $RUSTUP_TOOLCHAIN
 # Install the verus toolchain
 cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV cargo-v
```
stefano-garzarella: Do we still need to install `cargo-v`?

stefano-garzarella (changes requested):

scripts/vinstall.sh
ziqiaozhou: We do not need cargo-v for verification anymore. cargo-v only downloads corresponding verus toolchain based on verus lib version since verus's tools cannot be installed directly via cargo. We can remove cargo-v with the updated script to download verus directly.

Cargo.toml
ziqiaozhou: Thanks for catching that! I should still use = in the version specifier. 

We expect newer versions to work as well. Once this PR is merged and SVSM’s verification is integrated into Verus’s Verita tests, the Verus team will notify us promptly if any changes are needed for future versions.

.cargo/config.toml
ziqiaozhou: Yes. I updated the doc.

ziqiaozhou: We can revert the workaround:>

scripts/vinstall.sh
stefano-garzarella: Thanks for the clarification!

Cargo.toml
stefano-garzarella: Sorry, I'm confused, I still see `=` missing...

scripts/vinstall.sh
```diff
@@ -5,14 +5,54 @@
 #
 # Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
 # A script to install verus tools
-VERISMO_REV=1840262abbcf32f0bd5e622e47fb18d7dfaca795
-VERUS_RUST_VERSION=1.94.0
+
+VERUS_RELEASE=0.2026.04.12.f1166c4 # May need to update if vstd or other verus library changes
+VERUS_RUST_VERUS_RELEASE=1.94.0 # May need to update if verus version changes
```
stefano-garzarella: Why changing this variable from `VERUS_RUST_VERSION` to `VERUS_RUST_VERUS_RELEASE` ?

scripts/vinstall.sh
```diff
@@ -5,14 +5,54 @@
...
 VERUSFMT_REV=beff2fa686d856d5e60df368fd027d94ead11ac5 # v0.5.7
 
-# Install x86_64-unknown-none target for verus-compatible Rust version
-export RUSTUP_TOOLCHAIN=$VERUS_RUST_VERSION
-rustup target add x86_64-unknown-none --toolchain $RUSTUP_TOOLCHAIN
-# Install the verus toolchain
-cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV cargo-v
-cargo v install-verus
-cargo install --git https://github.com/verus-lang/verusfmt  --rev $VERUSFMT_REV
+# Install x86_64-unknown-none target for verus-compatible Rust VERUS_RELEASE
```
stefano-garzarella: Please try to keep changes as small as possible, why changing this comment?

scripts/vinstall.sh
stefano-garzarella, line 1: Can you explain a bit in the commit description why now we need this complex script and also what is doing?

stefano-garzarella (commented):
I suggest following the style of this project for commit message (e.g. "subsystem: ...")

scripts/vinstall.sh
ziqiaozhou: I added a reason why we need a script to download verus toolchain.

Actually we have three options:
1. Keep cargo v install-verus to hide how to install verus.
2. Using this long script to download a prebuilt, which looks dirty but makes it clear what are installed.
2. An unofficial way to install: Run cargo install to install verus, create a verus-root, and download z3 using a script from verus project. This is not encouraged since verus projects uses a special build process not directly via `cargo`.  It only works for some project (including svsm) that explicitly uses vstd as non-builtin and does not work for other projects.
```
RUSTC_BOOTSTRAP=1 cargo +$VERUS_RUST_VERSION install --git https://github.com/verus-lang/verus --rev f1166c4 cargo-verus verus rust_verify --force
touch ~/.cargo/bin/verus-root
curl -sL https://raw.githubusercontent.com/verus-lang/verus/5e663299cf8be6734156862c61c9a313c5e8783a/source/tools/get-z3.sh | bash
```

With Option 3, verus tool will warn users `warning: verus was compiled outside vargo, and we cannot determine whether it was built in debug mode, which will result in worse performance` since this is an unofficial way to install it.

I think Option 2 > 1 > 3 and so updated the script and so picked 2. Let me know if you prefer other option.

Cargo.toml
ziqiaozhou: Sorry, I forgot to commit the change. Now `=` is back

scripts/vinstall.sh
```diff
@@ -5,14 +5,53 @@
...
+    Linux)  PLATFORM_OS="linux" ;;
+    Darwin) PLATFORM_OS="macos" ;;
+    *) echo "Error: unsupported OS: $OS" >&2; exit 1 ;;
+esac
+
+PLATFORM="${PLATFORM_ARCH}-${PLATFORM_OS}"
+RELEASE_TAG="release/${VERUS_RELEASE}"
+ZIPFILE="verus-${VERUS_RELEASE}-${PLATFORM}.zip"
+DOWNLOAD_URL="https://github.com/verus-lang/verus/releases/download/${RELEASE_TAG}/${ZIPFILE}"
+curl -sfL "$DOWNLOAD_URL" -o "$ZIPFILE"
```
stefano-garzarella: Should we download it in the tmp dir too?

scripts/vinstall.sh
stefano-garzarella: Okay, thanks for explain better the situation. Let's go in this direction :-)

scripts/vinstall.sh
```diff
@@ -5,17 +5,14 @@
 #
 # Author: Ziqiao Zhou <ziqiaozhou@microsoft.com>
 # A script to install verus tools
-VERISMO_REV=4f504b72ddd7a6d5b194b65db159e55a41831ab9
-VERUS_RUST_VERSION=1.91.0
+VERISMO_REV=1840262abbcf32f0bd5e622e47fb18d7dfaca795
+VERUS_RUST_VERSION=1.94.0
 VERUSFMT_REV=beff2fa686d856d5e60df368fd027d94ead11ac5 # v0.5.7
```
stefano-garzarella, line 36: Since this script is becoming bigger, what about adding `set -e`, so if for example `curl` will fail to download, we stop immediately.

stefano-garzarella (commented):
LGTM, I left some minor comments, plus `claude` here is highlighting some typos in the commit messages:
> Typos across the series: There are typos in 3 of the 4 commit messages ("referece" in [1/4], "relect" and "alia" in [2/4], "erros" in [4/4], "vsinstall" in [3/4]).

stefano-garzarella (approved):

scripts/vinstall.sh
```diff
@@ -5,17 +5,55 @@
...
+esac
+
+PLATFORM="${PLATFORM_ARCH}-${PLATFORM_OS}"
+ZIPFILE="verus-${VERUS_RELEASE}-${PLATFORM}.zip"
+DOWNLOAD_URL="https://github.com/verus-lang/verus/releases/download/release/${VERUS_RELEASE}/${ZIPFILE}"
+TMPDIR=$(mktemp -d)
+
+# Download Verus prebuilt into a tmp folder
+curl -sfL "$DOWNLOAD_URL" -o "$TMPDIR/$ZIPFILE"
+unzip -q "$TMPDIR/$ZIPFILE" -d "$TMPDIR"
```
joergroedel: Sorry, but we are not re-introducing the binary downloads into our CI, this was a problematic practice noticed by the OpenSSF Scorecard. Please install verus by building it inside CI by a pinned hash (not a git tag).

ziqiaozhou: How about run `sha256sum --check` after downloading a prebuilt? download and build verus takes around 5min in CI.

joergroedel: I am willing to take the 5minutes penalty, the verus check is still optional in CI. We can also add caching later if needed, just like we do for out QEMU tests.

ziqiaozhou: Okay. The updated script will take around 17min in CI. 
Building verus takes around 5min
Building z3 takes around 10min. verus's official installation process uses a script to fetch z3 prebuilt and so I get rid of that step and build z3 from source code.

scripts/vinstall.sh
```diff
@@ -39,36 +46,86 @@ fetch_code() {
...
 
+# Install verusfmt
+cargo install --git https://github.com/verus-lang/verusfmt  --rev $VERUSFMT_REV
+
 # Create a temporary directory
 TMPDIR=$(mktemp -d)
 VERUS_DIR=$TMPDIR/verus
-# Build and install Verus from source
-fetch_code https://github.com/verus-lang/verus.git $VERUS_RELEASE
-git checkout $VERUS_RELEASE
```
stefano-garzarella: I'm really confused, why a commit that add just a way to download pre-build binaries, do all of this changes?

For example, why `git checkout $VERUS_RELEASE` is not needed anymore?

scripts/vinstall.sh
```diff
@@ -39,36 +46,86 @@ fetch_code() {
...
 cd $VERUS_DIR/source
 rustup component add rust-src rustc-dev llvm-tools-preview --toolchain $RUSTUP_TOOLCHAIN
-vargo build --release
-
-# Move the Verus and z3 binaries to the final installation directory
-for asset in "${VERUS_ASSETS[@]}"; do
-    echo "Installing $asset to ~/.cargo/bin/"
-    mv "$VERUS_DIR/source/target-verus/release/$asset" ~/.cargo/bin/
-done
+vargo build --release --vstd-no-verify
```
stefano-garzarella, line 137: why now we need `--vstd-no-verify` ?

scripts/vinstall.sh
```diff
@@ -7,28 +7,35 @@
...
+# Z3 version and commit hash
+VERUS_Z3_VERSION=4.12.5
+VERUS_Z3_REV=a7b564cafe3b96c8a868388bc4b96b319facea44
+
+# Parse arguments
+INSTALL_PREBUILT=false
+FORCE_INSTALL=false
+for arg in "$@"; do
+    case "$arg" in
+        --use-prebuilt) INSTALL_PREBUILT=true ;;
```
stefano-garzarella, line 31: Should we add this option in the documentation?

Also what `--force` is supposed to do?

scripts/vinstall.sh
ziqiaozhou: git checkout $VERUS_RELEASE is redundant with fetch_code, which I should change it in previous commit instead of this one. I moved this change to earlier commit.

scripts/vinstall.sh
ziqiaozhou: We can skip vstd verification step to save some time when build verus. I just realized that I forgot to add this flag in previous commit and so I added it here.

scripts/vinstall.sh
ziqiaozhou: --force will install all required verus tools (verus, rust_verify, z3, cargo-verus) in case that your installation is incomplete and some of tools are not at the desired version.
I updated the doc to explain the two options.

scripts/vinstall.sh
```diff
@@ -5,14 +5,80 @@
...
+        echo "Installing $asset to ~/.cargo/bin/"
+        mv "$src_dir/$asset" ~/.cargo/bin/
+    done
+}
+
+# Skip building Verus from source if the correct version is already installed
+if (verus --version | grep -q "${VERUS_VERSION:0:7}") &> /dev/null && ! $FORCE_INSTALL; then
+    echo "Verus version ${VERUS_VERSION:0:7} already installed, skipping build."
+    exit 0
+fi
```
stefano-garzarella: `VERUS_VERSION`, `FORCE_INSTALL` are added in a next commit, so here they are undefined.

stefano-garzarella (commented):
@ziqiaozhou thanks for adding the pre-build option!

The final code LGTM but I see something strange commit by commit, maybe rebase went wrong.

Feel free to squash in a single patch changes related to vinstall.sh if it helps

stefano-garzarella:
@ziqiaozhou is also the verification failure in the CI related to #1059 ?

stefano-garzarella (dismissed):

scripts/vinstall.sh
```diff
@@ -5,14 +5,127 @@
...
+    )
+    local src_dir=$1
+    for asset in "${VERUS_ASSETS[@]}"; do
+        echo "Installing $asset to ~/.cargo/bin/"
+        mv "$src_dir/$asset" ~/.cargo/bin/
+    done
+}
+
+# Skip building Verus from source if the correct version is already installed
+if (verus --version | grep -q "${VERUS_VERSION:0:7}") &> /dev/null && ! $FORCE_INSTALL; then
```
stefano-garzarella: why checking only the first 7 characters ?

scripts/vinstall.sh
```diff
@@ -5,14 +5,127 @@
...
-# Install the verus toolchain
-cargo install --git https://github.com/microsoft/verismo/ --rev $VERISMO_REV cargo-v
-cargo v install-verus
+
+fetch_code() {
+    url=$1
+    repo_name=$(basename "$url" .git)
+    commit=$2
+    git clone --no-checkout --depth 1 "$url" "$TMPDIR/$repo_name"
+    cd "$TMPDIR/$repo_name"
```
stefano-garzarella: I'd suggest to use `pushd` `popd`. It's fragile to change a directory inside a function and don't restore it.

scripts/vinstall.sh
```diff
@@ -5,14 +5,127 @@
...
+        "verus"
+        "rust_verify"
+        "z3"
+        "cargo-verus"
+        "verus-root"
+    )
+    local src_dir=$1
+    for asset in "${VERUS_ASSETS[@]}"; do
+        echo "Installing $asset to ~/.cargo/bin/"
+        mv "$src_dir/$asset" ~/.cargo/bin/
```
stefano-garzarella: can we use `install` ?

scripts/vinstall.sh
stefano-garzarella: Also I'd pass the destdir as parameter, using TMPDIR variable is fragile IMHO

ziqiaozhou:
> @ziqiaozhou is also the verification failure in the CI related to #1059 ?

Yes

scripts/vinstall.sh
ziqiaozhou: That is a good catch! I was using the commit hash to check version earlier before adding the prebuilt solution. I fixed it to check the full version.

stefano-garzarella (approved):

stefano-garzarella:
LGTM, let's wait #1059 merged, then we should rebase this.

stefano-garzarella (approved):

stefano-garzarella:
FYI I'm adding cache on top of this with #1079 (still testing it)

scripts/vinstall.sh
joergroedel: That is better, thanks.

joergroedel (approved):

PR #1039, opened 2026-04-19
task: refactor task scheduling

msft-jlange:
To support wait states and other scenarios in which tasks are not immediately scheduled for execution, it is helpful to change the scheduler logic so that it does not unconditionally prepare the current task to be resumed later on the current CPU.  Instead, modify the scheduler interface so that different scheduler entry points can control the behavior to reschedule when the time comes to select a different task for execution.

### DIFF
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -402,7 +402,7 @@ pub fn terminate() -> ! {
     current_task_terminated();
 
     // The current task will not run again, so switch to a different task.
-    select_new_task(false);
+    select_new_task(false, None);
     unreachable!("terminated task rescheduled");
 }
 
@@ -423,7 +423,7 @@ pub fn go_idle() {
 
     // Find another task to run.  If no other task is runnable, then the idle
     // thread will execute.
-    select_new_task(false);
+    select_new_task(false, None);
 }
 
 pub fn set_affinity(cpu_index: usize) {
@@ -446,7 +446,7 @@ pub fn set_affinity(cpu_index: usize) {
 
         // Find another task to run.  The scheduler will complete the affinity
         // change once a new task has been selected on this processor.
-        select_new_task(false);
+        select_new_task(false, None);
     }
 }
 
@@ -538,14 +538,16 @@ pub fn schedule() {
     // check if preemption is safe
     preemption_checks();
 
-    select_new_task(true);
+    select_new_task(true, None);
 }
 
 /// Select another task to run.  If rescheduling is requested, the current
 /// task will be placed back on the current processor's run queue so it can
 /// be eligible to run again.
-fn select_new_task(reschedule: bool) {
-    let guard = IrqGuard::new();
+fn select_new_task(reschedule: bool, irq_guard: Option<IrqGuard>) {
+    // If the caller has not already disabled interrupts, then disable them
+    // now.
+    let guard = irq_guard.unwrap_or_default();
 
     let work = this_cpu().schedule_prepare(reschedule);
 
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -110,15 +110,12 @@ impl RunQueue {
             .unwrap_or_else(|| self.idle_task.clone().unwrap())
     }
 
-    /// Update state before a task is scheduled out. Non-idle tasks in RUNNING
-    /// state will be put at the end of the run_list. Terminated tasks will be
-    /// stored in the terminated_task field of the RunQueue and be destroyed
-    /// after the task-switch.
-    fn handle_task(&mut self, task: TaskPointer) {
-        if task.is_running() && !task.is_idle_task() {
+    /// Place a task back onto the run queue so it can be scheduled again.
+    fn enqueue_task(&mut self, task: TaskPointer) {
+        // Task termination should not follow this path.
+        debug_assert!(!task.is_terminated());
+        if !task.is_idle_task() {
             self.run_list.push_back(task);
-        } else if task.is_terminated() {
-            self.terminated_task = Some(task);
         }
     }
 
@@ -137,8 +134,14 @@ impl RunQueue {
 
     /// Prepares a task switch. The function checks if a task switch needs to
     /// be done and return pointers to the current and next task. It will
-    /// also call `handle_task()` on the current task in case a task-switch
+    /// also call `enqueue_task()` on the current task in case a task-switch
     /// is requested.
+    ///
+    /// # Parameters
+    ///
+    /// - `reschedule`: Indicates whether the current task being rescheduled.
+    ///   If so, it will be reinserted on the run list of the current
+    ///   processor.
     ///
     /// # Returns
     ///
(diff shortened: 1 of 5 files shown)


kernel/src/task/schedule.rs
```diff
@@ -114,11 +114,11 @@ impl RunQueue {
...
     /// after the task-switch.
-    fn handle_task(&mut self, task: TaskPointer) {
-        if task.is_running() && !task.is_idle_task() {
+    fn enqueue_task(&mut self, task: TaskPointer) {
+        // Task termination should not follow this path.
+        debug_assert!(!task.is_terminated());
+        if !task.is_idle_task() {
             self.run_list.push_back(task);
-        } else if task.is_terminated() {
-            self.terminated_task = Some(task);
```
joergroedel, line 121: Where is this set now?

kernel/src/task/schedule.rs
```diff
@@ -769,10 +769,33 @@ global_asm!(
...
+    extern crate alloc;
     use crate::cpu::percpu::{PERCPU_AREAS, this_cpu};
+    use crate::task::KernelThreadStartInfo;
     use crate::task::set_affinity;
+    use crate::task::start_kernel_task;
+    use alloc::string::String;
+    use core::sync::atomic::AtomicU32;
+    use core::sync::atomic::Ordering;
+
+    static EMPTY_TASK_COUNTER: AtomicU32 = AtomicU32::new(0);
```
stefano-garzarella, line 823: Was your goal to check the value of `EMPTY_TASK_COUNTER` after scheduling the `empty_task` to see if it had executed (which is missing) or just something to do in the empty_task?

I mean, should we add some `assert_eq!` on that value in `test_task_termination()` ?

kernel/src/task/tasks.rs
```diff
@@ -1003,8 +1004,7 @@ unsafe fn run_kernel_task<T: KernelThreadStartParameter>(
 }
 
 fn task_exit() {
-    current_task_terminated();
-    schedule();
+    terminate();
```
stefano-garzarella, line 1007: Should we apply the same also to `sys_exit()` in `kernel/src/syscall/class0.rs` ?

kernel/src/task/schedule.rs
```diff
@@ -415,7 +435,7 @@ pub fn set_affinity(cpu_index: usize) {
 
         // Find another task to run.  The scheduler will complete the affinity
         // change once a new task has been selected on this processor.
-        schedule();
```
stefano-garzarella, line 418: so essentially we are skipping `preemption_checks()` now, is it okay?

kernel/src/task/schedule.rs
```diff
@@ -140,6 +140,12 @@ impl RunQueue {
     /// also call `handle_task()` on the current task in case a task-switch
```
stefano-garzarella: `handle_task()` -> `enqueue_task()`

kernel/src/task/schedule.rs
```diff
@@ -527,14 +527,16 @@ pub fn schedule() {
...
-    select_new_task(true);
+    select_new_task(true, None);
 }
 
 /// Select another task to run.  If rescheduling is requested, the current
 /// task will be placed back on the current processor's run queue so it can
 /// be eligible to run again.
-fn select_new_task(reschedule: bool) {
-    let guard = IrqGuard::new();
+fn select_new_task(reschedule: bool, irq_guard: Option<IrqGuard>) {
```
stefano-garzarella, line 547: IIUC `irq_guard` is always None in current code, so this is in preparation for something else, right?

kernel/src/task/schedule.rs
msft-jlange: My goal was to ensure that `empty_task()` had a non-empty function body, and giving it a counter to increment is a trivial way to ensure the function has logic.  The value of the counter is irrelevant.

kernel/src/task/tasks.rs
msft-jlange: Yes, that should be fixed.

kernel/src/task/schedule.rs
msft-jlange: Correct, there are two future PRs that both use this field.

kernel/src/task/schedule.rs
msft-jlange: We should perform those checks here too.

kernel/src/task/schedule.rs
msft-jlange: I missed this, so it's now coming back in `current_task_terminated()`.  However, I think this can be improved so this RQ field is not required at all, and I'll include that in a future PR.

kernel/src/task/schedule.rs
stefano-garzarella: Maybe I'm missing something, but this seems still unresolved.

kernel/src/task/schedule.rs
stefano-garzarella: Should we use `select_new_task()` here too?

kernel/src/task/schedule.rs
msft-jlange: Oh, I misread the line numbers and I thought you were talking about `go_idle()`, which had the same problem (and is now fixed).

kernel/src/task/schedule.rs
msft-jlange: You are correct.

stefano-garzarella (approved):

kernel/src/task/schedule.rs
```diff
@@ -769,10 +769,33 @@ global_asm!(
...
+
+    fn empty_task(_parameter: usize) {
+        EMPTY_TASK_COUNTER.fetch_add(1, Ordering::Relaxed);
+    }
+
+    #[test]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_task_termination() {
+        // Start a task that will immediately terminate.
+        start_kernel_task(
```
00xc, line 793: Shouldn't we have more thorough checks here? Something like spinning until the task is terminated and then verifying that `EMPTY_TASK_COUNTER` was incremented only once.

msft-jlange: As I wrote earlier, `EMPTY_TASK_COUNTER` exists just to give the task something to do so its body isn't empty.  Checking the counter isn't particularly helpful here since it only indicates that the task started, not that it terminated correctly.  The body of the task doesn't include any loop so there is no way for the task to cause the counter to increment multiple times; there is certainly no facility that could possibly cause a task to restart from the beginning.  Besides that, there is no way today to check to see whether a task is still running, so there is no way to spin for completion.  I do have an upcoming PR that implements "wait for task exit", but that is a blocking wait and not a spin wait, and it comes with its own test.

The real value of this test is not to verify any specific side-effects, but rather to verify that the code doesn't raise any assertions as a result of the termination - such as the asserts that are in the `Drop` implementation of `Task`.  The test is successful not because it verifies its result, but because it completes without blocking forever nor without asserting.  The same will be true of the "wait for task exit" test.

PR #1041, opened 2026-04-24
vsock/stream: fix SVSM crash when device is not available

luigix25:
When a vsock device is not attached to the guest and VsockStream::connect() is called, SVSM panics because `VSOCK_DEVICE` is accessed but it is uninitialized.

Add a check using try_get_inner() at the start of connect() to return an error instead of panicking.

### DIFF
--- a/kernel/src/vsock/stream.rs
+++ b/kernel/src/vsock/stream.rs
@@ -62,12 +62,16 @@ impl VsockStream {
     /// # Returns
     ///
     /// Returns a connected `VsockStream` on success, or an error if:
-    /// - The VSOCK device is not available (`VsockError::DriverError`)
+    /// - The VSOCK device is not available (`VsockError::DeviceNotAvailable`)
     /// - No free local ports are available
     /// - The connection fails
     pub fn connect(remote_port: u32, remote_cid: u32) -> Result<Self, SvsmError> {
-        let local_port = VSOCK_DEVICE.get_first_free_port()?;
-        VSOCK_DEVICE.connect(remote_cid, local_port, remote_port)?;
+        let device = VSOCK_DEVICE
+            .try_get_inner()
+            .map_err(|_| SvsmError::Vsock(VsockError::DeviceNotAvailable))?;
+
+        let local_port = device.get_first_free_port()?;
+        device.connect(remote_cid, local_port, remote_port)?;
 
         Ok(Self {
             local_port,
@@ -94,9 +98,13 @@ impl Read for VsockStream {
     /// # Errors
     ///
     /// Returns an error if:
-    /// - The VSOCK device is not available (`VsockError::DriverError`)
+    /// - The VSOCK device is not available (`VsockError::DeviceNotAvailable`)
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Err> {
-        match VSOCK_DEVICE.recv(self.remote_cid, self.local_port, self.remote_port, buf) {
+        let device = VSOCK_DEVICE
+            .try_get_inner()
+            .map_err(|_| SvsmError::Vsock(VsockError::DeviceNotAvailable))?;
+
+        match device.recv(self.remote_cid, self.local_port, self.remote_port, buf) {
             Ok(value) => Ok(value),
             Err(SvsmError::Vsock(VsockError::NotConnected)) => Ok(0),
             Err(SvsmError::Vsock(VsockError::PeerSocketShutdown)) => Ok(0),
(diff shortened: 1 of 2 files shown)


kernel/src/vsock/stream.rs
```diff
@@ -66,6 +66,9 @@ impl VsockStream {
     /// - No free local ports are available
     /// - The connection fails
     pub fn connect(remote_port: u32, remote_cid: u32) -> Result<Self, SvsmError> {
+        VSOCK_DEVICE
+            .try_get_inner()
+            .map_err(|_| SvsmError::Vsock(VsockError::DriverError))?;
```
stefano-garzarella: Why `DriverError` ? 

What about adding a specific error (if we don't have already one) about the device availability or remote peer availability.

luigix25: To be fair the doc mentioned a `DriverError` but, `DeviceNotAvailable` makes more sense.

joergroedel (dismissed):

kernel/src/vsock/stream.rs
```diff
@@ -62,10 +62,13 @@ impl VsockStream {
     /// # Returns
     ///
     /// Returns a connected `VsockStream` on success, or an error if:
-    /// - The VSOCK device is not available (`VsockError::DriverError`)
```
stefano-garzarella, line 65: mmm, so we had this comment before, but without really checking it? Or we do this check in some other place?

kernel/src/vsock/stream.rs
stefano-garzarella: Partially unrelated to this PR, but is this true ?

(maybe not to be fixed in this PR, but I noticed it while reviewing)

stefano-garzarella: Ditto.

kernel/src/vsock/stream.rs
```diff
@@ -62,10 +62,13 @@ impl VsockStream {
...
+    /// - The VSOCK device is not available (`VsockError::DeviceNotAvailable`)
     /// - No free local ports are available
     /// - The connection fails
     pub fn connect(remote_port: u32, remote_cid: u32) -> Result<Self, SvsmError> {
+        VSOCK_DEVICE
+            .try_get_inner()
+            .map_err(|_| SvsmError::Vsock(VsockError::DeviceNotAvailable))?;
         let local_port = VSOCK_DEVICE.get_first_free_port()?;
         VSOCK_DEVICE.connect(remote_cid, local_port, remote_port)?;
```
stefano-garzarella: nit: in this way we avoid the extra 2 `Deref`

```suggestion
        let device = VSOCK_DEVICE
            .try_get_inner()
            .map_err(|_| SvsmError::Vsock(VsockError::DeviceNotAvailable))?;
        let local_port = device.get_first_free_port()?;
        device.connect(remote_cid, local_port, remote_port)?;
```

kernel/src/vsock/stream.rs
stefano-garzarella: Just to be clear, IMHO is better to add `let device = VSOCK_DEVICE.try_get_inner()...` in every place to avoid panics and return errors in the future. I know that for now a VsockStream can only be created by `connect()` but in the future we may add new ways and forgot those checks.

luigix25: yep, will do.

kernel/src/vsock/stream.rs
luigix25: I think we had this check before, it was remove but the doc was not updated accordingly.

luigix25:
- Added `VSOCK_DEVICE` check to every function for defensive programming
- Updated the doc, removing a reference to an old error
- Updated Drop to match the same pattern with the other functions, reducing Deref.

kernel/src/vsock/stream.rs
```diff
@@ -117,20 +125,23 @@ impl Write for VsockStream {
...
 impl Drop for VsockStream {
     fn drop(&mut self) {
-        if VSOCK_DEVICE.try_get_inner().is_err() {
+        let Ok(device) = VSOCK_DEVICE.try_get_inner() else {
             return;
-        }
-
-        let _ = VSOCK_DEVICE.shutdown(self.remote_cid, self.local_port, self.remote_port, true);
+        };
+        let _ = device.shutdown(self.remote_cid, self.local_port, self.remote_port, true);
```
stefano-garzarella: nit:

```suggestion
        if let Ok(device) = VSOCK_DEVICE.try_get_inner() {
            let _ = device.shutdown(self.remote_cid, self.local_port,               
  self.remote_port, true);
        }    
```

stefano-garzarella (dismissed):
Just a minor nit, but I'm fine also with this version.

kernel/src/vsock/stream.rs
luigix25: done

stefano-garzarella (approved):

joergroedel (approved):

PR #1044, opened 2026-04-27
task: implement `wait_for_termination`

msft-jlange:
Implement a new function that permits waiting for a task to terminate.  No wait is performed if the task has already terminated.  If the task is still executing, the caller will enter a wait state and will be woken on the same CPU on which the terminating task last ran.

### DIFF
--- a/kernel/src/task/exec.rs
+++ b/kernel/src/task/exec.rs
@@ -52,7 +52,7 @@ fn task_name(binary: &str) -> String {
 ///
 /// # Returns
 ///
-/// [`Ok(tid)`] on success, [`Err(SvsmError)`] on failure.
+/// [`Ok(TaskPointer)`] on success, [`Err(SvsmError)`] on failure.
 pub fn exec_user(binary: &str, root: Arc<dyn Directory>) -> Result<TaskPointer, SvsmError> {
     let fh = open_read(binary)?;
     let file_size = fh.size();
--- a/kernel/src/task/exec.rs
+++ b/kernel/src/task/exec.rs
@@ -6,6 +6,7 @@
 
 extern crate alloc;
 
+use super::TaskPointer;
 use crate::address::{Address, VirtAddr};
 use crate::error::SvsmError;
 use crate::fs::{Directory, open_read};
@@ -52,7 +53,7 @@ fn task_name(binary: &str) -> String {
 /// # Returns
 ///
 /// [`Ok(tid)`] on success, [`Err(SvsmError)`] on failure.
-pub fn exec_user(binary: &str, root: Arc<dyn Directory>) -> Result<u32, SvsmError> {
+pub fn exec_user(binary: &str, root: Arc<dyn Directory>) -> Result<TaskPointer, SvsmError> {
     let fh = open_read(binary)?;
     let file_size = fh.size();
 
@@ -117,5 +118,5 @@ pub fn exec_user(binary: &str, root: Arc<dyn Directory>) -> Result<u32, SvsmErro
     finish_user_task(new_task.clone());
     schedule();
 
-    Ok(new_task.get_task_id())
+    Ok(new_task)
 }
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -274,15 +274,18 @@ impl TaskList {
     /// # Safety
     /// The caller must ensure that `task` is already a member of this task
     /// list.
-    unsafe fn terminate(&mut self, task: TaskPointer) {
+    unsafe fn terminate(&mut self, task: TaskPointer) -> Option<TaskPointer> {
         // Set the task state as terminated. If the task being terminated is the
         // current task then the task context will still need to be in scope until
         // the next schedule() has completed. Schedule will keep a reference to this
         // task until some time after the context switch.
-        task.set_task_terminated();
+        let wakeup = task.set_task_terminated();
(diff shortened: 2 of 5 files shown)


kernel/src/task/schedule.rs
```diff
@@ -372,6 +375,23 @@ pub fn is_current_task(id: u32) -> bool {
     }
 }
 
+/// Waits for a task to terminate
+#[allow(dead_code)]
```
stefano-garzarella: I think we can remove `#[allow(dead_code)]`, it's a pub function also exported with a `pub use`, so should be fine.

kernel/src/task/schedule.rs
```diff
@@ -544,7 +571,7 @@ pub fn schedule() {
 /// Select another task to run.  If rescheduling is requested, the current
 /// task will be placed back on the current processor's run queue so it can
 /// be eligible to run again.
-fn select_new_task(reschedule: bool, irq_guard: Option<IrqGuard>) {
+pub fn select_new_task(reschedule: bool, irq_guard: Option<IrqGuard>) {
```
stefano-garzarella: Why this is needed ?

kernel/src/task/schedule.rs
```diff
@@ -837,6 +871,33 @@ mod test {
         .expect("Failed to start test termination task");
     }
 
+    #[test]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_wait_for_task() {
```
stefano-garzarella: nit: the commit title and this test talk about `wait_for_task` but we are adding `wait_for_termination`

kernel/src/task/tasks.rs
```diff
@@ -752,6 +766,35 @@ impl Task {
...
+        // is marked terminated before its wait queue is examined, this
+        // sequence guarantees that a task will not block unless it is
+        // guaranteed to be woken.
+        let mut wait_queue = self.wait_queue.lock();
+        if self.is_terminated() {
+            return None;
+        }
+
+        // Join the wait queue of the target task.
+        wait_queue.wait_for_event(current_task);
```
stefano-garzarella, line 796: In `wait_for_event()` we have `assert!(self.waiter.is_none());` so IIUC `wait_for_exit()` and so `wait_for_termination()` can be called only by a single task. If there is a task waiting and a second one want to wait, we will panic, should we at least document this?

kernel/src/task/schedule.rs
msft-jlange: I added this because Clippy demanded it but now Clippy appears to be satisfied without it so I'll remove it.

kernel/src/task/schedule.rs
msft-jlange: That was needed for some early version of this code but it seems that it is no longer required.

kernel/src/task/tasks.rs
msft-jlange: The restriction on wait has been present since the wait queue was first introduced.  It should be improved but that's outside the scope of this PR.

kernel/src/task/schedule.rs
```diff
@@ -372,6 +375,22 @@ pub fn is_current_task(id: u32) -> bool {
     }
 }
 
+/// Waits for a task to terminate
+pub fn wait_for_termination(task: TaskPointer) {
```
n-ramacciotti, line 397: Would it make sense for this to return the exit code of the task?

kernel/src/task/tasks.rs
joergroedel, line 796: What prevents this code from waiting on the current task? IIUC this will dead-lock. Should we also handle the case of multiple waiters more gracefully instead of triggering the `assert!` in `wait_for_event()`?

kernel/src/task/schedule.rs
msft-jlange: If we want to make the exit code available, it should probably come from a different function which specifically deals with exit codes.  The wait function should only deal with the behavior of the wait.  For example, if we decide someday to add a timeout ("wait for termination but no more than X time") then the result of the wait will be whether or not it timed out.

kernel/src/task/tasks.rs
msft-jlange: Deadlock is not a safety concern, so safe code does not require deadlock detection.  This is also true of spin locks - there's nothing that stops code from trying to acquire a spin lock that is already held by the current thread.  I think this is an area where we should trust developers to consider deadlock when writing code instead of littering the code with deadlock detection.

As I wrote in another comment, the existing wait queue infrastructure that you wrote only comprehends a single waiter.  We definitely need to fix this someday but that's out of scope for this PR.

kernel/src/task/exec.rs
```diff
@@ -52,7 +53,7 @@ fn task_name(binary: &str) -> String {
 /// # Returns
 ///
 /// [`Ok(tid)`] on success, [`Err(SvsmError)`] on failure.
```
stefano-garzarella: doc-comment here needs an update

stefano-garzarella (commented):
nit: commit https://github.com/coconut-svsm/svsm/pull/1044/changes/f41908ff7d67a7a2a26873c59dcaf0c861c7a02c title still refers to `wait_for_task` instead of `wait_for_termination`

joergroedel (approved):

PR #1045, opened 2026-04-27
task: simplify `set_affinity`

msft-jlange:
Simplify the logic of `set_affinity`.  This eliminates all of the post-switch logic to finalize affinity changes, and instead simply inserts the running task on the run queue of the target processor and wakes it to ensure that run queue evaluation will be performed.


kernel/src/task/schedule.rs
```diff
@@ -434,19 +433,26 @@ pub fn set_affinity(cpu_index: usize) {
...
         let task = this_cpu().current_task();
-        task.set_task_blocked();
+        let target_cpu = PERCPU_AREAS.get_by_cpu_index(cpu_index);
+
+        // Disable interrupts to prevent delays in scheduling once the task
+        // has been queued on the target processor.
+        let guard = IrqGuard::new();
+
+        // Join this task to the run queue of the target CPU.
+        target_cpu.runqueue_mut().enqueue_task(task);
```
joergroedel, line 487: I think there is a race-condition here. After the task is sent to another CPU it may get its per-cpu mappings updated. This happens already before it is scheduled in, so not protected by the TASK_ACTIVE tracking. On the source CPU the `select_new_task` function will access per-cpu, and might use a wrong per-cpu mapping for that.

joergroedel (commented):
Besides the comment I ran some testing. In release builds I have seen a boot-lockup when starting the request-loop processing threads on AMD SEV-SNP.

kernel/src/task/schedule.rs
msft-jlange: I see your point.  What you're really pointing out is that the "task active" tracking that I added in my previous PR was not sufficient: we need to prevent not only a stack switch but also a CPU address space switch until the task is idle.  I would like to submit that as a separate PR because it's an issue with core task switching and isn't specifically related to affinity changes.

joergroedel (approved):

PR #1048, opened 2026-04-29
github/workflows: add apt update before apt install in publish-docs

luigix25:
GitHub Actions runners ship with a pre-populated apt cache that can become stale. When this happens, apt install fails because the cached index references package URLs that no longer exist on the mirror:

`E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?`

Add an explicit apt update step to ensure the package index is current.

This adds ~12 seconds to CI time, but does not impact PRs CI time as this workflow is only executed on pushes to the main branch.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -34,6 +34,9 @@ jobs:
             profile: minimal
             override: true
 
+      - name: Update apt package list
+        run: sudo apt update
+
       - name: Install TPM 2.0 Reference Implementation build dependencies
         run: sudo apt install -y build-essential cmake pkg-config
 


stefano-garzarella (approved):
This is also recommended here: https://docs.github.com/en/actions/how-tos/manage-runners/github-hosted-runners/customize-runners

> Always run sudo apt-get update before installing a package. In case the apt index is stale, this command fetches and re-indexes any available packages, which helps prevent package installation failures.

@joergroedel should we merge this before the release ?

Also should we do this also in `qemu.yml` and `rust.yml`? I'm worried about adding 12 sec on both, so not sure if postpone the change till we have an issue.

stefano-garzarella:
Or what about using this: https://github.com/marketplace/actions/cache-apt-packages

PR #1049, opened 2026-05-04
SVSM/apic: always prefer HV_IPI over x2APIC MSR write for ICR

Goodleon-Y:
Commit 60eb14cc gated hv_ipi() behind SEV_SNP_RESTR_INJ feature check. KVM without RESTR_INJ falls through to WRMSR on x2APIC ICR (MSR 0x830), which KVM rejects via GHCB MSR protocol (error 2).

HV_IPI (GHCB exit 0x0015) is a base SEV-ES protocol feature and does not require restricted injection. This patch always prefers HV_IPI, falling back to direct MSR write only if HV_IPI is unsupported.

Fixes: 60eb14cc ("SVSM/apic: avoid the use of the APIC emulation")

### DIFF
--- a/kernel/src/sev/snp_apic.rs
+++ b/kernel/src/sev/snp_apic.rs
@@ -65,15 +65,18 @@ impl ApicAccess for GHCBApicAccessor {
     }
 
     fn icr_write(&self, icr: u64) -> Result<(), SvsmError> {
-        // The #HV IPI can only be used if restricted injection is supported.
-        // Otherwise, the IPI must be sent via an X2APIC ICR write.
-        if self.use_restr_inj() {
-            current_ghcb().hv_ipi(icr)?;
-        } else {
-            self.apic_write(APIC_OFFSET_ICR, icr);
+        // HV_IPI (GHCB exit 0x0015) is a base SEV-ES protocol feature and
+        // does not require restricted injection. KVM rejects direct WRMSR to
+        // the x2APIC ICR MSR (0x830) via the GHCB MSR protocol (error 2),
+        // so HV_IPI is the only reliable path.
+        match current_ghcb().hv_ipi(icr) {
+            Ok(()) => Ok(()),
+            Err(_) => {
+                // Fallback: try direct MSR write if HV_IPI is unsupported
+                self.apic_write(APIC_OFFSET_ICR, icr);
+                Ok(())
+            }
         }
-
-        Ok(())
     }
 
     fn eoi(&self) {


kernel/src/sev/snp_apic.rs
```diff
@@ -65,15 +65,18 @@ impl ApicAccess for GHCBApicAccessor {
     }
 
     fn icr_write(&self, icr: u64) -> Result<(), SvsmError> {
-        // The #HV IPI can only be used if restricted injection is supported.
-        // Otherwise, the IPI must be sent via an X2APIC ICR write.
-        if self.use_restr_inj() {
-            current_ghcb().hv_ipi(icr)?;
-        } else {
-            self.apic_write(APIC_OFFSET_ICR, icr);
+        // HV_IPI (GHCB exit 0x0015) is a base SEV-ES protocol feature and
```
msft-jlange, line 68: HV_IPI is not a base protocol feature.  According to the AMD GHCB documentation, support for HV_IPI is part of the "SEV-SNP Restricted Injection" feature, and support is not guaranteed unless bit 2 is set in the GHCB FEATURES bitmap.  See Section 2.2 of the AMD GHCB specification.

Goodleon-Y: Thanks for the review. I looked into this carefully and believe the patch is correct as written, for the following reasons:

  1. HV_IPI is a base SEV-ES GHCB exit, not RESTR_INJ-specific

  The original code in commit 60eb14cc that gated hv_ipi() behind use_restr_inj() appears to have conflated two separate things:
  - HV_IPI (GHCB exit 0x0015): guest → hypervisor IPI delivery request, available since SEV-ES
  - Restricted Injection (GHCB FEATURES bit 2): SNP extension that changes how the hypervisor delivers interrupts into the guest, including the no_eoi_required optimization

  These are orthogonal: the guest can always request IPI delivery via HV_IPI, regardless of whether the hypervisor supports Restricted Injection for incoming interrupt delivery.

  2. Practical behavior on KVM

  On current KVM without RESTR_INJ advertising, the original code falls through to apic_write(APIC_OFFSET_ICR, icr) which issues a WRMSR to x2APIC ICR (MSR 0x830) via the GHCB MSR protocol. KVM rejects this
  WRMSR — the GHCB MSR protocol does not permit guest writes to the x2APIC ICR under SEV-SNP. The result is that all IPIs fail silently, breaking vCPU communication.

  The patch simply tries HV_IPI first (as the protocol intends) and falls back to direct MSR write only if HV_IPI itself is genuinely unsupported.

  3. The eoi() path is unaffected

  The eoi() method at line 79 already correctly distinguishes: it checks use_restr_inj() only to decide the no_eoi_required optimization. That check is appropriate because no_eoi_required is a
  RESTR_INJ-specific doorbell feature. This is a different situation from icr_write().

  Let me know if I'm missing a corner of the AMD spec where HV_IPI is explicitly tied to RESTR_INJ — happy to adjust the reasoning if there's a newer revision I haven't checked.

msft-jlange: KVM is not the only host that supports SNP, so the behavior of KVM is only valuable as a reference.  Only the GHCB specification can be relied upon to describe the behavior that is correct across all possible implementations of SNP hosts.

AMD APM volume 2, appendix B is titled "VMCB Layout".  I cannot find any references to GHCB from the APM.  The GHCB protocol is defined in a separate specification, titled, "SEV-ES Guest-Hypervisor Communication Block Standardization".

Section 4.1.11 of the GHCB specification states specifically that "The #HV Send IPI NAE event allows for an SEV-SNP guest to send an IPI to other vCPUs in the guest when the Restricted Injection feature is enabled."  It further says that "If the Restricted Injection feature is not enabled, the hypervisor may request standard VMGEXIT 
error processing protocol to signal a #GP exception."  This language seems to pretty clearly indicate that the #HV IPI` feature is strongly coupled to restricted injection

Goodleon-Y: You're right, thank you for the citation. I checked AMD doc #56421 §4.1.11 which explicitly states HV_IPI requires Restricted Injection. My commit message was wrong, and I'll close this PR.

  Following this to its logical conclusion: without RESTR_INJ, the spec-compliant fallback is WRMSR to x2APIC ICR (MSR 0x830) via the GHCB MSR protocol. That's what the current code does. But on the KVM
  version we tested, KVM rejects this WRMSR with error 2 via the GHCB MSR protocol.

  This leaves no working IPI path on KVM without RESTR_INJ — the original motivation for this patch. Sounds like the issue should be raised on the KVM side: should KVM be accepting x2APIC ICR MSR writes via
  GHCB when RESTR_INJ is not available.

PR #1050, opened 2026-05-04
github/workflows: Use the test-in-svsm script directly

n-ramacciotti:
The `test-qemu-nocc-svsm` script uses `grep` to
find a success string and then terminates.

This was necessary because SVSM looped without
doing anything after booting, but now it exits
on both success and failure during test in svsm.

Use the exit value of the `test-in-svsm` script
to determine the success of the nocc test.

These changes also make it possible to remove the entire
`test-qemu-nocc-svsm.sh` script and use `test-in-svsm.sh`
directly in the CI, as the former becomes only a wrapper
for `test-in-svsm`


stefano-garzarella:
IMO it is better to keep also the `grep` just to be sure that tests really run. For example, we can have a bug where QEMU shutdown for some reason returning 0, but tests never run.

Maybe we can now check both.

n-ramacciotti:
> IMO it is better to keep also the `grep` just to be sure that tests really run. For example, we can have a bug where QEMU shutdown for some reason returning 0, but tests never run.

IIUC, `test-in-svsm.sh` only returns 0 when SVSM writes 0x21 (actually 0x10, `QEMUExitValue::Success`), which can only  happen when testing inside svsm. Otherwise QEMU returns another value and `test-in-svsm` will return with exit value 1

stefano-garzarella:
> These changes also make it possible to remove the entire `test-qemu-nocc-svsm.sh` script and use `test-in-svsm.sh` directly in the CI, as the former becomes only a wrapper for `test-in-svsm`

This is a good point for this change, why not doing this also in the PR?

n-ramacciotti:
> > These changes also make it possible to remove the entire `test-qemu-nocc-svsm.sh` script and use `test-in-svsm.sh` directly in the CI, as the former becomes only a wrapper for `test-in-svsm`
> 
> This is a good point for this change, why not doing this also in the PR?

I wanted first to have another opinion on that as maybe it could have made sense to keep the script around instead of deleting it.

stefano-garzarella:
> > > These changes also make it possible to remove the entire `test-qemu-nocc-svsm.sh` script and use `test-in-svsm.sh` directly in the CI, as the former becomes only a wrapper for `test-in-svsm`
> > 
> > 
> > This is a good point for this change, why not doing this also in the PR?
> 
> I wanted first to have another opinion on that as maybe it could have made sense to keep the script around instead of deleting it.

If we want to delete that scripts, I think this change makes sense, otherwise I'll keep the `grep`.

n-ramacciotti:
> If we want to delete that scripts, I think this change makes sense, otherwise I'll keep the `grep`.

Personally, I don't use often that script when testing locally, but I imagine it could be useful for someone else.

> > IMO it is better to keep also the `grep` just to be sure that tests really run. For example, we can have a bug where QEMU shutdown for some reason returning 0, but tests never run.
> 
> IIUC, `test-in-svsm.sh` only returns 0 when SVSM writes 0x21 (actually 0x10, `QEMUExitValue::Success`), which can only happen when testing inside svsm. Otherwise QEMU returns another value and `test-in-svsm` will return with exit value 1

For the same reason (some bugs) that QEMU can return 0, it could also return 0x21. So I suppose it makes sense to keep `grep`.


I can close this PR, keeping the current behaviour.

stefano-garzarella:
IIRC @osteffenrh added `test-qemu-nocc-svsm.sh` because initially we just checked the boot, without running tests. Now that we run tests IMO is better to remove that scripts and uses `test-in-svsm.sh` everywhere.

@osteffenrh WDYT? Or there are other reasons to keep it?

stefano-garzarella:
Maybe the only reason is to keep the timeout handling, but we can either integrate it in `test-in-svsm.sh` or use https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idtimeout-minutes in the CI.

n-ramacciotti:
> Maybe the only reason is to keep the timeout handling, but we can either integrate it in `test-in-svsm.sh` or use https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idtimeout-minutes in the CI.

We can use the one at the step level as other steps might require some time (like building QEMU or SVSM) and then delete `test-qemu-nocc`. https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepstimeout-minutes

> @osteffenrh WDYT? Or there are other reasons to keep it?

Before doing any change, I'll wait.

osteffenrh:
IMO we should keep timeout handling in the script and not rely on the github workflow alone. That way it is possible to run the test script locally with the exact behavior. Easier for testing and higher re-usability.

I am not 100% sure QEMU exit code 0 alone is a good marker for a successful test. I don't have a clear example right now, but there are probably many cases that QEMU considers a "clean exit" which are not test PASS. We can for sure check for a specific code that SVSM returns though. With that, we would not need the grep anymore.

n-ramacciotti:
> We can for sure check for a specific code that SVSM returns though. With that, we would not need the grep anymore.

I think this is the current behaviour of `test-in-svsm.sh`. When all tests pass, SVSM writes 0x10 to the QEMU exit port. This results in QEMU exiting with 0x21, which is then checked by `test-in-svsm.sh`

n-ramacciotti:
v2:
- Used `test-in-svsm.sh` in CI
- Removed `test-qemu-nocc-svsm.sh`
- Moved timeout handling in `test-in-svsm.sh`

@stefano-garzarella @osteffenrh

scripts/test-in-svsm.sh
```diff
@@ -106,17 +106,22 @@ done
...
 # This results in QEMU returning 0x21 ((0x10 << 1) | 1)
 if [[ $svsm_exit_code -eq 0x21 ]]; then
     echo "All tests passed"
     exit_value=0
+elif [[ $svsm_exit_code -eq 124 ]]; then
+    echo "Test Failed: timeout"
+    exit_value=1
 else
     echo "Test Failed"
     exit_value=1
```
osteffenrh, line 157: Sorry to bother you again, and this is not actually directly related to your work, so feel free to say no.

I would like to print the svsm_exit_code in the "fail" case to debugging. But the way that variable is used is odd: it is uninitialized if QEMU exits with 0, which we need to treat as an error. The way the exit code is stored does not capture the "0" case.

Can we rework that? (aka capture all exit codes and print them in the fail case)?

n-ramacciotti: > it is uninitialized if QEMU exits with 0, which we need to treat as an error

IIUC, this is the only way that `svsm_exit_code` will not be initialized. What about adding `svsm_exit_code=0` before launching QEMU  (`launch_guest.sh`)?

n-ramacciotti:
v3:
- Print exit code when  failing
- Add zero init of `svsm_exit_code` to cover QEMU exit code 0

osteffenrh (approved):
thanks @n-ramacciotti 👍

scripts/test-in-svsm.sh
```diff
@@ -108,17 +108,22 @@ TEST_IO_PID=$!
 
 svsm_exit_code=0
 
-$SCRIPT_DIR/launch_guest.sh --igvm $SCRIPT_DIR/../bin/coconut-test-qemu.igvm \
-    --state "$TEST_DIR/svsm_state.raw" \
-    --vsock "$VSOCK_CID" \
-    --unit-tests $TEST_DIR/pipe \
-    $LAUNCH_GUEST_ARGS "$@" || svsm_exit_code=$?
+timeout --foreground 180s \
```
stefano-garzarella: Can we add a parameter to enable the timeout handling?

IMO the default behavior should be like before (no timeout), but we can add a parameter, to be used in the CI or internal tests, to enable the timeout handling.

This can be useful for example, if there is a deadlock in a test, I can potentially attach to the VM to understand what is going on.

stefano-garzarella: And if we add a parameter, maybe the time (180s) could be specified via that parameter instead of having it hardcoded in the script.

scripts/test-in-svsm.sh
```diff
@@ -96,6 +96,9 @@ mkfifo $TEST_DIR/pipe.out
 dd if=/dev/urandom of="$TEST_DIR/svsm_state.raw" bs=512 count=1024
 
 LAUNCH_GUEST_ARGS=""
+TIMEOUT_ENABLED=0
+TIMEOUT_SECONDS=""
+MAX_TIMEOUT_SECONDS=600
```
osteffenrh: Why do we need a maximum for the timeout?

scripts/test-in-svsm.sh
```diff
@@ -96,6 +96,9 @@ mkfifo $TEST_DIR/pipe.out
 dd if=/dev/urandom of="$TEST_DIR/svsm_state.raw" bs=512 count=1024
 
 LAUNCH_GUEST_ARGS=""
+TIMEOUT_ENABLED=0
```
osteffenrh: Do we need an extra variable for the enable-state tracking? `TIMEOUT_SECONDS` could just be empty or undef to mark "disabled"?

scripts/test-in-svsm.sh
```diff
@@ -124,22 +141,35 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
+if [[ $TIMEOUT_ENABLED -eq 1 ]]; then
+    echo "Running tests with timeout: $TIMEOUT_SECONDS"
+    TIMEOUT_CMD="timeout --foreground $TIMEOUT_SECONDS"
+else
+    echo "Running tests without timeout"
+    TIMEOUT_CMD=""
+fi
```
stefano-garzarella: Let's avoid to change the behaviour without timeout, we don't need to print this, and you can remove entirely the else, initializing TIMEOUT_CMD="" on top with the other variables.

scripts/test-in-svsm.sh
```diff
@@ -113,6 +116,20 @@ while [[ $# -gt 0 ]]; do
...
+      TIMEOUT_ENABLED=1
+      TIMEOUT_SECONDS="$2"
+      if ! [[ "$TIMEOUT_SECONDS" =~ ^[0-9]+$ ]]; then
+        echo "Error: $TIMEOUT_SECONDS is not a valid number of seconds"
+        exit 1
+      fi
+      if [[ "$TIMEOUT_SECONDS" -gt "$MAX_TIMEOUT_SECONDS" ]]; then
+        echo "Error: $TIMEOUT_SECONDS exceeds max ($MAX_TIMEOUT_SECONDS)"
+        exit 1
+      fi
```
stefano-garzarella: Why we need to check a maximum? (should be explained also in the commit)

IMO we can avoid this check and let `timeout` handle it.

scripts/test-in-svsm.sh
```diff
@@ -113,6 +116,20 @@ while [[ $# -gt 0 ]]; do
       shift
       shift
       ;;
+    --timeout)
+      TIMEOUT_ENABLED=1
+      TIMEOUT_SECONDS="$2"
+      if ! [[ "$TIMEOUT_SECONDS" =~ ^[0-9]+$ ]]; then
```
stefano-garzarella: Ditto, `timeout` will already do it, no? Why overcomplicating this script?

scripts/test-in-svsm.sh
```diff
@@ -124,22 +141,35 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
+if [[ $TIMEOUT_ENABLED -eq 1 ]]; then
```
stefano-garzarella: I think you can even remove the `TIMEOUT_ENABLED` variable too, you just need `TIMEOUT_CMD`. You can initialize this empty, or set it when `--timeout` is specified. You can check if it's empty or not when needed (e.g. to check the return code)

.github/workflows/qemu.yml
```diff
@@ -146,5 +146,8 @@ jobs:
           PR_NUM=${{ github.event.pull_request.number }}
           VSOCK_PORT=$(( ${PR_NUM:-0} + 1024 ))
           VSOCK_CID=$(( ${PR_NUM:-0} + 10 ))
-          QEMU=./citests/qemu/bin/qemu-system-x86_64 scripts/test-qemu-nocc-svsm.sh --vsock-port $VSOCK_PORT --vsock-cid $VSOCK_CID
+          QEMU=./citests/qemu/bin/qemu-system-x86_64 scripts/test-in-svsm.sh \
+            --vsock-port $VSOCK_PORT --vsock-cid $VSOCK_CID \
+            --nocc --timeout 180 \
+            </dev/null 2>&1
```
stefano-garzarella, line 152: why we are adding the redirect ?

scripts/test-in-svsm.sh
n-ramacciotti: > `TIMEOUT_SECONDS` could just be empty or undef to mark "disabled"?

right, thanks!

stefano-garzarella: see https://github.com/coconut-svsm/svsm/pull/1050#discussion_r3208051863, IMO you can remove all them and just use TIMEOUT_CMD

scripts/test-in-svsm.sh
n-ramacciotti: I wanted to set an upper bound to the waiting time, but yes it's actually not useful.

scripts/test-in-svsm.sh
n-ramacciotti: Right, thanks!

scripts/test-in-svsm.sh
n-ramacciotti: This was introduce for the second if (max timeout), but yes `timeout` will do it on its own. I will remove it

.github/workflows/qemu.yml
n-ramacciotti: It was present in `test-qemu-nocc-svsm.sh`, so I added it also here

stefano-garzarella: ah ok, @osteffenrh do you remember why?

In any case, please mention this change in the commit message.

n-ramacciotti:
update:
- Addressed comments (thanks for the good suggestions!)

stefano-garzarella (approved):

.github/workflows/qemu.yml
osteffenrh: I think that can go if the CI works without it.

osteffenrh:
I have the issue that a QEMU process stays behind in case of a timeout.

To make sure we can trigger a timeout:
```diff
diff --git a/kernel/src/testing.rs b/kernel/src/testing.rs
index 106e0642..5004c0ef 100644
--- a/kernel/src/testing.rs
+++ b/kernel/src/testing.rs
@@ -88,7 +88,7 @@ pub fn svsm_test_runner(test_cases: &[&test::TestDescAndFn]) {
 
     info!("All tests passed!");
 
-    exit(QEMUExitValue::Success);
+    //exit(QEMUExitValue::Success);
 }
 
 /// Exits the in-SVSM test session with the given exit value.
```

then
```bash
scripts/test-in-svsm.sh --nocc --vsock-port 1025 --vsock-cid 10 --timeout 1
```

Says:
```
...
[SVSM] test vsock::stream::tests::test_virtio_vsock_read ...
[SVSM] test vsock::stream::tests::test_virtio_vsock_write ...
[SVSM] All tests passed!
Test Failed: timeout
```
Ok.

But:
```
❯ pgrep -a qem
639217 /home/osteffen/work/coconut/main/out/qemu/bin/qemu-system-x86_64 -cpu max,smep=on -machine q35,confidential-guest-support=cgs0,memory-backend=mem0,igvm-cfg=igvm0,accel=tcg,x-svsm-virtio-mmio=on -object memory-backend-memfd,size=8G,id=mem0,share=true,prealloc=false,reserve=false -object igvm-cfg,id=igvm0,file=/home/osteffen/work/coconut/main/svsm/scripts/../bin/coconut-test-qemu.igvm -object nocc,id=cgs0 -smp 4 -no-reboot -netdev user,id=vmnic -device e1000,netdev=vmnic,romfile= -nographic -vga none -monitor none -serial stdio -serial null -serial null -chardev pipe,id=test,path=/tmp/tmp.f6oyflLuAL/pipe -serial chardev:test -device isa-debug-exit,iobase=0xf4,iosize=0x04 -device pc-testdev -global virtio-mmio.force-legacy=false -drive file=/tmp/tmp.f6oyflLuAL/svsm_state.raw,format=raw,if=none,id=svsm_storage,cache=none -device virtio-blk-device,drive=svsm_storage -device vhost-vsock-device,guest-cid=10
```

This might be related to qemu running as a sub-process of `launch_guest.sh`

This seems to make it work:
```diff
diff --git a/scripts/launch_guest.sh b/scripts/launch_guest.sh
index 591bc6c1..b5eb0c75 100755
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -179,8 +179,16 @@ if [ -t 0 ]; then
   stty intr ^]
 fi
 
+# In test mode (--unit-tests), use exec so QEMU replaces the shell process.
+EXEC_CMD=""
+if [ -n "$QEMU_EXIT_DEVICE" ]; then
+    EXEC_CMD="exec"
+fi
+
 # Temporarily use -vga none to avoid IGVM VGA init failure in QEMU 10.1
-$SUDO_CMD \
+$EXEC_CMD $SUDO_CMD \
   "$QEMU" \
     -cpu $CPU \
     -machine $MACHINE,$VIRTIO_ENABLE \
```

n-ramacciotti:
Good catch! it also seems to happen in `test-qemu-nocc-svsm.sh`.

Adding `exec` seems to make it work, but there is a problem
about restoring the original interrupt from `ctrl-]` to `ctrl-C`
when launching  `test-in-svsm.sh` directly, as it will remain
`ctrl-]` after execution.

Is it sufficient to reset that option in `test-in-svsm.sh`? 
There would also be the problem if someone used
`launch_guest.sh` directly with `--unit-tests`, but I don't know
how often this is done without `test-in-svsm.sh`

osteffenrh:
yeah, exactly, the restore would be missing.

But in the "test-in-svsm" case, where we use the timeout, we do not need the remapping of Ctrl+c, since this is not meant to be an interactive experience. We could skip changing the termnial settings when running launch_guest.sh in "testing" mode?

stefano-garzarella:
IIUC in `launch_guest.sh` we do the SIGINT remapping only when there is a tty attached, so it should work as is in the CI, but if we want to have it working also locally, can we just avoid that remapping too when it is started from `test-in-svsm.sh` instead of using `exec`?

stefano-garzarella:
aaaa, I think that behaviour is related to:
```
       -f, --foreground
              when  not  running timeout directly from a shell prompt, allow COMMAND to read from the TTY and
              get TTY signals; in this mode, children of COMMAND will not be timed out
```

stefano-garzarella:
```diff
diff --git a/scripts/test-in-svsm.sh b/scripts/test-in-svsm.sh
index 19f1dd14..bc20afef 100755
--- a/scripts/test-in-svsm.sh
+++ b/scripts/test-in-svsm.sh
@@ -116,7 +116,7 @@ while [[ $# -gt 0 ]]; do
       ;;
     --timeout)
       echo "Running tests with timeout: $2"
-      TIMEOUT_CMD="timeout --foreground $2"
+      TIMEOUT_CMD="timeout $2"
       shift
       shift
       ;;
@@ -142,7 +142,7 @@ $TIMEOUT_CMD \
         --state "$TEST_DIR/svsm_state.raw" \
         --vsock "$VSOCK_CID" \
         --unit-tests $TEST_DIR/pipe \
-        $LAUNCH_GUEST_ARGS "$@" || svsm_exit_code=$?
+        $LAUNCH_GUEST_ARGS "$@" < /dev/null || svsm_exit_code=$?
 
 # SVSM writes 0x10 to the QEMU exit port when all tests passed.
 # This results in QEMU returning 0x21 ((0x10 << 1) | 1)
```

@osteffenrh can you try if just this fix your issue ? (without the changes in `scripts/launch_guest.sh` )

edit: ` < /dev/null` is needed otherwise `qemu` hangs waiting stdin that is not available since `timeout` is now running it in background

osteffenrh:
Can do. -> https://github.com/coconut-svsm/svsm/issues/1056

osteffenrh:
We need the `--foreground` option AFAIK. Let me dig in and fix that. See the issue I  opened.

n-ramacciotti:
> We could skip changing the termnial settings when running launch_guest.sh in "testing" mode?

That makes more sense than what I was proposing (resetting it in `test-in-svsm.sh`)

> We need the `--foreground` option AFAIK

I introduced foreground to run `test-in-svsm` locally and being able to interrupt it even before timeout. I'm not sure about other ways to keep the interrupt

> edit:  < /dev/null is needed otherwise qemu hangs waiting stdin that is not available since timeout is now running it in background

I think `launch-guest.sh` stops before launching qemu here (tested with some `echo` before and after):
```
  stty intr ^]
 ```

stefano-garzarella:
@n-ramacciotti please continue the discussion in #1056, this PR is now closed, so there is not a lot of visibility

n-ramacciotti:
Yes, sorry. I will

PR #1051, opened 2026-05-04
task: fix conflict in per-CPU mapping update

msft-jlange:
When a task is moving from one processor to another, it is not sufficient to wait until its stack is no longer consumed before scheduling it.  Remapping of the current per-CPU virtual address space must also wait until the new task has stopped running on its previous processor.

### DIFF
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -478,6 +478,13 @@ unsafe fn switch_to(prev: *const Task, next: *const Task) {
     // the page table and stack information in those tasks are correct and
     // can be used to switch to the correct page table and execution stack.
     unsafe {
+        // Before manipulating the new task, wait until it has stopped running,
+        // and mark it as active again.  It is not safe to change per-CPU
+        // mappings or stack pointers until the task is exclusively claimed
+        // by the current CPU.
+        (*next).make_active();
+        (*next).update_cpu();
+
         let cr3 = (*next).page_table.lock().cr3_value().bits() as u64;
 
         // The location of a cpu-local stack that's mapped into every set of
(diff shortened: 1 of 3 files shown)


kernel/src/task/tasks.rs
```diff
@@ -227,9 +229,9 @@ impl TaskSchedState {
         self
     }
 
-    fn set_active(&self) {
-        if self.active.swap(true, Ordering::Release) {
-            panic!("attempted switch to an active task");
+    fn make_active(&self) {
+        while self.active.swap(true, Ordering::Release) {
```
joergroedel: The use of `Ordering::Release` deserves a comment. I think it is fine because the `Acquire` happens in assembly and the ordering is guaranteed by the x86 architecture, but that is not obvious from a first glance. Intuitively I had chosen `Ordering::AcqRel` to be safe.

joergroedel (commented):
I left a note where a code comment is needed, otherwise this looks good to me. My testing also passed.

kernel/src/task/tasks.rs
msft-jlange: Actually, `Ordering::Acquire` is what's needed here, which I think makes sense intuitively, because the write is giving the current thread the right to operate on the task, i.e. to acquire synchronization.  There is no need for release semantics here because previous stores don't have to drain before the swap but subsequent loads must wait for the swap to complete.  I will change this to `Ordering::Acquire`.  Do you still think this requires a comment given that it is intuitively consistent?

msft-jlange: As you pointed out to me separately, this has to be `Ordering::AcqRel` because Rust requires this for `swap` operations (or any operation that can do both a load and store).

kernel/src/task/schedule.rs
```diff
@@ -467,6 +478,13 @@ unsafe fn switch_to(prev: *const Task, next: *const Task) {
     // the page table and stack information in those tasks are correct and
     // can be used to switch to the correct page table and execution stack.
     unsafe {
+        // Before manipulating the new task, wait until it has stopped running,
+        // and mark it as active again.  It is not safe to change per-CPU
+        // mappings or stack pointers until the task is exclusively claimed
+        // by the current CPU.
+        (*next).make_active();
+        (*next).update_cpu();
```
joergroedel, line 486: What happens here if CPU A runs task T1 and CPU B runs task T2 and both CPUs decide at the same time that their tasks need to be migrated to the other CPU? It looks like both CPUs will wait for the incoming task to become inactive, which will then never happen This looks like a possible deadlock to me, am I right?

joergroedel (commented):
I think there is a possible deadlock here, can you please have a look?

msft-jlange:
Fixing this problem will require a different approach, so I will open a new PR once the new code is ready

Issue #532, opened 2024-11-23
split stage2 and svsm kernel into two crates


Freax13:
As of today, the stage2 and svsm targets are two binary targets of the same crate and share most of their code through a common lib.rs. This has various downsides:
1. stage2 is unnecessarily bloated.
2. We can't have stage2 or svsm-specific APIs in the shared code.
3. publicly exposing all modules in lib.rs hides a lot of dead-code warnings.

Instead, we should split the code into at least three crates, one for the stage2 target, one for the svsm target, and one for shared code.

There was some discussion on this in one of the [meetings](https://github.com/coconut-svsm/governance/blob/main/Meetings/devel-call-2024-09-25.md#stage-2-split).

Freax13:
I worked on this for a while, but unfortunately, I won't be able to finish my work [^1]. You can find my work-in-progress code [here](https://github.com/Freax13/svsm/tree/enhancement/split-kernel-stage2).

In the hopes that maybe someone else will be able to pick this up and finish it, here's what I did so far:
I started with one commit that [copies all code](https://github.com/Freax13/svsm/commit/9a246b171815cdda246fc3990af1e34ec7e52daf) from the svsm directories into a new stage2 directory. The goal behind this was to make it easy to update when rebasing without missing any new changes. Following this, there are a couple commits that fix build errors introduced by just doing a raw copy in the first commit.
Those first couple of commits don't build, but everything after that starting with https://github.com/Freax13/svsm/commit/0dbd9707874e9bbcbe1a6bcd29598a8dd273dd53 should build and pass the tests. There have been a few instances where some commits broke tests after a rebase, so I'd recommend keeping this structure until everything's done.
Most of the the remaining commits simply remove dead code, but some of them also rewrite existing code e.g. there's a commit replacing the very complex heap allocator and its SVSM-specific features with a simple bump allocator in stage2.

[^1]: Previously, my employer was paying me to work on the COCONUT SVSM project, but I started a new position at a new employer last week and my new role doesn't involve SVSM development.

msft-jlange:
Resolved by #1053.

PR #1053, opened 2026-05-05
stage2: remove stage2

msft-jlange:
With the addition of a new, simple boot loader, support for the old stage2 can be removed.  This eliminates the dual build of the svsm crate, so the kernel crate is now used for the kernel only.  This resolves issue #532 by eliminating stage2 altogether.


joergroedel (approved):
All testing passed.

PR #1058, opened 2026-05-11
scripts/launch_guest: drop nocc object for QEMU >= 11.0

luigix25:
QEMU 11.0 and later no longer require the nocc object or confidential-guest-support for the --nocc path. Skip both when running on QEMU >= 11.0.

QEMU 10.x still requires the nocc object, so keep it for backwards compatibility.

This is preliminary work for https://github.com/coconut-svsm/qemu/pull/32

### DIFF
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -129,9 +129,12 @@ fi
 
 case "$CGS" in
   nocc)
-    SNP_GUEST="-object nocc,id=cgs0"
+    SNP_GUEST=""
     CPU=max,smep=on
-    ACCEL=tcg
+    if (( QEMU_MAJOR < 11 )); then
+      ACCEL=tcg
+      SNP_GUEST="-object nocc,id=cgs0"
+    fi
     ;;
   sev)
     SNP_GUEST="-object sev-snp-guest,id=cgs0,cbitpos=$C_BIT_POS,reduced-phys-bits=$REDUCED_PHYS_BITS"
@@ -140,7 +143,10 @@ case "$CGS" in
     echo "Error: Unexpected CGS value '$CGS'"
     exit 1
 esac
-MACHINE=q35,confidential-guest-support=cgs0,memory-backend=mem0,igvm-cfg=igvm0,accel=$ACCEL
+MACHINE=q35,memory-backend=mem0,igvm-cfg=igvm0,accel=$ACCEL
+[ -n "$SNP_GUEST" ] && MACHINE+=",confidential-guest-support=cgs0"
+[ -n "$VIRTIO_ENABLE" ] && MACHINE+=",${VIRTIO_ENABLE}"
+
 MEMORY=memory-backend-memfd,size=8G,id=mem0,share=true,prealloc=false,reserve=false
 IGVM_OBJ="-object igvm-cfg,id=igvm0,file=$IGVM"
 
@@ -183,7 +189,7 @@ fi
 $SUDO_CMD \
   "$QEMU" \
     -cpu $CPU \
-    -machine $MACHINE,$VIRTIO_ENABLE \
+    -machine $MACHINE \
     -object $MEMORY \
     $IGVM_OBJ \
     $SNP_GUEST \
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -93,6 +93,10 @@ while [[ $# -gt 0 ]]; do
       shift
       shift
       ;;
+    --tcg)
+      ACCEL=tcg
+      shift
+      ;;
     --)
       shift
       break


scripts/launch_guest.sh
```diff
@@ -129,9 +129,13 @@ fi
 
 case "$CGS" in
   nocc)
-    SNP_GUEST="-object nocc,id=cgs0"
     CPU=max,smep=on
     ACCEL=tcg
+    if (( QEMU_MAJOR >= 11 )); then
+      SNP_GUEST=""
+    else
```
stefano-garzarella: Let's declare it empty in any case and add just an if block for old QEMU

scripts/launch_guest.sh
```diff
@@ -140,7 +144,13 @@ case "$CGS" in
     echo "Error: Unexpected CGS value '$CGS'"
     exit 1
 esac
-MACHINE=q35,confidential-guest-support=cgs0,memory-backend=mem0,igvm-cfg=igvm0,accel=$ACCEL
+
+if [ -n "$SNP_GUEST" ]; then
+  MACHINE=q35,confidential-guest-support=cgs0,memory-backend=mem0,igvm-cfg=igvm0,accel=$ACCEL
```
stefano-garzarella: IIUC the only difference is `confidential-guest-support=cgs0`, let's do something similar of what we did with `$VIRTIO_ENABLE` and set it in the `case` before, so it's more clear which case requires it.

scripts/launch_guest.sh
```diff
@@ -129,9 +129,13 @@ fi
 
 case "$CGS" in
   nocc)
-    SNP_GUEST="-object nocc,id=cgs0"
     CPU=max,smep=on
     ACCEL=tcg
```
stefano-garzarella: IIUC from @kraxel now we can use `kvm` as well, what about using it by default and add an option to enable tcg for some use cases (e.g. if the CI can't use KVM, but I'm not sure if the CI can't use KVM)

luigix25: yep, kvm works ( and so does the tcg ). I'll check if we can use kvm in the CI.

luigix25: kvm works in the CI!

stefano-garzarella: @luigix25 thanks for checking, so at that point, let's use kvm as default but up to you if you want to add a parameter to enable tcg. Maybe we can avoid the parameter, but we can leave the ACCEL variable, so if some one really want to use tcg, they can easily edit the script. (not a strong opinion on that, really as you prefer)

luigix25:
- Added a `tcg` option to launch_guest
- Applied @stefano-garzarella suggestions

scripts/launch_guest.sh
```diff
@@ -183,7 +189,7 @@ fi
 $SUDO_CMD \
   "$QEMU" \
     -cpu $CPU \
-    -machine $MACHINE,$VIRTIO_ENABLE \
+    -machine $MACHINE \
```
stefano-garzarella, line 196:
```
avoiding stray
commas (e.g. "accel=tcg,,") that QEMU would reject.
```

mm, so it this a pre-existing issue? If yes, please fix in a separate commit (strange we didn't see it)

scripts/launch_guest.sh
```diff
@@ -93,6 +93,11 @@ while [[ $# -gt 0 ]]; do
       shift
       shift
       ;;
+    --tcg)
+      CPU=max,smep=on
```
stefano-garzarella: Why we need `CPU=max,smep=on` for tcg ?
Is this true also for QEMU 11.0?

We should mention this in the commit message.

scripts/launch_guest.sh
stefano-garzarella: I just tried `./scripts/launch_guest.sh --nocc` on main and QEMU starts without issues with:
```
qemu-system-x86_64 -cpu max,smep=on -machine q35,confidential-guest-support=cgs0,memory-backend=mem0,igvm-cfg=igvm0,accel=tcg, -object memory-backend-memfd...
```

So it seems QEMU accepts it in some way, but I agree that is better what you are doing.

luigix25: yes and no. I _think_ QEMU can tolerate 1 extra comma, but not two.

stefano-garzarella: okay, so it's fine to keep as it is, maybe mention this in the commit message.

scripts/launch_guest.sh
luigix25: good question. for some reason the CPU model `EPYC-v4` lacks some features that are required by svsm to boot.

This applies to QEMU 11.0 as well.

stefano-garzarella: Okay, I'm not asking to put all the details ;-)
It's fine just to mention that is needed for TCG since `EPYC-v4` lacks something, etc. so we will remember that we did it on purpose.

luigix25:
- improved commit messages [@stefano-garzarella]

stefano-garzarella (dismissed):

scripts/launch_guest.sh
stefano-garzarella, line 164: Partially related: I just noticed that `-no-cc` with qemu 11 is now asking for sudo password also if in my case I don't need sudo to use `/dev/kvm`.

What about removing this and ask the caller to use sudo or not. IMHO is not elegant to use sudo in a script, but I know this is pre-existing, so I don't want to break user flows.

WDYT?

CCing @osteffenrh @joergroedel

luigix25: it asked for sudo password for no-cc until 68c8c650 ("scripts/launch_guest: avoid sudo when using TCG acceleration"). So it's relatively new. But yes, I agree that is not nice to always ask for sudo password.

osteffenrh: IMO the whole sudo thing is not needed. One can adjust permissions on `/dev/sev` the same way `/dev/kvm` uses. Running as root all the time just complicates things.

I would say lets drop that completely and mention in the readme how to ser permissions for the sev device correctly.

stefano-garzarella: @osteffenrh 100% agree, and if the user wants to use `sudo` because don't want to change permissions, it is up to them to call the script with `sudo`

luigix25: yep agree! I guess it's material for a different PR?

luigix25:
Using cpu model `EPYC-v4` using kvm in nocc mode, shows some warning from QEMU about missing CPU features.

So I reverted back to `max` model.

@stefano-garzarella PTAL

scripts/launch_guest.sh
stefano-garzarella: > yep agree! I guess it's material for a different PR?

yep, I think so. Let me know if you plan to work on that.

stefano-garzarella (approved):

scripts/launch_guest.sh
luigix25: yep, I can work on that

PR #1059, opened 2026-05-11
mm/alloc.verus: update phys_to_virt proof after stage2 removal

ziqiaozhou:
Fix broken proof due to changes made in #1053.

Adjust the allocation proof to match the updated phys_to_virt behavior from the stage2 removal. Since phys_to_virt no longer treats the stage2 identity mapping as a special case, the spec can be simplified by removing that exception.


stefano-garzarella (approved):
LGTM, but I'd like to wait Jon's opinion before merging.

@msft-jlange do you want to take a look since this is related to your work?

joergroedel (approved):

PR #1061, opened 2026-05-12
bldr: clean up SIPI logic

msft-jlange:
The transition to a single, simple boot loader provides some opportunities to clean up the rest of the boot logic.  Because the boot loader is so small, it can be moved to a low portion of memory (below 1 MB), which makes it possible to embed the SIPI startup stub in the boot loader itself instead of requiring an extremely awkward array of hand-patched assembly code to be embedded in the image builder.  This also provides an opportunity to unify all of the boot paths, between SIPI, SNP 32-bit startup, and TDX 32-bit startup.


boot/bldr/src/main.rs
```diff
@@ -21,17 +22,68 @@ use core::slice;
...
+         * 32-bit mode, and adjust the target address back by two bytes to
+         * account for the difference in instruction size that is used when
+         * calculating the displacement. */
+    startup_32_bsp:
+        jmp startup_bsp - 2
+        .word 0
+
+    startup_32_ap:
+        jmp startup_ap_no_gdt - 2
+        .word 0
```
peterfang: Can we just put these in the `.code32` area for simplicity?

boot/bldr/src/main.rs
```diff
@@ -21,17 +22,68 @@ use core::slice;
...
         .section .text
-        .section .startup.text,"ax"
+        .section .startup16.text,"ax"
+        .code16gcc
+
+        .globl base_entry
+    base_entry:
+        jmp startup_16
+        jmp startup_32_bsp
+        jmp startup_32_ap
```
peterfang, line 48: These are executed in 32-bit mode as well, so do they have the same displacement issue as the code below?

boot/bldr/src/main.rs
```diff
@@ -21,17 +22,68 @@ use core::slice;
...
+    gdt_start:
+        .quad	0
+        .quad	0x00cf9b000000ffff // flat 32-bit code segment
+        .quad	0x00cf93000000ffff // flat data segment
+        .quad	0x00af9b000000ffff // 64-bit code segment
+    gdt_end:
+
+    startup_16:
+        movl %cr0, %eax
+        orl $1, %eax // enable protected mode
```
peterfang: You can use a const that translates to `CR0Flags::PE.bits()` here

boot/bldr/src/main.rs
```diff
@@ -110,22 +167,71 @@ global_asm!(
...
+        wrmsr
+    5:
+        /* Enable paging using the transition page table */
+        movl transition_pt_root, %eax
+        movl %eax, %cr3
+        movl %cr4, %eax
+        orl ${CR4_PAE}, %eax
+        movl %eax, %cr4
+        movl %cr0, %eax
+        bts $31, %eax
```
peterfang: You can use `PG_SHIFT` here

boot/bldr/src/main.rs
```diff
@@ -42,6 +94,11 @@ global_asm!(
          * in EAX upon entry. */
         movl %eax, {PLATFORM}(%ebp)
 
+        /* Capture the address of the AP start context in case it is needed
+         * later. */
+        movl {AP_CTXT_ADDR}(%ebp), %edx
+        movl %edx, ap_ctxt_ptr
```
peterfang, line 93: Why not just capture `AP_INITIAL_RIP(ap_ctxt_ptr)` here?

tools/igvmbuilder/src/context.rs
```diff
@@ -386,6 +386,9 @@ pub fn construct_stage1_image(
         }
     }
 
+    // The AP entry point is two bytes beyond the standard entry point.
+    tdp_context.ap_entry = tdp_context.rip + 2;
```
peterfang, line 393: Can we point to the assembly in `boot/bldr/src/main.rs` in the comment? I think this might be very difficult to understand for future readers

tools/igvmbuilder/src/igvm_builder.rs
```diff
@@ -413,23 +413,34 @@ impl IgvmBuilder {
             // Populate the boot loader stack.
             let bldr_stack =
                 BootLoaderStack::new(&self.gpa_map, &boot_image_info, &init_page_table_info);
-            bldr_stack.add_directive(
-                self.gpa_map.bldr_stack.get_start(),
-                COMPATIBILITY_MASK.get(),
-                &mut self.directives,
-            );
+
+            // Add the contents of the stack page that incldues the launch
```
peterfang:
```suggestion
            // Add the contents of the stack page that includes the launch
```
Nit: typo

tools/igvmbuilder/src/igvm_builder.rs
```diff
@@ -413,23 +413,34 @@ impl IgvmBuilder {
...
+                    empty_stack_pages,
+                    COMPATIBILITY_MASK.get(),
+                    IgvmPageDataType::NORMAL,
+                )?;
+            }
 
             // Construct a native context object to capture the start context.
-            let start_rip = self.gpa_map.bldr_image.get_start();
-            let start_rsp = self.gpa_map.bldr_stack.get_end() - size_of::<BootLoaderStack>() as u64;
+            // The 32-bit entry point is two bytes into the image.
```
peterfang: Can we add more comments here also to help future readers?

boot/bldr/src/main.rs
msft-jlange: Yes.  I had been under the impression that `.code32` implied the existence of a different code section, which would cause problems here because the jumps from the 16-bit area to the 32-bit area must be encoded as short jumps, which is not possible when jumping across code segments.  However, `.code32` doesn't enforce a code section boundary, only a code generation directive, so this can be changed without breaking the assumptions.

boot/bldr/src/main.rs
msft-jlange: No, these are guaranteed to be short jumps and those are two bytes in every mode.

tools/igvmbuilder/src/igvm_builder.rs
msft-jlange: Can you give an example of what you'd like to see?  Is it enough to refer to the assembly code as you requested elsewhere?

boot/bldr/src/main.rs
msft-jlange: I don't understand the question.  At this point, the boot loader has no idea where the AP context structure will live - that's allocated by the IGVM builder in some other location.  The boot loader's initial parameters (which come from the bldr invocation stack frame) contain the address of the AP context structure that was allocated by the IGVM builder.  That address is stored in a global variable here because when it is needed again - at SIPI time - the boot loader's initial stack frame is gone, and the global variable is the only way to find the AP context structure.

This global variable is referenced from the SIPI path, which is invoked much later (during kernel execution, not during bldr execution).  The SIPI code needs to be able to look up the initial RIP that was specified by the kernel for the processor being started.  That code doesn't have access to the bldr initial stack, but it can see the global variable, so it reads the global to find out where the AP context structure was allocated, and then it can fetch the kernel-supplied RIP as the entry point for the AP.

peterfang: My question was not very clear. I was simply curious about why we don't just save `ap_start_context_addr + AP_INITIAL_RIP` in `ap_ctxt_ptr`. Looks like that's the only thing we need in `ApStartContext`

boot/bldr/src/main.rs
```diff
@@ -21,17 +22,57 @@ use core::slice;
...
+
         .code32
 
-        .globl startup_32
-        startup_32:
+    startup_32_bsp:
+        jmp startup_bsp - 2
+
+    startup_32_ap:
+        jmp startup_ap_no_gdt - 2
```
peterfang: Since they're now in the `.code32` area, I don't think the 2-byte adjustment is needed anymore
```suggestion
    startup_32_bsp:
        jmp startup_bsp

    startup_32_ap:
        jmp startup_ap_no_gdt
```

tools/igvmbuilder/src/igvm_builder.rs
peterfang: The updated version looks fine to me. Thanks for the documentation

boot/bldr/src/main.rs
msft-jlange: You are correct.

boot/bldr/src/main.rs
msft-jlange: > My question was not very clear. I was simply curious about why we don't just save `ap_start_context_addr + AP_INITIAL_RIP` in `ap_ctxt_ptr`. Looks like that's the only thing we need in `ApStartContext`

I'm less certain that this is the only field of `ApStartContext` that will ever be referenced from the AP start flow.  I don't see any particular advantage to performing the `AP_INITIAL_RIP` adjustment at the time the address is captured vs. the time it is used, so it seems advantageous to keep the door open for potential future use cases.

boot/bldr/src/main.rs
peterfang: The updated code seems to be in good shape now. My last comment is to maybe add a comment above `base_entry` to explain these 3 jumps. Basically, they must be short jumps (EB disp8) in order for this to work as expected. This implies these jumps have a limit: target_address <= jump_instruction_address + 2 + 127

joergroedel (dismissed):
Testing looks good.

peterfang (approved):

PR #1062, opened 2026-05-13
paging: refactor svsm kernel page table to creata a reusable pagetable crate

ziqiaozhou:
This PR factors the kernel’s page-table implementation into a new standalone paging crate, then updates the kernel to consume (and re-export) those generic paging/address/size/util building blocks. It also introduces Verus specification/proof scaffolding for alignment helpers and basic paging constants.

Changes:

- Added a new paging crate containing generic page-table types (GenericPageTable, PTEntry, PTPage, flags) and x86_64-specific flag/level definitions.
- Moved/re-exported common utilities and page size constants from kernel into paging, and updated kernel code to use those re-exports.
- Added Verus spec/proof modules for alignment helpers and page size facts, and wired paging/verus into kernel verification features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1583,6 +1583,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "paging"
+version = "0.1.0"
+dependencies = [
+ "bitflags",
+ "verify_external",
+ "verify_proof",
+ "verus_stub",
+ "vstd",
+ "zerocopy",
+]
+
+[[package]]
 name = "pastey"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2092,6 +2104,7 @@ dependencies = [
  "libtcgtpm",
  "log",
  "packit",
+ "paging",
  "release",
  "rustc-demangle",
  "rustc_version",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = [
     "boot/bootimg",
     "boot/defs",
     "cpuarch",
+    "paging",
     # binary targets
     "stage1",
     "kernel",
@@ -55,6 +56,7 @@ bldr = { path = "boot/bldr" }
 bootdefs = { path = "boot/defs" }
 bootimg = { path = "boot/bootimg" }
 cpuarch = { path = "cpuarch" }
+paging = { path = "paging" }
 test = { path = "test" }
 svsm = { path = "kernel" }
 elf = { path = "elf" }
@@ -139,6 +141,7 @@ missing_debug_implementations = { level = "deny", priority = 50 }
 single_use_lifetimes = { level = "warn", priority = 125 }
 trivial-numeric-casts = { level = "deny", priority = 10 }
 unsafe_op_in_unsafe_fn = { level = "deny", priority = 2 }
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(verus_only)'] }
 
 [workspace.lints.clippy]
 await_holding_lock = "warn"
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -16,6 +16,8 @@ doctest = true
 bootdefs.workspace = true
 bootimg.workspace = true
 cpuarch.workspace = true
+paging.workspace = true
+
 libaproxy = { workspace = true, optional = true }
 elf.workspace = true
 syscall.workspace = true
@@ -79,7 +81,7 @@ enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 vtpm = ["dep:libtcgtpm"]
 nosmep = []
 nosmap = []
-verus_all = ["verus_builtin", "verus_builtin_macros", "vstd", "verify_proof/verus", "verify_external/verus", "verus_stub/disable"]
+verus_all = ["verus_builtin", "verus_builtin_macros", "vstd", "verify_proof/verus", "verify_external/verus", "verus_stub/disable", "paging/verus"]
 verus = ["verus_all", "verify_proof/noverify", "verify_external/noverify"]
 noverify = []
 virtio-drivers = ["dep:virtio-drivers", "dep:safe-mmio"]
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -32,7 +32,7 @@ use crate::locking::{
     WriteLockGuard, WriteLockGuardIrqSafe,
 };
 use crate::mm::page_visibility::SharedBox;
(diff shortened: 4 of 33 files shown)


ziqiaozhou:
As I mentioned in last-week's dev meeting, I refactored the page table code to extract out the general part of the page table implementation from kernel/src/pagetable.rs, in order to 
1. make it more reusable; e.g., Litebox can reuse this paging crate if we make it general and verified.
2. distinguish the svsm-specific page table properties and general page table properties to simplify the verification efforts.
3. Once I make the general part verified, users of the crate gain more confidence to use our paging crate instead of other external page table implementations.

Goal of the PR: move general OS-independent and arch-independent page table logic into paging crate; keep the change as minimal as possible;
non-goal of the PR: make the paging crate perfect to fit all OS and arch; support different page granularities (4k vs 16k vs others supported by arm); support 32-bit page table.

Although I tried to keep implementation closer to its original, I still made some minor changes to deduplicate the code and make it easy to be used by non-svsm projects. 
- RawPageTablePart and PageTable share many logic in some similar functions (map/unmap_4k/2M, allocate_pte_xxx, etc.).  I merged them into a GenericPageTable struct that supports customized page levels. 
- Some free functions (e.g., supported_flags, private_pte_mask, etc.) are now become a trait associated functions  to ArchPagingMeta so that external users of the paging crate can pick their own way to enable arch features and does not need to init the static vars (e.g., PRIVATE_PTE_MASK or FEATURE_MASK).

ziqiaozhou:
@stefano-garzarella can you add verification label to this PR since I moved some verified code into that paging crate from svsm kernel crate? I was able to add that label but cannot do that since this week.
@joergroedel, I moved and deduplicated some code from your original pagetable.rs file. Please take a look to confirm their usage as still as before. I have checked all callers and confirmed that based on my understanding.

ziqiaozhou:
#1059 will fix the verification failure in this PR.

ziqiaozhou:
@joergroedel , svsm kernel picks flush_tlb_global_sync, flush_tlb_global_sync_range or flush_tlb_global_sync_page  based on the scope of the change, outside pagetable crate. I think I should move the flush_tlb_xxx out of the paging crate for split_4k to make the pagetable implementation consistant. 
NIT: split_4k is never used by any code inside svsm

ziqiaozhou:
Review tips: focus on changes in 2fc8a8e26ab0768e6c9be655a24cad48f0293eb5, 8da895ad9df2ec37d128b623943fd14ae0925abc, and f16e21e6c0bd457c8082dc78d9d35ee0b0ba626b. Other commits are just moving or copying files.

joergroedel (commented):
Overall looks great to me. I really like the new abstractions. A detail about the TLB flushing needs to be sorted out.

joergroedel (commented):
As I said before, the overall direction of the changes is great. The new crate looks a lot cleaner and is more generic than our previous implementation.

Two things are on my mind, though. I think I already mentioned the TLB-flushing question, but can not find the comment anymore, so I am putting it here again.

1. The page-table code itself should not flush the TLB itself on changes, that should be up to the caller. The caller usually has better knowledge of the flushing scope than the page-table code (CPU-set to flush on, global vs. non-global, PCID, ...). What the new paging crate should provide is tracking of which addresses or ranges need to be flushed.
2. COCONUT-SVSM has another abstraction in the page-table code: `PageTablePart`. This abstraction holds a sub-page-table which can be mapped into several top-level page-tables. Is there a plan to move that to the generic crate as well?

ziqiaozhou:
> As I said before, the overall direction of the changes is great. The new crate looks a lot cleaner and is more generic than our previous implementation.
> 
> Two things are on my mind, though. I think I already mentioned the TLB-flushing question, but can not find the comment anymore, so I am putting it here again.
> 
> 1. The page-table code itself should not flush the TLB itself on changes, that should be up to the caller. The caller usually has better knowledge of the flushing scope than the page-table code (CPU-set to flush on, global vs. non-global, PCID, ...). What the new paging crate should provide is tracking of which addresses or ranges need to be flushed.
> 2. COCONUT-SVSM has another abstraction in the page-table code: `PageTablePart`. This abstraction holds a sub-page-table which can be mapped into several top-level page-tables. Is there a plan to move that to the generic crate as well?

I rebased the branch and added a commit to defer the TLB flush and give caller a hint when it may need to flush in the paging crate.
cd46489e18b04f740eed126d4197dacf80af5883

msft-jlange:
The Technical Steering Committee has had some lengthy discussions about this proposed change, and we have concluded that we do not support moving the page table code out of the kernel into a separate crate.  Management of page tables is tightly coupled with memory management primitives, including address space management, TLB management, the tracking of allocations, and many other operations.  Moving page table management into a separate crate will introduce a layer of abstraction that will make it difficult to maintain the highest levels of efficiency and robustness in the memory manager.  Furthermore, depending on a separate crate that may be consumed in other projects will bind the memory manager to a contract that is not flexible enough to evolve rapidly as the memory manager changes over time.  Consequently, we believe that page table management logic should remain embedded in the memory manager where it can be maintained as a part of the memory management logic.

ziqiaozhou:
> uce a layer of abstraction that will make it difficult to maintain the highest levels of efficiency and robustness in the memory manager. Furthermore, depending on a separate crate that may be consumed in other projects will bind the memory manager to a contract that is not flexible enough to evolve rapidly as the memory manager changes over time. Consequently, we believe that page tabl

Thanks for the feedback. I lost the track of steering committee's discussion since I was not attending the meeting in the past few weeks due to a conflict.

The address definition is just moved out of kernel without any additional layers since I need to avoid cyclic dependency

As for the abstraction, I only have a small abstraction in paging/src/traits.rs which I think is relatively small since it highly follows SVSM's original design where almost all trait methods in trait [`ArchPagingMeta`](https://github.com/ziqiaozhou/svsm/blob/pgtable-refactor/paging/src/traits.rs#L110C11-L110C25) were free functions in svsm/kernel/src/mm/pagetable.rs before. The trait [`GenericPageTableFlags`](https://github.com/ziqiaozhou/svsm/blob/pgtable-refactor/paging/src/traits.rs#L241) is there since svsm may need to support ARM and so I just add that abstraction. The only trait that was specific for other kernel is [`PagingHandler`](https://github.com/ziqiaozhou/svsm/blob/pgtable-refactor/paging/src/traits.rs#L199) which only have three methods: alloc, dealloc, paddr_to_vaddr.

The only exception is the `TlbFlush` trait in commit cd46489, which is because I want to enforce the TLB flush as precise as possible. However, it can also be simplified if that abstraction looks too length which I also agree with. In that case, I should not return any result to indicate whether we need a TLB flush. That is also why I paused that commit for a while until I pushed it.

Issue #1057, opened 2026-05-11
AMD SEV-SNP: Use correct `XCR0` value in `snp_cpuid()`


joergroedel:
The `snp_cpuid()` function currently uses either `0` or `1` as the XCR0 input value for lookups in the CPUID table. This is a bug as it might cause COCONUT to allocate not enough memory for XSAVE areas, leading to memory corruption on context switches.

00xc:
Interestingly, we set XCR0=7, but we do not have any leaves in the CPUID page with that XCR0 value, we only have the following for `EAX=0xd` and `ECX=0` / `ECX=1`:

```
EAX_IN: 0x0000000d ECX_IN: 0x00000000 XCR0_IN: 0x00000001 XSS_IN: 0x00000000 EAX_OUT: 0x00000007 EBX_OUT: 0x00000240 ECX_OUT: 0x00000340 EDX_OUT: 0x00000000
EAX_IN: 0x0000000d ECX_IN: 0x00000001 XCR0_IN: 0x00000001 XSS_IN: 0x00000000 EAX_OUT: 0x0000000f EBX_OUT: 0x00000240 ECX_OUT: 0x00000000 EDX_OUT: 0x00000000
```

So if we use the actual XCR0 after it has been set in `xcr0_set()`, the CPUID page query will fail. So I think we cannot do better than what we have right now.

00xc:
If we had the right CPUID leaves in the CPUID page then something like this would be the thing to do:

```diff
diff --git a/kernel/src/platform/snp.rs b/kernel/src/platform/snp.rs
index bf89484ff6d5..97537c15ad89 100644
--- a/kernel/src/platform/snp.rs
+++ b/kernel/src/platform/snp.rs
@@ -16,6 +16,7 @@ use super::snp_fw::{
 use crate::address::{Address, PhysAddr, VirtAddr};
 use crate::boot_params::BootParams;
 use crate::console::init_svsm_console;
+use crate::cpu::control_regs::read_cr4;
 use crate::cpu::cpuid::cpuid_table;
 use crate::cpu::cpuid::cpuid_table_raw;
 use crate::cpu::cpuid::init_cpuid_table;
@@ -52,10 +53,11 @@ use syscall::GlobalFeatureFlags;
 
 #[cfg(test)]
 use bootdefs::platform::SvsmPlatformType;
-use core::arch::x86_64::CpuidResult;
+use core::arch::x86_64::{_xgetbv, CpuidResult};
[...]

PR #1063, opened 2026-05-13
kernel: cpu/sse: retrieve XSAVE area size from 0xD CPUID subleaves

00xc:
The size that needs to be allocated for the XSAVE area can traditionally be computed by querying CPUID with `EAX=0xD` `ECX=0`, which will return the required size based on the currently enabled floating point features (which are stored in XCR0).
    
On SEV-SNP, the CPUID instruction for this particular leaf is emulated by querying the CPUID page. These entries are indexed by the input EAX, ECX and XCR0, meaning that there would need to be a dedicated CPUID page entry for each potential combination of enabled features in XCR0.
    
Instead of modifying igvmbuilder to request each of these entries in the CPUID page, compute the XSAVE area size manually. This can be done by obtaining the size required for each enabled feature, querying CPUID with `EAX=0xD` and setting `ECX` to the corresponding feature bit (in the 2-63 range). This works for all platforms, not just SEV-SNP.

This mimicks what Linux does to compute the XSAVE area size under SEV-SNP as well.

In order for this to work, we need to have the right `EAX=0xD` subfunctions present in the CPUID page, which for now only adds the YMM subfunction.

This addresses #1057, although it does not prevent misuse of the `EAX=0xD` `ECX={0,1}` mechanism to retrieve the XSAVE area size. This can't easily be done because the output registers for these subleaves, besides `ECX`, have legitimate uses and do not depend on XCR0.

### DIFF
--- a/tools/igvmbuilder/src/cpuid.rs
+++ b/tools/igvmbuilder/src/cpuid.rs
@@ -31,14 +31,10 @@ impl SnpCpuidLeaf {
     }
 
     pub fn new2(eax_in: u32, ecx_in: u32) -> Self {
-        Self::new3(eax_in, ecx_in, 0)
-    }
-
-    pub fn new3(eax_in: u32, ecx_in: u32, xcr0: u64) -> Self {
         Self {
             eax_in,
             ecx_in,
-            xcr0,
+            xcr0: 0,
             xss: 0,
             eax_out: 0,
             ebx_out: 0,
@@ -86,8 +82,8 @@ impl SnpCpuidPage {
         cpuid_page.add(SnpCpuidLeaf::new2(7, 1))?;
         cpuid_page.add(SnpCpuidLeaf::new1(11))?;
         cpuid_page.add(SnpCpuidLeaf::new2(11, 1))?;
-        cpuid_page.add(SnpCpuidLeaf::new3(13, 0, 1))?;
-        cpuid_page.add(SnpCpuidLeaf::new3(13, 1, 1))?;
+        cpuid_page.add(SnpCpuidLeaf::new2(13, 0))?;
+        cpuid_page.add(SnpCpuidLeaf::new2(13, 1))?;
         // XSAVE size for YMM registers
         cpuid_page.add(SnpCpuidLeaf::new2(13, 2))?;
         cpuid_page.add(SnpCpuidLeaf::new1(0x80000000))?;
--- a/tools/igvmbuilder/src/cpuid.rs
+++ b/tools/igvmbuilder/src/cpuid.rs
@@ -88,6 +88,8 @@ impl SnpCpuidPage {
         cpuid_page.add(SnpCpuidLeaf::new2(11, 1))?;
         cpuid_page.add(SnpCpuidLeaf::new3(13, 0, 1))?;
         cpuid_page.add(SnpCpuidLeaf::new3(13, 1, 1))?;
+        // XSAVE size for YMM registers
+        cpuid_page.add(SnpCpuidLeaf::new2(13, 2))?;
         cpuid_page.add(SnpCpuidLeaf::new1(0x80000000))?;
         cpuid_page.add(SnpCpuidLeaf::new1(0x80000001))?;
         cpuid_page.add(SnpCpuidLeaf::new1(0x80000002))?;
--- a/cpuarch/src/snp_cpuid.rs
+++ b/cpuarch/src/snp_cpuid.rs
@@ -7,7 +7,7 @@
 const SNP_CPUID_MAX_COUNT: usize = 64;
 
 #[derive(Copy, Clone, Default, Debug)]
-#[repr(C, packed)]
+#[repr(C)]
 pub struct SnpCpuidFn {
     pub eax_in: u32,
     pub ecx_in: u32,
@@ -21,7 +21,7 @@ pub struct SnpCpuidFn {
 }
 
 #[derive(Copy, Clone, Debug)]
-#[repr(C, packed)]
+#[repr(C)]
 pub struct SnpCpuidTable {
     pub count: u32,
     pub reserved_1: u32,
--- a/kernel/src/cpu/cpuid.rs
+++ b/kernel/src/cpu/cpuid.rs
@@ -4,6 +4,7 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
+use super::sse::XSAVE_LEGACY_SIZE;
 use crate::address::VirtAddr;
 use crate::types::PAGE_SIZE;
 use crate::utils::immut_after_init::ImmutAfterInitCell;
@@ -19,6 +20,52 @@ pub fn register_cpuid_table(table: &'static SnpCpuidTable) {
     CPUID_PAGE
         .init(table)
         .expect("Could not initialize CPUID page");
+}
+
+/// Remove XSAVE feature bits from leaf 0xD if the corresponding size
+/// subleaf (2-63) is not present in the CPUID page. This prevents
+/// enabling a feature in XCR0/XSS whose XSAVE area size cannot be
+/// determined.
+///
+/// Updates as well ECX in subleaf 0, which reports maximum XSAVE
+/// area size.
+fn filter_xsave_features(table: &mut SnpCpuidTable) {
+    // Legacy x87 + SSE are always present
+    let mut avail = 3u64;
+    let mut max_size = XSAVE_LEGACY_SIZE;
+
+    // Gather available features from ECX={2-63} subleaves
+    for e in table.entries() {
+        // If the size subleaf is present and reports a non-zero size, the feature
+        // can be enabled.
+        if e.eax_in == 0xd && e.ecx_in >= 2 && e.ecx_in < 64 && e.eax_out != 0 {
+            avail |= 1u64 << e.ecx_in;
+            max_size = max_size.max(e.ebx_out + e.eax_out);
+        }
+    }
+
+    // Now remove the available feature bits from ECX={0,1} subleaves
+    for e in table.entries_mut().iter_mut().filter(|e| e.eax_in == 0xd) {
+        let (hi, lo) = match e.ecx_in {
+            // Subleaf 0: EAX/EDX contain XCR0 features
+            0 => (&mut e.edx_out, &mut e.eax_out),
+            // Subleaf 1: ECX/EDX contain XSS features
+            1 => (&mut e.edx_out, &mut e.ecx_out),
+            _ => continue,
+        };
+        *lo &= avail as u32;
+        *hi &= (avail >> 32) as u32;
+    }
+
+    // ECX in subleaf ECX=0 indicates maximum XSAVE area size, update it
+    // as well
+    if let Some(e) = table
+        .entries_mut()
+        .iter_mut()
+        .find(|e| e.eax_in == 0xd && e.ecx_in == 0)
+    {
+        e.ecx_out = max_size;
+    }
 }
 
 /// # Safety
@@ -42,6 +89,7 @@ pub unsafe fn init_cpuid_table(addr: VirtAddr) {
         }
     }
 
+    filter_xsave_features(table);
     register_cpuid_table(table);
 }
 
--- a/kernel/src/cpu/cpuid.rs
+++ b/kernel/src/cpu/cpuid.rs
@@ -139,21 +139,17 @@ impl CpuidLeaf {
     }
 }
 
-pub fn cpuid_table_raw(eax: u32, ecx: u32, xcr0: u64, xss: u64) -> Option<CpuidResult> {
+pub fn cpuid_table(eax: u32, ecx: u32) -> Option<CpuidResult> {
     CPUID_PAGE
         .entries()
         .iter()
-        .find(|f| eax == f.eax_in && ecx == f.ecx_in && xcr0 == f.xcr0_in && xss == f.xss_in)
+        .find(|f| eax == f.eax_in && ecx == f.ecx_in)
         .map(|f| CpuidResult {
             eax: f.eax_out,
             ebx: f.ebx_out,
             ecx: f.ecx_out,
             edx: f.edx_out,
         })
-}
-
-pub fn cpuid_table(eax: u32, ecx: u32) -> Option<CpuidResult> {
-    cpuid_table_raw(eax, ecx, 0, 0)
 }
 
 pub fn dump_cpuid_table() {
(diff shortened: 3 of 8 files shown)


msft-jlange:
I think this is an excellent change and I'm very happy to see it coming in, but I think it doesn't quite go far enough, and I think a bit of additional logic should be included in this PR instead of deferring to the future.

To maintain self-consistency, the code should build the maximum set of available extended features by determining which subleaves are present and which are absent.  For all subleaves that are absent, the corresponding features should be removed from the EAX/EDX values of `EAX=0xD,ECX=0x0` as well as from the ECX/EDX values of `EAX=0xD,ECX=0x1`.  Features should not be advertised as valid in either XCR0 or XSS unless the corresponding CPUID information is available.  This much really needs to be done now.

Also, since you are able to compute the maximum size of the XSAVE area with all features enabled (based on the subleaves you enumerate), we should probably also update the ECX value in the `EAX=0xD,ECX=0x0` leaf to match the computed maximum, since there's no reason to describe a size that is bigger than the set of features that we are willing to enable.  This is less critical to do now but it would be nice to do as part of this XSAVE CPUID effort.

tools/igvmbuilder/src/cpuid.rs
```diff
@@ -88,6 +88,8 @@ impl SnpCpuidPage {
         cpuid_page.add(SnpCpuidLeaf::new2(11, 1))?;
         cpuid_page.add(SnpCpuidLeaf::new3(13, 0, 1))?;
         cpuid_page.add(SnpCpuidLeaf::new3(13, 1, 1))?;
```
msft-jlange: Since we're now computing the XCR0 contribution to CPUID leaves in the kernel CPUID code, can we remove the `new3` variant of `SnpCpuidLeaf` here and just drop XCR0 from the input template altogether?

00xc:
> I think this is an excellent change and I'm very happy to see it coming in, but I think it doesn't quite go far enough, and I think a bit of additional logic should be included in this PR instead of deferring to the future.
> 
> To maintain self-consistency, the code should build the maximum set of available extended features by determining which subleaves are present and which are absent. For all subleaves that are absent, the corresponding features should be removed from the EAX/EDX values of `EAX=0xD,ECX=0x0` as well as from the ECX/EDX values of `EAX=0xD,ECX=0x1`. Features should not be advertised as valid in either XCR0 or XSS unless the corresponding CPUID information is available. This much really needs to be done now.

Makes sense, yes, I added it to this PR.

> Also, since you are able to compute the maximum size of the XSAVE area with all features enabled (based on the subleaves you enumerate), we should probably also update the ECX value in the `EAX=0xD,ECX=0x0` leaf to match the computed maximum, since there's no reason to describe a size that is bigger than the set of features that we are willing to enable. This is less critical to do now but it would be nice to do as part of this XSAVE CPUID effort.

Added this as well.

tools/igvmbuilder/src/cpuid.rs
00xc: After applying this change (https://github.com/coconut-svsm/svsm/commit/01e84e43d8723aaf512323615e5a31b81ac1a446) this the SVSM fails to boot. Reverting that commit fixes it.

```
[SVSM] ERROR: Panic on CPU[0]! COCONUT-SVSM Version: 2026.04-devel-54-g01e84e43
[SVSM] ERROR: Info: panicked at kernel/src/cpu/sse.rs:53:9:
xsave unsupported
[SVSM] ---BACKTRACE---:
[SVSM]   [ffffff8000076fe8] svsm::cpu::sse::sse_init+0xe8
[SVSM]   [ffffff800007553e] svsm_entry+0xb9e
[SVSM]   [ffffff8000000028] startup_64+0x28
[SVSM] ---END---
```

I think asking the SNP firmware for the `EAX=0xD` leaf with `XCR0=0` will generate zeroed out bits in output EAX.

00xc: For now I am removing the change in favor of adding a comment. If we really want to simplify CPUID page lookup we can just ignore XCR0, but keep requesting the leaf with XCR0=1 from igvmbuilder, although that might make things a bit confusing.

msft-jlange: I can't think of any good reason that making this change would cause the SVSM to fail to boot, so I worry that there is a hidden problem with your CPUID calculation logic.  The maximum supported information (maximum mask and maximum XSAVE size) shouldn't vary based on the XCR0 value, and the current information is something you calculate dynamically based on the list of maximums found in the CPUID information.  Can you at least figure out what part of the CPUID information is causing the problem when the change is present?  We really need to understand whether this is coming from SVSM code or from elsewhere (and if elsewhere, it would be really helpful to know if this is due to FW behavior or QEMU behavior, but that's less important with respect to this PR).

If you can explain for sure that this bug cannot be avoided with SVSM code, then I'd like to see two more changes.  One is a comment in the IGVM builder code that explains that a non-zero XCR0 is required (possibly for unknown reasons) - I don't see that in the current version of the code.  The other change would be to ensure that `EAX=0xD,ECX=0x2` also specifies a non-zero XCR0 for consistency.  I would expect all of the `EAX=0xD` leaves to have similar declarations in the list; it's odd to see that D/0 and D/1 specify one while D/2 specifies zero, without any explanation.  Ideally, all of them should specify 3 (since that value is architecturally guaranteed to be valid) or should specify 3 plus the bit for the particular subleaf (so 7 in the case of D/2), though I think a value of 3 everywhere would be the most sensible because it is consistent and thus more obvious.

00xc: This is happening before we compute the XSAVE size. This is the panic we are hitting:

```rust
    if cpu_has_feat(Feature::Xsave) && cpu_has_feat(Feature::XsaveOpt) {
        if cpu_has_feat(Feature::Xcr0X87)
            && cpu_has_feat(Feature::Xcr0Sse)
            && cpu_has_feat(Feature::Xcr0Avx)
        {
            SVSM_XCR0.fetch_or(XCR0_YMM_ENABLE, Ordering::Relaxed);
        }
        cr4_xsave_enable();
        xcr0_set();
    } else {
        panic!("xsave unsupported");
    }
```

Turns out the CPUID page entries still have XCR0=1 set, so the leaf lookup fails (because we search with XCR0=0):

```
[SVSM] EAX_IN: 0x0000000d ECX_IN: 0x00000000 XCR0_IN: 0x00000001 XSS_IN: 0x00000000 EAX_OUT: 0x00000007 EBX_OUT: 0x00000240 ECX_OUT: 0x00000340 EDX_OUT: 0x00000000
[SVSM] EAX_IN: 0x0000000d ECX_IN: 0x00000001 XCR0_IN: 0x00000001 XSS_IN: 0x00000000 EAX_OUT: 0x0000000f EBX_OUT: 0x00000240 ECX_OUT: 0x00000000 EDX_OUT: 0x00000000
```

It happens even if I disable `filter_xsave_features()`. Qemu seems to be hardcoding XCR0=1:

https://github.com/coconut-svsm/qemu/blob/34d36846c6bc670bdb632b94be11d386d4c925d0/target/i386/sev.c#L1499

I also tried other XCR0 values in igvmbuilder to see if the CPUID page changes, but the result is the same. So it's most likely qemu.

> The other change would be to ensure that EAX=0xD,ECX=0x2 also specifies a non-zero XCR0 for consistency.

Technically none of the output registers for ECX > 2 depend on XCR0, so it does make sense to have it the way it is now. I could add a comment though.

msft-jlange: > Turns out the CPUID page entries still have XCR0=1 set, so the leaf lookup fails (because we search with XCR0=0):

This puzzles me a lot.  First, if we search with XCR=0, and if QEMU always inserts the leaves with XCR0=1, then how does the initial content of the IGVM CPUID page make any difference?  I would expect the search to filter on XCR0 identically regardless of what's in the IGVM file.

And why does the CPU feature lookup require an exact match on XCR0 anyway?  There is no CPU feature that is advertised in a way that depends on the content of XCR0.  Shouldn't we modify the feature detection code so that XCR0 is ignored in the CPUID lookup?

> Technically none of the output registers for ECX > 2 depend on XCR0, so it does make sense to have it the way it is now. I could add a comment though.

I agree that it isn't required, which means you could set XCR0 to anything you want and still get the same result.  My opinion is that code readability is dramatically improved when the code is consistent.  So if XCR0=0 and XCR0=1 give you the same answer, but XCR0=1 creates consistency and thus improves readability, then why would we not just set XCR0=1?  Yes, you could address this with a comment, but I would argue that readability is maximized when comments aren't required.

00xc: > This puzzles me a lot. First, if we search with XCR=0, and if QEMU always inserts the leaves with XCR0=1, then how does the initial content of the IGVM CPUID page make any difference? I would expect the search to filter on XCR0 identically regardless of what's in the IGVM file.

It doesn't. Boot simply fails because my patch removes the XCR0 fixup when querying CPUID via `SnpPlatform::cpuid()`, so the query against `EAX=0xD` `ECX=0x0` `XCR0=0` fails, and thus `cpu_has_feat()` returns `false` in the code I pasted above, resulting in a panic.

> And why does the CPU feature lookup require an exact match on XCR0 anyway? There is no CPU feature that is advertised in a way that depends on the content of XCR0. Shouldn't we modify the feature detection code so that XCR0 is ignored in the CPUID lookup?

We just match on all input registers (EAX/ECX/XCR0/XSS) for the leaf, since in theory they influence the output. But they are really unused (especially after this PR), so I would agree that maybe it makes more sense to drop the match on XCR0 and XSS.

> I agree that it isn't required, which means you could set XCR0 to anything you want and still get the same result. My opinion is that code readability is dramatically improved when the code is consistent. So if XCR0=0 and XCR0=1 give you the same answer, but XCR0=1 creates consistency and thus improves readability, then why would we not just set XCR0=1? Yes, you could address this with a comment, but I would argue that readability is maximized when comments aren't required.

To me seeing XCR0=1 denotes some intent, or in other words, that the value needs to actively be set for a reason, so I would actually find it more confusing.

msft-jlange: > We just match on all input registers (EAX/ECX/XCR0/XSS) for the leaf, since in theory they influence the output. But they are really unused (especially after this PR), so I would agree that maybe it makes more sense to drop the match on XCR0 and XSS.

Good, so then will you plan to update your PR so that CPU feature detection (at least) ignores XCR0/XSS in the lookup?  If you don't think you can do that now, then can you explain what would need to happen first before this change becomes possible?

> To me seeing XCR0=1 denotes some intent, or in other words, that the value needs to actively be set for a reason, so I would actually find it more confusing.

If that is the case, then we still have to ensure there is an explicit comment explaining the difference, and that comment needs to reflect all of the reasoning that we have been discussing here.  It must be clear to anybody who reads the code exactly why the various leaves differ and what factors go into that difference - especially so that when a future attempt is made to clean this up, there is no confusion about what must be accounted for when the cleanup happens.  Alternatively, if we can just change our code to ignore XCR0 as described above, then we can just get rid of `new3` now as described earlier and there's nothing to comment.

I will also point out that we cannot rely on the quirky behavior of QEMU because the Hyper-V loader behaves differently.  Removing the reliance of XCR0 at the time of CPUID lookup means that we will be tolerant of whatever the loader chooses to do, and thus there will be no dependency on how the table is constructed by the IGVM file and its loader.

00xc: > Good, so then will you plan to update your PR so that CPU feature detection (at least) ignores XCR0/XSS in the lookup? If you don't think you can do that now, then can you explain what would need to happen first before this change becomes possible?

Yes, I'll update the PR.

> I will also point out that we cannot rely on the quirky behavior of QEMU because the Hyper-V loader behaves differently. Removing the reliance of XCR0 at the time of CPUID lookup means that we will be tolerant of whatever the loader chooses to do, and thus there will be no dependency on how the table is constructed by the IGVM file and its loader.

What does Hyper-V do in this case? Does it generate the EAX=0xD ECX=0x0 leaf with XCR0_IN=0 and the same output values?

00xc: I think we can mark this as resolved. The latest patch in this PR addresses your comment while allowing the SVSM to boot.

msft-jlange: > What does Hyper-V do in this case? Does it generate the EAX=0xD ECX=0x0 leaf with XCR0_IN=0 and the same output values?

Agreed.

> What does Hyper-V do in this case? Does it generate the EAX=0xD ECX=0x0 leaf with XCR0_IN=0 and the same output values?

Hyper-V behavior depends on whether we're importing a CPUID_DATA page or a CPUID_XF page.  COCONUT-SVSM only uses CPUID_DATA today, though I would like to start using both once QEMU can support CPUID_XF.

In the case of CPUID_DATA, Hyper-V consumes the values of EAX/ECX/XCR0/XSS from each CPUID leaf as specified in the IGVM file and obtains CPUID data as specified.  It writes back the EAX/EBX/ECX/EDX values but leaves the input specifiers alone.  This is what is passed to the SNP firmware for page import.  The CPUID lookup in Hyper-V does not require XCR0 to be architecturally valid in order to obtain a meaningful result.  I don't know offhand what it generates for XSAVE size if XCR0 is zero, but whatever it is doing is acceptable to the SNP firmware when the contents are validated.  Since features 0 and 1 do not contribute to XSAVE size (they are in the base XSAVE area), I expect that those bits are ignored by both the hypervisor's CPUID lookup and by the SNP firmware validation when it validates the "XSAVE size for the current XCR0" field of the result.

In the case of CPUID_XF, Hyper-V generates a page whose first entry is EAX=0xD,ECX=0,XCR0=3,XSS=0 and whose second entry is the same but with ECX=1.  Then it combines the max XCR0 and XSS values as obtained from those two leaves to compute the full set of supported features (XCR0 and XSS are defined to be disjoint, so they can be combined for this purpose).  For each supported leaf _n_, it adds EAX=0xD,ECX=_n_,XCR0=3,XSS=0 entries to the page and adds the correct CPUID value for each.  When complete, this is passed to the SNP firmware for page import.  In this way, the guest can obtain the correct CPUID leaves for all features that are supported by the host without having to know ahead of time what that list is.

Once QEMU can support both formats, I would like to start using CPUID_XF as the way that EAX=0xD data is obtained by COCONUT-SVSM.  This will result in two CPUID pages, each with separate data.  I would also like the kernel to parse that data and copy it into a separate table that is sorted, with a tree structure (for leaves with subleaves) so that lookup is much simpler.  That is not a high priority but it would be nice to do someday because it allows us to make CPUID support more robust, especially when we add paravisor functionality in the future.

00xc:
I think this should be ready to merge @msft-jlange

00xc:
Rebased on main to fix an import conflict.

joergroedel (commented):
Testing looks good, but needs a rebase.

joergroedel (approved):

PR #1064, opened 2026-05-13
cpu: Make sure guest's GIF is set

MelodyHuibo:
When the guest's Global Interrupt Flag (GIF) is 0, the virtual interrupt queued in Vintr_Ctrl will never be taken.

In order to be able to inject interrupts with Vintr_Ctrl when Alternate Injection is enabled, set vGIF in Vintr_Ctrl so queued virtual interrupts can be recognized by the guest.

### DIFF
--- a/kernel/src/cpu/vmsa.rs
+++ b/kernel/src/cpu/vmsa.rs
@@ -8,7 +8,7 @@ use crate::hyperv;
 use crate::sev::status::sev_flags;
 use crate::types::{GUEST_VMPL, SVSM_CS, SVSM_CS_ATTRIBUTES, SVSM_DS, SVSM_DS_ATTRIBUTES};
 use cpuarch::sev_status::SEVStatusFlags;
-use cpuarch::vmsa::{VMSA, VMSASegment};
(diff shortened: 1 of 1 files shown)


msft-jlange (approved):

kernel/src/cpu/vmsa.rs
```diff
@@ -170,6 +170,7 @@ pub fn init_guest_vmsa(v: &mut VMSA, rip: u64, alternate_injection: bool) {
     // Enable alternate injection if requested.
     if alternate_injection {
         sev_status.insert(SEVStatusFlags::ALT_INJ);
+        v.vintr_ctrl = VIntrCtrl::new().with_vgif(true);
```
stefano-garzarella, line 173: in `disable_alternate_injection()` we clear the `ALT_INJ` flag from `sev_features`, should we do the same for `vgif` in `vintr_ctrl`? (I'm not sure TBH, it's more of a question I'm asking myself to try to understand better this)

msft-jlange: When alternate injection is disabled, the VMSA VIntrCtrl is ignored, so it doesn't matter what vGIF is set.

PR #1065, opened 2026-05-14
error: avoid `SvsmReqError` outside of SVSM-specific paths

msft-jlange:
The `SvsmReqError` type is defined based on the SVSM protocol, and its use should only be a part of code paths that specifically support the SVSM protocol.  Common code paths that provide general services should use the system-wide error type `SvsmError`.

### DIFF
--- a/kernel/src/protocols/errors.rs
+++ b/kernel/src/protocols/errors.rs
@@ -83,6 +83,11 @@ impl From<SvsmError> for SvsmReqError {
                 ApicError::Registration => Self::protocol(SVSM_ERR_APIC_CANNOT_REGISTER),
             },
             SvsmError::Attestation(e) => Self::protocol(e as u64),
+            SvsmError::InvalidParameter => Self::invalid_parameter(),
+            SvsmError::InvalidFormat => Self::invalid_format(),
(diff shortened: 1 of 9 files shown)


kernel/src/protocols/errors.rs
```diff
@@ -83,6 +83,9 @@ impl From<SvsmError> for SvsmReqError {
                 ApicError::Registration => Self::protocol(SVSM_ERR_APIC_CANNOT_REGISTER),
             },
             SvsmError::Attestation(e) => Self::protocol(e as u64),
+            SvsmError::InvalidParameter => Self::invalid_parameter(),
+            SvsmError::InvalidFormat => Self::invalid_format(),
+            SvsmError::NotSupported => Self::invalid_request(),
```
stefano-garzarella: Should we also handle here `MessageDecryptionFailure` and `SnpGuestRequest` ?

msft-jlange: Yes, those were previously returned as `SvsmReqError::invalid_request()` so we should preserve that behavior.

stefano-garzarella (approved):

PR #1066, opened 2026-05-14
scripts/launch_guest: remove sudo from QEMU invocation

luigix25:
Running QEMU as root via sudo grants unnecessary privileges when the only resource requiring access is /dev/sev. ~Instead, document that the user should be in the `sev` group to access the device.~

Original discussion: https://github.com/coconut-svsm/svsm/pull/1058#discussion_r3225903449

### DIFF
--- a/Documentation/docs/installation/INSTALL.md
+++ b/Documentation/docs/installation/INSTALL.md
@@ -356,19 +356,24 @@ scripts/launch_guest.sh
 ```
 
 The path to QEMU can also be specified either by setting the `QEMU` variable, or
-by passing the path as a parameter:
+by passing the path as a parameter. Both can also include `sudo` to run QEMU
+with elevated privileges:
 
 ```shell
 QEMU=/path/to/qemu-system-x86_64 scripts/launch_guest.sh
 
 scripts/launch_guest.sh --qemu /path/to/qemu-system-x86_64
+
+QEMU="sudo /path/to/qemu-system-x86_64" scripts/launch_guest.sh
+
+scripts/launch_guest.sh --qemu "sudo /path/to/qemu-system-x86_64"
 ```
 
 The script supports a number of other options described in the table below
 
 | Command-line     | Variable | Default               | Description                                                                  |
 | ---------------- | -------- | --------------------- | ---------------------------------------------------------------------------- |
-| `--qemu [path]`  | QEMU     | qemu-system-x86_64    | Path to QEMU binary to use.                                                  |
+| `--qemu [path]`  | QEMU     | qemu-system-x86_64    | Path to QEMU binary to use. Can include `sudo` to run as root.               |
 | `--igvm [path]`  | IGVM     | bin/coconut-qemu.igvm | Path to the IGVM binary to launch.                                           |
 | `--image [path]` | IMAGE    | [None]                | The QEMU disk image to use. If unset then no disk is provided on the guest.  |
 | `--debugserial`  | N/A      | not set               | Define a second serial port that can be used with the COCONUT-SVSM GDB stub. |
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -161,12 +161,6 @@ if [ ! -z "$IMAGE" ]; then
     -device scsi-hd,drive=disk0"
 fi
 
-if [ "$EUID" -ne 0 ] && [ "$ACCEL" != "tcg" ]; then
-	SUDO_CMD="sudo"
-else
-	SUDO_CMD=""
-fi
-
 echo "============================="
 echo "Launching SVSM guest"
 echo "============================="
@@ -190,8 +184,7 @@ if [ -t 0 ]; then
 fi
 
 # Temporarily use -vga none to avoid IGVM VGA init failure in QEMU 10.1
-$SUDO_CMD \
-  "$QEMU" \
+$QEMU \
     -cpu $CPU \
     -machine $MACHINE \
     -object $MEMORY \
--- a/.github/workflows/qemu.yml
+++ b/.github/workflows/qemu.yml
@@ -129,6 +129,9 @@ jobs:
       - name: Install netcat used by test-in-svsm for vsock tests
         run: sudo apt install -y ncat
 
+      - name: Set up KVM permissions
+        run: sudo chmod 666 /dev/kvm
+
       - name: Set up vhost-vsock permissions
         run: sudo chmod 666 /dev/vhost-vsock
 


Documentation/docs/installation/INSTALL.md
stefano-garzarella, line 1: Please split in 2 commits, those changes are unrelated.

stefano-garzarella (changes requested):
`only resource requiring access is /dev/sev.` mm, not sure if we need to mention this, maybe also `/dev/kvm`, `/dev/vhost-vsock`, etc. require special permissions

stefano-garzarella:
Just to clarify: IMO, it’s fine to mention in the documentation that you can run QEMU with `sudo` (I usually avoid that and leave to the user the choice to use sudo or not), and even a user who doesn’t want to deal with permissions can run `sudo ./script/launch_guest.sh`. What I meant in the discussion is that having `sudo` in a script isn’t ideal, and we should avoid it.

But again, if this breaks some other workflow we can avoid this change at all.

CC @joergroedel

Documentation/docs/installation/INSTALL.md
stefano-garzarella: Or remove the changes in the documentation at all (up to you)

luigix25: dropped the doc.

luigix25:
- Added a chmod on the /dev/kvm device so that the CI doesn't fail (like we already did for /dev/vhost-vsock)
- Dropped documentation changes [@stefano-garzarella]
- Modified commit message

stefano-garzarella (approved):

joergroedel:
My setup has no `sev` group, `/dev/sev` is owned by `root:root`. Maybe remoe use of sudo by default, but keep a flag to re-enable it when needed?

luigix25:
> My setup has no `sev` group, `/dev/sev` is owned by `root:root`. Maybe remoe use of sudo by default, but keep a flag to re-enable it when needed?

@joergroedel what do you mean? you can still do `sudo launch_guest [...]`.
I already removed every reference to the `sev` group, but I forgot to update the PR description

joergroedel:
> @joergroedel what do you mean? you can still do `sudo launch_guest [...]`. I already removed every reference to the `sev` group, but I forgot to update the PR description

I use it mostly via `make test-in-svsm`, if I put sudo in front of that the build will also happen as root. At least in the Makefile we need an option to call launch_guest via sudo, so that the build happens as the normal user.

stefano-garzarella:
> > @joergroedel what do you mean? you can still do `sudo launch_guest [...]`. I already removed every reference to the `sev` group, but I forgot to update the PR description
> 
> I use it mostly via `make test-in-svsm`, if I put sudo in front of that the build will also happen as root. At least in the Makefile we need an option to call launch_guest via sudo, so that the build happens as the normal user.

Good point.

What about supporting `QEMU="sudo path/to/qemu" make test-in-svms` ?
IIUC now this doesn't work because we have `"$QEMU"` in the script.

Or do you prefer a parameter?

joergroedel:
> What about supporting `QEMU="sudo path/to/qemu" make test-in-svms` ? IIUC now this doesn't work because we have `"$QEMU"` in the script.
> 
> Or do you prefer a parameter?

Yes, that could work. We also used to have a SUDO=1 parameter to the Makefile, I am fine with either way, as long as it is documented.

stefano-garzarella (changes requested):
> > What about supporting `QEMU="sudo path/to/qemu" make test-in-svms` ? IIUC now this doesn't work because we have `"$QEMU"` in the script.
> > Or do you prefer a parameter?
> 
> Yes, that could work. We also used to have a SUDO=1 parameter to the Makefile, I am fine with either way, as long as it is documented.

makes sense to me (I slightly prefer to avoid SUDO=1, but not a strong opinion).
@luigix25 please implement one of those in this PR.

luigix25:
now it's possible to do `QEMU="sudo /path/to/qemu" make test-in-svsm`

@stefano-garzarella @joergroedel

Documentation/docs/installation/INSTALL.md
```diff
@@ -356,12 +356,15 @@ scripts/launch_guest.sh
 ```
 
 The path to QEMU can also be specified either by setting the `QEMU` variable, or
-by passing the path as a parameter:
+by passing the path as a parameter. The `QEMU` variable can also include `sudo`
```
stefano-garzarella: I guess it's the same also for the `--qemu` parameter, no?

scripts/launch_guest.sh
```diff
@@ -190,8 +184,7 @@ if [ -t 0 ]; then
 fi
 
 # Temporarily use -vga none to avoid IGVM VGA init failure in QEMU 10.1
-$SUDO_CMD \
-  "$QEMU" \
+exec $QEMU \
```
stefano-garzarella: Why using `exec` ? Just removing the double quote should be enough.

In addition with this change now the `trap 'stty "$STTY_ORIGINAL"' EXIT` is not working anymore. Indeed, now after running `/scripts/launch_guest.sh ...` and exiting with `ctrl-]`, `ctrl-c` is not working anymore.

stefano-garzarella (changes requested):
`ctrl-c` restore is broken, see the comment.

scripts/launch_guest.sh
stefano-garzarella: If you want to keep `exec` (IIRC @osteffenrh was suggesting to use it for timeout or something related to @n-ramacciotti work), we can avoid the remap at all and use the qemu monitor feature (I like this idea), in this way:
```diff
diff --git a/scripts/launch_guest.sh b/scripts/launch_guest.sh
index 1c24d33f..fa7e5d03 100755
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -17,7 +17,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 # in the guest.
 REDUCED_PHYS_BITS=1
 C_BIT_POS=$(cargo run --package cbit || true)
-COM1_SERIAL="-serial stdio" # console
+COM1_SERIAL="-serial mon:stdio" # console
 COM2_SERIAL="-serial null"  # debug
 COM3_SERIAL="-serial null"  # used by hyper-v
 COM4_SERIAL="-serial null"  # used by in-SVSM tests
@@ -171,16 +171,9 @@ echo "IMAGE:        ${IMAGE}"
 echo "============================="
 
 
-# Remap Ctrl-C to Ctrl-] to allow the guest to handle Ctrl-C,
-# if we are running with a TTY attached.
 if [ -t 0 ]; then
-  echo "Press Ctrl-] to interrupt"
+  echo "Press Ctrl-A, X to enter the QEMU monitor and exit"
   echo "============================="
-  # Store original terminal settings and restore it on exit
-  STTY_ORIGINAL=$(stty -g)
-  trap 'stty "$STTY_ORIGINAL"' EXIT
-
-  stty intr ^]
 fi
 
 # Temporarily use -vga none to avoid IGVM VGA init failure in QEMU 10.1
@@ -196,7 +189,6 @@ exec $QEMU \
     $IMAGE_DISK \
     -nographic \
     -vga none \
-    -monitor none \
     $COM1_SERIAL \
     $COM2_SERIAL \
     $COM3_SERIAL \

```

But we should check if `timeout` is still working with `exec`.

That said, IMO we can do this change (removing the mapping and using the monitor) with or without `exec`.

Documentation/docs/installation/INSTALL.md
luigix25: yep!

stefano-garzarella: okay, so should we change this sentence? Like: “Both can also include ...”

scripts/launch_guest.sh
stefano-garzarella: > But we should check if `timeout` is still working with `exec`.

I just tried `scripts/test-in-svsm.sh --nocc --timeout 0.2` and it's working also with `exec` used in `scripts/launch_guest.sh` and the changes suggested in the previous comment:
```
...
[SVSM] COCONUT Secure Virtual Machine Service Module Version 2026.03-devel-94-gf265e61cb+
qemu-system-x86_64: terminating on signal 15 from pid 55491 (timeout)
Test Failed: timeout
```

luigix25: thanks for all the testing and the suggestions. I ended up removing the quotes around QEMU. 
I'm _really_ used to `ctrl ]` and I feel like using the monitor it's a little more overhead.

luigix25:
[@stefano-garzarella]
- Removed quotes around $QEMU instead of using exec.
- improved doc and commit messages

scripts/launch_guest.sh
stefano-garzarella: > thanks for all the testing and the suggestions. I ended up removing the quotes around QEMU. I'm _really_ used to `ctrl ]` and I feel like using the monitor it's a little more overhead.

but this is how QEMU is usually used, and will allow us to remove custom code to do pretty much what QEMU is already doing. I'll open a PR, and we can discuss there, it's unrelated.

stefano-garzarella (dismissed):

Documentation/docs/installation/INSTALL.md
stefano-garzarella: Should we document that behavior also here?

scripts/launch_guest.sh
stefano-garzarella: Changes proposed at https://github.com/coconut-svsm/svsm/pull/1066#discussion_r3265041584 posted in https://github.com/coconut-svsm/svsm/pull/1084

Documentation/docs/installation/INSTALL.md
luigix25: done.

luigix25:
- improved documentation [@stefano-garzarella]

joergroedel (approved):

PR #1072, opened 2026-05-18
virtio-drivers: validate host provided len and token before using them

stefano-garzarella:
Both `len` and `token` are provided by the device. If the device is misbehaving the driver would panic.
Let's check them before using and return errors instead.

Same changes also posted upstream, see https://github.com/rcore-os/virtio-drivers/pull/245

CC @osteffenrh


stefano-garzarella:
FYI https://github.com/rcore-os/virtio-drivers/pull/245 is already merged upstream.

luigix25 (approved):

osteffenrh (approved):

joergroedel (approved):

PR #1073, opened 2026-05-18
Fixes various issues in the COCONUT code-base

joergroedel:
Fix a number of issues recently reported on the COCONUT-SVSM code-base. Details are in the commit logs.

### DIFF
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -22,12 +22,11 @@ use crate::utils::MemoryRegion;
 #[cfg(all(feature = "vtpm", not(test)))]
 use crate::vtpm::vtpm_get_manifest;
 
+use crate::sev::ghcb::GhcbError;
+use crate::types::PAGE_SHIFT;
 use alloc::{boxed::Box, vec::Vec};
 use uuid::{Uuid, uuid};
 use zerocopy::{FromBytes, FromZeros, Immutable, IntoBytes, KnownLayout};
-
-use crate::sev::ghcb::GhcbError;
-use crate::types::PAGE_SHIFT;
 
 pub const ATTEST_PROTOCOL_VERSION_MIN: u32 = 1;
 pub const ATTEST_PROTOCOL_VERSION_MAX: u32 = 1;
@@ -81,6 +80,14 @@ impl AttestServicesOp {
         }
 
         Ok(ops)
+    }
+
+    pub fn check_valid(&self) -> Result<(), SvsmReqError> {
+        if !self.is_reserved_clear() {
+            Err(SvsmReqError::invalid_parameter())
+        } else {
+            Ok(())
+        }
     }
 
     /// Checks if reserved fields are all set to zero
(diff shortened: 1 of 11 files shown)


luigix25:
CI failure shows a KVM error, probably it's related to the QEMU rebase but I'm not sure why.
We also switched from tcg to kvm in the CI.

msft-jlange:
> CI failure shows a KVM error, probably it's related to the QEMU rebase but I'm not sure why. We also switched from tcg to kvm in the CI.

This looks like the same random failure I saw last week after merging a PR.  That generated a guest shutdown in virtio tests as well.

luigix25:
> > CI failure shows a KVM error, probably it's related to the QEMU rebase but I'm not sure why. We also switched from tcg to kvm in the CI.
> 
> This looks like the same random failure I saw last week after merging a PR. That generated a guest shutdown in virtio tests as well.

yep, looks the same to me as well

kernel/src/protocols/attest.rs
```diff
@@ -163,6 +163,12 @@ impl AttestServicesOp {
             return Err(SvsmReqError::invalid_parameter());
         }
 
+        // Check for overflow when size is added
+        let _ = PhysAddr::from(
+            self.certificate_gpa
+                .checked_add(size as u64)
+                .ok_or_else(SvsmReqError::invalid_address)?,
+        );
```
stefano-garzarella: Instead of adding this block, can we just use `MemoryRegion::checked_new()` at the end of this function (instead of using `MemoryRegion::new()`) ?

Also, should we do the same in `get_report_region()` and `get_manifest_region()`

kernel/src/protocols/core.rs
```diff
@@ -369,7 +369,7 @@ fn core_pvalidate_one(entry: u64) -> Result<(), SvsmReqError> {
 fn core_pvalidate(params: &RequestParams) -> Result<(), SvsmReqError> {
     let gpa = PhysAddr::from(params.rcx);
 
-    if !gpa.is_aligned(8) || !valid_phys_address(gpa) {
+    if !gpa.is_aligned(8) || !valid_phys_region(&MemoryRegion::new(gpa, 8)) {
```
00xc: Doesn't the fact that this address is 8-byte aligned and that the struct has a size of 8 guarantee that the object does not cross a page boundary? If that is the case, I think valid_phys_address() does enough checks, so this wouldn't be strictly needed.

The second check below is definitely needed OTOH as the amount of entries is variable

kernel/src/protocols/attest.rs
joergroedel: Right, `MemoryRegion::checked_new()` is both better and simpler. Updated the patch.

kernel/src/protocols/core.rs
joergroedel: `valid_phys_address()` and `valid_phys_region()` also check for overlap with any VMSA pages. You are right about the page-crossing check, though.

joergroedel: Actually this does not check yet for overlaps with other CAA regions. I will fix that.

joergroedel: > Actually this does not check yet for overlaps with other CAA regions. I will fix that.

Okay, screw that. `valid_phys_region()` actually checks for overlaps with both, VMSA and CAA address regions. So the check is needed to make sure that the `pvalidate` requests do not fall into any registered VMSA or CAA.

stefano-garzarella (approved):

PR #1075, opened 2026-05-19
apic: support enable/disable for all interrupt vectors

MelodyHuibo:
Per the Alternate Injection spec, SVSM_APIC_CONFIGURE_VECTOR(SVSM APIC protocol call 4) uses bit 9 in ECX to indicate whether all vectors should be enabled or disabled. Add support for this.

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -681,6 +681,15 @@ impl LocalApic {
         }
     }
 
+    pub fn configure_all_vectors(&mut self, allowed: bool) {
+        if allowed {
+            self.allowed_irr[0] |= 1 << 31;
+            self.allowed_irr[1..].fill(u32::MAX);
+        } else {
+            self.allowed_irr.fill(0);
+        }
+    }
+
     fn signal_one_host_interrupt(&mut self, vector: u8, level_sensitive: bool) -> bool {
         let index = (vector >> 5) as usize;
         let mask = 1 << (vector & 31);
--- a/kernel/src/protocols/apic.rs
+++ b/kernel/src/protocols/apic.rs
@@ -8,6 +8,7 @@ use crate::cpu::percpu::this_cpu;
 use crate::platform::SVSM_PLATFORM;
 use crate::protocols::RequestParams;
 use crate::protocols::errors::SvsmReqError;
+use bitfield_struct::bitfield;
 
 const SVSM_REQ_APIC_QUERY_FEATURES: u32 = 0;
 const SVSM_REQ_APIC_CONFIGURE: u32 = 1;
@@ -17,6 +18,15 @@ const SVSM_REQ_APIC_CONFIGURE_VECTOR: u32 = 4;
 
 pub const APIC_PROTOCOL_VERSION_MIN: u32 = 1;
 pub const APIC_PROTOCOL_VERSION_MAX: u32 = 1;
+
+#[bitfield(u64)]
+struct VectorConfig {
+    vector: u8,
+    enable: bool,
+    all_vectors: bool,
+    #[bits(54)]
+    rsvd: u64,
+}
 
 fn apic_query_features(params: &mut RequestParams) -> Result<(), SvsmReqError> {
     // No features are supported beyond the base feature set.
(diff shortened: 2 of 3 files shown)


kernel/src/cpu/apic.rs
```diff
@@ -681,6 +681,14 @@ impl LocalApic {
         }
     }
 
+    pub fn configure_all_vectors(&mut self, allowed: bool) {
+        if allowed {
+            self.allowed_irr.fill(u32::MAX);
```
msft-jlange: Vectors 0-30 should always be considered illegal and therefore they should never be enabled even when enabling all vectors.  This call should only enable vectors 31-255.

MelodyHuibo: Good point. Fixed it. Please review the updated version, thank you

kernel/src/protocols/apic.rs
```diff
@@ -76,13 +76,21 @@ fn apic_configure_vector(params: &RequestParams) -> Result<(), SvsmReqError> {
...
+        return Err(SvsmReqError::invalid_parameter());
+    }
+
+    let allowed = (params.rcx & 0x100) != 0;
+    let all = (params.rcx & 0x200) != 0;
+
+    if all {
+        cpu.configure_apic_all_vectors(allowed)?;
+        Ok(())
+    } else {
```
00xc, line 98: This is getting a bit unwieldy. I would suggest adding a bitfield struct for the vector configuration parameter. For example:

```diff
diff --git a/kernel/src/protocols/apic.rs b/kernel/src/protocols/apic.rs
index 19f79eb8e340..ad24a3337185 100644
--- a/kernel/src/protocols/apic.rs
+++ b/kernel/src/protocols/apic.rs
@@ -4,6 +4,8 @@
 //
 // Author: Jon Lange (jlange@microsoft.com)
 
+use bitfield_struct::bitfield;
+
 use crate::cpu::percpu::this_cpu;
 use crate::platform::SVSM_PLATFORM;
 use crate::protocols::RequestParams;
@@ -18,6 +20,15 @@ const SVSM_REQ_APIC_CONFIGURE_VECTOR: u32 = 4;
 pub const APIC_PROTOCOL_VERSION_MIN: u32 = 1;
 pub const APIC_PROTOCOL_VERSION_MAX: u32 = 1;
 
+#[bitfield(u64)]
+struct VectorConfig {
+    vector: u8,
+    enable: bool,
+    all_vectors: bool,
+    #[bits(54)]
+    rsvd: u64,
+}
+
 fn apic_query_features(params: &mut RequestParams) -> Result<(), SvsmReqError> {
     // No features are supported beyond the base feature set.
     params.rcx = 0;
@@ -76,14 +87,19 @@ fn apic_configure_vector(params: &RequestParams) -> Result<(), SvsmReqError> {
     if !cpu.use_apic_emulation() {
         return Err(SvsmReqError::invalid_request());
     }
-    if params.rcx <= 0x1FF {
-        let vector: u8 = (params.rcx & 0xFF) as u8;
-        let allowed = (params.rcx & 0x100) != 0;
-        cpu.configure_apic_vector(vector, allowed)?;
-        Ok(())
+    let config = VectorConfig::from_bits(params.rcx);
+
+    if config.rsvd() != 0 {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+
+    if config.all_vectors() {
+        cpu.configure_all_vectors(config.enable())?;
     } else {
-        Err(SvsmReqError::invalid_parameter())
+        cpu.configure_apic_vector(config.vector(), config.enable())?;
     }
+
+    Ok(())
 }
 
 pub fn apic_protocol_request(request: u32, params: &mut RequestParams) -> Result<(), SvsmReqError> {
```

00xc: And looking again at it, I think we got the polarity of bit 8 inverted. The spec says if bit 8 is set it means the vector(s) are *not* permitted to be delivered. We already had it wrong before your PR but now we got it doubly wrong

msft-jlange: > And looking again at it, I think we got the polarity of bit 8 inverted. The spec says if bit 8 is set it means the vector(s) are _not_ permitted to be delivered. We already had it wrong before your PR but now we got it doubly wrong

I think the spec is wrong.  The original proposal I wrote said that 1=Enabled and 0=Disabled but the published spec has it reversed.  I will clarify with AMD.

msft-jlange: > > And looking again at it, I think we got the polarity of bit 8 inverted. The spec says if bit 8 is set it means the vector(s) are _not_ permitted to be delivered. We already had it wrong before your PR but now we got it doubly wrong
> 
> I think the spec is wrong. The original proposal I wrote said that 1=Enabled and 0=Disabled but the published spec has it reversed. I will clarify with AMD.

The spec is indeed incorrect.  The next published version of the spec will fix this issue, and will indicate that ECX[8]=1 means that delivery will be enabled and ECX[8]=0 means that delivery will be disabled.  That means that both the current PR and the existing code are correct.

00xc: I see. In any case, I think we should have the struct cleanup I proposed.

MelodyHuibo: Thank you very much for the suggestion, I just fixed it, please review and let me know if you have any other questions.

00xc (approved):

PR #1076, opened 2026-05-19
svsm: detect and display CPU vendor

msft-jlange:
During startup, display the CPU vendor for debug logging purposes.

### DIFF
--- a/kernel/src/platform/native.rs
+++ b/kernel/src/platform/native.rs
@@ -6,8 +6,13 @@
 // Author: Jon Lange <jlange@microsoft.com>
 // Author: Joerg Roedel <jroedel@suse.de>
 
+use super::CpuVendor;
+use super::PageEncryptionMasks;
+use super::PageStateChangeOp;
+use super::PageValidateOp;
+use super::SvsmPlatform;
 use super::capabilities::Caps;
-use super::{PageEncryptionMasks, PageStateChangeOp, PageValidateOp, SvsmPlatform};
+use super::cpuid;
 use crate::address::{PhysAddr, VirtAddr};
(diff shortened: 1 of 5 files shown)


joergroedel (approved):

kernel/src/platform/native.rs
```diff
@@ -84,6 +89,18 @@ impl SvsmPlatform for NativePlatform {
         Ok(())
     }
 
+    fn get_cpu_vendor(&self) -> CpuVendor {
+        if let Some(r) = cpuid(0, 0) {
+            match r.edx {
```
00xc, line 94: Maybe this could be a feature in cpu/features.rs

kernel/src/platform/native.rs
```diff
@@ -84,6 +89,18 @@ impl SvsmPlatform for NativePlatform {
         Ok(())
     }
 
+    fn get_cpu_vendor(&self) -> CpuVendor {
+        if let Some(r) = cpuid(0, 0) {
```
00xc, line 93: Just a nit, but you could use `Self::cpuid()` instead so that it's self-consistent

kernel/src/platform/native.rs
msft-jlange: It could be, but it doesn't apply outside of the `Native` platform and it's only used for debug logging purposes, so it is not out of place where it is.

kernel/src/platform/native.rs
msft-jlange: Given that `cpuid` is intended to be promoted as a function for global use, there doesn't seem to be anything wrong with using it as such instead of clarifying the scope with `Self`.

PR #1078, opened 2026-05-19
cpu/ipi: fix race conditions

msft-jlange:
Ensure that the count of pending IPI completers is only adjusted and never overwritten once any other CPU could possibly observe an IPI request.  Otherwise, any completions that might have been recorded before the pending count is written could be lost, leading to deadlock.

### DIFF
--- a/kernel/src/cpu/ipi.rs
+++ b/kernel/src/cpu/ipi.rs
@@ -365,6 +365,12 @@ fn send_ipi(target_set: IpiTarget<'_>, ipi_helper: &mut dyn IpiHelper) {
     let mut interrupt_target = target_set;
     let mut interrupt_set: MaybeUninit<CpuSet> = MaybeUninit::uninit();
 
+    // There should be no pending activity for the current IPI board.  This is
+    // a debug assert because it does not affect safety, and it should normally
+    // be guaranteed by the design of the IPI logic without depending on the
+    // behavior of any routine that invokes the IPI logic.
+    debug_assert_eq!(ipi_board.pending.load(Ordering::Relaxed), 0);
+
     // Enumerate all CPUs in the target set to advise that an IPI message has
     // been posted.
     let mut include_self = false;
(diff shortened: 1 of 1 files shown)


kernel/src/cpu/ipi.rs
```diff
@@ -398,6 +405,11 @@ fn send_ipi(target_set: IpiTarget<'_>, ipi_helper: &mut dyn IpiHelper) {
                     send_interrupt = true;
                 }
             }
+
+            // Record the count of targets that will need to respond before
+            // this IPI can complete.
+            ipi_board.pending.fetch_add(target_count, Ordering::Relaxed);
+
```
joergroedel, line 412: Although atomics do not panic on under/overflow in Rust, it would be more correct to change pending to `AtomicIsize` instead of leaving it as `AtomicUsize`, no? `Pending` can be decremented as soon as `flag_ipi()` queued the request, which is before `target_count` gets added. A remote CPU can pick up the request in this window and `fetch_sub()` on `0`, causing an underflow.

msft-jlange: > Although atomics do not panic on under/overflow in Rust, it would be more correct to change pending to `AtomicIsize` instead of leaving it as `AtomicUsize`, no?

No, I disagree.  As you point out, Rust atomics are expressly designed to ignore overflow/underflow and therefore the choice of signed/unsigned types is irrelevant if overflow/underflow is a concern.  Since that is the case, I believe the choice of a signed vs. unsigned type should be based entirely on what type of data is being expressed in the variable, since the arithmetic side effects that may arise are irrelevant.  In this case, the data being expressed is a count of CPUs, which is necessarily a positive number.  It is analogous to a length, which is consistently defined as `usize` in Rust.  We would like the flexibility to be able to move data in and out of the `pending` field in this case and copy to/from other values which may represent a count of CPUs, and this copy should be possible without requiring type conversions.  For example, the following should be possible:

`pending.fetch_add(PERCPU_AREAS.len(), Ordering::Relaxed)`

Since `PERCPU_AREAS.len()` is a `usize` by definition, this operation can only be done if `pending` is also a `usize` type.  If it were defined as an `isize` type, then the operation above would require a cast (e.g. `PERCPU_AREAS.len() as isize` or `PERCPU_AREAS.len().try_into().unwrap()`) which would be undesirable for reasons of conversion safety.

So what we're trading off here is the readability of usages of this field that are related to existing `usize` concepts vs. the potential for overflow/underflow which is defined by the standard to be benign in all cases and therefore not a problem.  That tradeoff has a pretty clear win in my mind.

joergroedel (approved):

PR #1079, opened 2026-05-19
github/manual-verify: cache verus toolchain

stefano-garzarella:
Cache the verus binaries (verus, rust_verify, z3, cargo-verus, verus-root,
verusfmt) using actions/cache, keyed on the SHA of vinstall.sh and the
workflow file itself like we already do for QEMU.

On cache hit, vinstall.sh detects the correct version is already installed
and exits early, skipping the ~19 min build from source.

### DIFF
--- a/.github/workflows/manual-verify.yml
+++ b/.github/workflows/manual-verify.yml
@@ -29,6 +29,28 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Calculate Verus cache key
+        run: |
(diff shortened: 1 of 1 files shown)


stefano-garzarella:
Based on #1038

stefano-garzarella:
After removing and adding back `verification` label the time went from 19 mins to 3 mins, so caching is working, but I keep this as draft till #1038 is merged

stefano-garzarella:
v2:
- rebased on main since #1038 is now merged

This PR now is ready to be merged.

joergroedel (approved):

PR #1085, opened 2026-05-20
igvmbuilder: configure target VTL based on the presence of firmware

msft-jlange:
When guest firmware is present, the IGVM platform directive should indicate that multiple privilege levels are required, but if no guest firmware is present, only a single privilege level is required.

### DIFF
--- a/tools/igvmbuilder/src/igvm_builder.rs
+++ b/tools/igvmbuilder/src/igvm_builder.rs
@@ -284,11 +284,17 @@ impl IgvmBuilder {
     }
 
     fn build_platforms(&mut self) {
+        // Set the highest supported VTL based on whether guest firmware is
+        // present.
+        let highest_vtl = match self.firmware {
+            Some(_) => 2,
+            None => 0,
+        };
         if COMPATIBILITY_MASK.contains(SNP_COMPATIBILITY_MASK) {
             self.platforms.push(IgvmPlatformHeader::SupportedPlatform(
                 IGVM_VHS_SUPPORTED_PLATFORM {
                     compatibility_mask: SNP_COMPATIBILITY_MASK,
-                    highest_vtl: 2,
+                    highest_vtl,
                     platform_type: IgvmPlatformType::SEV_SNP,
                     platform_version: 1,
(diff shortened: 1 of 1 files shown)


tools/igvmbuilder/src/igvm_builder.rs
```diff
@@ -321,7 +327,7 @@ impl IgvmBuilder {
             self.platforms.push(IgvmPlatformHeader::SupportedPlatform(
                 IGVM_VHS_SUPPORTED_PLATFORM {
                     compatibility_mask: NATIVE_COMPATIBILITY_MASK,
-                    highest_vtl: 0,
+                    highest_vtl,
```
luigix25, line 330: does it make sense to have multiple privilege levels in native mode? According to the igvm doc: (https://docs.rs/igvm_defs/latest/i686-pc-windows-msvc/igvm_defs/struct.IGVM_VHS_SUPPORTED_PLATFORM.html)

```
On platforms that do not support multiple VTLs, this value must be zero
```

msft-jlange: Similarly, it is not possible to host guest firmware from an SVSM if there is only a single privilege level.  For that reason, it's always safe to assume that native platforms will never support hosting a guest firmware image, and that if anyone ever tries to build a native image that includes guest firmware, then it should not be expected to function correctly.

luigix25: I agree completely that the firmware will never run on the native platform, and building with the firmware is a mistake. So why not set 0 directly? We can view this as a 'defence in depth': even if you build it incorrectly, we can still 'tolerate' the error.

msft-jlange: Fair enough.  I would prefer to make it illegal to build an IGVM file that requests native support with firmware, but this would require an overhaul of our current build targets, as well as the scripts that consume them, so I'll have to save that for the future.

luigix25: yep, that would be a very good idea

joergroedel (approved):

PR #1086, opened 2026-05-20
cpu/vc: remove CAA MSR handling

msft-jlange:
The SVSM never issues RDMSR or WRMSR accesses to the CAA MSR, so its #VC handler does not require code to handle emulation of that MSR.

### DIFF
--- a/kernel/src/cpu/vc.rs
+++ b/kernel/src/cpu/vc.rs
@@ -5,12 +5,9 @@
 // Author: Joerg Roedel <jroedel@suse.de>
 
 use super::idt::common::X86ExceptionContext;
-use crate::address::Address;
 use crate::address::VirtAddr;
-use crate::cpu::X86GeneralRegs;
 use crate::cpu::cpuid::cpuid_table_raw;
 use crate::cpu::percpu::current_ghcb;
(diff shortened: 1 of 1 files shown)


00xc:
We should at least keep a warning or a hard error in case this happens in the future

msft-jlange:
> We should at least keep a warning or a hard error in case this happens in the future

Two observations here.

One is that we don't have a generic warning mechanism for unhandled MSRs.  I don't believe there is anything special about this MSR that would be a cause for a warning/error compared to other MSRs like EFER or PL0_SSP which the SVSM uses and which are meaningless when intercepted and routed to the host via GHCB.  If we are going to come up with a warning mechanism on unhandled MSR intercepts, it should be comprehensive and not specific to the CAA MSR, and such a generic mechanism would certainly be outside the scope of this PR.

Second is that the CAA MSR is not architectural, but is a synthetic MSR that is standardized such that a guest OS can discover SVSM state if that SVSM enables paravisor functionality.  It has no meaning in the SVSM itself and thus I see no risk that somehow the CAA MSR would somehow be used in the SVSM in the future.

joergroedel (approved):

PR #1088, opened 2026-05-20
paging: move address, sizes, and utility types to paging crate

ziqiaozhou:
Create the paging crate and move foundational types from the kernel:
- address.rs (PhysAddr, VirtAddr, Address trait), remove unused VirtPhysPair which was introduced 2 years ago and is no longer used by svsm code.
- sizes.rs (PAGE_SIZE, PAGE_SHIFT, size constants)
- util.rs (align_up, align_down, page_align_up, overlap, bit_mask)
- Associated verus spec/proof files for verification
- paging: move Verus specs and proofs into dedicated dirs
- gate proofs and spec with verus_only

The kernel re-exports these types to maintain API compatibility. Unit tests for utility functions move to paging/src/util.rs.

Move helper functions for #1062. #1062 is based on this PR

### DIFF
--- a/paging/src/lib.rs
+++ b/paging/src/lib.rs
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! This crate provides page table–related functions and data structures.
+
+#![no_std]
+#![cfg_attr(verus_only, verifier::allow(unknown_automatic_derive))]
+#![cfg_attr(
+    verus_only,
+    allow(macro_expanded_macro_exports_accessed_by_absolute_paths)
+)]
+
+pub mod address;
+pub mod sizes;
+pub mod util;
--- a/paging/src/sizes.rs
+++ b/paging/src/sizes.rs
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//
+// Copyright (c) 2022-2023 SUSE LLC
+//
+// Author: Joerg Roedel <jroedel@suse.de>
+
+use verus_stub::*;
+
+#[cfg(verus_only)]
+#[path = "proofs/sizes.rs"]
+mod sizes_spec_defs;
+#[cfg(verus_only)]
+pub use sizes_spec_defs::*;
+#[cfg(verus_only)]
+verus! {
+    broadcast use sizes_spec_defs::group_types_proof;
+}
+
+verus! {
+
+pub const PAGE_SHIFT: usize = 12;
+pub const PAGE_SHIFT_2M: usize = 21;
+pub const PAGE_SHIFT_1G: usize = 30;
+pub const PAGE_SIZE: usize = 1 << PAGE_SHIFT;
+pub const PAGE_SIZE_2M: usize = 1 << PAGE_SHIFT_2M;
+pub const PAGE_SIZE_1G: usize = 1 << PAGE_SHIFT_1G;
+
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PageSize {
+    Regular,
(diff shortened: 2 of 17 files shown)


paging/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
```
00xc: Should probably add "OR Apache-2.0" to match everything else

paging/src/sizes.rs
```diff
@@ -0,0 +1,43 @@
...
+verus! {
+
+pub const PAGE_SHIFT: usize = 12;
+pub const PAGE_SHIFT_2M: usize = 21;
+pub const PAGE_SHIFT_1G: usize = 30;
+pub const PAGE_SIZE: usize = 1 << PAGE_SHIFT;
+pub const PAGE_SIZE_2M: usize = 1 << PAGE_SHIFT_2M;
+pub const PAGE_SIZE_1G: usize = 1 << PAGE_SHIFT_1G;
+
+}
```
00xc, line 28: These should probably be indented, no?

```suggestion
verus! {
    pub const PAGE_SHIFT: usize = 12;
    pub const PAGE_SHIFT_2M: usize = 21;
    pub const PAGE_SHIFT_1G: usize = 30;
    pub const PAGE_SIZE: usize = 1 << PAGE_SHIFT;
    pub const PAGE_SIZE_2M: usize = 1 << PAGE_SHIFT_2M;
    pub const PAGE_SIZE_1G: usize = 1 << PAGE_SHIFT_1G;
}
```

ziqiaozhou: verusfmt usually does not add indentation and even remove it inside the verus! macro, since the code within that macro is intended to be similar to standard Rust code. To keep it consistent with our spec code, which is formatted by both rustfmt and verusfmt, I suggest not adding indentation here.

In addition, this code was moved from the original type.rs. I moved the constants that are tightly coupled with page table functionality into the paging crate.

ziqiaozhou:
@joergroedel This PR is the first step for #1062 and #1100. Feel free to drop any comments if you have concerns on this change.

joergroedel (commented):
A review comment from codex which I am not sure is relevant. Dropping it here for reference:

  - [P2] Guard Verus specs with the cfg used by cargo-verus — /home/joro/src/svsm/paging/src/util.rs:12-18. When running the documented cargo verify --features verus flow, cargo-verus passes --cfg verus_keep_ghost (as the existing kernel code and build script expect), not verus_only. With these #[cfg(verus_only)] guards, the moved alignment specs/proofs are not included while the #[verus_spec] annotations still refer to align_up_requires, align_up_ens, etc., so verification builds of the new paging crate will fail once this crate is pulled in by kernel's verus feature.

ziqiaozhou:
> A review comment from codex which I am not sure is relevant. Dropping it here for reference:
> 
> * [P2] Guard Verus specs with the cfg used by cargo-verus — /home/joro/src/svsm/paging/src/util.rs:12-18. When running the documented cargo verify --features verus flow, cargo-verus passes --cfg verus_keep_ghost (as the existing kernel code and build script expect), not verus_only. With these #[cfg(verus_only)] guards, the moved alignment specs/proofs are not included while the #[verus_spec] annotations still refer to align_up_requires, align_up_ens, etc., so verification builds of the new paging crate will fail once this crate is pulled in by kernel's verus feature.

This is not related. verus_only and verus_keep_ghost are all applied when we run `cargo verus ...`. Since Feb, 2026, verus_only becomes a publicly documented setting attribute that only makes sense during verification, while verus_keep_ghost is not documented. `--features verus` is only used to pick dependencies.

PR #1091, opened 2026-05-25
vtpm: RFC — persistent vTPM state via a pluggable external TPM endpoint

Goodleon-Y:
[Draft PR — RFC for design feedback, not requesting merge.]

## Background

SEV-SNP does not currently provide trusted access from a CVM to
platform devices, including a physical TPM (discrete, firmware, or
otherwise). As a result, the in-CVM vTPM at VMPL0 cannot use a
physical TPM as a backing root of trust and cannot cross a cold boot
on its own. Today, the upstream path being developed by Stefano
Garzarella and others addresses this with encrypted-block + KBS:
durable storage for the vTPM blob plus a remote attestation-gated
key release for decryption. That path fits cloud deployments where
a KBS is already part of the trust infrastructure.

This RFC explores a **complementary** path for deployments where an
online KBS is unavailable, not a fit for the deployment model, or
overkill for the workload:

- on-prem / private cloud where a TPM is already the platform root of
  trust
- disconnected, long-running, or independent-audit workloads
- bring-up / development environments where setting up a KBS is
  disproportionate to the task

The approach: keep the vTPM in-CVM at VMPL0 (no change to today's
attestation surface), but route only its persistent state through a
seal/unseal cycle against an **external TPM 2.0 endpoint** reached
over an adversarial host channel. The endpoint is intentionally
abstract: a TPM 2.0 command/response pair is all the proxy assumes,
so the same code path can target:

- a software vTPM (`swtpm`) for development and CI
- a host-side firmware TPM (AMD-SP fTPM, Intel PTT) where one already
  exists on the platform
- a physical discrete TPM for deployments that want a hardware-rooted
  persistence anchor maintained outside the CVM
- a TPM-as-a-service endpoint where the operator wants centralized
  key custody without a full KBS stack

The choice is a deployment decision; the SVSM code is the same.

This is not a replacement for the KBS-stateful vTPM work. The two
approaches target different deployment regimes and can coexist on the
same SVSM build via different backends behind the `SealedBlobStore`
trait introduced here.

## What this RFC adds

- **libtcgtpm** (`d62abbb`): getter/setter accessors for TPM 2.0
  reference globals (`gp`, `gc`, `gr`) plus a bulk serialize/deserialize
  pair. Useful for any path that needs to lift the reference
  implementation's persistent state out of the in-CVM binary, not just
  this RFC.
- **vtpm/proxy.rs** (`48abdc3`, `6c08cd2`): a small `TpmTransport`
  trait with a `MockTransport` for unit tests and a VSOCK
  implementation (feature-gated). The TPM2 command builder
  (`build_get_random`, `build_nv_define_space`, `build_nv_increment`,
  `build_nv_read`, `build_create_primary`, `build_create`, `build_load`,
  `build_seal`, `build_unseal`, `build_flush_context`) is generic over
  the transport, so the proxy makes no assumption about which TPM 2.0
  endpoint is on the other side.
- **vtpm/sealed_store.rs** (`0bdc9df`): a `SealedBlobStore` trait with
  `StaticBufStore` (in-memory, test-only), a stubbed `IgvmVarStore`,
  and a VSOCK-backed `VsockHostStore` (feature-gated). Intent is to
  align with the upcoming block-layer abstraction so the same trait
  can later wrap a block device, and to leave room for a KBS-backed
  implementation to drop in as a peer backend.
- **vtpm/sealed.rs** (`72f12ef`): two-layer container.
  AES-256-GCM-encrypt the state payload (~1 KB at Tier-1.5, growing
  to ~19 KB once Tier-2 NV-block persistence is enabled), then
  TPM2_Seal only the 60-byte key bundle
  (`aes_key[32] || gcm_tag[16] || nonce[12]`). Keeps the TPM-sealed
  payload well under RSA-2048 size limits regardless of the
  underlying state size. The blob format is documented in
  `kernel/src/vtpm/sealed.rs` and is versioned. The current revision
  carries only the AES-256-GCM payload and the TPM-sealed key bundle;
  a counter-binding field for cross-tenant replay isolation on a
  shared NV counter backend is a planned format addition, not part of
  this revision (see open questions 5 and 8).
- **vtpm/state.rs** (`143ab03`): Rust wrappers around the new
  libtcgtpm accessors, plus a bulk serialize/deserialize path that
  matches the C-side layout.
- **vtpm boot integration** (`6f5db7b`): two boot modes plumbed
  through `kernel/src/svsm.rs`:
  - `Provision` — manufacture vTPM, extract state, seal via the
    configured TPM endpoint, persist the blob.
  - `Recover` — load the blob, unseal via the configured TPM endpoint,
    inject state, signal NV-on + light power-on.
- **vtpm/reseal.rs + Tier-2 persistence** (`b1fbbd1`): full
  guest-state persistence path covering DA lockout, PCR allocation,
  orderly-shutdown state, and the platform NV memory (TPM reference
  implementation's flat 16 KB `s_NV[]` buffer that backs every user NV
  index, evict object, and DA bookkeeping field). The libtcgtpm
  accessors gain section 0x1A (s_NV dump); the SVSM kernel gains
  `kernel/src/vtpm/reseal.rs` registering a runtime hook in the
  command path that, on a successful guest-side
  `TPM2_Shutdown(SU_STATE)`, re-extracts internal state, freshly
  re-encrypts under a new AES key + nonce (no key material is cached
  across calls), and pushes a new SealedBlob via `VsockHostStore`.
  Sealed blob grows from ~2.7 KB (Tier-1.5) to ~19 KB (Tier-2). The
  hook is feature-gated behind `vtpm-persist`; the previous Tier-1.5
  build path is unchanged when the feature is off.

## Known limitations / open questions

1. **Storage backend**: ships with `StaticBufStore` (test-only) and
   stubbed `IgvmVarStore`. Production needs either an IGVM-variable
   path, integration with the upcoming SVSM block layer, or a
   KBS-backed implementation. I deferred picking one so the trait
   shape can be reviewed first.
2. **vm_id**: currently zero in the boot wiring; should derive from
   IGVM `guest_context` once that surface stabilizes.
3. **Transport**: VSOCK implementation is feature-gated; the
   `TpmTransport` trait is intentionally minimal so alternative
   transports (virtio-serial, MMIO, in-process for `swtpm` co-tenants,
   etc.) are welcome.
4. **State extraction**: reaches into the Microsoft TPM reference
   globals via the libtcgtpm accessors added in commit 1. A
   TPM2_ContextSave-based path was considered but cannot reach
   hierarchy seeds and auth values by design. Would appreciate
   guidance on whether the maintainers prefer the direct-accessor
   approach or want a higher-level abstraction.
5. **HMAC channel auth**: the transport layer assumes an adversarial
   host channel, but per-session HMAC binding is only sketched.
   Enforcement would need a session-key provisioning path, which I
   left out of this RFC to keep the diff small.
6. **vTPM in user-mode**: per `DEVELOPMENT-PLAN.md`, the vTPM is
   slated to move to a user-mode task. The boot integration in commit
   7 is intentionally a thin shim (`vtpm_init_sealed`) so it can be
   re-targeted to a user-mode init task. Happy to refactor if there
   is a preferred timeline for that move.
7. **Tier-2 re-seal trigger boundary**: runtime re-seal fires on
   `TPM2_Shutdown(SU_STATE)` only. A guest that does not issue
   Shutdown (forced poweroff, hypervisor-side kill, guest crash,
   sudden power loss) loses that session's NV writes. This matches
   TCG semantics for ungraceful power loss but is worth flagging as
   an API contract.
8. **Tier-2 NV-counter / blob-save ordering**: rollback/replay
   resistance is sketched, not wired in this revision. The intended
   shape is to bind a monotonic counter from the external endpoint's
   NV into the blob and check it on unseal, so a stale or rolled-back
   blob fails closed. With that in place, the ordering matters: if the
   NV counter is advanced inside `seal_state` before the new SealedBlob
   is committed via the configured `SealedBlobStore` backend and the
   store-side write then fails, a later cold boot would see a counter
   ahead of the persisted blob and fall back to Provision instead of
   recovering. A save-then-commit ordering (defer `nv_increment` until
   the store has acknowledged the new blob) avoids that window. This is
   noted now so the counter-check and the commit ordering land
   together; both are tracked as follow-up alongside open question 5.

Seeking feedback on overall direction, whether `SealedBlobStore`
should live alongside the block-layer abstraction, the
`TpmTransport` shape (and whether keeping it endpoint-agnostic is
the right call), the libtcgtpm accessor surface, and whether the
Tier-2 runtime re-seal hook belongs in the kernel command path or
should be lifted to the planned user-mode vTPM task.

## Commit series

```
d62abbb libtcgtpm: Expose direct accessors for TPM persistent globals
48abdc3 vtpm: Add pluggable transport trait with mock implementation
6c08cd2 vtpm: Add VSOCK transport for TPM proxying
0bdc9df vtpm: Add pluggable storage trait for sealed vTPM state
72f12ef vtpm: Add two-layer SealedBlob (AES-256-GCM + TPM2_Seal)
143ab03 vtpm: Add TPM internal state extract/inject via libtcgtpm accessors
6f5db7b vtpm/svsm: Wire TPM-rooted persistence boot mode
b1fbbd1 vtpm: Add runtime re-seal hook and NV-block persistence
```

Each commit compiles and passes `cargo clippy --features vtpm` (and
`--features vtpm,vsock`, `--features vtpm,vtpm-persist` where
relevant) on its own. `make igvm` produces the three IGVM binaries
(`coconut-qemu.igvm`, `coconut-hyperv.igvm`, `coconut-vanadium.igvm`)
at HEAD with both `FEATURES=vtpm` and
`FEATURES=vtpm,vtpm-persist`.

End-to-end validated on an AMD EPYC server with SEV-SNP enabled,
against two TPM 2.0 endpoints sharing the same IGVM binary
(SHA-256 `6c1b09cd...`):

1. **swtpm as the endpoint** (software reference, via
   VSOCK→socat→TCP:2321): functional correctness baseline.
2. **Physical discrete TPM** — Infineon SLB 9672 SPI module
   (vendor "IFX"/"SLB9672", fw 1.84) on a Gigabyte CTM012 carrier,
   reached via `/dev/tpmrm0` and a host-side TPM-over-VSOCK bridge
   daemon that replaces socat/swtpm on the host side.

Both runs exercise the same L0–L5 guest validation suite and the
same Provision → cold-boot → Recover sequence:

- vTPM identity preserved across cold boot: EK name and EK
  qualified name are **byte-identical** between Provision and
  Recover (`000bdb93ab66...` / `000b0ae7f061...` on SLB 9672;
  `000b92934628...` / `000bfca90d16...` on swtpm).
- Guest-runtime NV writes persist: a user NV index at
  `0x1500016` written under Provision is readable after a
  cold reboot under Recover (Tier-2 / A3 path).
- The A3 runtime re-seal hook fires on guest-side
  `TPM2_Shutdown(SU_STATE)` and commits a fresh SealedBlob
  before SVSM exits (observed in serial log on both endpoints).

Sealed-blob sizes match across backends (~19.1 KB Provision,
~18.8 KB Recover; the observed sub-kilobyte delta between the two
cold-boot samples is consistent with an NV-section repack and is
not endpoint-specific). On SLB 9672 the external-TPM-side latency
is dominated by RSA-2048 `TPM2_CreatePrimary` (~6.6 s per call on
this device); the critical unseal step itself measures ~126 ms.
Per-command latency CSV (bridge-side wall clock) is included in
the artifact dump.

Reproduction artifacts (IGVM SHA-256, Launch Digest, boot/validate
logs, sealed blobs, TPM-endpoint latency CSV) available on request.

All commits are DCO-signed.

### DIFF
--- a/configs/qemu-target.json
+++ b/configs/qemu-target.json
@@ -14,7 +14,7 @@
     },
     "kernel": {
         "svsm": {
-            "features": "vtpm",
+            "features": "vtpm,virtio-drivers,vsock",
             "binary": false
         },
         "bldr": {
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -81,6 +81,26 @@ virtio-drivers = ["dep:virtio-drivers"]
 block = []
 vsock = []
 
+# Persistent vTPM via TPM-sealing of internal state to an off-CVM TPM
+# endpoint (research extension; mainline behaviour preserved when off).
+#
+# When OFF (default): the `vtpm` feature provides the standard ephemeral
+# vTPM that the mainline COCONUT-SVSM tree exposes — each cold boot
+# re-manufactures the simulator and produces a fresh EK/SRK/IAK.
+#
+# When ON: svsm.rs drives the Provision/Recover seal cycle through a
+# `VsockTransport`/`VsockHostStore` pair. The host helper (TPM endpoint
+# on a VSOCK port + blob store on two VSOCK ports) must be reachable;
+# unreachable endpoints are a fatal error, not a silent fallback,
+# because silent fallback would weaken the security model that treats
+# the physical TPM as a separate root of trust from the platform.
+#
+# Transport / store backends are currently fixed to VSOCK because that
+# is the only supported channel between an SEV-SNP CVM and a host-side
+# TPM endpoint; the `TpmTransport` and `SealedBlobStore` traits keep
+# the door open for future block-layer or KBS backends.
+vtpm-persist = ["vtpm", "vsock"]
+
 [dev-dependencies]
 sha2 = { workspace = true, features = ["force-soft"] }
 
--- a/kernel/src/cpu/x86/apic.rs
+++ b/kernel/src/cpu/x86/apic.rs
@@ -157,9 +157,9 @@ impl X86Apic {
     /// - `icr` - Value to write to the ICR register
     #[inline(always)]
     pub fn icr_write(&self, icr: u64) {
-        self.regs()
-            .icr_write(icr)
-            .expect("Failed to write APIC.ICR");
+        if let Err(e) = self.regs().icr_write(icr) {
+            log::warn!("Failed to write APIC.ICR: {e:?}");
+        }
     }
 
     /// Checks whether an IRQ vector is currently in service
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -51,7 +51,7 @@ pub mod virtio;
 pub mod vmm;
 #[cfg(feature = "vsock")]
 pub mod vsock;
-#[cfg(all(feature = "vtpm", not(test)))]
+#[cfg(feature = "vtpm")]
 pub mod vtpm;
 
 #[test]
--- a/kernel/src/protocols/mod.rs
+++ b/kernel/src/protocols/mod.rs
@@ -8,7 +8,7 @@ pub mod apic;
 pub mod attest;
 pub mod core;
 pub mod errors;
-#[cfg(all(feature = "vtpm", not(test)))]
+#[cfg(feature = "vtpm")]
 pub mod vtpm;
 
 extern crate alloc;
--- a/kernel/src/sev/snp_apic.rs
+++ b/kernel/src/sev/snp_apic.rs
@@ -65,15 +65,18 @@ impl ApicAccess for GHCBApicAccessor {
     }
 
     fn icr_write(&self, icr: u64) -> Result<(), SvsmError> {
-        // The #HV IPI can only be used if restricted injection is supported.
-        // Otherwise, the IPI must be sent via an X2APIC ICR write.
-        if self.use_restr_inj() {
-            current_ghcb().hv_ipi(icr)?;
-        } else {
-            self.apic_write(APIC_OFFSET_ICR, icr);
+        // HV_IPI (GHCB exit 0x0015) is a base SEV-ES protocol feature and
+        // does not require restricted injection. KVM rejects direct WRMSR to
+        // the x2APIC ICR MSR (0x830) via the GHCB MSR protocol (error 2),
+        // so HV_IPI is the only reliable path.
+        match current_ghcb().hv_ipi(icr) {
+            Ok(()) => Ok(()),
+            Err(_) => {
+                // Fallback: try direct MSR write if HV_IPI is unsupported
+                self.apic_write(APIC_OFFSET_ICR, icr);
+                Ok(())
+            }
         }
-
-        Ok(())
     }
 
     fn eoi(&self) {
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -72,16 +72,20 @@ use svsm::utils::ScopedMut;
 use svsm::utils::round_to_pages;
 #[cfg(all(feature = "virtio-drivers", any(feature = "block", feature = "vsock")))]
 use svsm::virtio::probe_mmio_slots;
-#[cfg(all(feature = "vtpm", not(feature = "vsock"), not(test)))]
+#[cfg(all(feature = "vtpm-persist", not(feature = "vsock"), not(test)))]
 use svsm::vtpm::MockTransport;
-#[cfg(all(feature = "vtpm", feature = "vsock", not(test)))]
+#[cfg(all(feature = "vtpm-persist", feature = "vsock", not(test)))]
 use svsm::vtpm::VsockTransport;
-#[cfg(all(feature = "vtpm", not(test)))]
-use svsm::vtpm::{SealedBlobStore, StaticBufStore, VtpmBootMode, vtpm_init_sealed};
-#[cfg(all(feature = "vtpm", feature = "vsock", not(test)))]
+#[cfg(all(feature = "vtpm", not(feature = "vtpm-persist"), not(test)))]
+use svsm::vtpm::vtpm_init;
+#[cfg(all(feature = "vtpm-persist", not(test)))]
+use svsm::vtpm::{SealedBlobStore, VsockHostStore, VtpmBootMode, vtpm_init_sealed};
+#[cfg(all(feature = "vtpm-persist", feature = "vsock", not(test)))]
 use svsm::vtpm::{VSOCK_HOST_CID, VSOCK_TPM_PORT};
 
 use alloc::string::String;
+#[cfg(all(feature = "vtpm-persist", not(test)))]
+use alloc::vec::Vec;
 use release::COCONUT_VERSION;
 
 #[cfg(feature = "attest")]
@@ -576,18 +580,37 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
         log::info!("attestation successful");
     }
 
-    #[cfg(all(feature = "vtpm", not(test)))]
+    // Mainline COCONUT-SVSM behaviour: ephemeral vTPM, no external
+    // dependencies, fresh seeds on every cold boot. This is the path
+    // upstream consumers get and the one preserved across the
+    // `vtpm-persist` feature being off.
+    #[cfg(all(feature = "vtpm", not(feature = "vtpm-persist"), not(test)))]
     {
-        // SealedBlob persistence is driven through the `SealedBlobStore`
-        // trait so that the same code path can target any backend
-        // (warm-reboot static buffer for dev/test, IGVM variable, the
-        // upcoming SVSM block layer, KBS-stateful vTPM service, ...).
-        // The default in this tree is the static buffer backend, which
-        // survives warm reboots within a single CVM lifetime.
-        static SEALED_BLOB_STORE: StaticBufStore = StaticBufStore::new();
-        let store: &dyn SealedBlobStore = &SEALED_BLOB_STORE;
+        vtpm_init().expect("vtpm_init failed");
+        log::info!("VTPM: ephemeral init complete");
+    }
 
-        let blob_opt = store.load().expect("SealedBlobStore::load failed");
+    // Persistent vTPM path — opt-in via the `vtpm-persist` feature.
+    // Drives the Provision/Recover seal cycle against an off-CVM TPM
+    // endpoint (transport) and host-side blob store, both currently
+    // reached over VSOCK. The host helper must be reachable: a
+    // missing/unreachable endpoint is a hard error, NOT a silent
+    // fallback to ephemeral — silently downgrading would void the
+    // security model the seal cycle promises, which treats the
+    // physical TPM as a separate root of trust (an attacker who can
+    // break the VSOCK channel would otherwise harvest a fresh vTPM
+    // identity by triggering the fallback).
+    #[cfg(all(feature = "vtpm-persist", not(test)))]
+    {
+        let store = VsockHostStore::new(
+            VSOCK_HOST_CID,
+            VsockHostStore::DEFAULT_LOAD_PORT,
+            VsockHostStore::DEFAULT_SAVE_PORT,
+        );
+
+        let blob_opt = store
+            .load()
+            .expect("VTPM: sealed-store load failed (host helper unreachable?)");
         let (boot_mode, blob_slice) = match blob_opt.as_deref() {
             Some(b) => (VtpmBootMode::Recover, Some(b)),
             None => (VtpmBootMode::Provision, None),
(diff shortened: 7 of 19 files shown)


joergroedel:
Thanks for your submission. Reading through the description above I have a couple of questions around security properties of the host-side parts and actual use-cases for these changes. Can you please join one of the next community calls to discuss these open questions?

Goodleon-Y:
> Thanks for your submission. Reading through the description above I have a couple of questions around security properties of the host-side parts and actual use-cases for these changes. Can you please join one of the next community calls to discuss these open questions?

Hi Joerg, thanks for going through this.

On the call: I'd rather answer in writing here first, so the technical details stay on the list and are easy to reference, and I can be more precise that way. Glad to join a later call if that's still helpful.

On the host-side security. The host components (the VSOCK helper that stores the blob and the bridge to the external TPM) are untrusted, at the same level as the rest of the SEV-SNP host. What the guest relies on today is confidentiality and integrity of the state at rest: the payload is AES-256-GCM encrypted and the AES key is TPM2_Sealed to the external endpoint, so the host only ever handles ciphertext plus a sealed key blob it cannot open. A host that flips bytes in either part fails the GCM tag or the unseal, so silent tampering is caught.

What a host can still do is withhold the blob or hand back an older one. As I flagged in open questions 5 and 8, rollback/replay resistance and channel authentication are the parts I left sketched rather than built in this RFC. The shape I'm aiming for is to bind a monotonic counter from the endpoint's NV into the blob and check it on unseal, so a stale or rolled-back blob fails closed, and to wrap the proxy channel in a salted HMAC session so the host can't splice or forge proxy traffic. I'd be glad to turn those two from a sketch into a concrete design note here if that's the most useful next step. With them in place the host reduces to untrusted storage and transport whose only remaining power is denial of service.

On the use-case, to be more concrete than the list above: the target is an on-prem or edge confidential VM that needs a stable TPM-rooted identity across reboots (its EK/AK, sealed secrets, measured-boot NV) in a setting where there is no KBS and no guarantee the network is up at boot. Today such a VM either loses vTPM state on every cold boot or has to re-enroll out of band. If the platform already has a discrete TPM, this lets that TPM be the persistence anchor without standing up KBS infrastructure. It is meant to sit beside the encrypted-block + KBS path, not replace it.

Happy to go deeper on either. Thanks again.

stefano-garzarella:
@Goodleon-Y thanks for this RFC, the design is interesting, but I have a question about the threat model, specifically around the trust boundary between the CVM and the TPM endpoint.

The core issue is that the TPM endpoint (whether swtpm or a physical TPM) is reachable only through the host. The host mediates every TPM command and response, and the AES-256 key is exposed in plaintext on the VSOCK channel in both directions during the seal (this may be not a problem if the sealing is done in a trusted environment) but also during the unseal IIUC.

This is not just a VSOCK problem: encrypting the VSOCK channel (e.g. with TLS) would not help, because the host-side process that terminates the channel (the socat/bridge to `/dev/tpm0` or swtpm) still handles the commands and responses in cleartext before forwarding them. The host is root on that process, so it can read its memory, ptrace it, or replace it entirely.

Could you clarify the intended threat model?

1. Is the host assumed to be trusted for confidentiality (just not for integrity, which AES-GCM covers)?
2. If not, how would you prevent the host from reading the unsealed key material? IIUC in the current architecture the TPM endpoint terminates on the host, so any channel encryption just moves the plaintext exposure from shared memory to the bridge process.
3. For the provisioning-on-a-trusted-host scenario: even if the blob is sealed to a specific physical TPM, the host retains the ability to unseal it at any time (especially with the current zero PCR policy). Is this considered acceptable?

For comparison, in the KBS-based approach the CVM generates a key pair, includes the public key in the SNP attestation report (which is signed by the hardware), and the KBS encrypts the secret to that public key after verifying the attestation. The host relays the encrypted blob but never sees the plaintext key, because only the CVM holds the private key.

I think understanding the intended trust boundary would help evaluate where this approach fits and what security guarantees it actually provides.

As Joerg suggested, I think joining one of the community calls would be really valuable to discuss these points. It would also be a good chance to get broader input from other contributors.

Goodleon-Y:
@stefano-garzarella, thanks for the careful read, this is exactly the boundary that matters. You're right, in the RFC the AES key is sealed to the TPM, so on unseal it comes back through the host bridge in cleartext, which makes the host trusted for confidentiality. I've reworked the key schedule so it isn't.

The AES key is now derived inside the CVM from two independent halves, and is never sealed or put on the channel:

```
aes_key = HKDF(salt ; k_psp* ‖ k_dtpm)
```

`k_dtpm` (32B random) is the only half sealed to the TPM, and the only key material that crosses the channel; assume the host unseals it at will, on its own it is useless. `k_psp*` is derived in-CVM via `MSG_KEY_REQ` from the SNP platform key, bound to `GUEST_FIELD_MEASUREMENT` at VMPL0. It never leaves the CVM, never hits the channel, and never touches the TPM.

So sniffing VSOCK, reading the bridge process, or unsealing the TPM blob all yield only `k_dtpm`. On your three questions: (1) no, the host is no longer trusted for confidentiality; (2) the only key material it can reach is `k_dtpm`; the protecting half is PSP-derived in-CVM and never crosses the boundary; (3) yes, the host can unseal the TPM part whenever it likes, and that is now harmless: the measurement binding moved out of the host-controlled PCR policy into the SNP derivation of `k_psp*`.

The host can't derive `k_psp*` itself: `MSG_KEY_REQ` is VMPCK-authenticated, so it can't issue it for the victim guest, and the MEASUREMENT mixing means any guest it does launch derives a different key. It is the same "host never sees the key" property as the KBS path, rooted in the platform key rather than a broker, which is the point when there is no broker to reach.

Validated on real SEV-SNP with a discrete TPM (Infineon SLB 9672), not swtpm: `k_psp*` derives through `MSG_KEY_REQ` against the PSP, re-derives byte-identically across a cold boot (a clean unseal is the witness, since the GCM tag only verifies on a matching key), and a build with a different launch measurement cannot recover another measurement's blob, so it fails closed.

The known cost of binding to MEASUREMENT is that an SVSM/TCB upgrade changes `k_psp*`, so state has to be migrated to the new measurement across an upgrade rather than recovered directly.

So it stays complementary to the encrypted-block + KBS path, for deployments with no online broker. Happy to take this to a call, and to share the reworked design or branch if that's useful.

stefano-garzarella:
@Goodleon-Y But if `k_psp*` already provides confidentiality, measurement binding, and platform binding... what does `k_dtpm` add?

Could we just encrypt the vTPM state with the PSP-derived key and store it anywhere, without involving an external TPM at all?

Goodleon-Y:
@stefano-garzarella  For confidentiality against the host, `k_psp*` is enough on its own and `k_dtpm` adds nothing there. It's there for two other reasons.

Rollback: SNP derivation gives you a key but no host-resettable monotonic state, so encrypting with the PSP key and storing the blob anywhere doesn't stop the host withholding it or serving an older one. The NV counter does: we bind its value into the blob and require strict equality on unseal, so a stale or rolled-back blob fails closed. That's why it's a TPM and not just a file.

Second root: `k_dtpm` keeps the other half of the key in the discrete TPM, independent of the PSP's derivation, so the state is exposed only if both roots break rather than the PSP alone. On a platform that already trusts a discrete TPM, anchoring the long-term vTPM identity there too costs no extra hardware, and the host driving the TPM still only sees that sealed half.

So encrypt-with-PSP-and-store is valid and simpler if you accept being rollback-able and resting the long-term identity's confidentiality on the single PSP root. The dTPM removes both, with hardware on-prem and edge boxes already have; its cost (MEASUREMENT binding, so upgrades need a migration) I noted before.

stefano-garzarella:
> @stefano-garzarella For confidentiality against the host, `k_psp*` is enough on its own and `k_dtpm` adds nothing there. It's there for two other reasons.
> 
> Rollback: SNP derivation gives you a key but no host-resettable monotonic state, so encrypting with the PSP key and storing the blob anywhere doesn't stop the host withholding it or serving an older one. The NV counter does: we bind its value into the blob and require strict equality on unseal, so a stale or rolled-back blob fails closed. That's why it's a TPM and not just a file.

Sorry, the TPM is host-controlled, no? So how can you trust the NV counter? As you mention the TPM can be a software TPM emulated by the host, so the host can do whatever, no?

> Second root: `k_dtpm` keeps the other half of the key in the discrete TPM, independent of the PSP's derivation, so the state is exposed only if both roots break rather than the PSP alone. On a platform that already trusts a discrete TPM, anchoring the long-term vTPM identity there too costs no extra hardware, and the host driving the TPM still only sees that sealed half.

Okay, so you must trust the TPM attached to the host. How?

joergroedel:
No response from submitter. Closing for now, please re-open if still relevant.

PR #1093, opened 2026-05-26
Use TOML formatter

osteffenrh:
Add a style for the TOML formatter `Taplo`, explain how
to use in in the docs, and reformat all `.toml` files
with it.

The style is what seems to be the most common standard in
the Rust ecosystem, Taplo default, 4 spaces indentaton,
comact arrays.

I hope to establish this as the default for all TOML files
here, which have very inconsistent formatting.

I did not add Taplo to the CI since it has a surprising
amount of dependencies and takes too long to install IMO.
(Maybe with some cargo caching?)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ aes-gcm = { version = "0.10.3", default-features = false }
 aes-kw = { version = "0.3.1", default-features = false }
 arbitrary = "1.3.0"
 base64 = { version = "0.22.1", default-features = false }
-bindgen = {version = "0.71.0", default-features = false }
+bindgen = { version = "0.71.0", default-features = false }
 bitfield-struct = "0.6.2"
 bitflags = "2.6"
 clap = { version = "4.4.14", default-features = false }
@@ -108,13 +108,18 @@ zeroize = { version = "1.8.2", default-features = false }
 
 # Verus repos
 verus_builtin = { version = "=0.0.0-2026-05-17-0151", default-features = false }
-verus_builtin_macros = { version = "=0.0.0-2026-07-12-0122", features = ["vpanic"], default-features = false }
+verus_builtin_macros = { version = "=0.0.0-2026-07-12-0122", features = [
+    "vpanic",
+], default-features = false }
 verus_state_machines_macros = { version = "=0.0.0-2026-06-14-0213", default-features = false }
-vstd = { version = "=0.0.0-2026-07-12-0122", features = ["alloc", "allow_panic"], default-features = false }
+vstd = { version = "=0.0.0-2026-07-12-0122", features = [
+    "alloc",
+    "allow_panic",
+], default-features = false }
 
 # Verification libs
-verify_proof = { path = "verification/verify_proof", default-features = false  }
-verify_external = { path = "verification/verify_external", default-features = false  }
+verify_proof = { path = "verification/verify_proof", default-features = false }
+verify_external = { path = "verification/verify_external", default-features = false }
 
 [patch.crates-io]
 # Force all dependencies to use the same version of safe-mmio, which has
--- a/taplo.toml
+++ b/taplo.toml
@@ -1,0 +1,19 @@
+# Exclude all git submodules
+exclude = ["libtcgtpm/deps/openssl/**", "libtcgtpm/deps/tpm-20-ref/**", "packit/**"]
+
+[formatting]
+
+# Use same indentation as Rust code = 4 spaces
+indent_string = "    "
+
+# Alphabetize dependencies.
+# Not enabling, to preserve grouping
+reorder_keys = false
+
+# No space after/before [ / ]
+compact_arrays = true
+
+# Keep arrays in one line if they fit
+array_auto_collapse = true
+
+column_width = 100
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,16 +2,18 @@
 
 [target.x86_64-unknown-none]
 rustflags = [
-	"-C", "force-frame-pointers",
-	"--cfg", "aes_force_soft",
-	"--cfg", 'aes_backend="soft"',
-	"--cfg", "polyval_force_soft",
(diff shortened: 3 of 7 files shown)


Documentation/docs/developer/CONTRIBUTING.md
```diff
@@ -49,6 +49,22 @@ before allowing them to be committed. It can be installed by running
...
+```bash
+cargo install taplo-cli
+```
+
+To format all `.toml` files in the repository run:
+
+```bash
+taplo fmt
+```
+
```
joergroedel: I don't think we want contributors to run that on their changes or checking it in CI. The formatting cleanups are nice and can be merged, but it is fine to run this regularly for cleanup. Can you please this documentation update? The rest is fine and can be merged.

osteffenrh:
v2:
- rebased
- Mention toml formatting in a new section in the doc.
  No longer requiring it for contribution.

osteffenrh:
forgot to add a reformatted toml file. sorry.

osteffenrh:
rebased.

taplo.toml
stefano-garzarella, line 1: Is there any `exclude` patterns we can specify?

I'm worried about git submodules like packit, openssl, tpm, etc.

Cargo.toml
```diff
@@ -105,13 +105,18 @@ zeroize = { version = "1.8.2", default-features = false }
 
 # Verus repos
 verus_builtin = { version = "=0.0.0-2026-04-12-0118", default-features = false }
-verus_builtin_macros = { version = "=0.0.0-2026-04-12-0118", features = ["vpanic"], default-features = false }
+verus_builtin_macros = { version = "=0.0.0-2026-04-12-0118", features = [
+    "vpanic",
+], default-features = false }
```
stefano-garzarella: Do you know which rule is enforcing this?
I quite prefer to have everything in a single line.

taplo.toml
osteffenrh: You want to skip formatting toml files in sub-modules? Makes sense.
Could use this:
```bash
taplo fmt $(git ls-files -co "*.toml")
```

Cargo.toml
osteffenrh: We can either increase the `column_width` and allow longer lines, but this will wrap it at some length anyway; or never break arrays into multiple lines (allow any line length) `array_auto_expand=false`

taplo.toml
stefano-garzarella: I'm fine also with this, but I think we should document it.

I was thinking more of using https://taplo.tamasfe.dev/configuration/file.html#exclude
So we can set the path to exclude, and we can continue to use `taplo fmt`.

No big preference on my side.

Cargo.toml
stefano-garzarella: What about setting `column_width` to 100 like default value of `max_width` for rustfmt? https://rust-lang.github.io/rustfmt/?version=v1.9.0&search=#max_width

No preference about `array_auto_expand`, as you prefer.

taplo.toml
osteffenrh: the `exclude` directive is a good idea. Will do.

Cargo.toml
osteffenrh: :heavy_check_mark: `column_width = 100`

> No preference about array_auto_expand, as you prefer.

Do you want everything on one line *always*, or just when it fits in to 100 chars?

osteffenrh:
v3:
- line length = 100
- exclude git submodules (explicitly listed in `taplo.toml`)
- try to keep arrays on same line

Cargo.toml
osteffenrh: in v3, the line you commented here will still be broken up (>100 chars)

stefano-garzarella:
> I did not add Taplo to the CI since it has a surprising
> amount of dependencies and takes too long to install IMO.
> (Maybe with some cargo caching?)

In the TSC call we also discussed eventually to use https://github.com/cargo-bins/cargo-binstall but now that I'm writing, IIRC @joergroedel pointed out some security reports about installing binaries in the CI.

Cargo.toml
stefano-garzarella: > ✔️ `column_width = 100`
> 
> > No preference about array_auto_expand, as you prefer.
> 
> Do you want everything on one line _always_, or just when it fits in to 100 chars?

not sure :-) as you prefer.

osteffenrh:
> > I did not add Taplo to the CI since it has a surprising
> > amount of dependencies and takes too long to install IMO.
> > (Maybe with some cargo caching?)
> 
> In the TSC call we also discussed eventually to use https://github.com/cargo-bins/cargo-binstall but now that I'm writing, IIRC @joergroedel pointed out some security reports about installing binaries in the CI.

Ok, let's postpone adding toml format checking to the CI.

Cargo.toml
osteffenrh: Then let's keep it as is.

Documentation/docs/developer/CONTRIBUTING.md
```diff
@@ -82,3 +82,24 @@ Fuzzing
 The SVSM project includes a number of fuzzing targets to test parts of the
 code-base. For details on how to run the fuzzers and extend the fuzzing
 functionality, please have a look at [FUZZING.md](FUZZING.md).
+
+TOML Formatting
+---------------
```
00xc: We discussed this again in the TSC call. The agreement was to not have any updates to `CONTRIBUTING.md` and leave everything else as-is. Later we can explore in a different PR how to enforce the formatting in CI, if at all.

00xc (changes requested):

Documentation/docs/developer/CONTRIBUTING.md
osteffenrh: Removed.

osteffenrh:
v4:
- Removed update to CONTRIBUTING.md
- rebased

00xc (approved):

PR #1095, opened 2026-05-28
kernel/attest: disable default fallback on serial port

luigix25:
This PR makes vsock the default transport mechanism for attestation and makes the serial fallback as an opt-in feature behind the `attest-serial` rust feature.

To summarize:
`make FEATURES=attest,virtio-drivers` will try to use vsock as a transport, if it's not available it will fail.
`make FEATURES=attest,virtio-drivers,attest-serial` will try to use vsock as a transport first, if it's not available it will fallback using the serial port.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -71,7 +71,8 @@ zeroize = { workspace = true, features = ["alloc", "derive"] }
 test.workspace = true
 
 [features]
-attest = ["dep:aes-kw", "dep:cocoon-tpm-crypto", "dep:cocoon-tpm-tpm2-interface", "dep:cocoon-tpm-utils-common", "dep:concat-kdf", "dep:kbs-types", "dep:libaproxy", "dep:serde", "dep:serde_json"]
+attest = ["vsock", "dep:aes-kw", "dep:cocoon-tpm-crypto", "dep:cocoon-tpm-tpm2-interface", "dep:cocoon-tpm-utils-common", "dep:concat-kdf", "dep:kbs-types", "dep:libaproxy", "dep:serde", "dep:serde_json"]
+attest-serial = ["attest"]
 
 default = []
 enable-gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
--- a/Documentation/docs/developer/ATTESTATION.md
+++ b/Documentation/docs/developer/ATTESTATION.md
@@ -247,10 +247,12 @@ launching the proxy. The supported backend attestation protocols include:
 
 SVSM communicates with the attestation proxy using one of two transport methods:
 
-- **vsock**: When the `vsock` feature is enabled, SVSM will first try to use vsock for communication with the host proxy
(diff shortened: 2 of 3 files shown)


kernel/Cargo.toml
```diff
@@ -80,6 +80,7 @@ noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
 block = []
 vsock = []
+no-serial-fallback = []
```
stefano-garzarella: The feature name is unclear IMO, it should refer to `attest` in some way.

stefano-garzarella (commented):
> When compiling SVSM with the vsock support, this becomes the preferred transport for the attestation proxy. If, for any reason, the attestation fails using vsock, it will try again using the serial port.
> 
> Add a new feature `no-serial-fallback` that selectively disables this fallback. In other words:
> 
>     * `attest` = only transport mechanism is the serial port
> 
>     * `attest` + `no-serial-fallback` = only transport mechanism is the serial port (no change)
> 
>     * `attest` + `vsock` = main transport is vsock that falls back on the serial port
> 
>     * `attest` + `vsock` + `no-serial-fallback` = only transport mechanism is vsock

IMO we can simplify this. Just add a `attest-serial` to support the serial port or not (disabled by default). And we should mark it deprecated or unsupported. I think we can also consider removing it at all.

luigix25:
> > When compiling SVSM with the vsock support, this becomes the preferred transport for the attestation proxy. If, for any reason, the attestation fails using vsock, it will try again using the serial port.
> > Add a new feature `no-serial-fallback` that selectively disables this fallback. In other words:
> > ```
> > * `attest` = only transport mechanism is the serial port
> > 
> > * `attest` + `no-serial-fallback` = only transport mechanism is the serial port (no change)
> > 
> > * `attest` + `vsock` = main transport is vsock that falls back on the serial port
> > 
> > * `attest` + `vsock` + `no-serial-fallback` = only transport mechanism is vsock
> > ```
> 
> IMO we can simplify this. Just add a `attest-serial` to support the serial port or not (disabled by default). And we should mark it deprecated or unsupported. I think we can also consider removing it at all.

Ok, this simplifies the logic a little bit

I have a question though: If attestation on serial port is disabled by default, should `attest` depend on `vsock`? otherwise we would have no default transport.

I guess this also implies that serial port would be just a fallback?

stefano-garzarella:
> I have a question though: If attestation on serial port is disabled by default, should `attest` depend on `vsock`? otherwise we would have no default transport.
> 
> I guess this also implies that serial port would be just a fallback?

Yep, I'm unsure whether we need to support serial-only attestation (without vsock). I'd lean toward making attest depend on vsock to keep things simple. We can always add serial-only support later if needed.

I think we should also document that serial port is just for testing/debugging for now.

luigix25:
[@stefano-garzarella]
- vsock is now the default transport mechanism
- serial is opt-in fallback when vsock is not available
- updated doc
- rebased to latest main

stefano-garzarella (approved):

Issue #807, opened 2025-09-15
Attestation: Zeroize secrets properly after use


osteffenrh:
Zero out all secrets and keys after use in `kernel/src/attest.rs` (and everywhere else).

PR #806 wraps the `Vec<>` containing secrets in a `Zeroizing`, but that does not take care of potential reallocations from within the `Vec`, see https://github.com/coconut-svsm/svsm/pull/806/files#r2343467453

We could use a container that does not automatically reallocate the data buffer, or create a specialized data type for secrets that zeroes itself automatically.

nicstange:
IIUC, the secret decryption part of the attestation code will receive further enhancements in the future anyway, c.f. https://github.com/coconut-svsm/svsm/pull/528#discussion_r2163386653

It would probably be good to determine the final length upfront then, if possible, and to avoid any subsequent reallocations.

stefano-garzarella:
> We could use a container that does not automatically reallocate the data buffer, or create a specialized data type for secrets that zeroes itself automatically.

From https://docs.rs/zeroize/latest/zeroize/#stackheap-zeroing-notes: The secrecy crate provides higher-level abstractions for eliminating usage patterns which can cause reallocations: https://crates.io/crates/secrecy

Maybe we can take a look at it, instead of creating something custom. @osteffenrh WDYT ?

stefano-garzarella:
#1097 improved a bit the status, but we need to complete it. E.g. in `decrypt()` we should do the same with keys, etc.

PR #1097, opened 2026-05-30
kernel/attest: add SecretSlice for zeroed attestation secrets

tanish111:
Introduce SecretSlice, a Box<[u8]> wrapper that zeroes its backing memory on drop, and use it to hold the secret returned from the attestation proxy in place of a plain Vec. This ensures the decrypted key material released after a successful attestation is wiped from memory once it goes out of scope.

SecretSlice takes ownership of an existing heap buffer via a From<Box<[u8]>> impl without copying, relies on zeroize's ZeroizeOnDrop derive for drop-time clearing, and implements Debug by hand to redact its contents. The attestation path moves the ciphertext into the SecretSlice and decrypts in place, reusing the same allocation. Unit tests cover construction, slice access and zeroization.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
+zeroize = { version = "1.8.2", default-features = false }
 embedded-io = { version = "0.6.1" }
 
 # Verus repos
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -62,6 +62,7 @@ vstd = { workspace = true, optional = true}
 verify_proof = { workspace = true, optional = true}
 verify_external = { workspace = true, optional = true}
 verus_stub = { workspace = true }
+zeroize = { workspace = true, features = ["alloc", "derive"] }
 
 [target."x86_64-unknown-none".dev-dependencies]
 test.workspace = true
--- a/kernel/src/attest.rs
+++ b/kernel/src/attest.rs
@@ -8,6 +8,7 @@
 extern crate alloc;
 
 use crate::{
+    crypto::SecretSlice,
     error::SvsmError,
     greq::{pld_report::*, services::get_regular_report},
     io::{DEFAULT_IO_DRIVER, Read, Write},
@@ -128,7 +129,7 @@ impl TryFrom<Tee> for AttestationDriver<'_> {
 
 impl AttestationDriver<'_> {
     /// Attest SVSM's launch state by communicating with the attestation proxy.
-    pub fn attest(&mut self) -> Result<Vec<u8>, SvsmError> {
+    pub fn attest(&mut self) -> Result<SecretSlice, SvsmError> {
         let negotiation = self.negotiation()?;
 
         Ok(self.attestation(negotiation)?)
@@ -152,7 +153,7 @@ impl AttestationDriver<'_> {
     /// Send an attestation request to the proxy. Proxy should reply with attestation response
     /// containing the status (success/fail) and an optional secret returned from the server upon
     /// successful attestation.
-    fn attestation(&mut self, n: NegotiationResponse) -> Result<Vec<u8>, AttestationError> {
+    fn attestation(&mut self, n: NegotiationResponse) -> Result<SecretSlice, AttestationError> {
         let curve =
             Curve::new(self.ecc.pub_key().get_curve_id()).map_err(AttestationError::Crypto)?;
 
@@ -185,13 +186,18 @@ impl AttestationDriver<'_> {
             return Err(AttestationError::PublicKeyMissing)?;
         };
 
-        let Some(mut secret) = response.secret else {
+        let Some(secret_enc) = response.secret else {
             return Err(AttestationError::SecretMissing);
         };
 
-        self.decrypt(&mut secret, decryption)?;
+        // `secret_enc` holds ciphertext from the attestation server. Move it
+        // into `SecretSlice` before decryption so the same buffer is decrypted
+        // in-place and zeroed on drop.
+        let mut secret_slice = SecretSlice::from(secret_enc.into_boxed_slice());
 
-        Ok(secret)
+        self.decrypt(&mut secret_slice, decryption)?;
+
+        Ok(secret_slice)
     }
 
     /// Decrypt a secret from the attestation server with the TEE private key. Secrets are
--- a/kernel/src/crypto/mod.rs
+++ b/kernel/src/crypto/mod.rs
@@ -6,6 +6,47 @@
 
 //! SVSM kernel crypto API
 
+extern crate alloc;
+
+use alloc::boxed::Box;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+/// Container for sensitive byte data.
+///
+/// The owned buffer is zeroed when the `SecretSlice` is dropped or explicitly
+/// [`zeroize`](Zeroize::zeroize)d.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct SecretSlice(Box<[u8]>);
+
+impl From<Box<[u8]>> for SecretSlice {
+    fn from(data: Box<[u8]>) -> Self {
+        Self(data)
+    }
+}
+
+impl core::ops::Deref for SecretSlice {
+    type Target = [u8];
+
+    fn deref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl core::ops::DerefMut for SecretSlice {
+    fn deref_mut(&mut self) -> &mut [u8] {
+        &mut self.0
+    }
+}
+
+impl core::fmt::Debug for SecretSlice {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("SecretSlice")
+            .field("len", &self.0.len())
+            .field("data", &"[REDACTED]")
+            .finish()
+    }
+}
+
 pub mod aead {
     //! API for authentication encryption with associated data
 
@@ -102,3 +143,40 @@ pub mod digest {
 // Crypto implementations supported. Only one of them must be compiled-in.
 
 pub mod rustcrypto;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloc::vec::Vec;
+    use zeroize::Zeroize;
+
+    #[test]
+    fn secretslice_stores_secret() {
+        let b: SecretSlice = Vec::from([1u8, 2, 3, 4, 5]).into_boxed_slice().into();
+        assert_eq!(b.len(), 5);
+        assert_eq!(&*b, &[1, 2, 3, 4, 5]);
+    }
+
+    #[test]
+    fn secretslice_empty_ok() {
+        let b: SecretSlice = Vec::new().into_boxed_slice().into();
+        assert_eq!(b.len(), 0);
+        assert!(b.is_empty());
+    }
+
+    #[test]
+    fn secretslice_deref_mut_mutates() {
+        let mut b: SecretSlice = Vec::from([1u8, 2, 3, 4]).into_boxed_slice().into();
+        b[0] = 42;
+        assert_eq!(&*b, &[42, 2, 3, 4]);
+    }
+
+    #[test]
+    fn secretslice_zeroize_wipes_data() {
+        let mut b: SecretSlice = Vec::from([1u8, 2, 3, 4, 5, 6, 7, 8])
+            .into_boxed_slice()
+            .into();
+        b.zeroize();
+        assert!(b.iter().all(|&byte| byte == 0));
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1968,6 +1968,7 @@ dependencies = [
  "virtio-drivers",
  "vstd",
  "zerocopy",
+ "zeroize",
 ]
 
 [[package]]
@@ -2648,6 +2649,20 @@ name = "zeroize"
 version = "1.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
+dependencies = [
+ "zeroize_derive",
+]
+
+[[package]]
+name = "zeroize_derive"
+version = "1.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "85a5b4158499876c763cb03bc4e49185d3cccbabb15b33c627f7884f43db852e"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
 
 [[package]]
 name = "zerotrie"


kernel/src/utils/util.rs
```diff
@@ -4,11 +4,14 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
+extern crate alloc;
+
 use crate::address::{Address, VirtAddr};
 use crate::types::PAGE_SIZE;
+use alloc::vec::Vec;
 use core::ops::{Add, BitAnd, Not, Sub};
-
```
luigix25, line 10: please avoid unrelated changes

kernel/Cargo.toml
```diff
@@ -62,6 +62,7 @@ vstd = { workspace = true, optional = true}
 verify_proof = { workspace = true, optional = true}
 verify_external = { workspace = true, optional = true}
 verus_stub = { workspace = true }
+zeroize = { version = "1.8.2", default-features = false, features = ["alloc", "derive"] }
```
luigix25: in SVSM we put all the dependencies in the root Cargo.toml and here use "workspace = true".
Please do the same

kernel/src/attest.rs
```diff
@@ -128,7 +129,7 @@ impl TryFrom<Tee> for AttestationDriver<'_> {
 
 impl AttestationDriver<'_> {
     /// Attest SVSM's launch state by communicating with the attestation proxy.
-    pub fn attest(&mut self) -> Result<Vec<u8>, SvsmError> {
+    pub fn attest(&mut self) -> Result<SecretBox<512>, SvsmError> {
```
luigix25: what about defining a constant?

kernel/src/attest.rs
```diff
@@ -152,7 +153,7 @@ impl AttestationDriver<'_> {
     /// Send an attestation request to the proxy. Proxy should reply with attestation response
     /// containing the status (success/fail) and an optional secret returned from the server upon
     /// successful attestation.
-    fn attestation(&mut self, n: NegotiationResponse) -> Result<Vec<u8>, AttestationError> {
+    fn attestation(&mut self, n: NegotiationResponse) -> Result<SecretBox<512>, AttestationError> {
```
luigix25: ditto

kernel/src/attest.rs
```diff
@@ -185,13 +186,15 @@ impl AttestationDriver<'_> {
             return Err(AttestationError::PublicKeyMissing)?;
         };
 
-        let Some(mut secret) = response.secret else {
+        let Some(secret) = response.secret else {
             return Err(AttestationError::SecretMissing);
         };
 
-        self.decrypt(&mut secret, decryption)?;
+        let mut secret_box = SecretBox::<512>::new(secret)?;
```
luigix25: ditto

kernel/src/attest.rs
```diff
@@ -185,13 +186,15 @@ impl AttestationDriver<'_> {
             return Err(AttestationError::PublicKeyMissing)?;
         };
 
-        let Some(mut secret) = response.secret else {
+        let Some(secret) = response.secret else {
```
luigix25: mmh, `response.secret` is a Vec<u8>, so we could still have some secret dangling?

luigix25 (changes requested):

kernel/src/attest.rs
tanish111: yes we should use constant it's already used 3 times.

kernel/src/attest.rs
tanish111: noted.

kernel/src/attest.rs
tanish111: @luigix25 `response.secret` is Vec but we will not have secret dangling bcz as soon as it's copied to SecretBox we use zeroize in this Vec.
Check this Diff:- line-138 in kernel/src/utils/util.rs

kernel/src/utils/util.rs
tanish111: noted.

kernel/src/attest.rs
tanish111: noted.

kernel/Cargo.toml
tanish111: noted.

kernel/src/attest.rs
luigix25: true, but someone had to write into that Vec no? we don't know if it has been reallocated in the past.

I'd like to hear more opinions though

stefano-garzarella:
> Related Issue:- #807

Please use one of the supported keyword by GH (see https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue). Usually we use `Closes: #807` in the PR description and/or in the commit message.

kernel/src/attest.rs
stefano-garzarella: or `type AttestSecret = SecretBox<512>;`

kernel/src/utils/util.rs
```diff
@@ -110,10 +113,55 @@ macro_rules! BIT_MASK {
...
+
+/// Secret Box which zeros our memory on drop. Useful for storing sensitive data like cryptographic keys.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct SecretBox<const N: usize> {
+    data: [u8; N],
+    len: usize,
+}
+
+impl<const N: usize> SecretBox<N> {
+    pub fn new(mut secret: Vec<u8>) -> Result<Self, SecretBoxError> {
```
stefano-garzarella: Please add documentation on `pub` methods, especially on this one, since is not documented that this call will zero the `secret` vec.

kernel/src/attest.rs
stefano-garzarella: IIUC `response.secret` is the secret still encrypted, so should be fine to have dangling copies.

luigix25: Update: we might have the secret dangling, _but_ it's still encrypted. So it's not a problem

kernel/src/attest.rs
stefano-garzarella: What is unclear to me is where 512 comes from.

kernel/src/utils/util.rs
tanish111: noted.

kernel/src/attest.rs
tanish111: @luigix25 another possible option is to change AttestationResponse 
https://github.com/coconut-svsm/svsm/blob/e972af7c4a6dac039771e687eb17cdb1cb6d9b1b/libaproxy/src/attestation.rs#L163

In this case we would need to define our custom Serialization and deserialization for SecretBox type.

kernel/src/attest.rs
tanish111: @stefano-garzarella For attestation, secrets are usually small: 16 bytes is enough for a 128-bit random secret, 32 bytes is common for AES-256 keys, and even multiple keys or bootstrap credentials are often well under 100 bytes. A 512-byte limit therefore provides plenty of headroom while preventing unbounded allocations during deserialization.

tanish111: It is very well possible to use response.secret.len instead of 512.

stefano-garzarella: I'm confused, why we need `SecretBox` at all. Why we can't just use `zeroize::Zeroizing<Vec<u8>` like for example `ecdh_c_1_1_cdh_compute_z()` does?

tanish111: @stefano-garzarella The motivation came from the zeroize docs [1]: Zeroizing<Vec<u8>> wipes the current buffer, but previous buffers created during reallocations may still contain plaintext data. SecretBox avoids reallocations entirely, which is the concern raised in [#807](https://github.com/coconut-svsm/svsm/issues/807).

[1] https://docs.rs/zeroize/latest/zeroize/

tanish111: Currently, the codebase does not modify the decrypted secret after it is created, so Zeroizing<Vec<u8>> would likely be sufficient. My concern is more about future changes: using a fixed-capacity type helps ensure we don't accidentally introduce a code path that causes reallocations and defeats the purpose of zeroizing sensitive data.

stefano-garzarella: > @stefano-garzarella For attestation, secrets are usually small: 16 bytes is enough for a 128-bit random secret, 32 bytes is common for AES-256 keys, and even multiple keys or bootstrap credentials are often well under 100 bytes. A 512-byte limit therefore provides plenty of headroom while preventing unbounded allocations during deserialization.

@tanish111 this limitation should at least be documented somewhere, but are we sure KBS is using any limitation on the secret size ?

stefano-garzarella: What about https://crates.io/crates/secrecy I mentioned in https://github.com/coconut-svsm/svsm/issues/807#issuecomment-4612515545 ?

kernel/src/utils/util.rs
stefano-garzarella: That said, why we need to zero the original `secret` vec ?

kernel/src/attest.rs
tanish111: @stefano-garzarella We could use [SecretBox](https://docs.rs/secrecy/0.10.3/secrecy/struct.SecretBox.html) from the secrecy crate, but I don't see enough benefit to justify introducing an additional dependency for something we can implement straightforwardly ourselves.

tanish111: Any thoughts @joergroedel / @luigix25 ?

luigix25: It's a small crate, brings no new dependencies, so I'd lean towards `secrecy`. Not a strong opinion

stefano-garzarella: I honestly still don't understand why we need `SecretBox` at all.

Can we just do:
``` 
let mut secret = Zeroizing::new(secret.into_boxed_slice());
self.decrypt(&mut secret, decryption)?;
Ok(secret)
``` 

And return `Zeroizing<Box<[u8]>>`

kernel/src/utils/util.rs
tanish111: @stefano-garzarella In this case the source Vec contains encrypted data, so zeroizing it is not strictly necessary. The goal was simply to ensure that once the plaintext secret enters SecretBox and no extra copies remain in memory.

kernel/src/attest.rs
tanish111: @luigix25 @joergroedel can you help me answer this I think we had a discussion around this approach. 
One possible reason I could recall is that it provides a reusable abstraction for handling sensitive data and could be generalized across repositories, rather than having each call site manually wrap secrets in Zeroizing<Box<[u8]>>.

stefano-garzarella: From https://docs.rs/zeroize/latest/zeroize/trait.Zeroize.html#impl-Zeroize-for-Box%3C%5BZ%5D%3E

> Unlike Vec, Box<[Z]> cannot reallocate, so we can be sure that we are not leaving values on the heap.

tanish111: @stefano-garzarella Yes, `Zeroizing<Box<[u8]>>` is definitely a valid approach. My thinking was more from a code organization perspective: having a dedicated SecretBox type gives us a single place to encapsulate behavior around sensitive data and makes it easier to evolve that behavior in the future without touching every call site. That said, if Zeroizing<Box<[u8]>> is preferred, I'm happy to switch to that.

kernel/src/utils/util.rs
stefano-garzarella: So, if it's not necessary, let's avoid that.

kernel/src/attest.rs
stefano-garzarella: I'm fine if we want that kind of abstraction, but we should not put arbitrary limit IMO, especially not documented. That said, I don't see a big difference on wrapping secrets from `Zeroizing<Box<[u8]>>` to `SecretBox<512>` except the second way add a limit to the slice.

stefano-garzarella: > @stefano-garzarella Yes, `Zeroizing<Box<[u8]>>` is definitely a valid approach. My thinking was more from a code organization perspective: having a dedicated SecretBox type gives us a single place to encapsulate behavior around sensitive data and makes it easier to evolve that behavior in the future without touching every call site. That said, if Zeroizing<Box<[u8]>> is preferred, I'm happy to switch to that.

Just to be clear, I like that kind of abstraction, but I don't like the fixed slice size.

tanish111: @stefano-garzarella We could keep the abstraction while removing the hardcoded fixed-size limit:

```rust
pub struct SecretBox(Zeroizing<Box<[u8]>>);
```

This preserves the benefits of having a dedicated type for sensitive data, provides a single place to encapsulate secret-handling behavior, and avoids introducing an arbitrary size limit such as `512`.

stefano-garzarella: @tanish111 I prefer to use `#[derive(Zeroize, ZeroizeOnDrop)]` instead of encapsulating Zeroizing, but I'd like to see the final code. In general, I’d avoid adding code that isn’t strictly necessary, but if you think it might be useful in the future and is general enough, that’s fine. I’d like to see it used in more places if possible.

kernel/src/utils/util.rs
tanish111: fixed.

kernel/src/attest.rs
tanish111: removed the limitation.

kernel/src/utils/util.rs
tanish111: removed.

kernel/src/attest.rs
tanish111: @stefano-garzarella I have updated the PR with recommended changes.

kernel/Cargo.toml
tanish111: updated.

tanish111:
@stefano-garzarella @luigix25 I have addressed most of the review comments and pushed the updated code. Could you please take another look?

stefano-garzarella:
> @stefano-garzarella @luigix25 I have addressed most of the review comments and pushed the updated code. Could you please take another look?

Thanks, but looking at the patch, `SecretBox` seems just a wrapper of `Zeroizing<Box<[u8]>>`. IMO we can avoid adding all of this code and just do what I suggested in https://github.com/coconut-svsm/svsm/pull/1097#discussion_r3348825416. It's a small patch, easy to understand, easy to maintain, etc.

If in the future we find some new reason to add a custom struct, we can add it, but for now I don't see any value in adding `SecretBox`. I think the question we should answer is:

    What `SecretBox` is adding that we can't achieve with just `Zeroizing<Box<[u8]>>`?

@luigix25 @joergroedel @osteffenrh WDYT?

tanish111:
@stefano-garzarella `Debug` is something we’ve implemented on top of `SecretBox `, ensuring the data is always redacted (e.g., `.field("data", &"[REDACTED]")`) so it never leaks into logs. As far as I can tell, `Zeroizing<Box<[u8]>>` doesn’t natively provide [1] this, and beyond that, it doesn’t seem to add any additional functional value.
[1] https://doc.rust-lang.org/nightly/core/fmt/trait.Debug.html#tymethod.fmt

stefano-garzarella:
> @stefano-garzarella `Debug` is something we’ve implemented on top of `SecretBox `, ensuring the data is always redacted (e.g., `.field("data", &"[REDACTED]")`) so it never leaks into logs. As far as I can tell, `Zeroizing<Box<[u8]>>` doesn’t natively provide [1] this, and beyond that, it doesn’t seem to add any additional functional value. [1] https://doc.rust-lang.org/nightly/core/fmt/trait.Debug.html#tymethod.fmt

This could be a good point to make more clear in the commit message. BTW if this is important for us, okay (I don't have a strong opinion on this TBH) let's add `SecretBox` (or `SecretSlice`, I think `SecretSlice` is a more accurate name in this case).  But please avoid to implement, len, is_empty, as_slice, etc. just implement Deref and DerefMut, something like this:
```rust
#[derive(Zeroize, ZeroizeOnDrop)]
pub struct SecretSlice(alloc::boxed::Box<[u8]>);

impl SecretSlice {
    pub fn new(secret: Vec<u8>) -> Self {
        Self(secret.into_boxed_slice())
    }
}

impl core::ops::Deref for SecretSlice {
    type Target = [u8];
    fn deref(&self) -> &[u8] {
        &self.0
    }
}

impl core::ops::DerefMut for SecretSlice {
    fn deref_mut(&mut self) -> &mut [u8] {
        &mut self.0
    }
}

impl core::fmt::Debug for SecretSlice {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        f.debug_struct("SecretSlice")
            .field("len", &self.0.len())
            .field("data", &"[REDACTED]")
            .finish()
    }
}
```

I'm also unsure about `new(secret: Vec<u8>)`, maybe we can pass the slice directly. Anyway, it's a minor.

kernel/src/utils/util.rs
```diff
@@ -110,10 +113,59 @@ macro_rules! BIT_MASK {
     }};
 }
 
+/// Container for sensitive byte data.
+///
+/// The stored bytes are zeroed when the `SecretBox` is dropped or explicitly
+/// [`zeroize`](Zeroize::zeroize)d.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct SecretBox {
```
tanish111: secretslice

kernel/src/utils/util.rs
```diff
@@ -110,10 +113,59 @@ macro_rules! BIT_MASK {
...
+
+impl SecretBox {
+    /// Takes ownership of `secret`; does not zeroize it.
+    pub fn new(secret: Vec<u8>) -> Self {
+        Self {
+            data: secret.into_boxed_slice(),
+        }
+    }
+
+    /// Returns the secret length in bytes.
```
tanish111: remove

tanish111:
@joergroedel @luigix25 We don't need a new serialization implementation for SecretSlice, as zeroize already provides this by default.

```
#[cfg(feature = "alloc")]
impl<Z> Zeroize for Box<[Z]>
where
    Z: Zeroize,
{
    /// Unlike `Vec`, `Box<[Z]>` cannot reallocate, so we can be sure that we are not leaving
    /// values on the heap.
    fn zeroize(&mut self) {
        self.iter_mut().zeroize();
    }
}
```
Docs:- https://docs.rs/zeroize/latest/src/zeroize/lib.rs.html#546

tanish111:
@stefano-garzarella I have pushed all the changes can you take a look at PR again.

tanish111:
We also need to find a way to test the PR. Any ideas?

kernel/src/utils/util.rs
```diff
@@ -110,10 +113,49 @@ macro_rules! BIT_MASK {
...
+/// Container for sensitive byte data.
+///
+/// The owned buffer is zeroed when the `SecretSlice` is dropped or explicitly
+/// [`zeroize`](Zeroize::zeroize)d.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct SecretSlice(alloc::boxed::Box<[u8]>);
+
+impl SecretSlice {
+    /// Takes ownership of `secret`; does not zeroize it.
+    pub fn new(secret: Vec<u8>) -> Self {
```
stefano-garzarella: How about passing the slice directly?

That way, we can easily reuse it with both Vec and Slice.

kernel/src/attest.rs
```diff
@@ -185,13 +186,15 @@ impl AttestationDriver<'_> {
...
             return Err(AttestationError::SecretMissing);
         };
 
-        self.decrypt(&mut secret, decryption)?;
+        let mut secret_box = SecretSlice::new(secret);
```
stefano-garzarella: Perhaps we should add a comment here to clarify what we’ve discussed in https://github.com/coconut-svsm/svsm/pull/1097/changes#r3331109971

In other words, I’d say we’re not particularly concerned with securing `secret`, since it contains the encrypted secret (maybe we can rename it in `secret_enc`), but we are concerned with protecting the decrypted data that will be contained in `secret_box`, given that we’re decrypting in-place.

Cargo.toml
```diff
@@ -94,10 +96,9 @@ serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
-virtfw-libefi = { version = "0.6" }
-virtfw-varstore = { version = "0.6.1" }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
+zeroize = { version = "1.8.2", default-features = false, features = ["alloc", "derive"] }
```
stefano-garzarella: please, avoid enabling `alloc` here, since this can also be used by other crates in the workspace using std. Enable them in `kernel/Cargo.toml` like we do for `zerocopy`.

kernel/src/utils/util.rs
stefano-garzarella, line 1: I'm not sure if this is the right module to put `SecretSlice`, what about `kernel/src/crypto/mod.rs` ?

kernel/src/attest.rs
stefano-garzarella:
```suggestion
        let mut secret_slice = SecretSlice::new(secret);
```

kernel/src/utils/util.rs
stefano-garzarella: In addition, this makes it the caller's responsibility to determine whether the original Vec should be considered secret or not (see the comment below regarding adding the comment)

stefano-garzarella (changes requested):
```
SVSM/attest: Store the attestation secret in a zeroizing SecretBox

Introduce SecretBox, a Box<[u8]> wrapper with Zeroize/ZeroizeOnDrop that
clears sensitive bytes on drop. Use it for the attestation secret returned
from the proxy instead of a plain Vec<u8>, so decrypted key material is
wiped once it goes out of scope. new() takes ownership of the input Vec
without zeroizing it; Debug redacts contents. Add zeroize as a workspace
dependency.
```
Please use our convention: `kernel/attest: ...`

Also this should be update since we changed SecredBox to SecretSlice

stefano-garzarella:
BTW this is just a step to fix #807 IMO, so we should just mention it in the cover and remove the fixes tag to avoid it will be closed by GH. In details, the issue says:

> Zero out all secrets and keys after use in kernel/src/attest.rs (and everywhere else).

Here we are just handling the secret, but not keys for examples used in decrypt().

I'm not asking to do it here (if you want will be great), I'm just making clear that #807 can't be closed after this version of the PR.

kernel/src/utils/util.rs
tanish111: yes we can move it to crypto/mod.rs.

Cargo.toml
tanish111: noted.

kernel/src/utils/util.rs
tanish111: Yes makes sense would update with Slice directly

tanish111:
@stefano-garzarella I’d prefer to test this and merge it first, then create a separate PR to make the keys safe. WDYT?

stefano-garzarella:
> @stefano-garzarella I’d prefer to test this and merge it first, then create a separate PR to make the keys safe. WDYT?

yep, fine by me, but remove `fixes` tag from the PR description; otherwise, GitHub will close that issue. Just mention it saying that this is the first step in that direction and follow-up PRs will come next.

00xc:
> We also need to find a way to test the PR. Any ideas?

This is what I used to verify that the approach is correct:

https://gist.github.com/00xc/c70b860a4f39760c7c03ebdd415c2911

Not sure if we can do the same though, as declaring a `#[global_allocator]` for just a single test will likely not work.

stefano-garzarella:
> > We also need to find a way to test the PR. Any ideas?
> 
> This is what I used to verify that the approach is correct:
> 
> https://gist.github.com/00xc/c70b860a4f39760c7c03ebdd415c2911
> 
> Not sure if we can do the same though, as declaring a `#[global_allocator]` for just a single test will likely not work.

Oh true, what about something like this (which is clearly UB, but maybe for a test on x86_64 should be fine):

```rust
    #[test]
    fn secretslice_drop_zeroizes() {
        let secret = SecretSlice::new(Vec::from([0x42u8; 512]));
        let ptr = secret.as_ptr();
        let len = secret.len();

        unsafe {
            assert!(
                core::slice::from_raw_parts(ptr, len)
                    .iter()
                    .all(|&b| b == 0x42)
            );
        }

        drop(secret);

        unsafe {
            assert!(
                core::slice::from_raw_parts(ptr, len)
                    .iter()
                    .all(|&b| b != 0x42),
                "found original fill byte after drop"
            );
        }
    }
```

My first attempt was to check all zeroes, but the allocator writes something (I guess some metadata afeter drop()) so just checking that original content is not preserved, should be fine.

I tested with this applied, and the test is failing as expected:
```diff
@@ -117,7 +117,6 @@ macro_rules! BIT_MASK {
 ///
 /// The owned buffer is zeroed when the `SecretSlice` is dropped or explicitly
 /// [`zeroize`](Zeroize::zeroize)d.
-#[derive(Zeroize, ZeroizeOnDrop)]
 pub struct SecretSlice(alloc::boxed::Box<[u8]>);
 
 impl SecretSlice {
```

tanish111:
@stefano-garzarella @00xc there was one buggy commit introduced I have fixed can you retest?

tanish111:
@00xc working on your suggestion

stefano-garzarella:
> @stefano-garzarella @00xc there was one buggy commit introduced I have fixed can you retest?

Okay, but please squash fixups next time.

tanish111:
@stefano-garzarella done : )

kernel/src/attest.rs
```diff
@@ -185,13 +186,19 @@ impl AttestationDriver<'_> {
...
 
-        self.decrypt(&mut secret, decryption)?;
+        // `secret_enc` holds ciphertext from the attestation server. We decrypt
+        // in-place over a plain buffer; it is not treated as secret material.
+        // After decryption, wrap the plaintext in `SecretSlice` so the
+        // decrypted data is zeroed on drop.
+        self.decrypt(&mut secret_enc, decryption)?;
 
-        Ok(secret)
+        let secret_box = SecretSlice::new(&secret_enc);
```
stefano-garzarella: Wait, why this change?
This is exactly what we should NOT do.

The decrypted data will be left on `secret_enc` which is a Vec<u8> that will not be zeroed on drop, so I'm really really confused about this change.

I think we agreed on creating the SecretSlice before the decryption, no?

kernel/src/attest.rs
stefano-garzarella: I still see `secret_box` used.

kernel/src/attest.rs
tanish111: @stefano-garzarella please ignore that push while squash I missed a commit please check now.

kernel/src/attest.rs
```diff
@@ -185,13 +186,19 @@ impl AttestationDriver<'_> {
...
+        let Some(secret_enc) = response.secret else {
             return Err(AttestationError::SecretMissing);
         };
 
-        self.decrypt(&mut secret, decryption)?;
+        // `secret_enc` holds ciphertext from the attestation server and is not
+        // treated as secret material. Copy it into `SecretSlice` before
+        // decryption so the decrypted plaintext only lives in `secret_box`,
+        // which is zeroed on drop.
+        let mut secret_box = SecretSlice::new(&secret_enc);
```
stefano-garzarella: Let's avoid the copy as we did before, using `secret_enc.into_boxed_slice()` here.

So you should change `new()` or maybe even better implementing `From<Box<[u8]>> for SecretSlice` and remove `new()` completely for now. We can add it later if we really need it.

kernel/src/crypto/mod.rs
stefano-garzarella, line 1:
```
    kernel/attest: add SecretSlice for zeroed attestation secrets
    
    Introduce SecretSlice, a Box<[u8]> wrapper that zeroes its buffer on
    drop and redacts Debug output. Decrypt attestation ciphertext in-place
    into a plain Vec, then copy the plaintext into SecretSlice so decrypted
    key material is wiped when it goes out of scope.
    
    Signed-off-by: tanish111 <tanishdesai37@gmail.com>
```

Again, I'm confused `Decrypt attestation ciphertext in-place into a plain Vec, then copy the plaintext`, this is really what we should not do and what was done in the previous version.

tanish111: yes would update this as well

kernel/src/crypto/mod.rs
```diff
@@ -6,6 +6,46 @@
...
+extern crate alloc;
+
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+/// Container for sensitive byte data.
+///
+/// The owned buffer is zeroed when the `SecretSlice` is dropped or explicitly
+/// [`zeroize`](Zeroize::zeroize)d.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct SecretSlice(alloc::boxed::Box<[u8]>);
```
stefano-garzarella: Why not `use alloc::boxed::Box` ?

00xc:
> > > We also need to find a way to test the PR. Any ideas?
> > 
> > 
> > This is what I used to verify that the approach is correct:
> > https://gist.github.com/00xc/c70b860a4f39760c7c03ebdd415c2911
> > Not sure if we can do the same though, as declaring a `#[global_allocator]` for just a single test will likely not work.
> 
> Oh true, what about something like this (which is clearly UB, but maybe for a test on x86_64 should be fine):
> 
> ```rust
>     #[test]
>     fn secretslice_drop_zeroizes() {
>         let secret = SecretSlice::new(Vec::from([0x42u8; 512]));
>         let ptr = secret.as_ptr();
>         let len = secret.len();
> 
>         unsafe {
>             assert!(
>                 core::slice::from_raw_parts(ptr, len)
>                     .iter()
>                     .all(|&b| b == 0x42)
>             );
>         }
> 
>         drop(secret);
> 
>         unsafe {
>             assert!(
>                 core::slice::from_raw_parts(ptr, len)
>                     .iter()
>                     .all(|&b| b != 0x42),
>                 "found original fill byte after drop"
>             );
>         }
>     }
> ```

This will likely break running tests with Miri (which is already kind of broken but I'd rather not break it further).

kernel/src/crypto/mod.rs
tanish111: updated @stefano-garzarella I initially thought this was the only use case, so I didn't see much point in importing it. But with this approach, the code looks much cleaner.

00xc:
I think the only approach to test this would be to add test hooks into the allocator, which seems too intrusive (and which would need to include the in-SVSM and not-in-SVSM divergence). I think we can trust the zeroize crate to do its job without having an explicit test.

stefano-garzarella:
> I think the only approach to test this would be to add test hooks into the allocator, which seems too intrusive (and which would need to include the in-SVSM and not-in-SVSM divergence). I think we can trust the zeroize crate to do its job without having an explicit test.

Agree on that, let's stay with what we have now.

Cargo.toml
```diff
@@ -94,10 +96,9 @@ serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
-virtfw-libefi = { version = "0.6" }
-virtfw-varstore = { version = "0.6.1" }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
+zeroize = { version = "1.8.2", default-features = false, features = ["derive"] }
```
stefano-garzarella:
```suggestion
zeroize = { version = "1.8.2", default-features = false }
```

stefano-garzarella: Or remove the `derive` in `kernel/Cargo.toml`, but since for now kernel is the only user of this, I'll remove from here.

stefano-garzarella:
@tanish111 a part of that nit comment, the rest LGTM, thanks!

tanish111:
@stefano-garzarella I have applied nit fix can you assist me with steps to merge this?

stefano-garzarella (approved):

stefano-garzarella:
@tanish111 thanks, merged!

PR #1098, opened 2026-06-01
scripts/check-signed-off: Avoid exiting on the first failure

n-ramacciotti:
The `check-signed-off` script skips all the remaining checks after the
first error is found.

Count the number of invalid checks in all the commits and then exit.

This is the same approach used in `check-spdx.sh` where all the new
files are checked.

Additionally, fix a minor issue where the incorrect failure message is 
shown.

### DIFF
--- a/scripts/check-signed-off.sh
+++ b/scripts/check-signed-off.sh
@@ -39,6 +39,8 @@ fi
 start=$1
 end=$2
 
+invalid_checks=0
+
 commits=$(git log --no-merges "${start}".."${end}" --format="%H")
 for c in ${commits[@]}; do
 
@@ -47,7 +49,7 @@ for c in ${commits[@]}; do
 	nonempty_body "$c"
 	if [ "$?" != "0" ]; then
 		echo "Message body is empty for commit $c"
-		exit 1
+		invalid_checks=$((invalid_checks + 1))
 	fi
 
 	commit_email=$(git show --no-patch --format="%ae" "$c" || exit 1)
@@ -60,7 +62,7 @@ for c in ${commits[@]}; do
 		echo "Author name mismatch on commit $c"
 		echo "    Commit author name: $commit_name"
 		echo "    Signed-off-by name(s):" $sign_names
-		exit 1
+		invalid_checks=$((invalid_checks + 1))
 	fi
 
 	matches_any "$commit_email" "$sign_emails"
@@ -68,9 +70,15 @@ for c in ${commits[@]}; do
 		echo "Author email mismatch on commit $c"
 		echo "    Commit author email: $commit_email"
 		echo "    Signed-off-by email(s):" $sign_emails
-		exit 1
+		invalid_checks=$((invalid_checks + 1))
 	fi
 
 done
 
+if (($invalid_checks > 0)); then
+	echo "Exiting with $invalid_checks failed checks"
+	exit 1
+fi
+
+echo "All commits have a Signed-off-by and a non-empty body"
 exit 0
--- a/scripts/check-signed-off.sh
+++ b/scripts/check-signed-off.sh
@@ -24,13 +24,10 @@ matches_any() {
 # Check that the body for the given commit is not empty
 nonempty_body() {
 	body=$(git show --no-patch --format="%b" "$1" | sed '/^ *$/d')
-	trailers=$(git show --no-patch --format="%(trailers:only)" "$1")
+	trailers=$(git show --no-patch --format="%(trailers:only)" "$1" | sed '/^ *$/d')
 
-	body_len=$(echo "$body" | wc -l)
-	trailer_len=$(echo "$trailers" | wc -l)
-
-	# If the body is the same length as the trailers it means the body is empty
-	[ "$body_len" = "$trailer_len" ] && return 1
+	# If the body is the same as the trailers it means that the body is empty
+	[ "$body" = "$trailers" ] && return 1
 	return 0
 }
 


scripts/check-signed-off.sh
```diff
@@ -42,15 +42,16 @@ fi
 start=$1
 end=$2
 
+invalid_checks=0
+
 commits=$(git log --no-merges "${start}".."${end}" --format="%H")
 for c in ${commits[@]}; do
-
```
luigix25, line 47: please avoid unrelated changes

scripts/check-signed-off.sh
```diff
@@ -42,15 +42,16 @@ fi
...
 	echo "Checking $c"
 
 	nonempty_body "$c"
 	if [ "$?" != "0" ]; then
-		echo "Message body is empty for commit $c"
-		exit 1
+		echo "Message body is empty"
```
luigix25: it could be a good idea to keep the commit hash in the log, wdyt?

scripts/check-signed-off.sh
```diff
@@ -60,20 +61,27 @@ for c in ${commits[@]}; do
...
 		echo "    Commit author name: $commit_name"
 		echo "    Signed-off-by name(s):" $sign_names
-		exit 1
+		invalid_checks=$((invalid_checks + 1))
 	fi
 
 	matches_any "$commit_email" "$sign_emails"
 	if [ "$?" != "0" ]; then
-		echo "Author email mismatch on commit $c"
+		echo "Author email mismatch"
```
luigix25: ditto

scripts/check-signed-off.sh
```diff
@@ -23,11 +23,11 @@ matches_any() {
 
 # Check that the body for the given commit is not empty
 nonempty_body() {
-	body=$(git show --no-patch --format="%b" "$1" | sed '/^ *$/d')
+	body=$(git show --no-patch --format="%b" "$1")
 	trailers=$(git show --no-patch --format="%(trailers:only)" "$1")
```
luigix25: what about

```shell
body=$(git show --no-patch --format="%b" "$1" | sed '/^ *$/d')
trailers=$(git show --no-patch --format="%(trailers:only)" "$1" | sed '/^ *$/d')
```

In this way we just modify one line

scripts/check-signed-off.sh
```diff
@@ -60,20 +61,27 @@ for c in ${commits[@]}; do
...
-		exit 1
+		invalid_checks=$((invalid_checks + 1))
 	fi
 
+	echo "================================================="
 done
 
+if (($invalid_checks > 0)); then
+	echo "Exiting with $invalid_checks failed checks"
+	exit 1;
```
luigix25: nit: `;` is not necessary

luigix25 (changes requested):
overall LGTM!

scripts/check-signed-off.sh
n-ramacciotti: Oh, sorry. I placed the `echo "===" ` there before, but I moved it at the end and I forgot to add the newline.

scripts/check-signed-off.sh
n-ramacciotti: The commit hash is already printed before. 
```
echo "Checking $c"
```
My idea is to have this structure:
```
Checking commit_hash_1
Message body is empty
=================================================
Checking commit_hash_2
=================================================
Exiting with 1 failed checks
```

With this, if multiple checks fail (for the same commit), the hash is shown only once.

Should I mention this in the commit message?

scripts/check-signed-off.sh
n-ramacciotti: The problem is that if `trailers` (or `body`) is empty, `echo` will still print a newline, which will result in `trailer_len` (or `body_len`) to be equal to 1

scripts/check-signed-off.sh
luigix25: ok, feel free to ignore

scripts/check-signed-off.sh
luigix25: oh.. right!

scripts/check-signed-off.sh
00xc: The cosmetic change to the output makes the diff harder to read in my opinion. It's not a big deal, but I'd rather this just update the fact that we do not exit early.

n-ramacciotti:
v2:
- removed output formatting changes

00xc (approved):

luigix25 (approved):

scripts/check-signed-off.sh
stefano-garzarella: > The problem is that if `trailers` (or `body`) is empty, `echo` will still print a newline, which will result in `trailer_len` (or `body_len`) to be equal to 1

This should be mentioned in the commit description, I was going to suggest exactly the same.

That said, what about removing line counting and just comparing strings in this way (untested):

```
body_stripped=$(echo "$body" | sed '/^ *$/d')
trailer_stripped=$(echo "$trailers" | sed '/^ *$/d')
[ "$body_stripped" = "$trailer_stripped" ] && return 1
```

n-ramacciotti:
v3:
- Comparing strings instead of counting lines @stefano-garzarella

stefano-garzarella (approved):

scripts/check-signed-off.sh
```diff
@@ -24,13 +24,10 @@ matches_any() {
 # Check that the body for the given commit is not empty
 nonempty_body() {
 	body=$(git show --no-patch --format="%b" "$1" | sed '/^ *$/d')
-	trailers=$(git show --no-patch --format="%(trailers:only)" "$1")
+	trailers=$(git show --no-patch --format="%(trailers:only)" "$1" | sed '/^ *$/d')
```
n-ramacciotti, line 27: [@stefano-garzarella]

If there is an empty line between two trailer lines, the trailer parsing will stop and only the second line will be considered a trailer. 

The first line would be considered a body line, resulting in a non-empty body.

This happens with or without [da4928a](https://github.com/coconut-svsm/svsm/pull/1098/changes/da4928a1876c4e1701a4639a33ffe271b1b2d241) as it is related to git parsing.

Is this a problem? I usually see them as a single block in commit messages, but I'm not sure.

With this issue, the sed manipulation here can be removed.

joergroedel (approved):

PR #1099, opened 2026-06-02
igvmbuilder: use correct VTL for VBS VP context

msft-jlange:
When building without guest firmware, the IGVM file designates VTL 0 as the highest enabled VTL.  The VBS VP context must specify a matching VTL.


luigix25 (approved):
LGTM! Thanks

stefano-garzarella (approved):

PR #1101, opened 2026-06-03
Preliminary macOS support for nocc mode

luigix25:
This PR allows to boot SVSM on macOS (and probably *BSD)

The changes are minimal: 
- Switch to a different memory-backend in QEMU for nocc mode.
- Switch to BSD-compatible commands (eg: `grep -P` is GNU-only)

Note: `virtio-vsock` and `virtio-blk` will NOT work as is, as they require vhost-device, that is linux-specific. We could fall back to vhost-user, but is not really needed for now.

Note 2: cross compiling on ARM is still not possible yet. There is some ongoing work but is not ready yet  (#971 #981)

Note 3: IGVM also relies on some GNU-only commands, I opened a PR to fix it (https://github.com/microsoft/igvm/pull/123)

### DIFF
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -29,6 +29,7 @@ CPU=EPYC-v4
 ACCEL=kvm
 IGVM_OBJ=""
 SNAPSHOT="on"
+MEMORY_SIZE="8G"
 
 STATE_DEVICE=""
 VSOCK_DEVICE=""
@@ -138,6 +139,7 @@ case "$CGS" in
   nocc)
     SNP_GUEST=""
     CPU=max,smep=on
+    MEMORY=memory-backend-ram,size=$MEMORY_SIZE,id=mem0,prealloc=false
     if (( QEMU_MAJOR < 11 )); then
       ACCEL=tcg
       SNP_GUEST="-object nocc,id=cgs0"
@@ -145,6 +147,7 @@ case "$CGS" in
     ;;
   sev)
     SNP_GUEST="-object sev-snp-guest,id=cgs0,cbitpos=$C_BIT_POS,reduced-phys-bits=$REDUCED_PHYS_BITS"
+    MEMORY=memory-backend-memfd,size=$MEMORY_SIZE,id=mem0,share=true,prealloc=false,reserve=false
     ;;
   *)
     echo "Error: Unexpected CGS value '$CGS'"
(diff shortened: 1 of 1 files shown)


scripts/launch_guest.sh
```diff
@@ -134,10 +134,13 @@ if (( QEMU_MAJOR < 10 || (QEMU_MAJOR == 10 && QEMU_MINOR < 1) )); then
   exit 1
 fi
 
+MEMORY=memory-backend-memfd,size=8G,id=mem0,share=true,prealloc=false,reserve=false
+
 case "$CGS" in
   nocc)
     SNP_GUEST=""
     CPU=max,smep=on
+    MEMORY=memory-backend-ram,size=8G,id=mem0,share=true,prealloc=false
```
stefano-garzarella: why `share=true` ?

scripts/launch_guest.sh
```diff
@@ -134,10 +134,13 @@ if (( QEMU_MAJOR < 10 || (QEMU_MAJOR == 10 && QEMU_MINOR < 1) )); then
   exit 1
 fi
 
+MEMORY=memory-backend-memfd,size=8G,id=mem0,share=true,prealloc=false,reserve=false
```
stefano-garzarella: Now we have the RAM size spread in 2 places, IMO better to add a variable to control it and use in both places.

stefano-garzarella: Also, I think you can move this under SEV, so it's more clear that we need this for SEV.

scripts/launch_guest.sh
luigix25: oops, leftover from copy-paste

luigix25:
[@stefano-garzarella]

Addressed review comments:
- Ram size is now a standalone variable
- memory-backend-memfd is now defined in the sev branch

scripts/launch_guest.sh
```diff
@@ -138,13 +139,15 @@ case "$CGS" in
   nocc)
     SNP_GUEST=""
     CPU=max,smep=on
+    MEMORY=memory-backend-ram,size=$MEMORY_SIZE,id=mem0,prealloc=false
```
stefano-garzarella, line 142: Why dropping `reserve=false` ?

luigix25: `reserve` is linux-specific, and won't work on macos. I don't think we need it for nocc mode, do we?

stefano-garzarella: > `reserve` is linux-specific, and won't work on macos. I don't think we need it for nocc mode, do we?

I'm fine to remove it, but it's a change from what we are doing now, so let's mention it in the commit message.

luigix25:
improved commit message @stefano-garzarella

stefano-garzarella (approved):

PR #1102, opened 2026-06-05
virtio/mmio: handle non-page-aligned MMIO device addresses

luigix25:
MMIO device addresses may not be page-aligned. This is the case for the microvm or ARM machine types in QEMU, that align the virtio-mmio device addresses to 512 bytes.

Previously the address was used directly without page-aligning, causing a panic on non-page-aligned addresses.

Split the address into a page-aligned base for mapping and an offset applied to the resulting virtual address, so the transport header points to the correct location within the mapped page.

If multiple devices reside in the same page, each gets its own mapping, which is slightly wasteful but avoids shared-ownership complexity.

This is part of an effort to allow SVSM to fully work on microvm.

### DIFF
--- a/kernel/src/virtio/mmio.rs
+++ b/kernel/src/virtio/mmio.rs
@@ -8,9 +8,10 @@ extern crate alloc;
 use alloc::vec::Vec;
 use core::ptr::NonNull;
 
+use virtio_drivers::transport::mmio::VirtIOHeader;
 use virtio_drivers::transport::{DeviceType, Transport, mmio::MmioTransport};
 
-use crate::address::PhysAddr;
+use crate::address::{Address, PhysAddr};
 use crate::boot_params::BootParams;
 use crate::fw_cfg::FwCfg;
 use crate::mm::{GlobalRangeGuard, map_global_range_4k_shared, pagetable::PTEntryFlags};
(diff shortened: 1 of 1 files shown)


kernel/src/virtio/mmio.rs
```diff
@@ -63,17 +64,33 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
...
+        // mapping, which is slightly wasteful but avoids shared-ownership
+        // complexity.
         let Ok(mem) = map_global_range_4k_shared(phys_addr, PAGE_SIZE, PTEntryFlags::data()) else {
             log::warn!("MmioSlots: Failed to map MMIO region at {addr:x}");
             continue;
         };
 
         // Not expected to fail, because mem exists.
-        let header = NonNull::new(mem.addr().as_mut_ptr()).unwrap();
+        let header = NonNull::new((mem.addr() + page_offset).as_mut_ptr::<VirtIOHeader>()).unwrap();
```
stefano-garzarella: Why `::<VirtIOHeader>` is now needed?

kernel/src/virtio/mmio.rs
```diff
@@ -63,17 +64,33 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
...
+
+        assert!(
+            raw_addr.is_aligned(core::mem::align_of::<VirtIOHeader>()),
+            "MMIO device address must be aligned to VirtIOHeader requirements"
+        );
+
+        assert!(
+            page_offset + core::mem::size_of::<VirtIOHeader>() <= PAGE_SIZE,
+            "MMIO device header at {raw_addr:#x} crosses a page boundary"
+        );
```
stefano-garzarella: It is unclear to me why we are adding asserts now while for other aligning failures (e.g. from `map_global_range_4k_shared()`) we were just ignoring the slot and printing a log message.

osteffenrh:
Nice,
LGTM!

kernel/src/virtio/mmio.rs
```diff
@@ -63,17 +64,33 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
     let mut slots = Vec::with_capacity(dev.len());
 
     for addr in dev {
-        let phys_addr = PhysAddr::from(addr);
-
+        let raw_addr = PhysAddr::from(addr);
+        let phys_addr = raw_addr.page_align();
+        let page_offset = raw_addr.page_offset();
```
stefano-garzarella: Those variable names are confusing IMO. What about:
```
let phys_addr = PhysAddr::from(addr);
let page_base = phys_addr.page_align();
let page_offset = phys_addr.page_offset();
```

kernel/src/virtio/mmio.rs
luigix25: good point, i'll rollback to simple warnings. `map_global_range_4k_shared` guaranteed us proper alignment, this is not true anymore. So I need to check this manually

kernel/src/virtio/mmio.rs
luigix25: oops, development leftover

luigix25:
[@stefano-garzarella]

- Renamed variables
- Removed cast
- Assert are now warnings

kernel/src/virtio/mmio.rs
```diff
@@ -64,16 +65,32 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
 
     for addr in dev {
         let phys_addr = PhysAddr::from(addr);
+        let page_base = phys_addr.page_align();
+        let page_offset = phys_addr.page_offset();
 
-        let Ok(mem) = map_global_range_4k_shared(phys_addr, PAGE_SIZE, PTEntryFlags::data()) else {
+        if !phys_addr.is_aligned(core::mem::align_of::<VirtIOHeader>()) {
```
stefano-garzarella:
```suggestion
        if !phys_addr.is_aligned_to::<VirtIOHeader>() {
```

kernel/src/virtio/mmio.rs
```diff
@@ -64,16 +65,32 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
...
+            log::warn!("MmioSlots: MMIO device at {phys_addr:x} is not properly aligned");
+            continue;
+        }
+
+        if page_offset + core::mem::size_of::<VirtIOHeader>() > PAGE_SIZE {
```
stefano-garzarella:
```suggestion
        if phys_addr.crosses_page(core::mem::size_of::<VirtIOHeader>()) {
```

luigix25:
@stefano-garzarella applied your suggestions, thanks

stefano-garzarella (approved):

PR #1103, opened 2026-06-08
task: streamline reference count management during task termination

msft-jlange:
Handle the reference count management of task `Arc` objects by passing task pointers back from the context switch routine instead of hanging references off of the current CPU pointer.  This makes reference count management a bit easier to comprehend.

Also, fix the reference management when a context switch moves to a task that is starting for the first time.  The previous code didn't have the logic for releasing terminated tasks, which meant they would be terminated at some point later, typically with interrupts disabled (which is problematic).

### DIFF
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -77,9 +77,6 @@ pub struct RunQueue {
     /// Idle task - runs when there is no other runnable task
     idle_task: Option<TaskPointer>,
 
-    /// Temporary storage for tasks which are about to be terminated
-    terminated_task: Option<TaskPointer>,
-
     /// Pointer to a task that should be woken when returning from idle
     wake_from_idle: Option<TaskPointer>,
 }
@@ -93,7 +90,6 @@ impl RunQueue {
             run_list: LinkedList::new(TaskRunListAdapter::new()),
             current_task: None,
             idle_task: None,
-            terminated_task: None,
             wake_from_idle: None,
         }
     }
@@ -416,10 +412,6 @@ fn current_task_terminated() {
     let cpu = this_cpu();
     let mut rq = cpu.runqueue_mut();
 
-    // Capture a reference to the current task so that it remains referenced
-    // until the next scheduling operation completes.
-    rq.terminated_task = rq.current_task.clone();
-
     let task_node = rq
         .current_task
         .as_mut()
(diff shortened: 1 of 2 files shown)


kernel/src/task/tasks.rs
```diff
@@ -994,27 +1001,31 @@ fn task_attach_console() {
...
     // disabled in the new task.
     // This only needs to be done for the first time a task runs. Any
     // subsequent task switches will go through schedule() and there the guard
     // is dropped, re-enabling IRQs.
-
     irqs_enable();
 
+    // Drop the interrupt guard before allowing the previous task reference to
+    // go out of scope.  This ensures that the task destructor will run with
+    // interrupts enabled.
```
stefano-garzarella: copy & paste issue? I can't see an IrqGuard here.

kernel/src/task/schedule.rs
```diff
@@ -617,30 +635,41 @@ fn select_new_task(reschedule: bool, irq_guard: Option<IrqGuard>) {
...
+            // beyond this call.  This call will return an `Arc` reference to
+            // the task that ran most recently, which may be the final
+            // reference to the task if the task was terminated.
+            let prev_task = switch_to(Some(current), next);
+
+            // The previously captured task pointer is known to be valid
+            // because the task is still executing.
+            sse_restore_context(u64::from((*current_ptr).xsa.vaddr()));
+
+            Some(prev_task)
```
stefano-garzarella:
```suggestion
            prev_task
```

kernel/src/task/schedule.rs
```diff
@@ -499,32 +491,56 @@ pub fn set_affinity(cpu_index: usize) {
...
+// pointer) or else it will be destroyed before it is entered.
 #[inline(always)]
-unsafe fn switch_to(prev: *const Task, next: *const Task) {
+unsafe fn switch_to(prev_task: Option<TaskPointer>, next_task: TaskPointer) -> Option<TaskPointer> {
+    // Capture a pointer to the new task and consume its `Arc`.  The caller
+    // guarantees that the task will still exist.
+    let next = Arc::as_ptr(&next_task);
+    drop(next_task);
+
+    // Consume the `Arc` describing the currently exeucting task without
```
stefano-garzarella:
```suggestion
    // Consume the `Arc` describing the currently executing task without
```

kernel/src/task/schedule.rs
```diff
@@ -499,32 +491,56 @@ pub fn set_affinity(cpu_index: usize) {
...
         );
+        complete_task_switch(new_prev)
+    }
+}
+
+/// # Safety
+/// The caller must guarantee that the task pointer argument is the one that
+/// was returned from the context switcher.
+pub unsafe fn complete_task_switch(prev: usize) -> Option<TaskPointer> {
+    // If there was a previous task, then reate a new `Arc` that describes
```
stefano-garzarella:
```suggestion
    // If there was a previous task, then create a new `Arc` that describes
```

stefano-garzarella (commented):
I'm not very familiar with this code, but overall it seems a nice improvement to me. I really only left a few minor comments.

kernel/src/task/tasks.rs
msft-jlange: Yes, copy/paste issue (from `select_new_task`).

kernel/src/task/schedule.rs
msft-jlange: This block has to evaluate to `Option<TaskPointer>` because the other side of the `if` is `None`.  The `None` path happens if the next task is the same as the current task, in which case no task switch is performed (and thus there is nothing to drop later).

stefano-garzarella: `prev_task` is already an `Option<TaskPointer>`, no?

I mean my point was that here we are creating an `Option<Option<TaskPointer>>` IIUC the code.

msft-jlange: > `prev_task` is already an `Option<TaskPointer>`, no?

Good point!

stefano-garzarella (approved):

joergroedel (approved):
Looks good to me overall. Codex found a corner case we should fix in a follow-on PR:

```
  - [P2] Avoid requiring discarded tasks to be terminated — /home/joro/src/svsm/kernel/src/task/tasks.rs:364-364
    This debug assertion also runs for tasks that were created but never started. For example, exec() calls create_user_task() and then performs several ELF
    segment and stack mappings before finish_user_task(); if any of those validations or mappings fail, the TaskPointer is dropped while its initial state is
    still RUNNING, causing a debug-build panic instead of returning the expected error. The termination-state check should only apply to tasks that have
    actually been scheduled/owned by the scheduler, or unstarted tasks need a different state before they can be dropped.
```

PR #1106, opened 2026-06-08
add support for cross-compiling libtcgtpm

kraxel:
This adds support for cross-compiling libtcgtpm using the gcc cross
compiler.  Requires a cross-compiler which can build freestanding code
like firmware or the linux kernel.  Many linux distros ship such a
compiler.  Use `dnf install gcc-x86_64-linux-gnu` to install it on
Fedora.

Together with https://github.com/coconut-svsm/svsm/pull/981 this
allows full-featured cross-builds of coconut svsm.


libtcgtpm/Makefile
```diff
@@ -1,3 +1,10 @@
+ifeq ($(shell uname -m),x86_64)
+CROSS_PREFIX :=
+else
+CROSS_PREFIX := x86_64-linux-gnu-
+endif
+CC := $(CROSS_PREFIX)gcc
```
luigix25, line 6: mmh, would this work on windows?

kraxel: I don't think so (unless using wsl).  I think there is no `uname` command on windows.

Can svsm compiled on windows today?  If so, should we maybe have CI check that?

luigix25: AFAIK svsm can be used on hyper-v. Not sure they are using internal tooling. Probably @msft-jlange can help us

+1 on adding the build target in the CI if possible

luigix25: Otherwise, we could just let the user define  `CROSS_PREFIX` when cross-compiling

stefano-garzarella: > Otherwise, we could just let the user define `CROSS_PREFIX` when cross-compiling

This is already supported, no?

stefano-garzarella: That said, we already had `CC = gcc` in `libtcgtpm/deps/libcrt/Makefile` so I think we are not introducing anything new.

stefano-garzarella (commented):
Overall LGTM, just a nit: can you follow our commit "subsystem: title" convention (e.g., `libtcgtpm: cross compiler setup`)

libtcgtpm/Makefile
luigix25: > > Otherwise, we could just let the user define `CROSS_PREFIX` when cross-compiling
> 
> This is already supported, no?

Right.

> That said, we already had `CC = gcc` in `libtcgtpm/deps/libcrt/Makefile` so I think we are not introducing anything new.

IIUC libcrt is optional, no?

Don't get me wrong, I LOVE the idea of cross compiling svsm

stefano-garzarella: > > > Otherwise, we could just let the user define `CROSS_PREFIX` when cross-compiling
> > 
> > 
> > This is already supported, no?
> 
> Right.
> 
> > That said, we already had `CC = gcc` in `libtcgtpm/deps/libcrt/Makefile` so I think we are not introducing anything new.
> 
> IIUC libcrt is optional, no?

ehm, I'm not sure if I get the question.

If you mean that it isn't always built, that's only true if the entire `libctgtpm` library isn't built (i.e., the `vtpm` feature isn't enabled for SVSM). If, on the other hand, `vtpm` feature is enabled, then `libcrt` must be built. It's essentially the libc required to run TPM and OpenSSL in SVSM.

We recently allowed to disable it (see commit 7c3b3c6560639678ab3533d769db38f283fec62d), but that feature is used just when building `libtcgtpm` for userspace targets like tests, doctests, etc. where the system libc will be used.

kraxel: Running on hyper-v doesn't imply building on windows.

The whole logic in `libtcgtpm/Makefile` doesn't look very windows-aware to me, but that is only used if the `vtpm` feature is turned on, so maybe cargo can build everything else just fine on windows.

stefano-garzarella (approved):
LGTM, thanks!

@kraxel since we also need #981 to support cross-compilation, what about opening another PR to add some instructions in `Documentation/` on how to cross-compile the entire project ?

kraxel:
> @kraxel since we also need #981 to support cross-compilation, what about opening another PR to add some instructions in `Documentation/` on how to cross-compile the entire project ?

There isn't that much do document actually.  You need the x86_64-unknown-none libs of course, but you need that on x86 too, so that isn't a cross-build specific requirement.  I have a PR open to tweak the fedora packages so this is actually installable on non-x86 arch (https://src.fedoraproject.org/rpms/rust/pull-request/68).

Then you need a cross-gcc for `libtcgtpm`, which is packaged as `gcc-x86_64-linux-gnu` in fedora.  That's it.  Simply typing `make` will build everything.

kraxel:
> Then you need a cross-gcc for `libtcgtpm`, which is packaged as `gcc-x86_64-linux-gnu` in fedora.

```
--- a/Documentation/docs/installation/INSTALL.md
+++ b/Documentation/docs/installation/INSTALL.md
@@ -204,6 +204,8 @@ Building the SVSM itself requires:
   [https://rustup.rs/](https://rustup.rs/) on how to get this environment installed.
 * `x86_64-unknown-none` target toolchain installed (`rustup target add x86_64-unknown-none`)
 * `binutils` >= 2.39
+* (optional) for the vtpm feature gcc is needed.  For cross-compiling
+  a cross-gcc is needed (/usr/bin/x86_64-linux-gnu-gcc, gcc-x86_64-linux-gnu.rpm).
 
 You may also need to install the TPM 2.0 Reference Implementation build
 dependencies. On OpenSUSE you can do this by:
```
?

stefano-garzarella:
> > Then you need a cross-gcc for `libtcgtpm`, which is packaged as `gcc-x86_64-linux-gnu` in fedora.
> 
> ```
> --- a/Documentation/docs/installation/INSTALL.md
> +++ b/Documentation/docs/installation/INSTALL.md
> @@ -204,6 +204,8 @@ Building the SVSM itself requires:
>    [https://rustup.rs/](https://rustup.rs/) on how to get this environment installed.
>  * `x86_64-unknown-none` target toolchain installed (`rustup target add x86_64-unknown-none`)
>  * `binutils` >= 2.39
> +* (optional) for the vtpm feature gcc is needed.  For cross-compiling
> +  a cross-gcc is needed (/usr/bin/x86_64-linux-gnu-gcc, gcc-x86_64-linux-gnu.rpm).
>  
>  You may also need to install the TPM 2.0 Reference Implementation build
>  dependencies. On OpenSUSE you can do this by:
> ```
> 
> ?

Yep, this should be enough. Thanks!

PR #1108, opened 2026-06-09
virtio-drivers: Use upstream via custom safe-mmio backend

osteffenrh:
Consume the current virtio-drivers master branch directly by implementing a custom MMIO backend for the safe-mmio crate (used by upstream virtio-drivers)
to route accesses through SVSM_PLATFORM.

~~The custom backend feature is not upstream yet (PR soon!), see the changes here:
https://github.com/osteffenrh/safe-mmio/commits/custom-mmio-backend/~~

The custom-mmio backend was added to safe-mmio in this pr: https://github.com/google/safe-mmio/pull/55
It is not published on crates.io yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,6 +90,7 @@ libfuzzer-sys = "0.4"
 log = "0.4.17"
 p384 = { version = "0.13.0" }
 rustc-demangle = { version = "0.1.26" }
+safe-mmio = { version = "0.3", features = ["custom-mmio"] }
 serde = { version = "1.0.215", default-features = false }
 serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
@@ -110,6 +111,12 @@ vstd = { version = "=0.0.0-2026-04-12-0118", features = ["alloc", "allow_panic"]
 # Verification libs
 verify_proof = { path = "verification/verify_proof", default-features = false  }
 verify_external = { path = "verification/verify_external", default-features = false  }
+
+[patch.crates-io]
+# Force all dependencies to use the same version of safe-mmio, which has
+# the new custom-mmio feature, so that injecting our MMIO backend works.
+# This can be removed once the feature has been released on creates.io.
+safe-mmio = { git = "https://github.com/google/safe-mmio", rev = "89f584caaf32fcd4341a5a716c97d67be83edb9c" }
 
 [profile.release]
 panic = 'abort'
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,6 @@ packit = { path = "packit" }
 userlib = { path = "user/lib" }
 userinit = { path = "user/init" }
 release = { path = "release" }
-virtio-drivers = { git = "https://github.com/coconut-svsm/virtio-drivers.git", rev = "dcbdbc8cefc0c985322382acdeb633a63a3a4cc1" }
 
 # crates.io
 aes-gcm = { version = "0.10.3", default-features = false }
@@ -97,10 +96,12 @@ sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
 virtfw-libefi = { version = "0.6" }
 virtfw-varstore = { version = "0.6.1" }
+# We can pull virtio-drivers from crates.io once the vsock connection manager
+# changes have been released.
+virtio-drivers = { git = "https://github.com/rcore-os/virtio-drivers.git", rev = "c8786e83e6a64e46ed5cab301048dbaf3aab1444" }
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
 zeroize = { version = "1.8.2", default-features = false }
-embedded-io = { version = "0.6.1" }
 
 # Verus repos
 verus_builtin = { version = "=0.0.0-2026-04-12-0118", default-features = false }
--- a/kernel/src/block/virtio_blk.rs
+++ b/kernel/src/block/virtio_blk.rs
@@ -22,7 +22,7 @@ extern crate alloc;
 use alloc::boxed::Box;
 
 pub struct VirtIOBlkDriver {
-    device: SpinLock<VirtIOBlk<SvsmHal, MmioTransport<SvsmHal>>>,
+    device: SpinLock<VirtIOBlk<SvsmHal, MmioTransport<'static>>>,
     _mmio_space: GlobalRangeGuard,
 }
 
--- a/kernel/src/virtio/mmio.rs
+++ b/kernel/src/virtio/mmio.rs
@@ -17,11 +17,11 @@ use crate::fw_cfg::FwCfg;
 use crate::mm::{GlobalRangeGuard, map_global_range_4k_shared, pagetable::PTEntryFlags};
 use crate::platform::SVSM_PLATFORM;
 use crate::types::PAGE_SIZE;
-use crate::virtio::hal::{SvsmHal, virtio_init};
+use crate::virtio::hal::virtio_init;
 
 #[derive(Debug)]
 pub struct MmioSlot {
-    pub transport: MmioTransport<SvsmHal>,
+    pub transport: MmioTransport<'static>,
     // Fields are ordered so that `mmio_range` is dropped last.
     // The MmioTransport destructor resets the device via MMIO writes, and
     // dropping `mmio_range` destroys the mapping, so `mmio_range` must be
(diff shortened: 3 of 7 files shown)


kraxel:
Looks good to me.  Is there any feedback from safe_mmio maintainers on the custom ops idea?

osteffenrh:
> Looks good to me. Is there any feedback from safe_mmio maintainers on the custom ops idea?

-> github.com/google/safe-mmio/pull/55

osteffenrh:
v2:
- safe-mmio merged the PR and has the custom-mmio backend feature in the github repo.
  Release to creates.io pending, I still need to provide a follow-up.
- rebased on current main
- this does not need any downstream forks (neither virtio-drivers now safe-mmio)

kernel/src/virtio/mmio.rs
```diff
@@ -102,7 +102,8 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
         // SAFETY: The address is valid, mapped by `map_global_range_4k_shared`, and verified
         // to be VirtIOHeader-aligned by the guard above.
         // The memory region has the same lifetime of the MmioSlot structure which will be consumed by the driver.
-        let Ok(transport) = (unsafe { MmioTransport::<SvsmHal>::new(header) }) else {
+        // TODO: mmio_size = PAGE_SIZE always OK?
```
luigix25: ~for now it's ok, but with device tree support this size should come from the hypervisor.~

luigix25 (commented):
Overall LGTM!

kernel/src/virtio/mmio.rs
luigix25: this is virtio code, so I think we can use sizeof(VirtioHeader).

osteffenrh: mmio_size must include the header and the device config space, which is at offset 0x100 into the mmio area.
Thinking about it, I guess we should leave it as PAGE_SIZE, since this is how QEMU spaces the devices right now. If at some point we have a way to get that information from QEMU automatically, we can use it here.

luigix25: please add a TODO so that we can modify it. This information should be available with the device tree support.

osteffenrh: Turned the TODO into a comment.

osteffenrh:
v3:
- Added subsystem "virtio" to the subject of the first 2 commits.

Cargo.toml
```diff
@@ -65,7 +65,10 @@ packit = { path = "packit" }
 userlib = { path = "user/lib" }
 userinit = { path = "user/init" }
 release = { path = "release" }
-virtio-drivers = { git = "https://github.com/coconut-svsm/virtio-drivers.git", rev = "dcbdbc8cefc0c985322382acdeb633a63a3a4cc1" }
+virtio-drivers = { git = "https://github.com/rcore-os/virtio-drivers.git", rev = "c8786e83e6a64e46ed5cab301048dbaf3aab1444" }
+safe-mmio = { git = "https://github.com/google/safe-mmio", rev = "89f584caaf32fcd4341a5a716c97d67be83edb9c", features = [
+  "custom-mmio",
+] }
```
luigix25: do we need to specify git here, if we already have:
```
[patch.crates-io]
# Make virtio-drivers use our version of safe-mmio and our MMIO backend, via Cargo feature unification.
safe-mmio = { git = "https://github.com/google/safe-mmio", rev = "89f584caaf32fcd4341a5a716c97d67be83edb9c" }
```
?

osteffenrh: Right, we can just include the crates.io version as a dependency.
The override below will redirect svsm's dependency to the git, as well as
virtio-drivers.
When the new release appears on creates.io we simply remove the `[patch...]` line.

osteffenrh:
v4:
- Clean up Cargo.toml changes: safe-mmio points to crates.io now and is patched to the git version.
- Squash Cargo.toml changes (1 commit less now)
- Cargo.lock changes are cleaner now.

Cargo.toml
```diff
@@ -110,6 +111,10 @@ vstd = { version = "=0.0.0-2026-04-12-0118", features = ["alloc", "allow_panic"]
 verify_proof = { path = "verification/verify_proof", default-features = false  }
 verify_external = { path = "verification/verify_external", default-features = false  }
 
+[patch.crates-io]
+# Make virtio-drivers use our version of safe-mmio and our MMIO backend, via Cargo feature unification.
```
stefano-garzarella: What `our version` means here?

I thought we need  `89f584caaf3` because they are not released yet on crates.io.

If this is the case, I'd update this comment, and if possible I'd specify which version we are waiting to be published.

Cargo.toml
```diff
@@ -97,6 +96,7 @@ sha2 = { version = "0.10.8", default-features = false }
 uuid = { version = "1.6.1", default-features = false }
 virtfw-libefi = { version = "0.6" }
 virtfw-varstore = { version = "0.6.1" }
+virtio-drivers = { git = "https://github.com/rcore-os/virtio-drivers.git", rev = "c8786e83e6a64e46ed5cab301048dbaf3aab1444" }
```
stefano-garzarella, line 101: I'd add a comment here to make it clear we are waiting the next release.

Cargo.toml
```diff
@@ -97,6 +96,7 @@ sha2 = { version = "0.10.8", default-features = false }
...
 # Add the derive feature by default because all crates use it.
 zerocopy = { version = "0.8.2", features = ["derive"] }
 embedded-io = { version = "0.6.1" }
```
stefano-garzarella: claude pointed out that we don't need `embedded-io` anymore since it is now pulled byt virtio-drivers dependency

kernel/src/block/virtio_blk.rs
```diff
@@ -22,10 +22,14 @@ extern crate alloc;
 use alloc::boxed::Box;
 
 pub struct VirtIOBlkDriver {
-    device: SpinLock<VirtIOBlk<SvsmHal, MmioTransport<SvsmHal>>>,
+    device: SpinLock<VirtIOBlk<SvsmHal, MmioTransport<'static>>>,
     _mmio_space: GlobalRangeGuard,
 }
 
+// SAFETY: MmioTransport contains NonNull which is !Send, but the MMIO region
+// is externally synchronized via SpinLock.
```
stefano-garzarella: So why we need this?

I just tried the following without any warning:
```diff
diff --git a/kernel/src/block/virtio_blk.rs b/kernel/src/block/virtio_blk.rs
index 15be9c7d..6aa7b90e 100644
--- a/kernel/src/block/virtio_blk.rs
+++ b/kernel/src/block/virtio_blk.rs
@@ -26,10 +26,6 @@ pub struct VirtIOBlkDriver {
     _mmio_space: GlobalRangeGuard,
 }
 
-// SAFETY: MmioTransport contains NonNull which is !Send, but the MMIO region
-// is externally synchronized via SpinLock.
-unsafe impl Send for VirtIOBlkDriver {}
-
 impl core::fmt::Debug for VirtIOBlkDriver {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         f.debug_struct("VirtIOBlkDriver").finish()
diff --git a/kernel/src/vsock/virtio_vsock.rs b/kernel/src/vsock/virtio_vsock.rs
index 1a5ac4e8..af21445a 100644
--- a/kernel/src/vsock/virtio_vsock.rs
+++ b/kernel/src/vsock/virtio_vsock.rs
@@ -29,10 +29,6 @@ pub struct VirtIOVsockDriver {
     _mmio_space: GlobalRangeGuard,
 }
 
-// SAFETY: MmioTransport contains NonNull which is !Send, but the MMIO region
-// is externally synchronized via SpinLock.
-unsafe impl Send for VirtIOVsockDriver {}
-
 impl core::fmt::Debug for VirtIOVsockDriver {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         f.debug_struct("VirtIOVsockDriver").finish()
```

Cargo.toml
osteffenrh: True, the comment is not very clear.

I want to say: The same version as coconut does, to make sure we can inject our MMIO backend ...

Will update.

kernel/src/block/virtio_blk.rs
osteffenrh: Applied.

osteffenrh:
v5:
- rebased
- improved comments in Cargo.toml
- removed unnecessary Send impl for blk and vsock drivers

Cargo.toml
```diff
@@ -111,6 +113,12 @@ vstd = { version = "=0.0.0-2026-04-12-0118", features = ["alloc", "allow_panic"]
 verify_proof = { path = "verification/verify_proof", default-features = false  }
 verify_external = { path = "verification/verify_external", default-features = false  }
 
+[patch.crates-io]
+# Force all dependencies to use the same version of safe-mmio, which has
+# the new custion-mmio feature, so that jinecting our MMIO backend works.
```
luigix25: /s/jinecting/injecting?

osteffenrh: fixed

luigix25 (approved):

stefano-garzarella (approved):

PR #1113, opened 2026-06-12
kernel: support dynamically-sized `SharedBox` + multi-page virtio DMA

00xc:
Add support in `SharedBox` to allocate dynamically-sized slices, just like `PageBox` does. This allows removing a long-standing TODO in the virtio HAL code where we can only allocate a single page for DMA.

This PR also includes some minor cleanups, like removing the unneeded `virtio_init()`.

### DIFF
--- a/kernel/src/mm/page_visibility.rs
+++ b/kernel/src/mm/page_visibility.rs
@@ -4,7 +4,6 @@
 //
 // Author: Jon Lange (jlange@microsoft.com)
 
-use core::mem::MaybeUninit;
 use core::ops::Deref;
 use core::ptr::NonNull;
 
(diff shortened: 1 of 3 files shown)


stefano-garzarella:
@osteffenrh can you take a look at the virtio part of this series (last 2 commits IIUC)?

osteffenrh:
Thanks for this!
virtio related changes are OK.

My only question: Are pages of a SharedBox guaranteed to be one chunk of physical memory?

00xc:
> Thanks for this! virtio related changes are OK.
> 
> My only question: Are pages of a SharedBox guaranteed to be one chunk of physical memory?

Ah, right. Currently yes, the kernel heap is one contiguous guest physical address region which is directly mapped.

00xc:
We agreed to leave this as it is. While pages are physically contiguous now, they may not be in the future, and since there is no immediate need for multi-page DMA, it's safer to leave the assertion.

PR #1114, opened 2026-06-13
task: support multiple waiters in WaitQueue

### LINKED ISSUE #1104: kernel/waitqueue: improve handling of multiple tasks
Currently, `WaitQueue` can only handle one task at a time. It will panic if a task attempts to wait and another task is already queued.

The plan is to allow multiple tasks to queue (or at least to avoid the panic).

This issue was first introduced in this [discussion](https://github.com/coconut-svsm/svsm/pull/1044#discussion_r3161992386).

Is anyone working on this?
@msft-jlange as you have recently worked in that area (in `wait_for_termination`), I was wondering if you have any plans to address it.

I'd be happy to help.

mvanhorn:
## Summary

`WaitQueue` used `Option<TaskPointer>` to hold at most one waiter and panicked if a second task called `wait_for_event` while one was already queued. This made it impossible for multiple tasks to call `wait_for_termination` on the same target task simultaneously.

## Why this matters

As described in #1104, the `wait_for_termination` API is the natural way to join a task, but any scenario where two tasks join the same target — e.g., a parent and a monitor both waiting on a worker — triggers an unconditional panic. The fix unlocks that usage pattern without any heap allocation in the wake path.

## Changes

- **`kernel/src/task/waiting.rs`**: Replace `waiter: Option<TaskPointer>` with `waiters: Vec<TaskPointer>`. Remove the `assert!(self.waiter.is_none())` guard. `wait_for_event` now pushes onto the vec; `wakeup` drains it (alloc-free at wake time) and returns the full list.
- **`kernel/src/task/tasks.rs`**: `set_task_terminated` now returns `Vec<TaskPointer>` instead of `Option<TaskPointer>`.
- **`kernel/src/task/schedule.rs`**: `TaskList::terminate` returns `Vec<TaskPointer>`; `current_task_terminated` iterates over all returned waiters and calls `enqueue_task` for each. Adds `test_wait_for_termination_multiple_waiters`: two tasks each call `wait_for_termination` on a third task; both must resume after it terminates.

Fixes #1104

Signed-off-by: Matt Van Horn <mvanhorn@gmail.com>

### DIFF
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -37,6 +37,7 @@ extern crate alloc;
 use super::tasks::TASK_ACTIVE_OFFSET;
 use super::tasks::TASK_CUR_CPU_OFFSET;
 use super::tasks::TaskExitStatus;
+use super::tasks::TaskWaitListAdapter;
 use super::{
     INITIAL_TASK_ID, KernelThreadStartInfo, Task, TaskListAdapter, TaskPointer, TaskRunListAdapter,
     UserExecInfo,
@@ -282,7 +283,7 @@ impl TaskList {
     /// # Safety
     /// The caller must ensure that `task` is already a member of this task
     /// list.
-    unsafe fn terminate(&mut self, task: TaskPointer) -> Option<TaskPointer> {
+    unsafe fn terminate(&mut self, task: TaskPointer) -> LinkedList<TaskWaitListAdapter> {
         // Set the task state as terminated. If the task being terminated is the
         // current task then the task context will still need to be in scope until
         // the next schedule() has completed. Schedule will keep a reference to this
@@ -292,7 +293,7 @@ impl TaskList {
         let mut cursor = unsafe { self.list().cursor_mut_from_ptr(task.as_ref()) };
         cursor.remove();
 
-        // Inform the caller about any task that may need to be woken.
+        // Inform the caller about any tasks that may need to be woken.
         wakeup
     }
 }
@@ -435,11 +436,11 @@ fn current_task_terminated() {
     // SAFETY: the scheduler guarantees that `current_task` always points to a
     // valid task, and every task has its pointer pushed into the global task
     // list during its creation.
-    let wakeup = unsafe { TASKLIST.lock().terminate(task_node.clone()) };
+    let mut wakeup = unsafe { TASKLIST.lock().terminate(task_node.clone()) };
 
-    // If another thread must be woken as a result of the termination, then
-    // schedule it now.
-    if let Some(wake_task) = wakeup {
+    // Remove each task from the wakeup list before making it runnable so it
+    // can immediately join another wait queue.
+    while let Some(wake_task) = wakeup.pop_front() {
         rq.prepare_run_task(wake_task);
     }
 }
@@ -904,10 +905,12 @@ const CONTEXT_SWITCH_RESTORE_TOKEN: VirtAddr = SVSM_CONTEXT_SWITCH_SHADOW_STACK.
 #[cfg(all(test, test_in_svsm))]
 mod test {
     use super::KernelThreadStartInfo;
+    use super::schedule;
     use super::set_affinity;
     use super::start_kernel_task;
     use super::wait_for_termination;
     use crate::cpu::percpu::{PERCPU_AREAS, this_cpu};
+    use core::sync::atomic::AtomicBool;
     use core::sync::atomic::AtomicU32;
     use core::sync::atomic::Ordering;
 
@@ -959,6 +962,96 @@ mod test {
         // Wait again for the task to terminate.  This should return
         // immediately.
         wait_for_termination(task);
+    }
+
+    #[cfg(test)]
+    static MULTI_WAITER_COUNTER: AtomicU32 = AtomicU32::new(0);
+    static MULTI_WAITER_WAITING_COUNTER: AtomicU32 = AtomicU32::new(0);
+    static MULTI_WAITER_TARGET_RELEASED: AtomicBool = AtomicBool::new(false);
+    // Distinguishes each waiter from the target task's single increment.
+    const MULTI_WAITER_TASK_INCREMENT: u32 = 10;
+
+    fn multi_waiter_target(_: usize) {
+        let current_cpu = this_cpu().get_cpu_index();
+        let target_cpu = PERCPU_AREAS.len() - 1;
+        set_affinity(if current_cpu == target_cpu {
+            0
+        } else {
+            target_cpu
+        });
+
+        while !MULTI_WAITER_TARGET_RELEASED.load(Ordering::Acquire) {
+            if PERCPU_AREAS.len() > 1 {
+                core::hint::spin_loop();
+            } else {
+                schedule();
+            }
+        }
+
+        MULTI_WAITER_COUNTER.fetch_add(1, Ordering::Relaxed);
+    }
+
+    fn waiting_task(target_id: usize) {
+        // Look up the target task by id and wait for it to terminate.
+        let task = super::TASKLIST
+            .lock()
+            .get_task(target_id as u32)
+            .expect("Failed to find multi-waiter target task");
+        super::preemption_checks();
+        let guard = task
+            .wait_for_exit()
+            .expect("Multi-waiter target terminated before waiter blocked");
+        MULTI_WAITER_WAITING_COUNTER.fetch_add(1, Ordering::Release);
+        super::select_new_task(false, Some(guard));
+        MULTI_WAITER_COUNTER.fetch_add(MULTI_WAITER_TASK_INCREMENT, Ordering::Relaxed);
+    }
+
+    #[test]
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_wait_for_termination_multiple_waiters() {
+        MULTI_WAITER_COUNTER.store(0, Ordering::Relaxed);
+        MULTI_WAITER_WAITING_COUNTER.store(0, Ordering::Relaxed);
+        MULTI_WAITER_TARGET_RELEASED.store(false, Ordering::Relaxed);
+
+        // Start the task that will be waited on.
+        let target = start_kernel_task(
+            KernelThreadStartInfo::new(multi_waiter_target, 0),
+            "multi-waiter target",
+        )
+        .expect("Failed to start target task");
+        let target_id = target.get_task_id() as usize;
+
+        // Start two waiter tasks that each wait on the same target task.
+        let waiter1 = start_kernel_task(
+            KernelThreadStartInfo::new(waiting_task, target_id),
+            "waiter 1",
+        )
+        .expect("Failed to start waiter 1");
+
+        let waiter2 = start_kernel_task(
+            KernelThreadStartInfo::new(waiting_task, target_id),
+            "waiter 2",
+        )
+        .expect("Failed to start waiter 2");
+
+        while MULTI_WAITER_WAITING_COUNTER.load(Ordering::Acquire) < 2 {
+            schedule();
+        }
+        assert_eq!(MULTI_WAITER_WAITING_COUNTER.load(Ordering::Acquire), 2);
+
+        // Both waiters have joined the target's wait queue, so allow the
+        // target to terminate before this task waits for it too.
+        MULTI_WAITER_TARGET_RELEASED.store(true, Ordering::Release);
+        wait_for_termination(target);
+        wait_for_termination(waiter1);
+        wait_for_termination(waiter2);
+
+        // Both waiter tasks must have run: multi_waiter_target adds 1,
+        // each waiting_task adds MULTI_WAITER_TASK_INCREMENT.
+        assert_eq!(
+            MULTI_WAITER_COUNTER.load(Ordering::Relaxed),
+            1 + (MULTI_WAITER_TASK_INCREMENT * 2)
+        );
     }
 
     #[test]
--- a/kernel/src/task/waiting.rs
+++ b/kernel/src/task/waiting.rs
@@ -4,26 +4,37 @@
 //
 // Author: Joerg Roedel <jroedel@suse.de>
 
-use super::tasks::TaskPointer;
+use intrusive_collections::LinkedList;
 
-#[derive(Debug, Default)]
+use super::tasks::{TaskPointer, TaskWaitListAdapter};
+
+/// A queue of tasks waiting for a single event. Multiple tasks may wait
+/// simultaneously.
+#[derive(Debug)]
 pub struct WaitQueue {
-    waiter: Option<TaskPointer>,
+    waiters: LinkedList<TaskWaitListAdapter>,
 }
 
 impl WaitQueue {
-    pub const fn new() -> Self {
-        Self { waiter: None }
+    pub fn new() -> Self {
+        Self {
+            waiters: LinkedList::new(TaskWaitListAdapter::new()),
+        }
     }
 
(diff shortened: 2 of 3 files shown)


msft-jlange:
Unfortunately, this design won't work because it requires allocation of memory when entering a wait state (as a result of `Vec::push`).  This makes it impossible for the memory allocator to participate in wait operations, which may be required as a result of system-wide synchronization (this obviously has nothing to with task termination, but `WaitQueue` will ultimately be used for all types of wait operations).

Please reconsider your design to implement a queue without requiring memory allocation.

kernel/src/task/schedule.rs
```diff
@@ -268,7 +268,7 @@ impl TaskList {
     /// # Safety
     /// The caller must ensure that `task` is already a member of this task
     /// list.
-    unsafe fn terminate(&mut self, task: TaskPointer) -> Option<TaskPointer> {
+    unsafe fn terminate(&mut self, task: TaskPointer) -> alloc::vec::Vec<TaskPointer> {
```
msft-jlange: As a general pattern, we prefer a `use` statement in the file so that types in declarations do not require namespace decoration.  So you would want `use alloc::vec::Vec` at the top of the file and then this just becomes `Vec<TaskPointer>`.

kernel/src/task/schedule.rs
```diff
@@ -936,6 +935,59 @@ mod test {
         wait_for_termination(task);
     }
 
+    static MULTI_WAITER_COUNTER: AtomicU32 = AtomicU32::new(0);
```
msft-jlange, line 968: Is this just used for testing?  If so, it should be guarded by `#[cfg(test)]`.  Otherwise, I don't understand what this is for.

kernel/src/task/schedule.rs
```diff
@@ -936,6 +935,59 @@ mod test {
...
+        MULTI_WAITER_COUNTER.fetch_add(1, Ordering::Relaxed);
+    }
+
+    fn waiting_task(target_id: usize) {
+        // Look up the target task by id and wait for it to terminate.
+        let target = super::TASKLIST.lock().get_task(target_id as u32);
+        if let Some(task) = target {
+            wait_for_termination(task);
+        }
+        MULTI_WAITER_COUNTER.fetch_add(10, Ordering::Relaxed);
```
msft-jlange: What does `10` mean?  Is it a magic value?

kernel/src/task/waiting.rs
```diff
@@ -4,26 +4,36 @@
...
         current_task.set_task_blocked();
-        self.waiter = Some(current_task);
+        self.waiters.push(current_task);
     }
 
-    pub fn wakeup(&mut self) -> Option<TaskPointer> {
-        self.waiter.take()
+    /// Wake all waiting tasks. Returns the drained list of waiters so the
+    /// caller can schedule each one. The drain itself does not allocate.
+    pub fn wakeup(&mut self) -> Vec<TaskPointer> {
```
msft-jlange: In general, some wake operations will want to wake the entire queue and some operations will only want to wake the head of the queue.  Waiting for task termination clearly falls into former category but turnstile-type wait operations can only wake one at a time.  While it's not necessary to implement this now, it would be nice to anticipate this by defining a parameter to this function (e.g. `wake_all: bool`) that permits both types.  If you don't want to implement both now, then you can just `assert!(wake_all)` for now and make it a problem for the future, but defining this now at least requires callers to be conscious of the type of wakeup so that we don't have to reconsider the behavior of entire code base when this single wakeup is finally implemented.

msft-jlange (changes requested):
Please reimplement as described without using memory allocation.

joergroedel:
> Unfortunately, this design won't work because it requires allocation of memory when entering a wait state (as a result of `Vec::push`). This makes it impossible for the memory allocator to participate in wait operations, which may be required as a result of system-wide synchronization (this obviously has nothing to with task termination, but `WaitQueue` will ultimately be used for all types of wait operations).
> 
> Please reconsider your design to implement a queue without requiring memory allocation.

I second this, the prefered data structure to use here is a linked list from `intrusive_collections`.

mvanhorn:
Reworked in d2626cd per both of your points: the Vec is gone. The waiters queue is now an intrusive linked list from intrusive_collections, with the LinkedListAtomicLink embedded in Task next to the existing runlist link, so wait_for_event() performs no allocation and the allocator can safely participate in waits. wakeup() takes the list and hands it back to the scheduler for wake-up. cargo check for the svsm kernel target passes locally.

stefano-garzarella:
@mvanhorn please rebase since there are conflicts and also check all comments and resolve ones fixed or explain why they are not fixed.

mvanhorn:
Rebased onto main. There was one conflict in `current_task_terminated`: main moved task wakeup to `rq.prepare_run_task`, while this branch changed `terminate()` to wake multiple waiters, so I kept the multi-waiter loop and scheduled each via `rq.prepare_run_task`. `cargo check` for `x86_64-unknown-none` passes, clippy and fmt clean. (Couldn't run the host unit tests locally - this is an arm64 box and the `syscall` crate uses x86 asm - so CI will cover those.)

Still working through @msft-jlange's review comments (the `Vec` import, the `#[cfg(test)]` guard, the magic value, and the wake-all vs wake-head question); I'll push those separately.

msft-jlange:
> Still working through @msft-jlange's review comments (the `Vec` import, the `#[cfg(test)]` guard, the magic value, and the wake-all vs wake-head question); I'll push those separately.

Thanks for continuing to work through these.  I haven't had a chance to look at your updates yet, but your description is certainly headed in the right direction.

One other issue - we require each commit in a PR to be separately stable, and your updates have not resolved the `Vec` issue in the first commit (they are resolved in a separate commit).  You should plan to squash those changes into a single commit so the PR has no intermediate unstable commits.

mvanhorn:
All four items addressed as of 8fd94da:

- Vec import / namespace decoration: moot after the previous push - waiters now use an intrusive linked list, so there's no Vec left in the file.
- cfg(test) guard: the counter and helpers live inside the `#[cfg(all(test, test_in_svsm))]` test module.
- Magic 10: now `MULTI_WAITER_TASK_INCREMENT`, and the final assertion is derived from it.
- Wake semantics: `wakeup()` takes `wake_all: bool` as you suggested. Termination passes true (wake the whole queue); false removes only the head waiter, so turnstile-type waits are supported without revisiting callers later.

Verified: cargo check and clippy clean for x86_64-unknown-none, fmt clean. The multi-waiter test is test_in_svsm, so it runs in-guest via CI.

kernel/src/task/waiting.rs
```diff
@@ -4,26 +4,49 @@
...
+            waiters
+        }
+    }
+}
+
+impl Default for WaitQueue {
+    fn default() -> Self {
+        Self::new()
     }
 }
```
msft-jlange: This is not needed after #1162 and should be removed.

kernel/src/task/schedule.rs
```diff
@@ -951,6 +950,63 @@ mod test {
...
+        )
+        .expect("Failed to start waiter 1");
+
+        start_kernel_task(
+            KernelThreadStartInfo::new(waiting_task, target_id),
+            String::from("waiter 2"),
+        )
+        .expect("Failed to start waiter 2");
+
+        // Wait for the target task from this task too, ensuring it has
```
msft-jlange: Nothing here ensures that waiter 1 and waiter 2 enter the wait state while the target task is still executing so your tests cannot prove that the waiters actually join the wait queue.  In fact, because your target task runs on the same CPU that the test function runs on, it is virtually guaranteed that the target task will terminate before either waiter thread is even created, so it is overwhelmingly likely that there is no wait queue testing occurring here.  A proper test would require the following sequence:
- Change the target task to set affinity to some processor other than the originating processor (use the same logic that is in `empty_task()`).
- Add a barrier to the target task so that it will spin-wait for the barrier to be signaled until it proceeds to termination.
- Signal the barrier from the test task after both waiters have been created but before the test task performs its own wait.  This will guarantee that both waiter 1 and waiter 2 have entered wait states and will also guarantee that the target task has not terminated before both waiters have entered their wait states.

msft-jlange:
@mvanhorn More actions are still required.
- All commits must be squashed before merging to maintain our stable commit rule.
- All comments must be resolved.  There are new comments since your last push that identify additional issues.  In addition, comments that appear to be addressed by your latest updates still need to be resolved.
- Rebasing is required to eliminate merge conflicts.

kernel/src/task/schedule.rs
```diff
@@ -434,9 +435,9 @@ fn current_task_terminated() {
     // list during its creation.
     let wakeup = unsafe { TASKLIST.lock().terminate(task_node.clone()) };
 
-    // If another thread must be woken as a result of the termination, then
-    // schedule it now.
-    if let Some(wake_task) = wakeup {
+    // Wake all tasks that were waiting for this task to terminate, scheduling
+    // each on the current runqueue.
+    for wake_task in wakeup {
```
msft-jlange: I think this is subject to race conditions.  Once a task has been woken, it can immediately attempt to enter another wait state.  However, for the duration of this loop, the task's wait list link is still inserted into the wakeup list here.  That means that an attempt by the task to enter a new wait state will panic when it tries to insert itself into a different wait queue.  I believe it is necessary to ensure that the task being woken is removed from the local wakeup list before the wakeup can be attempted.

mvanhorn:
Review feedback addressed and amended into the signed commit (9166b00, history kept squashed per repo convention). Verified locally: targeted x86_64 svsm build and clippy pass offline.

kernel/src/task/schedule.rs
```diff
@@ -960,6 +963,93 @@ mod test {
...
+    #[cfg_attr(not(test_in_svsm), ignore = "Can only be run inside guest")]
+    fn test_wait_for_termination_multiple_waiters() {
+        MULTI_WAITER_COUNTER.store(0, Ordering::Relaxed);
+        MULTI_WAITER_WAITING_COUNTER.store(0, Ordering::Relaxed);
+        MULTI_WAITER_TARGET_RELEASED.store(false, Ordering::Relaxed);
+
+        if PERCPU_AREAS.len() < 2 {
+            return;
+        }
+
```
msft-jlange: I wouldn't include this.  Obviously the contours of the test are quite different between one and many CPUs but the test can still execute on a single CPU; it just might not test everything we want.  We should consider the possibility that thread scheduling logic will change in the future, and it might turn out to be the case in the future that we can orchestrate the order of task execution such that we get very robust testing on a single CPU.  I wouldn't want to have to rewrite this test when that day comes.

mvanhorn:
Updated in ed3241d - dropped the single-CPU early return so the test runs everywhere, and the wait loop yields via schedule() on one CPU instead of spinning so the target task can make progress. Couldn't run the full build locally (macOS host), so leaning on CI to confirm.

mvanhorn:
Thanks for the detailed review. Pushed `5864ba1` for the wake-loop race you flagged:

- `terminate()` / `WaitQueue::wakeup()` now drain the wait list into an owned `Vec<TaskPointer>`, so every waiter is popped off (and unlinked from) the intrusive list before any of them is woken. That closes the window where a just-woken task re-entering a wait state would panic while its link was still in the wakeup list.
- `wakeup()` now `assert!(wake_all)` so callers stay conscious of the wakeup type until the wake-head path is actually needed.
- Also dropped the post-#1162 dead code, gated the test-only counter under `#[cfg(test)]`, and documented the per-waiter increment constant.

The multi-waiter test already pins the target to a separate CPU and holds its termination on the `TARGET_RELEASED` barrier until both waiters have signaled, matching the sequence you outlined — happy to tighten it further if you see a remaining gap.

mvanhorn:
Reworked per your comments in 34ceb87: terminate() hands back the wait-queue list and each waiter is unlinked before prepare_run_task, so a woken task can immediately join another wait queue. Also signed off the earlier commit that was missing DCO.

kernel/src/task/waiting.rs
```diff
@@ -29,18 +29,11 @@ impl WaitQueue {
...
-        if wake_all {
-            self.waiters.take()
-        } else {
-            let mut waiters = LinkedList::new(TaskWaitListAdapter::new());
-            if let Some(task) = self.waiters.pop_front() {
-                waiters.push_back(task);
-            }
-            waiters
-        }
+        assert!(wake_all);
```
00xc, line 36: I guess this is a leftover. Otherwise, I don't see the point of the bool parameter which is always asserted to be true.

msft-jlange: This was something I asked for in a previous comment.  In the future, we will want to support waking only a single waiter in addition to the current implementation that can wake all waiters.  I believe that defining the parameter now makes it easier to add the new behavior later so that all call sites are already using the correct parameters, and don't have to be revisited to account for the new behavior.  Since single-wake is not currently implemented, an assert is present to prevent accidental use of single-wake.

mvanhorn:
Squashed to a single commit at 1d97919 for the stable commit rule. The tree is byte-identical to the previous head (34ceb87), so this is purely a history rewrite; the commit body now folds in the re-entry fix, the wake_all rationale, and the test changes rather than leaving them as separate fixup commits.

On the `wake_all` parameter that @00xc flagged: I kept it per your earlier request, asserted true for now, so that adding wake-one later does not require touching every call site. Your reply on that thread covered it and it is resolved.

All nine review threads are resolved as of this push. I was not able to run the full build, clippy, or the test suite locally: the packit submodule is unavailable in my sandbox and cannot be fetched, so I am relying on CI for those. The affected files do pass rustfmt directly.

Let me know if you would rather the squash keep the three subjects as trailers instead of the prose body.

kernel/src/task/waiting.rs
00xc: Can we add a TODO comment then?

msft-jlange:
```suggestion
        // TODO: handle the case of waking a single waiter.
        assert!(wake_all);
```

msft-jlange:
@mvanhorn this is in pretty good shape.  There is one outstanding piece of feedback from @00xc requesting a TODO comment to be added (which I agree with).  The PR needs to be rebased to resolve merge conflicts.  I am going to perform testing on a single-processor system to ensure that your use of the scheduler works correctly (the code looks like it should work properly).  Once my testing is complete and you have added the comment and rebased the PR, then it should be ready to merge.

mvanhorn:
Both done. Added the TODO exactly as you suggested, and merged main to clear the conflicts. The only conflict was in `schedule.rs`, where main dropped the `String` and `AtomicBool` imports; kept them since the test module here still uses both.

I couldn't build locally to confirm (this clone is missing the `packit` submodule), so CI is the check. Interested in the single-processor results when you have them.

msft-jlange:
@mvanhorn All tests are passing (including single-CPU).  However, your merge broke the bisecting commits rule because you did a merge instead of a rebase.  Can you please rebase or otherwise adjust your commits so each commit is separately stable?

mvanhorn:
Rebased onto current main - the merge commit is gone and the branch is now a single commit on top of a9d05c68, so the bisecting rule holds.

I squashed rather than keeping the three commits, because they were not individually bisectable: the second and third were review fixups to the first, and the third specifically fixed a build break in the first (the `String` task names failed to resolve in the lib-test build). Keeping them separate would have left a commit that does not compile.

The rebase picked up 17fdd6d5 (userspace console handler removal) in `tasks.rs`. My changes do not touch `stdout_open` or the object handles, so there was no overlap to resolve. Net diff against main is unchanged: `schedule.rs`, `tasks.rs`, `waiting.rs`.

@00xc the TODO is in as you asked.

msft-jlange (approved):

joergroedel (approved):

PR #1115, opened 2026-06-14
kernel/tests: fix test_fpu_context_switch

n-ramacciotti:
From #926:
`
 CI is failing [1], I don't know why (yet), but at least now it fails in the right place. (before it failed after the print that all kernel tests succeeded, which was not true [2] as this is a failure in a kernel test)
 `

After some testing [3], a possible scheduling order is as follows: when the `SVSM test task` starts `task1` in `test_fpu_context_switch`, `task1` is scheduled. However, when `task1` starts `task2`, the `SVSM test task` is scheduled and continues until it finishes (printing `All tests passed!`, which is not true). Then `task2` is scheduled until it finishes, after which `task1` resumes.

This is not the correct behaviour as we want that all the tests to finish before printing `All tests passed!`.

Add `wait_for_termination` after starting `task1` and `task2`, so `SVSM test task` can continue at the right time.

Note: I'm not sure if the relative ordering between `task1` and `task2` is relevant for the test (and the assertion failure), but we must be sure that they are both terminated before `SVSM test task` continues.

[1] https://github.com/coconut-svsm/svsm/actions/runs/27468738903/job/81195646363?pr=926
[2] https://github.com/coconut-svsm/svsm/actions/runs/27468172423/job/81194123020
[3] https://github.com/n-ramacciotti/svsm/actions/runs/27494307809/job/81265414543?pr=22

### DIFF
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -1094,14 +1094,14 @@ mod tests {
     test_fpu:
         movq $0x3ff, %rax
         shl $52, %rax
-        // rax contains 1 in Double Precison FP representation
-        movd %rax, %xmm1
+        // rax contains 1.0 in Double Precision FP representation
+        movq %rax, %xmm1
         movapd %xmm1, %xmm3
 
         movq $0x400, %rax
         shl $52, %rax
-        // rax contains 2 in Double Precison FP representation
-        movd %rax, %xmm2
+        // rax contains 2.0 in Double Precision FP representation
+        movq %rax, %xmm2
 
         divsd %xmm2, %xmm3
         movq $0, %rax
@@ -1117,16 +1117,16 @@ mod tests {
         movq $1, %rax
         movq $0x3ff, %rbx
         shl $52, %rbx
-        // rbx contains 1 in Double Precison FP representation
-        movd %rbx, %xmm4
+        // rbx contains 1.0 in Double Precision FP representation
+        movq %rbx, %xmm4
         movapd %xmm4, %xmm6
         comisd %xmm4, %xmm1
         jnz 1f
 
         movq $0x400, %rbx
         shl $52, %rbx
-        // rbx contains 2 in Double Precison FP representation
-        movd %rbx, %xmm5
+        // rbx contains 2.0 in Double Precision FP representation
+        movq %rbx, %xmm5
         comisd %xmm5, %xmm2
         jnz 1f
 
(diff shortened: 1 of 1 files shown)


n-ramacciotti:
CI is failing also here for the same reason as in #926. 

Is it possible that the xstate is not being saved/restored properly or that it is being modified in the meantime?

The problem seems to be related to the introduction of `wait_for_termination`. However, even without the wait in `test_fpu_context_switch`, the same problem happens in the test in svsm PR (at the wrong point, as already mentioned).  Is it possible that it happens during the selection of a new task? (as in test in svsm probably happens during `exec_user`, seeing the log [2], which also calls `schedule` ( therefore `select_new_task`).

The existing tests on `wait_for_termination` do not cause a problem as they occur before the `test_fpu_context_switch` test.

Any help is appreciated!

kernel/src/task/tasks.rs
```diff
@@ -1090,14 +1090,14 @@ mod tests {
     test_fpu:
         movq $0x3ff, %rax
         shl $52, %rax
-        // rax contains 1 in Double Precison FP representation
-        movd %rax, %xmm1
+        // rax contains 1.0 in Double Precison FP representation
+        movq %rax, %xmm1
```
luigix25, line 1098: I think this is odd, but not wrong. If you compare the output of `movd` and `movq` they emit the exact same instruction. 

The emitted instruction changes if you tried `movd %ead, %xmm1`

For reference: https://godbolt.org/z/xPjWG5Mf5

xmm1 is a 128-bit long register, with `movq` the upper 64 bits are always zeroed.

n-ramacciotti: Good observation!

I did this test to see if the problem was in those functions or elsewhere, but it didn't solve the CI issue (as you noticed, they emit the same instruction)

However, I think this commit can still be maintained with an updated commit message for accuracy.

luigix25: > However, I think this commit can still be maintained with an updated commit message for accuracy.

No objections from my side!

n-ramacciotti:
v2:
- change `movq` commit description as suggested by @luigix25

n-ramacciotti:
(without this PR or #926) CI has never failed before as  `test_fpu_context_switch` never terminates: when creating `task2`, `svsm test task` is scheduled and it call `exit(qemu::success)`, stopping the test

This can be seen here: https://github.com/n-ramacciotti/svsm/actions/runs/27538565227/job/81394141092?pr=25

n-ramacciotti:
v3:
- rebased on latest main (with @luigix25 's fix)

luigix25 (approved):

kernel/src/task/tasks.rs
```diff
@@ -1094,14 +1094,14 @@ mod tests {
     test_fpu:
         movq $0x3ff, %rax
         shl $52, %rax
-        // rax contains 1 in Double Precison FP representation
-        movd %rax, %xmm1
+        // rax contains 1.0 in Double Precison FP representation
```
stefano-garzarella: Since we are here, what about fixing also `Precison` ?

```suggestion
        // rax contains 1.0 in Double Precision FP representation
```

It seems we have it in several other places

n-ramacciotti:
v4:
- Fixed comment (Precison->Precision) @stefano-garzarella

stefano-garzarella (approved):

PR #1117, opened 2026-06-15
Prepare switch to KVM Planes

joergroedel:
Prepare the switch to KVM planes:

* Update installation documentation with new required QEMU command line options
* Add the QEMU command line options to `launch_guest.sh`

### DIFF
--- a/Documentation/docs/installation/INSTALL.md
+++ b/Documentation/docs/installation/INSTALL.md
@@ -26,7 +26,7 @@ support patches applied is needed on the host machine.
 A repository based on the SNP-host patches with support for
 `guest-memfd` and SVSM support on-top is available here:
 [https://github.com/coconut-svsm/linux](https://github.com/coconut-svsm/linux).
-It is based on kernel 6.5 and code written by AMD to support [linux-svsm](https://github.com/AMDESE/linux-svsm/).
+It is based on kernel 7.1 and code written by AMD to support [linux-svsm](https://github.com/AMDESE/linux-svsm/).
 
 To use it, check out the svsm branch:
 
@@ -39,16 +39,20 @@ git checkout svsm
 Build, install and boot a kernel from that branch. For best chances of
 success use a kernel configuration provided by the distribution. Make
 sure the configuration includes support for AMD Secure Processor which is
-a requirement for SEV support (`CONFIG_KVM_AMD_SEV`). On openSUSE (other
-distributions may vary) the kernel configuration can be obtained by:
+a requirement for SEV support (`CONFIG_KVM_AMD_SEV`). On openSUSE the kernel configuration can be obtained by:
 
 ```shell
 gunzip -c /proc/config.gz > .config
 make olddefconfig
 ```
 
-After the new kernel is booted, the kernel log contains SEV-SNP
-initialization messages:
+Other distributions like Ubuntu ususally provide to configuration of the
+installed kernels in the `/boot` directory. Copy the configuration file
+matching your running kernel version to `.config` in the SVSM kernel source
+tree and run `make olddefconfig`.
+
+After the new kernel is compiled, installed, and booted, the kernel log contains
+SEV-SNP initialization messages:
 
 ```plain
 $ dmesg | grep SEV
@@ -68,10 +72,10 @@ Building QEMU
 
 COCONUT-SVSM is packaged during the build into a file conforming to the
 [Independent Guest Virtual Machine (IGVM)
-format](https://docs.rs/igvm_defs/0.1.3/igvm_defs/index.html). Current versions
-of QEMU do not support launching guests using IGVM, but a branch is available
-that includes this capability. This will need to be built in order to be able to
-launch COCONUT-SVSM.
+format](https://docs.rs/igvm_defs/0.4.0/igvm_defs/). Current versions
+of QEMU do not support launching an SVSM with guest support, but a branch is
+available that includes this capability. This will need to be built in order to
+be able to launch COCONUT-SVSM.
 
 First make sure to have all build requirements for QEMU installed. RPM
 and DEB based distributions provide ways to install build dependencies
@@ -81,6 +85,12 @@ enabled and then the packages can be installed by:
 ```shell
 sudo zypper refresh
 sudo zypper si -d qemu-kvm
+```
+
+On Debian and derived distributions you need to activate source repositories an install build dependencies with:
+
+```shell
+sudo apt build-dep qemu-system-x86
 ```
 
 Support for IGVM within QEMU depends on the IGVM library. This needs to be
@@ -105,7 +115,7 @@ and switch to the branch that supports IGVM:
 ```shell
 git clone https://github.com/coconut-svsm/qemu
 cd qemu
-git checkout svsm-igvm
+git checkout svsm
 ```
 
 Now the right branch is checked out and you can continue with the build.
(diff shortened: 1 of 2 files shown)


scripts/launch_guest.sh
```diff
@@ -153,7 +153,7 @@ case "$CGS" in
     echo "Error: Unexpected CGS value '$CGS'"
     exit 1
 esac
-MACHINE=q35,memory-backend=mem0,igvm-cfg=igvm0,accel=$ACCEL
```
stefano-garzarella, line 156: Should this change go in the `[ -n "$SNP_GUEST" ]` case? Or this will also work with native?

Documentation/docs/installation/INSTALL.md
```diff
@@ -68,10 +72,10 @@ Building QEMU
 
 COCONUT-SVSM is packaged during the build into a file conforming to the
 [Independent Guest Virtual Machine (IGVM)
-format](https://docs.rs/igvm_defs/0.1.3/igvm_defs/index.html). Current versions
-of QEMU do not support launching guests using IGVM, but a branch is available
-that includes this capability. This will need to be built in order to be able to
-launch COCONUT-SVSM.
+format](https://docs.rs/igvm_defs/0.4.0/igvm_defs/). Current versions
+of QEMU do not support launching and SVSM with guest support, but a branch is
```
stefano-garzarella:
```suggestion
of QEMU do not support launching an SVSM with guest support, but a branch is
```
?

Documentation/docs/installation/INSTALL.md
```diff
@@ -39,16 +39,20 @@ git checkout svsm
...
 ```
 
-After the new kernel is booted, the kernel log contains SEV-SNP
-initialization messages:
+Other distributions like Ubuntu ususally provide to configuration of the
+installed kernels in the `/boot` directory. Copy the configuration file
+matching your running kernel version to `.config` in the SVSM kernel source
+tree and run `make olddefconfig`.
+
+After the new kernel is compile, installed, and booted, the kernel log contains
```
stefano-garzarella:
```suggestion
After the new kernel is compiled, installed, and booted, the kernel log contains
```

scripts/launch_guest.sh
luigix25: I guess we drop all the reference to QEMU 10 as we won't support it anymore

scripts/launch_guest.sh
joergroedel: Thanks, fixed. I also added detection for the device-plane property.

scripts/launch_guest.sh
joergroedel: Changed it to check for QEMU 11.

luigix25: here as well

joergroedel: Removed, thanks.

scripts/launch_guest.sh
```diff
@@ -130,20 +130,23 @@ QEMU_BUILD=${QEMU_VERSION##*.}
...
-  echo "Error: SVSM requires QEMU 10.1 or newer (with patches)." >&2
+if (( QEMU_MAJOR < 11 )); then
+  echo "Error: SVSM requires QEMU 11.0 or newer (with patches)." >&2
   exit 1
 fi
 
+# Check for support for device-plane property
+QEMU_PLANE_PROPS=""
+QEMU_PLANES=$($QEMU -h | grep -o "device-plane=" || exit 0)
+if [ "$QEMU_PLANES" == "device-plane="]; then
```
stefano-garzarella:
```suggestion
if [ "$QEMU_PLANES" == "device-plane=" ]; then
```

Without this change plane 2 is not working, I saw:

```
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.06s
     Running `target/debug/cbit`
./scripts/launch_guest.sh: line 141: [: missing `]'
=============================
Launching SVSM guest
=============================
...
[SVSM] [CPU 0] Virtual memory pages used: 0 * 4K, 0 * 2M
[SVSM] VMSA PA: 0x8000f4b000
[SVSM] Launching Firmware
Failed to create plane 2
./scripts/launch_guest.sh: line 206: 28483 Aborted                    (core dumped) $QEMU -cpu $CPU -machine $MACHINE -object $MEMORY $IGVM_OBJ $SNP_GUEST -smp 4 -no-reboot $QEMU_NETDEV $IMAGE_DISK -nographic -vga none $COM1_SERIAL $COM2_SERIAL $COM3_SERIAL $COM4_SERIAL $QEMU_EXIT_DEVICE $QEMU_TEST_IO_DEVICE $VIRTIO_CONFIG $STATE_DEVICE $VSOCK_DEVICE "$@"
```

With that applied is working, but I'm wondering why there is a core dump if `,kernel-irqchip=split,device-plane=2` is not specified?

Anyway something to fix on QEMU patches, so not blocking this PR at all (I mean the core dump, but we should fix our script as suggested).

stefano-garzarella (changes requested):

scripts/launch_guest.sh
```diff
@@ -154,7 +157,7 @@ case "$CGS" in
     exit 1
 esac
 MACHINE=q35,memory-backend=mem0,igvm-cfg=igvm0,accel=$ACCEL
-[ -n "$SNP_GUEST" ] && MACHINE+=",confidential-guest-support=cgs0"
+[ -n "$SNP_GUEST" ] && MACHINE+=",confidential-guest-support=cgs0$QEMU_PLANE_PROPS"
```
stefano-garzarella: What about this to follow the style we have here:
```diff
--- a/scripts/launch_guest.sh
+++ b/scripts/launch_guest.sh
@@ -136,11 +136,7 @@ if (( QEMU_MAJOR < 11 )); then
 fi
 
 # Check for support for device-plane property
-QEMU_PLANE_PROPS=""
-QEMU_PLANES=$($QEMU -h | grep -o "device-plane=" || exit 0)
-if [ "$QEMU_PLANES" == "device-plane="]; then
-  QEMU_PLANE_PROPS=",kernel-irqchip=split,device-plane=2"
-fi
+QEMU_PLANES=$($QEMU -h | grep -o "device-plane=" || true)
 
 case "$CGS" in
   nocc)
@@ -157,7 +153,8 @@ case "$CGS" in
     exit 1
 esac
 MACHINE=q35,memory-backend=mem0,igvm-cfg=igvm0,accel=$ACCEL
-[ -n "$SNP_GUEST" ] && MACHINE+=",confidential-guest-support=cgs0$QEMU_PLANE_PROPS"
+[ -n "$SNP_GUEST" ] && MACHINE+=",confidential-guest-support=cgs0"
+[ -n "$SNP_GUEST" ] && [ -n "$QEMU_PLANES" ] && MACHINE+=",kernel-irqchip=split,device-plane=2"
 [ -n "$VIRTIO_ENABLE" ] && MACHINE+=",${VIRTIO_ENABLE}"
 
 IGVM_OBJ="-object igvm-cfg,id=igvm0,file=$IGVM"
```

scripts/launch_guest.sh
luigix25: do we really need to check for planes support?

If we want to launch a guest we need planes. If we are on nocc mode, planes are not available.

Therefore I think we can reuse `SNP_GUEST` to add the new properties. 

Retro-compatibility should not be a concern, right?

WDYT?

joergroedel: This is just for the transition period where people use an updated SVSM repo with the old Linux/QEMU. The old QEMU will not understand `device-plane` and errors out on that. Having the detection makes it work for the old and new setup.

scripts/launch_guest.sh
joergroedel: That's better, I applied your diff, thanks.

stefano-garzarella (approved):
Tested with the new linux and qemu on SNP machine, everything fine: linux boots, attestation works with vsock device, and test-in-svsm pass.

PR #1118, opened 2026-06-15
kernel: mm/pagetables: ensure progress in `PageTable::unmap_region()`

00xc:
If `PageTable::unmap_region()` attempts to unmap a non-existing page, it logs the error but does not advance the loop index, resulting in an infinite loop. Ensure that `vaddr` is advanced by the minimum supported page size.

Note that this function is currently unused.

### DIFF
--- a/kernel/src/mm/pagetable.rs
+++ b/kernel/src/mm/pagetable.rs
@@ -1238,6 +1238,7 @@ impl PageTable {
                 }
                 _ => {
                     log::error!("Can't unmap - address not mapped {vaddr:#x}");
+                    vaddr = vaddr + PAGE_SIZE;
                 }
             }
         }


kernel/src/mm/pagetable.rs
```diff
@@ -1270,37 +1281,55 @@ impl PageTable {
...
+                log::warn!("Page table walk ended in non-leaf PTE");
+                return Err(SvsmError::Mem);
             }
+
+            // If the page that this PTE maps falls entirely within the requested
+            // region, simply mark the PTE as read-only. This is guaranteed to be
+            // the case for 4k mappings, since `region` is not empty and 4k-aligned
+            let page_region = MemoryRegion::new(page, size);
+            if region.contains_region(&page_region) {
+                entry.make_ro();
```
peterfang: Don't you erase the HUGE bit here if this is a huge page?

kernel/src/mm/pagetable.rs
```diff
@@ -1270,37 +1281,55 @@ impl PageTable {
...
+                page = page + size;
+                continue;
+            }
+
+            // This is a huge page, and only part of it falls within the requested
+            // region. Split the page into 4k pages and change permissions up until
+            // we fulfill the request. This also means we are done iterating.
+            Self::do_split_4k(entry)?;
```
peterfang: This is dangerous. `do_split_4k()` only operates on 2M pages, which is not very clear from the function name itself. So if you want to do this, maybe don't allow 1G pages when you evaluate `self.walk_addr(page)`?

kernel/src/mm/pagetable.rs
```diff
@@ -1238,6 +1248,7 @@ impl PageTable {
                 }
                 _ => {
                     log::error!("Can't unmap - address not mapped {vaddr:#x}");
+                    vaddr = vaddr + PAGE_SIZE;
```
peterfang, line 1241: Great catch! But I'm now curious about the `log::error!()` above. Do we actually want unmapping to fail? Should we fail more loudly?

There are no callers to this function. So does it make sense to keep it?

kernel/src/mm/pagetable.rs
```diff
@@ -1270,37 +1281,55 @@ impl PageTable {
...
-                    if entry.huge() {
-                        return Err(SvsmError::Mem);
-                    }
-                }
-                _ => {}
+            // This must either be a last-level PTE or a huge PTE
+            if size != PAGE_SIZE && !entry.huge() {
+                log::warn!("Page table walk ended in non-leaf PTE");
+                return Err(SvsmError::Mem);
             }
```
peterfang: I think this should not happen due to the implementation of `walk_addr()`. I have trouble figuring out why it would return a non-leaf PTE? Is this just to ensure the function is implemented correctly?

00xc: It should not, this is just to guard against someone changing the implementation in the future. Maybe we can just panic instead.

kernel/src/mm/pagetable.rs
00xc: Right, I'll fix `PTEntry::make_ro()`

kernel/src/mm/pagetable.rs
00xc: Uh, right. I could make this work for 1G pages but I think the changes will be too intrusive and not worth it.considering we are not yet making use of huge pages in read only sections, and that there is a page table refactor on the way. I will just be dropping those changes and only keep the infinite loop fix.

kernel/src/mm/pagetable.rs
00xc: I'd say we keep it in case someone ends up using it. I have added a sentence in the commit message indicating that the function is currently unused.

peterfang: Sounds good to me

peterfang (approved):

PR #1119, opened 2026-06-15
kernel/task: preserve r12-r15 across context switch

luigix25:
`switch_context` is declared as extern "C", so the Rust compiler expects
it to follow the System V x86-64 ABI, which requires preserving all
callee-saved registers: rbx, rbp, r12, r13, r14, and r15.
    
However, commit ef4cd9ee ("task: streamline register preservation during
context switch") removed the push/pop of r12-r15 from the
`switch_context` assembly, keeping only rbx and rbp. This means that after
a context switch, r12-r15 contain stale values from whatever task last
ran on that CPU — not the values the compiler expects.
    
This issue, reported by Nicola, caused an assert in `test_fpu_context_switch`,
when he added a call to `wait_for_termination` to task1 and task2 to
address a scheduling problem.
    
Fix this by saving and restoring r12-r15 in the switch_context assembly
and adding the corresponding fields to X86TaskSwitchRegs. Also update
the gdbstub TaskContext-to-register conversion to report the now
available r12-r15 values instead of zeros.

This issue was found by @n-ramacciotti in #1115

/cc @msft-jlange @n-ramacciotti

### DIFF
--- a/kernel/src/debug/gdbstub.rs
+++ b/kernel/src/debug/gdbstub.rs
@@ -480,10 +480,10 @@ pub mod svsm_gdbstub {
                 0, // r9
                 0, // r10
                 0, // r11
-                0, // r12
-                0, // r13
-                0, // r14
-                0, // r15
+                value.regs.r12 as u64,
+                value.regs.r13 as u64,
+                value.regs.r14 as u64,
+                value.regs.r15 as u64,
             ];
             regs
         }
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -703,6 +703,10 @@ global_asm!(
         // stack.
         pushq   %rbp
         pushq   %rbx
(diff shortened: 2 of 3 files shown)


n-ramacciotti:
Thanks for this!!

I will rebase #1115 after this is merged!

n-ramacciotti:
From #1115:
`
CI has never failed before as test_fpu_context_switch never terminates: when creating task2, svsm test task is scheduled and it call exit(qemu::success), stopping the test
`

Is this masking worth mentioning? In the commits in my PR I only say that the order is wrong and that a wait should be added.

msft-jlange (approved):
Thanks for fixing this!  I mistakenly assumed the r12-r15 were caller save, but it turns out they are not.  Your fix looks good to me.

joergroedel (approved):

PR #1121, opened 2026-06-16
build: link virtio-drivers dependency to block/vsock features

armenon-rh:
Enabling the block or vsock subsystems alone does not automatically pull the virtio-drivers dependency. It silently skips the dependency during compilation.

Make the block and vsock Cargo features depend on the virtio-drivers feature in the kernel Cargo.toml, thereby allowing it to automatically pull virtio-drivers.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -80,10 +80,10 @@ verus_all = ["verus_builtin", "verus_builtin_macros", "vstd", "verify_proof/veru
 verus = ["verus_all", "verify_proof/noverify", "verify_external/noverify"]
 noverify = []
 virtio-drivers = ["dep:virtio-drivers"]
-block = []
-vsock = []
 uefivars = ["dep:virtfw-varstore", "dep:virtfw-libefi"]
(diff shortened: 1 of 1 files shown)


luigix25:
Hi @armenon-rh, this is done on purpose, because in the future we might have other transports like hyperv-vsock.

If you look in the code, vsock implementation relies on a `VsockDriver` trait, and the same thing applies to block. It's totally transport-agnostic.

I totally see your point that SVSM does not work without `virtio-drivers`, but as I mentioned, it's on purpose.

armenon-rh:
> Hi @armenon-rh, this is done on purpose, because in the future we might have other transports like hyperv-vsock.
> 
> If you look in the code, vsock implementation relies on a `VsockDriver` trait, and the same thing applies to block. It's totally transport-agnostic.
> 
> I totally see your point that SVSM does not work without `virtio-drivers`, but as I mentioned, it's on purpose.

Thanks @luigix25 for the explanation, I will close this PR. It makes total sense.

PR #1122, opened 2026-06-16
aproxy: Support Trustee KBS (CoCo) as the backend storage and verifier for attestation

armenon-rh:
Currently, attestation only works with the mock kbs server. Against the production grade Trustee KBS, attestation fails due to 
1. differences in payload format (JWS JSON vs binary),
2. hashing algorithms (SHA-384 vs SHA512) and
3. key binding structures.

This PR fixes the attestation flow against a trustee KBS (CoCo) server.
1. Use sev crate as a dependency for AttestationReport struct.
2. Create generic formats to incorporate different hash algorithms (sha512, sha384) and payload formats (binary vs JwsJson) combinations to create runtime_data.

Verified that the attestation token is successfully issued by the trustee KBS instance, see kbs container logs below

```
2026-06-16T07:48:09.055691Z  INFO actix_web::middleware::logger: 10.8.1.184 "POST /kbs/v0/attest HTTP/1.1" 200 2368 "-" "-" 0.164425
2026-06-16T07:48:09.056259Z  INFO kbs::attestation::backend: Cookie 60e3c85e8d4c471eaadd0ce24c5f6e5d request to get resource
2026-06-16T07:48:09.058087Z  INFO actix_web::middleware::logger: 10.8.1.184 "GET /kbs/v0/resource/coco/mykey/k2 HTTP/1.1" 200 554 "-" "-" 0.001929

```

### DIFF
--- a/Documentation/docs/developer/ATTESTATION.md
+++ b/Documentation/docs/developer/ATTESTATION.md
@@ -12,6 +12,7 @@
     - [Host Proxy Diagram](#host-proxy-diagram)
   - [Transport Methods](#transport-methods)
   - [Try for yourself](#try-for-yourself)
+  - [Try with Trustee KBS](#try-with-trustee-kbs)
 <!--toc:end-->
 
 ## Background
@@ -53,35 +54,21 @@ tuple [MAJOR, MINOR, PATCH]. The current version is 0.1.0.
 - `tee`: The TEE hardware architecture that SVSM is running on.
 
 The proxy will then complete the negotiation phase with the remote attestation
-server and reply with a list of negotiation parameters that must be included in
-the attestation evidence.
+server and reply with a challenge nonce that must be included in the attestation evidence.
 
 A `NegotiationResponse` is sent from the proxy to SVSM.
 An example `NegotiationResponse` is shown below.
 ```json
 {
     "challenge": "oFlY92ZdS3ymzxokYuDzxw==\",
-    "params": [
-                "EcPublicKeyBytes",
-                "Challenge"
-              ]
 }
 ```
 - `challenge`: The challenge nonce returned by the remote attestation server
 that will likely need to be hashed into the attestation evidence to ensure
 freshness. Represented as variably-sized array of bytes.
-- `params`: The negotiation parameters. Each `NegotiationParam` represents some
-form of data that must be hashed into the attestation evidence. This hash will
-be reconstructed by the remote attestation server when the evidence is presented
-from SVSM.
 
-The valid negotiation parameters are as follows:
-- `Challenge`: The bytes represented in the `challenge`.
-- `EcPublicKeyBytes`: The byte buffers of the public key's x and y coordinates
-(in that order).
-
-SVSM can then collect the attestation evidence (with the negotiation parameters
-embedded within the report data) and continue to the attestation phase.
+aproxy builds the entire KBS runtime JSON object, hashes it on the host using Sha384,
+and returns the hash as a single challenge for SVSM to put in its report.
 
 ### Attestation Phase
 
@@ -146,7 +133,7 @@ example).
 
 Valid evidence formats are the following:
 - `Snp`: SEV-SNP
-    - `report`: Attestation report bytes.
+    - `attestation_report`: Attestation report bytes.
     - `certs_buf`: Optional byte buffer representing SEV-SNP certificate chain
                    (ARK, ASK, VEK) set by the host hypervisor.
 
@@ -390,3 +377,8 @@ SEV-SNP machine with an SVSM-enabled kernel.
     cd kbs-test
     cargo run -- --measurement 9ef6c500d19addcd5937c6c8bd4e51b1893f048eea03d5407cfb0692c06615e3f6c044c667c32e520913d93234e836fe
     ```
+
+## Try with Trustee KBS
+
+You can also try with the official Trustee KBS and SVSM setup. See [KBS-SVSM](KBS-SVSM.md)
+
--- a/Documentation/docs/developer/ATTESTATION.md
+++ b/Documentation/docs/developer/ATTESTATION.md
@@ -12,6 +12,7 @@
     - [Host Proxy Diagram](#host-proxy-diagram)
   - [Transport Methods](#transport-methods)
   - [Try for yourself](#try-for-yourself)
+  - [Try with Trustee KBS](#try-with-trustee-kbs)
 <!--toc:end-->
 
 ## Background
@@ -60,25 +61,20 @@ A `NegotiationResponse` is sent from the proxy to SVSM.
 An example `NegotiationResponse` is shown below.
 ```json
 {
-    "challenge": "oFlY92ZdS3ymzxokYuDzxw==\",
-    "params": [
-                "EcPublicKeyBytes",
-                "Challenge"
-              ]
+    "challenge": "oFlY92ZdS3ymzxokYuDzxw==",
+    "hash_algo": "Sha384",
+    "payload_format": "JwsJson"
 }
 ```
 - `challenge`: The challenge nonce returned by the remote attestation server
 that will likely need to be hashed into the attestation evidence to ensure
 freshness. Represented as variably-sized array of bytes.
-- `params`: The negotiation parameters. Each `NegotiationParam` represents some
-form of data that must be hashed into the attestation evidence. This hash will
-be reconstructed by the remote attestation server when the evidence is presented
-from SVSM.
+- `hash_algo`: The cryptographic hash algorithm that SVSM must use (`Sha256`, `Sha384`, or `Sha512`).
+- `payload_format`: The structural layout format SVSM must construct before hashing (`RawBinary` or `JwsJson`).
 
-The valid negotiation parameters are as follows:
-- `Challenge`: The bytes represented in the `challenge`.
-- `EcPublicKeyBytes`: The byte buffers of the public key's x and y coordinates
-(in that order).
+The valid payload formats are as follows:
+  *   `RawBinary`: Instructs SVSM to format the public key coordinates and challenge nonce as a sequential binary layout.
+  *   `JwsJson`: Instructs SVSM to format, serialize, and hash a standard, JWS JSON map containing the public key and challenge nonce (specifically required by Trustee KBS).
 
 SVSM can then collect the attestation evidence (with the negotiation parameters
 embedded within the report data) and continue to the attestation phase.
@@ -126,7 +122,7 @@ An example `AttestationRequest` is shown below.
                        AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
                        AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
                        AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=\",
-            "certs_buf": null
+            "cert_chain": null
         }
     },
     "challenge": "oFlY92ZdS3ymzxokYuDzxw==\",
@@ -146,8 +142,8 @@ example).
 
 Valid evidence formats are the following:
 - `Snp`: SEV-SNP
-    - `report`: Attestation report bytes.
-    - `certs_buf`: Optional byte buffer representing SEV-SNP certificate chain
+    - `attestation_report`: Attestation report bytes.
+    - `cert_chain`: Optional byte buffer representing SEV-SNP certificate chain
                    (ARK, ASK, VEK) set by the host hypervisor.
 
 - `challenge`: The original challenge nonce given in the `NegotiationResponse`.
@@ -390,3 +386,7 @@ SEV-SNP machine with an SVSM-enabled kernel.
     cd kbs-test
     cargo run -- --measurement 9ef6c500d19addcd5937c6c8bd4e51b1893f048eea03d5407cfb0692c06615e3f6c044c667c32e520913d93234e836fe
     ```
+
+## Try with Trustee KBS
+
+You can also try with the official Trustee KBS and SVSM setup. See [KBS-SVSM](KBS-SVSM.md)
--- a/Documentation/docs/developer/KBS-SVSM.md
+++ b/Documentation/docs/developer/KBS-SVSM.md
@@ -1,0 +1,188 @@
+# Attestation in SVSM using Trustee KBS
+
+We need,
+1. SVSM (with attest feature),
+    > git clone https://github.com/coconut-svsm/svsm.git
+2. The aproxy on the host.
+3. Trustee KBS server (For now, the KBS softhsm container, is not upstream)
+    > git clone https://github.com/armenon-rh/trustee.git
+
+This document explains the steps to perform a complete attestation.
+
+# Prerequisites & Architectural Overview
+The Key Broker Service (KBS) handles confidential computing attestation and secret provisioning. Depending on your security requirements, it can be deployed in one of the two modes:
+
+## Mode A: Resource Key-Value Storage Backend (Default)
+Act as a direct storage repository for secrets.
+- Flow: SVSM <--> aproxy <--> KBS (Resource kvstorage backend)
+- Concept: The user provisions a secret into the KBS storage via kbs-client. SVSM later presents its attestation token to retrieve that raw secret.
+
+## Mode B: PKCS#11 Plugin Backend (softHSM)
+Offloads cryptographic operations and key management to a Software HSM.
+- Flow: SVSM <--> aproxy <--> KBS(PKCS#11 plugin) <--> softHSM
+- Concept: The user generates a 32-byte symmetric key, which softHSM wraps (encrypts) using its internal public key. The user uses the raw key to encrypt vTPM state, but implants the wrapped key into the metadata header. During boot, SVSM hands the wrapped key to KBS, which requests softHSM to unwrap it, passing back the raw key only after attestation.
+
+# Step by Step implementation
+
+## 1. Build the image
+
+An Ubuntu based KBS container image that also has softHSM and opensc (pkcs11-tool to interact with softHSM) packages installed.
+
+```bash
+cd trustee/
+podman build -t kbs-service -f kbs/docker/kbs-softhsm-pkcs11/Dockerfile .
+```
+
+## 2. Initialize Host storage directories
+
+Create the required host workspaces to persist softHSM tokens and KBS resource assets
+
+```bash
+mkdir -p $HOME/kbs-workspace/softhsm_tokens
+mkdir -p $HOME/kbs-workspace/kbs-storage
+```
+
+## 3. Configure Attestation Policy
+Create the repository directory and establish the Open Policy Agent (OPA) Rego policy file that dictates KBS access rules.
+
+```bash
+mkdir -p kbs-workspace/repository/kbs && cd kbs-workspace/repository/kbs
+
+cat <<EOF > resource-policy.rego
+package policy
+
+default allow = false
+
+allow if {
+    true
+}
+EOF
+
+cd -
+```
+
+## 4. Run the KBS container
+Choose the configuration profile that fits your desired operational mode:
+
+### Option A: Run as Resource Backend (KV Storage)
+
+```bash
+podman run -d \
+    --name kbs-pkcs11-service \
+    -p 8080:8080 \
+    -v $HOME/kbs-workspace/repository/kbs/resource-policy.rego:/etc/kbs/repository/kbs/resource-policy.rego:Z \
+    -v $HOME/kbs-workspace/kbs-storage/:/opt/confidential-containers/storage:Z \
+    kbs-service:latest /usr/local/bin/kbs --config-file /etc/kbs/kbs-config-resource.toml
+```
+
+### Option B: Run as PKCS#11 Plugin Backend
+```bash
+podman run -d \
+    --name kbs-resource-service \
+    -p 8080:8080 \
+    -v $HOME/kbs-workspace/softhsm_tokens/:/var/lib/softhsm/tokens:Z \
+    -v $HOME/kbs-workspace/repository/kbs/resource-policy.rego:/etc/kbs/repository/kbs/resource-policy.rego:Z \
+    -v $HOME/test/trustee/kbs/docker/kbs-softhsm-pkcs11/config_pkcs11.toml:/etc/kbs/kbs-config.toml \
+    kbs-service:latest
+```
+
+## 5. Generate Cryptographic assets
+```bash
+openssl rand 32 > rand.bin
+echo "My secret" > secret.txt
+```
+The secret will be stored in kbs resource backend, whereas if we want to use the pkcs11 plugin then the key will be wrapped.
+
+## 6. Provision or Wrap the cryptographic assets
+
+### Option A: Store Secret in Resource Backend
+
+```bash
+make -C kbs cli FEATURES
+./target/release/kbs-client --url http://0.0.0.0:8080 config  set-resource --path default/sample/test --resource-file secret.txt
+```
+The secret will be store in KBS.
+
+### Option B: Wrap Key via PKCS#11 Plugin
+```bash
+curl -X POST http://0.0.0.0:8080/kbs/v0/pkcs11/wrap-key \
+     -H "Content-Type: application/octet-stream" \
+     --data-binary @rand.bin \
+     --output wrapped_rand.bin
+```
+wrapped_rand.bin will contain the wrapped key
+
+## 7. Use [tpm provisioner](https://github.com/armenon-rh/tpm_provisioner) to create an NVChip file
+
+```bash
+cd ../
+git clone https://github.com/armenon-rh/tpm_provisioner.git && cd tpm_provisioner
+
+podman build -t tpm-provisioner -f Dockerfile .
+
+podman run -it --name tpm-lab tpm-provisioner
+
+cargo run && verify
+
+# Generate the ek public key from tpm, store it in ek.pub --> 1
+tpm2_createek -T mssim:host=localhost,port=2321 -c ek.ctx -u ek.pub -G rsa
+
+# print the public part of it the endorsement key  --> 2
+tpm2_print -t TPM2B_PUBLIC ek.pub
+
+# Retrieve the certificate from nvram, in DER format --> 3
+tpm2_nvread -T mssim:host=localhost,port=2321 -C o 0x1c00002 -o ek_cert.der
+
+# Extract public key from certificate, in PEM format --> 4
+openssl x509 -in ek_cert.der -inform DER -pubkey -noout > cert_key.pem
+
+# print the public key, modulus will be printed --> 5
+openssl rsa -pubin -in cert_key.pem -text -noout
+
+# modulus in 2 and 5 should match!
+
+# copy the important files to host
+# ek_cert.der, ek_cert.pem, local_ca.pem, cert_key.pem, NVChip
+podman cp tpm-lab:/app/tpm_provisioner/<file_name> /path/on/host/<file_name>
+
+```
+
+## 8. Use [vtpm-to-svsm-blk](https://github.com/armenon-rh/vtpm-to-svsm-blk) to create a virtio block image
+- use rand.bin from step 5 as key for encryption param
+- use wrapped_rand.bin from step 6 as wrapped key param
+vtpm_state.img is ready, which has payload encrypted and a header with the wrapped key.
+
+```bash
+git clone https://github.com/armenon-rh/vtpm-to-svsm-blk.git && cd vtpm-to-svsm-blk
+
+cargo build
+
+cargo run -- -k rand.bin -w wrapped_rand.bin -s /path/on/host/<NVChip> -o /path/on/host/to/store/vtpm_state.img
+
+# This will create a vTPM image, that can be attached to SVSM as a virtio block
+
+```
+
+## 9. Run Aproxy on the host
+
+```bash
+cd svsm/
+make aproxy
+./bin/aproxy --protocol kbs \
+    --url http://0.0.0.0:8080 \
+    --unix /tmp/svsm-proxy.sock \
+    --force > aproxy.log 2>&1 &
+```
+
+## 10. Launch the confidential guest
+
+Assuming that qemu is compiled using the [svsm](https://github.com/coconut-svsm/qemu.git) branch and SVSM is compiled using the latest [OVMF.FD file](https://github.com/coconut-svsm/edk2.git), and the fedora image is built using [image builder](https://github.com/stefano-garzarella/snp-svsm-vtpm/blob/main/build-vm-image.sh)
+
+```bash
+./scripts/launch_guest.sh --aproxy /tmp/svsm-proxy.sock \
+    --qemu $HOME/test/qemu/bin/qemu-system-x86_64 \
+    --image $HOME/test/snp-svsm-vtpm/images/fedora-luks.qcow2 \
+    --state ../vtpm_state.img
+```
+
+Note: Make sure to add the option --pcd PcdUninstallMemAttrProtocol=TRUE while building edk2 for Fedora 42 images. See [instructions](../installation/INSTALL.md#building-the-guest-firmware)
--- a/Documentation/docs/developer/KBS-SVSM.md
+++ b/Documentation/docs/developer/KBS-SVSM.md
@@ -1,0 +1,198 @@
+# Attestation in SVSM using Trustee KBS
+
+We need,
+1. SVSM (with attest feature),
+    > git clone https://github.com/coconut-svsm/svsm.git
+2. The aproxy on the host.
+3. Trustee KBS server (For now, the KBS softhsm container, is not upstream)
+    > git clone https://github.com/armenon-rh/trustee.git
+
+This document explains the steps to perform a complete attestation.
+
+# Prerequisites & Architectural Overview
+The Key Broker Service (KBS) handles confidential computing attestation and secret provisioning. Depending on your security requirements, it can be deployed in one of the two modes:
+
+## Mode A: Resource Key-Value Storage Backend (Default)
+Act as a direct storage repository for secrets.
+- Flow: SVSM <--> aproxy <--> KBS (Resource kvstorage backend)
+- Concept: The user provisions a secret into the KBS storage via kbs-client. SVSM later presents its attestation token to retrieve that raw secret.
+
+## Mode B: PKCS#11 Plugin Backend (softHSM)
+Offloads cryptographic operations and key management to a Software HSM.
+- Flow: SVSM <--> aproxy <--> KBS(PKCS#11 plugin) <--> softHSM
+- Concept: The user generates a 32-byte symmetric key, which softHSM wraps (encrypts) using its internal public key. The user uses the raw key to encrypt vTPM state, but implants the wrapped key into the metadata header. During boot, SVSM hands the wrapped key to KBS, which requests softHSM to unwrap it, passing back the raw key only after attestation.
+
+# Step by Step implementation
+
+## 1. Build the image
+
+An Ubuntu based KBS container image that also has softHSM and opensc (pkcs11-tool to interact with softHSM) packages installed.
+
+```bash
+cd trustee/
+podman build -t kbs-service -f kbs/docker/kbs-softhsm-pkcs11/Dockerfile .
+```
+
+## 2. Initialize Host storage directories
+
+Create the required host workspaces to persist softHSM tokens and KBS resource assets
+
+```bash
+mkdir -p $HOME/kbs-workspace/softhsm_tokens
+mkdir -p $HOME/kbs-workspace/kbs-storage
+```
+
+## 3. Configure Attestation Policy
+Create the repository directory and establish the Open Policy Agent (OPA) Rego policy file that dictates KBS access rules.
+
+```bash
+mkdir -p kbs-workspace/repository/kbs && cd kbs-workspace/repository/kbs
+
+cat <<EOF > resource-policy.rego
+package policy
+
+default allow = false
+
+allow if {
+    true
+}
+EOF
+
+cd -
+```
+
+## 4. Run the KBS container
+Choose the configuration profile that fits your desired operational mode:
+
+### Option A: Run as Resource Backend (KV Storage)
+
+```bash
+
+podman run -d \
+    --name kbs-resource-service \
+    -p 8080:8080 \
+    -v $HOME/kbs-workspace/repository/kbs/resource-policy.rego:/etc/kbs/repository/kbs/resource-policy.rego:Z \
+    -v $HOME/kbs-workspace/kbs-storage/:/opt/confidential-containers/storge:Z \
+    -v $HOME/test/trustee/kbs/docker/kbs-softhsm-pkcs11/config_resource.toml:/etc/kbs/kbs-config-resource.toml \ kbs-service:latest  /usr/local/bin/kbs --config-file /etc/kbs/kbs-config-resource.toml
+
+```
+
+### Option B: Run as PKCS#11 Plugin Backend
+```bash
+podman run -d \
+    --name kbs-resource-pkcs11 \
+    -p 8080:8080 \
+    -v $HOME/kbs-workspace/softhsm_tokens/:/var/lib/softhsm/tokens:Z \
+    -v $HOME/kbs-workspace/repository/kbs/resource-policy.rego:/etc/kbs/repository/kbs/resource-policy.rego:Z \
+    -v $HOME/kbs-workspace/kbs-storage/:/opt/confidential-containers/storge:Z \
+    -v $HOME/test/trustee/kbs/docker/kbs-softhsm-pkcs11/config_pkcs11.toml:/etc/kbs/kbs-config.toml \
+    kbs-service:latest
+```
+
+## 5. Generate Cryptographic assets
+```bash
+openssl rand 32 > rand.bin
+echo "My secret" > secret.txt
+```
+The secret will be stored in kbs resource backend, whereas if we want to use the pkcs11 plugin then the key will be wrapped.
+
+## 6. Provision or Wrap the cryptographic assets
+
+### Option A: Store Secret in Resource Backend
+
+```bash
+make -C kbs cli FEATURES
+./target/release/kbs-client --url http://0.0.0.0:8080 config  set-resource --path default/sample/test --resource-file secret.txt
+```
+The secret will be store in KBS.
+
+### Option B: Wrap Key via PKCS#11 Plugin
+```bash
+curl -X POST http://0.0.0.0:8080/kbs/v0/pkcs11/wrap-key \
+     -H "Content-Type: application/octet-stream" \
+     --data-binary @rand.bin \
+     --output wrapped_rand.bin
+```
+wrapped_rand.bin will contain the wrapped key
+
+## 7. Use [tpm provisioner](https://github.com/armenon-rh/tpm_provisioner) to create an NVChip file
+
+```bash
+cd ../
+git clone https://github.com/armenon-rh/tpm_provisioner.git && cd tpm_provisioner
+
+podman build -t tpm-provisioner -f Dockerfile .
+
+podman run -it --name tpm-lab tpm-provisioner
+
+# Refer the readme in tpm_provisioner project
+tpm_provisioner provision && verify;
+
+# Generate the ek public key from tpm, store it in ek.pub --> 1
+tpm2_createek -T mssim:host=localhost,port=2321 -c ek.ctx -u ek.pub -G rsa
+
+# print the public part of it the endorsement key  --> 2
+tpm2_print -t TPM2B_PUBLIC ek.pub
+
+# Retrieve the certificate from nvram, in DER format --> 3
+tpm2_nvread -T mssim:host=localhost,port=2321 -C o 0x1c00002 -o ek_cert.der
+
+# Extract public key from certificate, in PEM format --> 4
+openssl x509 -in ek_cert.der -inform DER -pubkey -noout > cert_key.pem
+
+# print the public key, modulus will be printed --> 5
+openssl rsa -pubin -in cert_key.pem -text -noout
+
+# modulus in 2 and 5 should match!
+
+# copy the important files to host
+# ek_cert.der, ek_cert.pem, local_ca.pem, cert_key.pem, NVChip
+podman cp tpm-lab:/app/tpm_provisioner/<file_name> /path/on/host/<file_name>
+
+```
+
+## 8. Use [vtpm-to-svsm-blk](https://github.com/armenon-rh/vtpm-to-svsm-blk) to create a virtio block image
+- use rand.bin from step 5 as key for encryption param
+- use wrapped_rand.bin from step 6 as wrapped key param
+vtpm_state.img is ready, which has payload encrypted and a header with the wrapped key.
+
+```bash
+git clone https://github.com/armenon-rh/vtpm-to-svsm-blk.git && cd vtpm-to-svsm-blk
+
+cargo build
+
+cargo run -- -k rand.bin -w wrapped_rand.bin -s /path/on/host/<NVChip> -o /path/on/host/to/store/vtpm_state.img
+
+# This will create a vTPM image, that can be attached to SVSM as a virtio block
+
+```
+
+## 9. Run Aproxy on the host
+
+```bash
+cd svsm/
+make aproxy
+
+# Based on what protocol you want to use (kbs or kbs-trustee)
+
+./bin/aproxy --protocol kbs \
+    --url http://127.0.0.1:8080 \
+    --vsock > aproxy.log 2>&1 &
+
+```
+
+## 10. Launch the confidential guest
+
+Assuming that qemu is compiled using the [svsm](https://github.com/coconut-svsm/qemu.git) branch and SVSM is compiled using the latest [OVMF.FD file](https://github.com/coconut-svsm/edk2.git), and the fedora image is built using [image builder](https://github.com/stefano-garzarella/snp-svsm-vtpm/blob/main/build-vm-image.sh)
+
+```bash
+
+./scripts/launch_guest.sh --vsock 3 \
+    --qemu $HOME/test/qemu/build/qemu-system-x86_64 \
+    --image $HOME/test/snp-svsm-vtpm/images/fedora-luks.qcow2
+    # provide VTPM state file if necessary
+    # --state ../vtpm_state.img
+
+```
+
+Note: Make sure to add the option --pcd PcdUninstallMemAttrProtocol=TRUE while building edk2 for Fedora 42 images. See [instructions](../installation/INSTALL.md#building-the-guest-firmware)
--- a/kernel/src/attest.rs
+++ b/kernel/src/attest.rs
@@ -33,12 +33,13 @@ use cocoon_tpm_utils_common::{
 use kbs_types::Tee;
 use libaproxy::*;
 use serde::Serialize;
-use sha2::{Digest, Sha512};
 use zerocopy::{FromBytes, IntoBytes};
 
 #[cfg(feature = "attest-serial")]
 // TODO: Make the IO port configurable/discoverable or drop the support entirely.
 const ATTEST_DEFAULT_SERIAL_IO_ADDR: u16 = 0x3e8; // COM3
+const CHALLENGE_LEN: usize = 48;
+const TEE_REPORT_DATA_LEN: usize = 64;
 
 enum Transport {
     Vsock(VsockStream),
@@ -174,7 +175,7 @@ impl AttestationDriver {
     fn attestation(&mut self, n: NegotiationResponse) -> Result<SecretSlice, AttestationError> {
         let pub_key = self.get_tpm_pub_key()?;
 
-        let evidence = evidence(&self.tee, hash(&n, &pub_key)?)?;
+        let evidence = evidence(&self.tee, prepare_report_data(&n)?)?;
 
         let req = AttestationRequest {
             tee: self.tee,
@@ -306,6 +307,8 @@ pub enum AttestationError {
     Crypto(CryptoError),
     /// Guest has failed attestation.
     Failed,
+    /// Negotiation Response challenge length from KBS is invalid
+    InvalidChallengeLength,
     // Unable to derive wrap key.
     KeyDerivation(concat_kdf::Error),
     /// Error deserializing the negotiation response from JSON bytes.
@@ -408,26 +411,14 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
     Ok(evidence)
 }
 
-/// Hash the negotiation parameters from the attestation server for inclusion in the
-/// attestation evidence.
-fn hash(
-    n: &NegotiationResponse,
-    pub_key: &TpmsEccPoint<'static>,
-) -> Result<Vec<u8>, AttestationError> {
-    let mut sha = Sha512::new();
-
-    for p in &n.params {
-        match p {
-            NegotiationParam::Challenge => {
-                sha.update(&n.challenge);
-            }
-            #[allow(irrefutable_let_patterns)]
-            NegotiationParam::EcPublicKeyBytes => {
-                sha.update(&*pub_key.x.buffer);
-                sha.update(&*pub_key.y.buffer);
-            }
-        }
+/// Take 48 byte negotiation challenge nonce from aproxy into 64 byte array required for the TEE attestation evidence report
+fn prepare_report_data(n: &NegotiationResponse) -> Result<Vec<u8>, AttestationError> {
+    if n.challenge.len() != CHALLENGE_LEN {
+        return Err(AttestationError::InvalidChallengeLength);
     }
 
-    try_to_vec(&sha.finalize()).or(Err(AttestationError::VecAlloc))
+    let mut report_data = [0u8; TEE_REPORT_DATA_LEN];
+    report_data[..CHALLENGE_LEN].copy_from_slice(&n.challenge);
+
+    Ok(report_data.to_vec())
 }
--- a/kernel/src/attest.rs
+++ b/kernel/src/attest.rs
@@ -130,6 +130,19 @@ impl TryFrom<Tee> for AttestationDriver {
 }
 
 impl AttestationDriver {
+    /// Extracts the public key coordinates as a TPM ECC point from the internal `EccKey`.
+    fn get_tpm_pub_key(&self) -> Result<TpmsEccPoint<'static>, AttestationError> {
+        let curve =
+            Curve::new(self.ecc.pub_key().get_curve_id()).map_err(AttestationError::Crypto)?;
+
+        let curve_ops = curve.curve_ops().map_err(AttestationError::Crypto)?;
+
+        self.ecc
+            .pub_key()
+            .to_tpms_ecc_point(&curve_ops)
+            .map_err(AttestationError::Crypto)
+    }
+
     /// Attest SVSM's launch state by communicating with the attestation proxy.
     pub fn attest(&mut self) -> Result<SecretSlice, SvsmError> {
         let negotiation = self.negotiation()?;
@@ -141,9 +154,12 @@ impl AttestationDriver {
     /// that should be included in attestation evidence (e.g. through SEV-SNP's REPORT_DATA
     /// mechanism).
     fn negotiation(&mut self) -> Result<NegotiationResponse, AttestationError> {
+        let pub_key = self.get_tpm_pub_key()?;
+
         let request = NegotiationRequest {
             version: (0, 1, 0), // Only version supported at present.
             tee: self.tee,
+            key: (self.ecc.pub_key().get_curve_id(), &pub_key).into(),
         };
 
         self.write(request)?;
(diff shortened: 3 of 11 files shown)


kernel/src/attest.rs
```diff
@@ -410,22 +409,10 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
 /// attestation evidence.
 fn hash(
     n: &NegotiationResponse,
-    pub_key: &TpmsEccPoint<'static>,
+    _pub_key: &TpmsEccPoint<'static>,
```
luigix25: can we drop this?

tools/aproxy/src/backend/kbs.rs
```diff
@@ -60,21 +64,34 @@ impl AttestationProtocol for KbsProtocol {
...
+        key_map.insert("crv".to_string(), json!("P-521"));
+        key_map.insert("kty".to_string(), json!("EC"));
+        key_map.insert("x".to_string(), json!(BASE64_URL_SAFE_NO_PAD.encode(&request.key.x)));
+        key_map.insert("y".to_string(), json!(BASE64_URL_SAFE_NO_PAD.encode(&request.key.y)));
+
+        runtime_data.insert("tee-pubkey".to_string(), json!(key_map));
+
+
+        let json_bytes = serde_json::to_vec(&runtime_data)?;
+        println!("DEBUG: Hashing this JSON: {}", String::from_utf8_lossy(&json_bytes));
```
luigix25: probably leftover from development

tools/aproxy/src/backend/kbs.rs
```diff
@@ -102,6 +119,9 @@ impl AttestationProtocol for KbsProtocol {
             },
         };
 
+        let json_payload = serde_json::to_string_pretty(&attestation)?;
+        println!("DEBUG: Sending Attestation Request to KBS:\n{}", json_payload);
+
```
luigix25: ditto

Cargo.toml
```diff
@@ -66,6 +66,7 @@ userlib = { path = "user/lib" }
 userinit = { path = "user/init" }
 release = { path = "release" }
 virtio-drivers = { git = "https://github.com/coconut-svsm/virtio-drivers.git", rev = "dcbdbc8cefc0c985322382acdeb633a63a3a4cc1" }
+serde_bytes = { version = "0.11", default-features = false }
```
luigix25: who is using this dependency?

tools/aproxy/src/backend/kbs.rs
```diff
@@ -218,33 +238,28 @@ fn unwrap_epk(resp: &Response) -> anyhow::Result<EcP256PublicKey> {
...
 enum KbsEvidence {
     Snp {
-        #[serde(rename = "snp-report")]
-        snp_report: String,
-        #[serde(rename = "certs-buf")]
-        certs_buf: Option<String>,
+        #[serde(rename = "attestation_report")]
+        snp_report: sev::firmware::guest::AttestationReport,
+        #[serde(rename = "cert_chain")]
+        certs_buf: Option<Vec<u8>>,
```
luigix25, line 243: is this parameter used? I see it is set to None

kernel/src/attest.rs
```diff
@@ -399,22 +409,10 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
...
 ) -> Result<Vec<u8>, AttestationError> {
-    let mut sha = Sha512::new();
-
-    for p in &n.params {
-        match p {
-            NegotiationParam::Challenge => {
-                sha.update(&n.challenge);
-            }
-            #[allow(irrefutable_let_patterns)]
-            NegotiationParam::EcPublicKeyBytes => {
```
luigix25, line 416: will `EcPublicKeyBytes` be used anywhere else?

luigix25 (changes requested):
I did a quick review.

Don't forget to run `make clippy` `cargo fmt` and `make test` before submitting :)

kernel/src/attest.rs
armenon-rh: Thanks for the review @luigix25 
yes we can drop this. Hashing is no longer done here, so maybe the function itself should be renamed.
I will do it.

tools/aproxy/src/backend/kbs.rs
armenon-rh: yes, this should be removed. Thanks

tools/aproxy/src/backend/kbs.rs
armenon-rh: yes.

Cargo.toml
armenon-rh: I see. None of the crates is using this. Previously AttestationReport struct was copied into svsm. After including sev library from virtee, I forgot to remove this. 
I shall remove it, thanks.

tools/aproxy/src/backend/kbs.rs
armenon-rh: Our implementation checks against the KBS verifier [SnpEvidence](https://github.com/confidential-containers/trustee/blob/8b2110ab5875fb7c3e98edae523706dcd57e8ce8/deps/verifier/src/snp/mod.rs#L38)

I recall I was getting [this error](https://github.com/confidential-containers/trustee/blob/8b2110ab5875fb7c3e98edae523706dcd57e8ce8/deps/verifier/src/snp/mod.rs#L351) when these fields were not in place. 

Indeed its set to None and sent by the aproxy, so maybe the KBS verifier [fetches](https://github.com/confidential-containers/trustee/blob/8b2110ab5875fb7c3e98edae523706dcd57e8ce8/deps/verifier/src/snp/mod.rs#L437-L463) the certificate automatically. I am not entirely sure.

kernel/src/attest.rs
armenon-rh: yes it is a variant of NegotiationParam, which no longer is used. aproxy only returns a challenge nonce. 
I shall remove it. Thanks.

luigix25:
@armenon-rh please resolve all the conversation that no longer applies. I run `make clippy` locally and it's failing:

```
error[E0152]: found duplicate lang item `panic_impl`
   --> boot/bldr/src/main.rs:403:1
    |
403 | / fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
404 | |     // SAFETY: raising an undefined instuction exception is always safe.
405 | |     unsafe { core::arch::asm!("ud2") }
406 | |     unreachable!("");
407 | | }
    | |_^
    |
    = note: the lang item is first defined in crate `std` (which `uuid` depends on)
    = note: first definition in `std` loaded from /home/lleonard/.rustup/toolchains/1.88.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-8df6be531efb3fd0.rlib
    = note: second definition in the local crate (`bldr`)
```

not sure why, i didn't investigate, but reverting your changes it passes

Documentation/docs/developer/KBS-SVSM.md
```diff
@@ -0,0 +1,201 @@
...
+export PYTHON_COMMAND=python3
+make -j16 -C BaseTools/
+source ./edksetup.sh --reconfig
+build -p OvmfPkg/OvmfPkgX64.dsc -a X64 \
+      -b DEBUG -t GCC \
+      -D DEBUG_ON_SERIAL_PORT \
+      -D DEBUG_VERBOSE \
+      -D TPM2_ENABLE
+      --pcd PcdUninstallMemAttrProtocol=TRUE
+```
```
luigix25: we already have a section where we talk about building edk2 in the file `installation/INSTALL.md`.
Can we just link it here somehow? If building instruction change, we'd need to two update two things

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -53,35 +54,22 @@ tuple [MAJOR, MINOR, PATCH]. The current version is 0.1.0.
...
 A `NegotiationResponse` is sent from the proxy to SVSM.
 An example `NegotiationResponse` is shown below.
 ```json
 {
     "challenge": "oFlY92ZdS3ymzxokYuDzxw==\",
-    "params": [
-                "EcPublicKeyBytes",
-                "Challenge"
-              ]
+
```
luigix25: nit: i think this blank like can be dropped

kernel/src/attest.rs
```diff
@@ -397,26 +405,9 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
...
-                sha.update(&*pub_key.y.buffer);
-            }
-        }
-    }
-
-    try_to_vec(&sha.finalize()).or(Err(AttestationError::VecAlloc))
+/// Take 48 byte negotiation challenge nonce from aproxy into 64 byte array required for the TEE attestation evidence report
+fn prepare_report_data(n: &NegotiationResponse) -> Result<Vec<u8>, AttestationError> {
+    let mut report_data = [0u8; 64];
+    report_data[..48].copy_from_slice(&n.challenge);
```
luigix25: should we check for n.challenge size?

tools/aproxy/src/backend/kbs.rs
```diff
@@ -92,7 +108,7 @@ impl AttestationProtocol for KbsProtocol {
         let attestation = Attestation {
             init_data: None,
             runtime_data: RuntimeData {
-                nonce: BASE64_STANDARD.encode(request.challenge),
+                nonce: self.original_nonce.take().unwrap_or_default(),
```
luigix25: should this be treated as an error? IIUC nonce is mandatory, right? I'm not an expert

armenon-rh:
> @armenon-rh please resolve all the conversation that no longer applies. I run `make clippy` locally and it's failing:
> 
> ```
> error[E0152]: found duplicate lang item `panic_impl`
>    --> boot/bldr/src/main.rs:403:1
>     |
> 403 | / fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
> 404 | |     // SAFETY: raising an undefined instuction exception is always safe.
> 405 | |     unsafe { core::arch::asm!("ud2") }
> 406 | |     unreachable!("");
> 407 | | }
>     | |_^
>     |
>     = note: the lang item is first defined in crate `std` (which `uuid` depends on)
>     = note: first definition in `std` loaded from /home/lleonard/.rustup/toolchains/1.88.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-8df6be531efb3fd0.rlib
>     = note: second definition in the local crate (`bldr`)
> ```
> 
> not sure why, i didn't investigate, but reverting your changes it passes

Hi Luigi,
I have mentioned in the commit message that make clippy does not work, because sev crate we added in the commit, pulls uuid which is std dependent. So we can run make clippy CARGO_HACK=1
I do not know if this is a good idea.

tools/aproxy/src/backend/kbs.rs
armenon-rh: removed this line in v2

kernel/src/attest.rs
armenon-rh: removed the extra param and changed the function name in v2.

kernel/src/attest.rs
armenon-rh: remove NegotiationParam in v2 as it is no longer used.

Documentation/docs/developer/KBS-SVSM.md
armenon-rh: yes I will do it.

tools/aproxy/src/backend/kbs.rs
armenon-rh: yes, it should be treated as an error. I will change it in v3

kernel/src/attest.rs
armenon-rh: yes. I shall add the check in v3.

Documentation/docs/developer/ATTESTATION.md
armenon-rh: yes

tools/aproxy/src/backend/kbs.rs
armenon-rh: We need this parameter to be sent.

Documentation/docs/developer/KBS-SVSM.md
armenon-rh: linked in v3.

luigix25 (commented):
I see a lot of changes in `Cargo.lock` file, I don't think this is expected..

other than that, I'm not an attestation expert, I don't see anything wrong with the code!

tools/aproxy/src/backend/kbs.rs
luigix25, line 35: is this doc up to date ? IIUC svsm is not doing hashing anymore because it's done here, right?

Documentation/docs/developer/KBS-SVSM.md
```diff
@@ -0,0 +1,188 @@
+# Attestation in SVSM using Trustee KBS
+
+We need,
+1. SVSM (with attest feature),
+    > git clone https://github.com/coconut-svsm/svsm.git
+2. The aproxy on the host.
+3. Trustee KBS server (For now, the KBS softhsm container, is not upstream)
+    > git clone https://github.com/armenon-rh/trustee.git
```
luigix25, line 8: this should be `kbs-softhsm` branch, right? is there an ongoing work to upstream it?

Documentation/docs/developer/ATTESTATION.md
luigix25, line 47: here we are missing the new `key` field

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -146,7 +133,7 @@ example).
 
 Valid evidence formats are the following:
 - `Snp`: SEV-SNP
-    - `report`: Attestation report bytes.
+    - `attestation_report`: Attestation report bytes.
     - `certs_buf`: Optional byte buffer representing SEV-SNP certificate chain
```
luigix25, line 137: `certs_buf` was renamed to `cert_chain`

Documentation/docs/developer/KBS-SVSM.md
```diff
@@ -0,0 +1,188 @@
...
+
+```bash
+cd ../
+git clone https://github.com/armenon-rh/tpm_provisioner.git && cd tpm_provisioner
+
+podman build -t tpm-provisioner -f Dockerfile .
+
+podman run -it --name tpm-lab tpm-provisioner
+
+cargo run && verify
```
luigix25, line 125: does this need to be run in the container or outside? I suspect outside

libaproxy/src/negotiation.rs
stefano-garzarella, line 15: Should we update this documentation mentioning why `key` is needed ?

kernel/src/attest.rs
```diff
@@ -408,26 +411,14 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
...
-            NegotiationParam::Challenge => {
-                sha.update(&n.challenge);
-            }
-            #[allow(irrefutable_let_patterns)]
-            NegotiationParam::EcPublicKeyBytes => {
-                sha.update(&*pub_key.x.buffer);
-                sha.update(&*pub_key.y.buffer);
-            }
-        }
+/// Take 48 byte negotiation challenge nonce from aproxy into 64 byte array required for the TEE attestation evidence report
```
stefano-garzarella, line 414: Are we sure this is a good idea?

IIUC we are trusting in some way the host to do things right, but what prevents the host to put in the hash another public key that the host control (e.g. it knows the private key)?

CC @tylerfanelli IIRC this was a design choice.

kernel/src/attest.rs
```diff
@@ -33,12 +33,13 @@ use cocoon_tpm_utils_common::{
...
 use libaproxy::*;
 use serde::Serialize;
-use sha2::{Digest, Sha512};
 use zerocopy::{FromBytes, IntoBytes};
 
 #[cfg(feature = "attest-serial")]
 // TODO: Make the IO port configurable/discoverable or drop the support entirely.
 const ATTEST_DEFAULT_SERIAL_IO_ADDR: u16 = 0x3e8; // COM3
+const CHALLENGE_LEN: usize = 48;
+const TEE_REPORT_DATA_LEN: usize = 64;
```
stefano-garzarella, line 42: Is this Trustee KBS specific?

tools/aproxy/src/backend/kbs.rs
```diff
@@ -60,21 +61,36 @@ impl AttestationProtocol for KbsProtocol {
...
+            json!(BASE64_URL_SAFE_NO_PAD.encode(&request.key.y)),
+        );
+
+        runtime_data.insert("tee-pubkey".to_string(), json!(key_map));
 
-        Ok(resp)
+        let json_bytes = serde_json::to_vec(&runtime_data)?;
+
+        let mut hasher = Sha384::new();
+        hasher.update(&json_bytes);
```
stefano-garzarella, line 88: Does it means that the order in the json matter ?
Should we add tests and/or tests and/or asserts and/or links to KBS docs about the order we need to follow?

stefano-garzarella: But as I said, IMO the hash must be done in SVSM and the protocol with aproxy should define the order, etc.

stefano-garzarella (commented):
@armenon-rh thanks for this, but I think there is a critical issue about moving the hashing in the proxy.

That said, maybe commit 2 should be split in multiple commits if possible.

Documentation/docs/developer/ATTESTATION.md
armenon-rh: Oh yes. I will update it.

Documentation/docs/developer/ATTESTATION.md
armenon-rh: yeah, we need to add "key" of type EcP256PublicKey. Thanks.

Documentation/docs/developer/KBS-SVSM.md
armenon-rh: yes, the POC was done, with changes in my public fork of trustee. I will be sending a PR to the confidential-containers/trustee project.

Documentation/docs/developer/KBS-SVSM.md
armenon-rh: actually this has been changed in the 
[tpm_provisioner project](https://github.com/armenon-rh/tpm_provisioner/pull/5/changes#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2R9)

We get a shell of the container with the previous command
So running
`tpm_provisioner provision && verify`
directly will give the desired vtpm state file with the endorsement key.
Yes, I need to update the README here, and also in the tpm_provisioner project. Maybe I will link to the readme in the tpm_provisioner project here. Thanks.

kernel/src/attest.rs
armenon-rh: yes.
The [verify](https://github.com/confidential-containers/trustee/blob/c3165a7ceecad6b2e6fbd8b706de3ee16bc6618d/kbs/src/attestation/coco/builtin.rs#L79) function in coco/builtin.rs uses Sha384 algorithm explicitly for all the tee types except IBM Z SE. The output is therefore 384 bits, or 48 bytes nonce/challenge from trustee back to svsm. I guess its similar hardcoded mapping in other coco integrations as well.

tools/aproxy/src/backend/kbs.rs
armenon-rh: yeah, I will update the doc. Hashing is done here, not in svsm.

kernel/src/attest.rs
armenon-rh: I understand.
SVSM blindly trusts the challenge that the aproxy sends it, to sign. So the host can practically use the (nonce + host_key) and create runtime_data and send it to SVSM.

We thought of this because we did not want to add json serialization in SVSM.

tools/aproxy/src/backend/kbs.rs
armenon-rh: yeah, the order matters. AFAIK, its alphabetical. Maybe we should use something like [serialize_canon_json()](https://github.com/confidential-containers/trustee/blob/c3165a7ceecad6b2e6fbd8b706de3ee16bc6618d/attestation-service/src/lib.rs#L32)

libaproxy/src/negotiation.rs
armenon-rh: The key is sent **in negotiation phase** so that aproxy backend can use it to pre calculate the sha384 of the json block before SVSM can sign it.

But there is a flaw in this design, like you mentioned, the host can tamper this report_data sent to SVSM. So we might as well drop sending the public key the negotiation phase.

SVSM can create a simple binary buffer independently when it receives the nonce from kbs
(nonce + svsm public key)  and compute the hash.
Put the hash in report_data and request hardware signature => signed report

SVSM sends attestation request to kbs via aproxy. This request has signed report and its public key.

aproxy remains the same. It will take the public key and nonce, format them into json/base64url object expected by kbs and transmits to KBS. But this time tee_evidence will have the signed report from svsm

The trustee kbs will require some changes then,
[parse_runtime_data()](https://github.com/confidential-containers/trustee/blob/c3165a7ceecad6b2e6fbd8b706de3ee16bc6618d/attestation-service/src/lib.rs#L321) function , instead of serializing and hashing the json representation, 
should parse the JSON fields, compute hash and compare it with the report_data hash.

I will try this approach. Thanks a lot.

kernel/src/attest.rs
tylerfanelli: Since the challenge is only one part of the report_data, how is that an issue? It's the server's responsibility to have access to the challenge and hash it correctly.

kernel/src/attest.rs
tylerfanelli: If it's Trustee KBS specific, it should be within the Trustee KBS's attestation module.

kernel/src/attest.rs
stefano-garzarella: @tylerfanelli Maybe I didn't get it completely, but here it seems the hash is done by the host. So it's not clear to me what prevent the host to put its public key in the hash...

libaproxy/src/negotiation.rs
stefano-garzarella: Yeah this makes sense to me. Also
Keep in mind that SVSM code should be as generic as possible and not tied to trustee KBS

stefano-garzarella: @armenon-rh IIUC in json array order is strictly preserved, this was the reason why using an array of `NegotiationParam` in the `NegotiationResponse`, so I guess what we have now can be used to achieve exactly what we need, no?

I think the proxy should instruct SVSM on how to build the hash putting the right order of `NegotiationParam` in `NegotiationResponse.params`.

armenon-rh: @stefano-garzarella yes, I am trying to do exactly that in the next revision.
I am adding a new NegotiationParam that will handle JWS formatted JSON. Unlike the already existing params  that handled binary data, this will handle json data and therefore will require sorting internally. To do that I am using a BTreeMap.

For example, 
The runtime_data that traditionally we sent contained binary data (challenge and pubkey)

But the Trustee KBS expects runtime_data as a JWS-compliant JSON payload
It contains:
 - "additional-evidence": An empty string placeholder ("").
 - "nonce": The challenge nonce received from the AS, encoded using standard Base64 (base64_encode).
 - "tee-pubkey": A nested JSON map representing SVSM's public key.

tee-pubkey was inturn made of
   - "alg": "ECDH-ES+A256KW".
   - "crv": "P-521".
   - "kty": "EC"
   - "x": The public key's x-coordinate, encoded using Base64url with no padding (base64url_encode).
   - "y": The public key's y-coordinate, encoded using Base64url with no padding (base64url_encode).

So we construct the above in SVSM, make sure its sorted using a BTreeMap and then hash it using sha384.
This was done in the aproxy till now (accidentally in the correct lexicographical order) . In the future revision I am shifting this logic to SVSM. Ofcourse, aproxy instructs SVSM to send such data via setting the NegotiationParam.

Thanks so much for the review.

tylerfanelli: +1. I was going to suggest perhaps sending the negotiation hash back to SVSM so it can reconstruct it and ensure the correct parameters, but I think that is more trouble than it is worth and the slight change to Trustee would be warranted.

stefano-garzarella: @armenon-rh seems reasonable to me, I'm still not sure why we need BTreeMap, since IIUC json arrays order is strictly preserved, so the proxy can put the right order of `NegotiationParam`. Maybe we need to extend them or `NegotiationResponse` to specify that everything should be in JSON format before the hash. Anyway, I'm sure I'm missing something, so if you think it's needed, go head. I'll look at the code to understand better.

Documentation/docs/developer/ATTESTATION.md
armenon-rh: We are no longer sending the key in negotiation request. That was flawed design. So closing this conversation.

Documentation/docs/developer/KBS-SVSM.md
armenon-rh: I have made changes in the original tpm_provisioner project.
Made a reference to it here.

tools/aproxy/src/backend/kbs.rs
armenon-rh: hashing is done in SVSM now. Order is preserved using BTreeMap.

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -60,25 +61,20 @@ A `NegotiationResponse` is sent from the proxy to SVSM.
...
-              ]
+    "challenge": "oFlY92ZdS3ymzxokYuDzxw==",
+    "hash_algo": "Sha384",
+    "payload_format": "JwsJson"
 }
 ```
 - `challenge`: The challenge nonce returned by the remote attestation server
 that will likely need to be hashed into the attestation evidence to ensure
 freshness. Represented as variably-sized array of bytes.
-- `params`: The negotiation parameters. Each `NegotiationParam` represents some
```
stefano-garzarella, line 73: I'm really confused why we are removing this.

E.g. for `RawBinary` how the order of params is specified?

Documentation/docs/developer/ATTESTATION.md
```diff
@@ -146,8 +142,8 @@ example).
 
 Valid evidence formats are the following:
 - `Snp`: SEV-SNP
-    - `report`: Attestation report bytes.
-    - `certs_buf`: Optional byte buffer representing SEV-SNP certificate chain
+    - `attestation_report`: Attestation report bytes.
+    - `cert_chain`: Optional byte buffer representing SEV-SNP certificate chain
```
stefano-garzarella, line 146: Can we avoid this renames in the svsm-proxy protocol and renames them in the `aproxy` ?

kernel/src/attest.rs
```diff
@@ -375,26 +380,110 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
...
+impl PayloadFormatter for RawBinaryFormatter {
+    fn format(
+        &self,
+        challenge: &[u8],
+        pub_key: &TpmsEccPoint<'_>,
+    ) -> Result<Vec<u8>, AttestationError> {
+        let mut buffer = Vec::new();
+        buffer.extend_from_slice(&pub_key.x.buffer);
+        buffer.extend_from_slice(&pub_key.y.buffer);
+        buffer.extend_from_slice(challenge);
```
stefano-garzarella, line 430: Exactly this, why this order is fixed? What happen if a new attestation protocol uses a different order or needs something else?

kernel/src/attest.rs
```diff
@@ -375,26 +380,110 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
...
+        let x_encoded = base64url_encode(&pub_key.x.buffer);
+        let y_encoded = base64url_encode(&pub_key.y.buffer);
+        let nonce_encoded = base64_encode(challenge);
+
+        let mut key_map = BTreeMap::new();
+        key_map.insert("alg".to_string(), json!("ECDH-ES+A256KW"));
+        key_map.insert("crv".to_string(), json!("P-521"));
+        key_map.insert("kty".to_string(), json!("EC"));
+        key_map.insert("x".to_string(), json!(x_encoded));
+        key_map.insert("y".to_string(), json!(y_encoded));
```
stefano-garzarella, line 452: What happen if another protocol uses different names? (e.g. "algorithm" instead of "alg")

stefano-garzarella (commented):
I'm concerned about the protocol between SVSM and aproxy. @tylerfanelli tried to make it generic, but we're adding a lot of non-generic changes here (fixed parameter order for the Raw type, fixed JSON field names for the JSON format). At this point, the protocol is no longer generic IIUC, so is it still worth using this intermediate protocol, or should we remove it and incorporate the attestation protocol directly into SVSM and implement a simple https proxy in the host?

tools/aproxy/src/backend/kbs.rs
armenon-rh: Shifted hashing back to SVSM.

Documentation/docs/developer/ATTESTATION.md
armenon-rh: yes

armenon-rh:
> I'm concerned about the protocol between SVSM and aproxy. @tylerfanelli tried to make it generic, but we're adding a lot of non-generic changes here (fixed parameter order for the Raw type, fixed JSON field names for the JSON format). At this point, the protocol is no longer generic IIUC, so is it still worth using this intermediate protocol, or should we remove it and incorporate the attestation protocol directly into SVSM and implement a simple https proxy in the host?

I totally understand. There are 2 enums that I have introduced and it can become confusing. I can explain the changes.

AFAIK, SVSM can be deployed anywhere. So when its communicating to a backend storage server (to retrieve the secret key) or a verifier to produce evidence, it needs to know what protocol to use. Since we can have multiple such backends, we have tools/aproxy/backend where all of them can be defined in the future. Apart from that it also helps in the communication (HTTP).

Previously we only used to communicate with the kbs-test backend. 
SVSM <--> Aproxy <--> kbs-test 
It accepted evidence that was created by concatenating raw binary public key + nonce, hashed using sha512. 
So SVSM only had the code to do exactly that. 

With this series, we can also communicate with the CoCo Trustee KBS backend.
SVSM <--> Aproxy <--> Trustee KBS

You are right about the order of parameters in the raw binary format. I can bring back the negotiation param and thereby order the payload in RawBinary hash according to the order mentioned in the vector. 
But while developing, I figured that most storage backends or verifiers available in the market for example Intel, Microsoft,.. predominantly uses the JSON format.
[Microsoft](https://learn.microsoft.com/en-us/dotnet/api/overview/azure/security.attestation-readme?view=azure-dotnet#set-an-attestation-policy-for-a-specified-attestation-type)
[Intel](https://docs.trustauthority.intel.com/main/articles/articles/ita/concept-attestation-tokens.html#attester-claims)
I can be totally wrong here, but this is the reason why I shifted the binary format inside
KbsBackendType::Legacy => (HashAlgo::Sha512, PayloadFormat::RawBinary) and fixed the order. It is backward compatible and the kbs-test works fine with SVSM.

About the fixed JSON field names,
If a future protocol uses a different JSON schema (e.g., "algorithm" instead of "alg", or different nesting structures), we cannot let SVSM dynamically rename keys based on aproxy input.
1. Doing so would require SVSM to parse a dynamic key-value map and construct arbitrary JSON payloads. If we introduce CBOR, then we would have to construct that payload dynamically.
2. Trying to make JSON schemas dynamically configurable by the untrusted host IMO could be a security risk. Instead, SVSM keeps its formatting statically typed and compiled.

If a future protocol uses a different schema, we simply:
   1. Add a new compile-time PayloadFormat variant (e.g., PayloadFormat::Cbor or PayloadFormat::AzureKv).
   2. Implement a dedicated, statically verified PayloadFormatter for that layout in SVSM.

I also agree that we can revert the renames in libaproxy and SVSM, going back to report and certs_buf. The naming is only required in aproxy (specifically in TrusteeEvidence).

So to summarize the flow, SVSM can run in any environment. Aproxy will be started on the host with appropriate command line arguments (kbs, kbs-trustee or anything else in the future). This will tell aproxy to fix a communication protocol (Payload format + hash algo). So when aproxy forwards the nonce from storage backend to SVSM, it explicitly also mentions the protocol.

SVSM constructs the runtime_data hash. This is statically fixed. To draw a parallel, just like the kernel uses a compiled-in driver (like VirtIO) to talk to a specific hardware, SVSM uses compile-time static PayloadFormatter for specific attestation standards.
Once the runtime_data hash is created, it will send the attestation request to /attest endpoint:
1. TEE evidence i.e. signed report_data
3. Cleartext runtime_data (contains nonce and tee-pubkey

The trustee kbs in our case, takes the cleartext runtime_data -> serializes into JWS JSON -> hashes it. Then compares this hash with Tee evidence report_data (hardware signed SNP report).

Please correct me if I am wrong. 
cc: @tylerfanelli 


Note to self:
[ ] Bring back NegotiationParam for raw binary format
[ ] Rename the fields only in aproxy

stefano-garzarella:
> AFAIK, SVSM can be deployed anywhere. So when its communicating to a backend storage server (to retrieve the secret key) or a verifier to produce evidence, it needs to know what protocol to use. Since we can have multiple such backends, we have tools/aproxy/backend where all of them can be defined in the future. Apart from that it also helps in the communication (HTTP).

Not really, the goal was to make SVSM completely attestation protocol-agnostic.

Let me summarize what @berrange, @tylerfanelli , and I discussed some time ago.
Basically, we discussed two options:
1 - Make SVSM completely agnostic regarding the attestation protocol to be used, implement the protocol (e.g. Trustee) entirely in the proxy, and have a generic protocol between SVSM and the proxy. This would ensure that any new attestation protocol to support or change to the them would not require modifications to SVSM, and thus would not affect launch measurements, etc.
2 - Use the proxy as a simple HTTP proxy and fully implement the attestation protocol in SVSM (e.g. Trustee). This drastically simplifies the protocol between the proxy and SVSM, but for every new protocol or change, it was necessary to update SVSM, launch measurements, etc.

We went with Option 1 mainly to avoid making changes to SVSM as much as possible. If we see that Option 1 is difficult to maintain, I'm open to switching to Option 2, but not to having something in between that just complicates things.

> 
> Previously we only used to communicate with the kbs-test backend. SVSM <--> Aproxy <--> kbs-test It accepted evidence that was created by concatenating raw binary public key + nonce, hashed using sha512. So SVSM only had the code to do exactly that.
> 
> With this series, we can also communicate with the CoCo Trustee KBS backend. SVSM <--> Aproxy <--> Trustee KBS

The goal is clear, I'm worried about the design of the protocol between SVSM and the proxy.

> 
> You are right about the order of parameters in the raw binary format. I can bring back the negotiation param and thereby order the payload in RawBinary hash according to the order mentioned in the vector. But while developing, I figured that most storage backends or verifiers available in the market for example Intel, Microsoft,.. predominantly uses the JSON format. [Microsoft](https://learn.microsoft.com/en-us/dotnet/api/overview/azure/security.attestation-readme?view=azure-dotnet#set-an-attestation-policy-for-a-specified-attestation-type) [Intel](https://docs.trustauthority.intel.com/main/articles/articles/ita/concept-attestation-tokens.html#attester-claims) I can be totally wrong here, but this is the reason why I shifted the binary format inside KbsBackendType::Legacy => (HashAlgo::Sha512, PayloadFormat::RawBinary) and fixed the order. It is backward compatible and the kbs-test works fine with SVSM.
> 
> About the fixed JSON field names, If a future protocol uses a different JSON schema (e.g., "algorithm" instead of "alg", or different nesting structures), we cannot let SVSM dynamically rename keys based on aproxy input.
> 
>     1. Doing so would require SVSM to parse a dynamic key-value map and construct arbitrary JSON payloads. If we introduce CBOR, then we would have to construct that payload dynamically.
> 
>     2. Trying to make JSON schemas dynamically configurable by the untrusted host IMO could be a security risk. Instead, SVSM keeps its formatting statically typed and compiled.
> 
> 
> If a future protocol uses a different schema, we simply:
> 
>     1. Add a new compile-time PayloadFormat variant (e.g., PayloadFormat::Cbor or PayloadFormat::AzureKv).
> 
>     2. Implement a dedicated, statically verified PayloadFormatter for that layout in SVSM.

This really brings us back to the discussion of the options I mentioned above. This seems to fall right in the middle of the two, and IMO, we should avoid it at all costs because it takes away the advantages of option 1 and adds the disadvantages of option 2. So IMHO it’s basically the worst-case scenario.

> 
> 
> I also agree that we can revert the renames in libaproxy and SVSM, going back to report and certs_buf. The naming is only required in aproxy (specifically in TrusteeEvidence).
> 
> So to summarize the flow, SVSM can run in any environment. Aproxy will be started on the host with appropriate command line arguments (kbs, kbs-trustee or anything else in the future). This will tell aproxy to fix a communication protocol (Payload format + hash algo). So when aproxy forwards the nonce from storage backend to SVSM, it explicitly also mentions the protocol.

IMO this should be avoided, if we implement option 1, SVSM - proxy protocol should don't know anything about the attestation protocol used, everything should be handled by the proxy and translated into the generic protocol.

Or we should switch to option 2 and have a simple http proxy and let SVSM handle everything.

I hope I've made it clear that I'm concerned we're no longer doing either Option 1 or Option 2, but something in between that loses the benefits of both.

Thanks,
Stefano

armenon-rh:
> > AFAIK, SVSM can be deployed anywhere. So when its communicating to a backend storage server (to retrieve the secret key) or a verifier to produce evidence, it needs to know what protocol to use. Since we can have multiple such backends, we have tools/aproxy/backend where all of them can be defined in the future. Apart from that it also helps in the communication (HTTP).
> 
> Not really, the goal was to make SVSM completely attestation protocol-agnostic.
> 
> Let me summarize what @berrange, @tylerfanelli , and I discussed some time ago. Basically, we discussed two options: 1 - Make SVSM completely agnostic regarding the attestation protocol to be used, implement the protocol (e.g. Trustee) entirely in the proxy, and have a generic protocol between SVSM and the proxy. This would ensure that any new attestation protocol to support or change to the them would not require modifications to SVSM, and thus would not affect launch measurements, etc. 2 - Use the proxy as a simple HTTP proxy and fully implement the attestation protocol in SVSM (e.g. Trustee). This drastically simplifies the protocol between the proxy and SVSM, but for every new protocol or change, it was necessary to update SVSM, launch measurements, etc.
> 
> We went with Option 1 mainly to avoid making changes to SVSM as much as possible. If we see that Option 1 is difficult to maintain, I'm open to switching to Option 2, but not to having something in between that just complicates things.
> 
> > Previously we only used to communicate with the kbs-test backend. SVSM <--> Aproxy <--> kbs-test It accepted evidence that was created by concatenating raw binary public key + nonce, hashed using sha512. So SVSM only had the code to do exactly that.
> > With this series, we can also communicate with the CoCo Trustee KBS backend. SVSM <--> Aproxy <--> Trustee KBS
> 
> The goal is clear, I'm worried about the design of the protocol between SVSM and the proxy.
> 
> > You are right about the order of parameters in the raw binary format. I can bring back the negotiation param and thereby order the payload in RawBinary hash according to the order mentioned in the vector. But while developing, I figured that most storage backends or verifiers available in the market for example Intel, Microsoft,.. predominantly uses the JSON format. [Microsoft](https://learn.microsoft.com/en-us/dotnet/api/overview/azure/security.attestation-readme?view=azure-dotnet#set-an-attestation-policy-for-a-specified-attestation-type) [Intel](https://docs.trustauthority.intel.com/main/articles/articles/ita/concept-attestation-tokens.html#attester-claims) I can be totally wrong here, but this is the reason why I shifted the binary format inside KbsBackendType::Legacy => (HashAlgo::Sha512, PayloadFormat::RawBinary) and fixed the order. It is backward compatible and the kbs-test works fine with SVSM.
> > About the fixed JSON field names, If a future protocol uses a different JSON schema (e.g., "algorithm" instead of "alg", or different nesting structures), we cannot let SVSM dynamically rename keys based on aproxy input.
> > ```
> > 1. Doing so would require SVSM to parse a dynamic key-value map and construct arbitrary JSON payloads. If we introduce CBOR, then we would have to construct that payload dynamically.
> > 
> > 2. Trying to make JSON schemas dynamically configurable by the untrusted host IMO could be a security risk. Instead, SVSM keeps its formatting statically typed and compiled.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If a future protocol uses a different schema, we simply:
> > ```
> > 1. Add a new compile-time PayloadFormat variant (e.g., PayloadFormat::Cbor or PayloadFormat::AzureKv).
> > 
> > 2. Implement a dedicated, statically verified PayloadFormatter for that layout in SVSM.
> > ```
> 
> This really brings us back to the discussion of the options I mentioned above. This seems to fall right in the middle of the two, and IMO, we should avoid it at all costs because it takes away the advantages of option 1 and adds the disadvantages of option 2. So IMHO it’s basically the worst-case scenario.
> 
> > I also agree that we can revert the renames in libaproxy and SVSM, going back to report and certs_buf. The naming is only required in aproxy (specifically in TrusteeEvidence).
> > So to summarize the flow, SVSM can run in any environment. Aproxy will be started on the host with appropriate command line arguments (kbs, kbs-trustee or anything else in the future). This will tell aproxy to fix a communication protocol (Payload format + hash algo). So when aproxy forwards the nonce from storage backend to SVSM, it explicitly also mentions the protocol.
> 
> IMO this should be avoided, if we implement option 1, SVSM - proxy protocol should don't know anything about the attestation protocol used, everything should be handled by the proxy and translated into the generic protocol.
> 
> Or we should switch to option 2 and have a simple http proxy and let SVSM handle everything.
> 
> I hope I've made it clear that I'm concerned we're no longer doing either Option 1 or Option 2, but something in between that loses the benefits of both.
> 
> Thanks, Stefano

Thank you so much for explaining the design choices. You have perfectly captured the architectural tradeoffs.

I understand the design choice sits right between option1 and option2
But that is because we have some constraints.

1. We cannot implement option 1 fully because,
	As I discovered from the previous review: The hash must be computed inside SVSM.
	If we delegate this to aproxy (like we had in v2 patches), then an untrusted host
	can hash its own key in runtime_data and sign the report using AMD Secure processor.
	SVSM cannot output a binary hash and expect the aproxy to translate it into
	JWS JSON hash later. Therefore, the exact payload structure must be formatted
	inside SVSM before hashing it.
	
	Now I agree that we also want to keep SVSM agnostic of the backend storage.
	SVSM does not know about Trustee or specific backend servers. Instead it is only
	aware of standard serialization formats (JwsJson or RawBinary).
	If we come across other backends expecting the same formats, we will be able to
	configure them without having to make any changes in SVSM.
	Maybe we can call it format aware and not backend aware.
	
2. We cannot implement option 2 fully, because AFAIK we do not have TCP/IP stack in SVSM.
   Adding that in SVSM can increase the attack surface.
   
The middleground is that the proxy must handle the HTTP layer but SVSM must handle the JwsJson formatting and hashing. This achieves security (from host key substitution) and network safety.

I totally understand the concerns. But I could not figure out a way to achieve both at the same time.

stefano-garzarella:
> I understand the design choice sits right between option1 and option2 But that is because we have some constraints.
> 
>     1. We cannot implement option 1 fully because,
>        As I discovered from the previous review: The hash must be computed inside SVSM.
>        If we delegate this to aproxy (like we had in v2 patches), then an untrusted host
>        can hash its own key in runtime_data and sign the report using AMD Secure processor.

We already discussed about that, and we agreed we can't do that. Indeed we introduced `NegotiationParam` exactly for that reason...

>        SVSM cannot output a binary hash and expect the aproxy to translate it into
>        JWS JSON hash later. Therefore, the exact payload structure must be formatted
>        inside SVSM before hashing it.

Right, so the proxy should tell to SVSM how to construct the json, what to put in some field, what name to use, etc. Like we did with `NegotiationParam` but extending them (e.g. introducing the name to be used in a json) or using some kind of template, or something else. IMO having protocol-specific stuff (order, names, etc.) in SVSM is wrong and must be avoided. If we need to do that, we should move to option 2.

>        Now I agree that we also want to keep SVSM agnostic of the backend storage.
>        SVSM does not know about Trustee or specific backend servers. Instead it is only
>        aware of standard serialization formats (JwsJson or RawBinary).
>        If we come across other backends expecting the same formats, we will be able to
>        configure them without having to make any changes in SVSM.
>        Maybe we can call it format aware and not backend aware.

I don't see the point of doing this TBH. You mean ask other attestation protocols to change to be adapted to our internal protocol? Why? At this point we should go through option 2, I don't understand why we have to complicate our own lives and those of others.

What are the advantages of this approach, compared with option 2?

> 
>     2. We cannot implement option 2 fully, because AFAIK we do not have TCP/IP stack in SVSM.
>        Adding that in SVSM can increase the attack surface.

We did that without any TCP/IP stack in SVSM. It was my first attempt, let me find the code: https://github.com/stefano-garzarella/RAClients

The protocol was very simple:
```
#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
pub enum HttpMethod {
    GET,
    POST,
}

#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
pub struct Request {
    pub endpoint: String,
    pub method: HttpMethod,
    pub body: Value,
}

#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
pub struct Response {
    pub status: u16,
    pub body: String,
}
```

The proxy will be used exactly to avoid TCP/IP in SVSM, otherwise we don't need the proxy at all if SVSM can talk TCP/IP.

SVSM will create the entire payload of the message, send it to the proxy with our custom protocol, the proxy will put it in a http packet and send it to the server, opposite on the other direction.

armenon-rh:
Hi,

I have implemented [option2](https://github.com/coconut-svsm/svsm/pull/1203).

In the new design, aproxy is a simple HTTP forwarder. SVSM does the job of constructing secure runtime_data and hashing it. A combination of Payloadformat and Hashalgo (mentioned using --protocol) allows the user to indicate what kind of data format is expected by the backend while bringing up aproxy.

Since we no longer import the external sev crate to match the AttestationReport in aproxy, SVSM defines a few structs in kernel/src/greq/pld_report.rs to help with serialization of binary data into JSON before sending it across to the backend.

I still need to add documentation in the PR, but I will start working on it once the design is approved.

Thanks
Arun.

kernel/src/attest.rs
```diff
@@ -360,12 +360,12 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
...
             // AttestationRequest.
-            let report =
+            let attestation_report =
                 try_to_vec(resp.report().as_bytes()).or(Err(AttestationError::VecAlloc))?;
 
             AttestationEvidence::Snp {
-                report,
-                certs_buf: None,
+                attestation_report,
+                cert_chain: None,
```
tylerfanelli, line 368: While I don't disagree with this change on principle, the SVSM kernel's field names should not be modified to fit Trustee. The SVSM kernel's fields are generic, and not tied to one server.

These field names are a bit more specific, however. In that case this change is fine.

Also, can you remove the link in the commit message?

libaproxy/src/lib.rs
```diff
@@ -16,7 +16,8 @@ pub use attestation::*;
 pub use negotiation::*;
 
 use alloc::{string::String, vec::Vec};
-use base64::{Engine, prelude::BASE64_STANDARD};
+pub use base64::Engine;
+pub use base64::prelude::{BASE64_STANDARD, BASE64_URL_SAFE_NO_PAD};
```
tylerfanelli, line 20: `pub use base64::{prelude::{BASE64_STANDARD, BASE64_URL_SAFE_NO_PAD}, Engine};`

tools/aproxy/src/main.rs
```diff
@@ -49,7 +49,10 @@ struct Args {
 /// to work.
 #[derive(Clone, Copy, Debug, ValueEnum)]
 enum ArgsBackend {
+    /// Legacy or Mock KBS (raw public key & challenge, SHA-512)
     Kbs,
+    /// Trustee KBS (JWS JSON, SHA-384)
+    KbsTrustee,
 }
```
tylerfanelli, line 56:
```suggestion
enum ArgsBackend {
    /// Legacy or Mock KBS (raw public key & challenge, SHA-512)
    SampleKbs,
    /// Trustee KBS (JWS JSON, SHA-384)
    TrusteeKbs,
}
```

tools/aproxy/src/backend/kbs.rs
```diff
@@ -16,8 +16,32 @@ use reqwest::StatusCode;
...
 
-#[derive(Clone, Copy, Debug, Default)]
-pub struct KbsProtocol;
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum KbsBackendType {
+    /// Trustee KBS backend using JWS-compliant JSON (SHA-384)
+    Trustee,
+    /// Legacy or Mock KBS client backend using raw key and challenge bytes (SHA-512)
+    Legacy,
+}
```
tylerfanelli, line 25: I think the Trustee KBS protocol should just be its own module. Call this the `sample_kbs` module.
```
tools/aproxy/src/backend/sample_kbs.rs
```
and the Trustee module `trustee`.
```
tools/aproxy/src/backend/trustee.rs
```

Since they diverge so drastically now.

tylerfanelli (changes requested):
Thanks @armenon-rh. I have one main suggestion for this PR. @stefano-garzarella I'd also like your thoughts.

In the negotiation phase of the protocol, we originally designed the NegotiationResponse to contain a set of negotiation parameters that the server expects to be included in the HOST_DATA hash.

```
struct NegotiationResponse {
    /// Challenge returned from the attestation server to verify freshness of attestation evidence.
    pub challenge: Vec<u8>,

    /// Parameters to be hashed in the specific order defined by the array
    pub params: Vec<NegotiationParam>,
}
``` 

However, perhaps this is **too generic**. AFAIU, each server implementation (Trustee KBS being at the forefront of my mind when considering this) has only one expected format that HOST_DATA hashes should follow. With this in mind, why don't we just throw out the negotiation parameters field altogether and instead specify which server implementation is being used, so SVSM can hash the parameters in the order and format expected by the server. For Trustee KBS, this could be as follows:
```
struct NegotiationResponse {
    /// Challenge returned from the attestation server to verify freshness of attestation evidence.
    pub challenge: Vec<u8>,

    /// The protocol (ordering and formatting of negotiation parameters) expected by the server.
    pub hash_protocol: HashProtocol,
}
``` 

`HashProtocol` can be an enum specifying the server implementation, like so:
```
pub enum HashProtocol {
    /// Sample KBS used for testing.
    SampleKbs,

    /// Trustee KBS.
    TrusteeKbs,
}
```

Then, SVSM can compile the HOST_DATA hash based on the result of the `hash_protocol` specified in the negotiation response. With this, the hash can be constructed in the correct ordering and formatting (in Trustee's case, JWS JSON).

The downside to this is we're essentially bringing server-specific traits to the SVSM kernel. However, I think this is a relatively small area (i.e. how the negotiation parameters are hashed) requiring this change and thus worth the trouble.

Furthermore, this change is extensible, as when another server proxy comes along, it can specify another `HashProtocol` enum value and write a module of how to compile the HOST_DATA hash in the way expected by that server.

stefano-garzarella:
> The downside to this is we're essentially bringing server-specific traits to the SVSM kernel. However, I think this is a relatively small area (i.e. how the negotiation parameters are hashed) requiring this change and thus worth the trouble.
> 
> Furthermore, this change is extensible, as when another server proxy comes along, it can specify another `HashProtocol` enum value and write a module of how to compile the HOST_DATA hash in the way expected by that server.

IMO, this means we lose all the benefits of having a generic protocol between SVSM and the proxy, since if we need to support a new attestation protocol or a new version of it, we have to update SVSM and then have a new launch measurement, etc.

At this point, what would be the advantage of having a complicated generic protocol between SVSM and the proxy compared to having a simple HTTP proxy and implementing the entire protocol in SVSM?

The advantages of the latter are a very simple proxy, a simple protocol between the proxy and SVSM, and the entire attestation protocol implemented in a single trusted location (i.e. SVSM), so we can avoid any bad assumption (e.g. making the host to perform the hash, etc.).

tools/aproxy/src/backend/kbs.rs
stefano-garzarella: Yeah, agree on this! `sample_kbs` or `kbs_test` to make it clear is the protocol talked about our `kbs-test` crate.

kernel/src/attest.rs
```diff
@@ -375,26 +375,40 @@ fn evidence(tee: &Tee, hash: Vec<u8>) -> Result<AttestationEvidence, Attestation
...
+    let mut h = D::new();
+    h.update(data);
+    try_to_vec(&h.finalize()).or(Err(AttestationError::VecAlloc))
+}
+
+trait HashAlgoExt {
+    fn digest(&self, data: &[u8]) -> Result<Vec<u8>, AttestationError>;
+}
+
+impl HashAlgoExt for HashAlgo {
```
armenon-rh, line 388: I can use kbs_types::HashAlgorithm and get rid of this dispatch.

tools/aproxy/src/main.rs
```diff
@@ -49,7 +49,10 @@ struct Args {
 /// to work.
 #[derive(Clone, Copy, Debug, ValueEnum)]
 enum ArgsBackend {
+    /// Legacy or Mock KBS (raw public key & challenge, SHA-512)
     Kbs,
```
armenon-rh, line 53: LegacyKbs

tools/aproxy/src/main.rs
```diff
@@ -49,7 +49,10 @@ struct Args {
...
+    /// Trustee KBS (JWS JSON, SHA-384)
+    KbsTrustee,
```
armenon-rh, line 55: TrusteeKbs

libaproxy/src/negotiation.rs
armenon-rh: 2 new enums added to negotiation response. They communicate payload format ( binary vs json) and hash algorithm (sha256, sha384) to SVSM. SVSM uses this to construct the report data (using public key and nonce) and hash it.

Documentation/docs/developer/KBS-SVSM.md
armenon-rh: https://github.com/confidential-containers/trustee/commit/51bd7bb544b8dc5f9b58e4bc5b533028299dd85a

I am not sure if these changes should be in trustee upstream. This was only created for convenience. 

It has the necessary "manual testing" configuration toml files, that help us set up the container and connect to it quickly with a few instructions.


Please guide @tylerfanelli

kernel/src/attest.rs
armenon-rh: Trustee expects the report data in a particular format and hashed with a particular hashing algorithm. 
Our Proxy, in this PR, communicates this information (PayloadFormat + HashAlgo) to SVSM. Therefore, yes, we do have format specific code in SVSM. 
But this is extensible to any other backend that we may introduce in the future (ex. Keylime).

tools/aproxy/src/main.rs
armenon-rh: done in v2.

kernel/src/attest.rs
armenon-rh: The hashing is now done in the SVSM and not the host aproxy. Therefore this attack is prevented.
Even if it manages to replace the key, at the remote verifier end, the hashes will not match thereby not leaking the secret. Fixed in v2.

libaproxy/src/lib.rs
armenon-rh: yes, done in v2.

kernel/src/attest.rs
armenon-rh: yes, done in v2.

tools/aproxy/src/backend/kbs.rs
armenon-rh: done in v2.
Introduced SampleKbs and TrusteeKbs generic protocol backend variants.

armenon-rh:
> Thanks @armenon-rh. I have one main suggestion for this PR. @stefano-garzarella I'd also like your thoughts.
> 
> In the negotiation phase of the protocol, we originally designed the NegotiationResponse to contain a set of negotiation parameters that the server expects to be included in the HOST_DATA hash.
> 
> ```
> struct NegotiationResponse {
>     /// Challenge returned from the attestation server to verify freshness of attestation evidence.
>     pub challenge: Vec<u8>,
> 
>     /// Parameters to be hashed in the specific order defined by the array
>     pub params: Vec<NegotiationParam>,
> }
> ```
> 
> However, perhaps this is **too generic**. AFAIU, each server implementation (Trustee KBS being at the forefront of my mind when considering this) has only one expected format that HOST_DATA hashes should follow. With this in mind, why don't we just throw out the negotiation parameters field altogether and instead specify which server implementation is being used, so SVSM can hash the parameters in the order and format expected by the server. For Trustee KBS, this could be as follows:
> 
> ```
> struct NegotiationResponse {
>     /// Challenge returned from the attestation server to verify freshness of attestation evidence.
>     pub challenge: Vec<u8>,
> 
>     /// The protocol (ordering and formatting of negotiation parameters) expected by the server.
>     pub hash_protocol: HashProtocol,
> }
> ```
> 
> `HashProtocol` can be an enum specifying the server implementation, like so:
> 
> ```
> pub enum HashProtocol {
>     /// Sample KBS used for testing.
>     SampleKbs,
> 
>     /// Trustee KBS.
>     TrusteeKbs,
> }
> ```
> 
> Then, SVSM can compile the HOST_DATA hash based on the result of the `hash_protocol` specified in the negotiation response. With this, the hash can be constructed in the correct ordering and formatting (in Trustee's case, JWS JSON).
> 
> The downside to this is we're essentially bringing server-specific traits to the SVSM kernel. However, I think this is a relatively small area (i.e. how the negotiation parameters are hashed) requiring this change and thus worth the trouble.
> 
> Furthermore, this change is extensible, as when another server proxy comes along, it can specify another `HashProtocol` enum value and write a module of how to compile the HOST_DATA hash in the way expected by that server.



> Thanks @armenon-rh. I have one main suggestion for this PR. @stefano-garzarella I'd also like your thoughts.
> 
> In the negotiation phase of the protocol, we originally designed the NegotiationResponse to contain a set of negotiation parameters that the server expects to be included in the HOST_DATA hash.
> 
> ```
> struct NegotiationResponse {
>     /// Challenge returned from the attestation server to verify freshness of attestation evidence.
>     pub challenge: Vec<u8>,
> 
>     /// Parameters to be hashed in the specific order defined by the array
>     pub params: Vec<NegotiationParam>,
> }
> ```
> 
> However, perhaps this is **too generic**. AFAIU, each server implementation (Trustee KBS being at the forefront of my mind when considering this) has only one expected format that HOST_DATA hashes should follow. With this in mind, why don't we just throw out the negotiation parameters field altogether and instead specify which server implementation is being used, so SVSM can hash the parameters in the order and format expected by the server. For Trustee KBS, this could be as follows:
> 
> ```
> struct NegotiationResponse {
>     /// Challenge returned from the attestation server to verify freshness of attestation evidence.
>     pub challenge: Vec<u8>,
> 
>     /// The protocol (ordering and formatting of negotiation parameters) expected by the server.
>     pub hash_protocol: HashProtocol,
> }
> ```
> 
> `HashProtocol` can be an enum specifying the server implementation, like so:
> 
> ```
> pub enum HashProtocol {
>     /// Sample KBS used for testing.
>     SampleKbs,
> 
>     /// Trustee KBS.
>     TrusteeKbs,
> }
> ```
> 
> Then, SVSM can compile the HOST_DATA hash based on the result of the `hash_protocol` specified in the negotiation response. With this, the hash can be constructed in the correct ordering and formatting (in Trustee's case, JWS JSON).
> 
> The downside to this is we're essentially bringing server-specific traits to the SVSM kernel. However, I think this is a relatively small area (i.e. how the negotiation parameters are hashed) requiring this change and thus worth the trouble.
> 
> Furthermore, this change is extensible, as when another server proxy comes along, it can specify another `HashProtocol` enum value and write a module of how to compile the HOST_DATA hash in the way expected by that server.

Thank you @Tyler for the detailed explanation. This is what I am adding.

```

/// A parameter that must be hashed into the negotiation hash.
#[derive(Serialize, Deserialize, Debug)]
pub enum NegotiationParam {
    /// Hash the challenge returned from attestation server.
    Challenge,
    /// Hash the EC public key's `Elliptic-Curve-Point-to-Octet-String` encoding.
    EcPublicKeyBytes,
}

pub enum PayloadFormat {
    /// Raw sequential binary representation of public key coordinates and challenge.
    RawBinary,
    /// JWS-compliant JSON formatted representation of runtime_data.
    JwsJson,
}

#[derive(Serialize, Deserialize, Debug)]
pub struct NegotiationResponse {
    /// Challenge returned from the attestation server to verify freshness of attestation evidence.
    #[serde(
        serialize_with = "serialize_base64",
        deserialize_with = "deserialize_base64"
    )]
    pub challenge: Vec<u8>,
    /// The hashing algorithm to use.
    pub hash_algo: HashAlgo,
    /// The payload formatting to use.
    pub payload_format: PayloadFormat,
}
```
NegotiationParam is still kept so that kbs-test work with the RawBinary formatter

Documentation/docs/developer/ATTESTATION.md
armenon-rh: After the [comment](https://github.com/coconut-svsm/svsm/pull/1122#pullrequestreview-5174191254) from Tyler, and the discussion we had, it is now clear that we need to specify a format, and the hash algorithm to svsm from aproxy. This is implemented in the latest commit v2.

kernel/src/attest.rs
armenon-rh: That will require a change in SVSM. The remote verifiers do not seldom change the format. If we need to add any other format used by a different backend like Keylime, then that is possible by extending the PayloadFormatter implementation.

armenon-rh:
> I see a lot of changes in `Cargo.lock` file, I don't think this is expected..

The sev crate included brings in some dependencies.

kernel/src/attest.rs
armenon-rh: If json format is used, ordering will almost always be alphaetical.
If a new remote verifier needs a payload format, then that needs to be supported in SVSM by extending PayloadFormatter.

PR #1123, opened 2026-06-16
aproxy: Support wrap-key PKCS#11 plugin KBS backend

armenon-rh:
The request from aproxy to KBS, post a successful attestation, could
be a secret resource stored on the storage backend or a simple call to
the numerous plugins that KBS offers.

SVSM intends to use the wrap-key endpoint of KBS to retrieve the
unwrapped key to decrypt the vTPM state file. The SecretRequest enum
captures these two variants, and can be extended in future.
This is added to AttestationRequest from SVSM and the aproxy can hit
the KBS wrap-key endpoint immediately post attestation.

This PR depends on: #1122. Please review only the commits ddcdde6500ccd6dce8544c112b15b5ec95a41dde and 86cdf74441a10f97bb237feacd2a77247e82c7ac


armenon-rh:
Not looking at this PR now.

PR #1124, opened 2026-06-16
kernel/vtpm: Support vtpm invocation using an external state file

armenon-rh:
Add vTPM state file as a parameter to vtpm_init() allowing it to initialize the vTPM with an external state file.

In order to create an external vTPM state file image, that can be attached to svsm as a virtio-blk device, use [vtpm-to-svsm-blk](https://github.com/armenon-rh/vtpm-to-svsm-blk)

This PR depends on: #1123. Please review only the commits 6cfe74e4133b769ca090bf8ac2f1f71ab4f118a4, 9af3f2c743971465e58c1ad23d0a41ed7036d3d7


armenon-rh:
Not looking into this PR now.

PR #1125, opened 2026-06-16
igvmbuilder: Move COCONUT-SVSM to 128MB physical address

joergroedel:
Move the COCONUT-SVSM memory region into guest RAM at 128MB from the previous GPA at 512GB. This will allow the SVSM to allocate more memory for itself at boot.

A recent EDK2 built from the latest `master` branch is needed to support this on the guest side.

### DIFF
--- a/tools/igvmbuilder/src/gpa_map.rs
+++ b/tools/igvmbuilder/src/gpa_map.rs
@@ -126,17 +126,14 @@ impl GpaMap {
 
         // Choose the kernel base and maximum size.
         let (kernel_base, kernel_max_size) = match options.hypervisor {
-            Hypervisor::Qemu => {
-                // Place the kernel area at 512 GB with a maximum size of 16 MB.
-                (0x0000008000000000, 0x01000000)
-            }
-            Hypervisor::HyperV => {
-                // Place the kernel area at 64 MB with a maximum size of 16 MB.
-                (0x04000000, 0x01000000)
-            }
             Hypervisor::Vanadium => {
                 // Place the kernel area at 8TiB-2GiB with a maximum size of 2 GiB.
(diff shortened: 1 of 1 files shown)


tools/igvmbuilder/src/gpa_map.rs
```diff
@@ -127,8 +127,8 @@ impl GpaMap {
         // Choose the kernel base and maximum size.
         let (kernel_base, kernel_max_size) = match options.hypervisor {
             Hypervisor::Qemu => {
-                // Place the kernel area at 512 GB with a maximum size of 16 MB.
-                (0x0000008000000000, 0x01000000)
+                // Place the kernel area at 128 MB with a maximum size of 16 MB.
```
msft-jlange: Can we do this for Hyper-V as well?  There will be no problem with 128 GB compared with 64 GB, and it would be nice to remove one more difference between hypervisors.

joergroedel: Sure, done.

msft-jlange: Can we replace the QEMU/Hyper-V arms with `_` so that we don't have two copies of "128 MB" in the match?  Make 128 MB be the default and let Vanadium be the only exception.  This would look a lot cleaner.

msft-jlange (approved):

tools/igvmbuilder/src/gpa_map.rs
joergroedel: Makes sense, PR is updated.

00xc:
SVSM tests and Linux boot test passed.

00xc (approved):

PR #1126, opened 2026-06-16
kernel: cpu/apic: various fixes

00xc:
Fix a variety of issues with the APIC emulation code:
* Ignore errors when requesting host EOI
* Verify `value != 0` for EOI writes (as mandated by spec).
* Verify verify must-be-zero bits for ICR writes (as mandated by spec).
* Fix index used in `get_ppr_with_tpr()`
* Fix NMI delivery in physical mode for current CPU.
* Handle LDR register reads (as mandated by spec).
* Fix returned error on invalid register access  (as mandated by spec).

### DIFF
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -561,7 +561,7 @@ impl LocalApic {
             IcrMessageType::Fixed | IcrMessageType::Nmi
         );
 
-        if !valid_type {
+        if !valid_type || icr.rsvd_13() || icr.rsvd_31_20() != 0 {
             return Err(SvsmError::Apic(Emulation));
         }
 
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -7,7 +7,7 @@
 use crate::cpu::idt::common::INT_INJ_VECTOR;
 use crate::cpu::percpu::{PERCPU_AREAS, PerCpuShared, current_ghcb, this_cpu};
 use crate::cpu::x86::apic_post_irq;
-use crate::error::ApicError::Emulation;
+use crate::error::ApicError::{Emulation, InvalidRegister};
 use crate::error::SvsmError;
 use crate::mm::GuestPtr;
 use crate::platform::guest_cpu::GuestCpuState;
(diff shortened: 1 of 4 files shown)


kernel/src/cpu/apic.rs
```diff
@@ -323,8 +324,7 @@ impl LocalApic {
     fn perform_host_eoi(vector: u8) {
         // Errors from the host are not expected and cannot be meaningfully
         // handled, so simply ignore them.
-        let _r = current_ghcb().specific_eoi(vector, GUEST_VMPL.try_into().unwrap());
-        assert!(_r.is_ok());
+        let _ = current_ghcb().specific_eoi(vector, GUEST_VMPL.try_into().unwrap());
```
msft-jlange: Can we at least log an error if the host doesn't cooperate here?

msft-jlange (approved):

kernel/src/cpu/apic.rs
```diff
@@ -561,7 +561,7 @@ impl LocalApic {
             IcrMessageType::Fixed | IcrMessageType::Nmi
         );
 
-        if !valid_type {
+        if !valid_type || icr.rsvd_13() || icr.rsvd_31_20() != 0 {
```
stefano-garzarella, line 564: Looking at KVM code, they have a similar check, but they include also bits 16, 17 (`remote_read_status` IIUC)  and reset bit 12 because of some Intel/AMD discrepancy:
```
#define X2APIC_ICR_RESERVED_BITS (GENMASK_ULL(31, 20) | GENMASK_ULL(17, 16) | BIT(13))

static int __kvm_x2apic_icr_write(struct kvm_lapic *apic, u64 data, bool fast)
{
	if (data & X2APIC_ICR_RESERVED_BITS)
		return 1;

	/*
	 * The BUSY bit is reserved on both Intel and AMD in x2APIC mode, but
	 * only AMD requires it to be zero, Intel essentially just ignores the
	 * bit.  And if IPI virtualization (Intel) or x2AVIC (AMD) is enabled,
	 * the CPU performs the reserved bits checks, i.e. the underlying CPU
	 * behavior will "win".  Arbitrarily clear the BUSY bit, as there is no
	 * sane way to provide consistent behavior with respect to hardware.
	 */
	data &= ~APIC_ICR_BUSY;

```

Should we do something similar?

stefano-garzarella (commented):
I left a minor comment that can be eventually fixed later.

kernel/src/cpu/apic.rs
00xc: Bits 16 and 17 should be checked I think.

As for bit 12, my guess is that we also need to explicitly reject it, since APIC emulation only makes sense on SEV-SNP with restricted injection. On Intel, the TDX module takes care of APIC reads/writes, so there is no need for SVSM involvement, and the divergence in how the ICR busy bit is treated is not relevant (we only need to emulate AMD behavior).

@msft-jlange is this correct?

joergroedel (approved):

00xc:
@joergroedel I think we raced :) I'll open a separate PR for Stefano's comment

kernel/src/cpu/apic.rs
00xc: @stefano-garzarella see #1134

PR #1127, opened 2026-06-16
xbuild: build test kernels directly

00xc:
Currently, building test kernels is a complex process: the user invokes the Makefile, the Makefile invokes xbuild with a test configuration, and the test configuration directs xbuild to call back into the Makefile to build the kernel component.

Instead, add the dedicated infrastructure in xbuild to not depend on the Makefile, removing the last users of the "make" build type in the build recipes. This is done by adding a new build type, "test", which invokes `cargo test --no-run` and picks up the built binary.

This PR also includes a cleanup of the TDX build, and a fix to correctly forward `RELEASE=1` for test kernels.

### Before

JSON recipe:

```json
        "svsm": {
            "type": "make",
            "output_file": "bin/test-kernel.elf"
        },
```

Makefile:

```makefile
bin/test-kernel.elf: bin
	RUSTDOC=true LINK_TEST=1 cargo +nightly test --package svsm ${CARGO_ARGS} ${SVSM_ARGS_TEST} \
		--target=x86_64-unknown-none \
		--config 'target.x86_64-unknown-none.runner=["sh", "-c", "cp $$0 ../$@"]'
```

### After

JSON recipe:

```json
        "svsm": {
            "type": "test",
            "toolchain": "nightly",
            "no_default_features": true,
            "env": {
                "RUSTDOC": "true",
                "LINK_TEST": "1"
            }
        },
```

No Makefile involvement

### DIFF
--- a/xbuild/src/main.rs
+++ b/xbuild/src/main.rs
@@ -143,6 +143,8 @@ struct ComponentConfig {
     #[serde(default)]
     features: Option<String>,
     #[serde(default)]
+    no_default_features: bool,
+    #[serde(default)]
     binary: bool,
     #[serde(default)]
     objcopy: Objcopy,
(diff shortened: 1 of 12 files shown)


xbuild/src/main.rs
n-ramacciotti, line 126: Not sure if this is the right PR, but since you're touching `xbuild`:

(in #926) When in `FsConfig`, I try to build with your suggestion (`/test/usermodule`) the copy fails as `test` dir is not available. 

Should this function (`copy`) also checks if the `dst.parent()` exists and create it if it doesn't?  (or doing this in fsConfig::build directly before calling `copy`)

In this case the dst.pop() in `Fsconfig::build` should also take into account multiple pop after dst.push() with multiple parts?

joergroedel: Yes, that behavior of xbuild makes sense. But it is not in the scope of this PR, so let's do this as a follow-on.

joergroedel (approved):

PR #1128, opened 2026-06-17
kernel/task: save exit code

### LINKED ISSUE #1105: kernel/task: save exit code
After the introduction of `wait_for_termination` in [#1044](https://github.com/coconut-svsm/svsm/pull/1044), it is possible to wait for the completion of a task. 

When a task terminates, it can be useful to save its exit code for later use.

One use case is to retrieve the exit code of task A after task B waits for task A to terminate
during userspace tests in svsm, to determine if a test has failed. In this case, the failure causes a panic,
which results in an `exit` syscall  with a non-zero code (see [#926](https://github.com/coconut-svsm/svsm/pull/926)).

Is anyone working on this? I'd be happy to help.

n-ramacciotti:
Add exit code handling to `Task`  introducing it in `TaskSchedState.state`, making it an `AtomicU64` from `AtomicU32` to store both `TaskState` and the exit code.

Add exit code handling to the termination functions chain.

I added a `[test]` commit to set and assert the exit value of `init`.

I'm not sure this is the right approach, so I marked this as draft:
- I think that `Taskstate::TERMINATED` and `exit_code` should be updated together. Another possibility was to introduce an additional field `exit_code` in `Task`, but with this I think an additional lock should be added to synchronize `TaskSchedState.state` and `Task.exit_code`. If this is not the case I can move the field to `Task`

I'd like some feedback!

Fixes: #1105 

[@stefano-garzarella ]

### DIFF
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -28,7 +28,7 @@ use crate::mm::GuestPtr;
 use crate::mm::PAGE_SIZE;
 use crate::platform::PageValidateOp;
 use crate::platform::SvsmPlatform;
-use crate::task::{is_task_fault, terminate};
+use crate::task::{TaskExitStatus, is_task_fault, terminate};
 use crate::tdx::ve::handle_virtualization_exception;
 use crate::utils::MemoryRegion;
 use crate::utils::immut_after_init::ImmutAfterInitCell;
@@ -194,7 +194,7 @@ extern "C" fn ex_handler_terminate(ctx: &mut X86ExceptionContext, vector: usize)
             vector,
             { ctx.frame.rip }
         );
-        terminate();
+        terminate(Some(TaskExitStatus::Exception));
     } else {
         // Kernel-mode tasks should never cause a termination class exception,
         // so any such exception is grounds for panic.
@@ -225,7 +225,7 @@ extern "C" fn ex_handler_double_fault(ctxt: &mut X86ExceptionContext) {
         log::error!(
             "Double-Fault at RIP {rip:#018x} RSP: {rsp:#018x} CR2: {cr2:#018x} - Terminating task"
         );
-        terminate();
+        terminate(Some(TaskExitStatus::Exception));
     } else {
         panic!(
             "Double-Fault at RIP {:#018x} RSP: {:#018x} CR2: {:#018x}",
@@ -245,7 +245,7 @@ extern "C" fn ex_handler_general_protection(ctxt: &mut X86ExceptionContext) {
         log::error!(
             "Unhandled General-Protection-Fault at RIP {rip:#018x} error code: {err:#018x} rsp: {rsp:#018x} - Terminating task"
         );
-        terminate();
+        terminate(Some(TaskExitStatus::Exception));
     } else if !handle_exception_table(ctxt) {
         panic!(
             "Unhandled General-Protection-Fault at RIP {:#018x} error code: {:#018x} rsp: {:#018x}",
@@ -290,7 +290,7 @@ extern "C" fn ex_handler_page_fault(ctxt: &mut X86ExceptionContext, vector: usiz
             log::error!(
                 "Unexpected user-mode page-fault at RIP {rip:#018x} CR2: {cr2:#018x} error code: {err:#018x} - Terminating task"
             );
-            terminate();
+            terminate(Some(TaskExitStatus::Exception));
         }
     } else if this_cpu()
         .handle_pf(
@@ -377,7 +377,7 @@ extern "C" fn ex_handler_ve(ctxt: &mut X86ExceptionContext) {
             log::error!(
                 "Failed to handle #VE from user-mode at RIP {rip:#018x} code: {code:#018x} - Terminating task"
             );
-            terminate();
+            terminate(Some(TaskExitStatus::Exception));
         } else {
             panic!(
                 "Failed to handle #VE from kernel-mode at RIP {:#018x} code: {:#018x}",
@@ -399,7 +399,7 @@ extern "C" fn ex_handler_vmm_communication(ctxt: &mut X86ExceptionContext, vecto
             log::error!(
                 "Failed to handle #VC from user-mode at RIP {rip:#018x} code: {code:#018x} - Terminating task"
             );
-            terminate();
+            terminate(Some(TaskExitStatus::Exception));
         } else {
             panic!(
                 "Failed to handle #VC from kernel-mode at RIP {:#018x} code: {:#018x}",
--- a/kernel/src/syscall/class0.rs
+++ b/kernel/src/syscall/class0.rs
@@ -9,17 +9,19 @@ use crate::address::VirtAddr;
 use crate::cpu::percpu::current_task;
 use crate::fs::find_dir;
 use crate::mm::guestmem::UserPtr;
+use crate::task::TaskExitStatus;
 use crate::task::exec_user;
 use crate::task::terminate;
 use core::ffi::c_char;
 use syscall::SysCallError;
 
 pub fn sys_exit(exit_code: u32) -> ! {
+    let code = (exit_code & 0xffff) as u16;
     log::info!(
-        "Terminating task {}, exit_code {exit_code}",
+        "Terminating task {}, exit_code {code}",
         current_task().get_task_name()
     );
-    terminate();
+    terminate(Some(TaskExitStatus::Exited(code)));
 }
 
 pub fn sys_exec(file: usize, root: usize, _flags: usize) -> Result<u64, SysCallError> {
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -36,6 +36,7 @@ extern crate alloc;
 
 use super::tasks::TASK_ACTIVE_OFFSET;
 use super::tasks::TASK_CUR_CPU_OFFSET;
+use super::tasks::TaskExitStatus;
 use super::{
     INITIAL_TASK_ID, KernelThreadStartInfo, Task, TaskListAdapter, TaskPointer, TaskRunListAdapter,
 };
@@ -440,10 +441,18 @@ fn current_task_terminated() {
     }
 }
 
-pub fn terminate() -> ! {
+/// Terminate the current task and optionally set its exit status.
+/// If no exit status is provided, then the task will terminate
+/// with the default value `Exited(0)`.
+pub fn terminate(exit_status: Option<TaskExitStatus>) -> ! {
     // Terminating a task will result in a task change, so preemption must
     // be allowable.
     preemption_checks();
+
+    if let Some(status) = exit_status {
+        current_task().set_exit_status(status);
+    }
+
     current_task_terminated();
 
     // The current task will not run again, so switch to a different task.
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -159,6 +159,40 @@ impl From<TaskState> for u32 {
     }
 }
 
+#[derive(PartialEq, Debug, Copy, Clone)]
+pub enum TaskExitStatus {
+    /// Task called sys_exit(code).
+    Exited(u16),
+    /// Task was killed by an x86 exception.
+    Exception,
+}
+
+impl From<u32> for TaskExitStatus {
+    fn from(v: u32) -> Self {
+        let (code, exception) = ((v >> 16) as u16, (v & 0xFFFF) as u16);
+        if exception != 0 {
+            Self::Exception
+        } else {
+            Self::Exited(code)
+        }
+    }
+}
+
+impl From<TaskExitStatus> for u32 {
+    fn from(s: TaskExitStatus) -> u32 {
+        match s {
+            TaskExitStatus::Exited(code) => (code as u32) << 16,
+            TaskExitStatus::Exception => 1,
+        }
+    }
+}
+
+impl Default for TaskExitStatus {
+    fn default() -> Self {
+        Self::Exited(0)
+    }
+}
+
 #[derive(Clone, Copy, Debug)]
 pub enum TaskError {
     // Attempt to close a non-terminated task
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -195,7 +195,7 @@ impl Default for TaskExitStatus {
 
 #[derive(Clone, Copy, Debug)]
 pub enum TaskError {
-    // Attempt to close a non-terminated task
+    // Attempt to close or to retrieve the exit status of a non-terminated task
     NotTerminated,
     // A closed task could not be removed from the task list
     CloseFailed,
@@ -355,6 +355,9 @@ pub struct Task {
 
     /// Queue of tasks waiting for completion of this task.
     wait_queue: SpinLockIrqSafe<WaitQueue>,
+
+    /// Exit status of the task
+    exit_status: AtomicU32,
 }
 
 // Expose the offsets of critical task fields to assembly.
@@ -529,6 +532,7 @@ impl Task {
             runlist_link: LinkedListAtomicLink::default(),
             objs: objtree,
             wait_queue: SpinLockIrqSafe::new(WaitQueue::new()),
+            exit_status: AtomicU32::new(TaskExitStatus::default().into()),
         }))
     }
 
@@ -646,6 +650,27 @@ impl Task {
 
     pub fn is_terminated(&self) -> bool {
         self.sched_state.get_state() == TaskState::TERMINATED
+    }
+
+    pub fn set_exit_status(&self, exit_status: TaskExitStatus) {
+        assert!(
+            !self.is_terminated(),
+            "The exit status can only be set for a task that is not yet terminated"
+        );
+        // Use Relaxed order since the exit status is set before the task
+        // is marked terminated with Release order
+        self.exit_status
+            .store(exit_status.into(), Ordering::Relaxed);
+    }
+
+    pub fn get_exit_status(&self) -> Result<TaskExitStatus, SvsmError> {
+        if self.is_terminated() {
+            // Use Relaxed order since the exit status is read after the task
+            // state is checked with Acquire order
+            Ok(self.exit_status.load(Ordering::Relaxed).into())
+        } else {
+            Err(TaskError::NotTerminated.into())
+        }
     }
 
     pub fn set_idle_task(&self) {
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -1131,7 +1131,7 @@ unsafe fn run_kernel_task<T: KernelThreadStartParameter>(
 }
 
 fn task_exit() {
-    terminate();
+    terminate(None);
 }
 
 #[cfg(all(test, test_in_svsm))]
--- a/kernel/src/task/mod.rs
+++ b/kernel/src/task/mod.rs
@@ -18,7 +18,7 @@ pub use schedule::{
 
 pub use tasks::{
     INITIAL_TASK_ID, KernelThreadStartInfo, TASK_FLAG_SHARE_PT, Task, TaskContext, TaskError,
-    TaskListAdapter, TaskPointer, TaskRunListAdapter, TaskState, is_task_fault,
+    TaskExitStatus, TaskListAdapter, TaskPointer, TaskRunListAdapter, TaskState, is_task_fault,
 };
 
 pub use exec::exec_user;
--- a/syscall/src/class0.rs
+++ b/syscall/src/class0.rs
@@ -8,6 +8,9 @@ use super::call::{SysCallError, syscall1, syscall3};
 use super::{SYS_EXEC, SYS_EXIT};
 use core::ffi::CStr;
 
+/// Terminate the current task with the given exit code. The kernel only
+/// supports exit codes on 16 bits, so the upper 16 bits of the provided
+/// code will be ignored.
 pub fn exit(code: u32) -> ! {
     // SAFETY: SYS_EXIT is supported syscall number by the svsm kernel.
     unsafe {


kernel/src/task/tasks.rs
```diff
@@ -157,9 +157,30 @@ impl From<TaskState> for u32 {
...
 
+fn compute_task_state(state: TaskState, exit_code: Option<u32>) -> u64 {
+    let state_val: u32 = state.into();
+    let shifted_state = (state_val as u64) << 32;
+    if let Some(code) = exit_code {
+        shifted_state | (code as u64)
+    } else {
+        shifted_state
+    }
+}
```
joergroedel: Why is the exit-code encoded together with the task state? I don't think there is a reason to update the exit_code atomically, no?

Just add another exit_code member with separate get/set methods. That also saves you from passing the exit-code around through multiple layers and extend function and method signatures for it.

stefano-garzarella: Yeah, agree, that was the question I asked yesterday during the GSoC call.

The only issue that comes to mind is the correlation between the state and the `exit_value`. I mean, `exit_value` is only valid in certain states, so should we handle the ordering with barriers or use `Option` (not sure is doable with atomics)?

joergroedel: The ideal solution would be to encode it in the enum as TERMINATED(u32), but that doesn't work because then the enum does not work with `#[repr(u32)]` anymore can not be stored in an atomic variable.

So given that the `exit_code` is only consumed on TERMINATED tasks I am fine with a less strict state tracking or using an `Option`. It is also fine to have it as an `u32` in the struct and return an `Option` in the getter (which returns `None` when not in terminated state). Something like that.

msft-jlange: Strictly speaking, the exit code has to be stored as `AtomicU32` because it is never possible to obtain a mutable reference to a task once it has been created, and interior mutability using locks is highly undesirable.  Using an atomic requires a solution where the exit code is stored as a bare value (not part of an enum) and thus the get function must use the current state to determine whether to return `None` or `Some(u32)`.

n-ramacciotti:
v2:
- Moved `exit_code` in an `AtomicU32`. 
- Implemented `get`, which returns a `Some` only when task state is `TERMINATED` .
- Implemented `set`, which can be called  to avoid passing `exit_code` to termination functions chain.

kernel/src/task/tasks.rs
```diff
@@ -608,6 +612,18 @@ impl Task {
...
+        self.exit_code.store(exit_code, Ordering::Relaxed);
+    }
+
+    pub fn get_exit_code(&self) -> Option<u32> {
+        if self.is_terminated() {
+            Some(self.exit_code.load(Ordering::Relaxed))
+        } else {
+            None
+        }
+    }
```
stefano-garzarella: IMO, this is very fragile. As you wrote in the commit description, a special ordering on setting/getting state and exit_code must be followed, and furthermore, there’s no guarantee that the exit code has actually been set.

If we want to support the case of termination without an exit_code, I think we need to introduce a new state. If, on the other hand, termination always involves an exit_code, then setting the state and the exit_code should be done in the same function, following the ordering you described in your comment to avoid future errors.

n-ramacciotti: >Just add another exit_code member with separate get/set methods. That also saves you from passing the exit-code around through multiple layers and extend function and method signatures for it.

My idea was to follow @joergroedel's suggestion to avoid passing the `exit_code` trough the "termination chain". However, as you said, in this context, I agree that it is necessary to do both set operations in the same function.

msft-jlange: I agree with @stefano-garzarella here: either the exit code is fundamentally a part of termination, in which case it must be set in all termination flows and is available for any terminated task, or the exit code is separate, in which case it can be set elsewhere but its availability must not be tied to task state.  If you want to keep it separate, it must be stored as an `Option` (or an equivalent data structure that can synchronize the state correctly to expose an `Option`-like value).  If you're going to store the exit code in the `Task` itself instead of the `TaskSchedState`, then you don't need an `AtomicU32` as long as you can support interior mutability in some way to enable the exit code to be set.  It would be reasonable to encode the exit code as a `SpinLock<Option<u32>>` (doesn't need to be IRQ-safe because it won't be subject to interrupt reentrancy and thus there's no requirement to prevent interrupt dispatch while the lock is held).

Part of making the decision here is a recognition that the core task object doesn't distinguish between kernel tasks and user tasks, so either an exit code must be fundamental to all types of tasks or it must be optional.  I think it's a little unnatural to expose exit codes from kernel tasks since many of them have no state to communicate out, and I suspect this may have been the basis for Joerg's concern about plubming exit codes through every termination path.  I believe your goal is to expose the exit code from a user task, and if that's the case, then storing it as an optional value would be cleanest.  If you design the user task interface such that all task termination will cause an exit code to be set, then you can guarantee that all terminated user tasks have an exit code and you can rely on that in the user API and make it appear non-optional there.

msft-jlange:
The TSC discussed the desired architecture for exit code handling and came to the following conclusions:
- All tasks should support exit codes, including both user tasks and kernel tasks.
- Exit codes should be required for user tasks.  This should be enforced in the syscalls that cause user task termination.
- Exit codes should not be required for kernel tasks.  If a kernel task wants to set its own exit code, it is free to do so with an explicit call.  If a kernel task does not set its own exit code, then the exit code will default to zero.
- Obtaining the exit code of a non-terminated task should fail.  In the case of the `get_exit_code()` kernel function, it is preferable to return a `Result<u32, SvsmError>` so that error information can be delivered if the error code is not available due to task state or any other reason.

n-ramacciotti:
v3:
- Introduced `terminated_with_exit_code`: can be used by both user and kernel tasks:
  - Another way could be to use an `Option` in `terminate`:  in user task `Some(exit_code)` is passed and it is setted in the termiation flow with `relaxed`. Kernel tasks will pass `None` to terminate and can use an explicit `set_exit_code` (with ~~`release`~~ `relaxed`) if they want an exit code which is not zero.
- Return `Result` in `get_exit_code`

luigix25:
IIUC this is not an RFC anymore, but in commit 2 you still mention an RFC, maybe a leftover?

n-ramacciotti:
> IIUC this is not an RFC anymore, but in commit 2 you still mention an RFC, maybe a leftover?

@luigix25 The PR is technically ready. I have one doubt about the exit code(s) that should be assigned for exception handling. I will remove it from the commit message and open a review where it can be discussed in more detail.

kernel/src/cpu/idt/svsm.rs
```diff
@@ -39,6 +39,8 @@ use core::slice;
 use crate::syscall::*;
 use syscall::*;
 
+const USER_EXCEPTION_EXIT_CODE: u32 = 1;
```
n-ramacciotti: I'm not sure what value(s) should be returned when terminating user tasks during exception.

kernel/src/task/tasks.rs
```diff
@@ -614,6 +618,18 @@ impl Task {
         self.sched_state.get_state() == TaskState::TERMINATED
     }
 
+    pub(super) fn set_exit_code(&self, exit_code: u32) {
+        self.exit_code.store(exit_code, Ordering::Relaxed);
+    }
```
stefano-garzarella: This code is still fragile IMO.

We should not introduce any new method to set the `exit_code`, but we should change `set_task_terminated()` to have it as parameter and set it with the right ordering required by `get_exit_code()`.

I mean something like this (untested, made with claude):
```diff
diff --git a/kernel/src/task/schedule.rs b/kernel/src/task/schedule.rs
index 883b08de..a1cd9a53 100644
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -280,12 +280,16 @@ impl TaskList {
     /// # Safety
     /// The caller must ensure that `task` is already a member of this task
     /// list.
-    unsafe fn terminate(&mut self, task: TaskPointer) -> Option<TaskPointer> {
+    unsafe fn terminate(
+        &mut self,
+        task: TaskPointer,
+        exit_code: u32,
+    ) -> Option<TaskPointer> {
         // Set the task state as terminated. If the task being terminated is the
         // current task then the task context will still need to be in scope until
         // the next schedule() has completed. Schedule will keep a reference to this
         // task until some time after the context switch.
-        let wakeup = task.set_task_terminated();
+        let wakeup = task.set_task_terminated(exit_code);
         // SAFETY: `task` must be a task pointer that is part of this list.
         let mut cursor = unsafe { self.list().cursor_mut_from_ptr(task.as_ref()) };
         cursor.remove();
@@ -420,7 +424,7 @@ pub fn wait_for_termination(task: TaskPointer) {
 /// # Panic
 ///
 /// This function must only be called after scheduling is initialized, otherwise it will panic.
-fn current_task_terminated() {
+fn current_task_terminated(exit_code: u32) {
     let cpu = this_cpu();
     let mut rq = cpu.runqueue_mut();
 
@@ -431,7 +435,7 @@ fn current_task_terminated() {
     // SAFETY: the scheduler guarantees that `current_task` always points to a
     // valid task, and every task has its pointer pushed into the global task
     // list during its creation.
-    let wakeup = unsafe { TASKLIST.lock().terminate(task_node.clone()) };
+    let wakeup = unsafe { TASKLIST.lock().terminate(task_node.clone(), exit_code) };
 
     // If another thread must be woken as a result of the termination, then
     // schedule it now.
@@ -444,22 +448,21 @@ fn current_task_terminated() {
 /// If this is called without passing by [`terminate_with_exit_code`]
 /// the task will terminate with the default value (0).
 pub fn terminate() -> ! {
+    terminate_with_exit_code(0)
+}
+
+/// Terminate the current task and set its exit code.
+pub fn terminate_with_exit_code(exit_code: u32) -> ! {
     // Terminating a task will result in a task change, so preemption must
     // be allowable.
     preemption_checks();
-    current_task_terminated();
+    current_task_terminated(exit_code);
 
     // The current task will not run again, so switch to a different task.
     select_new_task(false, None);
     unreachable!("terminated task rescheduled");
 }
 
-/// Terminate the current task and set its exit code.
-pub fn terminate_with_exit_code(exit_code: u32) -> ! {
-    current_task().set_exit_code(exit_code);
-    terminate()
-}
-
 pub fn go_idle() {
     // Entering an idle state will result in a task change, so preemption must
     // be allowable.
diff --git a/kernel/src/task/tasks.rs b/kernel/src/task/tasks.rs
index a227aa27..c13a0dd8 100644
--- a/kernel/src/task/tasks.rs
+++ b/kernel/src/task/tasks.rs
@@ -593,7 +593,8 @@ impl Task {
         self.sched_state.set_state(TaskState::RUNNING);
     }
 
-    pub fn set_task_terminated(&self) -> Option<TaskPointer> {
+    pub fn set_task_terminated(&self, exit_code: u32) -> Option<TaskPointer> {
+        self.exit_code.store(exit_code, Ordering::Relaxed);
         self.sched_state
             .panic_on_idle("Trying to terminate idle task")
             .set_state(TaskState::TERMINATED);
@@ -618,10 +619,6 @@ impl Task {
         self.sched_state.get_state() == TaskState::TERMINATED
     }
 
-    pub(super) fn set_exit_code(&self, exit_code: u32) {
-        self.exit_code.store(exit_code, Ordering::Relaxed);
-    }
-
     pub fn get_exit_code(&self) -> Result<u32, SvsmError> {
         if self.is_terminated() {
             Ok(self.exit_code.load(Ordering::Relaxed))
```

kernel/src/task/tasks.rs
```diff
@@ -614,6 +618,18 @@ impl Task {
...
+
+    pub fn get_exit_code(&self) -> Result<u32, SvsmError> {
+        if self.is_terminated() {
+            Ok(self.exit_code.load(Ordering::Relaxed))
```
stefano-garzarella: here, and where we set it we should add a comment like you did in the commit description, to explain why `Ordering::Relaxed` is fine and also that the barrier is done by set_state/get_state

kernel/src/task/tasks.rs
msft-jlange: > We should not introduce any new method to set the `exit_code`, but we should change `set_task_terminated()` to have it as parameter and set it with the right ordering required by `get_exit_code()`.

I disagree with this point of view.  We said that the exit code should be optional.  Passing it as a parameter to `set_task_terminated()` makes it a required parameter, because it means that all termination paths are now required to supply a value.  We said that making the exit code optional would also result in a default value of zero, and while that could be achieved by requiring a parameter to `set_task_terminated()` and having callers pass zero if they have no better idea, it forces every single termination point to make a decision about what the correct default value is in that circumstance.  It is much more robust to initialize the value to zero and to have an explicit method to set the value to a specific value when the caller knows that a specific value is desired.

kernel/src/cpu/idt/svsm.rs
stefano-garzarella: Oh, I didn't think about that. IMO, this is not right; there should not be any overlap between exceptions and user-set values.

Also, I think it will be useful when we will implement `sys_wait()` (or something similar) to determine whether the task terminated correctly (and the code set by `sys_exit()`) or due to an exception. So maybe we should either have another field like `exit_reason` or use an enum like this to encode both:
```
pub enum TaskExitStatus {
    /// Task called sys_exit(code).
    Exited(u32),
    /// Task was killed by an x86 exception with the given vector number.
    Exception(u8),
}
```

Then we can put it in an SpinLock similar to what @msft-jlange suggested in https://github.com/coconut-svsm/svsm/pull/1128#discussion_r3446059461

kernel/src/task/tasks.rs
stefano-garzarella: `terminate()` is still there without any parameter and kernel tasks can use it without specifying any exit code.

kernel/src/cpu/idt/svsm.rs
msft-jlange: Long-term, we said this would be related to the signal that is raised by the exception, but we haven't implemented signal delivery yet.  We should define a user signal exit value (probably much greater than `1`) and that definition should be in a file that is shared between user mode and kernel mode, and that will enable us to build on top of that definition in the future once signal delivery is implemented.

msft-jlange: > So maybe we should either have another field like `exit_reason` or use an enum like this to encode both:

An enum is a nice idea but whatever we define has to be something that can be represented as a `u32` and the enum you propose cannot be.  We could consider inflating the exit code to a `u64` and then implement `From` to convert between that `u64` and an enum, but the primitive type must be something that can be represented as an atomic in order to achieve the mutability requirements.

kernel/src/task/tasks.rs
stefano-garzarella: > it forces every single termination point to make a decision about what the correct default value is in that circumstance.

I don't get this TBH. All termination points actually pass through `terminate()` which is still there. @n-ramacciotti added `terminate_with_exit_code()` to support terminating with an exit code different from 0. My proposal doesn't change the API exposed to the tasks, just change the underlying API, to be more robust, since `get_exit_code()` imply a specific ordering.

kernel/src/cpu/idt/svsm.rs
stefano-garzarella: @msft-jlange Can we just use `SpinLock<TaskExitStatus>` as you proposed in https://github.com/coconut-svsm/svsm/pull/1128#discussion_r3446059461 for the `Option<u32>` ?

msft-jlange: > @msft-jlange Can we just use `SpinLock<TaskExitStatus>` as you proposed in [#1128 (comment)](https://github.com/coconut-svsm/svsm/pull/1128#discussion_r3446059461) for the `Option<u32>` ?

In our earlier discussion, we said that we didn't want to store the exit code as an `Option` because every task would have an exit code.  We said that we would instead define a default exit code of zero to be used if no exit code was set explicitly.  Using an `Option` takes us away from that path because it implies that we will encode `None` instead of zero if no exit code was set.

As I think about it further, I think the idea of an enum is inconsistent with the idea that the default exit code is zero, because zero isn't an enumerated type.  We could conceivably redefine the default value to something that means there is no exit code, but that's really the same as defining it to be `Option::None` and now we're back to the `Option` definition that we said we didn't want.  So I believe the only design that is consistent with our prior discussion is an integer type for which zero is the default.  If you don't agree, then we need another TSC discussion to rehash what we previously discussed.

kernel/src/task/tasks.rs
msft-jlange: > > it forces every single termination point to make a decision about what the correct default value is in that circumstance.
> I don't get this TBH. All termination points actually pass through `terminate()` which is still there.

I think what you're saying is that my claim may be true - that all termination points must make a decision - but that in the current construction of the code, there is only one such point (`terminate()`) and therefore there is no undue burden because there's only one place to maintain.  That might be true, but I still find awkward because we are pushing the definition of "default" away from the construction of the task and into the termination path.  I'd like to understand why you think this is a better approach.  What problem is solved?  Are you simply worried about guaranteeing that there is no potential to set the exit code in a way that is not synchronized with task termination?  I'm not convinced that is a guarantee that is necessary, especially given the fact that the exit code is optional and is therefore simply a usage convention and therefore not necessarily a robust construction.  In fact, I could argue that no synchronization whatsoever is necessary for exit code handling - we don't even need to guarantee that it can only be read from a terminated task - because the convention used by all consumers will be that they only care about an exit code for a terminated task and would never even bother to query it if the task is not terminated.

kernel/src/cpu/idt/svsm.rs
stefano-garzarella: We can safely set the default for `TaskExitStatus` to `Exited(0)` and set it when we create the task or handle it in the API exposed to tasks (see `terminate()` in the other thread https://github.com/coconut-svsm/svsm/pull/1128#discussion_r3482155762).

Honestly, I don’t understand the connection between a default value and what we use to store it. I agree that we should provide a default and don't ask kernel tasks to set an exit value, but I really don't see how this is related on how we store that value. Maybe I'm missing something. Feel free to clarify that here or in the TSC if you prefer.

msft-jlange: > We can safely set the default for `TaskExitStatus` to `Exited(0)` and set it when we create the task or handle it in the API exposed to tasks (see `terminate()` in the other thread [#1128 (comment)](https://github.com/coconut-svsm/svsm/pull/1128#discussion_r3482155762)).
> 
> Honestly, I don’t understand the connection between a default value and what we use to store it. I agree that we should provide a default and don't ask kernel tasks to set an exit value, but I really don't see how this is related on how we store that value. Maybe I'm missing something. Feel free to clarify that here or in the TSC if you prefer.

In our discussion we agreed that the default value should be zero, not `Exited(0)`.  We also discussed the handling of exception termination and said that we should follow the Linux model, which adds a constant value to the signal number (I think the constant is 0x80 but I could be wrong), not that we would introduce a new enumerated type to distinguish exception termination from graceful termination.  So the enum proposal here is a new idea.

Our discussions were also based on the idea that the exit code must be encoded as an integer type so that we could interior mutability for free based on atomics.

Introducing an enumerated type will require reconsideration of the definition of zero as the default (because zero and `Exited(0)` are not the same value) as well as reconsideration of whether to use a spin lock instead of an atomic to achieve interior mutability on the task object.  We can certainly have that discussion in the TSC if you want.

kernel/src/task/tasks.rs
stefano-garzarella: > I think what you're saying is that my claim may be true - that all termination points must make a decision - but that in the current construction of the code, there is only one such point (`terminate()`) and therefore there is no undue burden because there's only one place to maintain. That might be true, but I still find awkward because we are pushing the definition of "default" away from the construction of the task and into the termination path. I'd like to understand why you think this is a better approach.

That was the point I was missing. Now I understand what you mean. Sorry, I hadn't quite realized that you meant you wanted the default to be specified when the task was created, not when it was terminated.

> What problem is solved? Are you simply worried about guaranteeing that there is no potential to set the exit code in a way that is not synchronized with task termination? 

Yep, exactly that one, especially after reading commit https://github.com/coconut-svsm/svsm/pull/1128/changes/f0c6ec0c012691e025b3a5b7dbc88e6fc2b6c171 description:

```
For the atomic context, use `Ordering::Relaxed` for both methods as: `set`
will be executed before setting the state (with `Ordering::Release`) and
`get` will be executed after checking the state (with `Ordering:Acquire`).
```

> I'm not convinced that is a guarantee that is necessary, especially given the fact that the exit code is optional and is therefore simply a usage convention and therefore not necessarily a robust construction. In fact, I could argue that no synchronization whatsoever is necessary for exit code handling - we don't even need to guarantee that it can only be read from a terminated task - because the convention used by all consumers will be that they only care about an exit code for a terminated task and would never even bother to query it if the task is not terminated.

Okay, now I understand. Since a task can't do anything else after calling `terminate()`, it follows that the `exit_code` can only be set before the state changes. Got it.

kernel/src/cpu/idt/svsm.rs
stefano-garzarella: > In our discussion we agreed that the default value should be zero, not `Exited(0)`. We also discussed the handling of exception termination and said that we should follow the Linux model, which adds a constant value to the signal number (I think the constant is 0x80 but I could be wrong), not that we would introduce a new enumerated type to distinguish exception termination from graceful termination. So the enum proposal here is a new idea.

Yes, I remember that, but I thought it was more about the user interface, not how we store it internally in the kernel.

> Our discussions were also based on the idea that the exit code must be encoded as an integer type so that we could interior mutability for free based on atomics.

Yep, and I agreed on that, but now the exceptions is chaning my mind.

> Introducing an enumerated type will require reconsideration of the definition of zero as the default (because zero and `Exited(0)` are not the same value) as well as reconsideration of whether to use a spin lock instead of an atomic to achieve interior mutability on the task object. We can certainly have that discussion in the TSC if you want.

Yes, I think it’s best to clarify how to handle exceptions in the next TSC. I agree that implementing signals, etc., can come later, but IMO, retrieving an exit code from a task that terminated with an exception should fail, just as it would when the task hasn’t yet terminated (obviously with a different error).

Essentially, I only realized this while doing this review (my fault for not looking closely at the code before our discussion) that we currently terminate a user task when there’s an exception with the same TERMINATED state. I'm not saying we have to change this, but IMO we should record that the task ended with an exception in some way: by adding another field, extending `exit_code` to u64, using an enum, etc. So, I think `exit_code: AtomicU32` alone isn't enough.

stefano-garzarella: @n-ramacciotti we discussed this today in the TSC and we agreed on keep it simple for now and extend later eventually.

What we can do in case of exception is to follow Linux. From `_exit(2)` man page:
```
The value status & 0xFF is returned to the parent process as the process's exit status,  and  can
be collected by the parent using one of the wait(2) family of calls.
```

So we should follow it and leave the first 8 bits for user codes.

n-ramacciotti: >What we can do in case of exception is to follow Linux. From _exit(2) man page:
> ...
> So we should follow it and leave the first 8 bits for user codes.

I'm not sure I understand the man page correctly. During normal exits, Linux does `(code & 0xff) << 8`, while for exception, IIUC, it leaves the value as it is. So, following the Linux approach, the user code shouldn't be stored in the first 8 bits.

msft-jlange: > I'm not sure I understand the man page correctly. During normal exits, Linux does `(code & 0xff) << 8`, while for exception, IIUC, it leaves the value as it is. So, following the Linux approach, the user code shouldn't be stored in the first 8 bits.

What you want to do is to mimic the behavior of `waitpid()`, and the more I think about it, the more I think we want to do that with a Rust enum for better ergonomics.  That allows us to skip macros like `WIFSIGNALED` and so on.  So I'm now starting to lean towards an enum that can encode either a 16-bit exit code or a 16-bit signal code, and implements `From` to convert in both directions between the enum and a `u32`.  That will allow us to store the enum in an `AtomicU32` and will enable us to deliver the value as an enum when it is read.

Linux uses a 16-bit value with 8-bit signal and 8-bit exit codes, but since we're already on the path to use a 32-bit value to store the encoded value, we might as well take advantage of the space.

kernel/src/task/tasks.rs
msft-jlange: There is no barrier required at all here.  We are using an atomic just because it is an ergonomic way to achieve interior mutability, and when an atomic is used this way, specific memory ordering is never required.  That's why `Ordering::Relaxed` is appropriate.

stefano-garzarella: I didn't ask to put any barrier, I asked to add here a comment to explain why Relaxed is fine, like we have in the commit message.

kernel/src/cpu/idt/svsm.rs
stefano-garzarella: > What you want to do is to mimic the behavior of `waitpid()`, and the more I think about it, the more I think we want to do that with a Rust enum for better ergonomics.

Yes, the enum would be great, especially for use in the kernel. I still slightly prefer saving the entire enum and dealing with the conversion later, when we will implement the `wait*()` syscalls within their handlers. But I understand that we've already decided to save an `AtomicU32`, so the enum and `From` are fine with me too.

n-ramacciotti:
v4:
- Introduced `TaskExitStatus` enum with `from` u32 trait implementation in both directions:
  - The conversion follows Linux style, (exit_code | exception), where if exception is not zero the result is an Exception(code).
- Added comments on relaxed ordering

kernel/src/task/schedule.rs
```diff
@@ -451,6 +455,17 @@ pub fn terminate() -> ! {
...
+pub fn terminate_with_exit_code(exit_code: u16, is_exception: bool) -> ! {
+    let exit_status = if is_exception {
+        TaskExitStatus::Exception(exit_code)
+    } else {
+        TaskExitStatus::Exited(exit_code)
+    };
+    current_task().set_exit_status(exit_status);
+    terminate()
+}
+
```
stefano-garzarella: Why not creating `TaskExitStatus` in the caller and pass it to this function?

I mean something like `terminate_with(TaskExitStatus::Exited(exit_code))` on exit or  `terminate_with(TaskExitStatus::Exception(exception_code))` on exceptions/signals

The point of the enum IMO was exactly that.

n-ramacciotti: Yeah, right!

n-ramacciotti:
v5:
- Addressed @stefano-garzarella's comment:
  - changed `terminated_with_exit_code` to `terminate_with_exit_status` with `TaskExitStatus` directly.
- Changed `exit_code` to `exit_status` in all commits.

kernel/src/task/tasks.rs
```diff
@@ -159,6 +159,34 @@ impl From<TaskState> for u32 {
     }
 }
 
+#[derive(Debug)]
+pub enum TaskExitStatus {
```
stefano-garzarella, line 163: I'm a bit confused.

If we use `TaskExitStatus::Exception(0)` and store in a u32, IIUC this will be converted to `TaskExitStatus::Exited(0)`. IMO we should prevent this.

What about using `Exception(NonZeroU16)` ?

Otherwise we should reserve one of the bit for the exception to be set to 1.

kernel/src/task/tasks.rs
```diff
@@ -159,6 +159,34 @@ impl From<TaskState> for u32 {
...
+    /// Task called sys_exit(code).
+    Exited(u16),
+    /// Task was killed by an x86 exception with the given vector number.
```
stefano-garzarella: `with the given vector number` this is not really what we are doing.

kernel/src/task/tasks.rs
```diff
@@ -642,6 +646,25 @@ impl Task {
         self.sched_state.get_state() == TaskState::TERMINATED
     }
 
+    pub fn set_exit_status(&self, exit_status: TaskExitStatus) {
+        if !self.is_terminated() {
```
stefano-garzarella: Would be better to add an assert or debug_assert here?

kernel/src/task/tasks.rs
```diff
@@ -523,6 +526,7 @@ impl Task {
             runlist_link: LinkedListAtomicLink::default(),
             objs: objtree,
             wait_queue: SpinLockIrqSafe::new(WaitQueue::new()),
+            exit_status: AtomicU32::new(0),
```
stefano-garzarella: What about using `TaskExitStatus::Exited(0)` here. Or maybe we can even implement the Default trait for `TaskExitStatus` and use it here.

kernel/src/syscall/class0.rs
```diff
@@ -9,17 +9,19 @@ use crate::address::VirtAddr;
...
 use crate::mm::guestmem::UserPtr;
+use crate::task::TaskExitStatus;
 use crate::task::exec_user;
-use crate::task::terminate;
+use crate::task::terminate_with_exit_status;
 use core::ffi::c_char;
 use syscall::SysCallError;
 
 pub fn sys_exit(exit_code: u32) -> ! {
+    let code = (exit_code & 0xffff) as u16;
```
stefano-garzarella, line 19: This should be documented somewhere, maybe in `syscall/src/class0.rs`

kernel/src/cpu/idt/svsm.rs
```diff
@@ -194,7 +196,7 @@ extern "C" fn ex_handler_terminate(ctx: &mut X86ExceptionContext, vector: usize)
             vector,
             { ctx.frame.rip }
         );
-        terminate();
+        terminate_with_exit_status(TaskExitStatus::Exception(USER_EXCEPTION_EXIT_CODE));
```
stefano-garzarella: If we have to use the same value in all cases, I'd say for now we might skip setting a value and just have `TaskExitStatus::Exception`, always encoding it as 1 << 16.

We can add the encoded value later, once we know what to do with it.

stefano-garzarella: @n-ramacciotti I'm not sure about my suggestion :-) 
@joergroedel @msft-jlange do you have any better ideas?

kernel/src/task/tasks.rs
n-ramacciotti: Yeah, right. I copy-pasted your suggestion and didn't change the comment. I will remove that part for now

kernel/src/task/tasks.rs
n-ramacciotti: What about using an additional enum, something like:
```
TaskExitStatus::Exception(ExceptionCode::CodeX)
```
Where we can force `CodeX` to be different from 0?

kernel/src/cpu/idt/svsm.rs
msft-jlange: We haven't defined what exceptions codes are yet.  For now, we can define `TaskStatus::Exception` as a standalone enum with no embedded integer.  When the time comes that we define exception codes, we can expand the `TaskStatus` enum to include the values that make sense at the time.

Stefano's statement
> We can add the encoded value later, once we know what to do with it.

is exactly right.  We shouldn't use this PR to define unrelated concepts that aren't currently defined.

kernel/src/task/tasks.rs
msft-jlange: We haven't defined what exception codes are.  It's better just to leave this as `Exception` until we know what we are doing.

kernel/src/task/schedule.rs
```diff
@@ -451,6 +455,12 @@ pub fn terminate() -> ! {
     unreachable!("terminated task rescheduled");
 }
 
+/// Terminate the current task and set its exit status.
+pub fn terminate_with_exit_status(exit_status: TaskExitStatus) -> ! {
```
msft-jlange: It would be better to redefine the existing `terminate()` function to take `Option<TaskExitStatus>`.  The whole point of `Option` is to be able to avoid having to define two different code paths to describe the presence and absence of some concept - this is exactly what `Some` and `None` are for.

n-ramacciotti:
v6:
- Removed embedded code in `Exception` variant. Now it's always encoded as 1.
- Introduced `Option` in `terminate`
- Added `assertion` in `set_exit_status`
- Added `Default` implementation for `TaskExitStatus`
- Added comment in `syscall/src/class0.rs`

kernel/src/task/tasks.rs
```diff
@@ -159,9 +159,43 @@ impl From<TaskState> for u32 {
     }
 }
 
+#[derive(Debug)]
```
stefano-garzarella: Maybe we can derive also `Clone` and `Copy` like we do for the other enums here.

Not sure if derive also `PartialEq` since it can be useful later when we will use get_exit_status() to check exit status, but we can do it later if you prefer.

stefano-garzarella (dismissed):
LGTM! @n-ramacciotti thanks, I left just a minor comment.

n-ramacciotti:
v7:
- derived `PartialEq`, `Clone` and `Copy` for `TaskExitStatus` as suggested by @stefano-garzarella

stefano-garzarella (approved):

msft-jlange (approved):

joergroedel (approved):

PR #1129, opened 2026-06-17
task: introduce `PENDING` state

msft-jlange:
Define a new pending state to describe a task that has been created but never inserted into any run queue.  This ensures that if the task object is dropped before it is ever run, and thus never goes through a termination flow, there will not be an assertion failure due to the fact that the task was not explicitly terminated.

### DIFF
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -46,9 +46,13 @@ use crate::sev::ghcb::{GHCB, GhcbPage};
 use crate::sev::hv_doorbell::{HVDoorbell, allocate_hv_doorbell_page};
 use crate::sev::utils::RMPFlags;
 use crate::sev::vmsa::{VMSAControl, VmsaPage};
-use crate::task::{
-    KernelThreadStartInfo, RunQueue, Task, TaskPointer, schedule, schedule_task, scheduler_idle,
-};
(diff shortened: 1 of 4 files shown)


joergroedel:
On KVM SMP-guest (SEV-SNP) boots this fails with:

```
[SVSM] ERROR: Panic on CPU[1]! COCONUT-SVSM Version: 2026.05-devel-95-gf54eedbc
[SVSM] ERROR: Info: panicked at kernel/src/task/schedule.rs:122:9:
assertion failed: task.is_running()
[SVSM] ---BACKTRACE---:
[SVSM]   [ffffff800005093b] svsm::task::schedule::RunQueue::enqueue_task+0x9b
[SVSM]   [ffffff80000529a6] svsm::task::schedule::schedule_task+0x36
[SVSM]   [ffffff800004c7a3] svsm::cpu::percpu::cpu_idle_loop+0xb3
[SVSM]   [ffffff800007d540] svsm::task::tasks::task_exit+0x0
[SVSM] ---END---
```

The firmware has already started to run when this happens.

msft-jlange:
> On KVM SMP-guest (SEV-SNP) boots this fails with:

I believe I've fixed this with my latest push.

joergroedel (approved):

PR #1130, opened 2026-06-17
kernel: protocols/attest: spec-align `SVSM_ATTEST_SERVICES` request

00xc:
The handling of `SVSM_ATTEST_SERVICES` is not fully compliant to the SVSM spec on several counts:
    
1. The returned error for an insufficiently large guest-provided buffer is incorrect, as the spec states that it should be `SVSM_ERR_INVALID_PARAMETER`; this is also what the Linux driver checks for. The current code returns a protocol-specific error.
2. The output registers are only written on success, but they should be written on the error path as well so that the lower-privilege guest can re-allocate its buffers to the required size. The Linux guest driver also relies on this behavior if the aforementioned error is returned.
    
Fix this by unifying the extended and non-extended attestation paths, always setting the output registers, and returning `SVSM_ERR_INVALID_PARAMETER` as appropriate.

The first commit fixes the core issue. Commits 2 and 3 bring in additional cleanups.

### DIFF
--- a/kernel/src/error.rs
+++ b/kernel/src/error.rs
@@ -61,25 +61,6 @@ pub enum ApicError {
     Registration,
 }
 
-/// Errors related to Attestation handling. These may originate from multiple
-/// layers in the system. Added to protocol base for protocol error.
-#[repr(u64)]
-#[derive(Clone, Copy, Debug)]
-pub enum AttestError {
-    /// An error related to attestation report.
-    Report = 0,
-
-    /// An error related to attestation manifest.
-    Manifest = 1,
-
-    /// An error related to attestation certificates.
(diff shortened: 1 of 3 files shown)


stefano-garzarella:
@IT302 if you have time, can you take a look at this PR ?

stefano-garzarella (commented):
LGTM, I'd like to wait a bit if @IT302 has any comment since he touched this code recently.

stefano-garzarella (approved):
@IT302 I'm going to merge this, if you find something to improve, we can open a PR on top.

IT302:
Looks good to me. Apologies for the delayed comment—I missed the mentions earlier.

PR #1133, opened 2026-06-18
kernel: sev/hv_doorbell: treat NoEoiRequired as a single field

00xc:
The #HV doorbell page has a byte-sized field to determine whether an an EOI is required to service an interrupt. The GHCB spec says:

    When it is possible to perform an EOI without exiting the guest,
    NoEoiRequired will be set to a non-zero value by the host,
    indicating that no explicit EOI is required. When the guest
    wishes to perform an EOI, it should attempt to atomically change
    NoEoiRequired from non-zero to zero.

However, the current code in the SVSM only looks at bit 0 in the field, and performs a CAS loop to only update this bit, which does not match what the spec says.

This is the pseudocode provided in the GHCB spec:

    NoEoiRequired := XCHG(Doorbell.NoEoiRequired, 0)
    IF NoEoiRequired = 0
        WRMSR(X2APIC_EOI)
    FI

Any reasonable hypervisor will write bit 0, but do not rely on this behavior and simplify the code. Swap the whole field to 0 and treat any non-zero value as "no EOI required".

### DIFF
--- a/kernel/src/sev/hv_doorbell.rs
+++ b/kernel/src/sev/hv_doorbell.rs
@@ -156,28 +156,9 @@ impl HVDoorbell {
 
     pub fn no_eoi_required(&self) -> bool {
         // Check to see if the "no EOI required" flag is set to determine
-        // whether an explicit EOI can be avoided.
-        let mut no_eoi_required = self.no_eoi_required.load(Ordering::Relaxed);
-        loop {
-            // If the flag is not set, then an explicit EOI is required.
-            if (no_eoi_required & 1) == 0 {
-                return false;
-            }
-            // Attempt to atomically clear the flag.
-            match self.no_eoi_required.compare_exchange_weak(
-                no_eoi_required,
-                no_eoi_required & !1,
-                Ordering::Relaxed,
-                Ordering::Relaxed,
-            ) {
-                Ok(_) => break,
-                Err(new) => no_eoi_required = new,
-            }
-        }
-
-        // If the flag was successfully cleared, then no explicit EOI is
-        // required.
-        true
+        // whether an explicit EOI can be avoided. If the flag was successfully
+        // cleared, then no explicit EOI is required, and required otherwise.
+        self.no_eoi_required.swap(0, Ordering::Relaxed) != 0
     }
 }
 


msft-jlange:
The intention of the specification is that `NoEoiRequired` is defined as a single bit and the remainder of that byte is reserved for future use.  The pseudocode is certainly vague on this point but the specific notation of the XCHG is intended to convey that only the bit itself is to be exchanged, not the other bits that happen to be around it.  Keeping the SVSM behavior as it exists today will always work correctly, at last as long as the SVSM only ever examines the low bit instead of the entire byte (if not, that's a bug).  Changing the SVSM behavior to write an entire byte would interfere with other possible definitions that may arise in the future.

00xc:
> The intention of the specification is that `NoEoiRequired` is defined as a single bit and the remainder of that byte is reserved for future use.

The spec quite literally defines `NoEoiRequired` as a full byte:

> PendingEvent is a 16-bit field in bytes 0..1 of the page, and NoEoiRequired is byte 2 of the page. Bytes 3..63 are reserved to accommodate future expansion of required functionality.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -155,29 +155,7 @@ impl HVDoorbell {
     }
 
     pub fn no_eoi_required(&self) -> bool {
-        // Check to see if the "no EOI required" flag is set to determine
-        // whether an explicit EOI can be avoided.
```
msft-jlange, line 159: Can we keep this comment for clarity?

kernel/src/sev/hv_doorbell.rs
```diff
@@ -155,29 +155,7 @@ impl HVDoorbell {
...
-                Ordering::Relaxed,
-                Ordering::Relaxed,
-            ) {
-                Ok(_) => break,
-                Err(new) => no_eoi_required = new,
-            }
-        }
-
-        // If the flag was successfully cleared, then no explicit EOI is
-        // required.
```
msft-jlange, line 179: Can we keep this comment as well?  It can be combined with the previous comment.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -155,29 +155,7 @@ impl HVDoorbell {
     }
 
     pub fn no_eoi_required(&self) -> bool {
-        // Check to see if the "no EOI required" flag is set to determine
-        // whether an explicit EOI can be avoided.
-        let mut no_eoi_required = self.no_eoi_required.load(Ordering::Relaxed);
-        loop {
-            // If the flag is not set, then an explicit EOI is required.
```
msft-jlange, line 162: Can we keep this comment as well?  It can be combined with the previous comment.

msft-jlange:
> The spec quite literally defines `NoEoiRequired` as a full byte:

I discussed this with AMD and there is a spec clarification coming.  It will continue to define `NoEoiRequired` as a full byte but will state that the only legal values are zero and one.  This is logically equivalent to saying that `NoEoiRequired` is a single bit and the remaining bits are MBZ.  The effect is that both bitwise and bytewise CAS operations will correctly interpret and update the `NoEoiRequired` state, so guests are free to take either approach.  As such, your proposed change is perfectly fine but also not strictly necessary.  I agree that your version of the code is much simpler so I'm fine merging it but I don't believe it's strictly necessary.

msft-jlange (approved):

joergroedel (approved):

PR #1135, opened 2026-06-19
kernel/protocols: add vtpm to core query protocol

n-ramacciotti:
The query protocol is used to check the availability of a protocol at a given version.

It should take into account every available protocol, but currently it does not handle the vtpm protocol. Add vtpm protocol management.

### DIFF
--- a/kernel/src/protocols/core.rs
+++ b/kernel/src/protocols/core.rs
@@ -22,6 +22,11 @@ use crate::protocols::uefivars::{UEFI_MM_PROTOCOL_VERSION_MAX, UEFI_MM_PROTOCOL_
 use crate::protocols::{
     RequestParams, SVSM_APIC_PROTOCOL, SVSM_ATTEST_PROTOCOL, SVSM_CORE_PROTOCOL,
 };
+#[cfg(all(feature = "vtpm", not(test)))]
+use crate::protocols::{
+    SVSM_VTPM_PROTOCOL,
+    vtpm::{VTPM_PROTOCOL_VERSION_MAX, VTPM_PROTOCOL_VERSION_MIN},
+};
 use crate::requests::SvsmCaa;
 use crate::sev::utils::{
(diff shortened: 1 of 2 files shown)


stefano-garzarella (dismissed):

kernel/src/protocols/core.rs
```diff
@@ -14,11 +14,15 @@ use crate::mm::{GuestPtr, valid_phys_region, writable_phys_addr};
...
 use crate::protocols::SVSM_UEFI_MM_PROTOCOL;
+#[cfg(all(feature = "vtpm", not(test)))]
+use crate::protocols::SVSM_VTPM_PROTOCOL;
 use crate::protocols::apic::{APIC_PROTOCOL_VERSION_MAX, APIC_PROTOCOL_VERSION_MIN};
 use crate::protocols::attest::{ATTEST_PROTOCOL_VERSION_MAX, ATTEST_PROTOCOL_VERSION_MIN};
 use crate::protocols::errors::SvsmReqError;
 #[cfg(all(feature = "uefivars", not(test)))]
 use crate::protocols::uefivars::{UEFI_MM_PROTOCOL_VERSION_MAX, UEFI_MM_PROTOCOL_VERSION_MIN};
+#[cfg(all(feature = "vtpm", not(test)))]
+use crate::protocols::vtpm::{VTPM_PROTOCOL_VERSION_MAX, VTPM_PROTOCOL_VERSION_MIN};
```
00xc: You can just do:

```suggestion
#[cfg(all(feature = "vtpm", not(test)))]
use crate::protocols::{SVSM_VTPM_PROTOCOL, vtpm::{VTPM_PROTOCOL_VERSION_MAX, VTPM_PROTOCOL_VERSION_MIN}};
```

And avoid the two conditional includes. Arguably the uefivars protocol should do the same but that's outside the scope of this PR.

n-ramacciotti: Right, thanks!

n-ramacciotti:
v2:
- Addressed @00xc's comment

00xc (approved):

PR #1136, opened 2026-06-22
kernel: zeroize more secrets

### LINKED ISSUE #807: Attestation: Zeroize secrets properly after use

00xc:
* Zeroize VMPCK after use.
* Enable `zeroize` feature for `aes-gcm` crate
* Zeroize KEK and CEK after use during attestation secret retrieval
* Update `aes-kw` and enable `zeroize` feature

Closes: #807

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,6 +4,7 @@
 rustflags = [
 	"-C", "force-frame-pointers",
 	"--cfg", "aes_force_soft",
+	"--cfg", 'aes_backend="soft"',
 	"--cfg", "polyval_force_soft",
 ]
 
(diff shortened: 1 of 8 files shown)


00xc:
I think this addresses all the remaining points for #807 @osteffenrh @stefano-garzarella

stefano-garzarella:
> I think this addresses all the remaining points for #807 @osteffenrh @stefano-garzarella

I think so, so I added `Closes` tag in the PR descrition. Thanks for this!

stefano-garzarella (approved):

PR #1137, opened 2026-06-22
kernel: migrate CPUID feature detection to cpufeature crate

tanish111:
SVSM maintained hand-written CPUID checks, with each CPU feature requiring custom leaf/register/bit logic. This was repetitive and error-prone as new features were added. Replace it with the shared, auto-generated cpufeature crate to provide a single, consistent source of CPU feature detection.

cpufeature crate, generated from the X86-CPUID
database, as the single source of truth for standard feature
descriptors. Call sites pass cpufeature::leaves constants directly
to cpu_has_feat() and cpu_get_feat() instead of a local Feature
enum, so leaf numbers, registers, and bit positions are no longer
hand-maintained in the kernel.

Platform CPUID goes through CpuidBackend: native and TDP use the
default backend; SNP routes architectural leaves through the CPUID
table and hypervisor leaves through GHCB. platform::cpuid() takes a
CpuidFeature descriptor, so feature checks and platform dispatch
share one path. features.rs keeps only leaves outside the database
(HYPERV_INTERFACE, PHYS_ADDR_SIZES).

This removes about 110 lines of custom detection code, eliminates
the macro-generatedd lookup table, and simplifies future feature
additions by relying on external cpufeature crate instead
of extending another local table.

cpufeature crate:- https://github.com/coconut-svsm/cpufeature

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,8 @@ virtio-drivers = { git = "https://github.com/rcore-os/virtio-drivers.git", rev =
 zerocopy = { version = "0.8.2", features = ["derive"] }
 zeroize = { version = "1.8.2", default-features = false }
 
+cpufeature = { git = "https://github.com/coconut-svsm/cpufeature.git", rev = "0182d61954372552fe2a40477340e949471a49ee" }
+
 # Verus repos
 verus_builtin = { version = "=0.0.0-2026-04-12-0118", default-features = false }
 verus_builtin_macros = { version = "=0.0.0-2026-04-12-0118", features = ["vpanic"], default-features = false }
--- a/kernel/src/cpu/features.rs
+++ b/kernel/src/cpu/features.rs
@@ -5,18 +5,49 @@
 // Authors: Joerg Roedel <jroedel@suse.de>
 //          Carlos López <clopez@suse.de>
 
-use core::arch::x86_64::CpuidResult;
+use crate::platform::cpuid;
+use crate::utils::immut_after_init::ImmutAfterInitCell;
+use cpufeature::CpuidFeature;
+use cpufeature::CpuidRegister;
+use cpufeature::leaves::{
+    CET_SS, INVLPGB_MAX_PAGES, PTE_CBIT_POS, X86_FEATURE_PGE, X86_FEATURE_SMAP, X86_FEATURE_SMEP,
+    X86_FEATURE_UMIP, X86_FEATURE_X2APIC, X86_FEATURE_XMM, X86_FEATURE_XSAVE, X86_FEATURE_XSAVEOPT,
+    XCR0_AVX, XCR0_SSE, XCR0_X87,
+};
 
-use crate::{platform::cpuid, utils::immut_after_init::ImmutAfterInitCell};
+/// CPUID leaf 0 — CPU vendor ID string (EBX/ECX/EDX).
+///
+/// Not exposed under that name in the cpufeature database (leaf 0 is
+/// `MAX_STD_LEAF` on EAX), so keep a local descriptor for vendor detection.
+pub const CPU_VENDOR_ID: CpuidFeature = CpuidFeature {
+    leaf: 0,
+    subleaf: 0,
+    register: CpuidRegister::Edx,
+    shift: 0,
+    width: 32,
+};
 
-/// The CPUID output register for a particular feature
-#[derive(Clone, Copy, Debug)]
-enum CpuidReg {
-    Eax,
-    Ebx,
-    Ecx,
-    Edx,
-}
+/// Raw EAX from CPUID leaf 0x8000_0008 — processor capacity parameters.
+/// Not in the cpufeature database as a full-register field (only the individual
+/// bit ranges are), so keep a local descriptor for the full EAX value.
+const PHYS_ADDR_SIZES: CpuidFeature = CpuidFeature {
+    leaf: 0x8000_0008,
+    subleaf: 0,
+    register: CpuidRegister::Eax,
+    shift: 0,
+    width: 32,
+};
+
+/// CPUID.4000_0001:EAX — Hyper-V interface signature ("Hv#1").
+const HYPERV_INTERFACE: CpuidFeature = CpuidFeature {
+    leaf: 0x4000_0001,
+    subleaf: 0,
+    register: CpuidRegister::Eax,
+    shift: 0,
+    width: 32,
+};
+
+const HYPERV_INTERFACE_SIGNATURE: u32 = 0x3123_7648;
 
 /// A discoverable CPU feature.
 ///
@@ -26,78 +57,35 @@ enum CpuidReg {
 struct CpuFeat {
     /// Raw value
     val: ImmutAfterInitCell<u32>,
-    /// CPUID leaf
-    leaf: u32,
-    /// CPUID subleaf
-    subleaf: u32,
-    /// CPUID output register
-    reg: CpuidReg,
-    /// Bitshift to apply to raw value
-    shift: u8,
-    /// Bitmask to apply to raw value
-    bitsize: u8,
+    /// Location of the feature within a CPUID leaf
+    feature: CpuidFeature,
     /// Expected value after shift + mask
     expected: u32,
 }
 
 impl CpuFeat {
-    /// Create a new feature, indicated by a single bit in the specified
-    /// register in the given CPUID leaf.
-    const fn new_bit(leaf: u32, reg: CpuidReg, bit: u8) -> Self {
-        Self::new(leaf, reg, bit, 1, 1)
+    /// Create a new feature, indicated by a single bit (expected value 1).
+    const fn new_bit(feature: CpuidFeature) -> Self {
+        Self::new(feature, 1)
     }
 
-    /// Create a new feature, indicated by the full value of a register
-    /// in the given CPUID leaf.
-    const fn new_u32(leaf: u32, reg: CpuidReg, expected: u32) -> Self {
-        Self::new(leaf, reg, 0, u32::BITS as u8, expected)
-    }
-
-    /// Create a new CPU feature, detected by querying the given CPUID leaf, and applying
-    /// a bitshift and bitmask on the specified output register to compare it to the given
-    /// expected value.
-    const fn new(leaf: u32, reg: CpuidReg, shift: u8, bitsize: u8, expected: u32) -> Self {
-        // This method is only called from const context, so these assertions have no
-        // runtime effect.
-        assert!((shift as u32) < u32::BITS);
-        assert!((bitsize as u32) <= u32::BITS);
+    /// Create a new CPU feature from a cpufeature descriptor, comparing the
+    /// extracted field against the given expected value.
+    const fn new(feature: CpuidFeature, expected: u32) -> Self {
         Self {
             val: ImmutAfterInitCell::uninit(),
-            leaf,
-            subleaf: 0,
-            reg,
-            shift,
-            bitsize,
+            feature,
             expected,
         }
-    }
-
-    /// Create a copy of the given CPU feature, but with the specified
-    /// subleaf.
-    const fn with_subfn(mut self, subleaf: u32) -> Self {
-        self.subleaf = subleaf;
-        self
-    }
-
-    /// Get the CPUID register that corresponds to this feature
-    const fn get_reg(&self, cpuid: &CpuidResult) -> u32 {
-        match self.reg {
-            CpuidReg::Eax => cpuid.eax,
-            CpuidReg::Ebx => cpuid.ebx,
-            CpuidReg::Ecx => cpuid.ecx,
-            CpuidReg::Edx => cpuid.edx,
-        }
-    }
-
-    const fn mask(&self) -> u32 {
-        ((1u64 << self.bitsize) - 1) as u32
     }
 
     fn get_or_init(&self) -> u32 {
         if let Ok(val) = self.val.try_get_inner() {
             return *val;
         }
-        let val = cpuid(self.leaf, self.subleaf).map_or(0, |c| self.get_reg(&c));
+        let val = cpuid(&self.feature).map_or(0, |c| {
+            self.feature.register.select(c.eax, c.ebx, c.ecx, c.edx)
+        });
         // If init() fails it means the cell got initialized by someone else
         // concurrently, which is always benign.
         let _ = self.val.init(val);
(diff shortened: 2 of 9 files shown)


tanish111:
@luigix25 a test PR.

kernel/src/platform/native.rs
```diff
@@ -59,6 +60,13 @@ impl NativePlatform {
     }
 }
 
+impl CpuidBackend for NativePlatform {
+    fn cpuid(feature: &CpuidFeature) -> Option<core::arch::x86_64::CpuidResult> {
+        // SAFETY: CPUID is always safe on native hardware.
+        Some(unsafe { core::arch::x86_64::__cpuid_count(feature.leaf, feature.subleaf) })
+    }
+}
+
```
luigix25: I'm not really convinced about this.
we already have a cpuid function from the SvsmPlatform.

should `CpuidBackend` be a "supertrait" for `SvsmPlatform`?
and each architecture, that cannot rely on the default implementation, should override that function.

@joergroedel wdyt?

tanish111: @luigix25 looks like a good idea to me. (idk why I didn't think this in first place)

kernel/src/platform/native.rs
```diff
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 //
 // Copyright (c) Microsoft Corporation
-// Copyright (c) SUSE LLC
```
luigix25, line 4: avoid unrelated changes

luigix25 (changes requested):
I think we should we drop `cpuid` from the `SvsmPlatform` trait entirely.
We still have code that is calling the "old" cpuid.

Please update the commit message as we are not doing anything related to the shadowstack.

luigix25 (changes requested):
Once again avoid unrelated changes, and split the PR into multiple commits as we said during the call.

Ideally one commit per feature.

One the last commit should drop the `cpuid` function from SvsmPlatform so that every commit compiles.

tanish111:
@luigix25 it's giving clippy errors with this code 
```
// Authors: Joerg Roedel <jroedel@suse.de>
//          Carlos López <clopez@suse.de>
```
Can you check it once.

stefano-garzarella:
> @luigix25 it's giving clippy errors with this code
> 
> ```
> // Authors: Joerg Roedel <jroedel@suse.de>
> //          Carlos López <clopez@suse.de>
> ```
> 
> Can you check it once.

CI runs `make clippy` on PRs and I didn't see any error about that. What change triggered it?

Anyway, also if it's the case, it's unrelated so it should be in a separate commit/PR.

tanish111:
@stefano-garzarella sorry clippy was not the culprit it's [pre-commit hooks](https://github.com/coconut-svsm/svsm/blob/main/scripts/pre-commit)
```
	if  ! [[ "${arr[4]}" =~ ^//[[:space:]]Author: ]] ; then
		echo "${header_warning}"
		return 1
	fi
```
thanks @luigix25 for quick debug.
(for now would remove pre-commit hook locally and would work on long term fix)

stefano-garzarella:
@tanish111 please could you explain more clearly in the commit and also in the PR description why we're doing this? (It's not at all clear to me, though I admit I haven't seen the changes; but since I don't understand the purpose of these changes, it's hard for me to follow)

tanish111:
@stefano-garzarella The motivation is to stop maintaining hand-written CPUID checks in SVSM. Right now, each cpu feature is detected with custom leaf/register/bit logic, which is repetitive and easy to get wrong as we add more features. 
Instead, we're using a shared [cpufeature](https://github.com/coconut-svsm/cpufeature) crate that is auto-generated from the [X86-CPUID](https://x86-cpuid.org/) database. That gives us a single, consistent way to enumerate CPU features, rather than scattering open-coded CPUID logic across the kernel. Also this is part of my GSoC project.

stefano-garzarella:
@tanish111 Thanks, I see now!

This is exactly the information that the PR and commit descriptions should include. This makes it easier for reviewers or anyone who looks at the commit in the future to understand the reasoning behind a change.

tanish111:
CPUID leaf 0x4000_0001 (Hyper-V interface signature "Hv#1" in EAX) is defined by the Hyper-V TLFS but is not in [x86-cpuid-db](https://gitlab.com/x86-cpuid.org/x86-cpuid-db) only 0x4000_0000 (vendor ID) is. So cpufeature has no generated descriptor for it, and we define it locally in hyperv until it can be added upstream.

tanish111:
@luigix25 it's ready to review

osteffenrh:
> Instead, we're using a shared [cpufeature](https://github.com/coconut-svsm/cpufeature) crate that is auto-generated from the [X86-CPUID](https://x86-cpuid.org/) database.

I missed all the initial discussion, maybe it got discussed already:
Why reinvent a cpu feature detection crate instead of using https://crates.io/crates/cpufeatures ?

tanish111:
@osteffenrh crate you mentioned only covers ~20 hardcoded crypto/SIMD features as booleans. It still lacks support for arbitrary fields like multi-bit PCID/address-width values and many more.

Also it hardcodes raw CPUID execution, which doesn’t work for SNP guests that need to validate/source CPUID via the CPUID page rather than trusting the instruction directly, our CpuidBackend trait makes that pluggable.

00xc (commented):
I don't understand why we need to remove `features.rs`, since it already contains some of the functions you're introducing with a new name (e.g. `cpu_has_feat()` -> `has_cpuid_feature()`). I'd rater have the existing functions updated to use the new backend you're introducing (with the proper CPUID leaf definitions instead of manually inserted ones).

00xc:
> Also it hardcodes raw CPUID execution, which doesn’t work for SNP guests that need to validate/source CPUID via the CPUID page rather than trusting the instruction directly, our CpuidBackend trait makes that pluggable.

Technically it does work since we handle CPUID in the #VC exception handler, but it's definitely undesirable to take an exception for no good reason.

tanish111:
@00xc is there a specific reason we want to keep features.rs?
I agree that removing it means we lose the CPUID caching it had, and that’s a real downside. But in exchange we don’t have to register every feature in define_cpu_feats! before we can use it anywhere in the tree. Also CpuidFeature (which crate provides) and CpuFeat are identical when you compare the fields (leaf, subleaf, reg, shift, width), so keeping both felt redundant. The cpufeature crate already covers most of what features.rs did it had the registry (removing define_cpu_feats!) and the per-feature struct. That’s why I dropped the file and moved those helper functions to the platform layer instead.
Here is the PR for cpufeature crate for implementation details:- https://github.com/coconut-svsm/cpufeature/pull/1

tanish111:
Also open to ideas on how we can implement caching without having a central registry that need to be updated every time we want to use a new CPUID. One possible solution is to add cache at query time with a global lookup table. (For the first time it fills look-up table and on next call onwards use it to return cpuid)

00xc:
> @00xc is there a specific reason we want to keep features.rs?

To minimize the overall diff size. The API you're replacing already exists.

> I agree that removing it means we lose the CPUID caching it had, and that’s a real downside.

I am fine with removing the caching. What I was suggesting is that we can keep the existing public API and only change what's behind that API.

> But in exchange we don’t have to register every feature in define_cpu_feats! before we can use it anywhere in the tree. Also CpuidFeature (which crate provides) and CpuFeat are identical when you compare the fields (leaf, subleaf, reg, shift, width), so keeping both felt redundant. The cpufeature crate already covers most of what features.rs did it had the registry (removing define_cpu_feats!) and the per-feature struct. That’s why I dropped the file and moved those helper functions to the platform layer instead. Here is the PR for cpufeature crate for implementation details:- [coconut-svsm/cpufeature#1](https://github.com/coconut-svsm/cpufeature/pull/1)

That is all fine, we can rewrite the guts of `features.rs` and replace it with your crate (and even add caching back in later). That is what I meant by "I'd rater have the existing functions updated to use the new backend you're introducing (with the proper CPUID leaf definitions instead of manually inserted ones)".

00xc:
> Also open to ideas on how we can implement caching without having a central registry that need to be updated every time we want to use a new CPUID. One possible solution is to add cache at query time with a global lookup table. (For the first time it fills look-up table and on next call onwards use it to return cpuid)

That is what is what the existing code does.

For this to work efficiently you need to find a way to generate the lookup table at compile time (that is what `define_cpu_feats!()` does), since you need it to be `const` for the lookup to be cheap. Also IIRC `const` variables are inlined into each user (at least in release builds), as opposed to `static`, which generates a variable in e.g. the `.bss` section, and increases binary size.

tanish111:
@00xc I've updated the PR to keep the existing public API (cpu_has_feat / cpu_get_feat), but callers still need to change because the signature now takes &CpuidFeature constants from the crate instead of Feature. (The caller updates are mostly import + Feature::Foo → &X86_FEATURE_FOO; the behavioral logic stays the same.)

An alternative would be to keep define_cpu_feats! and the Feature enum, with each variant mapping to a cpufeature constant internally that would avoid touching call sites. I don't think that's worth it: we'd be bookkeeping the same CPUID leaf/register/bit data twice, and readers would have to mentally map two names (Feature::Xsave vs X86_FEATURE_XSAVE) for the same thing. Passing the crate constants directly keeps a single source of truth. wdyt?

Also PR is ready for review can you ptal?

kernel/src/cpu/features.rs
```diff
@@ -5,170 +5,51 @@
...
-}
-
-/// Gets the raw value of the given feature
-pub fn cpu_get_feat(feat: Feature) -> u32 {
-    CPU_FEATS[feat as usize].get()
+pub fn cpu_has_feat(feature: &CpuidFeature) -> bool {
+    if core::ptr::eq(feature, &HYPERV_INTERFACE) {
+        cpuid_field(feature) == Some(HYPERV_INTERFACE_SIGNATURE)
+    } else {
+        cpuid_field(feature) == Some(1)
```
00xc: Or maybe we can consider the feature present if it is `Some` and not zero.

```rust
cpuid_field(feature).is_some_and(|f| f != 0)
```

kernel/src/cpu/sse.rs
```diff
@@ -5,13 +5,16 @@
...
 use core::arch::x86_64::{_xgetbv, _xsetbv};
 use core::sync::atomic::{AtomicU64, Ordering};
-
-use super::features::{Feature, cpu_has_feat};
-
+use cpufeature::CpuidFeature;
+use cpufeature::leaves::XSAVE_SZ;
+use cpufeature::leaves::{
+    X86_FEATURE_XMM, X86_FEATURE_XSAVE, X86_FEATURE_XSAVEOPT, XCR0_AVX, XCR0_SSE, XCR0_X87,
+};
```
00xc: These imports can be merged

kernel/src/cpu/features.rs
tanish111: Yes this works as well

00xc:
We had a discussion regarding this PR during the TSC meeting.

The sentiment is that the existing CPUID caching is not only a performance feature but also a security feature, since it prevents the hypervisor from offering inconsistent results throughout the lifetime of the SVSM.

I think the best way forward is to replace the existing manual definitions with those in your crate, but keep the macro that generates the lookup table and the logic to cache queries. That way we get the benefits of both approaches: automatic feature definitions plus caching.

tanish111:
@00xc thanks for reviews I would update it accordingly

00xc:
> For this to work efficiently you need to find a way to generate the lookup table at compile time (that is what `define_cpu_feats!()` does), since you need it to be `const` for the lookup to be cheap. Also IIRC `const` variables are inlined into each user (at least in release builds), as opposed to `static`, which generates a variable in e.g. the `.bss` section, and increases binary size.

Not sure what I was thinking here, but this is wrong. The lookup table needs to be `static` because it needs to be updated at runtime with the cached values. `define_cpu_feats!()` does generate a `static` table.

tanish111:
@00xc I have pushed suggested changes can you ptal?

kernel/src/platform/native.rs
```diff
@@ -87,14 +90,10 @@ impl SvsmPlatform for NativePlatform {
...
     fn get_cpu_vendor(&self) -> CpuVendor {
-        if let Some(r) = cpuid(0, 0) {
-            match r.edx {
-                0x69746e65 => CpuVendor::AMD,
-                0x49656e69 => CpuVendor::Intel,
-                _ => CpuVendor::Unknown,
-            }
-        } else {
-            CpuVendor::Unknown
+        match Self::cpuid(&MAX_STD_LEAF).map(|r| r.edx) {
```
luigix25: mmh, this is more of an hack isn't it?

Cargo.toml
```diff
@@ -106,6 +106,9 @@ virtio-drivers = { git = "https://github.com/rcore-os/virtio-drivers.git", rev =
 zerocopy = { version = "0.8.2", features = ["derive"] }
 zeroize = { version = "1.8.2", default-features = false }
 
+#Github repos
```
luigix25: this comment can be dropped

kernel/src/platform/native.rs
luigix25: `MAX_STD_LEAF` is technically correct, but semantically wrong. I think we should use a new const

tanish111:
@luigix25 I have resolved all comments. can you review and approve it if everything looks good?

kernel/src/platform/native.rs
```diff
@@ -87,7 +90,7 @@ impl SvsmPlatform for NativePlatform {
     }
 
     fn get_cpu_vendor(&self) -> CpuVendor {
-        if let Some(r) = cpuid(0, 0) {
+        if let Some(r) = Self::cpuid(&MAX_STD_LEAF) {
```
luigix25: as I already said (https://github.com/coconut-svsm/svsm/pull/1137#discussion_r3602102267), `MAX_STD_LEAF` is technically correct, but not semantically. We should use another constant.

luigix25 (changes requested):
While we are here: it's a good practice to write a mini changelog between versions: for reference see #1166 (v1, v2 ecc). So it's easier for reviewers understand the changes.

kernel/src/platform/native.rs
tanish111: @luigix25 should I create a new const just like [HYPERV_INTERFACE](https://github.com/coconut-svsm/svsm/pull/1137/changes#diff-23555b376665b1c4868eef8baa451a5b5aaa1ebc85357b1984f6a6bf0f4b8015R30). Something maybe CPU_VENDOR_ID?

luigix25: yes, please.

tanish111: fixed.

peterfang (approved):
The TDX part looks good to me. Thanks!

luigix25 (approved):

00xc (approved):

PR #1138, opened 2026-06-23
kernel/protocols: Introduce OCP support

n-ramacciotti:
A guest OS, running inside a CVM does not know the state information, such as log data or memory usage, of the SVSM. These sources of information could be useful for the guest OS to understand the environment it is running in.  It might also be useful for the guest OS to be able to configure some of these sources.

A new protocol, called Observability and Configuration Protocol (OCP), has been defined to enable information to be retrieved from an SVSM source, and to allow the configuration of sources in SVSM. All the sources and source categories available in SVSM can be listed using this protocol, and each of them can be read and potentially written.

In this PR the SVSM side of the protocol is implemented.

More details can be found in the commit messages.

- Linux Device Driver: see https://github.com/coconut-svsm/linux/pull/20
  **note**: This require the vm to have a custom guest kernel with the above commits.

- Userspace utility: see https://github.com/n-ramacciotti/svsm-tools/pull/1

Tested on a SNP machine. (Thanks @joergroedel )

/cc @stefano-garzarella @kraxel 

This started as part of my work for Google Summer of code [1] based on the draft protocol posted in the mailing list [2]. An update version of the proposal is in one commit of this PR that updates the documentation.

[1] https://wiki.qemu.org/Google_Summer_of_Code_2026
[2] https://lore.kernel.org/coconut-svsm/aeiLnWdClMC6YuLi@sirius.home.kraxel.org/T/#t


tlendacky:
Is this meant to be a Coconut specific protocol or a general any SVSM implementation protocol. If the latter, then it needs to be worked out properly and added to the SVSM specification. If the former, then it will need to use one of the reserved reference implementation protocol numbers and needs to be documented for Coconut.

n-ramacciotti:
v2:
- Introduced `OcpSourceOperations` trait: now each source must be implement it in order to be used in the protocol. It contains `read`, `write` and `get_source_entry` methods. The latter is useful to retrieve the information and interate over the array. `read` and `write` have default implementations which return an error.
- Source name len is now up to 123 bytes. The last one is reserved for null terminator. If one source has a smaller name, the other bytes are padded with 0. It still panics on len discrepancies.
- Changed `ocp_list_request` to take into consideration the new trait and `get_source_entry_method`. Tested with the linux PR.
- Introduced `SVSM_OCP_READ` and `SVSM_OCP_WRITE`. Not yet tested with linux. Their structure is based on the new trait. They are basically a wrapper for the source trait implementation + input checks.
- Introduce a `[test]` commit with a fake log buffer to test the list request.
- For now, the SOURCE array is built at compile time in a static value of trait objects.
- Each source should implement `Sync` for safe access.

n-ramacciotti:
> Is this meant to be a Coconut specific protocol or a general any SVSM implementation protocol. If the latter, then it needs to be worked out properly and added to the SVSM specification. If the former, then it will need to use one of the reserved reference implementation protocol numbers and needs to be documented for Coconut.

@tlendacky  This is part of my work for google summer of code in the coconut-svsm[1].

This is based on the draft proposal of @joergroedel posted in the mailing list [2].

I believe this will be general for any SVSM implementation protocol.

Sorry for not mentioning this earlier in the PR description. Now it's updated.

[1] https://wiki.qemu.org/Google_Summer_of_Code_2026
[2] https://lore.kernel.org/coconut-svsm/aeiLnWdClMC6YuLi@sirius.home.kraxel.org/T/#t

kernel/src/protocols/ocp.rs
```diff
@@ -0,0 +1,62 @@
...
+            // Compile time fialure if the length is greater than
+            // that value as we want a null terminated string
+            panic!("Name length must be positive and less than 124 bytes");
+        }
+
+        let mut i = 0;
+        while i < len {
+            name_bytes[i] = bytes[i];
+            i += 1;
+        }
```
00xc: I think this just works, no?

```rust
name_bytes[..len].copy_from_slice(bytes);
```

kernel/src/protocols/ocp.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub const fn new(writable: bool, name: &str) -> Self {
+        let mut name_bytes = [0u8; OCP_SOURCE_NAME_LEN];
+        let bytes = name.as_bytes();
+        let len = bytes.len();
+
+        if len == 0 || len >= OCP_SOURCE_NAME_LEN {
+            // Compile time fialure if the length is greater than
+            // that value as we want a null terminated string
+            panic!("Name length must be positive and less than 124 bytes");
+        }
```
00xc: This is not guaranteed to happen at compile time. `const` just indicates that the function **can** be used in const context, not that it will.

kernel/src/protocols/ocp.rs
```diff
@@ -0,0 +1,62 @@
...
+impl OcpSource {
+    pub const fn new(writable: bool, name: &str) -> Self {
+        let mut name_bytes = [0u8; OCP_SOURCE_NAME_LEN];
+        let bytes = name.as_bytes();
+        let len = bytes.len();
+
+        if len == 0 || len >= OCP_SOURCE_NAME_LEN {
+            // Compile time fialure if the length is greater than
+            // that value as we want a null terminated string
+            panic!("Name length must be positive and less than 124 bytes");
```
00xc: Instead of hardcoding "124" in the error message I would suggest printing the actual value of `OCP_SOURCE_NAME_LEN`

kernel/src/protocols/ocp.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    let mut buffer = Vec::with_capacity(buffer_size);
+
+    for &entry in OCP_SOURCES[first..end].iter() {
+        buffer.extend_from_slice(entry.get_source_entry().as_bytes())
+    }
+
+    copy_slice_to_guest(buffer.as_bytes(), gpa_buffer)?;
+
+    params.rcx = entries_to_return as u64;
```
00xc: I'm not a great fan of allocating memory on each request. I believe you can create a `PerCPUMappingGuard` and use that to write each entry in a loop. Same thing for similar operations below.

00xc (commented):
I realize this is in RFC status still, so feel free to discard my comments if the code changes significantly.

kernel/src/protocols/ocp.rs
n-ramacciotti: Right. I will update the comment

kernel/src/protocols/ocp.rs
n-ramacciotti: This cannot be done because:
```bash
error[E0015]: cannot call non-const formatting macro in constant functions
  --> kernel/src/protocols/ocp.rs:84:13
   |
84 | ...   panic!("Name length must be positive and less than {} bytes", OCP_SOURCE_NAME_LEN);
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants

```

kernel/src/protocols/ocp.rs
n-ramacciotti: This is a similar error as in https://github.com/coconut-svsm/svsm/pull/1138#discussion_r3494677478: 
```bash
error[E0015]: cannot call non-const operator in constant functions
  --> kernel/src/protocols/ocp.rs:87:19
   |
87 |         name_bytes[..len].copy_from_slice(bytes);
   |                   ^^^^^^^
   |
   = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants

For more information about this error, try `rustc --explain E0015`.
```

kernel/src/protocols/ocp.rs
n-ramacciotti: I will look into this. Thanks!

kernel/src/protocols/ocp.rs
00xc: Then let's just print the constant name (e.g. "Name length must not be zero nor exceed OCP_SOURCE_NAME_LEN"). This is an error only a developer should see anyway.

kernel/src/protocols/ocp.rs
n-ramacciotti: A `PerCPUPageMappingGuard` is created internally by `copy_slice_to_guest`. So, I could loop over each entry using `copy_slice_to_guest` for each one directly, but I don't think that is what you meant.

IIUC, you're suggesting something similar to `core_pvalidate`, where we can create the mapping guard, get a `GuestPtr` and loop over each entry using `GuestPtr::write`. Right?

00xc: > IIUC, you're suggesting something similar to core_pvalidate, where we can create the mapping guard, get a GuestPtr and loop over each entry using GuestPtr::write. Right?

Yes, correct. That way you don't have to construct an intermediate heap-allocated buffer, and you don't map/unmap memory (with the corresponding TLB flush) for each entry.

For the record, cleaning up the `PerCPUPageMappingGuard` / `GuestPtr` combination is on my TODO list, it's not very ergonomic :)

n-ramacciotti: > Yes, correct. That way you don't have to construct an intermediate heap-allocated buffer, and you don't map/unmap memory (with the corresponding TLB flush) for each entry.

Right! I'll take this approach. Thanks!

kraxel:
> [2] https://lore.kernel.org/coconut-svsm/aeiLnWdClMC6YuLi@sirius.home.kraxel.org/T/#t

The discussion on the protocol died off without conclusion, so I'll use the opportunity to resume it here.

I still think it makes sense to include some information what type of data is returned by a source in the SVSM_OCP_LIST reply (i.e. what became `struct OcpSource` in this PR).  Specifically I think the entries should include enough metadata that the linux kernel driver can proxy the data between svsm and linux userspace without additional knowledge about the sources.  Also linux userspace should be able to collect and store metrics without additional knowledge about the source.  Only when analyzing the collected data you need to know what the data you got from the svsm actually represents.

The other two ideas from my mail (`SVSM_OCP_READ_REGISTER`) + (`SVSM_OCP_READ_BATCH`) can be added to the protocol later on, so lets put those aside for now.

n-ramacciotti:
v3:
- Addressed @00xc's comments:
  - improved panic message and comment when creating `OcpSourceFlags`
  - Introduced `PerCPUMappingGuard` and `GuestPtr` to avoid allocating on the heap. (For now only in `SVSM_OCP_LIST` handler)

n-ramacciotti:
v4:
- Introduced a new read-only source, which returns the running version of SVSM
- Introduced a new field `kind`, which indicates the type of data the source holds
- Simplified `SVSM_OCP_READ` and `SVSM_OCP_WRITE` with removal of heap allocation. Now the copy is source dependent and must be done inside the trait implementation

n-ramacciotti:
v5:
- Introduced max buffer size
- Introduced testing in linux fork for read operation. It successfully prints coconut version.

kernel/src/protocols/ocp.rs
```diff
@@ -0,0 +1,76 @@
...
+/// OCP source entry structure.
+#[repr(C)]
+#[derive(Debug, IntoBytes, Immutable)]
+pub struct OcpSource {
+    /// Source flags.
+    flags: OcpSourceFlags,
+    /// Name of the source encoded as UTF-8.
+    name: [u8; OCP_SOURCE_NAME_LEN],
+    /// Type of the source.
+    kind: OcpSourceType,
```
kraxel: Minor nit: I'd put the two `u32` fields type + flags first and the name as last element.

kernel/src/protocols/ocp.rs
```diff
@@ -74,3 +79,14 @@ const _: () = assert!(
...
+
+/// Operations required for an OCP source
+pub trait OcpSourceOperations: Sync {
+    fn read(&self, _offset: u32, _gpa: PhysAddr, _size: u32) -> Result<u32, SvsmReqError> {
+        Err(SvsmReqError::unsupported_call())
+    }
+    fn write(&self, _offset: u32, _gpa: PhysAddr, _size: u32) -> Result<u32, SvsmReqError> {
+        Err(SvsmReqError::unsupported_call())
+    }
+    fn get_source_entry(&self) -> &OcpSource;
```
kraxel: Does it make sense to allow the trait support multiple sources, so objects implementing the trait are not limited to one?

kernel/src/protocols/ocp.rs
n-ramacciotti: Makes sense!

kernel/src/protocols/ocp.rs
n-ramacciotti: Wouldn't we lose direct indexing of the source array in this case?

kraxel: It would make things a bit more complicated, yes.  You'd need an array with pointer and source index for example.  Not fully sure this is a good path forward, maybe it is easier to just create one object per source.

I was wondering if -- for example -- the snp platform object wants allow multiple things being observed such as platform feature flags and cbit position.  How would that be done?  I see two possible approaches:
* Implement the Trait for the platform object + allow multiple sources, or
* Create one object (may or may not be embedded in the platform struct) per source.

n-ramacciotti:
v6:
- moved type field as first element in `OcpSource`
- introduced a test source for write operations. Read a counter and increase it multiple times. this will be dropped it's here for testing purposes. (can be tested with the linux PR in coconut fork)
- rebased on latest main

Todo: 
- [x] improve commit messages

kernel/src/protocols/ocp.rs
n-ramacciotti: @kraxel @stefano-garzarella @joergroedel when introducing multiple sources per object with the two indexes, does it make sense to split the `LIST` request in two different calls:
- one that returns the "object index(es)" and the number of sources for the object(s) + some information on the object category;
- one that returns the sources (updated with index and sub-index) for the object(s)

Or is it better to keep the current behavior, where the `LIST` call returns directly an array of sources (updated with index and sub-index fields)? I'm wondering as currently `R8` contains the number of array entries to return, but now each array entry can contains multiple sources. 

Additionally, as the objects array can potentially have "holes" as an ("super") index doesn't represent necessarily the position in the array, does it make sense to change the array of objects to a sort of ordered hash map? This map should also be populated at runtime and should probably be accessed in mutual exclusion when updated, for example if a source is added later.

kraxel: I like the idea to have two separate list operations, one for objects and one for sources of a given object.  I also like the idea of having an object category (svsm, testdummy, vtpm, process, ...), which is easier to implement with a separate object-list operation.

I think we should also clarify the possible lifetimes of objects and sources.  Objects can clearly come and go.  Sources which belong to an object will be added and removed together with the object.  But do we also want allow objects changing the list of sources during the lifetime of the object?  I can't think of a case where we will need that flexibility, so I'd tend to keep things simple and declare the list of sources for a given object to be constant.

n-ramacciotti: >  I can't think of a case where we will need that flexibility, so I'd tend to keep things simple and declare the list of sources for a given object to be constant.

I agree on this

stefano-garzarella: @n-ramacciotti I agree with @kraxel. I'd go on that direction!

n-ramacciotti:
v7:
- Introduced multiple sources for each object
- Used a `BtreeMap` instead of an array for keeping object. 
- Introduced two different list operations (which probably can be simplified as they are very similar):
  - one for objects details
  - one for sources in a given object
- Add a test object for:
  - returning svsm version running
  - writing and reading an atomic value
- Remove `const` from `new` of `OcpSource` as now sources are built at runtime and can use more features (like panic formatting and slice [.. ])

Todo:
 - [x] Test on SNP -> this requires new linux driver version 
 - [x] Improve commit messages
 - [x] Implement something similar to `VsockDriver::get_first_free_port` to store in an atomic variable and fetch a free (super) index?
 - [x] gate the OCP behind a feature?

n-ramacciotti:
v8:
- Added documentation
- Introduced feature
- tested on snp
- improved commits messages
- added utility to get free index
- refactored in multiple modules
- introduced utilities to interact with the map

TODO:
- [x] Introduce log buffer. Since this PR is getting big, I'm not sure if it makes sense to add it in this PR or open a new draft based on this PR.

n-ramacciotti:
v9:
- Introduced log buffer as source
- Introduced `GuestBuffer`, similarly to `UserBuffer`, to read and write data to the guest through fs objects
- Updated doc on log buffer
- fixed typo in commit a69a1a317975d1d23075d356ac14db4214e104f2

**note**: there is a commit marked with test as:
  - I'm not sure where that specific code should be placed (also the `SvsmObject` name is a placeholder)
  - Part of that commit (the source with write permission) should be dropped

Other than that, the PR is ready for review.

kernel/src/protocols/ocp/source.rs
```diff
@@ -79,4 +80,34 @@ const _: () = assert!(
...
+/// Type of objects the SVSM contains.
+pub enum OcpObjectType {
+    Svsm = 0,
+}
+
+#[repr(C)]
+#[derive(Debug, IntoBytes, Immutable)]
+pub struct OcpObjectDetails {
+    category: OcpObjectType,
+    index: u32,
```
kraxel: Rename to `sup_index` for consistency with `OcpSource` ?
Also place field first in struct (again like `OcpSource`)?

kernel/src/protocols/mod.rs
```diff
@@ -22,6 +22,7 @@ pub const SVSM_ATTEST_PROTOCOL: u32 = 1;
 pub const SVSM_VTPM_PROTOCOL: u32 = 2;
 pub const SVSM_APIC_PROTOCOL: u32 = 3;
 pub const SVSM_UEFI_MM_PROTOCOL: u32 = 4;
+pub const SVSM_OBSERVABILITY_CONFIGURATION_PROTOCOL: u32 = 5;
```
kraxel, line 26: This chunk can go into a separate pull request which does nothing but reserving the protocol number.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -180,11 +181,44 @@ fn ocp_read_request(params: &mut RequestParams) -> Result<(), SvsmReqError> {
...
+
+    if bytes_to_write as usize > OCP_BUFFER_MAX_SIZE {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+
+    let Some(object) = get_object(sup_index) else {
+        return Err(SvsmReqError::invalid_parameter());
+    };
+
+    if bytes_to_write == 0 {
```
kraxel: add check here whenever the writable bit is set in the object flags?

kraxel (commented):
Some minor comments, looks good to me overall.

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: My idea was to check this inside `object.write` where each source could check its flags, but I agree that we should check if the source is writable before returning something (0 in this case). We could also introduce another trait method (something like `is_source_writable(sub_index)`) or adding flags to the object.

kraxel: Oh, right, writable is per-source not per-object so this is a bit more complicated.

n-ramacciotti: We could also introduce something like `get_object_source`, where we can obtain a source from an object. Here we could use another trait, this time specific to sources, to perform the read. I should think about this more though.

n-ramacciotti:
v10:
- Applied @kraxel suggestions + other changes:
- Renamed `OcpSource` in `OcpSourceDetails`
- reordered fields in `OcpObjectDetail`
- Removed 0 byte write check
- Introduced a second trait for sources
- Introduced a new utility in `guestmem` to create a mapping guard on a checked memory region. This simplify the two list operatios. It can also be used in `copy_from_guest` and other points to simplify the code. I have not added it to those functions yet.
- Added check on map len when iterating

- For testing, I called three times `add_svsm_object`. This is to have multiple objects to test better the protocol. it's not a problem as internally it get a new `sup_index` on every call. This will be dropped.

Todo:
- [x] Move protocol number in another PR, that only reserve the number
- [ ] Add to documentation that a source index is not reused

n-ramacciotti:
v11:
- Solved conflict on `kernel/src/fs/buffer.rs`
- Updated doc with `Indexes are stable during runtime.`
- I opened #1181  with a variation of d5beedb
- I opened #1182 to reserve OCP number

n-ramacciotti:
v12:
- Rebased on latest main to solve a conflict in `kernel/src/lib.rs`

The new userspace utility to interact with the protocol is here: https://github.com/n-ramacciotti/svsm-tools/pull/1

kernel/src/protocols/ocp/details.rs
```diff
@@ -10,6 +10,7 @@ use zerocopy::{Immutable, IntoBytes};
 
 const OCP_SOURCE_NAME_LEN: usize = 112;
 pub const OCP_SOURCE_ENTRY_SIZE: usize = 128;
+pub const OCP_SOURCE_DETAILS_SIZE: usize = 12;
```
kraxel: I'd suggest to rename these to `OCP_SOURCE_DETAILS_SITE` and `OCP_OBJECT_DETAILS_SIZE` to make naming more consistent with the struct names and less confusing.

kernel/src/ocp_test.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    fn add_source(&mut self, source: Arc<dyn OcpSourceOperations>) {
+        // todo remove mut
+        self.ocp_source_entries.push(source);
+        self.details.increase_count();
+    }
+}
+
+impl OcpObjectOperations for OcpSvsmObject {
+    fn read(
```
kraxel: There is no code specifiic to `OcpSvsmObject` here, I think this can be the default implementatiob for the `OcpObjectOperations` trait (same for write).

n-ramacciotti: > There is no code specifiic to OcpSvsmObject here

There is `self.ocp_source_entries.get(sub_index as usize)`. However, we could introduce a new trait method `get_sources_at_index` that is called inside the default implementation.

n-ramacciotti:
v13:
- marked commit that add `checked_guest_region_guard` with `[hack]`. This is to avoid rebasing on the new `GuestPtr` infrastructure PR. It will be dropped when that PR is ready
- addressed @kraxel's comments:
  - renamed sources and object details
  - use trait default implementation for read and write + a new trait method.
- updated documentation with new information

TODO:
- [x] `num_entries` in list operations should be changed to `num_bytes` requested/returned

n-ramacciotti:
v14:
- rebased on latest main and pick `GuestPtr` changes
- changed `num_entries` in `num_bytes` in request params for both list operations
- updated description with max number of bytes and the new parameter type.

n-ramacciotti:
[FIRST VERSION AFTER GSoC] 
v15:
- update according to new spec (also summarized in a commit)
- drop index and sub index. Now the protocol uses names.
- drop `OcpObjectDetails` and renamed `OcpSourceDetails` in `OcpSource`
- Updated `OcpSourceType` with the new type (`Object`, `Bytes`, `SInteger8bit`) instead of `String`, `StaticString` etc
- Removed `write_source` as not needed
- Removed `BtreeMap` for a `Vec`
- Introduced a new utility to extract names from OcpIdentifier (`ObjectName/SourceName`)

TODO:
- [ ] solve todos in traits:
  - check offset and write permission in default implemetnation of object write
  - change `get_object_sources` to avoid the `Iter`
- [ ] Update commits description
- [ ] Check object uniqueness

kernel/src/protocols/ocp/api.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use super::source::OcpSource;
+
+use crate::{address::PhysAddr, protocols::errors::SvsmReqError};
+
+use alloc::sync::Arc;
+use core::{fmt::Debug, slice::Iter};
+
+/// Operations required for an OCP object
+pub trait OcpObjectOperations: Debug + Send + Sync {
```
00xc: Or just `OcpObject`?

kernel/src/protocols/ocp/api.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn get_object_name(&self) -> &str;
+    fn get_object_details(&self) -> &OcpSource;
+    fn get_object_source_by_name(&self, name: &str) -> Option<Arc<dyn OcpSourceOperations>>;
+    // todo: change return type of get_object_sources
+    fn get_object_sources(&self) -> Iter<'_, Arc<dyn OcpSourceOperations>>;
+    fn get_object_source_count(&self) -> usize;
+}
+
+/// Operations required for an OCP source
+pub trait OcpSourceOperations: Debug + Send + Sync {
```
00xc: Or just `OcpSource`?

kernel/src/protocols/ocp/source.rs
```diff
@@ -0,0 +1,91 @@
...
+    SInteger8Bit = 2,
+    SInteger16Bit = 3,
+    SInteger32Bit = 4,
+    SInteger64Bit = 5,
+}
+
+/// OCP source details structure.
+#[repr(C)]
+#[derive(Debug, IntoBytes, Immutable)]
+pub struct OcpSource {
```
00xc: Or perhaps `OcpSourceInfo` / `OcpSourceDetails`?

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,276 @@
...
+
+pub fn get_ocp_object(name: &str) -> Option<Arc<dyn OcpObjectOperations>> {
+    let map = OCP_SOURCES.lock_read();
+    for obj in map.iter() {
+        if obj.get_object_name() == name {
+            return Some(obj.clone());
+        }
+    }
+    None
+}
```
00xc:
```suggestion
pub fn get_ocp_object(name: &str) -> Option<Arc<dyn OcpObjectOperations>> {
    OCP_SOURCES
        .lock_read()
        .find(|obj| obj.get_object_name() == name)
        .map(|obj| obj.clone())
}
```

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,276 @@
...
+            {
+                break;
+            }
+            entries_written += 1;
+        }
+        (
+            entries_written * OCP_SOURCE_SIZE,
+            sources_len * OCP_SOURCE_SIZE,
+        )
+    };
```
00xc: These branches are really long, maybe you can factor them out into separate helper functions.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -20,10 +20,11 @@ use crate::{
 
 use super::source::{OCP_NAME_LEN, OCP_SOURCE_SIZE, OcpSource};
 
-use core::ffi::CStr;
+use core::{ffi::CStr, str};
 
 // OCP protocol services
 const SVSM_OCP_LIST: u32 = 0;
+const SVSM_OCP_READ: u32 = 1;
```
00xc, line 26: The message for the commit introducing this constant says this should be 2, it's probably a leftover.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -153,9 +154,76 @@ fn ocp_list_request(params: &mut RequestParams) -> Result<(), SvsmReqError> {
...
+        .ok_or(SvsmReqError::invalid_parameter())?;
+
+    let obj_name =
+        str::from_utf8(&bytes[..slash_pos]).map_err(|_| SvsmReqError::invalid_parameter())?;
+
+    let source_name =
+        str::from_utf8(&bytes[slash_pos + 1..]).map_err(|_| SvsmReqError::invalid_parameter())?;
+
+    Ok((obj_name, source_name))
+}
```
00xc: I think you could also convert the whole thing to a str via `CStr::to_str()` and then do `str.split_once()`.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -220,10 +221,52 @@ fn ocp_read_request(params: &mut RequestParams) -> Result<(), SvsmReqError> {
...
+    let Some(object) = get_ocp_object(obj_name) else {
+        return Err(SvsmReqError::invalid_parameter());
+    };
+
+    let bytes_copied = object.write(offset, gpa_buffer, bytes_to_write, source_name)?;
+
+    params.r8 = bytes_copied as u64;
+
+    Ok(())
+}
```
00xc, line 267: Correct me if I'm wrong, but it seems that a lot of this logic is shared with `ocp_read_request()`. Maybe you can extract into a helper function the part regarding reading the name from guest memory, parsing the name and source, and getting the corresponding object.

kernel/src/protocols/ocp/source.rs
00xc: Or `OcpSourceDetails` / `OcpSourceInfo`? I think this name is better suited for the trait. This just represents information about a source, but it is not the source object itself.

n-ramacciotti: Yes, it makes sense! I'll change this to `OcpSourceInfo` and use the other for the trait

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: Right!

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: Yes, thanks for noticing!

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: It makes much more sense! I can also use the new helper to extract the source directly instead of the object and get rid of the `read` and `write` methods in the `object` trait!

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: Right, thanks!

n-ramacciotti:
v16:
- addressed @00xc's comments:
  - Improved comments
  - Changed `OcpSource` in `OcpSourceInfo` and used the former for the trait. Moved the trait in `source.rs`
  - Added utility to extract `source` and `object` from the guest provided gpa
  - Removed `read` and `write` methods from `OcpObject` trait.
  - Refactored `list` branches in two helpers
  - Improvement on `ObjectName/SourceName` names extraction.
- Added utilities to check if the offset and size required are aligned to source type.
- Renamed trait methods to reduce noise

TODO:
- [ ] check object uniqueness
- [ ] change `get_sources` to avoid `Iter`

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,276 @@
...
+    let name_ptr = name_guard.guest_slice::<u8>(gpa_obj_name.page_offset(), OCP_NAME_LEN)?;
+    name_ptr.read_to_slice(&mut name_slice)?;
+
+    let obj_name = CStr::from_bytes_until_nul(&name_slice)
+        .map_err(|_| SvsmReqError::invalid_parameter())?
+        .to_str()
+        .map_err(|_| SvsmReqError::invalid_parameter())?;
+
+    get_ocp_object(obj_name).ok_or(SvsmReqError::invalid_parameter())
+}
```
00xc, line 76: I think the guest access logic is equivalent to `read_from_guest()`. Also, the error conversion can be collapsed into a single `map_err()`:

```suggestion
fn get_requested_object(gpa: PhysAddr) -> Result<Arc<dyn OcpObject>, SvsmReqError> {
    let name_slice = read_from_guest::<[u8; OCP_NAME_LEN]>(gpa)?;
    let obj_name = CStr::from_bytes_until_nul(&name_slice)
        .and_then(|s| s.to_str())
        .map_err(|_| SvsmReqError::invalid_parameter())?;
    get_ocp_object(obj_name).ok_or(SvsmReqError::invalid_parameter())
}
```

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,276 @@
...
+        .ok_or(SvsmReqError::invalid_parameter())?;
+
+    let Some(object) = get_ocp_object(obj_name) else {
+        return Err(SvsmReqError::invalid_parameter());
+    };
+
+    object
+        .get_source(source_name)
+        .ok_or(SvsmReqError::invalid_parameter())
+}
```
00xc, line 108: Same thing here, I think you can just use `read_from_guest()`.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+        if entries_ptr
+            .write(entries_written, entry.get_info())
+            .is_err()
+        {
+            break;
+        }
+        entries_written += 1;
+    }
+    Ok((entries_written, objects_len))
```
00xc, line 131: You can use `take()` to limit the iterator. Also, shouldn't the function return an error if writing to guest memory fails? Maybe you can `map_err()` and return the right `SvsmReqError` along the number of entries successfully written.

```suggestion
    for entry in objects.iter().take(num_entries) {
        entries_ptr
            .write(entries_written, entry.get_info())
            .map_err(...)?;
        entries_written += 1;
    }
    Ok((entries_written, objects_len))
```

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,276 @@
...
+            break;
+        }
+        if entries_ptr
+            .write(entries_written, entry.get_info())
+            .is_err()
+        {
+            break;
+        }
+        entries_written += 1;
+    }
```
00xc, line 156: Same thing here, the iterator can be cut short with `take()`, and failing to write should probably result in an error.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    let buffer_size = (params.r8 & LOW_32_BITS) as usize;
+
+    if buffer_size == 0 || buffer_size > OCP_BUFFER_MAX_SIZE {
+        return Err(SvsmReqError::invalid_parameter());
+    }
+
+    if !buffer_size.is_multiple_of(OCP_SOURCE_SIZE) {
+        return Err(SvsmReqError::invalid_parameter());
+    }
```
00xc, line 177: This is a stylistic choice, but I would prefer these checks to be merged into a single if statement.

kernel/src/ocp_test.rs
```diff
@@ -0,0 +1,150 @@
...
+use crate::address::PhysAddr;
+use crate::fs::{FsObj, GuestBuffer, open_read};
+use crate::mm::guestmem::copy_slice_to_guest;
+use crate::protocols::errors::SvsmReqError;
+
+use core::slice::Iter;
+
+#[derive(Debug)]
+struct OcpSvsmObject {
+    ocp_source_entries: Vec<Arc<dyn OcpSource>>,
```
00xc, line 29: Or just `sources`?
```suggestion
    sources: Vec<Arc<dyn OcpSource>>,
```

kernel/src/protocols/ocp/source.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// Operations required for an OCP source
+pub trait OcpSource: Debug + Send + Sync {
+    fn read(&self, _offset: u32, _gpa: PhysAddr, _size: u32) -> Result<u32, SvsmReqError> {
+        Err(SvsmReqError::unsupported_call())
+    }
+
+    fn write(&self, _offset: u32, _gpa: PhysAddr, _size: u32) -> Result<u32, SvsmReqError> {
+        Err(SvsmReqError::unsupported_call())
+    }
```
00xc, line 136: These two methods have somewhat confusing names. For example, it is not clear if `read()` means "read from the provided GPA" or "read the source into the provided GPA".

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: > I think the guest access logic is equivalent to `read_from_guest()`

Right!

> Also, the error conversion can be collapsed into a single `map_err()`:

It seems that this cannot be done as `from_bytes_until_nul` and `to_str`  have different errors and `and_then` requires them to be the same:

expected `Result<_, FromBytesUntilNulError>`, found `Result<&str, Utf8Error>`

kernel/src/protocols/ocp/source.rs
n-ramacciotti: Right, I'll change them to `write_to_guest` and `read_from_guest`

kernel/src/ocp_test.rs
n-ramacciotti: I can change it, but my original plan was to use this PR for the basic protocol infrastructure and open another one to add the initial sources. This was only introduced to have some sources for testing. I should probably move the description of these sources to this commit instead of the one that introduced the general documentation.

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: > Also, shouldn't the function return an error if writing to guest memory fails? 

Yes, right

>  Maybe you can `map_err()` and return the right `SvsmReqError` along the number of entries successfully written.

Do you mean something like `Result<(usize, usize), (SvsmReqError, usize, usize)>` and logging the error while still returning to the guest the number of entries already written?

00xc: Yeah, I think you can even do with just one of those usizes. The function should just report how many entries were written out of those requested, right?

kernel/src/protocols/ocp/requests.rs
00xc: > It seems that this cannot be done as from_bytes_until_nul and to_str have different errors and and_then requires them to be the same:

Ah, right. You can convert them both to options via `ok()` if you want, then you can do `map()` + `ok_or()`, but I am fine leaving it as it is.

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: The list operation should also return the total number of entries available, so that the guest can retry the operation, as the requested number can be less than the total number

n-ramacciotti:
v17:
- addressed @00xc's comments:
  - Renamed `OcpSource` trait methods to `write_to_guest` and `read_from_guest`
  - Renamed `ocp_source_entries` in `sources`
  - Merged checks in single if where possible
  - Used `read_from_guest` instead of the manual counterpart
  - Avoided multiple `map_err()`
  - Now a failure when writing entries returns an error and the number of entries written up to that moment + total entries. The error is logged and the number of entries written is returned to the guest
  - Introduced `take` on iterator
- Used `checked_page_align_up` instead of `page_align_up`

TODO:
- [ ] check object uniqueness.
- [ ] Remove trait method that returns iterator? (https://github.com/coconut-svsm/svsm/pull/1138#discussion_r4063075764)

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: I applied this change. However, I'm not sure if `get_sources` should return an iterator, especially in the case that the sources of an object can change

00xc: I am not sure what use cases you have in mind, but I think an iterator is fine. You can make it generic, so that the backing storage of the sources can be changed transparently:

```rust
fn iter_sources(&self) -> impl Iterator<Item = &Arc<dyn OcpSource>>>;
```

Alternatively you can add a `for_each()` method to `OcpObject` that just takes a callback and applies it to every item, but that is less flexible, as it does not allow the caller to stop the iteration early, unless you establish a particular sentintel value, e.g.

```rust
/// Applies `f` to every source until `f` returns `Some`.
fn for_each_source<F, R>(&self, f: F) -> Option<R>
where
    F: Fn(&Arc<dyn OcpSource>) -> Option<R>;
```

But if you have a particular way of traversing the list of sources, then I don't see why you would not be able to create an iterator out of that and go with the first option above.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -67,8 +77,102 @@ fn get_requested_source(gpa_id: PhysAddr) -> Result<Arc<dyn OcpSource>, SvsmReqE
         .ok_or(SvsmReqError::invalid_parameter())
 }
 
-pub fn ocp_protocol_request(request: u32, _params: &mut RequestParams) -> Result<(), SvsmReqError> {
+fn list_objects_info(
+    entries_ptr: GuestPtr<'_, [OcpSourceInfo]>,
+    num_entries: usize,
+) -> Result<(usize, usize), (SvsmReqError, usize, usize)> {
```
00xc, line 83: Let's document the meaning of the return value, since it's not a common function signature.

kernel/src/protocols/ocp/requests.rs
```diff
@@ -67,8 +77,102 @@ fn get_requested_source(gpa_id: PhysAddr) -> Result<Arc<dyn OcpSource>, SvsmReqE
...
+        entries_written += 1;
+    }
+    Ok((entries_written, objects_len))
+}
+
+fn list_object_sources_info(
+    entries_ptr: GuestPtr<'_, [OcpSourceInfo]>,
+    gpa_obj_name: PhysAddr,
+    num_entries: usize,
+) -> Result<(usize, usize), (SvsmReqError, usize, usize)> {
```
00xc, line 102: Same here

kernel/src/protocols/ocp/requests.rs
```diff
@@ -0,0 +1,248 @@
...
+    let (entries_written, total_entries) = match result {
+        Ok(values) => values,
+        Err((e, entries, total)) => {
+            // In this branch, no entries means the object was not found
+            if total == 0 {
+                return Err(e);
+            }
+            log::error!("Failed to write one entries to guest memory: {e:?}");
+            (entries, total)
+        }
```
00xc, line 170: You can match the no entries branch separately if you want, e.g.

```suggestion
        Ok(values) => values,
        Err(e, _, 0) => return Err(e),
        Err((e, entries, total)) => {
            log::error!("Failed to write one entries to guest memory: {e:?}");
            (entries, total)
        }
```

kernel/src/protocols/ocp/requests.rs
n-ramacciotti: > But if you have a particular way of traversing the list of sources, then I don't see why you would not be able to create an iterator out of that and go with the first option above.

I'll follow that option, but `impl Trait` cannot be used with `dyn`. So I think we can use a `box` to wrap the iterator

00xc: Ah right. Maybe the for_each() approach is better then, up to you.

n-ramacciotti:
v18:
- addressed @00xc's comments:
  - added documentation for `Result<(usize,usize),(SvsmReqError,usize,usize)>` and divided the error branch in two parts in `list_request`
  - Introduced `for_each_source` instead of `get_sources`.
- Introduced object uniqueness based on name

kernel/src/protocols/ocp/source.rs
```diff
@@ -0,0 +1,151 @@
...
+        && mem::offset_of!(OcpSourceInfo, name) == 0x08
+        && mem::size_of::<OcpSourceInfo>() == OCP_SOURCE_SIZE
+);
+
+/// Operations required for an OCP object
+pub trait OcpObject: Debug + Send + Sync {
+    fn get_name(&self) -> &str;
+    fn get_info(&self) -> &OcpSourceInfo;
+    fn get_source(&self, name: &str) -> Option<Arc<dyn OcpSource>>;
+    /// Applies `f` to every source until `f` returns `false`.
```
00xc, line 121:
```suggestion
    /// Applies `f` to every source until `f` returns `Ok(false)` or an error.
    /// In the latter case, the error is returned.
```

PR #1141, opened 2026-06-25
xbuild/fs: Handle multi-level path

n-ramacciotti:
Currently, when attempting to build the filesystem, the build fails if the specified file path is multi-level (like `test/usermodule`) as it is not guaranteed that all the directories exist.

Check for the presence of the directories and create them if they are not already present before copying the file.

Avoid modifying `dst` directly as a single `push` with multiple levels requires multiple `pop` operations.

### DIFF
--- a/xbuild/src/fs.rs
+++ b/xbuild/src/fs.rs
@@ -27,7 +27,7 @@ impl FsConfig {
     pub fn build(
         &self,
         args: &Args,
-        mut dst: PathBuf,
+        dst: PathBuf,
         cmd_feats: &mut Features,
     ) -> BuildResult<Option<PathBuf>> {
         if dst.try_exists()? {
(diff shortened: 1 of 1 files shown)


xbuild/src/fs.rs
```diff
@@ -51,19 +51,26 @@ impl FsConfig {
...
             if dst_file.starts_with("/") {
                 dst_file = dst_file.strip_prefix("/").unwrap();
             }
-            dst.push(dst_file);
-            comp.config.objcopy.copy(&bin, &dst, args)?;
+
+            let current_dst = dst.join(dst_file);
+
+            if let Some(parent) = current_dst.parent() {
+                if !parent.exists() {
```
00xc: I think you can safely drop this check, `create_dir_all()` will be a no-op if the directory already exists.

n-ramacciotti:
v2:
- removed check on `parent.exists()` @00xc

00xc (approved):

PR #1147, opened 2026-06-30
scripts: align pre-commit SPDX check with check-spdx.sh

tanish111:
The pre-commit hook previously used its own check_file_header() that enforced a rigid 5-line layout (SPDX, blank comment, Copyright, blank comment, Author:) on every staged .rs file. That broke commits on unrelated changes because many existing files in the tree never had an Author: line or did not follow that exact layout — a full-tree audit found 100+ files missing Copyright or Author in the expected form.

This PR drops the Author and fixed-layout checks and reuses scripts/check-spdx.sh for SPDX validation instead of duplicating the logic. A new --files option is added to check-spdx.sh to validate a list of files; the pre-commit hook calls it for each staged file, using the same SPDX + MIT rules already applied in CI. rustfmt and clippy checks are unchanged.

It also prints an additional line _All checked file(s) contain a valid SPDX-License-Identifier header._ if header check passes for all staged files.

Reported-by: @luigix25

### DIFF
--- a/scripts/check-spdx.sh
+++ b/scripts/check-spdx.sh
@@ -7,6 +7,7 @@
 #
 # Modes:
 #   --all				check every tracked file.
+#   --files <path> [<path> ...]	check one or more files.
 #   <start_commit> [<end_commit>]	check only files added in this commit range.
 #					<end_commit> defaults to HEAD.
 
@@ -17,12 +18,20 @@ cd "$(git rev-parse --show-toplevel)"
 
 if [ $# -lt 1 ]; then
 	echo "Usage: $0 --all"
+	echo "       $0 --files <path> [<path> ...]"
 	echo "       $0 <start_commit> [<end_commit>]"
 	exit 1
 fi
 
 if [ "$1" = "--all" ]; then
 	files=$(git ls-files)
+elif [ "$1" = "--files" ]; then
+	shift
+	if [ $# -lt 1 ]; then
+		echo "Usage: $0 --files <path> [<path> ...]"
+		exit 1
+	fi
+	files=$*
 else
 	start=$1
 	end=${2:-HEAD}
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -5,29 +5,12 @@
 #
 # Author: Joerg Roedel <jroedel@suse.de>
 
-check_file_header() {
-	header_warning="Header format incorrect in $1, follow the header structure in other files"
-	IFS=$'\n'
-	arr=($(head -n 5 "$1"))
-	if  ! [[ "${arr[0]}" == "// SPDX-License-Identifier: MIT OR Apache-2.0" ||
-	         "${arr[0]}" == "// SPDX-License-Identifier: MIT" ]] ; then
-		echo "${header_warning}"
-		return 1
-	fi
-	if  ! [[ "${arr[2]}" =~ ^//[[:space:]]Copyright ]] ; then
-		echo "${header_warning}"
-		return 1
-	fi
-	if  ! [[ "${arr[4]}" =~ ^//[[:space:]]Author: ]] ; then
-		echo "${header_warning}"
-		return 1
-	fi
-	return 0
-}
-
 RET=0
+CHECK_SPDX="$(git rev-parse --show-toplevel)/scripts/check-spdx.sh"
 
-for file in `git diff --diff-filter=d --name-only --staged`; do
+staged_files=$(git diff --diff-filter=d --name-only --staged)
+
+for file in $staged_files; do
     ext=${file##*.}
     if [ "$ext" == "rs" ]; then
         rustfmt --check $file > /dev/null 2>&1
@@ -35,12 +18,12 @@ for file in `git diff --diff-filter=d --name-only --staged`; do
             echo "$file needs rustfmt checking"
             RET=1
         fi
-	check_file_header $file
-        if [ "$?" == "1" ]; then
-            RET=1
-        fi
     fi
 done
+
+if [ -n "$staged_files" ]; then
+    "$CHECK_SPDX" --files $staged_files || RET=1
+fi
 
 make clippy || exit 1
 


tanish111:
@luigix25 @stefano-garzarella here is a fix PR for pre-commit script. I ran the script for all *.rs files in repo and here is the report.
SPDX header and clippy passed for all files.
rustfmt fails for 5 files they are boot/bootimg/src/lib.rs, elf/src/file.rs, elf/src/lib.rs, kernel/src/address.verus.rs and kernel/src/mm/alloc_inner.verus.rs. 
Should I fix fmt in the same PR or keep files unchanged?

luigix25 (changes requested):
We have a script in place to check the spdx_header, I think we can reuse it here.

tanish111:
@luigix25 I considered it, but it does not fit the pre-commit flow. The script only supports checking all tracked files or a commit range, which works for CI because it has the PR base and head SHAs. In a pre-commit hook, no new commit exists yet, and we only want to validate the currently staged files (`git diff --cached`). Using `--all` would unnecessarily scan the entire tree on every commit attempt. For that reason, the pre-commit hook inlines the SPDX check while applying the same validation rules as `check-spdx.sh`.

tanish111:
Also The old check_file_header() enforced a rigid 5-line layout for every staged .rs file: an SPDX license line, a blank comment line, a Copyright line, another blank comment line, and an Author: line. That broke pre-commit on unrelated commits because many tracked files never had an Author: line, and some did not follow that exact layout. A full-tree audit found 100+ files missing Copyright or Author in the expected form, so any commit touching those files would fail even when the change itself was fine.
The hook now only validates SPDX and MIT in the first 5 lines, matching scripts/check-spdx.sh. The rustfmt and clippy checks are unchanged. Copyright and Author lines are still encouraged for new files but are no longer enforced at commit time.

luigix25:
> @luigix25 I considered it, but it does not fit the pre-commit flow. The script only supports checking all tracked files or a commit range, which works for CI because it has the PR base and head SHAs. In a pre-commit hook, no new commit exists yet, and we only want to validate the currently staged files (`git diff --cached`). Using `--all` would unnecessarily scan the entire tree on every commit attempt. For that reason, the pre-commit hook inlines the SPDX check while applying the same validation rules as `check-spdx.sh`.

we can use `git diff --staged`, no? (should be the same as --cached)

I know that the script does not support it, but we can add it. In this way we don't need to maintain 2 files. There is an high risk that we will forget to update one and they go out of sync

tanish111:
@luigix25 yes it's a high risk considering script is part of CI but much better than maintaining 2 files. I couldn't make a call thanks for helping pushing the updated script.

luigix25:
> rustfmt fails for 5 files they are boot/bootimg/src/lib.rs, elf/src/file.rs, elf/src/lib.rs, kernel/src/address.verus.rs and kernel/src/mm/alloc_inner.verus.rs.

No, this is a bug in our rustfmt, I'll open a PR.

tanish111:
@luigix25 pushed updated code using script.

luigix25:
@tanish111 sorry Tanish, I got confused. The solution is much easier: we can just add a `--file` option to check-spdx, that checks the spdx header for that file and in the loop that just call check-spdx on that file.

untested: 
```bash
+elif [ "$1" = "--file" ]; then
+       if [ $# -lt 2 ]; then
+               echo "Usage: $0 --file <path>"
+               exit 1
+       fi
+       files=$2
```

the `--staged` option is useless, we already have a way to detect the modified files.

scripts/check-spdx.sh
```diff
@@ -7,6 +7,7 @@
 #
 # Modes:
 #   --all				check every tracked file.
+#   --staged				check staged files (git diff --staged).
```
luigix25: be careful with formatting

luigix25 (commented):
It's good practice for commit message to not exceed 80 columns

tanish111:
@luigix25 pushed suggested changes.

luigix25 (approved):

tanish111:
@stefano-garzarella can you review this PR as well?
(@luigix25 has approved the changes)

scripts/pre-commit
```diff
@@ -35,11 +16,8 @@ for file in `git diff --diff-filter=d --name-only --staged`; do
             echo "$file needs rustfmt checking"
             RET=1
         fi
-	check_file_header $file
```
stefano-garzarella, line 38: IIUC before this patch we did the check only for rust files, while now we do for all files. Is that on purpose?

scripts/check-spdx.sh
stefano-garzarella, line 74: This message now will be printed to every file we check since we are going to call this script for each file in the stage, is really what we want or should we add a `--quiet` parameter to use in this case?

scripts/check-spdx.sh
```diff
@@ -7,6 +7,7 @@
 #
 # Modes:
 #   --all				check every tracked file.
+#   --file <path>		check a single file.
```
stefano-garzarella: instead of `--file`, what about `--files` so we can pass multiple files in a single call?

scripts/pre-commit
luigix25: to be fair, that script checks the file extension, so I think it makes more sense in this way

scripts/check-spdx.sh
stefano-garzarella: Or support multiples `--file` (e.g. ./scripts/check-spdx.sh --file file1 --file file2)`. I mean the goal is to call this script only one time from the other script.

scripts/pre-commit
stefano-garzarella: ah okay, since it's a behaviour change, IMO should be at least mentioned in the commit description or moved to another commit (not a strong opinion about splitting the commits since it's just an utility script)

scripts/check-spdx.sh
tanish111: @stefano-garzarella I had pushed another implementation which uses script only once.
Link:- https://github.com/coconut-svsm/svsm/commit/ee99d18fe7fe6662478f2631249f242726bf66d6
do we want something like this?
cc: @luigix25

scripts/check-spdx.sh
tanish111: @stefano-garzarella yes it's worth adding --quiet. Also clippy dumps alot in stdout should we suppress that as well for pre-commits as for every commit or rebase this hook floods terminal

scripts/check-spdx.sh
stefano-garzarella: @tanish111 I don't have a strong opinion on that, both your version or supporting multiple `--file` are fine by me.

scripts/check-spdx.sh
stefano-garzarella: @tanish111 I'm not an user of this script, so not a strong opinion on my side, but seems reasonable to me reduce noise and print less.

scripts/check-spdx.sh
tanish111: @stefano-garzarella I updated the code with --files <file-1> <file-2> ... 
This makes a single call to the script and adds new functionality to header scripts, allowing individual files to be tested without staging/committing them, which was previously not supported.

scripts/check-spdx.sh
tanish111: @stefano-garzarella I tried this initially because I thought it would be useful, but suppressing the Clippy output makes it look like the terminal is stuck for 30–40 seconds until all Clippy checks finish, which is quite annoying. Also, we don't need `--quiet` for the header script anymore since, after the recent changes, it only runs once instead of once per file.

scripts/pre-commit
```diff
@@ -5,44 +5,26 @@
...
-        if [ "$?" == "1" ]; then
-            RET=1
-        fi
     fi
 done
 
-make clippy || exit 1
```
luigix25, line 45: why you dropped clippy?

tanish111: @luigix25 sorry this was not intended rebasing to remove quiet mode changes dropped entire line.

tanish111: it's fixed.

luigix25 (approved):

stefano-garzarella (approved):

stefano-garzarella:
LGTM, but I'm not using it, so I'll leave to @00xc and/or @joergroedel to merge this.

joergroedel (approved):

PR #1148, opened 2026-06-30
rustfmt: Add edition to rustfmt.toml

luigix25:
Set the edition field to "2024" to match the workspace Cargo.toml edition setting. Unlike cargo fmt, standalone rustfmt does not read the edition from Cargo.toml and defaults to 2015.

Adding the edition explicitly ensures correct syntax parsing when using rustfmt directly.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,1 +1,2 @@
+edition = "2024"
 style_edition = "2024"


00xc (approved):
Huh, interesting. I'm sure there's a reason to have separate items for `edition` and `style_edition`, but it's not obvious to me. In any case, LGTM.

luigix25:
> Huh, interesting. I'm sure there's a reason to have separate items for `edition` and `style_edition`, but it's not obvious to me. In any case, LGTM.

Same for me, but if you run `rustfmt elf/src/file.rs --check` it outputs some formatting errors.

PR #1150, opened 2026-06-30
irq: assume the availablity of interrupts on all systems

msft-jlange:
Now that SNP requires the use of restricted injection, it is safe to assume that all all platforms support the use of SVSM interrupts, and all abstractions can be removed.

### DIFF
--- a/kernel/src/cpu/ipi.rs
+++ b/kernel/src/cpu/ipi.rs
@@ -10,7 +10,6 @@ use super::idt::common::IPI_VECTOR;
 use super::percpu::{PERCPU_AREAS, PerCpuShared, this_cpu};
 use super::x86::apic_post_irq;
 use crate::error::SvsmError;
-use crate::platform::SVSM_PLATFORM;
 use crate::types::{TPR_IPI, TPR_SYNCH};
 use crate::utils::{ScopedMut, ScopedRef};
 
@@ -18,7 +17,8 @@ use core::cell::{Cell, UnsafeCell};
 use core::mem;
 use core::mem::MaybeUninit;
 use core::ptr;
-use core::sync::atomic::{AtomicU32, AtomicUsize, Ordering};
+use core::sync::atomic::AtomicUsize;
+use core::sync::atomic::Ordering;
 use cpuarch::x86apic::ApicIcr;
 use cpuarch::x86apic::IcrDestFmt;
 
@@ -326,8 +326,6 @@ impl Default for IpiBoard {
 // is consumed as a dynamic dispatch trait to avoid explosion due to multiple
 // generic message implementations.
 fn send_ipi(target_set: IpiTarget<'_>, ipi_helper: &mut dyn IpiHelper) {
-    assert!(ipi_available());
-
     let sender_cpu_index = this_cpu().get_cpu_index();
 
     // Raise TPR to synch level to prevent reentrant attempts to send an IPI.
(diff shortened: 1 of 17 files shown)


kernel/src/task/tasks.rs
```diff
@@ -710,11 +709,7 @@ impl Task {
         let (mapping, bounds) = Task::allocate_stack_common()?;
         // Do not run user-mode with IRQs enabled on platforms which are not
         // ready for it.
```
stefano-garzarella: I guess we need to update (or remove) this comment.

kernel/src/platform/native.rs
```diff
@@ -46,11 +46,11 @@ use crate::mm::virt_to_phys;
 #[cfg(test)]
 use bootdefs::platform::SvsmPlatformType;
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Default)]
```
stefano-garzarella: Are we sure about deriving `Default` here?

Asking because in `NativePlatform::new()` we have the X2APIC check that will be bypassed if `NativePlatform::default()` will be used.

kernel/src/cpu/ipi.rs
```diff
@@ -629,26 +628,6 @@ pub fn ipi_available() -> bool {
...
 }
 
-/// Request IPI blocking on the current CPU and wait until all other CPUs have
-/// done the same.
-pub fn wait_for_ipi_block() {
-    // Mark this CPU as wanting to block IPIs and wait until all other CPUs
-    // have done the same.  Note that while waiting, additional IPIs may still
-    // be received, which is necessary because other CPUs may not have yet
-    // gotten to the point that they are willing to stop using IPIs.
-    IPI_AVAILABLE_CPU_COUNT.fetch_sub(1, Ordering::Release);
```
stefano-garzarella, line 639: `IPI_AVAILABLE_CPU_COUNT` is now only incremented. Do we still need it?

kernel/src/cpu/ipi.rs
stefano-garzarella: I think this is related to `IPI_AVAILABLE_CPU_COUNT` I mentioned before, so I'm not sure if we still need `ipi_available()` and `ipi_start_cpu()` as well.

stefano-garzarella (commented):
Should we also clean boot_params and igvm_builder?
```
$ git grep suppress_svsm_interrupts
boot/defs/src/boot_params.rs:    pub suppress_svsm_interrupts_on_snp: u8,
kernel/src/boot_params.rs:    pub fn suppress_svsm_interrupts_on_snp(&self) -> bool {
kernel/src/boot_params.rs:        self.boot_param_block.suppress_svsm_interrupts_on_snp != 0
tools/igvmbuilder/src/igvm_builder.rs:        let suppress_svsm_interrupts_on_snp = match self.options.hypervisor {
tools/igvmbuilder/src/igvm_builder.rs:            suppress_svsm_interrupts_on_snp,
```

kernel/src/platform/native.rs
msft-jlange: Clippy complains when a type implements `new()` with no parameters but does not implement `Default`.  I believe that `derive(Default)` will cause `Default::default()` to result in a call to `new()` - I don't know how else to interpret the clippy lint.

msft-jlange:
> Should we also clean boot_params and igvm_builder?

Yes, good catch!

kernel/src/cpu/ipi.rs
msft-jlange: > `IPI_AVAILABLE_CPU_COUNT` is now only incremented. Do we still need it?

We do not.

kernel/src/cpu/ipi.rs
msft-jlange: > I think this is related to `IPI_AVAILABLE_CPU_COUNT` I mentioned before, so I'm not sure if we still need `ipi_available()` and `ipi_start_cpu()` as well.

They are related and this can be removed.

kernel/src/platform/native.rs
stefano-garzarella: > Clippy complains when a type implements `new()` with no parameters but does not implement `Default`. I believe that `derive(Default)` will cause `Default::default()` to result in a call to `new()` - I don't know how else to interpret the clippy lint.

I tried in this playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e991c60ed09b58cb0be332b673778fa9

The print in `new()` is skipped when creating the instance with `default()`.
IIUC `#[derive(Default)]` does not call `new()`, but it generates its own implementation that constructs the struct by calling `Default::default()` on each field. Since `NativePlatform` doesn't have any fiels, I guess it justs returns `Self {}`.

IMHO, if we care about the check in `NativePlatform::new()`, we should do something similar of what you did for `SvsmPlatformCell` in https://github.com/coconut-svsm/svsm/pull/1150/changes#diff-e30e223f5755a1686826aa3a8381b9005d995d6a618e503f0312642b79206de4R342-R346 which is also what they suggest in the lint description: https://rust-lang.github.io/rust-clippy/master/#new_without_default

Or just use `#[allow(clippy::new_without_default)]`

msft-jlange: OK, I understand the issue now.

> Or just use `#[allow(clippy::new_without_default)]`

I don't know why this lint is enabled for us, but I'm extremely reluctant to suppress individual lints in individual places without a clear justification, and I don't know how to justify this suppression for this particular type.  I think it would be reasonable to open a separate PR to disable this lint globally; that's something we discuss.

Meanwhile, I'm adding an explicit `Default` implementation as you have suggested.

stefano-garzarella: > > Or just use `#[allow(clippy::new_without_default)]`
> 
> I don't know why this lint is enabled for us, but I'm extremely reluctant to suppress individual lints in individual places without a clear justification, and I don't know how to justify this suppression for this particular type.

Yeah, agree.

> I think it would be reasonable to open a separate PR to disable this lint globally; that's something we discuss.

Agree also on this.

> 
> Meanwhile, I'm adding an explicit `Default` implementation as you have suggested.

Thanks!

stefano-garzarella (dismissed):
LGTM!

@msft-jlange sorry, I merged something and we have conflicts now. Also I was having trouble booting Linux with KVM and planes support, but after rebasing to current main, everything is running smoothly.

joergroedel (approved):

PR #1152, opened 2026-07-02
kernel: remove several unnecessary heap allocations

00xc:
In preparation to introduce fallible allocations, remove several uses of unnecessary heap allocations, as these will introduce extra points of failure without good reason. This will also minimize the diff when switching to the new infrastructure.


osteffenrh (approved):

kernel/src/vmm/execloop.rs
```diff
@@ -4,12 +4,12 @@
 //
 // Author: Jon Lange (jlange@microsoft.com)
 
-use super::{GuestExitMessage, GuestRegister, set_guest_register};
```
stefano-garzarella, line 7: After this patch `GuestRegister` and `set_guest_register()` are unused, so we can potentially remove the entire `kernel/src/vmm/registers.rs`.

Does it makes sense, or we should leave it for future useage?

stefano-garzarella (approved):
LGTM, just a minor cleanup that can be done also later.

kernel/src/vmm/execloop.rs
msft-jlange: Let's leave these for future use.  I think we can be pretty confident that these will need to come back when paravisor support is implemented.  If the future support doesn't rely on these routines, they can be removed at that time.

PR #1153, opened 2026-07-03
Move ELF loading into context of new task

joergroedel:
Cleanup the task setup and launch process and move loading of the ELF file for a user-space process into the context of the user-mode task.

### DIFF
--- a/kernel/src/cpu/shadow_stack.rs
+++ b/kernel/src/cpu/shadow_stack.rs
@@ -97,7 +97,9 @@ pub enum ShadowStackInit {
         /// The address of the first instruction that will be executed by the task.
         entry_return: usize,
         /// The address of the function that's executed when the task exits.
-        exit_return: Option<usize>,
+        exit_return: usize,
+        /// Whether there is an iret frame on the bottom of the stack.
+        iret_frame: bool,
     },
     /// A shadow stack to be used during context switches.
     ///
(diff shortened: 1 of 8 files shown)


kernel/src/task/exec.rs
```diff
@@ -44,27 +58,20 @@ fn task_name(binary: &str) -> String {
...
-/// # Arguments
-///
-/// * binary: Path to file in the file-system
-///
-/// # Returns
-///
-/// [`Ok(TaskPointer)`] on success, [`Err(SvsmError)`] on failure.
-pub fn exec_user(binary: &str, root: Arc<dyn Directory>) -> Result<TaskPointer, SvsmError> {
-    let fh = open_read(binary)?;
+pub fn exec(info: UserExecInfo) -> Result<u64, SvsmError> {
```
luigix25, line 71: you are adding a new pub function here, is it worth to add some documentation? or maybe mark it as non-pub?

kernel/src/task/exec.rs
```diff
@@ -44,27 +58,20 @@ fn task_name(binary: &str) -> String {
...
+    let fh = open_read(&info.binary)?;
     let file_size = fh.size();
 
     let current_task = current_task();
+
```
luigix25, line 76: nit: blank line

kernel/src/task/exec.rs
```diff
@@ -44,27 +58,20 @@ fn task_name(binary: &str) -> String {
...
     let vstart = current_task.mmap_kernel_guard(
         VirtAddr::new(0),
         Some(&fh),
         0,
         file_size,
         VMFileMappingFlags::Read,
     )?;
+
```
luigix25, line 84: nit: blank line

kernel/src/task/tasks.rs
```diff
@@ -1042,25 +1025,56 @@ unsafe fn setup_new_task_common(xsa_addr: u64, prev_task: Option<TaskPointer>) {
...
-    prev: usize,
-    entry: fn(T),
-    xsa_addr: u64,
-    start_parameter: usize,
+/// This is called from assembly code. The caller needs to ensure the
+/// parameters are passed correctly.
+#[unsafe(no_mangle)]
+unsafe fn run_user_task(
+    info_ptr: *mut UserExecInfo,
+    _unsused: u64,
```
luigix25: /s/_unsused/_unused?

kernel/src/task/tasks.rs
```diff
@@ -1042,25 +1025,56 @@ unsafe fn setup_new_task_common(xsa_addr: u64, prev_task: Option<TaskPointer>) {
...
+        Err(e) => {
+            log::error!("Failed to load ELF binary: {e:?}");
+            terminate();
+        }
+    };
+
+    // Do not run user-mode with IRQs enabled on platforms which are not
+    // ready for it.
+    let iret_rflags: usize = if SVSM_PLATFORM.use_interrupts() {
+        2 | EFLAGS_IF
```
luigix25: is it worth a const for this 2?

kernel/src/task/tasks.rs
```diff
@@ -1042,25 +1025,56 @@ unsafe fn setup_new_task_common(xsa_addr: u64, prev_task: Option<TaskPointer>) {
...
+    } else {
+        2
+    };
+
+    ctxt.frame = X86InterruptFrame {
+        rip: entry as usize,
+        cs: (SVSM_USER_CS | 3).into(),
```
luigix25: ditto for 3

kernel/src/task/exec.rs
joergroedel: Yes. So far it is only used in the same module, but going forward there might be other uses. Added the documentation.

kernel/src/task/exec.rs
joergroedel: There for readability :)

kernel/src/task/exec.rs
joergroedel: There for readability :)

kernel/src/task/tasks.rs
joergroedel: Fixed.

kernel/src/task/tasks.rs
joergroedel: Yes, although this is pre-existing. I hacked it up, but this needs more consolidation into a control-register module. I will leave that for another PR.

kernel/src/task/tasks.rs
joergroedel: Done as well.

kernel/src/task/tasks.rs
```diff
@@ -663,41 +663,53 @@ impl Task {
...
         // defined in switch_context below.
         let stack_tos = percpu_mapping.virt_addr() + bounds.end().bits();
+
         // Make space for the task termination handler
-        let stack_offset = size_of::<KernelTaskContext>();
+        let stack_offset = size_of::<T>();
         let stack_ptr = stack_tos
             .checked_sub(stack_offset)
             .unwrap()
             .as_mut_ptr::<u8>();
```
00xc: Now you can just do `as_mut_ptr::<T>()` here. Then writing the structure to the pointer below is even simpler, since the cast is not needed, e.g.

```rust
unsafe { stack_ptr.write(*context) };
```

kernel/src/cpu/shadow_stack.rs
```diff
@@ -116,58 +118,63 @@ pub fn init_shadow_stack(
     init: ShadowStackInit,
 ) -> (Option<VirtAddr>, VirtAddr) {
     // Initialize the shadow stack.
-    let mut chunk = [0; 24];
-    let (base_token_addr, ssp) = match init {
+    let mut chunk = [0; 32];
```
00xc: There is a lot of magic numbering with 32 and 24, maybe these should be constants or derived from other constants

kernel/src/task/tasks.rs
```diff
@@ -442,11 +442,17 @@ impl Task {
...
         // Determine which kernel-mode entry/exit routines will be used for
         // this task.
-        let (entry_return, exit_return) = match args.start_info {
-            ThreadStartInfo::User(_) => (return_new_task as *const () as usize, None),
-            ThreadStartInfo::Kernel(ref info) => {
-                (info.start_routine, Some(task_exit as *const () as usize))
-            }
+        let (entry_return, exit_return, iret_frame) = match args.start_info {
+            ThreadStartInfo::User(_) => (
+                thread_entry_asm as *const () as usize,
```
00xc: `thread_entry_asm` can be pulled out of the match statement, since it's common to both arms.

kernel/src/task/tasks.rs
```diff
@@ -713,13 +720,15 @@ impl Task {
         let task_context = KernelTaskContext {
             task_context: TaskContext {
                 regs: X86TaskSwitchRegs {
-                    rsi: entry,
-                    rdx: xsa_addr,
-                    rcx: start_parameter,
+                    rsi: xsa_addr,
+                    // Put these into callee-safed registers
```
00xc: `safed` -> `saved`

kernel/src/task/tasks.rs
```diff
@@ -995,23 +1006,8 @@ fn task_attach_console() {
...
-    // xsa_addr are valid.
-    unsafe {
-        // Needs to be the first function called here.
-        setup_new_task_common(xsa_addr, complete_task_switch(prev));
-    }
-    task_attach_console();
-}
-
-unsafe fn setup_new_task_common(xsa_addr: u64, prev_task: Option<TaskPointer>) {
+unsafe fn complete_new_thread(prev: usize, xsa_addr: u64) {
```
00xc, line 1059: Some documentation about what this function does would be useful, even if it's not public

kernel/src/task/tasks.rs
```diff
@@ -1032,25 +1031,22 @@ unsafe fn setup_new_task_common(xsa_addr: u64, prev_task: Option<TaskPointer>) {
     }
 }
 
+/// Runs the first time a new task is scheduled, in the context of the new
+/// task. Any first-time initialization specific to user-tasks need to happen
+/// here.
+/// # Safety
+/// This is called from assembly code. The caller needs to ensure the
+/// parameters are passed correctly.
```
00xc, line 1097: What parameters though?

kernel/src/cpu/shadow_stack.rs
joergroedel: I realized the same, but this needs a general cleanup which I plan to do as a separate PR. It is out of scope for this one.

joergroedel: It turned out to be only one patch to clean this up, so I included it in this PR instead of opening a new one.

stefano-garzarella (approved):

PR #1154, opened 2026-07-03
Fix Guestmem label and page-fault handler

joergroedel:
Fix two issues and add missing `Copy + Clone` annotations to ELF code. The issues fixes here:

* Fix the faulting label in `copy_bytes()` so it points to the instruction that will fault.
* Fix the page-fault handler so that it correctly handles and resolves faults when kernel mode accesses user-mode addresses.

### DIFF
--- a/elf/src/program_header.rs
+++ b/elf/src/program_header.rs
@@ -15,7 +15,7 @@ use bitflags::bitflags;
 bitflags! {
 /// Attributes of an ELF64 program header, to specify whether
 /// the segment is readable, writable, and/or executable
-    #[derive(Debug)]
+    #[derive(Copy, Clone, Debug)]
     pub struct Elf64PhdrFlags : Elf64Word {
         const EXECUTE = 0x01;
         const WRITE   = 0x02;
@@ -24,7 +24,7 @@ bitflags! {
 }
 
 /// Program header entry in an ELF64 file
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
 pub struct Elf64Phdr {
     /// Type of the program header entry
     pub p_type: Elf64Word,
--- a/kernel/src/cpu/idt/svsm.rs
+++ b/kernel/src/cpu/idt/svsm.rs
@@ -277,34 +277,29 @@ extern "C" fn ex_handler_page_fault(ctxt: &mut X86ExceptionContext, vector: usiz
     let err = ctxt.error_code;
     let vaddr = VirtAddr::from(cr2);
 
-    if user_mode(ctxt) {
-        let kill_task: bool = if is_task_fault(vaddr) {
-            current_task()
-                .fault(vaddr, (err & PageFaultError::W.bits() as usize) != 0)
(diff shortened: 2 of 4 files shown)


kernel/src/mm/guestmem.rs
```diff
@@ -249,8 +249,8 @@ unsafe fn copy_bytes(src: *const u8, dst: *mut u8, size: usize) -> Result<(), Sv
     // SAFETY: Safe as long as the function's safety requirements are met. Any
     // fault that might happen is handled via the exception handlers.
     unsafe {
-        asm!("1:cld
```
00xc, line 252: You can actually remove this `cld` altogether, Rust guarantees that the direction flag is clear upon entry to the assembly block:

https://doc.rust-lang.org/reference/inline-assembly.html#r-asm.rules.x86-df

elf/src/program_header.rs
```diff
@@ -15,7 +15,7 @@ use bitflags::bitflags;
 bitflags! {
 /// Attributes of an ELF64 program header, to specify whether
 /// the segment is readable, writable, and/or executable
-    #[derive(Debug)]
+    #[derive(Copy, Clone, Debug)]
```
luigix25, line 18: might be a stupid question: why is this needed?

kernel/src/cpu/idt/svsm.rs
```diff
@@ -277,34 +277,29 @@ extern "C" fn ex_handler_page_fault(ctxt: &mut X86ExceptionContext, vector: usiz
...
-            true
-        };
+    let resolved = if is_task_fault(vaddr) {
+        current_task()
+            .fault(vaddr, (err & PageFaultError::W.bits() as usize) != 0)
+            .is_ok()
+    } else {
+        this_cpu()
+            .handle_pf(vaddr, (err & PageFaultError::W.bits() as usize) != 0)
+            .is_err()
```
luigix25: shouldn't this be `is_ok()`? IIUC `handle_pf` returns an error if it's not able to handle it, right?

kernel/src/mm/guestmem.rs
joergroedel: Done.;

elf/src/program_header.rs
joergroedel: I am about to make updates to the ELF loader code. These updates require Copy on ElfPhdrFlags.

kernel/src/cpu/idt/svsm.rs
joergroedel: Yes, it should. This is the reason CI failed because of this bug the code got into a page-fault loop :)
This is fixed now.

kernel/src/mm/guestmem.rs
stefano-garzarella: I was just about to ask why we were removing it. A line in the commit message would have been appreciated.

stefano-garzarella (approved):

PR #1155, opened 2026-07-03
Cleanup ELF loading and handle non page-aligned section ends corretly

### LINKED ISSUE #483: Fix ELF Parser Alignment Handling, Remove Custom Linker Script, and Improve Handling of .bss Section in Binary Parsing

joergroedel:
This cleans up the ELF loading code for user-mode binaries and adds handling for non page-aligned section-ends. When a writable section does not end at a page boundary, the rest of the page is zeroed out.

With these changes the user-mode binaries do no longer need a custom linker script which enforces the alignment. So this is removed as well.

These changes depend on #1152 and #1153.
Fixes: #483


kernel/src/mm/guestmem.rs
```diff
@@ -269,6 +269,44 @@ unsafe fn copy_bytes(src: *const u8, dst: *mut u8, size: usize) -> Result<(), Sv
...
+/// # Safety
+///
+/// The caller must make sure that writing to `dst` does not harm memory safety.
+unsafe fn clear_bytes(dst: *mut u8, size: usize) -> Result<(), SvsmError> {
+    let mut rcx: u64;
+
+    // SAFETY: The safety requirements of the function need to be met. Will
+    // write only `size` zero bytes at `dst`.
+    unsafe {
+        asm!("   testq  %rcx, %rcx
```
stefano-garzarella: can we use `rep` operation like in `copy_bytes` ?

kernel/src/task/exec.rs
```diff
@@ -10,6 +10,7 @@ use super::TaskPointer;
 use crate::address::{Address, VirtAddr};
 use crate::error::SvsmError;
 use crate::fs::{Directory, open_read};
+use crate::mm::guestmem::zero_user_mem;
```
stefano-garzarella: nit:
```suggestion
use crate::mm::zero_user_mem
```

kernel/src/mm/guestmem.rs
joergroedel: No, the `rep` prefix is not supported by the `mov` instruction, only by `movs` (which has two memory operands and can thus not be used here).

kernel/src/task/exec.rs
joergroedel: Updated, thanks.

kernel/src/mm/guestmem.rs
joergroedel: Turns out there is a way using the `stosb` instruction, which supports the `rep` prefix.

joergroedel: Updated PR using `stosb`, thanks.

stefano-garzarella (approved):

PR #1157, opened 2026-07-04
task: assign a PCID to each task page-table root

tanish111:
Process Context IDs let the CPU tag TLB entries with the active page-table root. Without that, every context switch that reloads CR3 must invalidate the whole TLB, even when the outgoing and incoming tasks are unrelated and their mappings could safely coexist in the cache. Giving each live task root its own PCID (1..4095; 0 reserved for per-CPU transitional tables) is the foundation for cheaper switches later and keeps stale translations from one dead root from matching a reused tag.

Each task root gets a TaskPcid from a global pool of 4096 IDs (PCID 0 stays reserved for per-CPU transitional page tables). The allocator keeps a simple in_use bitmap: allocation scans from 1 upward for the first free slot and marks it taken; no flush happens at allocation time because a freshly handed-out ID either was never used or was fully cleaned when the previous owner died. Freeing is tied to task lifetime via RAII — when the Task is destroyed, TaskPcid’s Drop runs release(), which first broadcasts a PCID-targeted TLB shootdown on every CPU (flush_tlb_pcid_sync, INVPCID on native / INVLPGB+TLBSYNC on SNP) so no stale translations remain tagged with that ID, then clears the in_use bit so the number can be reused. That guarantees the invariant that any PCID loaded into CR3 for a new root has no valid entries under its tag anywhere in the system.

### DIFF
--- a/kernel/src/cpu/tlb.rs
+++ b/kernel/src/cpu/tlb.rs
@@ -6,6 +6,7 @@
 
 use crate::address::{Address, VirtAddr};
 use crate::cpu::control_regs::{read_cr3, read_cr4, write_cr3, write_cr4};
+use crate::cpu::features::{Feature, cpu_has_feat};
 use crate::cpu::ipi::{IpiMessage, IpiTarget, send_multicast_ipi};
 use crate::platform::SVSM_PLATFORM;
 use crate::types::PageSize;
@@ -20,6 +21,13 @@ static FLUSH_SMP: AtomicBool = AtomicBool::new(false);
 /// need to be flushed exceeds this value, a complete TLB flush will
 /// be performed instead.
 const TLB_FLUSH_ALL_THRESHOLD: usize = 256;
+
+/// INVPCID descriptor (Intel SDM Vol. 2A, `INVPCID`).
+#[repr(C, align(16))]
+struct InvpcidDesc {
+    pcid: u64,
+    reserved: u64,
+}
 
 /// Defines the scope of a TLB flush.
 #[derive(Copy, Clone, Debug)]
@@ -40,6 +48,9 @@ pub struct TlbFlushScope {
     pub global: bool,
     /// Indicates the range of virtual addresses that should be flushed.
     pub range: TlbFlushRange,
+    /// When set, flush all non-global entries for this PCID (INVPCID or
+    /// INVLPGB, depending on CPU features).
+    pub pcid: Option<u16>,
 }
 
 impl TlbFlushScope {
@@ -48,6 +59,16 @@ impl TlbFlushScope {
         Self {
             global: true,
             range: TlbFlushRange::All,
+            pcid: None,
+        }
+    }
+
+    /// Creates a TLB flush scope targeting all non-global entries tagged with the given PCID.
+    pub const fn pcid(pcid: u16) -> Self {
+        Self {
+            global: false,
+            range: TlbFlushRange::All,
+            pcid: Some(pcid),
         }
     }
 
@@ -63,6 +84,7 @@ impl TlbFlushScope {
         Self {
             global: true,
             range: TlbFlushRange::Range { region, pgsize },
+            pcid: None,
         }
     }
 
@@ -94,6 +116,10 @@ impl TlbFlushScope {
 
     /// Flushes all the entries in the TLB for the current CPU.
     fn flush_percpu_all(&self) {
+        if let Some(pcid) = self.pcid {
+            flush_pcid_percpu(pcid);
+            return;
+        }
         match self.global {
             true => __flush_tlb_global_percpu(),
             false => __flush_tlb_percpu(),
@@ -130,6 +156,11 @@ pub fn set_tlb_flush_smp() {
     FLUSH_SMP.store(true, Ordering::Relaxed);
 }
 
+/// Flushes all non-global TLB entries tagged with `pcid` on all CPUs.
+pub fn flush_tlb_pcid_sync(pcid: u16) {
+    TlbFlushScope::pcid(pcid).flush_all_cpus();
+}
+
 pub fn flush_tlb_global_sync() {
     TlbFlushScope::all().with_global(true).flush_all_cpus();
 }
@@ -164,6 +195,25 @@ pub fn flush_tlb_global_percpu_page(vaddr: VirtAddr, pgsize: PageSize) {
 
 pub fn flush_tlb_percpu() {
     TlbFlushScope::all().with_global(false).flush_percpu();
+}
+
+fn flush_pcid_percpu(pcid: u16) {
+    if cpu_has_feat(Feature::Invpcid) {
+        let desc = InvpcidDesc {
+            pcid: u64::from(pcid),
+            reserved: 0,
+        };
+        // SAFETY: INVPCID type 1 (single-context invalidation) flushes all
+        // non-global TLB entries for the PCID in desc. Type in any GPR,
+        // descriptor in memory. The descriptor must be 16-byte aligned;
+        // alignment is guaranteed by InvpcidDesc's #[repr(C, align(16))] attribute.
+        unsafe {
+            asm!("invpcid ({0}), {1}",
+                 in(reg) &raw const desc,
+                 in(reg) 1u64,
+                 options(att_syntax));
+        }
+    }
 }
 
 fn __flush_tlb_global_percpu() {
--- a/kernel/src/sev/tlb.rs
+++ b/kernel/src/sev/tlb.rs
@@ -113,6 +113,18 @@ pub fn flush_address_sync(va: VirtAddr) {
 }
 
 pub fn flush_tlb_scope(flush_scope: &TlbFlushScope) {
+    if let Some(pcid) = flush_scope.pcid {
+        // PCID-scoped flush via INVLPGB. RDX layout matches Linux (pcid << 16 | asid):
+        // PCID in bits 31:16, ASID in bits 15:0. Set both valid_pcid and valid_asid in
+        // RAX; without valid_asid the hardware mis-scopes the flush and a raw PCID in
+        // RDX would be interpreted as an ASID instead of occupying bits 31:16.
+        let rax = InvlpgbRax::new()
+            .with_valid_asid(true)
+            .with_valid_pcid(true);
+        do_invlpgb(rax.into_bits(), 0, u64::from(pcid) << 16);
+        do_tlbsync();
+        return;
+    }
     match flush_scope.range {
         TlbFlushRange::All => flush_tlb_sync(flush_scope.global),
         TlbFlushRange::Range { region, pgsize } => {
--- a/kernel/src/task/pcid.rs
+++ b/kernel/src/task/pcid.rs
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: MIT
+//
+// Copyright (c) 2026 Tanish Desai
+//
+// Author: Tanish Desai
+
+use crate::cpu::features::{Feature, cpu_has_feat};
+use crate::cpu::tlb::flush_tlb_pcid_sync;
+use crate::error::SvsmError;
+use crate::flat_bitmap_allocator;
+use crate::mm::alloc::AllocError;
+
+const PCID_COUNT: usize = 4096;
+
+/// Bitmap of allocated PCIDs. PCID 0 is reserved as the overflow PCID and is
+/// claimed on first allocation (see [`alloc_pcid`]).
+static PCID_ALLOC: flat_bitmap_allocator!(PCID_COUNT) = <flat_bitmap_allocator!(PCID_COUNT)>::new();
+
+/// True when PCIDs can be used for task page-table roots.
+///
+/// Requires `Pcid` (CR4.PCIDE) plus a targeted flush path: INVPCID (`Invpcid`)
+/// or AMD INVLPGB (`Invlpgb`).
+pub fn pcid_supported() -> bool {
+    cpu_has_feat(Feature::Pcid)
+        && (cpu_has_feat(Feature::Invpcid) || cpu_has_feat(Feature::Invlpgb))
+}
+
+fn alloc_pcid() -> Result<u16, SvsmError> {
+    loop {
+        let idx = PCID_ALLOC
+            .alloc()
+            .ok_or(SvsmError::Alloc(AllocError::OutOfMemory))?;
+        // PCID 0 is reserved; leave it allocated and take the next free id.
+        if idx != 0 {
+            return Ok(idx as u16);
+        }
+    }
+}
+
+fn release_pcid(pcid: u16) {
+    flush_tlb_pcid_sync(pcid);
+    PCID_ALLOC.free(usize::from(pcid));
+}
+
+/// Owned PCID for a task page-table root.
+#[derive(Debug)]
+pub struct TaskPcid {
+    pcid: u16,
+}
+
+impl TaskPcid {
+    pub fn new() -> Result<Self, SvsmError> {
+        Ok(Self {
+            pcid: alloc_pcid()?,
+        })
+    }
+
+    pub fn pcid(&self) -> u16 {
+        self.pcid
+    }
+}
+
+impl Drop for TaskPcid {
+    fn drop(&mut self) {
+        release_pcid(self.pcid);
+    }
+}
--- a/kernel/src/task/schedule.rs
+++ b/kernel/src/task/schedule.rs
@@ -538,7 +538,21 @@ unsafe fn switch_to(prev_task: Option<TaskPointer>, next_task: TaskPointer) -> O
     // the page table and stack information in those tasks are correct and
     // can be used to switch to the correct page table and execution stack.
     unsafe {
-        let cr3 = (*next).page_table.lock().cr3_value().bits();
+        let cr3 = {
+            let mut cr3 = (*next).page_table.lock().cr3_value().bits();
+            if let Some(pcid) = (*next).pcid() {
+                cr3 |= usize::from(pcid);
+                // Set bit 63 so that writing CR3 does not flush the TLB
+                // entries associated with the target PCID. Without this bit,
+                // every write to CR3 invalidates all non-global entries for
+                // the loaded PCID, defeating the purpose of tagging address
+                // spaces with a PCID across context switches. Targeted
+                // invalidation is performed explicitly via INVPCID / INVLPGB
+                // when a mapping actually changes.
+                cr3 |= 1usize << 63;
+            }
+            cr3
+        };
 
         // Switch to new task
         let new_prev = switch_context(
--- a/kernel/src/cpu/control_regs.rs
+++ b/kernel/src/cpu/control_regs.rs
@@ -37,6 +37,10 @@ pub fn cr4_init() {
 
     cr4.insert(CR4Flags::PGE); // Enable Global Pages
 
+    if cpu_has_feat(Feature::Pcid) {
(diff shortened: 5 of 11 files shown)


tanish111:
I used Linux as a reference for both instruction encodings. For INVPCID, I looked at [`arch/x86/include/asm/invpcid.h`](https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/invpcid.h), which uses a 16-byte descriptor (`{pcid, addr}`) passed as a memory operand with the invalidation type in any GPR — generic across all 4 INVPCID types since it has several call sites. I followed the same descriptor layout and operand convention in `flush_pcid_percpu`, but narrowed it to type 1 (single-context) only, since my PCID allocator only ever needs to flush everything tagged with one PCID, never a single address.
Linux Code:- 
```c
struct { u64 d[2]; } desc = { { pcid, addr } };
asm volatile("invpcid %[desc], %[type]":: [desc] "m" (desc), [type] "r" (type) : "memory");
```

For INVLPGB and TLBSYNC, I referenced [`arch/x86/include/asm/tlb.h`](https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/tlb.h). Linux's `__invlpgb()` always sets `INVLPGB_FLAG_ASID` in `rax` and packs `edx` as `pcid << 16 | asid`. I mirrored both of these in my `flush_tlb_scope` PCID branch.
Linux Code:- 
```c
u64 rax = addr | flags | INVLPGB_FLAG_ASID;
u32 ecx = (stride << 31) | (nr_pages - 1);
u32 edx = (pcid << 16) | asid;
asm volatile(".byte 0x0f, 0x01, 0xfe" :: "a" (rax), "c" (ecx), "d" (edx));
```

tanish111:
@luigix25 can you take first pass before I mark it ready for review

kernel/src/task/pcid.rs
```diff
@@ -0,0 +1,76 @@
...
+/// Requires `Pcid` (CR4.PCIDE) plus a targeted flush path: INVPCID (`Invpcid`)
+/// or AMD INVLPGB (`Invlpgb`).
+pub fn pcid_supported() -> bool {
+    cpu_has_feat(Feature::Pcid)
+        && (cpu_has_feat(Feature::Invpcid) || cpu_has_feat(Feature::Invlpgb))
+}
+
+#[derive(Debug)]
+struct PcidAllocator {
+    in_use: [bool; PCID_COUNT],
```
luigix25: not sure it's worth fixing, but we are using a whole page just for pcid tracking.

kernel/src/task/schedule.rs
```diff
@@ -526,7 +526,13 @@ unsafe fn switch_to(prev_task: Option<TaskPointer>, next_task: TaskPointer) -> O
...
     // can be used to switch to the correct page table and execution stack.
     unsafe {
-        let cr3 = (*next).page_table.lock().cr3_value().bits();
+        let cr3 = {
+            let mut cr3 = (*next).page_table.lock().cr3_value().bits();
+            if let Some(pcid) = (*next).pcid() {
+                cr3 |= usize::from(pcid);
+            }
+            cr3
+        };
```
luigix25, line 555: Claude pointed out this to me, and then I checked in the intel manual (https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html) Section `5.10.4.1`

Without bit 63 set, all writes to CR3 will flush all the entries from the _old_ PCID.

Open question for experts: Should we also set bit 63 to preserve the entries when a context switch happens?

kernel/src/sev/tlb.rs
```diff
@@ -113,6 +113,15 @@ pub fn flush_address_sync(va: VirtAddr) {
...
 pub fn flush_tlb_scope(flush_scope: &TlbFlushScope) {
+    if let Some(pcid) = flush_scope.pcid {
+        let rax = InvlpgbRax::new()
+            .with_valid_asid(true)
+            .with_valid_pcid(true);
+        // EDX: PCID in bits 31:16, ASID in bits 15:0 (Linux: pcid << 16 | asid).
+        do_invlpgb(rax.into_bits(), 0, u64::from(pcid) << 16);
+        do_tlbsync();
+        return;
+    }
```
luigix25, line 127: not an expert, so can you explain what are you doing here?

kernel/src/cpu/tlb.rs
```diff
@@ -130,6 +156,10 @@ pub fn set_tlb_flush_smp() {
     FLUSH_SMP.store(true, Ordering::Relaxed);
 }
 
+pub fn flush_tlb_pcid_sync(pcid: u16) {
```
luigix25, line 160: you are adding a new public function, please document it

kernel/src/cpu/tlb.rs
```diff
@@ -94,6 +116,10 @@ impl TlbFlushScope {
 
     /// Flushes all the entries in the TLB for the current CPU.
     fn flush_percpu_all(&self) {
+        if let Some(pcid) = self.pcid {
+            flush_pcid_percpu(pcid);
+            return;
+        }
```
luigix25, line 122: why this change?

kernel/src/cpu/tlb.rs
```diff
@@ -48,6 +59,16 @@ impl TlbFlushScope {
         Self {
             global: true,
             range: TlbFlushRange::All,
+            pcid: None,
+        }
+    }
+
+    /// Flush all non-global TLB entries tagged with the given PCID.
```
luigix25: please be more precise in the doc, you are not flushing anything here.

what about 
`/// Creates a TLB flush scope targeting all entries tagged with the given PCID.`
?

kernel/src/cpu/tlb.rs
```diff
@@ -48,6 +59,16 @@ impl TlbFlushScope {
...
+    pub const fn pcid(pcid: u16) -> Self {
+        Self {
+            global: true,
```
luigix25: why `global:true`? Not an expert but in the doc you wrote "non-global", so I'm confused.

kernel/src/cpu/tlb.rs
```diff
@@ -166,6 +196,24 @@ pub fn flush_tlb_percpu() {
     TlbFlushScope::all().with_global(false).flush_percpu();
 }
 
+fn flush_pcid_percpu(pcid: u16) {
+    if cpu_has_feat(Feature::Invpcid) {
+        let desc = InvpcidDesc {
+            pcid: u64::from(pcid),
+            reserved: 0,
+        };
+        // SAFETY: INVPCID type 1 — type in any GPR, descriptor in memory.
```
luigix25: should we worry about alignments?

kernel/src/cpu/tlb.rs
```diff
@@ -166,6 +196,24 @@ pub fn flush_tlb_percpu() {
...
+        unsafe {
+            asm!("invpcid ({0}), {1}",
+                 in(reg) &raw const desc,
+                 in(reg) 1u64,
+                 options(att_syntax));
+        }
+    } else {
+        __flush_tlb_global_percpu();
+    }
```
luigix25: do we need this else branch? who is going to use this?

kernel/src/cpu/tlb.rs
```diff
@@ -166,6 +196,24 @@ pub fn flush_tlb_percpu() {
...
+    if cpu_has_feat(Feature::Invpcid) {
+        let desc = InvpcidDesc {
+            pcid: u64::from(pcid),
+            reserved: 0,
+        };
+        // SAFETY: INVPCID type 1 — type in any GPR, descriptor in memory.
+        unsafe {
+            asm!("invpcid ({0}), {1}",
+                 in(reg) &raw const desc,
+                 in(reg) 1u64,
```
luigix25, line 213: I think it's worth mentioning what type 1 is and does

luigix25 (changes requested):
In the commit message you mention GSoC, this is not necessary.

You are doing changes that are not trivial at all, *PLEASE* explain them.

kernel/src/sev/tlb.rs
tanish111: @luigix25 I have referred the linux implementation as shared in previous comment as well https://github.com/coconut-svsm/svsm/pull/1157#issuecomment-4881573711.
Linux's `rax = addr | flags | INVLPGB_FLAG_ASID` becomes `InvlpgbRax::new().with_valid_asid(true).with_valid_pcid(true) `. We always set the ASID-valid bit because Linux had to fix the same issue in [634ab761](https://github.com/torvalds/linux/commit/634ab761) without INVLPGB_FLAG_ASID the hardware doesn't scope the invalidation correctly, and passing raw pcid in EDX would land it in the ASID field instead of bits 31:16.

kernel/src/cpu/tlb.rs
tanish111: @luigix25 it's safe to remove this.

kernel/src/cpu/tlb.rs
tanish111: @luigix25 INVPCID reads a 16-byte descriptor from memory and that address must be 16-byte aligned .  We achieve this using `#[repr(C, align(16))]`.

kernel/src/task/pcid.rs
tanish111: Need to discuss this further. Since we're only using a very small subset of PCIDs (around 8–10), we should also look for a way to avoid reserving all 4096 slots upfront. Perhaps we could use a data structure that allocates entries lazily, only when they're needed.

kernel/src/cpu/tlb.rs
luigix25: I think this should be mentioned in the safety comment. "alignment is guaranteed by.."

kernel/src/cpu/tlb.rs
tanish111: @luigix25 this is new branch which will handle TLB flush requests with pcid. If TlbFlushScope has pcid set in it I would run `flush_pcid_percpu` and not global `__flush_tlb_global_percpu`.  I wanted to keep the main caller unchanged `(TaskPcid::drop → flush_tlb_pcid_sync → flush_percpu_all)`.

kernel/src/cpu/tlb.rs
tanish111: @luigix25 I think it makes sense to set `global: false` although it's only for semantics, since that flag isn't used on the code path when a PCID is set. `global: true` would mean flushing global pages also, and PCID-based flushing doesn't support that today.

kernel/src/sev/tlb.rs
luigix25: This should be documented. It's not clear from code

kernel/src/task/pcid.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl PcidAllocator {
+    const fn new() -> Self {
+        Self {
+            in_use: [false; PCID_COUNT],
+        }
+    }
+
+    fn alloc(&mut self) -> Result<u16, SvsmError> {
+        for pcid in 1..PCID_COUNT as u16 {
```
msft-jlange: We can do better than a linear search across an array of `bool`.  A `BitmapAllocatorTree<BitmapAllocator64>` would be much faster, not to mention significantly more compact.

kernel/src/task/pcid.rs
```diff
@@ -0,0 +1,76 @@
...
+            let idx = usize::from(pcid);
+            if !self.in_use[idx] {
+                self.in_use[idx] = true;
+                return Ok(pcid);
+            }
+        }
+        Err(SvsmError::Alloc(AllocError::OutOfMemory))
```
msft-jlange: So does this mean that we are limiting the kernel to a maximum number of processes that cannot exceed the number of PCIDs?  That seems overly restrictive.  We need a better story for the situation in which we try to create a new process and all PCIDs are in use.

kernel/src/task/pcid.rs
tanish111: Yes it's a good idea I would update it with `BitmapAllocatorTree<BitmapAllocator64>`

kernel/src/task/pcid.rs
tanish111: @msft-jlange yes we are moving ahead with this limitation considering svsm has a max of about 4-8 tasks it wouldn't hurt current use cases in any way but yes I agree it's not a good story.

kernel/src/task/pcid.rs
tanish111: One thing to note is that `BitmapAllocatorTree<BitmapAllocator64>` can allocate up to 1024 PCIDs, which is well beyond our current needs but still below the architectural maximum of 4096. Using `BitmapAllocatorTree<BitmapAllocatorTree<BitmapAllocator64>>` increases the capacity to 16,384 entries, which is far more than the 4096 PCIDs we actually need. Any suggestion on what should we go ahead with?

kernel/src/task/pcid.rs
msft-jlange: Why can we not say that if a task cannot allocate a PCID, then it can simply run with PCID disabled?  We are already using an `Option` to hold the PCID in a task, so surely we can support a mix of `Some` and `None` on the same system.

tanish111: To support that, we’d need to toggle CR4.PCIDE at runtime every-time we fallback. Right now, CR4 is configured only once during initialization. (without resetting cr4 if we don't set `self.pcid` during context switch it falls back to 0 and cpu cannot distinguish their TLB entries by PCID leading to accessing corrupt memory) 

I discussed a few alternatives with @joergroedel and @luigix25 during my mentorship meeting. One option is to associate the PCID with TaskMM rather than individual Task, so all threads belonging to the same task share a single PCID.

This also aligns well with the current SVSM design, which supports at most 1024 TaskMM instances—well below the architectural limit of 4096 PCIDs—making the allocation story much simpler.

msft-jlange: > To support that, we’d need to toggle CR4.PCIDE at runtime every-time we fallback.

That's not required.  If you reserve PCID zero to mean "overflow PCID", then you can simply ensure that you set CR3[63]=0 when you load CR3 for a task that does not have a unique PCID, and all of the non-global entries will be flushed.  The result is that every task that has PCID=0 would get a full TLB flush on every CR3 load, exactly as if there were no separate PCIDs.  The fact is that the PCID architecture is designed so that you can set CR4.PCIDE=1 and get no impact whatsoever in TLB management until you start opting in by setting CR3[PCID] to non-zero and by setting CR3[63]=1.  So it's trivial to let PCID and non-PCID address spaces coexist.

> This also aligns well with the current SVSM design, which supports at most 1024 TaskMM instances—well below the architectural limit of 4096 PCIDs—making the allocation story much simpler.

I don't know how long the limit of 1024 instances will persist.  If we are able to expand that limit to 8192 or higher, then we would end up being stuck because of the PCID assignment limitation.  In general, the code is most robust when we don't design in any implicit relationships between resource limits in one area and resource limits in another area.

tanish111: > That's not required.  If you reserve PCID zero to mean "overflow PCID", then you can simply ensure that you set CR3[63]=0 when you load CR3 for a task that does not have a unique PCID, and all of the non-global entries will be flushed.  The result is that every task that has PCID=0 would get a full TLB flush on every CR3 load, exactly as if there were no separate PCIDs.  The fact is that the PCID architecture is designed so that you can set CR4.PCIDE=1 and get no impact whatsoever in TLB management until you start opting in by setting CR3[PCID] to non-zero and by setting CR3[63]=1.  So it's trivial to let PCID and non-PCID address spaces coexist.

Yes thanks for clarification this makes sense. Making pcid default to 0 and remove it from panic if we exhaust all pcid is sufficient to allow unlimited tasks (first 4096 using pcid while others fall back to global flush)  
> 
> 
> 
> I don't know how long the limit of 1024 instances will persist.  If we are able to expand that limit to 8192 or higher, then we would end up being stuck because of the PCID assignment limitation.  In general, the code is most robust when we don't design in any implicit relationships between resource limits in one area and resource limits in another area.
 

 Thanks, this is a good point to note. I’ll apply this principle in future PRs as well.

_

tanish111:
v2:
- moved PCID ownership from Task to TaskMM so tasks sharing an address space share a PCID
- replaced the [bool; 4096] PCID array with a lazily-initialized bitmap allocator
- set CR3[63] on PCID-tagged roots to avoid flushing that PCID's TLB entries on switch
- fall back to PCID 0 instead of failing task creation when PCIDs are exhausted

kernel/src/task/pcid.rs
```diff
@@ -0,0 +1,83 @@
...
+pub fn pcid_supported() -> bool {
+    cpu_has_feat(Feature::Pcid)
+        && (cpu_has_feat(Feature::Invpcid) || cpu_has_feat(Feature::Invlpgb))
+}
+
+/// Bitmap tracking used PCIDs. Capacity is 64 * 16 * 16 = 16384, which covers
+/// the whole `PCID_COUNT` range. Wrapped in an `Option` so the allocator has a
+/// `const` initializer (`None`); the bitmap is created on first use via its
+/// derived `Default`.
+type PcidBitmap = BitmapAllocatorTree<BitmapAllocatorTree<BitmapAllocator64>>;
```
luigix25: this is an hack, it works, but long term we should make the bitmap allocator more flexible.

tanish111:
v3:

- Resolved merge in kernel/src/cpu/features.rs onto the cpufeature leaf descriptors.
- Refactored PCID support features (Invlpgb, Pcid, Invpcid) to use the cpufeature crate (INVLPGB, X86_FEATURE_PCID, X86_FEATURE_INVPCID

kernel/src/task/pcid.rs
tanish111: [Restating idea discussed in meeting for future reference]:- Would go ahead with a new flat_bitmap_allocator which is consumed only by PcidBitmap.

tanish111:
v4:
- Added FlatBitmapAllocator: atomic single-bit flat bitmap (alloc/free via CAS).
- Switched PCID allocation to it (dropped SpinLock + BitmapAllocatorTree).

tanish111:
@msft-jlange I’ve updated the data structure and incorporated all the suggestions mentioned earlier. When you have a chance, could you please take a look and review it?

tanish111:
v5:

- Added a fuzzer for the PCID bitmap allocator
- Fixed stale TLB entries after a task moves CPUs and unmaps a page: `VMR::remove()` now calls `flush_tlb_pcid_sync_range()` instead of `INVLPG`
- Unmaps flush the task’s PCID on every CPU
- Made `pcid_supported()` stricter: PCIDs are used only when the CPU has `INVPCID`
- Dropped unused is_used() from flat_bitmap_allocator.rs

tanish111:
v6:
- merge remote branch to latest remote main

luigix25:
@tanish111 current cpu model in launch_script (EPYC-v4) does not expose pcid nor invpcid.
One possible workaround is this: `CPU=EPYC-v4,+pcid,+invpcid`, otherwise we need to switch to a newer cpu model, like `EPYC-Milan`, `EPYC-Genoa` or `EPYC-Turin`

tanish111:
@luigix25 I think `CPU=EPYC-v4,+pcid,+invpcid` this would be safest and fastest considering other tests are written with assumption we are using EPYC-v4. Switch to a newer cpu model looks like a larger change to me. 
Open for discussion with other maintainers, if it aligns with current plans of SVSM happy to update to a newer cpu model as well. 
cc: @joergroedel  @00xc @stefano-garzarella

PR #1160, opened 2026-07-07
xbuild/test: Replace dashes in package name

n-ramacciotti:
When specifying `--message-format=json-render-diagnostics`, Cargo produces an output where potential '-' in target.name are converted in '_' for lib crates. [1]

This seems to be the expected behavior. [2]

Update `find_test_executable` to account for that.

[1] https://github.com/rust-lang/cargo/issues/13867
[2] https://github.com/rust-lang/cargo/pull/13887

**note**: I discovered this when trying to use `kapi-tester` as crate name instead of `kapi_tester` in #926. 
For `test-in-svsm`, is it correct to assume that the `target.crate_types` will only contain `lib` as bin is tagged with `test=false`? Otherwise, the match guard should be changed to account for other cases

### DIFF
--- a/xbuild/src/cargo.rs
+++ b/xbuild/src/cargo.rs
@@ -20,9 +20,28 @@ enum CargoMessage {
     Other,
 }
 
+/// The kind of a cargo target as reported in `compiler-artifact` messages.
+#[derive(Deserialize, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+enum ArtifactKind {
+    Bin,
+    Lib,
+    Rlib,
+    Dylib,
+    Cdylib,
+    Staticlib,
+    ProcMacro,
+    Test,
+    Bench,
+    Example,
+    #[serde(other)]
+    Other,
+}
+
 #[derive(Deserialize)]
 struct ArtifactTarget {
     name: String,
+    kind: Vec<ArtifactKind>,
 }
 
 #[derive(Deserialize)]
@@ -34,9 +53,8 @@ struct ArtifactProfile {
 /// the path to the test executable produced for package `pkg`.
 ///
 /// Cargo emits one JSON object per line. Look for the `compiler-artifact`
-/// record whose target name matches `pkg`, was built in test mode, and has
-/// an `executable` path. If cargo produces several such records (e.g. lib +
-/// integration tests) the last one wins.
+/// record whose target name matches `pkg`, is a lib target, was built in test
+/// mode, and has an `executable` path.
 pub fn find_test_executable(stdout: &[u8], pkg: &str) -> Option<PathBuf> {
     std::str::from_utf8(stdout)
         .ok()?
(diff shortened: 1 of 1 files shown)


00xc:
> **note**: I discovered this when trying to use `kapi-tester` as crate name instead of `kapi_tester` in #926.
> For `test-in-svsm`, is it correct to assume that the `target.crate_types` will only contain `lib` as bin is tagged with `test=false`? Otherwise, the match guard should be changed to account for other cases

Right, yes, because as you mention, we have `test = false`:

https://github.com/coconut-svsm/svsm/blob/27cb6d407c55732058488493ffb6112c517dccae/kernel/Cargo.toml#L6-L9

I guess we could make it more robust to only match on `lib` targets, but it is not strictly necessary. 

```diff
diff --git a/xbuild/src/cargo.rs b/xbuild/src/cargo.rs
index 7feff197988d..702a9bffd0e7 100644
--- a/xbuild/src/cargo.rs
+++ b/xbuild/src/cargo.rs
@@ -20,9 +20,28 @@ enum CargoMessage {
     Other,
 }
 
+/// The kind of a cargo target as reported in `compiler-artifact` messages.
+#[derive(Deserialize, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+enum ArtifactKind {
+    Bin,
+    Lib,
+    Rlib,
+    Dylib,
+    Cdylib,
+    Staticlib,
+    ProcMacro,
+    Test,
+    Bench,
+    Example,
+    #[serde(other)]
+    Other,
+}
+
 #[derive(Deserialize)]
 struct ArtifactTarget {
     name: String,
+    kind: Vec<ArtifactKind>,
 }
 
 #[derive(Deserialize)]
@@ -34,9 +53,8 @@ struct ArtifactProfile {
 /// the path to the test executable produced for package `pkg`.
 ///
 /// Cargo emits one JSON object per line. Look for the `compiler-artifact`
-/// record whose target name matches `pkg`, was built in test mode, and has
-/// an `executable` path. If cargo produces several such records (e.g. lib +
-/// integration tests) the last one wins.
+/// record whose target name matches `pkg`, is a lib target, was built in test
+/// mode, and has an `executable` path.
 pub fn find_test_executable(stdout: &[u8], pkg: &str) -> Option<PathBuf> {
     std::str::from_utf8(stdout)
         .ok()?
@@ -47,7 +65,12 @@ pub fn find_test_executable(stdout: &[u8], pkg: &str) -> Option<PathBuf> {
                 target,
                 profile,
                 executable,
-            } if target.name == pkg && profile.test => executable,
+            } if target.name == pkg.replace('-', "_")
+                && target.kind.contains(&ArtifactKind::Lib)
+                && profile.test =>
+            {
+                executable
+            }
             _ => None,
         })
         .next_back()
```

00xc (dismissed):
LGTM. I'll wait for your reply in case you want to add the `lib` matching on top. Otherwise we can just merge this directly.

n-ramacciotti:
v2:
- applied @00xc's suggestion:
  - introduced `ArtificactKind` to check for `lib` presence

00xc (approved):

joergroedel (approved):

PR #1161, opened 2026-07-07
kernel/platform: allow unused_unsafe on __cpuid_count call

luigix25:
`__cpuid_count` is safe on nightly but unsafe on stable. This causes an unused_unsafe warning when running test-in-svsm that uses nightly.

Add #[allow(unused_unsafe)] to satisfy both toolchains.

### DIFF
--- a/kernel/src/platform/mod.rs
+++ b/kernel/src/platform/mod.rs
@@ -186,8 +186,13 @@ pub trait SvsmPlatform: Sync {
     where
         Self: Sized,
     {
+        // Stable clippy and the nightly rust compiler disagree on the safety
+        // of __cpuid_count(). Silence both until this is resolved.
+        #[allow(unused_unsafe)]
         // SAFETY: CPUID is always safe
-        unsafe { Some(__cpuid_count(eax, ecx)) }
+        unsafe {
+            Some(__cpuid_count(eax, ecx))
+        }
     }
 
     /// Write a host-owned MSR.


kernel/src/platform/mod.rs
```diff
@@ -187,6 +187,7 @@ pub trait SvsmPlatform: Sync {
         Self: Sized,
     {
         // SAFETY: CPUID is always safe
+        #[allow(unused_unsafe)]
```
stefano-garzarella, line 191: Can you add a comment here with the motivations you mentioned in the commit description?

luigix25: done

luigix25:
- Fixed cargo fmt problems
- Added comment [@stefano-garzarella]

kernel/src/platform/mod.rs
```diff
@@ -187,7 +187,12 @@ pub trait SvsmPlatform: Sync {
         Self: Sized,
     {
         // SAFETY: CPUID is always safe
-        unsafe { Some(__cpuid_count(eax, ecx)) }
+        // __cpuid_count is unsafe on stable but safe on nightly,
+        // causing an unused_unsafe warning on test-in-svsm.
+        #[allow(unused_unsafe)]
+        unsafe {
+            Some(__cpuid_count(eax, ecx))
+        }
```
stefano-garzarella: Sorry, I just noticed we do the same in `kernel/src/cpu/vc.rs` and I like what we did there.
So what about following it both having the safety comment after the `#[allow(unused_unsafe)]` plus the message we used there which is a bit more clear ?

luigix25:
- Now reusing the same comment from `kernel/src/cpu/vc.rs` [@stefano-garzarella]

stefano-garzarella (approved):

PR #1162, opened 2026-07-07
clippy: allow `new_without_default`

msft-jlange:
Many types implement `T::new()` with no parameters, and there is no expectation that `T::default()` is also available.  However, by default, clippy requires every type that implements `T::new()` to also implement `T::default()`, and this results in unnecessary errors and unnecessary code introduced to correct those errors.  This clippy warning can be suppressed without negative effects, and the unneeded `Default` implementatiosn can be removed.

Note that when a type `U` implements `Default` and `U` contains an instance of `T`, then `T` is required to implement `Default`, but the compiler will generate an error in that case, and there is no need to rely on clippy to find problems like this.


Cargo.lock
```diff
@@ -1443,9 +1443,9 @@ checksum = "f52b00d39961fc5b2736ea853c9cc86238e165017a493d1d5c8eac6bdc4cc273"
...
 [[package]]
 name = "memmap2"
-version = "0.9.11"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d1219ed1b7f229ee7104d281dd01d6802fe28bb6e95d292942c4daacdeb798c0"
+checksum = "f49388d20533534cd19360ad3d6a7dadc885944aa802ba3995040c5ec11288c6"
 dependencies = [
  "libc",
 ]
```
luigix25, line 1451: i suspect something went wrong during a rebase

packit
luigix25, line 1: ditto

luigix25 (changes requested):

00xc (approved):

stefano-garzarella:
We have at least another one in `elf/`, plus 2 in `packit`, but it's a submodule, so if we want we should do in that repo:
```
$ grep -rn -A3 'impl Default for' --include='*.rs' | grep 'Self::new()'
elf/src/relocation.rs-168-        Self::new()
packit/src/archive.rs-93-        Self::new()
packit/src/header.rs-58-        Self::new()
```

So I think adding this is to this PR is enough, or we can do as followup:
```diff
--- a/elf/src/relocation.rs
+++ b/elf/src/relocation.rs
@@ -163,12 +163,6 @@ impl Elf64X86RelocProcessor {
     }
 }
 
-impl Default for Elf64X86RelocProcessor {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl Elf64RelocProcessor for Elf64X86RelocProcessor {
     /// Applies a relocation operation for x86_64 ELF files.
     ///
```

joergroedel (approved):

joergroedel:
> We have at least another one in `elf/`, plus 2 in `packit`, but it's a submodule, so if we want we should do in that repo:
> 
> ```
> $ grep -rn -A3 'impl Default for' --include='*.rs' | grep 'Self::new()'
> elf/src/relocation.rs-168-        Self::new()
> packit/src/archive.rs-93-        Self::new()
> packit/src/header.rs-58-        Self::new()
> ```
> 
> So I think adding this is to this PR is enough, or we can do as followup:
> 
> ```diff
> --- a/elf/src/relocation.rs
> +++ b/elf/src/relocation.rs
> @@ -163,12 +163,6 @@ impl Elf64X86RelocProcessor {
>      }
>  }
>  
> -impl Default for Elf64X86RelocProcessor {
> -    fn default() -> Self {
> -        Self::new()
> -    }
> -}
> -
>  impl Elf64RelocProcessor for Elf64X86RelocProcessor {
>      /// Applies a relocation operation for x86_64 ELF files.
>      ///
> ```

Sorry, missed that. We should remove it in a follow-on PR.

stefano-garzarella:
> Sorry, missed that. We should remove it in a follow-on PR.

Don't worry, posted here: https://github.com/coconut-svsm/svsm/pull/1163

PR #1164, opened 2026-07-09
boot_params: get debug com port from IGVM command line

msft-jlange:
The IGVM command line is intended to convey debug configuration options, so obtain the debug com port identifier from the command line instead of relying on a static configuration in the IGVM file.  This eliminates one more hypervisor-specific difference from the IGVM file.  If the command line is not specified (QEMU does not specify a command line), then the com port defaults to 1, which is the value that QEMU expects, thus ensuring that QEMU can get the correct behavior without requiring a command line to be specified.

### DIFF
--- a/boot/bootimg/src/lib.rs
+++ b/boot/bootimg/src/lib.rs
@@ -169,11 +169,11 @@ where
         idt_vaddr,
         boot_params_virt_addr: boot_params_vaddr,
         vtom: boot_image_params.vtom,
-        debug_serial_port: boot_image_params.boot_params.debug_serial_port,
         ap_start_context_addr: boot_image_params.ap_start_context_addr,
         use_alternate_injection: boot_image_params.boot_params.use_alternate_injection != 0,
         kernel_page_table_vaddr: kernel_page_tables.root_vaddr(),
         vmsa_in_kernel_heap,
+        _reserved: Default::default(),
     };
     add_page_contents(add_page_data, launch_info_paddr, launch_info.as_bytes())?;
 
--- a/boot/defs/src/boot_params.rs
+++ b/boot/defs/src/boot_params.rs
@@ -118,12 +118,13 @@ pub struct BootParamBlock {
     /// of the memory map (which is in IGVM format).
     pub memory_map_offset: u32,
 
+    /// The offset, in bytes, of the command line, or zero if no command line
+    /// is present.
+    pub command_line_offset: u32,
+
     /// The offset, in bytes, of the guest context, or zero if no guest
     /// context is present.
     pub guest_context_offset: u32,
-
-    /// The port number of the serial port to use for debugging.
-    pub debug_serial_port: u16,
 
     /// Indicates whether the guest can support alternate injection.
     pub use_alternate_injection: u8,
@@ -141,7 +142,7 @@ pub struct BootParamBlock {
     /// range.
     pub vmsa_in_kernel_range: u8,
 
-    pub _reserved: u8,
+    pub _reserved: [u8; 7],
 
     /// Metadata containing information about the firmware image embedded in the
     /// IGVM file.
--- a/boot/defs/src/boot_params.rs
+++ b/boot/defs/src/boot_params.rs
@@ -125,12 +125,13 @@ pub struct BootParamBlock {
     /// of the memory map (which is in IGVM format).
     pub memory_map_offset: u32,
 
+    /// The offset, in bytes, of the command line, or zero if no command line
+    /// is present.
+    pub command_line_offset: u32,
+
     /// The offset, in bytes, of the guest context, or zero if no guest
     /// context is present.
     pub guest_context_offset: u32,
-
-    /// The port number of the serial port to use for debugging.
-    pub debug_serial_port: u16,
 
     /// Indicates whether the guest can support alternate injection.
     pub use_alternate_injection: u8,
@@ -144,8 +145,6 @@ pub struct BootParamBlock {
     /// Indicates whether the VMSA is placed at the top of the kernel GPA
     /// range.
     pub vmsa_in_kernel_range: u8,
-
-    pub _reserved: [u8; 2],
 
     /// Metadata containing information about the firmware image embedded in the
     /// IGVM file.
--- a/boot/defs/src/kernel_launch.rs
+++ b/boot/defs/src/kernel_launch.rs
@@ -36,9 +36,9 @@ pub struct KernelLaunchInfo {
     pub vtom: u64,
     pub kernel_page_table_vaddr: u64,
     pub ap_start_context_addr: u32,
-    pub debug_serial_port: u16,
     pub vmsa_in_kernel_heap: bool,
     pub use_alternate_injection: bool,
+    pub _reserved: [bool; 2],
 }
 
 pub const INITIAL_KERNEL_STACK_WORDS: usize = 3;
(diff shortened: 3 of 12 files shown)


msft-jlange:
This PR requires a change to QEMU such that QEMU ignores the command line parameter directive instead of treating it as a fatal error.  Since the kernel doesn't require any command line to be present, this is a safe change to QEMU, but it has to propagate before this PR can merge.

stefano-garzarella:
> This PR requires a change to QEMU such that QEMU ignores the command line parameter directive instead of treating it as a fatal error. Since the kernel doesn't require any command line to be present, this is a safe change to QEMU, but it has to propagate before this PR can merge.

yeah, we were discussing exactly about this on a thread related to device tree support. I don't remember the current status, but @luigix25 was working on that, so maybe it's the time to merge it at least downstream.

stefano-garzarella:
Could also be related to this: https://github.com/microsoft/igvm/issues/120

luigix25:
> > This PR requires a change to QEMU such that QEMU ignores the command line parameter directive instead of treating it as a fatal error. Since the kernel doesn't require any command line to be present, this is a safe change to QEMU, but it has to propagate before this PR can merge.
> 
> yeah, we were discussing exactly about this on a thread related to device tree support. I don't remember the current status, but @luigix25 was working on that, so maybe it's the time to merge it at least downstream.

Upstream QEMU already supports optional igvm directives!
https://gitlab.com/qemu-project/qemu/-/commit/3f24c8a74ce89686143be2517b4c1e98651bb913

as per the igvm spec, you need to set the 31st bit in the directive (https://docs.rs/igvm_defs/0.4.0/igvm_defs/struct.IgvmVariableHeaderType.html)

As Stefano pointed out, I think some changes are required in igvm, as currently there is no way to set the optional bit using that crate.

We could start by backporting that commit in our fork, but i don't think it's super urgent as we need to wait for a new igvm release, that includes that option, first.

stefano-garzarella:
> 
> Upstream QEMU already supports optional igvm directives! https://gitlab.com/qemu-project/qemu/-/commit/3f24c8a74ce89686143be2517b4c1e98651bb913
> 
> as per the igvm spec, you need to set the 31st bit in the directive (https://docs.rs/igvm_defs/0.4.0/igvm_defs/struct.IgvmVariableHeaderType.html)
> 
> As Stefano pointed out, I think some changes are required in igvm, as currently there is no way to set the optional bit using that crate.
> 
> We could start by backporting that commit in our fork, but i don't think it's super urgent as we need to wait for a new igvm release, that includes that option, first.

So IIUC IGVM spec supports "optional" directives, QEMU supports it, but here, in `igvmbuilder` we don't have any way to specify that a directive is optional because `igvm` crate doesn't support it (tracked by https://github.com/microsoft/igvm/issues/120). @luigix25 is this correct?

I think in the meantime we should change QEMU to continue the boot also if a non-optional directive is not recognized by QEMU. This will avoid breaking changes and will work for now. IMO QEMU should do a best effort, will be up to SVSM to fail the boot if something is missing. @kraxel WDTY?

luigix25:
> > Upstream QEMU already supports optional igvm directives! https://gitlab.com/qemu-project/qemu/-/commit/3f24c8a74ce89686143be2517b4c1e98651bb913
> > as per the igvm spec, you need to set the 31st bit in the directive (https://docs.rs/igvm_defs/0.4.0/igvm_defs/struct.IgvmVariableHeaderType.html)
> > As Stefano pointed out, I think some changes are required in igvm, as currently there is no way to set the optional bit using that crate.
> > We could start by backporting that commit in our fork, but i don't think it's super urgent as we need to wait for a new igvm release, that includes that option, first.
> 
> So IIUC IGVM spec supports "optional" directives, QEMU supports it, but here, in `igvmbuilder` we don't have any way to specify that a directive is optional because `igvm` crate doesn't support it (tracked by [microsoft/igvm#120](https://github.com/microsoft/igvm/issues/120)). @luigix25 is this correct?
> 

yep, correct

joergroedel:
> The IGVM command line is intended to convey debug configuration options, so obtain the debug com port identifier from the command line instead of relying on a static configuration in the IGVM file. This eliminates one more hypervisor-specific difference from the IGVM file. If the command line is not specified (QEMU does not specify a command line), then the com port defaults to 1, which is the value that QEMU expects, thus ensuring that QEMU can get the correct behavior without requiring a command line to be specified.

The command-line may contain security-relevant settings, how is it measured? Will it be part of the launch measurement or is there another way?

kraxel:
> I think in the meantime we should change QEMU to continue the boot also if a non-optional directive is not recognized by QEMU. This will avoid breaking changes and will work for now. IMO QEMU should do a best effort, will be up to SVSM to fail the boot if something is missing. @kraxel WDTY?

The optional flag is there to signal to the hypervisor whenever it is save to ignore it if unknown or not.  IMHO qemu should implement this correctly.  This needs fixing on the igvm / svsm side to correctly tag the command line as optional.  Implementing igvm cmd line support in qemu is obviously an option too.

stefano-garzarella:
> > I think in the meantime we should change QEMU to continue the boot also if a non-optional directive is not recognized by QEMU. This will avoid breaking changes and will work for now. IMO QEMU should do a best effort, will be up to SVSM to fail the boot if something is missing. @kraxel WDTY?
> 
> The optional flag is there to signal to the hypervisor whenever it is save to ignore it if unknown or not. IMHO qemu should implement this correctly. This needs fixing on the igvm / svsm side to correctly tag the command line as optional. Implementing igvm cmd line support in qemu is obviously an option too.

I need to check better the spec, but I saw it more like @chris-oo comment here https://github.com/microsoft/igvm/issues/120#issuecomment-4651502604

We don't trust the hypervisor by design, so I don't see any real advantage in QEMU failing to create the CVM, rather than doing its best and providing what it can.

That said, if the specification clearly states that if the loader encounters a non-optional directive it cannot handle, then it must fail, okay, then there's no doubt about what to do.

stefano-garzarella:
> That said, if the specification clearly states that if the loader encounters a non-optional directive it cannot handle, then it must fail, okay, then there's no doubt about what to do.

I found this TODO in the code https://github.com/microsoft/igvm/blob/3d1a950116d624746bf3342e09ea0a3416d89dd3/igvm_defs/src/lib.rs#L219-L223:
```
// TODO: The format_version in the fixed_header should describe the required
//       minimum version of the file. The loader should be free to ignore
//       additional headers it does not understand. The file builder is required
//       to set the format_version to the correct minimum version required to
//       boot. Maybe IGVM_FIXED_HEADER_EXTENSION?
```

But it's a TODO, so it's not entirely clear to me whether the loader is allowed to ignore directives it doesn't recognize or not.

msft-jlange:
> The command-line may contain security-relevant settings, how is it measured? Will it be part of the launch measurement or is there another way?

The command line doesn't contain settings that are relevant for security.  Unmeasured parameters are not allowed to contain security-sensitive information.

What we're using this for is to determine which COM port should be used for console output.  That should not be security-sensitive.  If the host proposes a COM port that the SVSM believes impacts security, then the SVSM should simply reject that COM port option.

stefano-garzarella:
FYI we just backported IGVM_VHT_OPTIONAL_BIT support from QEMU upstream to our downstream. See https://github.com/coconut-svsm/qemu/pull/34 So when `igvm` will support it, we should be ready (thanks @luigix25 for the effort)

PR #1165, opened 2026-07-11
userspace: remove direct console access

n-ramacciotti:
From PR #52 (https://github.com/coconut-svsm/svsm/pull/52#discussion_r3559615916 which probably can't be accessed directly as marked as solved):
`
yes, exactly. I'm not sure this is a good behavior. My idea was to leave only one handler (on of CONSOLE_HANDLE or LOG_HANDLE), removing the ability for userspace to print directly to the console, but only through the console file handler, which leave the final decision to kernel (enable_console_log). Essentially, IMO, userspace should not be able to print directly neither it should know where the message is logged (console or log file).
`

Remove the console handler, as this allows userspace to bypass the kernel's decision to log to the console or to the log buffer only.

~~Reduce log level so that `log::debug` are available in the binary. The minimum log level can still be set at runtime.~~ (I will move this to another PR)

~~Reintroduce `$crate` in macro definitions.~~


n-ramacciotti:
v2:
- removed one commit. It has been moved in https://github.com/coconut-svsm/svsm/pull/1166

@vsntk18 please take a look

n-ramacciotti:
v3:
- Updated `LINE_BUFFER_SIZE` to 224
- Fixed typo in commit message

kernel/src/fs/buffer.rs
```diff
@@ -140,28 +141,26 @@ impl Buffer for UserBuffer {
     }
 }
 
-// With the value of 223 the Buffer struct will be exactly 256 bytes
+// With the value of 224 the Buffer struct will be exactly 256 bytes
 // large, avoiding memory waste due to internal fragmentation.
-const LINE_BUFFER_SIZE: usize = 223;
+const LINE_BUFFER_SIZE: usize = 224;
```
vsntk18, line 146: Can we use something like assert_eq!(size_of::< LineBuffer > (), 256) for sanity check?

n-ramacciotti: Makes sense! I will add some assertions

n-ramacciotti:
v4:
- addressed @vsntk18's comment:
  - add compile time assertion on `LineBuffer` size

vsntk18:
Thanks for cleaning up the mess. Changes generally look good from my point of view.

But we have now unconditionally revoked the direct console access to user space (even when enable-console-log) is on. Are you sure this is fine? I am not aware of all the use cases. So it's better to discuss this with Joerg and Stefano?

An alternative would be to have a feature called enable-console (instead of enable-console-log),  if enabled, write_console() syscall is still available to the user.

n-ramacciotti:
> But we have now unconditionally revoked the direct console access to user space (even when enable-console-log) is on. Are you sure this is fine? I am not aware of all the use cases.

Good question! From #12:
`
The SVSM prints all log messages directly to the serial console. In a production environment this could be a security problem as it reveals information about the guest.
`

With your PR, the information are now stored in the log file and with `enable-console-log` are still printed to console (also the one coming from userspace). I believe the ultimate goal is to avoid printing to the console entirely, but for the moment it is ok to do so (with `enable-console-log`). However, I don't think that userspace should be able to write to console when kernel didn't enable the feature.

Edit: There might be some particular cases where writing to console will still be needed, but I think those are related to kernel-side operations and userspace should not be involved at all.

>  So it's better to discuss this with Joerg and Stefano

I agree. I open this PR so we can discuss with others. @stefano-garzarella

kernel/src/fs/log_buffer.rs
```diff
@@ -69,15 +69,11 @@ impl File for LogFile {
     }
 }
 
-pub fn stdout_open(taskname: String) -> (Arc<dyn Obj>, Arc<dyn Obj>) {
-    let console_file: Arc<dyn File> = Arc::new(LogFile::new(taskname.clone(), true));
-    let log_file: Arc<dyn File> = Arc::new(LogFile::new(taskname.clone(), false));
+pub fn stdout_open(taskname: String) -> Arc<dyn Obj> {
+    let log_file: Arc<dyn File> = Arc::new(LogFile::new(taskname.clone()));
```
luigix25: `clone` can be dropped

kernel/src/fs/buffer.rs
luigix25, line 146: probably i'm missing something, but why are you changing from 223 to 224?

syscall/src/logging.rs
```diff
@@ -6,9 +6,8 @@
 
 use crate::{FsObjHandle, ObjHandle, SysCallError, write};
 
-static LOG_HANDLE: FsObjHandle = FsObjHandle::new(ObjHandle::new(1));
+static LOG_HANDLE: FsObjHandle = FsObjHandle::new(ObjHandle::new(0));
```
luigix25, line 9: i think this should be moved to commit 1, otherwise bisect is broken.

IOW: when running userspace tests in commit 1 this happens:


```
[SVSM test task] Running tests in svsm for uapi_tester
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
[uapi_tester] ERROR: logging failed
```

kernel/src/fs/buffer.rs
```diff
@@ -185,19 +184,15 @@ impl LineBuffer {
...
-        } else {
+        let _ = log_write(self.prefix.as_bytes())?;
+        let _ = log_write(&self.buffer[..self.fill])?;
+
+        #[cfg(feature = "enable-console-log")]
+        {
             console_write(self.prefix.as_bytes());
             console_write(&self.buffer[..self.fill]);
         }
+
```
luigix25: nit: blank line

kernel/src/log_buffer.rs
```diff
@@ -28,7 +28,7 @@ impl LogBuffer {
 impl fmt::Write for LogBuffer {
     /// Writes all bytes from the slice to the log buffer
     fn write_str(&mut self, s: &str) -> fmt::Result {
-        let obj_handle = ObjHandle::new(1);
```
luigix25, line 31: i think this change should be mentioned in the commit message

luigix25:
while we are here: I'm looking at `DISABLE_CONSOLE_LOG` in the Makefile, but I don't think it works, does it?

IIUC `enable-console-log` is enabled by default on every build

kernel/src/fs/buffer.rs
n-ramacciotti: It was already 224 before the introduction of `is_console` field. So when I removed the field, I updated also that value. Strictly speaking, `LineBuffer` would still be 256 as the compiler can introduce some padding.

luigix25: Oh, I see now. Probably it's worth mentioning this in the commit message!

vsntk18:
> while we are here: I'm looking at `DISABLE_CONSOLE_LOG` in the Makefile, but I don't think it works, does it?
> 
> IIUC `enable-console-log` is enabled by default on every build

hmm, Yeah, after adding it to features explicitly in configs/* file,  It became pointless.

n-ramacciotti:
v5:
- addressed @luigix25's comments:
  - removed `DISABLE_CONSOLE_LOG`
  - solved broken git bisection. Now in the first commit both `LOG_HANDLE` and `CONSOLE_HANDLE` have 0 as obj identifier, but the latter is never used, so I decide to keep it and remove it in the subsequent commit.
  - updated commits

luigix25 (approved):

n-ramacciotti:
v6:
- moved @luigix25's suggestion of removing `DISABLE_CONSOLE_LOG` in #1166 
- moved "$crate" introduction in macros in #1166

Since these changes are not related to removing direct access to console they can be merged in the other PR that is now ready for review

n-ramacciotti:
v7:
- rebased on latest main and solved conflicts on `LogFile::new()`, `LienBufer::new()`, and `stdout_open` (where String was substituted)

n-ramacciotti:
v8:
- rebased on main after #1166

joergroedel (approved):

PR #1166, opened 2026-07-13
userspace: Introduce log crate

n-ramacciotti:
Introduce `log` crate in `userlib` and make it available for user binaries. `log::set_logger` behaves internally as `immut_after_init` when initializing, so there is no need to introduce that and successive calls will return an error. Call the `install_logger` function before entering main. This function causes a panic on multiple calls.

Reduce log level so that `log::debug` are available in the binary. The minimum log level can still be set at runtime.

~~Depends on https://github.com/coconut-svsm/svsm/pull/1165: only the last two commits here are related to this PR~~

other fixes:
- introduce `$crate` in macros
- remove `DISABLE_CONSOLE_LOG`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,11 +33,6 @@ TARGET_PATH=debug
 OBJCOPY_ELF_ARGS := --strip-debug
 endif
 
-ifeq ($(DISABLE_CONSOLE_LOG),)
-SVSM_ARGS += --features enable-console-log
-XBUILD_ARGS += -f enable-console-log
-endif
-
 ifdef OFFLINE
 CARGO_ARGS += --locked --offline
 endif
--- a/kernel/src/log_buffer.rs
+++ b/kernel/src/log_buffer.rs
@@ -51,13 +51,7 @@ fn lb_write(args: fmt::Arguments<'_>) {
 }
 
 #[derive(Clone, Copy, Debug)]
-struct BufferLogger {}
-
-impl BufferLogger {
-    const fn new() -> Self {
-        Self {}
-    }
-}
+struct BufferLogger;
 
 impl log::Log for BufferLogger {
     fn enabled(&self, _metadata: &log::Metadata<'_>) -> bool {
@@ -94,24 +88,21 @@ impl log::Log for BufferLogger {
     fn flush(&self) {}
 }
 
-static BUFFER_LOGGER: ImmutAfterInitCell<BufferLogger> = ImmutAfterInitCell::uninit();
+static BUFFER_LOGGER: BufferLogger = BufferLogger;
 
+/// Set the log buffer as logger. This function will panic if it is
+/// called more than once.
 pub fn install_buffer_logger() {
     let logbuf = SpinLock::new(LogBuffer::new());
-    let _res = LOGGER.init(logbuf).map_err(|_| SvsmError::LogError);
+    LOGGER
+        .init(logbuf)
+        .expect("Failed to initialize log buffer");
 
-    BUFFER_LOGGER
-        .init_from_ref(&BufferLogger::new())
-        .expect("log init error");
-
-    if let Err(e) = log::set_logger(&*BUFFER_LOGGER) {
-        // Failed to install the ConsoleLogger, presumably because something had
-        // installed another logger before. No logs will appear at the console.
-        // Print an error string.
-        _print(format_args!(
-            "ERROR: failed to install console logger: {e:?}"
-        ));
-    }
+    // As described in the `log` crate, this function can be called only once
+    // and successive calls will return an error. Panic on that error.
+    // Essentially, it handles initialization in the same way as `immut_after_init`,
+    // so we can avoid using it. Therefore, the logger will be immutable after initialization.
+    log::set_logger(&BUFFER_LOGGER).expect("Failed to install the BufferLogger");
 
     // Log levels are to be configured via the log's library feature configuration.
     log::set_max_level(log::LevelFilter::Trace);
--- a/user/lib/src/lib.rs
+++ b/user/lib/src/lib.rs
@@ -40,6 +40,6 @@ macro_rules! declare_main {
 #[cfg(all(not(test), target_os = "none"))]
 #[panic_handler]
 fn panic(info: &core::panic::PanicInfo<'_>) -> ! {
-    println!("Panic: {}", info);
+    log::error!("Panic: {}", info);
     exit(!0);
 }
--- a/user/lib/src/lib.rs
+++ b/user/lib/src/lib.rs
@@ -15,6 +15,8 @@ pub use locking::*;
 pub use log_buffer::*;
 pub use syscall::*;
 
+pub use log;
+
 #[cfg(all(feature = "test_runner", not(test), target_os = "none"))]
 pub mod testing;
 #[cfg(all(feature = "test_runner", not(test), target_os = "none"))]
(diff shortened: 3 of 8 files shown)


kernel/Cargo.toml
```diff
@@ -43,7 +43,7 @@ gdbstub_arch = { workspace = true, optional = true }
 igvm_defs = { workspace = true, features = ["unstable"] }
 intrusive-collections.workspace = true
 kbs-types = { workspace = true, optional = true, features = ["alloc"] }
-log = { workspace = true, features = ["max_level_info", "release_max_level_info"] }
+log = { workspace = true, features = ["max_level_trace", "release_max_level_info"] }
```
n-ramacciotti: Is it better to move the features in the root `Cargo.toml`? As this is needed also for user crates. So this could avoid duplication and potentially makes things more coherent

n-ramacciotti:
v2:
- removed dependency on #1165:
  - it was only based on the name of a file (`logging.rs` which here is `log_buffer.rs`
- Added some commits from #1165 which are just fixes and not related to disabling direct access to console for userspace

It is now ready for review.

luigix25:
Please update the PR description, IIUC this does not depend on the other PR anymore.

n-ramacciotti:
> Please update the PR description, IIUC this does not depend on the other PR anymore.

right, thanks!

n-ramacciotti:
v3:
- Added a new commit with `$crate::exit` in `declare_main` macro definition  to avoid the import
- use `$crate::install_logger` in `declare_main` to avoid the import
- moved `log` usage in another commit 
- update `install_logger` description

n-ramacciotti:
v4:
- added commit b9b3d255887755d8f217e0010cbf5ede54276b95

Some commits are not directly related to userspace logging. Is it better to split this into multiple PRs?

kernel/Cargo.toml
stefano-garzarella: I'm not sure, I see a value to keep different levels

Makefile
stefano-garzarella, line 1: from commit 047d30752da1ce8077b2e2a6a396353f310af085:
```
makefile: Remove broken DISABLE_CONSOLE_LOG

The new variable `DISABLE_CONSOLE_LOG`, when defined, should prevent the
enabling of `enable_console_log`. However, since that feature is enabled
by default for all the targets, the variable doesn't work. Remove it.

Suggested-by: Luigi Leonardi <leonardi@redhat.com>
Signed-off-by: Nicola Ramacciotti <niko.ramak@gmail.com>
```
In `kernel/Cargo.toml` I can't see `enable-console-log` enabled by default, so I'm a bit confused.
Do you mean that this is enabled by all `configs/*json`?

kernel/Cargo.toml
```diff
@@ -43,7 +43,7 @@ gdbstub_arch = { workspace = true, optional = true }
 igvm_defs = { workspace = true, features = ["unstable"] }
 intrusive-collections.workspace = true
 kbs-types = { workspace = true, optional = true, features = ["alloc"] }
-log = { workspace = true, features = ["max_level_info", "release_max_level_info"] }
```
stefano-garzarella, line 46: Why this change?
Do you mean that for debug build is useful to print also trace messages? I'm not sure about that TBH.

kernel/src/log_buffer.rs
```diff
@@ -94,24 +88,20 @@ impl log::Log for BufferLogger {
...
-            "ERROR: failed to install console logger: {e:?}"
-        ));
-    }
+    LOGGER
+        .init(logbuf)
+        .expect("Failed to initialize log buffer");
+
+    // Failed to install the BufferLogger, presumably because something had
+    // installed another logger before. This behaves internally as `immut_after_init`,
+    // so there is no need to use it here.
```
stefano-garzarella: Should this comment be updated? especially because now there is a panic.

user/lib/src/lib.rs
```diff
@@ -40,6 +40,6 @@ macro_rules! declare_main {
 #[cfg(all(not(test), target_os = "none"))]
 #[panic_handler]
 fn panic(info: &core::panic::PanicInfo<'_>) -> ! {
-    println!("Panic: {}", info);
+    log::error!("Panic: {}", info);
```
stefano-garzarella, line 43: mmm, are we sure about that? What happen if we panic before or while setting the logger?

Makefile
n-ramacciotti: > Do you mean that this is enabled by all configs/*json?

yes (and also in `FEATURES_TEST`)

stefano-garzarella: okay, please make it more clear in the commit message

kernel/Cargo.toml
n-ramacciotti: Without this change, `log::debug` (and trace, but those are not present) prints are not included in the binary.

Additionally, we set when installing the logger:
```rust
// Log levels are to be configured via the log's library feature configuration.
log::set_max_level(log::LevelFilter::Trace);
 ```
 
 But `debug` and `trace` have been already filtered out. So I decided to use `max_level_trace` to keep things coherent. We could also use `max_level_debug`

kernel/src/log_buffer.rs
n-ramacciotti: I will update the comment.

user/lib/src/lib.rs
n-ramacciotti: I followed the kernel handler, where `log::error` is used. If a panic happens before setting the logger, it's a nop operation and nothing is printed. I'm not sure about what happens (or if it can happen) while setting it.

I can leave the `println` there.

kernel/Cargo.toml
stefano-garzarella: See commit b5cdca1cb8ba4ef4051e2b8520baaf0f30a9b3a4

stefano-garzarella: > Additionally, we set when installing the logger:

Why we set trace?

user/lib/src/lib.rs
stefano-garzarella: If we already do that in kernel, it is fine, I was just worried about that.

kernel/Cargo.toml
n-ramacciotti: > See commit https://github.com/coconut-svsm/svsm/commit/b5cdca1cb8ba4ef4051e2b8520baaf0f30a9b3a4

Is this the same reason for why it has been converted to `max_level_info`? In this case I can drop the commit and update also `userlib/src/cargo.toml` with the same level

> Why we set trace?

The default value is `off` so nothing is logged unless another value is set. I'm not sure why `Trace` was selected here. Maybe it was introduced with `max_level_trace` and in https://github.com/coconut-svsm/svsm/commit/b5cdca1cb8ba4ef4051e2b8520baaf0f30a9b3a4 was not changed

stefano-garzarella: > > See commit [b5cdca1](https://github.com/coconut-svsm/svsm/commit/b5cdca1cb8ba4ef4051e2b8520baaf0f30a9b3a4)
> 
> Is this the same reason for why it has been converted to `max_level_info`? In this case I can drop the commit and update also `userlib/src/cargo.toml` with the same level

I just saw commit ec376f3287f554371a256ca30cbae44855f72463

n-ramacciotti: > In this case I can drop the commit and update also userlib/src/cargo.toml with the same level

I will drop the commit. Thanks for sharing these commits.

n-ramacciotti:
v5:
- updated "makefile: Remove broken DISABLE_CONSOLE_LOG" commit description with where the feature is enabled 
- restored `max_level_info` in kernel and introduced in `userlib`
- updated comment on `set_logger`

stefano-garzarella (approved):

joergroedel (approved):
Looks good to me, including testing.

PR #1167, opened 2026-07-13
virtio/mmio: account for page offset in transport size

luigix25:
Since commit https://github.com/luigix25/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855 ("virtio/mmio: handle
non-page-aligned MMIO device addresses") added support for unaligned
MMIO slots, the assumption that each slot occupies exactly one page no
longer holds, making the comment about QEMU placing each slot on its
own page obsolete.

This also means that using a whole page for `MmioTransport` is wrong:
if the offset is non-zero and the header starts at page_base + offset,
passing PAGE_SIZE as the size causes the transport to go past the
boundaries of the allocated region.

Fix this by subtracting the page offset from the total size.

Also use `mem.size()` instead of `PAGE_SIZE`. This is not a functional
change, but makes code easier to maintain as the size is derived from
the mapped region rather than repeated as a literal.

/cc @osteffenrh

### DIFF
--- a/kernel/src/virtio/mmio.rs
+++ b/kernel/src/virtio/mmio.rs
@@ -99,8 +99,11 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
         // SAFETY: The address is valid, mapped by `map_global_range_4k_shared`, and verified
         // to be VirtIOHeader-aligned by the guard above.
         // The memory region has the same lifetime of the MmioSlot structure which will be consumed by the driver.
-        // Note: QEMU places each MMIO slot on its own page for us, thus mmio_size = PAGE_SIZE.
-        let Ok(transport) = (unsafe { MmioTransport::new(header, PAGE_SIZE) }) else {
+        // TODO: Currently the hypervisor does not advertise the size of the mmio space (header + config space).
+        //       The size will be provided by the device tree. For now, let's just make sure we don't go past
+        //       the boundaries of the allocated space.
+        let Ok(transport) = (unsafe { MmioTransport::new(header, mem.size() - page_offset) })
+        else {
             // Currently QEMU advertises _all_ slots, regardless they are empty or not.
             log::debug!("MmioSlots: {addr:x} empty");
             continue;


osteffenrh (approved):
Thanks!

kernel/src/virtio/mmio.rs
stefano-garzarella, line 1: Sorry, I don't understand the commit description: `mem` is allocated as `map_global_range_4k_shared(page_base, PAGE_SIZE, PTEntryFlags::data())` so `mem.size()` is exactly `PAGE_SIZE`, so this is just a cosmetic/semantic change, but in the end there is no functional change, right? (should be clarified IMO).

That said, perhaps there is a problem introduced by commit https://github.com/coconut-svsm/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855: `header` is at `mem.addr() + page_offset`, so I'm not sure it is fine to use `PAGE_SIZE` or `mem.size()` as size for `MmioTransport::new()`, but we should consider the offset, or am I missing something?

luigix25: > Sorry, I don't understand the commit description: `mem` is allocated as `map_global_range_4k_shared(page_base, PAGE_SIZE, PTEntryFlags::data())` so `mem.size()` is exactly `PAGE_SIZE`, so this is just a cosmetic/semantic change, but in the end there is no functional change, right? (should be clarified IMO).

No functional change, correct.

> That said, perhaps there is a problem introduced by commit [a109b5d](https://github.com/coconut-svsm/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855): `header` is at `mem.addr() + page_offset`, so I'm not sure it is fine to use `PAGE_SIZE` or `mem.size()` as size for `MmioTransport::new()`, but we should consider the offset, or am I missing something?

To be picky, this problem was introduced by commit 516425a2d717fee0cf70fcdd55f98399678ce980. But yes, I'll look into that

osteffenrh: Oh, totally right!

The mmio area does not need to be page-aligned anymore. So `PAGE_SIZE` is already wrong in the first place. But `mem.size()` is the same, as you pointed out. So this does not help.
We need to know how large the full mmio area is (header/transport config space + device config space). OR at least an upper bound for this that is valid for all devices that we expect to find behind a transport.

stefano-garzarella: > > Sorry, I don't understand the commit description: `mem` is allocated as `map_global_range_4k_shared(page_base, PAGE_SIZE, PTEntryFlags::data())` so `mem.size()` is exactly `PAGE_SIZE`, so this is just a cosmetic/semantic change, but in the end there is no functional change, right? (should be clarified IMO).
> 
> No functional change, correct.

So, this should be mentioned in the commit description IMO if we want to merge this version.

> 
> > That said, perhaps there is a problem introduced by commit [a109b5d](https://github.com/coconut-svsm/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855): `header` is at `mem.addr() + page_offset`, so I'm not sure it is fine to use `PAGE_SIZE` or `mem.size()` as size for `MmioTransport::new()`, but we should consider the offset, or am I missing something?
> 
> To be picky, this problem was introduced by commit [516425a](https://github.com/coconut-svsm/svsm/commit/516425a2d717fee0cf70fcdd55f98399678ce980). 

Why ?

osteffenrh: in microvm the spacing is 512 bytes. So I assume all devices fit into that? We could just default to that, maybe. Since we do not know the actual size. Maybe with a check that mmio areas do not overlap.

osteffenrh (changes requested):

kernel/src/virtio/mmio.rs
luigix25: This is tricky: `header` considers the offset, so the config space starts at that address and is one page long.
So this means that we can theoretically go past the page boundaries.

Ideally we'd need to set the proper size of the mmio area, but we don't have that information for now. (this will come with device tree support). The only thing we can do IMO, is to subtract the offset to page size.

WDYT?

stefano-garzarella: > > > That said, perhaps there is a problem introduced by commit [a109b5d](https://github.com/coconut-svsm/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855): `header` is at `mem.addr() + page_offset`, so I'm not sure it is fine to use `PAGE_SIZE` or `mem.size()` as size for `MmioTransport::new()`, but we should consider the offset, or am I missing something?
> > 
> > 
> > To be picky, this problem was introduced by commit [516425a](https://github.com/coconut-svsm/svsm/commit/516425a2d717fee0cf70fcdd55f98399678ce980).
> 
> Why ?

Just to be clear, I mentioned that commit because we have this in the PR descritpion:

> Since commit https://github.com/coconut-svsm/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855 ("virtio/mmio: handle non-page-aligned MMIO device addresses") added support for unaligned MMIO slots, the assumption that each slot occupies exactly one page no longer holds, making the comment about QEMU placing each slot on its own page obsolete.

So I thought before that commit we never supported unaligned MMIO, but maybe I'm missing something.

luigix25: > > > > That said, perhaps there is a problem introduced by commit [a109b5d](https://github.com/coconut-svsm/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855): `header` is at `mem.addr() + page_offset`, so I'm not sure it is fine to use `PAGE_SIZE` or `mem.size()` as size for `MmioTransport::new()`, but we should consider the offset, or am I missing something?
> > > 
> > > 
> > > To be picky, this problem was introduced by commit [516425a](https://github.com/coconut-svsm/svsm/commit/516425a2d717fee0cf70fcdd55f98399678ce980).
> > 
> > 
> > Why ?
> 
> Just to be clear, I mentioned that commit because we have this in the PR descritpion:
> 
> > Since commit [a109b5d](https://github.com/coconut-svsm/svsm/commit/a109b5d55e7b52aa2c35a225c3ce41902178f855) ("virtio/mmio: handle non-page-aligned MMIO device addresses") added support for unaligned MMIO slots, the assumption that each slot occupies exactly one page no longer holds, making the comment about QEMU placing each slot on its own page obsolete.
> 
> So I thought before that commit we never supported unaligned MMIO, but maybe I'm missing something.

Before that commit we did not support unaligned MMIO, correct. What I mean is that first we allowed unaligned mmio and then we switched to upstream virtio-drivers.

If you look closely, with the downstream virtio-drivers, we did not have to set a size for the mmio area when creating an MmioTransport. This only came with the upstream virtio-drivers.

osteffenrh: size was assumed to be PAGE_SIZE in the old version of the drivers.

stefano-garzarella: > This is tricky: `header` considers the offset, so the config space starts at that address and is one page long. So this means that we can theoretically go past the page boundaries.
> 
> Ideally we'd need to set the proper size of the mmio area, but we don't have that information for now. (this will come with device tree support). The only thing we can do IMO, is to subtract the offset to page size.
> 
> WDYT?

Yep, but please add a TODO comment mentioning this is the best effort we can do for now, device tree support will allow us to do better

luigix25: Fixed!

luigix25:
- Subtracting page_offset to mem.size() to not go past beyond the boundaries [@stefano-garzarella]
- Added TODO to remind us that we need to update this when we'll support device tree
- updated commit message

kernel/src/virtio/mmio.rs
```diff
@@ -99,8 +99,11 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
         // SAFETY: The address is valid, mapped by `map_global_range_4k_shared`, and verified
         // to be VirtIOHeader-aligned by the guard above.
         // The memory region has the same lifetime of the MmioSlot structure which will be consumed by the driver.
-        // Note: QEMU places each MMIO slot on its own page for us, thus mmio_size = PAGE_SIZE.
-        let Ok(transport) = (unsafe { MmioTransport::new(header, PAGE_SIZE) }) else {
+        // TODO: Currently the hypervisor does not advertise the size of the mmio space (header + config space).
+        //       The size will be provided by the device tree. For now, let's just make sure we don't go past
+        //       the boundaries of the allocated space.
+        let Ok(transport) = (unsafe { MmioTransport::new(header, mem.size() - page_offset) })
```
stefano-garzarella, line 105: What about mentioning the `PAGE_SIZE` -> `mem.size()` change in the commit message?

luigix25:
- updated commit message [@stefano-garzarella]

stefano-garzarella (approved):

osteffenrh (approved):

PR #1169, opened 2026-07-15
github/worflows: update QEMU branch reference from svsm-igvm to svsm

luigix25:
The coconut-svsm/qemu repository default branch was changed from svsm-igvm to svsm.

Update QEMU_REF accordingly.

### DIFF
--- a/.github/workflows/qemu.yml
+++ b/.github/workflows/qemu.yml
@@ -11,7 +11,7 @@ env:
   # IGVM C wrapper might require a different Rust version than SVSM
   RUST_VERSION_IGVM: 1.84.1
   QEMU_REPO: https://github.com/coconut-svsm/qemu.git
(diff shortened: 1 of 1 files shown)


stefano-garzarella:
> The coconut-svsm/qemu repository default branch was changed from svsm-igvm to svsm.

It’s best to provide more context here:

since the last rebase that added support for KVM planes, the `svsm` and `svsm-igvm` branches of the QEMU fork have been identical. Since our documentation specifies that we should use `svsm`, we want to avoid having to manage two branches and would like to remove `svsm-igvm`, so let’s prepare the CI.

luigix25:
updated the commit message, thanks @stefano-garzarella for the hint!

stefano-garzarella (approved):

stefano-garzarella (approved):

PR #1170, opened 2026-07-15
kernel: fs/filesystem: simplify file handle structures

00xc:
Reorganize and simplify file handle code:
* Allow performing some operations on a file handle without its lock (e.g. querying read/write permissions, file size). The handle lock is only used to synchronize updates to the file offset for the next read/write, not to the file contents themselves.
* Use the returned value from `File::truncate()` to simplify `FileHandle::truncate()`.
* Remove `RawFileHandle`, as it loses its purpose after the previous changes.

### DIFF
--- a/kernel/src/fs/filesystem.rs
+++ b/kernel/src/fs/filesystem.rs
@@ -21,8 +21,12 @@ use alloc::vec::Vec;
 #[derive(Debug)]
 struct RawFileHandle {
     file: Arc<dyn File>,
-    /// current file offset for the read/write operation
-    current: usize,
+    /// Current file offset for read/write operations. Use a spinlock
+    /// to serialize updates when reading or writing using this handle.
+    /// Synchronized access to the file itself from other handles is
+    /// handled by the File trait itself, as it is Sync and its methods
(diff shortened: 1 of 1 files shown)


kernel/src/fs/filesystem.rs
```diff
@@ -17,28 +17,45 @@ extern crate alloc;
...
+    /// Check whether FileHandle is open for reading.
+    #[inline]
+    pub fn readable(&self) -> bool {
+        self.read
+    }
+
+    /// Check whether FileHandle is open for writing.
+    #[inline]
+    pub fn writable(&self) -> bool {
+        self.read
```
stefano-garzarella:
```suggestion
        self.write
```

stefano-garzarella (changes requested):
LGTM! There is just a copy&paste (I guess) issue to fix.

kernel/src/fs/filesystem.rs
00xc: Woops, yes, fixed.

stefano-garzarella (approved):

PR #1171, opened 2026-07-16
kernel: fs: do not clone task name on stdout open

00xc:
When creating the stdout files, the task name is cloned so that the log prefix can be preformatted. However, the prefix itself is heap-allocated, so a simple &str is enough.


00xc:
Closing in favor of a more complete PR that will remove more allocations

PR #1172, opened 2026-07-16
kernel: remove more unnecessary allocations

00xc:
In preparation to introduce fallible allocations, remove more unnecessary allocations, as these will introduce extra points of failure without good reason. This will also minimize the diff when switching to the new infrastructure:
* Do not ask for an owned `String` when creating a new `LineBuffer`.
* Store task names in `Arc<str>`, which allows cloning the name without allocating.
* Allow reusing a `String` allocation in `exec_user()`.
* Allow using some `Directory` methods without an owned `FileName` (which aliases to `String`).
* Remove unnecessary uses of `FileName` in some tests.


stefano-garzarella (approved):
Thanks, I can't see anything wrong, feel free to merge or wait other comments.

PR #1174, opened 2026-07-22
virtio: remove shared page

luigix25:
Currently a shared page is created, on all platforms, for the MMIO header + config space.
This is not really necessary, because all the mmio reads and writes on SNP go through the GHCB, therefore no mapping is needed. It is only required for native mode.

This PR moves the mapping creation to the `SvsmPlatform` trait. In this way it's only created when necessary.

/cc @osteffenrh who came up this idea during an offline discussion.

### DIFF
--- a/kernel/src/block/virtio_blk.rs
+++ b/kernel/src/block/virtio_blk.rs
@@ -23,7 +23,7 @@ use alloc::boxed::Box;
 
 pub struct VirtIOBlkDriver {
     device: SpinLock<VirtIOBlk<SvsmHal, MmioTransport<'static>>>,
-    _mmio_space: GlobalRangeGuard,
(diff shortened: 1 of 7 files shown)


luigix25:
QEMU can expose devices at physical address 0.

virtio-drivers::MmioTransport takes a NonNull argument for the address, and because of the above, we cannot guarantee a non-null address. So, some changes are required in virtio-drivers first.

I'm not fully sure it's worth it.

Closing this for now.

osteffenrh:
but we could turn the shared mapping into a private one, at least?

osteffenrh:
Thanks for testing this out. has been on my mind for a while.

luigix25:
> but we could turn the shared mapping into a private one, at least?

yes indeed! Doesn't make any sense to have it shared!

PR #1177, opened 2026-07-23
kernel/vsock: fix flaky double_connect test

stefano-garzarella:
Replace ncat with an inline python vsock server that gives us direct control over the listening socket lifecycle.
This is in preparation of the fix for the test_virtio_vsock_double_connect test that was flaky because it relied on timing for the second connect() to fail, assuming the server had already closed its listening socket by the time the guest attempted the second connection:
```
        test vsock::stream::tests::test_virtio_vsock_double_connect ...
        ERROR: Panic on CPU[3]! COCONUT-SVSM Version: 2026.06-devel-120-g54cd9681
        ERROR: Info: panicked at kernel/src/vsock/stream.rs:184:14:
        The second connection operation was expected to fail, but it succeeded.: VsockStream { local_port: 1025, r
emote_port: 12345, remote_cid: 2 }
        ---BACKTRACE---:
          [ffffff8000092792] core::panicking::panic_bounds_check+0x0
          [ffffff8000042d24] svsm::vsock::stream::tests::test_virtio_vsock_double_connect+0xf4
          [ffffff800001d8f9] <svsm::vsock::stream::tests::test_virtio_vsock_double_connect::{closure#0} as core::ops::function::FnOnce<()>>::call_once+0x9
          [ffffff8000011f66] svsm::testing::svsm_kernel_test_runner+0xc6
          [ffffff8000052109] svsm::svsm_bin::test_in_svsm_task+0x19
          [ffffff80000953d4] @ thread_entry_asm+0x18
        ---END---
```

### DIFF
--- a/kernel/src/vsock/stream.rs
+++ b/kernel/src/vsock/stream.rs
@@ -178,7 +178,14 @@ mod tests {
 
         let remote_port = start_vsock_server_host();
 
-        VsockStream::connect(remote_port, VMADDR_CID_HOST).expect("connection failed");
+        let mut stream =
+            VsockStream::connect(remote_port, VMADDR_CID_HOST).expect("connection failed");
+
+        // Read the message to ensure the server has accepted the connection
+        // and closed its listening socket before we attempt the second connect.
+        let mut buffer = [0u8; 11];
+        stream.read(&mut buffer).expect("read failed");
+        drop(stream);
 
         VsockStream::connect(remote_port, VMADDR_CID_HOST)
             .expect_err("The second connection operation was expected to fail, but it succeeded.");
--- a/scripts/test-in-svsm.sh
+++ b/scripts/test-in-svsm.sh
@@ -30,7 +30,7 @@ check_required_tools() {
 
 # Only the non-coreutils binaries are listed here, since coreutils is assumed to
 # be installed. Add any future non-coreutils dependencies to this list.
-check_required_tools ncat xxd ss
+check_required_tools python3 xxd
 
 test_io(){
     PIPE_IN=$1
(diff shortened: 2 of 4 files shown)


stefano-garzarella:
CC @luigix25

scripts/test-in-svsm.sh
```diff
@@ -30,7 +30,41 @@ check_required_tools() {
...
+    conn.sendall(b"hello_world")
+    # virtio-vsock in SVSM does not handle half-duplex connections,
+    # so keep the connection open until the peer closes it.
+    while conn.recv(1024):
+        pass
+except OSError:
+    pass
+conn.close()
+' "$1" "$2"
+}
```
00xc: Isn't this "big" enough to deserve a standalone .py script?

stefano-garzarella: maybe yes, I put there to have all test handlers in the same file, but I can move it in a standalone python file if you prefer

kernel/src/vsock/stream.rs
luigix25, line 1: probably this test can be dropped altogether: this was added when vsockstream had to be closed manually using a "close" function. This is not the case anymore. The only way to create a new connection is to create a new stream.

stefano-garzarella: but can it be useful as regression test?

luigix25: yes, at the end of the day we are still testing that reconnecting on a previously used port with no-one listening, fails.

scripts/test-in-svsm.sh
```diff
@@ -30,7 +30,41 @@ check_required_tools() {
...
+    pipe.write(struct.pack(">I", port))
+conn, _ = sock.accept()
+sock.close()
+try:
+    conn.sendall(b"hello_world")
+    # virtio-vsock in SVSM does not handle half-duplex connections,
+    # so keep the connection open until the peer closes it.
+    while conn.recv(1024):
+        pass
+except OSError:
```
luigix25: not a python expert, is this needed? does python sends a half shutdown after there are no more sends?

stefano-garzarella: I don't fully understand your question, but here if we remove this loop, the socket will be closed and Linux will send the shutdown. Is it okay for SVSM?

luigix25: no it is not, SVSM is not able to handle half-duplex connections ( as you correctly wrote in the comment )

scripts/test-in-svsm.sh
stefano-garzarella: Done

stefano-garzarella:
v2:
- moved the inline python script in a standalone python file [@00xc]

luigix25 (approved):

joergroedel (approved):

PR #1179, opened 2026-07-28
scripts: Add script to run all fuzzers

joergroedel:
Add a script which runs all fuzzers in parallel for default runtime of 600 seconds. The runtime is configurable with the -r/--runtime parameter. Log files will be written to the logs/fuzz/ directory.

The script will query a list of available fuzzers, build them, and run all of them in parallel for the configured amount of time. When one fuzzer reports a failure it will immediately terminate all others and exit with an error message. Written with codex.


scripts/run-fuzzers.sh
```diff
@@ -0,0 +1,195 @@
...
+		echo "Fuzzer '$fuzzer' failed with exit code $status"
+		echo "See log: $log_file"
+		failed=1
+		stop_fuzzers
+		wait 2> /dev/null || true
+		exit 1
+	fi
+done
+
+if [ "$failed" -eq 0 ]; then
```
stefano-garzarella: At this point `failed` is alwasy 0, because we exit when we set it at 1, so I guess we can just remove `failed` at all.

joergroedel: Right, removed `failed` and the check.

stefano-garzarella (approved):

PR #1180, opened 2026-07-29
virtio/mmio: map MMIO regions as private

luigix25:
In SNP mode, all MMIO accesses are routed through GHCB, so having a shared mapping for the MMIO area is not required. Note that the mapping cannot be removed altogether because the driver expects non-null virtual addresses, and hypervisors can expose devices at physical address 0.

In native mode, there is no concept of shared page, and mapping it as private or shared is equivalent.

Switch to private mapping as it's sufficient and we can reduce the amount of unencrypted pages.

Suggested-by: Oliver Steffen <osteffen@redhat.com>

### DIFF
--- a/kernel/src/virtio/mmio.rs
+++ b/kernel/src/virtio/mmio.rs
@@ -14,7 +14,7 @@ use virtio_drivers::transport::{DeviceType, Transport, mmio::MmioTransport};
 use crate::address::{Address, PhysAddr};
 use crate::boot_params::BootParams;
 use crate::fw_cfg::FwCfg;
-use crate::mm::{GlobalRangeGuard, map_global_range_4k_shared, pagetable::PTEntryFlags};
+use crate::mm::{GlobalRangeGuard, map_global_range_4k_private, pagetable::PTEntryFlags};
 use crate::platform::SVSM_PLATFORM;
 use crate::types::PAGE_SIZE;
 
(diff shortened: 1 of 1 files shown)


osteffenrh (approved):

kernel/src/virtio/mmio.rs
```diff
@@ -88,15 +88,16 @@ pub fn probe_mmio_slots(boot_params: &BootParams<'_>) -> MmioSlots {
         // If multiple devices reside in the same page, each gets its own
         // mapping, which is slightly wasteful but avoids shared-ownership
         // complexity.
-        let Ok(mem) = map_global_range_4k_shared(page_base, PAGE_SIZE, PTEntryFlags::data()) else {
```
stefano-garzarella, line 91: Can you add a comment here to explain why we are mapping it as private since it's not intuitive IMO.

luigix25: done

luigix25:
- added a comment explaining why we are using a private mapping [@stefano-garzarella ]

stefano-garzarella (approved):

joergroedel (commented):
No, that is not how it should be done, sorry.

In general this is not a breaking change in the current code-base, but it will break should we ever decide to take #VC/#VE on MMIO accesses.

A better improvement here would be to split out the `mmio_read()` and `mmio_write()` methods of the platform trait into a separate mmio trait. An object (different types per platform) implementing this trait represents one MMIO region and would own the mapping.

luigix25:
Closing this for now. I'll reopen/open a new PR with Joerg's suggestions

PR #1183, opened 2026-08-04
kernel/attest: add some unit tests

luigix25:
This PR adds some unit tests for the attestation protocol and early boot attestation. More details in the commit messages. This is a rework of the PR #1002 

I dropped commit "test/attest: add kbs-test flow for SNP attestation" as the doc changes are already available. For the script, that takes care of setting up attestation infrastructure, I'll open a followup PR. 

I dropped some tests:
- `rejects_unknown_request` it had a really low value
- `reserved_clear_rejects_each_field` was doing the same work of `check_valid_rejects_nonzero_reserved`
- guidtable tests: even though the table is part of the svsm spec, it was testing rust's internal implementation.

All commit messages have been reworded and fixed clippy and cargo fmt checks

### DIFF
--- a/kernel/src/attest.rs
+++ b/kernel/src/attest.rs
@@ -422,3 +422,75 @@ fn hash(
 
     try_to_vec(&sha.finalize()).or(Err(AttestationError::VecAlloc))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloc::vec;
+    use cocoon_tpm_tpm2_interface::{Tpm2bEccParameter, TpmBuffer};
+
+    fn make_ecc_point(x: &[u8], y: &[u8]) -> TpmsEccPoint<'static> {
+        TpmsEccPoint {
+            x: Tpm2bEccParameter {
+                buffer: TpmBuffer::Owned(x.to_vec()),
+            },
+            y: Tpm2bEccParameter {
+                buffer: TpmBuffer::Owned(y.to_vec()),
+            },
+        }
+    }
+
+    mod negotiation_hash {
+        use super::*;
+
+        // The challenge size is server-dictated; 48 is an arbitrary choice.
+        const CHALLENGE_LEN: usize = 48;
+        // P-521 coordinate size.
+        const COORD_LEN: usize = 66;
+
+        /// hash() feeds NegotiationParams into SHA-512 in the order they
+        /// appear in `response.params`. Verify that [Challenge, EcPublicKeyBytes]
+        /// and [EcPublicKeyBytes, Challenge] each produce the correct digest and
+        /// that the two digests differ — the server-negotiated ordering must
+        /// affect the resulting attestation evidence.
+        #[test]
+        fn hash_respects_param_ordering() {
+            let challenge = vec![0xdd; CHALLENGE_LEN];
+            let x = vec![0x10; COORD_LEN];
+            let y = vec![0x20; COORD_LEN];
+            let pub_key = make_ecc_point(&x, &y);
+
+            let response_chal_first = NegotiationResponse {
+                challenge: challenge.clone(),
+                params: vec![
+                    NegotiationParam::Challenge,
+                    NegotiationParam::EcPublicKeyBytes,
+                ],
+            };
+            let response_key_first = NegotiationResponse {
+                challenge: challenge.clone(),
+                params: vec![
+                    NegotiationParam::EcPublicKeyBytes,
+                    NegotiationParam::Challenge,
+                ],
+            };
+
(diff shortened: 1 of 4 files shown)


kernel/src/attest.rs
```diff
@@ -422,3 +422,75 @@ fn hash(
 
     try_to_vec(&sha.finalize()).or(Err(AttestationError::VecAlloc))
 }
+
+#[cfg(test)]
```
stefano-garzarella, line 402: Should we add `attest` in `FEATURES_TEST` in the Makefile, so these new tests can be executed with `make test` ?

kernel/src/protocols/attest.rs
```diff
@@ -556,3 +584,109 @@ pub fn attest_protocol_request(
...
+            let mut bytes = base_op().as_bytes().to_vec();
+            bytes[offset_of!(AttestSingleServiceOp, manifest_ver)] = 1;
+            let op = AttestSingleServiceOp::ref_from_bytes(&bytes).unwrap();
+            let err = op.check_valid().unwrap_err();
+            assert!(is_invalid_parameter(&err));
+        }
+
+        /// AttestSingleServiceOp embeds an AttestServicesOp. Parsing must
+        /// validate the reserved fields of the inner struct too, not just
+        /// its own reserved_5.
```
stefano-garzarella: Should we test `reserved_5` like the others in `check_valid_rejects_nonzero_reserved()` ?

stefano-garzarella (commented):
nit: the right tag is `Co-developed-by` (lowercase 'd') as documented in `Documentation/docs/developer/CONTRIBUTING.md`:

> At the end of the patch, it needs to be signed off with a
```Signed-off-by``` tag as created by ```git commit -s```. If a patch
was written by more than one person, then additional developers can be
added via a ```Co-developed-by``` tag.

File: kernel/src/attest.rs (hunk 1 above)
Comment (luigix25) [reply kernel/src/attest.rs:402]: right good point! we need to move the attestation in `svsm.rs` after test are started, otherwise test-in-svsm would wait for the aproxy connection.

File: kernel/src/protocols/attest.rs (hunk 2 above)
Comment (luigix25) [reply kernel/src/protocols/attest.rs]: on a second thought we should just check `reserved_5`. All the other fields are checked in `check_valid_rejects_nonzero_reserved` WDYT?

Comment (stefano-garzarella) [reply kernel/src/protocols/attest.rs]: yep, I meant: "since we are checking all the others, should we check also this one?"

**Comment by luigix25** [comment]
Addressed @stefano-garzarella comments:

- Disabled attestation for test-in-svsm, otherwise running unit test with `attest` enabled would not work. This can be reverted once the followup of this PR, that adds a fake attestation, is completed.
- Checking for reserved_5
- Rebased to latest upstream

File: kernel/src/svsm.rs [hunk 3]
```diff
@@ -13,7 +13,7 @@ use core::arch::global_asm;
 use core::panic::PanicInfo;
 use core::ptr::NonNull;
 use svsm::address::{Address, PhysAddr, VirtAddr};
-#[cfg(feature = "attest")]
+#[cfg(all(feature = "attest", not(test_in_svsm)))]
 use svsm::attest::AttestationDriver;
```
stefano-garzarella: What about moving this inside the block where we do attestation  (line 556) ?

In this way we reduce the `cfg` around the code.

BTW this is pre-existing, feel free to ignore, we can fix later.

kernel/src/svsm.rs
```diff
@@ -78,7 +78,7 @@ use svsm::vtpm::vtpm_init;
 
 use release::COCONUT_VERSION;
 
-#[cfg(feature = "attest")]
+#[cfg(all(feature = "attest", not(test_in_svsm)))]
 use kbs_types::Tee;
```
stefano-garzarella: Ditto.

kernel/src/svsm.rs
stefano-garzarella, line 1:
```
kernel/svsm: skip attestation when running test-in-SVSM

Add `attest` to FEATURES_TEST so attestation unit tests are compiled in and can
run. Guard the attestation block and its imports to prevent the attestation
process from being triggered during test runs as they would fail, as there is
no proxy waiting for connections.
```
nit: the commit title IMO is confusing, what we are doing here IIUC is to enable `attest` when doing `make test` or `make test-in-svsm`. The fact that we are skipping attestation for test-in-svsm is a detail, not the end goal of this patch.

luigix25:
@stefano-garzarella 
- Changed commit title
- moved imports into the attestation block

stefano-garzarella (approved):

PR #1184, opened 2026-08-05
kernel/virtio: update safe-mmio to version 0.3.1

luigix25:
safe-mmio version 0.3.1 includes the `custom-mmio` feature that we need. This allows us to drop the `[patch.crates-io]` section and consume the crate directly from crates.io.

This new version removed the custom_mmio submodule and re-exports MmioOps directly from the crate root. The trait is also no longer unsafe to implement, so drop the unsafe qualifier.


luigix25:
cc @osteffenrh

stefano-garzarella (approved):

osteffenrh (approved):
thx :-D

PR #1186, opened 2026-08-06
kernel: protocols/core: require CAA GPA to be 4K-aligned

00xc:
The SVSM spec (Table 6 REMAP_CA, Table 10 CREATE_VCPU) specifies that the GPA for the Calling Area must be aligned to 4 KB, but the current code only enforces an 8-byte alignment. Moreover, the spec refers to the Calling Area as a full page, meaning that the entirety of that page should be bounds- and overlap-checked.

Fix the `REMAP_CA` and `CREATE_VCPU` handlers to check for the correct alignment, bounds and overlaps. Since the GPA is now guaranteed to be page-aligned on the success path, there is no longer a need to compute a page offset.

### DIFF
--- a/kernel/src/protocols/core.rs
+++ b/kernel/src/protocols/core.rs
@@ -103,13 +103,14 @@ fn core_create_vcpu(params: &RequestParams) -> Result<(), SvsmReqError> {
     let apic_id: u32 = (params.r8 & 0xffff_ffff) as u32;
 
     // Check alignment
-    if !paddr.is_page_aligned() || !pcaa.is_aligned(8) {
+    if !paddr.is_page_aligned() || !pcaa.is_page_aligned() {
         return Err(SvsmReqError::invalid_address());
     }
 
     let vmsa_region =
         MemoryRegion::checked_new(paddr, PAGE_SIZE).ok_or(SvsmReqError::invalid_address())?;
-    let caa_region = MemoryRegion::checked_new(pcaa, 8).ok_or(SvsmReqError::invalid_address())?;
+    let caa_region =
+        MemoryRegion::checked_new(pcaa, PAGE_SIZE).ok_or(SvsmReqError::invalid_address())?;
 
     // Check for region overlap
     if vmsa_region.overlap(&caa_region)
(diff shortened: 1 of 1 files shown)


kernel/src/protocols/core.rs
stefano-garzarella, line 1: LGTM, but Claude found a potential followup in `kernel/src/cpu/percpu.rs`:

>   - C1 (kernel/src/cpu/percpu.rs:1424): Incomplete change — CAA overlap checks in 
>   VmsaRegistry still use 8-byte regions. The commit updates core_create_vcpu and
>   core_remap_ca to treat the CAA as a full page (PAGE_SIZE), but
>   PERCPU_VMSAS.register() (line 1424, 1427) and PERCPU_VMSAS.remap_ca() (line 1455)
>   still construct MemoryRegion::new(caa, 8) and MemoryRegion::new(new_caa, 8) for
>   their overlap checks. This means the registry's internal overlap detection is
>   inconsistent with the caller's: two CAAs (or a CAA and VMSA) whose pages overlap but
>   whose first 8 bytes don't would pass the registry check. Similarly, overlaps() at
>   line 1411 uses MemoryRegion::new(vmsa.caa, 8). All four sites need updating to
>   PAGE_SIZE.

I guess we should do:
```diff
--- a/kernel/src/cpu/percpu.rs
+++ b/kernel/src/cpu/percpu.rs
@@ -1408,7 +1408,7 @@ impl PerCpuVmsas {
     pub fn overlaps(&self, region: &MemoryRegion<PhysAddr>) -> bool {
         self.vmsas.lock_read().iter().any(|vmsa| {
             region.overlap(&MemoryRegion::new(vmsa.paddr, PAGE_SIZE))
-                || region.overlap(&MemoryRegion::new(vmsa.caa, 8))
+                || region.overlap(&MemoryRegion::new(vmsa.caa, PAGE_SIZE))
         })
     }
 
@@ -1421,10 +1421,10 @@ impl PerCpuVmsas {
     ) -> Result<(), SvsmError> {
         let mut guard = self.vmsas.lock_write();
         let vmsa_region = MemoryRegion::new(paddr, PAGE_SIZE);
-        let caa_region = MemoryRegion::new(caa, 8);
+        let caa_region = MemoryRegion::new(caa, PAGE_SIZE);
         if guard.iter().any(|vmsa| {
             let vr = MemoryRegion::new(vmsa.paddr, PAGE_SIZE);
-            let cr = MemoryRegion::new(vmsa.caa, 8);
+            let cr = MemoryRegion::new(vmsa.caa, PAGE_SIZE);
             vr.overlap(&vmsa_region)
                 || vr.overlap(&caa_region)
                 || cr.overlap(&vmsa_region)
@@ -1452,7 +1452,7 @@ impl PerCpuVmsas {
         let mut guard = self.vmsas.lock_write();
         if guard.iter().any(|vmsa| {
             let r1 = MemoryRegion::new(vmsa.paddr, PAGE_SIZE);
-            vmsa.caa == new_caa || r1.overlap(&MemoryRegion::new(new_caa, 8))
+            vmsa.caa == new_caa || r1.overlap(&MemoryRegion::new(new_caa, PAGE_SIZE))
         }) {
             return Err(SvsmError::InvalidAddress);
         }
```

00xc: This causes a boot failure. I think the change is not correct because the guest is allowed to use the rest of the page for other purposes, and particularly OVMF uses it to place the buffer for CORE_PVALIDATE.

With the change, the overlap checks in `valid_phys_region()` are too strict and `core_pvalidate()` fails very early.

stefano-garzarella: > This causes a boot failure. I think the change is not correct because the guest is allowed to use the rest of the page for other purposes, and particularly OVMF uses it to place the buffer for CORE_PVALIDATE.

@00xc I'm confused, is OVFM violating SVSM spec or this is allowed?

00xc: This is allowed:
> The initially configured SVSM Calling Area is a page of memory that lies outside the initial SVSM
memory range and has not had its VMPL permissions restricted in any way. The address is
guaranteed to be aligned to a 4 KB boundary, so the remainder of the page may be used by the
guest for memory-based parameter passing if desired.

stefano-garzarella: Okay, I (and I guess Claude too) was confused by what we have in the cover letter/commit: 

> the spec refers to the Calling Area as a full page

Maybe we should clear it a bit in the commit, BTW I'm going to approve, up to you to fix it or not.

stefano-garzarella (approved):

joergroedel (approved):

PR #1187, opened 2026-08-06
kernel/alloc.verus: reenable verification in refill_page_list

ziqiaozhou:
Enable verification by adding proper loop invariant. Remove old proof code when it was implemented as a recursive function.

### DIFF
--- a/kernel/src/mm/alloc.rs
+++ b/kernel/src/mm/alloc.rs
@@ -911,15 +911,12 @@ impl HeapMemoryRegion {
     }
 
     /// Refills the free page list for a given order.
-    #[verus_verify(external_body)]
     #[verus_spec(ret =>
         requires
             old(self).wf_next_pages(),
             0 <= order <= MAX_ORDER,
         ensures
             old(self).ens_refill_page_list(*final(self), ret.is_ok(), order),
-        decreases
-            MAX_ORDER - order,
     )]
     fn refill_page_list(&mut self, order: usize) -> Result<(), AllocError> {
         proof! {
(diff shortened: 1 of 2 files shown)


stefano-garzarella:
mmmm, why CI didn't start? Maybe some GH issue

luigix25:
> mmmm, why CI didn't start? Maybe some GH issue

yep, they had an outage yesterday

stefano-garzarella:
I just pushed an empty commit, and force pushed removing it. It seems it triggered the CI.

kernel/src/mm/alloc.rs
```diff
@@ -934,22 +931,46 @@ impl HeapMemoryRegion {
...
+                |ord| self.next_page[*ord] != NO_PAGE,
+            )
             .ok_or(AllocError::OutOfMemory)?;
-
-        proof_decl! {
-            let tracked mut perm = PgUnitPerm::empty(arbitrary());
-        }
-
         // Split the page down to the order we want
+        #[cfg_attr(verus_keep_ghost, verus_spec(iter=>
```
stefano-garzarella, line 955: From Claude reviewing:

> - (kernel/src/mm/alloc.rs:955): The loop invariant uses
>   #[cfg_attr(verus_keep_ghost, verus_spec(iter=> ...)], while the analogous loop in
>   mark_compound_page (line 818) uses #[cfg_attr(verus_keep_ghost_body, 
>   verus_spec(...))]. Both functions are now non-external_body, so both should use the
>   same attribute gate. This inconsistency is not a bug (the correct cfg for a verified
>   body is verus_keep_ghost), but either mark_compound_page should be migrated to
>   verus_keep_ghost in a follow-up, or this commit should document why the two differ.
>   As-is it creates a confusing precedent for future contributors.

kernel/src/mm/alloc.verus.rs
stefano-garzarella, line 336: > - [Pre-existing] (kernel/src/mm/alloc.verus.rs:335-336): rhs_pfn and new_order
>   are defined but never used in ens_split_page_ok. The parameter pfn is also
>   effectively unused (only consumed by the dead rhs_pfn binding). These are harmless
>   dead bindings in spec code, but they clutter the spec and could mislead readers
>   about what the postcondition actually constrains.

stefano-garzarella (approved):
Thanks LGTM! 
Claude pointed out 2 small issues. They seems real, but please double check. In any case feel free to fix here or in another PR if they need to be fixed.

joergroedel (approved):

kernel/src/mm/alloc.rs
joergroedel: @ziqiaozhou Please check and fix in a follow-on PR, if needed.

kernel/src/mm/alloc.verus.rs
joergroedel: @ziqiaozhou Please check and fix in a follow-on PR, if needed.

PR #1188, opened 2026-08-11
kernel/sec/ghcb: add missing tlb flush in GhcbPage::drop.

ziqiaozhou:
Security issue without the fix:

If the page is allocated again for other purpose, the CPU may see a stale TLB cache and causes unexpected behavior (e.g., write unencrypted content to a private page).


ziqiaozhou:
@joergroedel  Since #1062 may not be merged, I created this PR to fix the TLB issue I noticed when I refactored the page table code.

msft-jlange (approved):

joergroedel (approved):

PR #1190, opened 2026-08-11
Verus: cleanup spec

ziqiaozhou:
This PR remove unused spec functions and variables inside spec functions. They were not detected by cargo clippy since they are inside verus!. This is to address the comment in #1187


stefano-garzarella:
@ziqiaozhou thanks for this! I'm not reviewing this since is still draft, so let me know when I should review it.

stefano-garzarella (approved):

Issue #659, opened 2025-03-27
GuestPtr vs UserPtr request for clarification and Documentation


deeglaze:
In https://github.com/coconut-svsm/svsm/pull/653 I'm attempting to add arbitrary length reads and writes to contiguous guest memory that is outside the purview of both the SVSM kernel and SVSM userspace.

Whereas SVSM controls the page tables for its own userspace, it doesn't control the page tables for GUEST_VMPL. This is where I understand page mapping guards to be important, to directly map physical memory. The mm::memory::valid_phys_address check ensures that an address is outside memory claimed for SVSM use (both kernel and userspace).

In this commit, (GitHub doesn't point to the PR that introduced it, so I can't find discussion around it)
https://github.com/coconut-svsm/svsm/commit/3121843f01a2926d35fea81e92ce910871d7177a I'm confused about wrapping GuestPtr. It seems to me that SVSM user space reads and writes should trap and return the equivalent of -EFAULT when a read or write goes bad, rather than require the pages to be mapped ahead of time in a guard. It also doesn't make sense to have a construct that checks physical address range since user vs. kernel should be determined by a supervisor bit in the PTE that stac/clac change the faulting behavior for.

Whereas the fault-checking read/write is there in copy_bytes for both GuestPtr and UserPtr, I think we should have separate concepts. First, GuestPtr and UserPtr are translated to constructs on top of slices so we can read/write a dynamically sized amount of memory (Vectors).

The GuestPtr and UserPtr type bounds do not account for the direct byte representation needed when crossing ABI boundaries. We don't always need a FromBytes bound and IntoBytes bound, so reads and writes need an asymmetric API.

/cc @vijaydhanraj for comment about guestmem UserPtr in case I misunderstand

00xc:
I believe everything here was addressed with #1191, closing. Feel free to reopen if there is anything missing.

PR #1191, opened 2026-08-11
kernel: mm/guestmem: `GuestPtr` / `UserPtr` rework

00xc:
Rework the `GuestPtr` and `UserPtr` interfaces.

The SVSM kernel accesses three distinct categories of memory that require different safety properties: its own kernel address space, lower-VMPL guest physical memory, and SVSM userspace. The existing `GuestPtr<T>` type conflates all of these under a single (misleading) name and does not enforce the different invariants each category requires.

Rework the code as follows:
* Rename `GuestPtr` to `TryPtr`, a low level primitive with fault handling and unsafe methods.
* `UserPtr`: a `TryPtr` wrapper for userspace access, with address space bounds checks, SMAP handling and safe methods.
* The new `GuestPtr`: a `TryPtr` wrapper for lower-VMPL memory access, with address space bounds checks, an explicit lifetime bound to a `PerCPUPageMappingGuard` and safe methods.
* Add slice support for `TryPtr`, `UserPtr` and `GuestPtr`.
* Migrate existing users to the new APIs.

I believe this PR addresses everything noted in #659

### DIFF
--- a/kernel/src/mm/guestmem.rs
+++ b/kernel/src/mm/guestmem.rs
@@ -20,7 +20,7 @@ use core::arch::asm;
 use core::borrow::Borrow;
 use core::ffi::c_char;
 use core::mem::{MaybeUninit, size_of};
-use core::ptr::{self, NonNull, with_exposed_provenance, with_exposed_provenance_mut};
+use core::ptr::{self, NonNull};
 use core::slice;
 use syscall::PATH_MAX;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, TryFromBytes};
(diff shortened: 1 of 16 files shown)


kernel/src/mm/guestmem.rs
```diff
@@ -431,42 +432,390 @@ impl<T> GuestPtr<T> {
...
+        unsafe { self.ptr.read(index) }
+    }
+
+    /// Writes `val` to element at `index`.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SvsmError::InvalidAddress`] if `index >= len`, or
+    /// [`SvsmError::Fault`] if the access faults.
+    pub fn write(&self, index: usize, val: T) -> Result<(), SvsmError>
```
n-ramacciotti, line 752: Would it make sense to also add `write_ref`?

00xc: Yes, but looking at it again, I think `write()` and `write_ref()` can be merged into a single method by using a generic parameter. I'll add this to the next version.

kernel/src/mm/guestmem.rs
```diff
@@ -402,71 +407,390 @@ impl<T> GuestPtr<T> {
...
+    ///
+    /// # Safety
+    ///
+    /// The caller must verify that the entire backing memory is appropriate
+    /// to read.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SvsmError::Fault`] if the copy faults.
+    pub unsafe fn read_to_vec(&self) -> Result<Vec<T>, SvsmError> {
```
joergroedel: This lacks a `T : FromBytes`, no? Similar to the `T : IntoBytes` in `write_from_slice()`

kernel/src/mm/guestmem.rs
```diff
@@ -696,8 +1028,10 @@ pub unsafe fn copy_from_guest(src: PhysAddr, dst: *mut u8, size: usize) -> Resul
 /// If the physical address region is not allocated to the guest, it returns
 ///   `Err(SvsmError::InvalidAddress)`.
 pub fn copy_slice_from_guest(src: PhysAddr, dst: &mut [u8]) -> Result<(), SvsmError> {
-    // SAFETY: The safety property of slices ensures the memory is owned and mutable.
-    unsafe { copy_from_guest(src, dst.as_mut_ptr(), dst.len()) }
+    let src_end = src.checked_add(dst.len()).ok_or(SvsmError::Mem)?;
+    PerCPUPageMappingGuard::create(src.page_align(), src_end.page_align_up(), 0)?
```
joergroedel: `src_end.page_align_up()` can overflow, I think this needs to be checked before passed to `PerCPUPageMappingGuard::create()`.

joergroedel (commented):
Looks good overall, left two comments.

kernel/src/mm/guestmem.rs
```diff
@@ -402,71 +407,393 @@ impl<T> GuestPtr<T> {
...
+    ///   fall outside guest memory.
+    /// * [`SvsmError::Mem`] if integer overflow occurs.
+    pub fn guest_slice<T>(
+        &self,
+        offset: usize,
+        len: usize,
+    ) -> Result<GuestPtr<'_, [T]>, SvsmError> {
+        let start = self.phys_base().checked_add(offset).ok_or(SvsmError::Mem)?;
+        let size = len.checked_mul(size_of::<T>()).ok_or(SvsmError::Mem)?;
+        let region = checked_guest_region(start, size)?;
```
joergroedel, line 655: This works when the virtual mapping is also physically contiguous. But there is the `create_4k_pages()` method, which allows to create mappings of non-contiguous physical memory. It is not a problem because `create_4k_pages()` has no users, but I suggest you remove it with this PR to avoid any future checking problems.

kernel/src/mm/guestmem.rs
```diff
@@ -584,57 +892,119 @@ impl UserPtr<c_char> {
...
+impl<T> UserPtr<[T]> {
+    /// Constructs a `UserPtr<[T]>` pointing to `len` elements at `v`,
+    /// checking that the entire object falls within userspace.
+    #[inline]
+    pub fn new(v: VirtAddr, len: usize) -> Result<Self, SvsmError> {
+        let userspace = MemoryRegion::from_addresses(USER_MEM_START, USER_MEM_END);
+        let size = len
+            .checked_mul(size_of::<T>())
+            .ok_or(SvsmError::InvalidAddress)?;
+        let region = MemoryRegion::new(v, size);
```
joergroedel: I think this needs `MemoryRegion::checked_new()` to catch overflows, no?

00xc: Yes, also in `UserPtr::<T>::new()`

joergroedel (approved):
I added a commit on-top to fix a small issue found by cargo-verify. Otherwise changes look great, thanks for the effort!

PR #1193, opened 2026-08-12
kernel/persistence: update usable inode start to 16

stefano-garzarella:
CocoonFS reserves inodes 0 through 15 for internal filesystem use (SpecialInode::ReservedLast = 15 in cocoon-tpm-storage src/fs/cocoonfs/inode_index.rs). The Demo inode was already correctly allocated at 16; update the comment to match.

### DIFF
--- a/kernel/src/persistence.rs
+++ b/kernel/src/persistence.rs
@@ -1077,7 +1077,7 @@ pub fn persistence_read_inode_sync(inode: u64) -> Result<Option<Zeroizing<Vec<u8
 
 /// Persistence inode numbers allocated statically for specific SVSM uses.
 ///
-/// Usable inode numbers start at `6`.
+/// Usable inode numbers start at `16`.
 #[derive(Debug)]
 #[repr(u64)]
 pub enum SvsmPersistenceStaticInode {


stefano-garzarella:
CC @nicstange for confirmation. I discovered this while working on vTPM state.
About that, what I should do with the demo when I have the vTPM state?
drop it and use 16 for the vTPM NVData? or use 17 and leave 16 for the demo or future use cases ?

nicstange:
> CC @nicstange for confirmation. I discovered this while working on vTPM state. About that, what I should do with the demo when I have the vTPM state? drop it and use 16 for the vTPM NVData? or use 17 and leave 16 for the demo or future use cases ?

Yeah, drop the demo, I'd say, unless it would still be useful for the attestation development work. For the inode number, it's completely up to you. Maybe make the upper 32bits non-zero -- the idea was to split the 64 bit inode numbers in two halves: the upper 32 bits would get statically allocated to the respective SVSM service and the lower half would get managed by that service then.

Perhaps it would make sense to define something like
```
enum SvsmInodeSpaceHi32 {
   Tpm = 1,
   Uefi = 2,
   ...
}
```
at a central place, like in `persistence.rs`? Then we could have
```
const TPM_STATE_INODE_NUMBER: u64 = (SvsmInodeSpaceHi32::Tpm as u64) << 32;
```

nicstange (approved):
Change LGTM, thanks!

joergroedel (approved):

PR #1195, opened 2026-08-13
kernel/vtpm: enable persistent NV state

stefano-garzarella:
This PR add a `CFileBuffer` struct to implement fopen, fclose, fread, fwrite, etc. C API used by TCG TPM to persist NV data through the CocoonFS persistence API.

When the `persistence` feature is not enabled or the CocoonFS storage is not available (e.g. no block device), the NV data is kept in memory so the TPM is re-manufactured on every boot as before.

~TODO:~
- ~[x] handle better inode allocation as suggested by @nicstange in https://github.com/coconut-svsm/svsm/pull/1193#issuecomment-5268050029~

How I tested it:
```
## shell1
## build svsm, tools and start aproxy

export FW_FILE=/usr/share/edk2/ovmf/OVMF.amdsev.fd

cd svsm
cargo xbuild -f attest,vsock,attest-serial,virtio-drivers,persistence configs/qemu-target.json
make bin/aproxy bin/igvmmeasure

bin/aproxy --protocol kbs --url http://0.0.0.0:8080 --vsock


## shell2
## create cocoonfs state image with encryption key and start
## kbs-test registering the expected launch measurement and the key
## to decrypt the state

HEX_SECRET=793212d775af0b2a3e6186a9b701bfefd440d4a048e58fbac4240b913efb58f8

cd cocoon-tpm
cargo run -p cocoonfs-cli -- -i ~/cocoonfs.img -f mkfs -K $HEX_SECRET -H sha2 -C aes -t 128 -I 'ddeeff' -s 8M

cd kbs-test
SVSM=/path/to/svsm
MEASUREMENT="$($SVSM/bin/igvmmeasure --check-kvm $SVSM/bin/coconut-qemu.igvm measure -b)"
cargo run -- --measurement $MEASUREMENT --secret $HEX_SECRET


## shell3
## start the CVM with vsock and state file

cd svsm
export QEMU=/path/to/qemu-system-x86_64
export IMAGE=/path/to/guest.qcow2

./scripts/launch_guest.sh --vsock 42 --state ~/cocoonfs.img --snapshot off

# wipe any pre-existing tpm2 key slots
guest$ systemd-cryptenroll /dev/sda3 --wipe-slot=tpm2

# seal the LUKS passphrase with the TPM also using some PCRs as policies
guest$ systemd-cryptenroll /dev/sda3 --tpm2-device=auto \
  --tpm2-pcrs=0,1,4,5,7,9

guest$ poweroff

# after the reboot, the disk will be automatically unlocked and mounted
```


stefano-garzarella:
Mark this as Draft since I still need to address a todo, but I'd like some comments about:
- TCG TPM interface (i.e. enabing FILE_BACKED_NV=YES and intercepting fopen/fwrite/fread/etc.)
- CocoonFS API usage

CC @nicstange @jejb @armenon-rh

kernel/src/svsm.rs
```diff
@@ -567,8 +567,6 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
         #[cfg(feature = "persistence")]
         if let Some(persistence_bootstrap_info) = persistence_bootstrap_info {
             persistence_init(persistence_bootstrap_info, &secret).unwrap();
-            // TODO: remove once we have real persistence users.
-            persistence_demo();
```
nicstange: ACK, but `./Documentation/docs/developer/PERSISTENCE.md` should have its mention of the demo at the very end removed as well. Probably best to just squash that into this revert here.

kernel/src/vtpm/tcgtpm/mod.rs
```diff
@@ -190,36 +191,47 @@ impl VtpmInterface for TcgTpm {
...
-        // 2. Make sure it does not fail if it is re-manufactured
-        // 3. Teardown to indicate it needs to be manufactured
-        // 4. Manufacture it for the first time
-        // 5. Power it on indicating it requires startup. By default, OVMF will start
+        // 1. Check if the NV state needs manufacturing
+        // 2. If manufacturing is needed:
+        //    a. Manufacture it for the first time
+        //    b. Make sure it does not fail if it is re-manufactured
+        //    c. Teardown to indicate it needs to be manufactured
+        //    d. Manufacture it for the first time
```
nicstange: I know you carried this logic over from what had been there, but does this manufacture-remanufacture dance make much sense for us?

kernel/src/vtpm/tcgtpm/cfile.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        if !persistence_available() {
+            return None;
+        }
+
+        // Treat errors as fatal: retries are already exhausted inside
+        // persistence_read_inode_sync(), and returning None would let
+        // NVMem.c create an empty NV image and overwrite the valid
+        // persistent state, destroying TPM keys and seeds.
+        persistence_read_inode_sync(self.inode).expect("CocoonFsBackend: failed to load NV inode")
```
nicstange: Regarding the `expect()`, I would also add a comment that the `load()` gets effectively called only once at startup, so the panic would happen then or never.

kernel/src/vtpm/tcgtpm/cfile.rs
```diff
@@ -0,0 +1,190 @@
...
+    fn create(backend: B) -> Self {
+        Self {
+            buf: Zeroizing::new(Vec::new()),
+            pos: 0,
+            backend,
+        }
+    }
+
+    pub fn save(&self) -> Result<(), SvsmError> {
+        self.backend.save(self.buf.clone())
```
nicstange: `persistence_write_inode_sync()` returns the input buffer back on success, and the TCG TPM reference implementation would enter failure mode upon encountering a `NVCommit()` failure from command execution anyway.  As an optimization for avoiding the buffer `clone()` here, how about
- making `self.buf` and `Option`, which is always `Some`, except for if a write error has happened,
- temporarily steal it here and install it back when done with the writing,
-  and make `save()`,  `fread()`, `fwrite()`, `fseek()` and `ftell()` return an error if it's `None`?

kernel/src/vtpm/tcgtpm/cfile.rs
```diff
@@ -0,0 +1,190 @@
...
+        dst[..n].copy_from_slice(&self.buf[self.pos..self.pos + n]);
+        self.pos += n;
+        n / element_size
+    }
+
+    pub fn fwrite(&mut self, src: &[u8], element_size: usize) -> usize {
+        if element_size == 0 || src.is_empty() {
+            return 0;
+        }
+        let Some(end) = self.pos.checked_add(src.len()) else {
```
nicstange: In the `fread()` counterpart, the length is aligned to the `element_size`, should that be done here as well?

nicstange (commented):
Hi Stefano,

very nice! :)

I would add a warning somewhere about the known security limitations at this development stage though. Two main issues come to my mind:
- IIRC, the storage key obtained from the KBS is currently not authenticated, hence the whole persistence thing is prone to MITM attacks (at manufacturing time at least). The `Documentation/docs/developer/PERSISTENCE.md` mentions it already, but perhaps it makes sense to repeat it at a more prominent place.
- Even if that was sorted out, the attestation (using the vtpm's `TPM2_Quote`) cannot be relied upon at this point. That's because we don't have any rollback protection implemented yet, and one could start two CVMs off the same persistent state, which is in violation of the TPM spec effectively requiring unique TPM instances. To be more specific, an adversary could spin up a "good" VM yielding the expected PCRs, and a rogue one, both from the same state. He could then obtain and mix `TPM2_Quote` results from either and provide them to a relying party. Like a quote on the PCRs from the good one, and a TPM key certification quote from the bad one. As these `TPM2_Quote` results would all contain the same `resetCount` (taken from NV) as part of the `TPMS_ATTEST`' `clockinfo` field, the relying party would wrongly assume that the quotes all originate from the same `_TPM_Init` cycle from a unique instance. It might consider this instance to be in a "good" state (because of the PCRs reported from the good VM) and proceed to encrypting some secret to the TPM key controlled by the rogue one.

Hope it makes some sense...

Thanks!

Nicolai

kernel/src/vtpm/tcgtpm/cfile.rs
stefano-garzarella: yeah, I was thinking the same since I didn't like the `clone()` too. Since we are in 2, let's do that ;-)

kernel/src/vtpm/tcgtpm/mod.rs
stefano-garzarella: IIUC this was done by @cclaudio following what the TCG TPM simulator does, so I'm not sure if we should change it, in anycase I don't think it's material for this PR TBH.

What issues do you see on doing this?

kernel/src/svsm.rs
stefano-garzarella: Oh, I thought all demo stuff was contained in a single commit so I just reverted that one. I'll update it.

kernel/src/vtpm/tcgtpm/cfile.rs
stefano-garzarella: Good point. This really depends on when TPM code calls `fopen()`, but so far is in that way, so I'll add it.

kernel/src/vtpm/tcgtpm/mod.rs
nicstange: Skimming through the TCG code, the remanufacturing seems redundant/pointless to me. But agreed, it would probably make sense to handle in a separate  PR.

kernel/src/vtpm/tcgtpm/cfile.rs
stefano-garzarella: yep, good catch, that `size`/`count` behavior made me crazy....

stefano-garzarella:
> Hi Stefano,
> 
> very nice! :)

thanks to you for CocoonFS :-)

> 
> I would add a warning somewhere about the known security limitations at this development stage though. Two main issues come to my mind:
> 
>     * IIRC, the storage key obtained from the KBS is currently not authenticated, hence the whole persistence thing is prone to MITM attacks (at manufacturing time at least). The `Documentation/docs/developer/PERSISTENCE.md` mentions it already, but perhaps it makes sense to repeat it at a more prominent place.

yeah, this really depends on the attestation protocol IMO and in Trustee KBS case we should disable the manufacturing in SVSM. Indeed @armenon-rh is working in an external tool that can be used to manufacture the TPM offline in a controlled environment without runinng SVSM at all. I'll document that, and in the future maybe add a way to disable it.

> 
>     * Even if that was sorted out, the attestation (using the vtpm's `TPM2_Quote`) cannot be relied upon at this point. That's because we don't have any rollback protection implemented yet, and one could start two CVMs off the same persistent state, which is in violation of the TPM spec effectively requiring unique TPM instances. To be more specific, an adversary could spin up a "good" VM yielding the expected PCRs, and a rogue one, both from the same state. He could then obtain and mix `TPM2_Quote` results from either and provide them to a relying party. Like a quote on the PCRs from the good one, and a TPM key certification quote from the bad one. As these `TPM2_Quote` results would all contain the same `resetCount` (taken from NV) as part of the `TPMS_ATTEST`' `clockinfo` field, the relying party would wrongly assume that the quotes all originate from the same `_TPM_Init` cycle from a unique instance. It might consider this instance to be in a "good" state (because of the PCRs reported from the good VM) and proceed to encrypting some secret to the TPM key controlled by the rogue one.

Yep, I think this is a common issue of software TPMs and IIRC Dionna mentioned that the TCG was working on extending the spec to cover vTPM cases and defining something for the rollback. That said, I will document these limitations.

Thanks,
Stefano

kernel/src/vtpm/tcgtpm/mod.rs
stefano-garzarella: But this is exactly what the TCG TPM simulator code does, see https://github.com/TrustedComputingGroup/TPM/blob/3b1f0e623f244abe2015df9e3ba86de5c9118524/TPMCmd/Simulator/src/TPMCmds.c#L235

stefano-garzarella: (not sure why TBH)

stefano-garzarella:
v2:
- fixed `make doc` issue
- added `InodeNamespace` following https://github.com/coconut-svsm/svsm/pull/1193#issuecomment-5268050029 [@nicstange]
- added documentation about current limitations [@nicstange]
- fixed `fread()` length alignment [@nicstange]
- avoided buffer `clone()` [@nicstange]
- refactored a bit the code to avoid too many `cfg` in cfile.rs

stefano-garzarella:
v3:
- gated `mod cfile` to make clippy happy

kernel/src/persistence.rs
```diff
@@ -1075,11 +1075,35 @@ pub fn persistence_read_inode_sync(inode: u64) -> Result<Option<Zeroizing<Vec<u8
...
+/// The 64-bit CocoonFS inode number is split into two halves: the upper
+/// 32 bits are statically assigned to each service here, and the lower
+/// 32 bits are managed by the service itself. Namespace 0 is reserved:
+/// although CocoonFS currently only reserves the first 16 inodes for
+/// internal use, all inodes with the upper 32 bits set to zero are kept
+/// reserved for simplicity.
+#[derive(Clone, Copy, Debug)]
+#[repr(u32)]
+pub enum InodeNamespace {
+    Reserved = 0,
```
00xc: Can't we just **not** define `Reserved`? E.g. define the `Tpm = 1` variant only, that way we don't need the `debug_assert!()` below.

kernel/src/persistence.rs
```diff
@@ -1075,48 +1075,36 @@ pub fn persistence_read_inode_sync(inode: u64) -> Result<Option<Zeroizing<Vec<u8
...
+    Tpm = 1,
 }
```
00xc, line 1090: We could just *not* define the `Reserved` variant, that way we would not need the debug assert below. Alternatively, this could be converted to a compile-time assertion:

```diff
diff --git a/kernel/src/persistence.rs b/kernel/src/persistence.rs
index 27cc7c6b3951..73f46fbf357a 100644
--- a/kernel/src/persistence.rs
+++ b/kernel/src/persistence.rs
@@ -1090,7 +1090,6 @@ pub enum InodeNamespace {
 
 impl InodeNamespace {
     fn inode(self, local: u32) -> u64 {
-        debug_assert!(!matches!(self, Self::Reserved));
         (self as u64) << 32 | local as u64
     }
 }
@@ -1105,6 +1104,9 @@ pub trait Inode: Into<u32> {
     const NAMESPACE: InodeNamespace;
 
     fn inode(self) -> u64 {
+        const {
+            assert!(!matches!(Self::NAMESPACE, InodeNamespace::Reserved));
+        }
         Self::NAMESPACE.inode(self.into())
     }
 }
```

kernel/src/vtpm/tcgtpm/cfile.rs
```diff
@@ -0,0 +1,245 @@
...
+        self.pos = end;
+
+        Ok(n / element_size)
+    }
+
+    pub fn fseek(&mut self, offset: isize, whence: i32) -> Result<usize, SvsmError> {
+        let buf = self
+            .buf
+            .as_ref()
+            .ok_or(SvsmError::FileSystem(FsError::BadHandle))?;
```
00xc: This construct is used in pretty much every method, maybe some helpers could be useful, e.g.

```rust
impl<B: StorageBackend> CFileBuffer<B> {
    fn buf(&self) -> Result<&Zeroizing<Vec<u8>>, SvsmError> {
        self.buf.as_ref().ok_or(FsError::BadHandle.into())
    }

    fn buf_mut(&mut self) -> Result<&mut Zeroizing<Vec<u8>>, SvsmError> {
        self.buf.as_mut().ok_or(FsError::BadHandle.into())
    }
    ...
}
```

kernel/src/persistence.rs
stefano-garzarella: Yeah, I can add the compile-time assertion, but should we keep also the debug_assert() to cover cases where the `Inode` trait is not used?

Not a strong opinion, at runtime will fail if some reserved id are used, so I can remove it.

joergroedel (commented):
Two findings from codex. The first one looks valid, not sure about the second one. Please check:


The persistence integration can leave an unmanufactured image looking valid after an interrupted first boot, and it cannot replace oversized NV images correctly. Both cases prevent reliable vTPM state initialization across boots.

Full review comments:

  - [P2] Persist the manufacturing state before trusting the NV image — /home/joro/src/svsm/kernel/src/vtpm/tcgtpm/mod.rs:213-215. If the VM stops during first boot after NVMem.c::_plat__NVEnable() flushes the initial zero-filled image but before TPM_Manufacture() completes, the next boot sees an exact-sized file and _plat__NVNeedsManufacture() returns false. This branch then skips manufacturing and starts the TPM with unmanufactured NV state; persist a manufacturing marker atomically or validate the loaded state rather than inferring completion from file size.

  - [P2] Truncate oversized NV images when reinitializing — /home/joro/src/svsm/kernel/src/vtpm/tcgtpm/cfile.rs:193-196. When a persisted NVChip inode is larger than the current NV_MEMORY_SIZE—for example after a format/configuration change—NVMem.c opens it with r+b and overwrites the fixed-size image, but this implementation retains the trailing bytes because it only grows the buffer. The inode therefore remains oversized, causing _plat__NVEnable() to manufacture a fresh TPM on every subsequent boot instead of saving the replacement state.

stefano-garzarella:
v4:
- added compile-time assertion in the `Inode` trait [@00xc]
- added `buf` and `buf_mut` helpers in `CFileBuffer` [@00xc]

stefano-garzarella:
> * [P2] Persist the manufacturing state before trusting the NV image — /home/joro/src/svsm/kernel/src/vtpm/tcgtpm/mod.rs:213-215. If the VM stops during first boot after NVMem.c::_plat__NVEnable() flushes the initial zero-filled image but before TPM_Manufacture() completes, the next boot sees an exact-sized file and _plat__NVNeedsManufacture() returns false. This branch then skips manufacturing and starts the TPM with unmanufactured NV state; persist a manufacturing marker atomically or validate the loaded state rather than inferring completion from file size.

I thought the checks in the TCG TPM code was more than just file size, but this is true, there is a NvFileCommit() call just after the file creation full of `0xFF`.

In long term maybe we should provide our own `_plat__*` functions, since on some of them I see:
```
// This version just pulls in data from a file. In a real TPM, with NV on chip,
// this function would verify the integrity of the saved context. If the NV
// memory was not on chip but was in something like RPMB, the NV state would be
// read in, decrypted and integrity checked.
```

I need to figure out a better way to prevent this; maybe we could prevent `fflush` from actually writing anything until we're sure the manufacturing process has completed successfully. Or using something like the the marker suggested by codex. Maybe the easiest way is to use another inode to write when the manufacturing is done and use it as marker.

@nicstange WDYT?

> 
> * [P2] Truncate oversized NV images when reinitializing — /home/joro/src/svsm/kernel/src/vtpm/tcgtpm/cfile.rs:193-196. When a persisted NVChip inode is larger than the current NV_MEMORY_SIZE—for example after a format/configuration change—NVMem.c opens it with r+b and overwrites the fixed-size image, but this implementation retains the trailing bytes because it only grows the buffer. The inode therefore remains oversized, causing _plat__NVEnable() to manufacture a fresh TPM on every subsequent boot instead of saving the replacement state.

it could also be a real problem, I think we can just call `_plat__NVDisable(1, 0)`, where 1 means delete the file content, before calling `TPM_Manufacture()`. I'll check that better, and at this point I'll re-valuate @nicstange comment in https://github.com/coconut-svsm/svsm/pull/1195#discussion_r3781353561 looking in the C code, it seems they are called just for coverage test, so I agree that we can simplify that.

kernel/src/vtpm/tcgtpm/mod.rs
kraxel: Given that the vTPM seems to be a bit more complex, can we split + fast-track the non-vTPM changes in another PR so I have something to work with for uefi vars persistance?

stefano-garzarella: sure, I can split the first 3 commits in another PR, so you can reserve another inodenamespace number for uefi.

@kraxel will that help? (I can also reserve 2 for UEFI in that PR if you prefer)

kraxel: Sounds good.  Reserving 2 for UEFI is fine too.

stefano-garzarella:
Marking this as draft since it now depends on #1199

kernel/src/vtpm/tcgtpm/mod.rs
cclaudio: > IIUC this was done by @cclaudio following what the TCG TPM simulator does, so I'm not sure if we should change it

I also prefered to not change that procedure. IIRC, the TCG TPM can be configured to work in different scenarios. Although the simulator steps to initialize the TPM seem redundant, I guess they are required to make sure that the TPM will get to a reliable state at the end of the initialization - i.e. it will survive or die ... gracefully :-)

stefano-garzarella: @cclaudio thanks for replying!!
Looking better in the TCG TPM simulator code, it seems they are called just for coverage test, so I'll try to simplify it so we can also reduce writes on persistent storage.

stefano-garzarella:
v5:
- rebased on top of #1199 
- simplified manufacturing [@nicstange @cclaudio]
- added a marker (new inode) to avoid P1 in https://github.com/coconut-svsm/svsm/pull/1195#pullrequestreview-4970207707 [@joergroedel]
- called `_plat__NVDisable()` before manufacturing to avoid P2 in https://github.com/coconut-svsm/svsm/pull/1195#pullrequestreview-4970207707 [@joergroedel]

stefano-garzarella:
v6:
- rebased on main since #1199 is now merged
- no changes from v5

joergroedel:
Codex reports this as the last remaining issue:

>A persistent-storage read failure for the manufacturing marker is treated as an absent marker, which can trigger destructive remanufacturing of otherwise valid TPM state.
>
>Review comment:
>
>  - [P1] Propagate marker read errors before remanufacturing — /home/joro/src/svsm/kernel/src/vtpm/tcgtpm/persistence/mod.rs:75-77  If the NV inode loads successfully but reading the manufacturing-marker inode returns an error, matches! converts that error to false, causing init() to call manufacture() and truncate the valid NV state. This can permanently destroy TPM keys and seeds after a transient or marker-specific storage failure; propagate the read error or fail closed, as CocoonFsBackend::load() already does for the NV inode.

Not sure how we want to handle that case. On the one hand it only enables a DoS attack, on the other side the SVSM has options to handle the case. And it seems better to not overwrite a valid NV state due to previous read errors. WDYT?

stefano-garzarella:
> Codex reports this as the last remaining issue:
> 
> > A persistent-storage read failure for the manufacturing marker is treated as an absent marker, which can trigger destructive remanufacturing of otherwise valid TPM state.
> > Review comment:
> > 
> > * [P1] Propagate marker read errors before remanufacturing — /home/joro/src/svsm/kernel/src/vtpm/tcgtpm/persistence/mod.rs:75-77  If the NV inode loads successfully but reading the manufacturing-marker inode returns an error, matches! converts that error to false, causing init() to call manufacture() and truncate the valid NV state. This can permanently destroy TPM keys and seeds after a transient or marker-specific storage failure; propagate the read error or fail closed, as CocoonFsBackend::load() already does for the NV inode.
> 
> Not sure how we want to handle that case. On the one hand it only enables a DoS attack, on the other side the SVSM has options to handle the case. And it seems better to not overwrite a valid NV state due to previous read errors. WDYT?

Yeah, so I'll propagate that error and fails to init/manufacture the TPM at all, right?

joergroedel:
> Yeah, so I'll propagate that error and fails to init/manufacture the TPM at all, right?

Yeah, that is probably the best reaction to that error.

stefano-garzarella:
v7:
- `ManufacturedMarker::exists()` now returns an error to fix https://github.com/coconut-svsm/svsm/pull/1195#issuecomment-5354813827 [@joergroedel]
- updated `toc` header in ATTESTATION.md

joergroedel (approved):
Looks good now and my testing also passed.

PR #1196, opened 2026-08-14
kernel: test: do not force-exclude vtpm and uefivars

00xc:
Do not force-exclude building the vtpm and uefivars code for testing, as there is no particular reason to do so. In the past, vtpm code caused issues when building userspace tests, but this was fixed with 7c3b3c656063 ("libtcgtpm: do not link libcrt for userspace targets").

Remove the special handling for these modules, and use only the existing features to gate building the code, ensuring that the functionality is not broken in the future for test builds.

The user can always disable building these modules via e.g. `FEATURES_TEST` in the Makefile.

### DIFF
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -17,11 +17,11 @@ use crate::greq::{
     services::get_regular_report,
 };
 use crate::mm::guestmem::{copy_slice_to_guest, read_bytes_from_guest, read_from_guest};
-#[cfg(all(feature = "uefivars", not(test)))]
+#[cfg(feature = "uefivars")]
 use crate::protocols::uefivars::uefi_mm_get_manifest;
 use crate::protocols::{RequestParams, errors::SvsmReqError};
 use crate::utils::MemoryRegion;
-#[cfg(all(feature = "vtpm", not(test)))]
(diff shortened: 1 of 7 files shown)


kernel/src/protocols/attest.rs
stefano-garzarella: What about replacing also this?


```suggestion
#[cfg(any(feature = "vtpm", feature = "uefivars"))]
```

stefano-garzarella (commented):
LGTM, just 1 more thing we may change.

Thanks for this!

kernel/src/protocols/attest.rs
stefano-garzarella: not sure what is happening to GH, but my real suggestion is:
```diff
--- a/kernel/src/protocols/attest.rs
+++ b/kernel/src/protocols/attest.rs
@@ -214,8 +214,7 @@ impl GuidTable {
         }
     }
 
-    // Only used when features are enabled.
-    #[allow(dead_code)]
+    #[cfg(any(feature = "vtpm", feature = "uefivars"))]
     fn push(&mut self, guid: uuid::Uuid, data: Vec<u8>) {
         self.entries.push(GuidTableEntry { guid, data })
     }
```

stefano-garzarella (approved):

PR #1198, opened 2026-08-19
aproxy: various improvements

luigix25:
This PR adds some improvements to aproxy:
- don't abort on attestation failure
- move protocol version to a separate const
- remove unnecessary clone
- fix log message
- introduce log levels

### DIFF
--- a/tools/aproxy/Cargo.toml
+++ b/tools/aproxy/Cargo.toml
@@ -16,7 +16,9 @@ anyhow = "1.0.103"
 base64 = { workspace = true, features = ["std"] }
 clap = { version = "4.5", features = ["derive"] }
 const_format = "0.2"
+env_logger = "0.11"
 libaproxy.workspace = true
+log.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 vsock = "0.5.1"
--- a/tools/aproxy/src/backend/kbs.rs
+++ b/tools/aproxy/src/backend/kbs.rs
@@ -244,7 +244,9 @@ impl TryFrom<&AttestationRequest> for KbsEvidence {
 
                 Ok(Self::Snp {
                     snp_report: BASE64_STANDARD.encode(report),
-                    certs_buf: certs_buf.clone().map(|certs| BASE64_STANDARD.encode(certs)),
+                    certs_buf: certs_buf
+                        .as_ref()
+                        .map(|certs| BASE64_STANDARD.encode(certs)),
                 })
             }
             _ => Err(anyhow!("invalid TEE")),
--- a/tools/aproxy/src/backend/kbs.rs
+++ b/tools/aproxy/src/backend/kbs.rs
@@ -113,6 +113,7 @@ impl AttestationProtocol for KbsProtocol {
             .context("unable to POST to KBS /attest endpoint")?;
 
         if http_resp.status() != StatusCode::OK {
+            log::warn!("KBS /attest returned {}", http_resp.status());
             return Ok(AttestationResponse {
                 success: false,
                 secret: None,
@@ -146,6 +147,7 @@ impl AttestationProtocol for KbsProtocol {
 
         // Unsuccessful attempt at retrieving secret.
         if http_resp.status() != StatusCode::OK {
+            log::warn!("KBS /resource returned {}", http_resp.status());
             return Ok(AttestationResponse {
                 success: false,
                 secret: None,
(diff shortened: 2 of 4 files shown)


00xc (dismissed):

tools/aproxy/src/backend/kbs.rs
```diff
@@ -244,7 +244,9 @@ impl TryFrom<&AttestationRequest> for KbsEvidence {
 
                 Ok(Self::Snp {
                     snp_report: BASE64_STANDARD.encode(report),
-                    certs_buf: certs_buf.clone().map(|certs| BASE64_STANDARD.encode(certs)),
+                    certs_buf: certs_buf
```
stefano-garzarella, line 251: I'd move this change to another commit, it seems unrelated to the changes to Cargo.toml in https://github.com/coconut-svsm/svsm/pull/1198/changes/6af691877436494bd28e5f0c4ed29ffe81aed2ae

Maybe you can have 3 commits: fixes in kbs, main, and cargo.toml

luigix25: yep right. I split that commit into two.

tools/aproxy/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "aproxy"
 version = "0.1.0"
 edition.workspace = true
 
-[target.'cfg(all(target_os = "linux"))'.dependencies]
```
stefano-garzarella: Same pattern in other tools, maybe we should fix also them:

```
$ git grep "cfg(all(target_os"
tools/igvmbuilder/Cargo.toml:[target.'cfg(all(target_os = "linux"))'.dependencies]
tools/igvmmeasure/Cargo.toml:[target.'cfg(all(target_os = "linux"))'.dependencies]
```

tools/aproxy/src/main.rs
```diff
@@ -60,7 +60,9 @@ fn accept_loop<S: Read + Write>(
     for stream in incoming {
         let mut stream = stream.context("Failed to accept connection")?;
         let mut http_client = backend::HttpClient::new(url.to_string(), backend.into())?;
-        attest::attest(&mut stream, &mut http_client)?;
+        if let Err(e) = attest::attest(&mut stream, &mut http_client) {
```
stefano-garzarella, line 64: If we have some issue in this phase, should we also send `AttestationResponse` with `success: false` to SVSM?

It could be a follow up of course

tools/aproxy/Cargo.toml
stefano-garzarella: I'm also not really sure if we need it at all here in `aproxy` TBH

stefano-garzarella: ah okay, we need it (I did that change in those tools xD), maybe we should put here the same comment we used in commit cd84c5ff67b7cf15e72d64db18c96bfb7312319e
```
# specify dependencies' target to avoid feature unification with SVSM
# see https://doc.rust-lang.org/cargo/reference/features.html#feature-unification
[target.'cfg(all(target_os = "linux"))'.dependencies]
```

luigix25: yep, right. I'll add the comment.

tools/aproxy/src/main.rs
luigix25: yep good idea. This need some rework of the error handling in the attestation function.

tools/aproxy/Cargo.toml
stefano-garzarella: feel free to open another PR with just that commit and fix all the places together

luigix25: Opened here: #1200

luigix25:
- Dropped commit that was moved to #1200
- Will address @stefano-garzarella's suggestion as a follow-up

tools/aproxy/src/main.rs
```diff
@@ -60,7 +60,9 @@ fn accept_loop<S: Read + Write>(
...
+            eprintln!("attestation failed: {e:#}");
```
stefano-garzarella: Would be better to start to introduce `log` and use it in this case, but also improving logging in general, e.g. adding a log entry when there is a connection incoming from SVSM, when we fail to connect to KBS, etc.

It is okay also in a follow up (please open issues to track follow ups)

luigix25:
- Introduced logging [@stefano-garzarella]

stefano-garzarella (approved):

PR #1202, opened 2026-08-20
add persistence support for uefi variable store

kraxel:


kernel/src/protocols/uefivars.rs
```diff
@@ -129,9 +131,16 @@ mod persistance {
...
-            persistence_unlink_inode_sync(inode.inode(), true)?;
+
+        if !stale.is_empty() {
+            let mut ops = PersistenceMultiOp::new();
+            for inode in stale {
+                let r = inode.inode()..=inode.inode();
+                ops.stage_inode_range_unlinking(r)?;
+            }
+            persistence_commit_multi_op_sync(ops, true)?;
         }
```
nicstange: No need to stage at intermediate `stale`, you could directly stage at a `PersistenceMultiOp` instance when walking the inodes. If empty in the end, the commit will be a nop.

nicstange (commented):
Checked current `PersistenceMultiOp` usage only, find a comment inline.

A high-level comment: with `PersistenceMultiOp`, it would in principle be possible to keep the index-inode and the per-var inodes on storage in sync at all times, which could alleviate the need for the garbage collection -- perhaps worth considering...

kernel/src/protocols/uefivars.rs
kraxel: Updated all this.  The varstore crate does more detailed update tracking, and with that the write callback does all updates needed with a single MultiOp now.

PR #1203, opened 2026-08-20
kernel/attest: Add Trustee KBS as an attestation backend to SVSM

armenon-rh:
Remote Attestation currently fails when interacting with a production-grade [Trustee KBS (CoCo) server](https://github.com/confidential-containers/trustee).
This occurs due to backend differences in payload formats (JWS JSON vs binary), hashing algorithms (SHA-384 vs SHA-512), and key binding structures.

Because the runtime_data hash verified by the remote backend must be constructed inside SVSM to prevent tampering by an untrusted host, SVSM must handle these backend-specific formats directly.
Representing nested JWS JSON structures using the legacy NegotiationParam vector adds unnecessary complexity, driving a design change to make aproxy a simple HTTP forwarder instead.

This series:
1. Implements a simple HTTP forwarder in libaproxy.
2. Allow user to specify the protocol that the backend uses and the connection parameters via command line. Remove anything extra in aproxy backend.
3. Implement the payload formatting and hashing logic in SVSM. Make it extensible such that any backend can be added in the future.
4. Add KBS Trustee backend while also keeping the [kbs-test](https://github.com/coconut-svsm/kbs-test) backend compatible.

Todo:

- [ ] Add guide docs to run the setup.


armenon-rh:
Not looking at this PR now.

PR #1205, opened 2026-08-21
kernel/svsm: gracefully handle missing vsock device during attestation

luigix25:
Currently, when SVSM is compiled with attestation or persistence enable, and launched without a vsock device attached it panics because it is expecting such device to be present.

Relax this constraint and if such device is not present, just skip attestation.


kernel/src/svsm.rs
```diff
@@ -470,13 +470,25 @@ fn free_init_bsp_stack() {
...
+        Ok(p) => p,
+        Err(SvsmError::Vsock(VsockError::DeviceNotAvailable)) => {
+            log::warn!(
+                "attestation: vsock device not available, skipping attestation and persistence"
+            );
+            return;
+        }
+        Err(e) => panic!("attestation: failed to initialize transport: {e:?}"),
+    };
+
```
stefano-garzarella, line 488: Should we document this new behaviout in ATTESTATION.md ?

kernel/src/svsm.rs
```diff
@@ -470,13 +470,25 @@ fn free_init_bsp_stack() {
...
     let secret = proxy.attest().unwrap();
```
stefano-garzarella, line 489: Should we do something similar also here?
I mean, if the vsock device is available, but the proxy is not, here we will panic, right?

luigix25: right, will do it.

stefano-garzarella: on second thought I'm not sure about this, here we can fail for several reasons, also because the attestation failed.
In any case, I think we should document this behaviour too.

luigix25: there are several reasons yes. What if we check just for Vsock-related errors and panic for the attestation ones?

stefano-garzarella: yeah, maybe it is fine for now, but in a long term I'd like to define a clear policy on which kind of error we should just log and which other we should panic, but of course it's not for this PR.

@joergroedel any thoughts on this?

luigix25:
- Added some documentation [@stefano-garzarella]
- Not panicking for all vsock-related errors. [@stefano-garzarella]

kernel/src/svsm.rs
```diff
@@ -466,6 +466,37 @@ fn free_init_bsp_stack() {
...
+    #[cfg(feature = "persistence")]
+    let persistence_bootstrap_info = persistence_discover().unwrap();
+
+    let mut proxy = match AttestationDriver::try_from(Tee::Snp) {
+        Ok(p) => p,
+        Err(SvsmError::Vsock(e)) => {
+            log::warn!("attestation: vsock error, skipping attestation and persistence: {e:?}");
+            return;
+        }
+        Err(e) => panic!("attestation: failed to initialize: {e:?}"),
```
stefano-garzarella, line 486: We just discussed this in the TSC and we agreed that for now we can avoid the panic in any case, including attestation failure, just log it as error and skip the `persistence_init`.

kernel/src/svsm.rs
```diff
@@ -466,6 +466,37 @@ fn free_init_bsp_stack() {
...
+    };
+
+    let secret = proxy.attest().unwrap();
```
stefano-garzarella, line 489: Ditto, let's log it as error.

kernel/src/svsm.rs
```diff
@@ -466,6 +466,37 @@ fn free_init_bsp_stack() {
...
+    log::info!("attestation successful");
+
+    #[cfg(not(feature = "persistence"))]
+    let _ = secret;
+    #[cfg(feature = "persistence")]
+    if let Some(persistence_bootstrap_info) = persistence_bootstrap_info {
```
stefano-garzarella, line 495: Maybe the `secret` can be the result of run_attestation() where you can return also the error, so we can have a single `log::err!` in the caller.

I mean, I'd move persistence related calls out of this function.

luigix25:
@stefano-garzarella
- Not panicking on any attestation error
- Secret is now the return value of the `run_attestation` function

Documentation/docs/developer/ATTESTATION.md
stefano-garzarella, line 1: Please move this change in a standalone commit.

kernel/src/svsm.rs
```diff
@@ -548,22 +562,21 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
...
+            .ok();
 
         #[cfg(not(feature = "persistence"))]
         let _ = secret;
         #[cfg(feature = "persistence")]
-        if let Some(persistence_bootstrap_info) = persistence_bootstrap_info {
+        if let (Some(persistence_bootstrap_info), Some(secret)) =
+            (persistence_bootstrap_info, secret)
+        {
             persistence_init(persistence_bootstrap_info, &secret).unwrap();
```
stefano-garzarella: mmm, this is not really readable IMO.

What about something like this? (clippy-only tested)
```diff
diff --git a/kernel/src/svsm.rs b/kernel/src/svsm.rs
index 6cc17bf3..b70afeaa 100755
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -567,17 +567,16 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
         #[cfg(feature = "persistence")]
         let persistence_bootstrap_info = persistence_discover().unwrap();
 
-        let secret = run_attestation()
-            .inspect_err(|e| log::error!("attestation failed: {e:?}"))
-            .ok();
-
-        #[cfg(not(feature = "persistence"))]
-        let _ = secret;
-        #[cfg(feature = "persistence")]
-        if let (Some(persistence_bootstrap_info), Some(secret)) =
-            (persistence_bootstrap_info, secret)
-        {
-            persistence_init(persistence_bootstrap_info, &secret).unwrap();
+        match run_attestation() {
+            Ok(secret) => {
+                #[cfg(feature = "persistence")]
+                if let Some(persistence_bootstrap_info) = persistence_bootstrap_info {
+                    persistence_init(persistence_bootstrap_info, &secret).unwrap();
+                }
+            }
+            Err(e) => {
+                log::error!("attestation failed: {e:?}");
+            }
         }
     }
```

I put the final solution, but maybe you can do it since the first commit.

stefano-garzarella (changes requested):

luigix25:
- Moved doc changes to a different commit
- Applied suggested style changes from @stefano-garzarella

Documentation/docs/developer/ATTESTATION.md
stefano-garzarella, line 15: This needs to be updated.

(not sure why we have this in this file, maybe we can remove, but of course not in this PR)

luigix25: oops, done!

luigix25:
- Added Section to the ToC

stefano-garzarella (approved):

PR #1207, opened 2026-08-26
Monotonic timer API and Secure TSC support

vsntk18:
This PR aims to add monotonic timer APIs to provide timekeeping services in Coconut-SVSM.

Time abstraction can be helpful for implementing features like time based preemptions or
even provide time based services for user space.

This PR provides structures like Instant and Deadline that can be used to implement
time based services.

Monotonic timer is backed by SecureTSC. SecureTSC is implemented for SEV-SNP platform in
this PR.

In addition, this PR also adds some tests for basic functionality.


vsntk18:
About the unit of Frequency of Secure TSC read from GUEST_TSC_FREQ MSR:
https://docs.amd.com/v/u/en-US/40332_4.10_APM_Vol1-5_PUB, section 15.36.18 says:

> Guests that run with Secure TSC enabled may read the GUEST_TSC_FREQ MSR (C001_0134h)
> which returns the effective frequency in MHz of the guest view of TSC

kernel/src/sev/secure_tsc.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        let raw = read_msr(MSR_GUEST_TSC_FREQ);
+        if raw != 0 {
+            let factor = self.tsc_factor.load(Ordering::Relaxed) as u128;
+            let adjustment = ((raw as u128) * factor) / 100_000;
+            let freq = raw.saturating_sub(adjustment as u64);
+            self.freq_raw.store(freq, Ordering::Relaxed);
+            self.freq_valid.store(true, Ordering::Release);
+            freq
```
joergroedel: As we discussed previously, `MSR_GUEST_TSC_FREQ` return frequency in Mhz, no Hz. So for the correct frequency it needs to be multiplied by `1.000.000`.

kernel/src/svsm.rs
```diff
@@ -497,6 +498,10 @@ fn svsm_init(launch_info: &KernelLaunchInfo) {
         .env_setup_svsm()
         .expect("SVSM platform environment setup failed");
 
+    if let Err(e) = SVSM_PLATFORM.configure_secure_tsc(launch_info.use_secure_tsc) {
+        log::warn!("Secure TSC config failed: {e:#?}");
+    }
```
joergroedel, line 503: This needs to be dependent on the SNP platform.

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,527 @@
...
+        log::info!("SecureTSC not enabled, monotonic clock not initialized");
+        return Ok(());
+    }
+
+    // Get the TSC frequency
+    let freq_hz = SECURE_TSC_ACCESSOR.read_tsc_frequency();
+    if freq_hz == 0 {
+        log::warn!("SecureTSC frequency is 0, monotonic clock not initialized");
+        return Ok(());
+    }
```
joergroedel, line 442: This should return a failure, and the caller can then decide to continue and print a message that no clocksource could be initialized.

kernel/src/platform/snp.rs
```diff
@@ -403,6 +409,38 @@ impl SvsmPlatform for SnpPlatform {
...
+        }
+
+        SECURE_TSC_ACCESSOR.set_tsc_info(secure_tsc_info);
+        SECURE_TSC_ACCESSOR.set_use_secure_tsc(true);
+
+        let base = SECURE_TSC_ACCESSOR.read_tsc_frequency();
+        if base == 0 {
+            SECURE_TSC_ACCESSOR.set_use_secure_tsc(false);
+            return Err(SvsmError::NotSupported);
+        }
```
joergroedel, line 439: When the `SecureTSC` bit is set in `SEV_FEATURES`, a failure to initialize the feature should be treated as fatal.

joergroedel (commented):
Thanks for these changes. I left a few comments.

kernel/src/svsm.rs
vsntk18: Actually there is a generic implementation for configure_secure_tsc() is provided in SvsmPlatform trait. And, SNP platform overrides this.

kernel/src/sev/secure_tsc.rs
vsntk18: ok.

kernel/src/time/monotonic.rs
vsntk18: done

kernel/src/platform/snp.rs
vsntk18: ok.

00xc:
Hi @vsntk18 could you update the PR description to cover the what, how and why of the changes? It adds 900+ lines of code, so I think we need a proper cover letter

vsntk18:
> Hi @vsntk18 could you update the PR description to cover the what, how and why of the changes? It adds 900+ lines of code, so I think we need a proper cover letter

Ok, will update the description.

tools/igvmbuilder/src/igvm_builder.rs
```diff
@@ -270,13 +270,15 @@ impl IgvmBuilder {
             debug_serial_port: self.options.get_port_address(),
             firmware: fw_info,
             vmsa_in_kernel_range: self.gpa_map.vmsa_in_kernel_range as u8,
+            use_secure_tsc: 1,
```
msft-jlange, line 273: This needs to be an optional configuration, since not all deployments will want the burden of secure TSC setup.

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,532 @@
...
+/// `MONOTONIC_CLOCK.elapsed_since_boot()` for time since boot.
+pub static MONOTONIC_CLOCK: MonotonicClock = MonotonicClock::new();
+
+/// Monotonic clock accessor.
+///
+/// Provides methods to query the current time and compute durations.
+/// The clock uses SecureTSC as the underlying time source.
+#[derive(Debug)]
+pub struct MonotonicClock {
+    _private: (),
```
msft-jlange, line 80: Why is this needed?

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,532 @@
...
+
+    #[inline]
+    fn ticks_to_duration(&self, ticks: u64) -> Duration {
+        let nanos = ((ticks as u128) * (self.nanos_per_tick_scaled as u128)) >> 32;
+        Duration::from_nanos(u64::try_from(nanos).unwrap_or(u64::MAX))
+    }
+}
+
+/// Global clock parameters, initialized once at boot.
+static CLOCK_PARAMS: ImmutAfterInitCell<ClockParams> = ImmutAfterInitCell::uninit();
```
msft-jlange, line 65: Why is this a separate object from the `MONOTONIC_CLOCK` global?  Couldn't the clock parameters simply be embedded in the global clock, since they are both singleton objects?

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,532 @@
...
+
+impl Default for MonotonicClock {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// A monotonic point in time.
+///
+/// An `Instant` is measured in nanoseconds since the monotonic clock's boot-time
```
msft-jlange, line 169: It's difficult for me to see the value in `Instant` as distinct from a simple `u64`.  It seems that `Instant` is simply a count of nanoseconds (which is always guaranteed to fit in a `u64`) with a couple of utility functions that could just as easily be implemented on `MonotonicClock` instead of being properties of the `Instant`.  Why invent a separate unique type?

tools/igvmbuilder/src/context.rs
```diff
@@ -342,6 +342,7 @@ pub fn construct_vmsa(
     features.set_snp(true);
     features.set_restrict_injection(true);
     features.set_debug_swap(true);
+    features.set_secure_tsc(true);
```
msft-jlange, line 345: This needs to be an optional configuration, since not all deployments will want the burden of secure TSC setup.

tools/igvmbuilder/src/igvm_builder.rs
vsntk18: done!

kernel/src/time/monotonic.rs
vsntk18: Not needed, removed.

kernel/src/time/monotonic.rs
vsntk18: That sounds better, yes. Moved this inside MonotonicClock struct.

kernel/src/time/monotonic.rs
vsntk18: I think the idea is to provide an equivalent of std::time::Instant in coconut-svsm: https://doc.rust-lang.org/std/time/struct.Instant.html.  So this Instant type provides the same methods/functionality as std::time::Instant.

tools/igvmbuilder/src/context.rs
vsntk18: done!

vsntk18:
Addressed review comments from Joerg and Jon.

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub const ZERO: Self = Self {
+        nanos_since_boot: 0,
+    };
+
+    /// Create an instant from a raw nanosecond value since boot.
+    ///
+    /// This is intended for code that persists or exchanges monotonic deadlines
+    /// in their raw representation.
+    #[inline]
+    pub const fn from_nanos(nanos: u64) -> Self {
```
msft-jlange, line 192: If this is intended to be an implementation of the type `std::time::Instant`, why include all of these additional methods that are not part of `Instant`?  There is nothing in the standard `Instant` definition that has any relation to boot time or any other epoch - in fact, `Instant` seems very clear that it is deliberately designed to avoid references to any absolute time with the exception of `Instant::now()`.  It is up to `Duration` to provide the relationship between different points in time.

If we believe that an implementation equivalent to `std::time::Instant` is valuable, then we should constrain the implementation to only those methods for the sake of consistency with the standard library.  Otherwise we should define a different type to encapsulate the different properties.

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,536 @@
...
+        Self {
+            nanos_since_boot: nanos,
+        }
+    }
+
+    /// Return the raw nanosecond value since boot.
+    #[inline]
```
msft-jlange, line 199: Why are these marked for explicit inlining?  What's wrong with letting the compiler make the decision based on its analysis of what code is best?

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub fn elapsed(&self) -> Duration {
+        MONOTONIC_CLOCK.now().duration_since(*self)
+    }
+
+    /// Try to calculate the duration elapsed since this instant.
+    ///
+    /// Returns `None` if the clock has not been initialized.
+    #[inline]
+    pub fn try_elapsed(&self) -> Option<Duration> {
```
msft-jlange, line 229: I expect consumers of `Instant` to rely on correct timing in all cases.  Wouldn't it be better to implement `elapsed()` instead of `try_elapsed()`, where `elapsed()` returns time based on TSC, monotonic or not?  It should be up to the caller to make a separate determination of whether `Instant` represents a secure monotonic timer or some less trustworthy timer.

kernel/src/time/monotonic.rs
```diff
@@ -0,0 +1,536 @@
...
+///
+/// An `Instant` is measured in nanoseconds since the monotonic clock's boot-time
+/// epoch. The backing clock source is deliberately hidden from callers so timer
+/// users can work with deadlines without depending on TSC details.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Instant {
+    nanos_since_boot: u64,
+}
+
+impl Instant {
```
msft-jlange, line 181: Why no implementation for `Instant::now()`?

kernel/src/cpu/vmsa.rs
```diff
@@ -174,4 +183,5 @@ pub fn init_guest_vmsa(v: &mut VMSA, rip: u64, alternate_injection: bool) {
     }
 
     v.sev_features = sev_status.as_sev_features();
+    configure_secure_tsc(v);
```
msft-jlange, line 186: This seems problematic to me.  The guest OS most likely doesn't contain the logic required to manage secure TSC correctly - it certainly doesn't know how to query for TSC_INFO and set up AP VMSAs with the correct TSC base - and if we force secure TSC on the guest BSP, I expect lots of things will break.  I would strongly recommend limiting this PR to TSC functionality within the SVSM, and leaving guest support for secure TSC in a future PR.

kernel/src/greq/pld_tsc_info.rs
```diff
@@ -0,0 +1,90 @@
...
+use core::mem::{offset_of, size_of};
+
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+use crate::error::SvsmError;
+
+const TSC_INFO_REQ_SIZE: usize = 128;
+
+/// MSG_TSC_INFO_REQ payload format.
+#[repr(C, packed)]
```
msft-jlange, line 18: This doesn't need to be `packed` if it's just a `u8` slice because that's naturally packed.  Marking a structure as `packed` can cause compilation challenges and it's better to avoid it unless natural alignment creates compatibility issues.

kernel/src/greq/pld_tsc_info.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl Default for SnpTscInfoRequest {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// MSG_TSC_INFO_RSP payload format.
+#[repr(C, packed)]
```
msft-jlange, line 40: This also doesn't need to be `packed` because it contains explicit padding, and therefore the natural alignment produces the correct layout.

tools/igvmbuilder/src/igvm_builder.rs
```diff
@@ -270,13 +272,15 @@ impl IgvmBuilder {
             debug_serial_port: self.options.get_port_address(),
             firmware: fw_info,
             vmsa_in_kernel_range: self.gpa_map.vmsa_in_kernel_range as u8,
+            use_secure_tsc,
```
msft-jlange, line 275:
```suggestion
            use_secure_tsc: self.options.secure_tsc.into(),
```

You can let the compiler convert the `bool` to the correct integer type, and then you can get rid of the explicit conversion above.

PR #1210, opened 2026-09-04
Support cocoon-tpm's OpenSSL backend

osteffenrh:
Add support for cocoon-tpm's OpenSSL backend (cocoon-tpm PR: https://github.com/coconut-svsm/cocoon-tpm/pull/30).

cocoon-tpm-crypto is the designated crypto library for SVSM. It supports
multiple backends: pure Rust, BoringSSL, and — with the above PR — OpenSSL.
This PR integrates the new OpenSSL backend on the SVSM side.

Previously, the vTPM code (libtcgtpm) maintained its own OpenSSL build —
including a git submodule, build config, and a bundled copy of libcrt (a
minimal C runtime). With cocoon-tpm now providing an OpenSSL backend, this
duplication is eliminated: a single libcrypto.a is built by cocoon-tpm's
ossl-bare-sys crate and shared by both cocoon-tpm-crypto and libtcgtpm.
The OpenSSL submodule and build infrastructure are removed from libtcgtpm.

The OpenSSL backend is selected via the cargo feature openssl. The vtpm
feature now implies openssl, since the TCG TPM reference code requires
OpenSSL at the C level. Without vtpm, the pure-Rust backend remains the
default.

To build OpenSSL for the freestanding SVSM environment, cocoon-tpm's
ossl-bare-sys uses a target-integration pattern: the upstream crate depends
on a stub crate (cocoon-tpm-ossl-bare-sys-target-integration) that is
overridden via [patch.crates-io] in each target environment. The SVSM
implementation of this crate provides the OpenSSL build configuration
(Configure flags and target config file) and depends on our libcrt crate for
the C runtime headers needed during the OpenSSL build.

libcrt itself is extracted from libtcgtpm into a standalone crate
(libcrt). It has no Rust code — it just builds the C runtime and provides
it at link time. Both the target-integration crate and libtcgtpm depend on
it. Include paths and library locations are communicated between crates via
cargo metadata (`DEP_CRT_*`, `DEP_OSSL_*` environment variables).

This PR does not consolidate all of SVSM's crypto usage under
cocoon-tpm-crypto; that is left as follow-up work.

The vtpm and openssl features do not work for host-side targets like tests
and doc tests, since they pull in the target-integration crate for OpenSSL
which is specific to the bare-metal environment of SVSM.

## Commit Outline

The first two commits refactor the random number generator pool in the crypto crate,
making it more generic and easier to add other RNG implementations later.
- kernel: utils: add static array backed object pool
- kernel: crypto: refactor RNG pool

The following commit is only temporary and shall be dropped before this PR is merged,
which requires that the cocoon-tpm PR is merged and a new release is published on crates.io first.
- workspace: override cocoon-tpm source to use osteffen's fork (drop)

Debug improvement, possible with the cocoon-tpm PR.
- kernel: print cocoon-tpm crypto backend name

libcrt refactoring: Pull the C runtime out from libtcgtpm into its own
top-level crate, allowing multiple consumers. It does not contain any Rust code,
just builds the C code and provides a static library to the linker. Include path
and library location are communicated between build.rs scripts via environment variables.
- workspace: add libcrt from libtcgtpm as its own crate
- libcrt: add missing visibility pop to match push in libcrt.h
- libtcgtpm: use libcrt from dedicated crate

The center piece of this PR: Add the target integration library for cocoon-tpm OpenSSL backend.
Provides OpenSSL build config and libcrt (by depending on the libcrt crate).
Bisect-ability between these three commits is broken: The RNG implementation is needed to build with
the openssl feature enabled (fixed by the second one) and the openssl feature conflicts with the vtpm
feature due to duplicate linkage against libcrypto.a instances. This is resolved by the last commit that
drops the openssl copy from libtcgtpm and uses the cocoon-tpm provided one instead.
- workspace: add cocoon-tpm OpenSSL backend integration lib
- kernel: crypto: add RNG implementation using OpenSSL
- libtcgtpm: use libcrypto from cocoon-tpm-ossl-bare-sys


osteffenrh:
cc: @nicstange

osteffenrh:
cocoon-tpm-ossl-bare-sys-target-integration: Change `edition` from 2021 to 2024 as done on the cocoon-tpm side (see https://github.com/coconut-svsm/cocoon-tpm/pull/30#issuecomment-5758049142)

kernel/src/utils/pool.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+impl<T: Send, const N: usize> Drop for PoolGuard<'_, T, N> {
+    fn drop(&mut self) {
+        if let Some(obj) = self.data.take() {
+            self.pool.put(obj);
+        }
+    }
+}
```
00xc, line 99: You can avoid using an `Option` to store the `T` by using `ManuallyDrop`, e.g.

```rust
pub struct PoolGuard<'a, T: Send, const N: usize> {
    pool: &'a Pool<T, N>,
    data: ManuallyDrop<T>,
}
```

Then `drop()` just becomes:

```suggestion
impl<T: Send, const N: usize> Drop for PoolGuard<'_, T, N> {
    fn drop(&mut self) {
        // SAFETY: `self.data` cannot be used again, since this is
        // the drop() implementation.
        let obj = unsafe { ManuallyDrop::take(&mut self.data) };
        self.pool.put(obj);
    }
}
```

This avoids the extra storage for the `Option` discriminant and the unnecessary unwraps.

osteffenrh: Done.

osteffenrh:
- Applied suggestion to use `ManuallyDrop` for the object pool (@00xc)
- Amended commits with `Cargo.lock` changes
- Rebased to latest main

PR #1211, opened 2026-09-04
kernel/svsm: log CPU vendor after buffer logger initialization

luigix25:
The CPU vendor log was being called before install_buffer_logger() was executed, causing the message to be lost. Move the log statement to after the logger is installed so CPU vendor information appears in test output.

### DIFF
--- a/kernel/src/svsm.rs
+++ b/kernel/src/svsm.rs
@@ -340,7 +340,6 @@ unsafe fn svsm_start(
     platform
         .env_setup(debug_serial_port, launch_info.vtom.try_into().unwrap())
         .expect("Early environment setup failed");
-    log::info!("CPU vendor {:?}", platform.get_cpu_vendor());
 
     // Load symbol info now that there is a console
     // SAFETY: the launch info here is the launch info passed by the boot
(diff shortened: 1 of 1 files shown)


luigix25:
cc @n-ramacciotti

n-ramacciotti (approved):

kernel/src/svsm.rs
```diff
@@ -410,6 +414,7 @@ unsafe fn svsm_start(
     // has been already created and console is
     // initialized in env_setup_late() before this
     install_buffer_logger();
+    log::info!("CPU vendor {:?}", platform.get_cpu_vendor());
```
00xc:
```suggestion
    log::info!("CPU vendor: {:?}", platform.get_cpu_vendor());
```

00xc (dismissed):

kernel/src/svsm.rs
luigix25: done, thanks

00xc (approved):

stefano-garzarella (approved):

PR #1212, opened 2026-09-04
cargo: fix some clippy warnings

luigix25:
running `make bin/coconut-test-qemu.igvm` with latest `rustc`/`clippy` version, returns several errors.
This PR fixes most of them.


test/Cargo.toml
```diff
@@ -3,4 +3,7 @@ name = "test"
 version = "0.1.0"
```
osteffenrh, line 3: Commit message: Can you mention what the change is first, instead of the log error message, putting this "Apply workspace lint rules to the test and verus_stubs crates." as the first line?

verification/verus_macro_stub/src/lib.rs
```diff
@@ -4,7 +4,6 @@
 //
 // Author: Luigi Leonardi <leonardi@redhat.com>
```
osteffenrh, line 5: Commit message: How about "remove unused extern crate" in the first line?

test/Cargo.toml
stefano-garzarella: IMHO it is better to explain the problem first, then what the commit is doing to fix it.

This is something also suggested in the Linux's doc: https://www.kernel.org/doc/html/latest/process/submitting-patches.html

> Once the problem is established, describe what you are actually doing about it in technical detail. It’s important to describe the change in plain English for the reviewer to verify that the code is behaving as you intend it to.

(Obviously, I'm not saying we should follow that documentation strictly, but I use it as a sort of guideline for writing good commit messages)

verification/verus_macro_stub/src/lib.rs
stefano-garzarella: It's already in the title, IMHO it's enough

test/Cargo.toml
stefano-garzarella: About the commit title, we never used `global: ` in the title. In this cases I usually don't put any prefix, so here maybe you can do `Inherit workspace lints in test and verus crates`. Or in other projects I saw using `chore: ` for something affecting the whole workspace.

luigix25: What about workspace:?

stefano-garzarella: > What about workspace:?

yep, anyway, if you prefer global, you can use it, it was just an observation that I never saw that used.

luigix25:
- Fixed some more warnings
- Replaced `global` with `cargo`

stefano-garzarella:
Which clippy/toolchain version starts to complain about those? (would be nice to have it specified in commits and PR description)

luigix25:
> Which clippy/toolchain version starts to complain about those? (would be nice to have it specified in commits and PR description)

good question, i don't know. I noticed it in the CI, and then reproduced it locally by running `rustup update`.

Since the toolchain is pinned by us, I think this is due to a newer clippy version, but I don't really know which version introduced these warnings.

00xc (commented):
The PR looks fine to me, and these are cleanups that we would have likely needed to do at some point anyway. But I'd also like to point out that fixing warnings that only appear on nightly (which is we use for `bin/coconut-test-qemu.igvm`) is typically a game of cat and mouse. In fact there are other nightly warnings remaining that are not fixable (because we use certain APIs that are deprecated in nightly, but their newer alternative is not stable in our pinned stable version).

user/uapi_tester/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition.workspace = true
 
 [[bin]]
-name = "uapi_tester"
+name = "uapi-tester"
```
stefano-garzarella, line 7: I know that clippy now is happy, but should we rename also the crate in `uapi-tester` ?

luigix25: you mean rename the folder?

stefano-garzarella:
```
Cargo.lock:name = "uapi_tester"
Cargo.toml:    "user/uapi_tester",
Cargo.toml:uapi_tester = { path = "user/uapi_tester" }
Makefile:       ${CARGO} clippy ${CLIPPY_OPTIONS} --workspace $(BARE_METAL_EXCLUDES) --exclude svsm-fuzz --exclude uapi_tester -- ${CLIPPY_ARGS}
configs/test/hyperv-test-target.json:            "uapi_tester": {
configs/test/hyperv-test-target.json:                "path": "/test/uapi_tester",
configs/test/qemu-test-target.json:            "uapi_tester": {
configs/test/qemu-test-target.json:                "path": "/test/uapi_tester",
configs/test/vanadium-test-target.json:            "uapi_tester": {
configs/test/vanadium-test-target.json:                "path": "/test/uapi_tester",
user/uapi_tester/Cargo.toml:name = "uapi_tester"
user/uapi_tester/src/lib.rs:pub mod uapi_tester;
```

stefano-garzarella: But I'm not sure what it's common in rust (I mean using - or _ in crates names).
My concern was that the binary is now `uapi-tester` but we still use `uapi_tester` in the configurations, which IMO is a bit confusing.

luigix25: right good point, I'm thinking about dropping this patch for now.

luigix25:
- Dropped kebab-name case for binaries

00xc (approved):

stefano-garzarella (approved):

PR #1213, opened 2026-09-07
kernel: cpu/features: allow arbitrary feature presence detection

00xc:
Currently, `cpu_has_feat()` returns a boolean indicating if the provided CPU feature is present (i.e. if it can be used in the current CPU). However, the general case for detecting if a feature is present is to compare its CPUID value to 1, which only works for features that have a bit-width of 1. This makes it really easy to misuse `cpu_has_feat()`.

Instead, rework the `CpuFeat` struct to hold a pointer to a function that can be used to determine if a feature is enabled. The general case will return true for any non-zero value, which is more robust than comparing to 1. Add a method for the special case where we want to compare to a different value.

Note that `InvlpgbMax` is special, as a zero is a valid value, so its detection function always returns true.

### DIFF
--- a/kernel/src/cpu/features.rs
+++ b/kernel/src/cpu/features.rs
@@ -59,24 +59,24 @@ struct CpuFeat {
     val: ImmutAfterInitCell<u32>,
     /// Location of the feature within a CPUID leaf
     feature: CpuidFeature,
-    /// Expected value after shift + mask
-    expected: u32,
+    /// Function to detect if feature is present, after shift + mask
+    present: fn(u32) -> bool,
 }
 
 impl CpuFeat {
-    /// Create a new feature, indicated by a single bit (expected value 1).
-    const fn new_bit(feature: CpuidFeature) -> Self {
-        Self::new(feature, 1)
-    }
-
-    /// Create a new CPU feature from a cpufeature descriptor, comparing the
-    /// extracted field against the given expected value.
-    const fn new(feature: CpuidFeature, expected: u32) -> Self {
+    /// Create a new feature, considered to be present if its value is non-zero.
+    const fn new(feature: CpuidFeature) -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


nagyation:
I did some digging (best effort) to understand more about the problem, and I think this PR highlights a legitimate potential problem, but I don't think it ultimately fixes it. InvlpgbMax, Cbit & PhysAddrSizes are still not properly checked for presence.

If the main concern is to avoid silently failing later with features of bit-width > 1, then we could add a compile-time check to prevent that case from happening. Otherwise we need find a solution that properly works with the whole feature list.
  
InvlpgbMax and the others could still have a correct cpu_has_feat by checking Invlpgb itself instead of InvlpgbMax. This could technically be done by adding them to the CpuFeat list, and then having the presence function check the value of the other one, something like:

``` rust
    Invlpgb => CpuFeat::new(INVLPGB), 
    InvlpgbMax => CpuFeat::new(INVLPGB_MAX_PAGES)
          .with_present(|_| cpu_has_feat(Feature::Invlpgb)),
```

Other solution comes to my mind is to split the ValueFeatures from the Features list.

kernel/src/cpu/features.rs
```diff
@@ -59,26 +59,26 @@ struct CpuFeat {
     val: ImmutAfterInitCell<u32>,
     /// Location of the feature within a CPUID leaf
     feature: CpuidFeature,
-    /// Expected value after shift + mask
-    expected: u32,
+    /// Function to detect if feature is present, after shift + mask
+    present: fn(u32) -> bool,
```
nagyation, line 63: I'd rather change the function pointer to take &Self instead of the extracted value, to give the user more flexibility instead of pre-reading the value if it's not going to be used.

`present: fn(&CpuFeat) -> bool,`

PR #1215, opened 2026-09-10
kernel/mm: Add context-specific virtual memory ranges

joergroedel:
The existing per-task kernel mapping scheme divides the available virtual address space into fixed-size regions. This imposes a fixed task limit and reserves the same amount of virtual address space regardless of each task's actual needs.

Replace that scheme with dynamically allocated, context-specific virtual memory ranges. The shared `UniqueVaAllocator` assigns globally unique virtual addresses. Each `TaskMM` owns a `ContextVMR` for kernel-task mappings, which remain visible only in the associated `TaskMM` context.

This series:

- Adds a page-granular unique virtual-address allocator (`UniqueVaAllocator`) with aligned, hinted, and fixed-address allocation.
- Adds `ContextVMR` and mapping guards, including automatic address release and page-table cleanup.
- Migrates task kernel stacks, shadow stacks, and kernel mmap requests to ContextVMR.
- Removes the fixed per-task virtual-region allocator and its reserved guard mappings.
- Makes page-fault handling usable by both regular and context-specific virtual memory ranges.

### Update on v2:

- Used the new `UniqueVaAllocator` for `struct VMR` as well. This needed some functional enhancements to the new VA allocator.
- `ContextVMRAllocator` and `ContextVMR` now also support shared mappings. This is not used in this PR, but will be needed for future per-CPU related PRs.


00xc (commented):
I don't particularly like that we now have 2 VMRs with quite a bit of duplicate logic (including duplicate mmap methods, duplicate guard types, etc.). I think we should keep the VA allocation and mapping management separate like we had before.

Can't we just replace `KTASK_VADDR_BITMAP`'s bitmap allocator with the new `UniqueVaAllocator`? That would solve the granularity problem.

joergroedel:
> I don't particularly like that we now have 2 VMRs with quite a bit of duplicate logic (including duplicate mmap methods, duplicate guard types, etc.). I think we should keep the VA allocation and mapping management separate like we had before.
> 
> Can't we just replace `KTASK_VADDR_BITMAP`'s bitmap allocator with the new `UniqueVaAllocator`? That would solve the granularity problem.

I agree that having a new VMR type is not ideal, but merging the functionality at this point would be a premature optimization. The two VMR types serve two very different purposes and will evolve in different directions.

The `struct VMR` will be enhanced to support POSIX-like `mmap` semantics, something that is unnecessary for `struct ContextVMR`. The `ContextVMR` shares a single address allocator between multiple instances, `VMR` has only one built-in.

One of the next steps for `ContextVMR` is to enhance it support the PerCPU portion of the address space, which again adds functionality not needed for `VMR`.

After all these changes are implemented and once the dust has settled we can re-evaluate what parts of both implementations can be merged together.

00xc:
> The `ContextVMR` shares a single address allocator between multiple instances, `VMR` has only one built-in.

Right, but what I am suggesting is that address allocation should be decoupled from the `VMR` (like `KTASK_VADDR_BITMAP`). Then we can use the `VMR` implementation both for globally visible ranges and per-CPU ones.

joergroedel:
> Right, but what I am suggesting is that address allocation should be decoupled from the `VMR` (like `KTASK_VADDR_BITMAP`). Then we can use the `VMR` implementation both for globally visible ranges and per-CPU ones.

Okay, this turned out better than I anticipated. Updated the PR to use the new `UniqueVaAllocator` for `VMR` and `ContextVMR`.

00xc:
> > Right, but what I am suggesting is that address allocation should be decoupled from the `VMR` (like `KTASK_VADDR_BITMAP`). Then we can use the `VMR` implementation both for globally visible ranges and per-CPU ones.
> 
> Okay, this turned out better than I anticipated. Updated the PR to use the new `UniqueVaAllocator` for `VMR` and `ContextVMR`.

Right. I see we still have two `VMRs`, I wonder if we can abstract the allocation strategy away to support different types of semantics. Maybe this can be done on top of #1217, e.g. something like

```rust
pub struct VMR<V: VirtualRange> {
    ...
}

impl<V: VirtualRange> VMR<V> {
    pub fn insert_at(&self, addr: VirtAddr, mapping: Mapping) -> Result<VirtAddr, SvsmError> {
        let addr = V::get_allocator().alloc_at(addr, mapping.mapping_size())?;
        self.insert_allocated(addr, mapping)
    }

    /// Inserts a mapping at or above `hint` with the requested alignment.
    pub fn insert_aligned(
        &self,
        hint: VirtAddr,
        mapping: Mapping,
        align: usize,
    ) -> Result<VirtAddr, SvsmError> {
        let addr = V::get_allocator()
            .alloc_aligned(hint, mapping.mapping_size(), align)
            .ok_or(SvsmError::Mem)?;
        self.insert_allocated(addr, mapping)
    }
```

That would require making `VirtualRange` support different allocators, not just `VRangeAllocator`, but that should be an easy change.

00xc:
> Maybe this can be done on top of #1217, e.g. something like
> 
> ```rust
> pub struct VMR<V: VirtualRange> {
>     ...
> }
> ```

Hm, this will not work for the per-CPU ranges, since there are different allocation sizes per sub-range, at least with my current design.

I still think we can keep the address allocation decoupled from `VMR` (as it is in my PR), and just allow different allocators by making `VirtualRange` more generic, as I suggested.

```rust
pub trait VirtualRange {
    // ...

    type Allocator: VirtualAllocator;

    fn get_allocator() -> impl core::ops::Deref<Target = Self::Allocator>;
}

pub trait VirtualAllocator {
    fn alloc(...)
    fn alloc_exact(...)
}
```

Then it's just a matter of using an allocator with the right semantics for each virtual range (e.g. your `ContextVMRAllocator` for the per-task range).

joergroedel:
> > > Right, but what I am suggesting is that address allocation should be decoupled from the `VMR` (like `KTASK_VADDR_BITMAP`). Then we can use the `VMR` implementation both for globally visible ranges and per-CPU ones.
> > 
> > 
> > Okay, this turned out better than I anticipated. Updated the PR to use the new `UniqueVaAllocator` for `VMR` and `ContextVMR`.
> 
> Right. I see we still have two `VMRs`, I wonder if we can abstract the allocation strategy away to support different types of semantics. Maybe this can be done on top of #1217, e.g. something like
> 
> ```rust
> pub struct VMR<V: VirtualRange> {
>     ...
> }
> 
> impl<V: VirtualRange> VMR<V> {
>     pub fn insert_at(&self, addr: VirtAddr, mapping: Mapping) -> Result<VirtAddr, SvsmError> {
>         let addr = V::get_allocator().alloc_at(addr, mapping.mapping_size())?;
>         self.insert_allocated(addr, mapping)
>     }
> 
>     /// Inserts a mapping at or above `hint` with the requested alignment.
>     pub fn insert_aligned(
>         &self,
>         hint: VirtAddr,
>         mapping: Mapping,
>         align: usize,
>     ) -> Result<VirtAddr, SvsmError> {
>         let addr = V::get_allocator()
>             .alloc_aligned(hint, mapping.mapping_size(), align)
>             .ok_or(SvsmError::Mem)?;
>         self.insert_allocated(addr, mapping)
>     }
> ```
> 
> That would require making `VirtualRange` support different allocators, not just `VRangeAllocator`, but that should be an easy change.

I am all for making the code more generic and avoid duplication, but in this case there are a few tradeoffs to think about:

1. The approach above requires to keep track of allocated addresses at two places. One is the allocator behind `V` and the other is behind the implementation of `insert_allocated()` in the `VMR` itself. This is needed for the `ContextVMR` because it needs to keep track of the global address space and the local mappings, but unnecessary overhead for `VMR`.
2. The implementation of `VMR` is going to be a highly performance sensitive piece of the COCONUT kernel. The `ContextVMR` not so much. For that reason there is value in logic duplication when it allows better optimizations on the `VMR` side.

I am not sure what the right tradeoff here is, I will put this PR on the list for tomorrow.

00xc:
> I am all for making the code more generic and avoid duplication, but in this case there are a few tradeoffs to think about:
> 
> 1. The approach above requires to keep track of allocated addresses at two places. One is the allocator behind `V` and the other is behind the implementation of `insert_allocated()` in the `VMR` itself. This is needed for the `ContextVMR` because it needs to keep track of the global address space and the local mappings, but unnecessary overhead for `VMR`.
>
> 2. The implementation of `VMR` is going to be a highly performance sensitive piece of the COCONUT kernel. The `ContextVMR` not so much. For that reason there is value in logic duplication when it allows better optimizations on the `VMR` side.

I am not sure if the optimization is worth the duplicate implementation. In any case, this is a question of whether the address allocator can keep metadata associated with an allocation. I would need to look into it deeper, but we could perhaps we could parameterize on that in a single `VMR` implementation.

> I am not sure what the right tradeoff here is, I will put this PR on the list for tomorrow.

Sure, let's discuss it on the call.

PR #1216, opened 2026-09-10
kernel: hyperv: map code page into global shared range

00xc:
Currently, the Hyper-V hypercall code page is mapped into a fixed address, given by `SVSM_HYPERCALL_CODE_PAGE`, which falls within the shared portion of the address space. However, the global ranges abstraction already allows doing this without needing to statically reserve a particular virtual address for the page.

Convert the mapping of the code page to use a `GlobalRangeGuard`, removing the use of the low-level page table method `map_4k()`. While we are at it, allocate the physical page using a `PageBox`, and make sure it is never freed by storing it, along the `GlobalRangeGuard`, into `HYPERV_HYPERCALL_CODE_PAGE`.


kernel/src/hyperv/hv.rs
```diff
@@ -280,29 +279,51 @@ pub const HV_STATUS_SUCCESS: u16 = 0;
...
+    ///
+    /// Keep `_page` declared below `guard` in order to ensure a safe
+    /// drop behavior: the exec mapping should be torn down before the
+    /// physical page is returned to the page allocator.
+    _page: PageBox<[u8; PAGE_SIZE]>,
+}
+
+impl HypercallCodePage {
+    fn new() -> Result<Self, SvsmError> {
+        let page = PageBox::<[u8; PAGE_SIZE]>::try_new_zeroed()?;
```
msft-jlange: The page contents are totally irrelevant because the page is overwritten when it is mapped by the hypervisor.  For that reason, it would be better to allocate this with `try_new_uninit()` (and similarly to store the `PageBox` as a `MaybeUninit`) as this would skip the wasted zero-filling.

msft-jlange (approved):

stefano-garzarella:
> Convert the mapping of the code page to use a `GlobalMappingGuard`, removing the use of the low-level page table method `map_4k()`. While we are at it, allocate the physical page using a `PageBox`, and make sure it is never freed by storing it, along the `GlobalMapppingGuard`, into `HYPERV_HYPERCALL_CODE_PAGE`.

nit: I guess you meant `GlobalRangeGuard`, right? anyway, I don't want to block this for that, feel free to merge as is.

00xc:
> > Convert the mapping of the code page to use a `GlobalMappingGuard`, removing the use of the low-level page table method `map_4k()`. While we are at it, allocate the physical page using a `PageBox`, and make sure it is never freed by storing it, along the `GlobalMapppingGuard`, into `HYPERV_HYPERCALL_CODE_PAGE`.
> 
> nit: I guess you meant `GlobalRangeGuard`, right? anyway, I don't want to block this for that, feel free to merge as is.

Right, done!

stefano-garzarella (approved):

PR #1217, opened 2026-09-13
kernel: mm: bitmap-based virtual address allocation overhaul

00xc:
This PR reworks how virtual address subranges are described, and how portions of those ranges are allocated and freed.

These changes result in a strongly-typed description of how each sub-address space is managed, and removes some duplicated code.

## Background

Currently, in the SVSM codebase there are two ways in which we allocate virtual addresses:
1. Finding a suitable range of addresses in `VMR`'s RBTree. This is used e.g. to insert file mappings, and is more heavyweight.
2. Using a bitmap to allocate portions of a mapping already inserted in the `VMR`. The mapping in the `VMR` is not backed by any physical pages, and the bitmap is used to determine which virtual addresses may be used to install temporary mappings in the page table.

This PR reworks path number 2 above.

## Approach

The series makes the following changes, in order:
* ~Introduces `RawLockGuard::map()` (needed down the line)~
* Introduces a type to describe a portion of the virtual address space, `AddrSpaceDescriptor`.
* Renames `VirtualRange` -> `SubVmAllocator`.
* Renames `VRangeAlloc` -> `SubVmAlloc`.
* Introduces a new trait, `SubVmRange`, which associates a portion of the virtual address space (`AddrSpaceDescriptor`) with a particular allocator (`SubVmAllocator`).
* Makes `SubVmAlloc` generic over `SubVmRange`. This allows allocating from any `SubVmRange` (i.e. per-CPU or global, regular or huge-sized).
* Replaces duplicate code with the new `SubVmAlloc`.
* Makes the existing `SubVmAllocator` generic over `SubVmRange`, greatly simplifying its implementation.

## Design

The main idea is to have a strongly-typed description of each subregion of the virtual address, and what mechanism is used to reserve addresses in each.

At the core of the new design there are 4 elements:
* **`AddrSpaceDescriptor`**: a struct to describe range of virtual addresses (similar to `MemoryRegion<VirtAddr>`).
* **`SubVmAllocator`**: an address allocator. Wrapper around `BitmapAllocator1024` that knows from which `AddrSpaceDescriptor` it is allocating from.
* **`SubVmAlloc`**: an smart pointer for an allocation in a `SubVmAllocator`.
* **`SubVmRange`**: a new trait that is used to tie a portion of the virtual address space (`AddrSpaceDescriptor`) with an allocator (`SubVmAllocator`).

For example, the 4k temporary 4K per-CPU mappings are described as follows:

```rust
/// Descriptor for the virtual address subregion
pub const PERCPU_TEMP_4K: AddrSpaceDescriptor =
    AddrSpaceDescriptor::new(SVSM_PERCPU_TEMP_BASE, SIZE_LEVEL1);

pub struct PerCpu4kRange;

impl SubVmRange for PerCpu4kRange {
    /// Which virtual address subregion we allocate from
    const DESCRIPTOR: AddrSpaceDescriptor = PERCPU_TEMP_4K;

    /// The size of each allocation (one 4K page)
    const GRANULE: usize = PAGE_SIZE;

    /// An accessor into this subregion's allocator
    fn get_allocator() -> impl core::ops::DerefMut<Target = SubVmAllocator<Self>> {
        this_cpu().vrange_4k_mut()
    }
}
```

Both `SubVmAllocator` and `SubVmAlloc` are aware of which `SubVmRange` they refer to. For example, the old `SubVmAlloc` only supported per-CPU allocations, and needed separate methods to allocate from the 4K and 2M subregions:

```rust
impl SubVmAlloc {
    /// Returns a virtual memory region in the 4K virtual range.
    pub fn new_4k(size: usize, align: usize) -> Result<Self, SvsmError> {
        // Each bit in our bitmap represents a 4K page
        if (size & (PAGE_SIZE - 1)) != 0 {
            return Err(SvsmError::Mem);
        }
        let page_count = size >> PAGE_SHIFT;
        let addr = this_cpu().vrange_4k_mut().alloc(page_count, align)?;
        let region = MemoryRegion::new(addr, size);
        Ok(Self {
            region,
            huge: false,
        })
    }

    /// Returns a virtual memory region in the 2M virtual range.
    pub fn new_2m(size: usize, align: usize) -> Result<Self, SvsmError> {
        // ...
    }
```

The new version can allocate generically from any range (4K, 2M, per-CPU or global) with a single method. The caller only needs to specify the range as a generic parameter:

```rust
impl<A: SubVmRange> SubVmAlloc<A> {
    /// Returns a virtual memory region in the given virtual range.
    pub fn new(count: usize, align: usize) -> Result<Self, SvsmError> {
        let addr = A::get_allocator().alloc(count, align)?;
        let region = MemoryRegion::new(addr, count * A::GRANULE);
        Ok(Self {
            region,
            _phantom: PhantomData,
        })
    }
}
```


00xc:
Setting to draft until #1218 is merged

PR #1218, opened 2026-09-13
kernel: locking/spinlock: implement `RawLockGuard::map()`

00xc:
Add a method to map a component of the backing data into a new guard. Acquiring this guard can never fail, as it preserves the access permissions the previous guard had.

In order to do this, update `RawLockGuard` to store a `NonNull` into the data, instead of a regular mutable reference, in order to relax the borrow checker when implementing the method.


kernel/src/locking/spinlock.rs
```diff
@@ -32,11 +34,46 @@ use core::sync::atomic::{AtomicU64, Ordering};
...
+    }
+}
+
+// SAFETY: RawLockGuard does not automatically implement Sync because it
+// contains a `NonNull`, which is guarded by the lock's behavior.
+unsafe impl<T: Sync, I> Sync for RawLockGuard<'_, T, I> {}
+
+// SAFETY: RawLockGuard does not automatically implement Send because it
+// contains a `NonNull`, which is guarded by the lock's behavior.
+unsafe impl<T: Send, I> Send for RawLockGuard<'_, T, I> {}
```
joergroedel: `I` also needs to be the `Sync` or `Send` for the lock to be `Sync` or `Send`. Otherwise this allows migration of percpu guards across CPUs.

joergroedel (approved):

PR #1219, opened 2026-09-14
kernel/percpu: Introduce linker-backed GS-relative per-CPU keys

joergroedel:
## Summary

Introduce a linker-backed per-CPU key infrastructure and migrate CPU-local state out of the legacy `PerCpu` structure into the modules that own it.

Each CPU receives a private copy of the `.percpu` linker section. `%GS.base` contains the relocation delta between the linker template and that CPU’s mapped copy, allowing Rust and assembly code to access CPU-local values directly through `%gs` without depending on `this_cpu()`.

## Details

- Add `PerCpuKey<T>` and the `percpu!` declaration macro.
- Retain per-CPU header containing the mapped area’s self-pointer.
- Support statically initialized and runtime-initialized keys.
- Use atomic one-time initialization for runtime-created values.
- Support exported backing symbols for direct GS-relative assembly access.
- Provide host-side template access so the API can be unit tested without modifying host TLS state.
- Initialize and install each CPU’s per-CPU area during CPU bring-up.
- Preserve values that must be prepared by the boot CPU in `PerCpu` staging fields until the target CPU can initialize its own key area.
- Sequence GDT, TSS, and GS initialization so selector loading does not invalidate the GS base before early per-CPU accesses.

Most of the fields in `struct PerCpu` are migrated with this PR. The fields that remain need changes in the boot flow to be migrated safely. This will be done in a follow-on PR.

The changes do not touch the `PerCpuShared` structure yet. This will be subject of another follow-on PR.


kernel/src/cpu/percpu/key.rs
```diff
@@ -188,7 +195,17 @@ impl<T> PerCpuStorage<T> {
     #[doc(hidden)]
     pub const fn new(value: T) -> Self {
         Self {
-            value: UnsafeCell::new(value),
+            state: AtomicU8::new(PERCPU_INITIALIZED),
+            value: UnsafeCell::new(MaybeUninit::new(value)),
```
00xc, line 199: Is there a good reason to not use `ImmutAfterInitCell` here?

kernel/src/cpu/percpu/key.rs
```diff
@@ -0,0 +1,580 @@
...
+        unsafe extern "C" {
+            static percpu_start: u8;
+        }
+
+        let template_start = ptr::addr_of!(percpu_start) as usize;
+        let template_addr = ptr::from_ref(self.template) as usize;
+        let offset = template_addr
+            .checked_sub(template_start)
+            .expect("per-CPU variable is outside the .percpu section");
+        (percpu.percpu_area.base() + offset) as *const T
```
00xc, line 260: Maybe we should have a const-time assertion of that `PerCpuStorage::value` sits at offset 0. Or even better, the offset could be added to the final calculation.

kernel/src/cpu/shadow_stack.rs
```diff
@@ -23,6 +23,18 @@ pub const BUSY: usize = 1;
 pub static IS_CET_SUPPORTED: AtomicBool = AtomicBool::new(false);
 pub static IS_CET_ENABLED: AtomicBool = AtomicBool::new(false);
 
+percpu! {
+    static INIT_SHADOW_STACK: Option<VirtAddr>;
+}
+
+pub fn init_initial_shadow_stack(stack: Option<VirtAddr>) {
+    assert!(INIT_SHADOW_STACK.init(stack).is_ok());
```
00xc:
```suggestion
    INIT_SHADOW_STACK.init(stack).unwrap();
```

kernel/src/cpu/percpu/key.rs
```diff
@@ -0,0 +1,580 @@
...
+        for cpu in 0..cpu_count {
+            set_affinity(cpu);
+            CPU_LOCAL_VALUE.with(|value| {
+                assert_eq!(value.load(Ordering::Relaxed), 0);
+                value.store(cpu + 1, Ordering::Relaxed);
+            });
+
+            assert!(!RUNTIME_VALUE.is_initialized());
+            assert!(RUNTIME_VALUE.init(AtomicUsize::new(cpu + 10)).is_ok());
+            assert!(RUNTIME_VALUE.init(AtomicUsize::new(0)).is_err());
```
00xc:
```suggestion
            RUNTIME_VALUE.init(AtomicUsize::new(cpu + 10)).unwrap();
            RUNTIME_VALUE.init(AtomicUsize::new(0)).unwrap_err();
```

kernel/src/cpu/isst.rs
```diff
@@ -4,6 +4,38 @@
...
+use crate::cpu::tss::IST_DF;
+use crate::locking::RWLock;
+
+percpu! {
+    static ISST: RWLock<Isst>;
+    static DOUBLE_FAULT_SHADOW_STACK: Option<VirtAddr>;
+}
+
+pub fn init_double_fault_shadow_stack(stack: Option<VirtAddr>) {
+    assert!(DOUBLE_FAULT_SHADOW_STACK.init(stack).is_ok());
```
00xc:
```suggestion
    DOUBLE_FAULT_SHADOW_STACK.init(stack).unwrap();
```

kernel/src/cpu/isst.rs
```diff
@@ -4,6 +4,38 @@
...
+}
+
+pub fn double_fault_shadow_stack() -> Option<VirtAddr> {
+    DOUBLE_FAULT_SHADOW_STACK.with(|stack| *stack)
+}
+
+pub fn init_isst(double_fault_shadow_stack: Option<VirtAddr>) {
+    assert!(ISST.init(RWLock::new(Isst::default())).is_ok());
```
00xc:
```suggestion
    (ISST.init(RWLock::new(Isst::default())).unwrap();
```

kernel/src/task/schedule.rs
```diff
@@ -786,7 +797,7 @@ global_asm!(
 
         // Check to see whether the task is moving across CPUs.  If so, its
         // per-CPU page table state must be updated.
-        movq    {PERCPU_SHARED_OFFSET}(%rdx), %r8
+        movq    %gs:__svsm_percpu_shared(%rip), %r8
```
00xc, line 800: I think we could pass in to this asm block the offset of `PerCpuStorage::value`, and then do:
```
movq %gs:__svsm_percpu_shared+{PERCPU_STORAGE_VALUE_OFFSET}(%rip), %rdi
```

That, paired with the other suggestion I made for `ptr_for()` should allow removing the `repr(C)` from `PerCpuStorage` and all the implicit reliance on the layout.

kernel/src/task/schedule.rs
```diff
@@ -45,32 +45,49 @@ use super::{
...
+percpu! {
+    static CURRENT_STACK: RWLock<MemoryRegion<VirtAddr>>;
+}
+
+pub(crate) fn init_current_stack() {
+    assert!(
+        CURRENT_STACK
+            .init(RWLock::new(MemoryRegion::new(VirtAddr::null(), 0)))
+            .is_ok()
+    );
```
00xc:
```suggestion
    CURRENT_STACK
        .init(RWLock::new(MemoryRegion::new(VirtAddr::null(), 0)))
        .unwrap();
```

kernel/src/cpu/percpu/key.rs
```diff
@@ -0,0 +1,580 @@
...
+        ALIGNED_VALUE.with(|value| {
+            assert_eq!(ptr::from_ref(value) as usize % 64, 0);
+            assert_eq!(value.0, [0; 64]);
+        });
+    }
+
+    #[test]
+    fn test_runtime_initialization() {
+        assert!(!RUNTIME_VALUE.is_initialized());
+        assert!(RUNTIME_VALUE.init(AtomicUsize::new(11)).is_ok());
```
00xc:
```suggestion
        RUNTIME_VALUE.init(AtomicUsize::new(11)).unwrap();
```

kernel/src/cpu/smp.rs
```diff
@@ -8,22 +8,38 @@ use super::idt::load_static_idt;
...
 use core::arch::global_asm;
 use cpuarch::x86::EFERFlags;
 
+percpu! {
+    #[percpu_asm_symbol("__svsm_percpu_shared")]
+    static PERCPU_SHARED: &'static PerCpuShared;
+}
+
+pub(super) fn init_percpu_shared(percpu_shared: &'static PerCpuShared) {
+    assert!(PERCPU_SHARED.init(percpu_shared).is_ok());
```
00xc:
```suggestion
    PERCPU_SHARED.init(percpu_shared).unwrap();
```

kernel/src/cpu/tss.rs
```diff
@@ -28,6 +29,58 @@ pub struct X86Tss {
 
 pub const TSS_LIMIT: u64 = core::mem::size_of::<X86Tss>() as u64;
 
+percpu! {
+    static TSS: X86Tss = X86Tss::new();
+    static DOUBLE_FAULT_STACK: Option<VirtAddr>;
+}
+
+pub fn init_double_fault_stack(stack: Option<VirtAddr>) {
+    assert!(DOUBLE_FAULT_STACK.init(stack).is_ok());
```
00xc:
```suggestion
    DOUBLE_FAULT_STACK.init(stack).unwrap();
```

kernel/src/cpu/tss.rs
```diff
@@ -28,6 +29,58 @@ pub struct X86Tss {
...
+}
+
+pub fn double_fault_stack() -> Option<VirtAddr> {
+    DOUBLE_FAULT_STACK.with(|stack| *stack)
```
00xc: This pattern is present in several places. Maybe we could have a `get()` method when `T: Copy` and simply do e.g. `DOUBLE_FAULT_STACK.get()`.

kernel/src/sev/hv_doorbell.rs
```diff
@@ -51,14 +56,14 @@ pub struct HVExtIntInfo {
...
+pub fn allocate_hv_doorbell_page(ghcb: &GHCB) -> Result<*mut HVDoorbell, SvsmError> {
     let page = SharedBox::<HVDoorbell>::try_new_zeroed()?;
 
     let vaddr = page.addr();
     let paddr = virt_to_phys(vaddr);
     ghcb.register_hv_doorbell(paddr)?;
 
-    Ok(page)
+    Ok(page.leak().as_ptr())
 }
```
00xc, line 67: I don't love the fact that we need to revert to using pointers and introducing unsafe code. As far as I can tell this allows implementing `current_hv_doorbell()` without changing callers, but that breaks precisely the point of having a `PerCpuKey::with()` method.

We should be able to keep the `SharedBox` with not that many changes:

```diff
diff --git a/kernel/src/cpu/apic.rs b/kernel/src/cpu/apic.rs
index 5379f50f4878..5a05374b3fd8 100644
--- a/kernel/src/cpu/apic.rs
+++ b/kernel/src/cpu/apic.rs
@@ -16,7 +16,7 @@ use crate::platform::SVSM_PLATFORM;
 use crate::platform::guest_cpu::GuestCpuState;
 use crate::requests::SvsmCaa;
 use crate::sev::ghcb::with_current_ghcb;
-use crate::sev::hv_doorbell::{HVExtIntStatus, current_hv_doorbell};
+use crate::sev::hv_doorbell::{HVDoorbell, HVExtIntStatus, with_current_hv_doorbell};
 use crate::types::GUEST_VMPL;
 
 use core::ptr::NonNull;
@@ -344,7 +344,7 @@ impl LocalApic {
     ) {
         // Make sure any interrupts being presented by the host have been
         // consumed.
-        self.consume_host_interrupts();
+        with_current_hv_doorbell(|hv| self.consume_host_interrupts(hv));
 
         // Consume any pending IPIs.
         if cpu_shared.ipi_pending() {
@@ -795,8 +795,7 @@ impl LocalApic {
         }
     }
 
-    fn consume_host_interrupts(&mut self) {
-        let hv_doorbell = current_hv_doorbell();
+    fn consume_host_interrupts(&mut self, hv_doorbell: &HVDoorbell) {
         let vmpl_event_mask = hv_doorbell.per_vmpl_events.swap(0, Ordering::Relaxed);
         // Ignore events other than for the guest VMPL.
         if vmpl_event_mask & (1 << (GUEST_VMPL - 1)) == 0 {
@@ -878,8 +877,7 @@ impl LocalApic {
         }
     }
 
-    fn handoff_to_host(&mut self) {
-        let hv_doorbell = current_hv_doorbell();
+    fn handoff_to_host(&mut self, hv_doorbell: &HVDoorbell) {
         let descriptor = &hv_doorbell.per_vmpl[GUEST_VMPL - 1];
         // Establish the IRR as holding multiple vectors regardless of the
         // number of active vectors, as this makes transferring IRR state
@@ -945,7 +943,7 @@ impl LocalApic {
         }
 
         // Hand the current APIC state off to the host.
-        self.handoff_to_host();
+        with_current_hv_doorbell(|hv| self.handoff_to_host(hv));
 
         let _ = Self::clear_guest_eoi_pending(caa);
 
diff --git a/kernel/src/cpu/percpu/key.rs b/kernel/src/cpu/percpu/key.rs
index cb87e3f6b42a..01561df1fc4e 100644
--- a/kernel/src/cpu/percpu/key.rs
+++ b/kernel/src/cpu/percpu/key.rs
@@ -351,6 +351,20 @@ impl<T: Sync + 'static> PerCpuKey<T> {
         Ok(())
     }
 
+    pub fn try_with<F, R>(&'static self, f: F) -> Option<R>
+    where
+        F: FnOnce(&T) -> R,
+    {
+        let storage = self.storage();
+        if storage.state.load(Ordering::Acquire) != PERCPU_INITIALIZED {
+            return None;
+        }
+        // SAFETY: PerCpuStorage requires T: Sync and only shared references are
+        // handed to callers. The Acquire load above observed the initialized
+        // state, so the MaybeUninit contains a fully initialized value.
+        Some(f(unsafe { (*storage.value.get()).assume_init_ref() }))
+    }
+
     /// Borrow this CPU's value for the duration of `f`.
     ///
     /// # Panics
diff --git a/kernel/src/mm/page_visibility.rs b/kernel/src/mm/page_visibility.rs
index bd8cf0b71dc3..fcd7534e8039 100644
--- a/kernel/src/mm/page_visibility.rs
+++ b/kernel/src/mm/page_visibility.rs
@@ -105,6 +105,7 @@ pub unsafe fn make_page_private(vaddr: VirtAddr) -> Result<(), SvsmError> {
 }
 
 /// SharedBox is a safe wrapper around memory pages shared with the host.
+#[repr(transparent)]
 pub struct SharedBox<T> {
     ptr: NonNull<T>,
 }
diff --git a/kernel/src/platform/snp.rs b/kernel/src/platform/snp.rs
index 5552642eb32a..2f323097dcd5 100644
--- a/kernel/src/platform/snp.rs
+++ b/kernel/src/platform/snp.rs
@@ -39,7 +39,7 @@ use crate::mm::PerCPUPageMappingGuard;
 use crate::mm::memory::write_guest_memory_map;
 use crate::platform::IrqGuard;
 use crate::sev::ghcb::{GHCBIOSize, register_ghcb, setup_ghcb, with_current_ghcb};
-use crate::sev::hv_doorbell::{HVDoorbell, setup_hv_doorbell, try_current_hv_doorbell};
+use crate::sev::hv_doorbell::{HVDoorbell, current_hv_doorbell_ptr, setup_hv_doorbell};
 use crate::sev::msr_protocol::{
     GHCBHvFeatures, hypervisor_ghcb_features, request_termination_msr, verify_ghcb_version,
 };
@@ -56,7 +56,6 @@ use crate::utils::immut_after_init::ImmutAfterInitCell;
 use bootdefs::platform::SvsmPlatformType;
 use core::arch::x86_64::CpuidResult;
 use core::mem::MaybeUninit;
-use core::ptr;
 use core::sync::atomic::{AtomicU32, Ordering};
 use cpufeature::CpuidFeature;
 use cpufeature::backend::CpuidBackend;
@@ -128,11 +127,7 @@ impl SvsmPlatform for SnpPlatform {
     }
 
     fn idle_halt(&self, _guard: &IrqGuard) {
-        let hv_doorbell = try_current_hv_doorbell();
-        let ptr = match hv_doorbell {
-            Some(doorbell) => ptr::from_ref(doorbell),
-            None => ptr::null(),
-        };
+        let ptr = current_hv_doorbell_ptr();
         // SAFETY: The correct #HV doorbell address was calculated above.
         unsafe {
             snp_idle_halt(ptr);
diff --git a/kernel/src/sev/ghcb.rs b/kernel/src/sev/ghcb.rs
index 888c32129397..8914eb8addb2 100644
--- a/kernel/src/sev/ghcb.rs
+++ b/kernel/src/sev/ghcb.rs
@@ -14,7 +14,7 @@ use crate::mm::validate::{
 };
 use crate::mm::virt_to_phys;
 use crate::platform::PageStateChangeOp;
-use crate::sev::hv_doorbell::{HVDoorbell, try_current_hv_doorbell};
+use crate::sev::hv_doorbell::{HVDoorbell, current_hv_doorbell_ptr};
 use crate::sev::utils::raw_vmgexit;
 use crate::types::{Bytes, GUEST_VMPL, PAGE_SIZE_2M, PageSize};
 use crate::utils::MemoryRegion;
@@ -24,7 +24,6 @@ use core::arch::global_asm;
 use core::arch::x86_64::CpuidResult;
 use core::mem::{self, MaybeUninit, offset_of};
 use core::ops::Deref;
-use core::ptr;
 use core::sync::atomic::{AtomicU8, AtomicU16, AtomicU32, AtomicU64, Ordering};
 
 use super::msr_protocol::{invalidate_page_msr, register_ghcb_gpa_msr, validate_page_msr};
@@ -852,11 +851,7 @@ pub fn switch_to_vmpl(vmpl: u32) {
     // The switch to a lower VMPL must be done with an assembly sequence in
     // order to ensure that any #HV that occurs during the sequence will
     // correctly block the VMPL switch so that events can be processed.
-    let hv_doorbell = try_current_hv_doorbell();
-    let ptr = match hv_doorbell {
-        Some(doorbell) => ptr::from_ref(doorbell),
-        None => ptr::null(),
-    };
+    let ptr = current_hv_doorbell_ptr();
     // SAFETY: FFI call. Parameters and return values are checked.
     unsafe {
         if !switch_to_vmpl_unsafe(ptr, vmpl) {
diff --git a/kernel/src/sev/hv_doorbell.rs b/kernel/src/sev/hv_doorbell.rs
index c6a8af328d3a..16c359761db0 100644
--- a/kernel/src/sev/hv_doorbell.rs
+++ b/kernel/src/sev/hv_doorbell.rs
@@ -12,12 +12,12 @@ use crate::sev::ghcb::GHCB;
 use bitfield_struct::bitfield;
 use core::arch::asm;
 use core::ptr;
-use core::sync::atomic::{AtomicPtr, AtomicU8, AtomicU32, Ordering};
+use core::sync::atomic::{AtomicU8, AtomicU32, Ordering};
 use zerocopy::FromBytes;
 
 percpu! {
     #[percpu_asm_symbol("__svsm_percpu_hv_doorbell")]
-    static HV_DOORBELL: AtomicPtr<HVDoorbell> = AtomicPtr::new(ptr::null_mut());
+    static HV_DOORBELL: SharedBox<HVDoorbell>;
 }
 
 #[bitfield(u8)]
@@ -56,14 +56,14 @@ pub struct HVExtIntInfo {
 
 /// Allocates a new HV doorbell page and registers it on the hypervisor
 /// using the given GHCB.
-pub fn allocate_hv_doorbell_page(ghcb: &GHCB) -> Result<*mut HVDoorbell, SvsmError> {
+pub fn allocate_hv_doorbell_page(ghcb: &GHCB) -> Result<SharedBox<HVDoorbell>, SvsmError> {
     let page = SharedBox::<HVDoorbell>::try_new_zeroed()?;
 
     let vaddr = page.addr();
     let paddr = virt_to_phys(vaddr);
     ghcb.register_hv_doorbell(paddr)?;
 
-    Ok(page.leak().as_ptr())
+    Ok(page)
 }
 
 #[repr(C)]
@@ -167,34 +167,41 @@ impl HVDoorbell {
     }
 }
 
-/// Gets the HV doorbell page configured for this CPU.
+/// Calls the specified callback with the HV doorbell page configured for this CPU.
 ///
 /// # Panics
 ///
 /// Panics if te HV doorbell page has not been set up beforehand.
-pub fn current_hv_doorbell() -> &'static HVDoorbell {
-    try_current_hv_doorbell().expect("HV doorbell page dereferenced before allocating")
+pub fn with_current_hv_doorbell<F, R>(f: F) -> R
+where
+    F: FnOnce(&HVDoorbell) -> R,
+{
+    HV_DOORBELL.with(|ghcb| f(ghcb))
 }
 
-pub fn try_current_hv_doorbell() -> Option<&'static HVDoorbell> {
-    let ptr = HV_DOORBELL.with(|doorbell| doorbell.load(Ordering::Acquire));
-    let ptr = ptr::NonNull::new(ptr)?;
+/// Calls the specified callback with the HV doorbell page configured for this CPU,
+/// or returns `None` if the page has not been set up beforehand.
+pub fn try_with_current_hv_doorbell<F, R>(f: F) -> Option<R>
+where
+    F: FnOnce(&HVDoorbell) -> R,
+{
+    HV_DOORBELL.try_with(|ghcb| f(ghcb))
+}
 
-    // SAFETY: The pointer is installed only from a leaked SharedBox, which
-    // remains valid for the lifetime of the kernel.
-    unsafe { Some(ptr.as_ref()) }
+pub fn current_hv_doorbell_ptr() -> *const HVDoorbell {
+    try_with_current_hv_doorbell(ptr::from_ref).unwrap_or(ptr::null())
 }
 
 pub fn setup_hv_doorbell() -> Result<(), SvsmError> {
     let doorbell = crate::sev::ghcb::with_current_ghcb(allocate_hv_doorbell_page)?;
-    HV_DOORBELL.with(|current| current.store(doorbell, Ordering::Release));
+    HV_DOORBELL.init(doorbell).unwrap();
     Ok(())
 }
 
 pub fn process_hv_events_if_required() {
-    if let Some(doorbell) = try_current_hv_doorbell() {
-        crate::cpu::irq_state::with_irq_state(|irq_state| doorbell.process_if_required(irq_state));
-    }
+    try_with_current_hv_doorbell(|doorbell| {
+        crate::cpu::irq_state::with_irq_state(|irq_state| doorbell.process_if_required(irq_state))
+    });
 }
 
 /// # Safety
diff --git a/kernel/src/sev/snp_apic.rs b/kernel/src/sev/snp_apic.rs
index 192e31ecbee4..e1cad9d7e4ef 100644
--- a/kernel/src/sev/snp_apic.rs
+++ b/kernel/src/sev/snp_apic.rs
@@ -11,7 +11,7 @@ use crate::cpu::x86::x2apic::MSR_X2APIC_BASE;
 use crate::cpu::x86::{ApicAccess, MSR_APIC_BASE};
 use crate::error::SvsmError;
 use crate::sev::ghcb::with_current_ghcb;
-use crate::sev::hv_doorbell::current_hv_doorbell;
+use crate::sev::hv_doorbell::with_current_hv_doorbell;
 
 use core::sync::atomic::{AtomicBool, Ordering};
 
@@ -73,7 +73,7 @@ impl ApicAccess for GHCBApicAccessor {
 
     fn eoi(&self) {
         // Issue an explicit EOI unless no explicit EOI is required.
-        if !self.use_restr_inj() || !current_hv_doorbell().no_eoi_required() {
+        if !self.use_restr_inj() || !with_current_hv_doorbell(|hv| hv.no_eoi_required()) {
             self.apic_write(APIC_OFFSET_EOI, 0);
         }
     }
```

kernel/src/cpu/percpu/key.rs
```diff
@@ -0,0 +1,580 @@
...
+    /// # Returns
+    ///
+    /// `Ok(())` when the value was installed. If the value is already
+    /// initialized or initialization is in progress, returns `Err(value)`.
+    ///
+    /// # Panics
+    ///
+    /// On the SVSM target, this function may only be called after per-CPU
+    /// setup has installed the current CPU's `%GS` base.
+    pub fn init(&'static self, value: T) -> Result<(), T> {
```
00xc, line 329: Are there any users that require retrieving the `T` back? If not, I'd say we just return a `Result<(), SvsmError>` here, and then callers can simply use `unwrap()` or `?`.

kernel/src/cpu/percpu/key.rs
joergroedel: The main reason is the defined layout of this type, as you noted in the other comments. But there are also others:

- `ImmutAfterInitCell` must be thread-safe (thus required `T: Send + Sync`) while this type only needs to be re-entrant safe (thus only needs `T : Sync`).
- This type needs to support static initialization and does not need a `Drop` implementation.

kernel/src/task/schedule.rs
joergroedel: I doubt that this makes the code more readable. Also, always needing to pass this offset to assembly when it needs to access a percpu variable sounds annoying. I'd rather add a compile-time assertion to make sure that `value` is at offet `0`. Maybe this even works without the `#[repr(C)]`.

kernel/src/cpu/percpu/key.rs
joergroedel: I agree that a compile-time assertion is needed.

kernel/src/cpu/percpu/key.rs
joergroedel: All callers check only for `is_ok()`, so I think you are right.

kernel/src/task/schedule.rs
00xc: Without the `repr(C)` the compile-time assertion will spuriously fail I think. That's why I suggest allowing the compiler to use whatever offset it likes, and then just using that offset where relevant, which would be more robust against changes.

joergroedel: Makes sense. So my preference is to keep `#[repr(C)]` then and add the assertion. Or is there a problem with using C struct representation?

kernel/src/sev/hv_doorbell.rs
joergroedel: Yeah, the reason for tracking the pointer now as a per-cpu key is that it can be dereferenced directly in assembly code. Storing the pointer still in a `SharedBox` would not give the necessary guarantees to make this safe.

On the other side only using the pointer has the issue of more unsafe code and no automatic lifetime tracking. I will think about a better solution.

kernel/src/cpu/percpu/key.rs
joergroedel: Done.

kernel/src/cpu/percpu/key.rs
joergroedel: Done.

kernel/src/sev/hv_doorbell.rs
joergroedel: This one should be fixed. It required to make `SharedBox` use `#[repr(transparent)]`.

00xc: > This one should be fixed. It required to make `SharedBox` use `#[repr(transparent)]`.

Yes, that was part of the diff, maybe it was not too visible

PR #1221, opened 2026-09-15
kernel: mm/vm: support globally unique address allocations

00xc:
A generic rework of #1215.

## Background

A `VMR` manages a set of virtual address, their backing mappings, and the page table parts needed for those mappings to be visible.

Currently, in the SVSM codebase there are two ways in which we allocate virtual addresses:
1. Finding a suitable range of addresses in `VMR`'s RBTree. This is used e.g. to insert file mappings, and is more heavyweight.
2. Using a bitmap to allocate portions of a mapping already inserted in the `VMR`. The mapping in the `VMR` is not backed by any physical pages, and the bitmap is used to determine which subrange of virtual addresses may be used to install temporary mappings in the page table.

This PR reworks path number 1.

## Approach

The core idea is to make `VMR` generic over a `VmRange`:

```rust
pub struct Vmr<V: VmRange> {
    alloc: V::Allocator,
    pgtbl_parts: RWLock<Vec<PageTablePart>>,
}

pub trait VmRange {
    /// The region of the virtual address space covered by this range.
    const DESCRIPTOR: AddrSpaceDescriptor;

    /// Page-table flags applied to every mapping in this range.
    const PT_FLAGS: PTEntryFlags;

    /// Whether mappings in this range are per-CPU. This selects the TLB
    /// flushing behaviour used when mappings are removed.
    const PER_CPU: bool = false;

    /// The allocator that reserves and tracks the [`Mapping`]s of this range.
    type Allocator: VmAllocator;
}
```

This allows to reuse the `Vmr` implementation independently of the allocation strategy used.

Particularly, the address range for a task is globally unique within the global task address range. This can easily be implemented by adding a `VmAllocator` implementation that reaches into a global static, ensuring synchronization across all `VMR` instances.

Both the private and shared (i.e. global) virtual address allocators are lifted from #1215, modulo some renaming and integration with the new traits.


kernel/src/mm/vm/range.rs
```diff
@@ -267,231 +326,129 @@ impl VMR {
...
-        let mut tree = self.tree.lock_write();
-        let addr = base.pfn();
-
-        let mut cursor = tree.find_mut(&addr);
-        let node = cursor.remove().ok_or(SvsmError::Mem)?;
-        self.unmap_vmm(&node);
+    /// The removed mapping on success, `SvsmError::Mem` on error
+    pub fn remove(&self, base: VirtAddr) -> Result<Mapping, SvsmError> {
+        let mapping = self.alloc.free(base).ok_or(SvsmError::Mem)?;
+        self.unmap_mapping(base, &mapping);
```
joergroedel: The ordering is wrong with the relaxed locking (allocator locked separately). The address must be freed as the last action, after the unmapping and TLB flush happened to avoid address re-use while the TLB is still dirty.

kernel/src/mm/vm/shared.rs
```diff
@@ -0,0 +1,388 @@
...
+            self.shared.free(base);
+            return Err(error);
+        }
+        Ok(base)
+    }
+
+    fn free(&self, base: VirtAddr) -> Option<Mapping> {
+        let mapping = self.local.lock_write().remove(base.as_usize());
+        self.shared.free(base);
+        mapping
```
joergroedel: This needs a check for whether the removal from the local data structure succeeded before freeing the global one. Only then it is guaranteed that a given context can only free its own mappings. That problem is likely also in my PR, it requires the `free()` method of `UniqueVaAllocator` to return the `Option` from the `remove()` method.

kernel/src/mm/vm/range.rs
```diff
@@ -4,110 +4,191 @@
...
-    /// combination of [`PTEntryFlags::GLOBAL`] and [`PTEntryFlags::USER`].
-    pt_flags: PTEntryFlags,
+/// A statically-described virtual memory range.
+///
+/// A [`Vmr`] manages the mappings of a region of the virtual address space.
+/// The covered region, its page-table flags and its per-CPU property are
+/// provided by the [`VmRange`] type parameter, and mappings are reserved and
+/// tracked by its [`VmRange::Allocator`].
+#[derive(Debug)]
+pub struct Vmr<V: VmRange> {
```
joergroedel, line 178: I prefer to keep the name as `VMR`.

joergroedel (commented):
This PR shows that a single VMR implementation is possible. I still need to review the various struct/trait relationships. Currently I identified two gaps:

1) The split locking between page-tables and the allocators calls for problems in the future. There are already bugs due to that in this PR, which I pointed out. Going forward COCONUT needs a clear and hard to misuse locking scheme which takes the right ordering between unmaps, TLB flushes, page freeing, and address allocation into account.
2) The implementation as-is does not support mappings shared between all VMR instances bound to a single allocator. I added this to `ContextVMR`, mainly to support percpu-shared mappings in the future.

I will follow-up with my thoughts on the allocator/trait and API design.

kernel/src/mm/vm/range.rs
00xc: Right, this was done to not conflict during the migration steps from the old struct to the new struct. I can revert it to `VMR` on the same commit that the old version is removed if you want, although the diff size will increase.

00xc:
> This PR shows that a single VMR implementation is possible. I still need to review the various struct/trait relationships. Currently I identified two gaps:
> 
> 1. The split locking between page-tables and the allocators calls for problems in the future. There are already bugs due to that in this PR, which I pointed out. Going forward COCONUT needs a clear and hard to misuse locking scheme which takes the right ordering between unmaps, TLB flushes, page freeing, and address allocation into account.

~If the pattern is always the same (global lock -> local lock) we can make methods for the second instance (local) receive a lock guard on the first one (global) as a parameter. This would prove that both locks are held at the same time and in order.~

EDIT: I solved this in a different way, at least for the `free()` path.

> 2. The implementation as-is does not support mappings shared between all VMR instances bound to a single allocator. I added this to `ContextVMR`, mainly to support percpu-shared mappings in the future.

Perhaps I'm not understanding, but it does. All instances of `TaskVmAllocator` share a reference to `TASK_VM_ALLOCATOR` (see `TaskVmAllocator::new()`).

More shared allocators for other ranges can be implemented following the same pattern (e.g. introduce `PerCpuVmAllocator` wrapping `SharedVmAllocator`, and make them all point to the same static `GlobalVmAllocator`).

> I will follow-up with my thoughts on the allocator/trait and API design.

Great, thanks!

kernel/src/mm/vm/range.rs
00xc: Fixed this along the other issue you pointed out by making `VmAllocator::free()` take a callback. That way the teardown can happen with the right lock held.

kernel/src/mm/vm/shared.rs
00xc: Fixed. The method now grabs the local lock, removes the mapping, drops the lock, invokes a callback and then commits the removal to the global instance.

PR #1222, opened 2026-09-16
kernel: cpu/percpu: fix accesses to current page table

00xc:
Multiple users across the codebase update the page tables by doing `this_cpu().get_pgtable()`. While the method returns the initial page table for the current CPU, after switching to task context that is no longer the active page table.
    
Updating the page table via `get_pgtable()` only works when a) the CPU has not switched to task context yet, or b) when the virtual address being updated is part of the shared virtual range.
    
Make the updates more robust by adding `with_pgtable()` / `with_pgtable_mut()`, which run a callback with precisely the currently active page table. This is either the task's page table, after switching to task context, or the initial page table before that.

Add some cleanups for accessing the current task while we are at it.


kernel/src/cpu/percpu.rs
```diff
@@ -819,9 +808,9 @@ impl PerCpu {
         Ok(())
     }
 
-    fn finish_page_table(&self) {
-        let pgtable = self.get_pgtable();
-        self.vm_range.populate(pgtable);
+    fn install_pagetable(&self, mut pgtable: PageBox<PageTable>) {
```
nagyation, line 826: `set_pgtable` is public while this is private, which mean users can set but can't populate. We should pick one option and apply for both. private is better since it's not used outside of the module yet.

kernel/src/cpu/percpu.rs
```diff
@@ -739,7 +739,20 @@ impl PerCpu {
         Ok(())
     }
 
-    pub fn get_pgtable(&self) -> &'static mut PageTable {
+    /// Run callback `f` with the currently active page table.
+    pub fn with_pgtable<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut PageTable) -> R,
+    {
+        if read_cr3().bits() == self.cr3.load(Ordering::Relaxed) {
```
nagyation, line 749: I think we should make `with_pgtable` a private function and wrap it with a `pub fn with_current_pgtable` similar to `current_task` to avoid being used & failing this condition on remote cpu.

00xc: Not sure what you mean, I don't think there is a way currently to access a remote CPU to call this method.

00xc: Ah, I see what you mean now, you want to avoid the case I fixed where the BSP accesses the remote AP's page table, right.

nagyation: Yup! Just to avoid it happening later by design.

PR #1224, opened 2026-09-17
cargo: update Cargo.lock

luigix25:
Commit e44df296 ("kernel: remove unused rustc_version dependency") removed a dependency but did not update Cargo.lock accordingly.


00xc (approved):

stefano-garzarella:
mmmm, why our CI didn't catch it?

In `.github/workflows/rust.yml` we have `cargo update --workspace --locked`, so this should be caught by the CI, but IIUC we call it after `make clippy` that updated the Cargo.lock, so we should move that step just after the checkout. I'll prepare a PR to fix that

stefano-garzarella:
> mmmm, why our CI didn't catch it?
> 
> In `.github/workflows/rust.yml` we have `cargo update --workspace --locked`, so this should be caught by the CI, but IIUC we call it after `make clippy` that updated the Cargo.lock, so we should move that step just after the checkout. I'll prepare a PR to fix that

Fixed in #1226

stefano-garzarella (approved):

PR #1226, opened 2026-09-18
github/workflows: fix Cargo.lock check

stefano-garzarella:
CI recently failed to detect an outdated Cargo.lock (see #1224) because we run the check after other commands (e.g. `cargo clippy`) have silently regenerated the Cargo.lock file.

Add a `git restore Cargo.lock` step to restore the committed version before checking it.


stefano-garzarella:
The CI should fail now (as it should have done for #1212) , so this needs to be rebased after we merge #1224

luigix25 (approved):

stefano-garzarella:
v2:
- rebased since #1224 is now merged. CI should now be green as expected.

00xc (approved):

PR #1230, opened 2026-09-21
kernel/attest: run attestation in the CI

### LINKED ISSUE #773: attest: add more testing
#528 is now merged, so we should improve a bit our tests including testing the attestation protocol.

Discussing this with @tylerfanelli, the idea could be to use the `kbs-test` in our test infra to make sure everything is working.

luigix25:
Fixes: #773


scripts/test-in-svsm.sh
```diff
@@ -68,7 +88,80 @@ test_io(){
...
+    fi
+
+    $SVSM_DIR/bin/aproxy --protocol kbs \
+        --url "http://127.0.0.1:$KBS_PORT" --vsock &
+    APROXY_PID=$!
+
+    # aproxy binds its vsock socket during startup and the guest only connects
+    # once it has booted, so there is no need to poll for the socket here.
+    # Just catch the case where aproxy failed to bind and exited.
+    sleep 1
```
stefano-garzarella: I don't understand the comment, why `sleep 1` is better than checking if `aproxy` is listening to the port?

